PR #13184, opened 2023-02-18
feat(protocol): partially randomize prover reward

### LINKED ISSUE #13154: feat: Reward the provers with some random mechanism to promote decentralization

According to Taiko's current L1 contract, the earlier the proof is submitted, the higher the reward. And for each block, a maximum of 5 provers will be rewarded.

The potential problem is that, assuming I have a batch of the highest performing provers sufficient to submit the proof on time for all the proposed blocks, I will get all the rewards.  Other lower-performing provers will not have a chance to receive rewards, and they will quit forever. This is not necessarily desirable for decentralization of provers.

One possible solution is to provide a random mechanism for selecting provers, as described below: 
1. Set a time window in which all provers can submit proofs. For example, the number of forkchoices can be 100 or more. 
2. Randomly select 5 or more provers to grant a reward. 
3. Allow the provers of the same address to submit multiple proofs. A nonce value may be required to distinguish the proofs.

The advantages of this solution: 
1. Provers with lower performance have a chance to get a reward and they can join. This makes the network of provers more decentralized. 
2. The high-performing prover has a greater chance of receiving a reward because it can submit multiple times and is more easily selected.

The disadvantages of this solution: 
1. The average time to finalize the block might be some longer.

In fact, a similar mechanism is used in POW. We will not use POW here, but we can still introduce a random mechanism to achieve more decentralization.

### Additional context

Additional context here.

dantaik:
Introduce a new config: proverRewardRandomizedPercentage. If it is set to 0, then the proof rewards are distributed as before this PR, if it is 100, then the proof rewards are distributed randomly. We can choose a value between 0 and 100 so the reward are partially random and partially fixed.


packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -148,6 +148,41 @@ library LibVerifying {
...
+
+        uint256 sum;
+        uint256 i;
+
+        if (randomized > 0) {
+            unchecked {
+                // calculate the randomized weight
+                uint256 seed = block.prevrandao;
+                for (i = 0; i < numProvers; ++i) {
+                    weights[i] = uint16(seed * (1 + i));
```
dantaik, line 168:
```
weights[i] = uint16(seed * (1 + i));
```
Not sure if this is a good way to create a random uint16 for each prover.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -166,30 +208,27 @@ library LibVerifying {
...
-            }
+        uint256[] memory weights = getProverRewardBips(config, count);
 
-            if (tkoToken.balanceOf(fc.provers[start + i]) == 0) {
-                // Reduce reward to 1 wei as a penalty if the prover
-                // has 0 TKO balance. This allows the next prover reward
-                // to be fully paid.
-                proverReward = uint256(1);
+        for (uint256 i = 0; i < count; ++i) {
+            uint256 proverReward = (reward * weights[i]) / 10000;
```
davidtaikocha, line 222: Not sure if divide `10000` is correct here, since weight is `(a number less than one) * 100 * randomized + (another number less than one) * 100 * (100 - randomized)`. So `weights[i] / 10000` can not represent the actual  % of total rewards that the current prover will get? （Maybe we need to re-calculate the sum of weights here?）

packages/protocol/test/tokenomics/proofReward.test.ts
```diff
@@ -215,61 +215,61 @@ describe("tokenomics: proofReward", function () {
...
-                );
-                const previousProverBalance = await tkoTokenL1.balanceOf(
-                    forkChoice.provers[i - 1]
-                );
-
-                expect(previousProverBalance.gt(proverBalance)).to.be.eq(true);
-            }
-        }
-    });
+    // it(`asserts that with N provers, where N is config.maxProofsPerForkChoice all provers who submit proofs are paid with decreasing weight from the first prover to the Nth`, async function () {
```
dantaik, line 218: Disable this test as it fails on the main branch.

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: I think it is correct, David

davidtaikocha (dismissed):

dionysuzx (dismissed):

davidtaikocha (dismissed):

davidtaikocha (dismissed):

cyberhorsey (dismissed):

packages/protocol/contracts/L1/libs/LibVerifying.sol
cyberhorsey: i think it looks OK.

cyberhorsey (approved):

davidtaikocha (approved):

Brechtpd:
I have some doubts about the PoWification of proof generation.
- Rewards less efficient provers which could be good, but it seems inevitable that these will still end up being unprofitable because the proof price will adjust down until the price where the more efficient provers are just profitable. So seems likely to me that these provers will still not want to prove blocks because it will unlikely to be profitable for them.
- The above could be untrue if the protocol knows it overpays for the proof generation, and faster provers cannot take advantage of these higher price. But higher prover costs will increase transaction costs so not great for users and and a potential disadvantage compared to other rollups. But that may also not true because
- The protocol doesn't know if indeed different provers are submitting proofs. You would think that the most efficient prover submits all the proofs until it's not profitable anymore, and so less efficient provers which would be good for "decentralization" are still out-competed. But the extra transaction fees paid by the users actually did not help get multiple different provers to proof the block, so the users pay a higher transaction fee for no benefit at all.
- It actually seems good to me that inefficient provers are not profitable and will stop generating proofs? Does multiple provers proving the same block really benefit users?
- The PoW factor also seems to make uncle rewards unnecessary because they have the same goal with a different implementation. Which is actually a good thing because that means always only a single proof to check on-chain. 

I'm not sure it's possible to come up with a good mechanism that achieves the specified goal of actual multiple distinct provers proving the same block without knowing the identity of provers.

PR #13195, opened 2023-02-21
refactor(protocol): extract two anchor related functions

dantaik:


dionysuzx (dismissed):

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -386,6 +317,106 @@ library LibProving {
         });
     }
 
+    // TODO(brecht): can we write a circuit for this function?
```
Brechtpd, line 320: Yeah sure, doing it in the smart contract just easier and I thought would have been cheap enough to not make much of a difference, but it's more expensive than I expected so seems to be worth it do it in the circuit (well unless maybe the cheap precompile gets added to Ethereum to do this).

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -386,6 +317,106 @@ library LibProving {
...
+            !proofVerifier.verifyMKP({
+                key: LibRLPWriter.writeUint(0),
+                value: anchorReceipt,
+                proof: evidence.proofs[zkProofsPerBlock + 1],
+                root: evidence.header.receiptsRoot
+            })
+        ) revert L1_ANCHOR_RECEIPT_PROOF();
+    }
+
+    // TODO(brecht): can we write a circuit for this function?
```
Brechtpd, line 381: I don't think it will be worth it do this in the circuit. Hopefully we can do away with invalid blocks totally by having the proof on the proposer side.  But also because proposers will never propose invalid blocks anyway so it doesn't matter how expensive this is (as long as the gas cost is < Ethereum block gas limit).

davidtaikocha (dismissed):

PR #13196, opened 2023-02-21
refactor(protocol): hash public inputs into instance

dantaik:


davidtaikocha (approved):

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -285,6 +285,13 @@ library LibProving {
 
         if (!skipZKPVerification) {
             for (uint256 i = 0; i < config.zkProofsPerBlock; ++i) {
+                bytes32 instance = keccak256(
```
johntaiko, line 288: We should decide that `blockHash`, `prover` adn `txListHash` are enough to verify the proof.
As previous version, we need more information, like: history hashes, previous state root etc.

johntaiko: Maybe these fields are not necessary as @Brechtpd said

smtmfft (approved):

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: These are witness, we do not need them.

PR #13197, opened 2023-02-21
feat(protocol):  WIP: use signature ZKP and anchor ZKP (breaking)

dantaik:
This PR assumes signatures verification and anchor transaction verification can both be done in circuits. As you can see, the contract can be greatly simplified and gas cost can be lowered.

Tests are not fixed, please take a review and provide some feedback.

@Brechtpd @davidtaikocha @smtmfft


packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -84,6 +84,10 @@ contract TaikoL1 is
      *          will be the first transaction in the block -- if there are
      *          n transactions in `txList`, then there will be up to n+1
      *          transactions in the L2 block.
+     *
+     *        - inputs[2] a byte32 that determins which verifier to run.
+     *
+     *        - inputs[3] a ZKP that verifies all signatures.
```
Brechtpd, line 90: Maybe good to add here that it also validates the RLP encoding of the transaction (and also things like data compression) which would be the reason invalid blocks cannot happen anymore.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -118,6 +120,24 @@ library LibProposing {
...
+                resolver.resolve("proof_verifier", false)
+            );
+            bool verified = proofVerifier.verifyZKP({
+                verifierId: string(
+                    abi.encodePacked(
+                        "plonk_verifier_propose",
+                        bytes32(inputs[3])
+                    )
+                ),
+                zkproof: inputs[4],
```
Brechtpd, line 134:
```suggestion
                        bytes32(inputs[2])
                    )
                ),
                zkproof: inputs[3],
```
From the function comments above this seems to make more sense, but I may miss something.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -118,6 +120,24 @@ library LibProposing {
             );
         }
 
+        {
+            IProofVerifier proofVerifier = IProofVerifier(
+                resolver.resolve("proof_verifier", false)
+            );
+            bool verified = proofVerifier.verifyZKP({
```
Brechtpd, line 127: Hopefully we can postpone this proof verification here because quite expensive:
- Only if this is invalid, a "prover" can run this proof verification in the smart contract and prove that the block is invalid.
- Otherwise this proof is part of the main proof and is proven together in the aggregation circuit off-chain with the proofs of the other sub-circuits.

dantaik: Good point, will change it later.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -82,6 +82,13 @@ contract TaikoL1 is
      *          will be the first transaction in the block -- if there are
      *          n transactions in `txList`, then there will be up to n+1
      *          transactions in the L2 block.
+     *
+     *        - inputs[2] a byte32 that determins which verifier to run.
+     *
+     *        - inputs[3] a ZKP that verifies all transactions are valid -- they
```
davidtaikocha, line 88: So the reason why we don't verify the "sig zkp" in `proposeBlock` is because we want to reduce the cost of honest proposers? (not sure the actual cost of this new circuits) Since if we do so, then we can totally remove the `block id` concept (like `state.nextBlockId` / `state.latestVerifiedId`) in protocol, and only keep the `block height` concept (like `state.latestVerifiedHeight`) ? and the `proveBlockInvalid` mechanism can also be removed? which may make our protocol more simpler.

dantaik:
@Brechtpd please take another look at the PR.

packages/protocol/contracts/test/libs/TestLibProving.sol
```diff
@@ -232,20 +221,19 @@ library TestLibProving {
...
+                    verifierId: string(
+                        abi.encodePacked(
+                            "full_zkp_verifier_id_",
+                            i,
+                            "_prove_",
+                            evidence.circuits[i]
+                        )
+                    ),
+                    zkproof: evidence.proofs[i],
+                    instance: instance
```
Brechtpd, line 234: Seems like the sig proof isn't part of the circuit input here? Or will be current implementation verify that proof on-chain as well?

dantaik: > Seems like the sig proof isn't part of the circuit input here? Or will be current implementation verify that proof on-chain as well?

I am assuming the hash of the signature proof doesn't have to be part of the input of the main circuit and the main circuit always **duplicates** the verification of all signatures (and other validity rules). Unless you want to the main circuits to verify the signature proof itself instead of having the signature circuits built in. Let me know your preference, Brecht.

Brechtpd: Ah I see, I think that's a good way to do!

Brechtpd: Oh wait, there may be a problem with this. The proposer could submit an incorrect proof for a valid block. But then the prover has the option to either prove the block, or verify the incorrect proof and show that the block is invalid. This would make the chain nondeterministic.

dantaik: Then I guess the only way to fix it is to hash the signature zk-proof into the main circuit's input and recursively prove the signature proof.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -149,19 +153,18 @@ contract TaikoL1 is
...
-     *        - inputs[0] An Evidence object with various information regarding
-     *          the block to be proven and the actual proofs.
-     *        - inputs[1] The target block to be proven invalid.
-     *        - inputs[2] The receipt for the `invalidBlock` transaction
-     *          on L2. Note that the `invalidBlock` transaction is supposed to
-     *          be the only transaction in the L2 block.
+     *        - inputs[0] The target block's metadata.
+     *        - inputs[1] The circuit ID which is submitted by the block
+     *          proposer when the block is proposed.
+     *        - inputs[2] The signature zk-proof which is submitted by the
```
davidtaikocha, line 159: some not-in-depth thoughts: if we use a two-stage-challenge strategy for `proveBlockInvalid` since the proof verification is not cheap, can we use the same challenge strategy for `proveBlockValid`? Like there is a 5 mins challenge window for valid block proof? which may also make `proveBlockValid` a lot cheaper, since most of the provers are honest?

Brechtpd: Not sure I understand the challenge window part. But all proofs needs to be validated on-chain, and sooner is better. If the proof isn't' validated on-chain it's just an optimistic rollup with a validity proof as the fraud proof, so you don't get the benefits of a zk-rollup because the fraud proof delay needs to remain the same, only the downsides.

If I misunderstand and the idea is to batch prove multiple proofs that where first only the proof data is submitted on-chain and then another prover takes that data and proves that those proofs are valid then that would indeed be more efficient, at the cost of slower verified status on-chain. So you could see that batching working over block proofs, but you could see the proposer proofs as just any other block proof that could also get batch proven this way over multiple blocks, but in the end it's kind of the same I think. But this batching over multiple blocks is very likely the way to go in the future, proof generation time just needs to be fast enough.

davidtaikocha: Yah right what I mean is separating the proof submission and proof validation, at the cost of slower block status validation👍

Issue #258, opened 2022-11-11
feat(protocol): allow a prover to auction for right to prove a block exclusively


dantaik:
The idea is to allow a prover to auction for the right to prove a block exclusively with the following exceptions:

- if another prover submits the proof really early, e.g., earily than ½ of avgProofTime, it is still allowed.
- if this current auction winner failed to submit a valid proof before a deadline (1.25 * aveProofTime), then anyone can submit proofs.

The auction shall also be limited to a small window, e.g., ¼ * avgProofTime.

dantaik:
I'm closing the PR and mark this feature as low priority. One reason not to offer this feature is that we need to test our core protocol logics thoroughly before we add more features.

d1onys1us:
is this still a good-first-issue? i don't think so, so i think we should remove the label. cc @cyberhorsey

PR #13200, opened 2023-02-22
feat(protocol): allow prover to submit multiple proofs per block

### LINKED ISSUE #13154: feat: Reward the provers with some random mechanism to promote decentralization

dantaik:
This is a feature request from ZKPool. With a nonce for a proof, one single prover can reuse the same address to submit multiple proofs for the same block. This will save them from managing multiple addresses.


davidtaikocha (approved):

Brechtpd:
Kind of feels like admitting that the system doesn't necessarily attract actual different provers, but users still have to pay transaction fees to pay `n` times the actual proof generation cost + proof verification cost. Seems like inevitably the only prover benefiting from this is the most efficient prover which will still be the only prover that will be profitable because that one will run off with all the profitable rewards, anyone else will likely stop proving (which is fine, we only need a single proof). I guess the extra cost is not that big of a deal when proving costs are low, but as a type 1 zkEVM proving costs will already be higher than others, so if that cost is significant this would make it potentially `n` times more severe. PoW style reward mechanisms feels inefficient in a zkp system.

dantaik:
> 



> Kind of feels like admitting that the system doesn't necessarily attract actual different provers, but users still have to pay transaction fees to pay `n` times the actual proof generation cost + proof verification cost. Seems like inevitably the only prover benefiting from this is the most efficient prover which will still be the only prover that will be profitable because that one will run off with all the profitable rewards, anyone else will likely stop proving (which is fine, we only need a single proof). I guess the extra cost is not that big of a deal when proving costs are low, but as a type 1 zkEVM proving costs will already be higher than others, so if that cost is significant this would make it potentially `n` times more severe. PoW style reward mechanisms feels inefficient in a zkp system.

If we always reward the first/earlier provers more than others (or with a higher odds of doing so), then smart provers will try to prove as many blocks as possible first, then to prove blocks that have already been proven by themselves or others to get extra rewards. 

Mutiple proofs per block allows slower provers to have a chance while faster provers are working on new blocks.

dantaik:
I think now we are in favor of this feature https://github.com/taikoxyz/taiko-mono/issues/258 (draft impl https://github.com/taikoxyz/taiko-mono/pull/13205) in alpha-3, thus this PR doesn't have to be merged at all.

PR #13203, opened 2023-02-22
feat(bridge-ui): release token button

shadab-taiko:


RogerLamTd:
@shadab-taiko 
Screenshot 2023-02-23 at 1 49 53 PM
I'm having trouble looking at the bridge ui, not sure why

packages/bridge-ui/src/components/Transaction.svelte
```diff
@@ -109,6 +152,29 @@
...
+        if(token === ethers.constants.AddressZero && amount.eq(0)) {
+          transaction.status = MessageStatus.FailedReleased;
+        }
+      } else {
+        const srcBridgeContract = new ethers.Contract(
+          chains[transaction.fromChainId].bridgeAddress,
+          Bridge,
+          $providers.get(chains[transaction.message.srcChainId.toNumber()].id)
+        )
+        const {token, amount} = await srcBridgeContract.isEtherReleased(transaction.msgHash);
```
cyberhorsey, line 172: i think this only returns true or false?
```ts

    function isEtherReleased(bytes32 msgHash) public view returns (bool) {
        return _state.etherReleased[msgHash];
    }
    ```

File: packages/bridge-ui/src/proof/service.ts [hunk 2]
```diff
@@ -86,6 +77,60 @@ class ProofService implements Prover {
 
     return signalProof;
   }
+
+  async GenerateProof(opts: GenerateProofOpts): Promise<string> {
+    const key = ProofService.getKey(opts);
```
cyberhorsey, line 82: i think its more idiomatic to use `this.` and not `ProofService.`

packages/bridge-ui/src/proof/service.ts
```diff
@@ -86,6 +77,60 @@ class ProofService implements Prover {
...
+      opts.destBridgeAddress,
+      [key],
+      block.hash,
+    ]);
+
+    if (proof.storageProof[0].value !== "0x3") {
+      throw Error("invalid proof");
+    }
+
+    return ProofService.getSignalProof(proof, blockHeader);
```
cyberhorsey, line 132: `this.`

cyberhorsey (changes requested):

packages/bridge-ui/src/proof/service.ts
shadab-taiko: Since these methods don't change the `instance` and are rather mostly pure (with side-effects) I think it will be more appropriate to have them as `static` methods than `instance` methods, what do you think @cyberhorsey ?

shadab-taiko: Can also make them utils/helpers with better names

dantaik (approved):

cyberhorsey (approved):

PR #13204, opened 2023-02-22
feat(protocol): add isEtherReleased to Bridge

cyberhorsey:
BridgeUI needs a way to determine if ether is released without crawling every block for events. It already exists for ERC20 but not for ETH.


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -178,4 +178,8 @@ contract Bridge is EssentialContract, IBridge {
     ) public pure returns (bytes32) {
         return LibBridgeStatus.getMessageStatusSlot(msgHash);
     }
+
+    function isEtherReleased(bytes32 msgHash) public view returns (bool) {
```
dantaik, line 182: Please move this function before any pure function, otherwise running 'pnpm lint:sol' will have warnings.

dantaik (approved):

packages/protocol/contracts/bridge/Bridge.sol
dantaik: I edited it.

davidtaikocha (approved):

PR #13205, opened 2023-02-22
feat(protocol): Claiming Blocks / Single Prover

cyberhorsey:
This PR  removes multiple provers per fork choice and introduces a concept of `claim` before `proving`.

There is a claimWindow + a claimDelay set. After a block is proposed, proposers have `claimWindow` seconds to bid for the block. if `claimWindow` seconds passes, the highest bidder has exclusive rights to prove the block. If `claimWindow` seconds has not passed but there is a big and `claimDelay` seconds passes, the window ends early and proving can begin.

If the claimer does not prove a block in `proofWindow` time, anyoen can submit a proof, and the first one to do so wins half that proposer's deposit, and other half is burnt.. Otherwise, proposer gets their deposit back.


dantaik:
Also see this one: https://github.com/taikoxyz/taiko-mono/issues/258

dantaik:
> If the user doesnt submit proof in time, two options: either anyone can submit the proof, and get the reward PLUS that user's deposit (burning half), or we can allow another user to claim. (i like option A, I think option B could halt the protocol w/ spammers attacking us and losing deposit)

Lets go with option A but the auction should have an minimal price increase of 10% to settle quickly.


> If the user doesnt submit proof in time, they are added to a badClaimers list and the next time they try to claim a block, they have to pay deposit * timesTheyClaimedAndDidntSubmit as the base deposit instead.

I'd argument against this feature as a bad claimer will simply switch to user another address to get around of this deposit penalty and we end up creating some code that's not going to be used. Losing the deposit should be a sufficient penalty already.

dantaik:
> If the user doesnt submit proof in time, they are added to a badClaimers list and the next time they try to claim a block, they have to pay deposit * timesTheyClaimedAndDidntSubmit as the base deposit instead.

I'd argument against this feature as a bad claimer will simply switch to user another address to get around of this deposit penalty and we end up creating some code that's not going to be used. Losing the deposit should be a sufficient penalty already.

dantaik:
> Optional constraints:
> - a `claimGap` which says users can not stake claims on sequential proofs, to enforce decentrilization (ie: you can not do blockId 1, 2, 3, 4 in a row, you must have a gap of N blockIds between claims). User can still claim multiple blocks separately: say there are `2048` proposed blocks, you could claim `2048 / claimGap` amount of them yourself, and KNOW your proof will given a reward.

I'm not a fan of this idea with the same reason: they can just get around of this constraint by using multiple addresses.

cyberhorsey:
> > If the user doesnt submit proof in time, they are added to a badClaimers list and the next time they try to claim a block, they have to pay deposit * timesTheyClaimedAndDidntSubmit as the base deposit instead.
> 
> I'd argument against this feature as a bad claimer will simply switch to user another address to get around of this deposit penalty and we end up creating some code that's not going to be used. Losing the deposit should be a sufficient penalty already.

I agree, will remove it

cyberhorsey:
> > If the user doesnt submit proof in time, two options: either anyone can submit the proof, and get the reward PLUS that user's deposit (burning half), or we can allow another user to claim. (i like option A, I think option B could halt the protocol w/ spammers attacking us and losing deposit)
> 
> Lets go with option A but the auction should have an minimal price increase of 10% to settle quickly.
> 
> There can also be two variables here, one is the action period, for example, 5 minutes. The other one is the auction delay (1 minutes) after which the best bidder, if any, will win and the auction stops even before 5 minutes time out.

Sounds good, I will add the delay timer

RogerLamTd:
> ... the highest bidder has exclusive rights to prove the block. If `claimWindow` seconds has not passed but there is a big and `claimDelay` seconds passes, the window ends early and proving can begin.

@cyberhorsey  "there is a big" do you mean bid? Also, are proposers bidding for the right to prove or are provers (or are they not mutually exclusive)? Is the idea instead that the proposers get priority in bidding?

cyberhorsey:
> 

A proposer can be a prover. Its simply an address bidding for the right to prove, which would then make you a prover.

I opened a new PR since Daniel did the single prover thing in a diff PR

PR #13208, opened 2023-02-23
feat(protocol): update `PlonkVerifier` to accept new public inputs

### LINKED ISSUE #393: feat(protocol): implement ZKP verifier

davidtaikocha:


johntaiko:
**Attention: It's not the final version, don't approve**

Because the contract of verifier is too large and more than 24kb, so I will optimize the size in next version

davidtaikocha:
> **Attention: It's not the final version, don't approve**
> 
> Because the contract of verifier is too large and more than 24kb, so I will optimize the size in next version

Yah after the verification contract bytecode size issue fixed, i will change this PR's status to "Ready for review".

packages/protocol/contracts/libs/LibBlockHeader.sol
```diff
@@ -34,7 +34,15 @@ library LibBlockHeader {
...
         BlockHeader memory header
     ) internal pure returns (bytes32) {
-        bytes[] memory list;
+        bytes memory rlpHeader = LibRLPWriter.writeList(
+            getBlockHeaderRLPItemsList(header)
+        );
+        return keccak256(rlpHeader);
+    }
+
+    function getBlockHeaderRLPItemsList(
```
dantaik, line 43: make it private and rename to `_getBlockHeaderRLPItemsList`?

dantaik (dismissed):

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -459,4 +451,24 @@ library LibProving {
             header.mixHash != meta.mixHash
         ) revert L1_META_MISMATCH();
     }
+
+    function _getInstance(
+        Evidence memory evidence
+    ) internal pure returns (bytes32 instance) {
```
dantaik, line 457: @smtmfft @davidtaikocha I'm wonder, if the circuits really need all inputs to be RLP encoded first, won't packing different bytes32 together be more efficient?

johntaiko: I think there are two reasons:
1. RLP encoding can be used for more complex data structures, such as list, nested list, etc. (Assuming pi will change in the future
2. The main reason ABI encoding and RLP encoding will require two different circuits, but they actually do the same thing. RLP is more versatile than ABI with a little overhead. More circuits make optimization difficult.

packages/protocol/contracts/libs/LibBlockHeader.sol
davidtaikocha: we still need it to be `internal` since Its also used by `LibProving._getInstance`?

dantaik (approved):

johntaiko (approved):

PR #13214, opened 2023-02-26
feat(protocol): sync signal service storage roots

dantaik:
@Brechtpd @cyberhorsey 

This is a very immature PR that shows the idea of synchronising not only block hash across L1/L2, but also the storageRoot of the signal service.

### The benefits
- Now the signal proof can be shorter as the account proof can be skipped entirely.

### The downsides
- One extra bytes32 needed to prove a block ,and one extra SSTORE needed when a block is proven; one extra SSTORE when one or multiple blocks are proven.
- The new anchor circuit will also need to prove the signal service storage root based on the parent block's block hash.

### Assumptions
I assume the following code can read the storage roof of an account -- from ChatGPT:):
```solidity
function getStorageRoot() public view returns (bytes32 root) {
    assembly {
        root := sload(0)
    }
}
```

@Brechtpd @cyberhorsey please let me know what you think.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -70,7 +70,7 @@ library TaikoData {
 
     // 3 + n slots
     struct ForkChoice {
-        bytes32 blockHash;
+        IHeaderSync.SyncData l2SyncData;
         uint64 provenAt;
         address[] provers;
     }
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -6,7 +6,8 @@
 
 pragma solidity ^0.8.18;
 
-import {IHeaderSync} from "../common/IHeaderSync.sol";
+import {SyncData} from "../common/IHeaderSync.sol";
+import {BlockHeader} from "../libs/LibBlockHeader.sol";
 
 library TaikoData {
     struct Config {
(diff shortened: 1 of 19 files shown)


Brechtpd:
I think this could be a good idea! The only thing I'm wondering if it's worth the extra complexity and what the gas savings will be.

> I assume the following code can read the storage roof of an account -- from ChatGPT:):
> 
> ```solidity
> function getStorageRoot() public view returns (bytes32 root) {
>     assembly {
>         root := sload(0)
>     }
> }
> ```

I'm pretty sure this will just read the data stored at address 0 in the storage tree, so this won't return the account storage root. I'm not aware of any way to access that account storage root accept going through the full MPT manually.

So for L1 -> L2 the most optimal way would be to add a MPT check in the circuits so only the correct storage root can be inserted in the L2 contract. This is a bit tricky however because the storage root isn't known on L1 so we cannot prepare the anchor transaction data directly so will need to handle this special case where the anchor transaction contains some unknown data that will have to be replaced on the fly.

For L2 -> L1 just an extra MPT check + add the value to the public input should work, so should be easy.

For gas it's nice that the bridge proofs are always the same so less complexity there. But the overall gas used may actually go up as the account storage proof now needs to be stored after each proof (which in the best case could be a couple of blocks). On the bridge side the gas cost to cache the account storage root is only paid when somebody that is bridging actually wants to execute his bridge transaction and pay the extra gas cost, with everybody else just waiting on the cached value to be there to reduce costs. So in theory the caching of the account storage root hash in the bridging contract could be very infrequent and only done for very few blocks.

Maybe one other thing to consider is that this bridging mechanism, and so this smart contract, would now be part of the core protocol. I think that's fine as the smart contract is very simple and doesn't really do anything.

I guess for future proofing, two things to keep track, depending on how and when Ethereum will change (and of course these may take very long to be added, so probably don't have to take these too much into account):
- There may be a precompile to verify these Merkle proofs (I can't find the EIP unfortunately, so not sure what the API will be): make sure that this change remains compatible with that if possible (otherwise the data savings gains may not offset the computational savings).
- Probably very little gas savings with Verkle trees, but who knows when this will be added

PR #13216, opened 2023-02-26
refactor(protocol): simplify protocol for single prover/proof

dantaik:
This is an attempt to simply the protocol design by:

- [X] Allowing only one zk-proof per block fork-choice (besides the oracle proof, if enabled). This is also a preparation for a future release that enables proof-auction.
- [X] Removing halt because of above
- [X] Oracle prover is enabled if "oracle_prover" is not address(0), no flag needed.
- [X]  Inlining a few small functions
- [X] Do not mint TKO reward to provers when blocks are verified, instead, we keep their outstanding balances and provers can withdraw later.
- [x] Fixing tests

I would even suggest that we should get this version tested and merged ASAP, maybe it can make into the alpha-2 testnet release.

**Please review carefully**

### This branch

```
·-------------------------------------------|---------------------------|---------------|----------------------------·
|           Solc version: 0.8.18            ·  Optimizer enabled: true  ·  Runs: 10000  ·  Block limit: 6718946 gas  │
············································|···························|···············|·····························
|  Methods                                                                                                           │
··························|·················|·············|·············|···············|··············|··············
|  Contract               ·  Method         ·  Min        ·  Max        ·  Avg          ·  # calls     ·  usd (avg)  │
··························|·················|·············|·············|···············|··············|··············
|  TestTaikoL1            ·  proposeBlock   ·      89859  ·     124071  ·       117964  ·          23  ·          -  │
··························|·················|·············|·············|···············|··············|··············
|  TestTaikoL1            ·  proveBlock     ·     425690  ·     425762  ·       425741  ·          11  ·          -  │
··························|·················|·············|·············|···············|··············|··············
|  TestTaikoL1            ·  verifyBlocks   ·      54302  ·      88172  ·        65094  ·          11  ·          -  │
··························|·················|·············|·············|···············|··············|··············
```

### main

```
·-------------------------------------------|---------------------------|---------------|----------------------------·
|           Solc version: 0.8.18            ·  Optimizer enabled: true  ·  Runs: 10000  ·  Block limit: 6718946 gas  │
············································|···························|···············|·····························
|  Methods                                                                                                           │
··························|·················|·············|·············|···············|··············|··············
|  Contract               ·  Method         ·  Min        ·  Max        ·  Avg          ·  # calls     ·  usd (avg)  │
··························|·················|·············|·············|···············|··············|··············
|  TestTaikoL1            ·  proposeBlock   ·      93401  ·     127637  ·       122756  ·          29  ·          -  │
··························|·················|·············|·············|···············|··············|··············
|  TestTaikoL1            ·  proveBlock     ·     468502  ·     468766  ·       468565  ·          11  ·          -  │
··························|·················|·············|·············|···············|··············|··············
|  TestTaikoL1            ·  verifyBlocks   ·      65026  ·      93158  ·        73942  ·          11  ·          -  │

```


packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -272,8 +159,29 @@ library LibVerifying {
...
+                // reward the prover
+                if (reward != 0) {
+                    uint256 _reward = reward;
+                    if (taikoToken.balanceOf(fc.prover) == 0) {
+                        // Reduce reward to 1 wei as a penalty if the prover
+                        // has 0 TKO balance. This allows the next prover reward
+                        // to be fully paid.
+                        _reward = uint256(1);
+                    }
+                    taikoToken.mint(fc.prover, _reward);
```
Brechtpd, line 171: May as well do only rewards if `balance > 0`, because not the max gas consumption is still much higher in the worst case.

Perhaps we shouldn't even do the transfers as part of the block verification if an ERC20 is used because quite expensive! (I don't know if it would be worthwhile to make it a pull API where anyone could claim the rewards themselves, depends I guess how much state is need to make this possible, if ETH is used it is probably cheaper to just transfer here with are small gas limit).

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -221,14 +209,6 @@ contract TaikoL1 is
         });
     }
 
-    /**
-     * Check if the L1 is halted.
-     * @return True if halted, false otherwise.
-     */
-    function isHalted() public view returns (bool) {
```
Brechtpd, line 228: Halting based some scenario's still a good idea I think if it doesn't add too much complexity to the protocol, but probably good to remove for the testnet anyway because it could be abused.

cyberhorsey:
I haven't had time to review this but I think a lot of it is duplicated from my other PR? But I guess this one can be merged for alpha-2.

packages/protocol/contracts/L1/TaikoL1.sol
dantaik: Agreed, but lets add that feature later. I figured it's a mistake to complicate the design/impl before a simpler version has been well tested.

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: Changed as suggested

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -125,6 +108,20 @@ library LibVerifying {
...
     }
 
+    function withdrawBalance(
+        TaikoData.State storage state,
+        AddressResolver resolver
+    ) public {
+        uint256 balance = state.balances[msg.sender];
+        if (balance == 0) return;
+
+        state.balances[msg.sender] = 0;
```
davidtaikocha, line 118: Shall we allow users to leave 1 TKO in their balance when withdrawing? Otherwise after a withdraw, their next reward will be reduced to 1 TKO, is this kind of penalty what we want?

dantaik: Good point, I'll change it a bit.

davidtaikocha (dismissed):

cyberhorsey:
@dantaik fixed the test you were talking about, we also can remove the `multiple provers` test as well now from `proofReward.test.ts` so I did that. However you have a failing test now in `tokenomics` with regards to your `withdrawBalance` function.

davidtaikocha (dismissed):

davidtaikocha (approved):

Brechtpd (dismissed):

packages/protocol/contracts/libs/LibSharedConfig.sol
```diff
@@ -55,12 +51,10 @@ library LibSharedConfig {
                 blockTimeCap: 60 seconds, // owner:daniel
                 proofTimeCap: 30 minutes, // owner:daniel
                 bootstrapDiscountHalvingPeriod: 30 days, // owner:daniel
-                initialUncleDelay: 10 minutes,
                 proverRewardRandomizedPercentage: 20, // owner: daniel
```
davidtaikocha, line 54:
```suggestion
```

davidtaikocha (dismissed):

davidtaikocha (approved):

johntaiko (approved):

Issue #13057, opened 2023-01-25
feat(protocol): Change require strings to custom errors


cyberhorsey:
### Describe the feature request

Custom Errors are significantly lower on gas and bytes, and would reduce the size and gas costs of our deployments and contract calls by quite a bit.

### Describe alternatives you've considered

Description of the alternatives you've considered here.

### Additional context

Additional context here.

cyberhorsey:
@dantaik do you agree?

dantaik:
[This PR](https://github.com/taikoxyz/taiko-mono/pull/13086) is the first one to use custom errors. We need another one for Bridge related contracts. @RogerLamTd could you please review the first PR and continue with the second oen?

dantaik:
> [This PR](https://github.com/taikoxyz/taiko-mono/pull/13086) is the first one to use custom errors. We need another one for Bridge related contracts. @RogerLamTd could you please review the first PR and continue with the second oen?

The bridge part is also done in my PR.

dantaik:
@cyberhorsey ,with [this PR](https://github.com/taikoxyz/taiko-mono/pull/13143), maybe we can quickly change all string-based errors to custom errors.

dantaik:
With https://github.com/taikoxyz/taiko-mono/pull/13146, the outstanding work is changing Bridge contracts to use custom errors.

dantaik:
Update: now only the bridge related files have string-based error messages.

PR #13220, opened 2023-02-27
feat(protocol): Change require to custom err in bridge contracts

adaki2004:
As per issue:
https://github.com/taikoxyz/taiko-mono/issues/13057

Bridge related require messages are changed in code + unit tests.

Created a gas usage comparison between current version vs. new. Please see differences in yellow (cheaper) or purple (expensive). Left is the new, right is the old.
Képernyőfotó 2023-02-27 - 9 35 07

It might only be slightly more expensive because only called 4 times.


packages/protocol/contracts/bridge/BridgeCustomErrors.sol
```diff
@@ -0,0 +1,17 @@
...
+//  _____     _ _         _         _
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+
+pragma solidity ^0.8.18;
+
+abstract contract BridgeCustomErrors {
+    // Libraries having custom errors in place
+    error B_CANNOT_RECEIVE();
```
dantaik, line 11: sort these errors?

packages/protocol/contracts/bridge/BridgeCustomErrors.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+//  _____     _ _         _         _
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+
+pragma solidity ^0.8.18;
+
+abstract contract BridgeCustomErrors {
+    // Libraries having custom errors in place
```
davidtaikocha, line 10: I believe the reason why we want to define a `BridgeCustomErrors` is that we want all custom errors which may be encountered when calling `Bridge` contract to be a part of the `Bridge` contract's compiled ABI, which will be helpful for the clients that only import the `Bridge` ABI to parse those errors. so i think we should also add those custom errors defined in `LibBridgeXXX` to here?

adaki2004: Make sense and done.

davidtaikocha (dismissed):

packages/protocol/contracts/bridge/BridgeCustomErrors.sol
adaki2004: Done.

cyberhorsey:
LGTM just needs merge conflict resolved!

davidtaikocha (dismissed):

adaki2004:
Integration tests has to be adjusted.

cyberhorsey (dismissed):

davidtaikocha (dismissed):

dantaik (approved):

cyberhorsey (approved):

PR #13230, opened 2023-02-28
feat(protocol): auction for block proving rights

cyberhorsey:
There is a claimWindow + a claimDelay set. After a block is proposed, bidders have `claimWindow` seconds to bid for the block. if `claimWindow` seconds passes, the highest bidder has exclusive rights to prove the block. If `claimWindow` seconds has not passed but there is a bid and `claimDelay` seconds passes, the window ends early and proving can begin.

If the claimer does not prove a block in `proofWindow` time, anyone can submit a proof, and the first one to do so wins half that claimer's deposit, and other half is burnt.. Otherwise, claimer gets their deposit back.


RogerLamTd:
> If the claimer does not prove a block in `proofWindow` time, anyone can submit a proof, and the first one to do so wins half that proposer's deposit, and other half is burnt.. Otherwise, proposer gets their deposit back.

@cyberhorsey 
Would you say that this implementation encourages proposers to prove their own blocks so they essentially spend only half the deposit to propose a block?
-  I guess not necessarily because they would also have to bid for proving rights
-  Unless I'm misunderstanding and the "proposer" in question is the claimer?

RogerLamTd:
> > > If the claimer does not prove a block in `proofWindow` time, anyone can submit a proof, and the first one to do so wins half that proposer's deposit, and other half is burnt.. Otherwise, proposer gets their deposit back.
> > 
> > 
> > @cyberhorsey Would you say that this implementation encourages proposers to prove their own blocks so they essentially spend only half the deposit to propose a block?
> > 
> > * I guess not necessarily because they would also have to bid for proving rights
> > * Unless I'm misunderstanding and the "proposer" in question is the claimer?
> > 
> > Would [this ](https://github.com/taikoxyz/taiko-mono/blob/9a834c6ad960b637cd4b112d2fb93e832ba4268b/packages/protocol/contracts/L1/TaikoL1.sol#L229-L231) line not make it possible for a proposer to bid unrealistically high amounts for proving rights if they can guarantee they can prove the block in proofWindow at essentially no cost (except proving computation)?
> > 
> > * and if the token reward is equal or greater than the value of 1/2 proposer deposit, could an individual theoretically keep proposing and proving cyclically at almost no cost (or even at a profit)?
> 
> Sorry it should say claimer not proposer there. It is unrelated to proposing fee

Ok that makes sense! I guess my second question still stands then, if the claimer makes back their deposit and then a token reward on top of it, would the described situation of cyclical proposing/claiming/proving at a profit be possible? If so is that behaviour we want to encourage?

cyberhorsey:
> > > > If the claimer does not prove a block in `proofWindow` time, anyone can submit a proof, and the first one to do so wins half that proposer's deposit, and other half is burnt.. Otherwise, proposer gets their deposit back.
> > > 
> > > 
> > > @cyberhorsey Would you say that this implementation encourages proposers to prove their own blocks so they essentially spend only half the deposit to propose a block?
> > > 
> > > * I guess not necessarily because they would also have to bid for proving rights
> > > * Unless I'm misunderstanding and the "proposer" in question is the claimer?
> > > 
> > > Would [this ](https://github.com/taikoxyz/taiko-mono/blob/9a834c6ad960b637cd4b112d2fb93e832ba4268b/packages/protocol/contracts/L1/TaikoL1.sol#L229-L231) line not make it possible for a proposer to bid unrealistically high amounts for proving rights if they can guarantee they can prove the block in proofWindow at essentially no cost (except proving computation)?
> > > 
> > > * and if the token reward is equal or greater than the value of 1/2 proposer deposit, could an individual theoretically keep proposing and proving cyclically at almost no cost (or even at a profit)?
> > 
> > 
> > Sorry it should say claimer not proposer there. It is unrelated to proposing fee
> 
> Ok that makes sense! I guess my second question still stands then, if the claimer makes back their deposit and then a token reward on top of it, would the described situation of cyclical proposing/claiming/proving at a profit be possible? If so is that behaviour we want to encourage? (and if not, how do we discourage it/ is it already being discouraged by the tokenomics of the token reward)

I'm not sure what the question is, can you explain? Nothing stops anyone from being both a proposer and a prover with the current design either?

RogerLamTd:
> > > > > If the claimer does not prove a block in `proofWindow` time, anyone can submit a proof, and the first one to do so wins half that proposer's deposit, and other half is burnt.. Otherwise, proposer gets their deposit back.
> > > > 
> > > > 
> > > > @cyberhorsey Would you say that this implementation encourages proposers to prove their own blocks so they essentially spend only half the deposit to propose a block?
> > > > 
> > > > * I guess not necessarily because they would also have to bid for proving rights
> > > > * Unless I'm misunderstanding and the "proposer" in question is the claimer?
> > > > 
> > > > Would [this ](https://github.com/taikoxyz/taiko-mono/blob/9a834c6ad960b637cd4b112d2fb93e832ba4268b/packages/protocol/contracts/L1/TaikoL1.sol#L229-L231) line not make it possible for a proposer to bid unrealistically high amounts for proving rights if they can guarantee they can prove the block in proofWindow at essentially no cost (except proving computation)?
> > > > 
> > > > * and if the token reward is equal or greater than the value of 1/2 proposer deposit, could an individual theoretically keep proposing and proving cyclically at almost no cost (or even at a profit)?
> > > 
> > > 
> > > Sorry it should say claimer not proposer there. It is unrelated to proposing fee
> > 
> > 
> > Ok that makes sense! I guess my second question still stands then, if the claimer makes back their deposit and then a token reward on top of it, would the described situation of cyclical proposing/claiming/proving at a profit be possible? If so is that behaviour we want to encourage? (and if not, how do we discourage it/ is it already being discouraged by the tokenomics of the token reward)
> 
> I'm not sure what the question is, can you explain? Nothing stops anyone from being both a proposer and a prover with the current design either?

Sorry I kinda made it too convoluted, basically my question is whether someone being both a proposer and prover and outcompeting other bidders at no cost (if they can 100% prove in proofWindow) such that they propose and prove their own blocks with no interference could be a problem/attack vector

cyberhorsey:
> 

It is intended for neither the proposer nor prover to "cost", and both should be profitable. Proposers get the fees from the transactions but pay a proposing fee and deposit to do so. Provers pay a deposit to be able to prove a block. 

If they were both not profitable, nobody would want to do either one!  
  
Regarding "no interference", well, that is a decentralized network IMO. If one guy chooses to out deposit everyone, is that much different than the largest ETH whale shaking his big stack and reaping the rewards for it?  
  
The current design before this PR also has the same issues - except it's just "fastest prover wins".

RogerLamTd:
> > 
> 
> It is intended for neither the proposer nor prover to "cost", and both should be profitable. Proposers get the fees from the transactions but pay a proposing fee and deposit to do so. Provers pay a deposit to be able to prove a block.
> 
> If they were both not profitable, nobody would want to do either one!
> 
> Regarding "no interference", well, that is a decentralized network IMO. If one guy chooses to out deposit everyone, and nobody else steps up, is that much different than the largest ETH whale shaking his big stack and reaping the rewards for it? It's a good question.
> 
> The current design before this PR also has the same issues - except it's just "fastest prover wins".

Ah ok gotcha, so it's more just part of what it means to be a decentralized network. Thanks for the explanation!

packages/protocol/contracts/L1/libs/LibClaiming.sol
```diff
@@ -0,0 +1,119 @@
...
+        uint256 baseDepositRequired = config.baseClaimDepositInWei;
+        if (state.timesProofNotDeliveredForClaim[tx.origin] > 0) {
+            baseDepositRequired =
+                config.baseClaimDepositInWei *
+                (state.timesProofNotDeliveredForClaim[tx.origin] + 1);
+        }
+
+        // if user hasnt sent enough to meet their personal base deposit amount
+        // we dont allow them to claim.
+        if (msg.value < baseDepositRequired) {
```
davidtaikocha, line 86: Actually I'm not sure that whether we should use TKO token here, to make this auction a part of its utility?

packages/protocol/contracts/L1/libs/LibClaiming.sol
```diff
@@ -0,0 +1,119 @@
...
+                state.proposedBlocks[blockId % config.maxNumBlocks].proposedAt >
+            config.claimAuctionWindowInSeconds
+        ) {
+            revert L1_CLAIM_AUCTION_WINDOW_PASSED();
+        }
+
+        // if user has claimed and not proven a block before, we multiply
+        // requiredDeposit by the number of times they have falsely claimed.
+        uint256 baseDepositRequired = config.baseClaimDepositInWei;
+        if (state.timesProofNotDeliveredForClaim[tx.origin] > 0) {
```
davidtaikocha, line 78: shall we use `msg.sender` instead of `tx.origin` here? since other part of our protocol allows proposer / prover to be `msg.sender`?

packages/protocol/contracts/L1/libs/LibClaiming.sol
```diff
@@ -0,0 +1,119 @@
...
+        if (currentClaim.claimer != address(0)) {
+            if (
+                msg.value <
+                currentClaim.deposit + config.minimumClaimBidIncreaseInWei
+            ) {
+                revert L1_INVALID_CLAIM_DEPOSIT();
+            } else {
+                // otherwise we have a new high bid, and can refund the previous claimer
+                // refund the previous claimer
+                currentClaim.claimer.sendEther(currentClaim.deposit);
```
davidtaikocha, line 100: And if we change to use `msg.sender` to record claimers, maybe we need to`try-catch` this `sendEther` call here, otherwise, if the last claimer write some code like `require(true == false)` in its `receive()`, then he will never be outbidded. But its fine if we use `tx.origin` here.

packages/protocol/contracts/L1/libs/LibClaiming.sol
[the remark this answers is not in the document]
cyberhorsey: I think you should be able to bid as much as you want, but i am open the conversation!
Should we only allow bids == baseDepositRequired?

packages/protocol/contracts/L1/libs/LibClaiming.sol
cyberhorsey: done

packages/protocol/contracts/L1/libs/LibClaiming.sol
cyberhorsey: We can't try catch internal methods and don't want to make libaddress external as will increase gas cost elsewhere, I will change to do a regular .call and ignore return value, allow to fail. Is this OK?

packages/protocol/contracts/L1/libs/LibClaiming.sol
cyberhorsey: @dantaik @adaki2004 what do you guys think about using the TKO token here? Maybe we can move the actual deposit transfer outside of the claim function somehow to save gas if we do that?

packages/protocol/contracts/L1/libs/LibClaiming.sol
davidtaikocha: yah i think it's actually what we did in [LibBridgeRelease](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/bridge/libs/LibBridgeRelease.sol#L76)

packages/protocol/contracts/L1/libs/LibClaiming.sol
adaki2004: IMO, this looks good as is and can save some gas elsewhere because wondering if removing or delaying deposit to a later state could cause other side effects (deposit amount already gone.. ? to punish/mitigate that might need new, more complex logic, delayed proving.. (?))

cyberhorsey: Updated to use TKOToken.

packages/protocol/contracts/L1/libs/LibClaiming.sol
```diff
@@ -0,0 +1,147 @@
...
+        ) {
+            revert L1_CLAIM_AUCTION_WINDOW_PASSED();
+        }
+
+        uint256 baseFee = claimBaseFee(state);
+        // if user hasnt sent enough to meet their personal base deposit amount
+        // we dont allow them to claim.
+        if (bid < baseFee) {
+            revert L1_INVALID_CLAIM_DEPOSIT();
+        }
```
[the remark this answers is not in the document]
cyberhorsey: done, ty!

cyberhorsey:
Tests are failing because I Merged the `major_protocol_upgrade` branch onto this branch. I removed the hardhat tests as well.

Issue #4772, opened 2022-12-31
fix(bridge-ui): History tx in the web not show up


mdhs01:
### Describe the bug

Trying to bridge $HORSE Ethereum A1 to Taiko A1, Notification for this tx was succesfull but i cant see history tx in the web.

### Steps to reproduce

Steps to reproduce here.

### Additional context

Additional context here.

dantaik:
Could you please provide your bridge transaction hash?

dantaik:
I believe we have an issue for this - loading data from relayer.

cyberhorsey:
This has been resolved.

PR #13244, opened 2023-03-02
feat(bridge-ui): fetch transactions from relayer api

shadab-taiko:
Addresses #4772 (and duplicate #12976)


packages/bridge-ui/src/relayer-api/service.ts
```diff
@@ -0,0 +1,192 @@
...
+     throw new Error("Address need to passed to fetch transactions");
+    }
+    let params = `address=${address}`;
+
+    if(chainID) {
+      params += `&chainID=${chainID}`;
+    }
+
+    
+    const requestURL = `${import.meta.env.VITE_RELAYER_URL}events?${params}`;
```
cyberhorsey, line 31: 1) can we use `import.meta.env.VITE_RELAYER_URL` in the constructor to set a `this.baseUrl` property? If possible we should be using as much of the env vars in `App.svelte` when initializing hte app as possible, and it also makes it more testable.

2) can we use axios's idiomatic get params here? ie:  
```ts
const requestURL = `${this.baseURL}events`;
const params = {
   address: address,
chainID: chainID
}
await axios.get(requestURL, params);
):
```

cyberhorsey (changes requested):

cyberhorsey:
Just one comment, otherwise LGTM!

packages/bridge-ui/src/relayer-api/service.ts
shadab-taiko: Done

cyberhorsey (approved):

dantaik (approved):

PR #13246, opened 2023-03-02
docs(website): update introduction page

archie0x:
Thought it was a good idea to remove the info about the Alpha-1 testnet since it's been deprecated and people still come and ask whether they can use it.


umededoteth:
Another option would be to keep the page as it is but add a note that the testnet has been deprecated.

dantaik (approved):

dionysuzx (approved):

PR #13247, opened 2023-03-03
feat(protocol): cross-sync signal service storage root

dantaik:
This PR shows the idea of synchronizing not only the block header hashes across chains, but also the SignalServcie Storage Root (SSSR or Sssr for short).

- L1->L2: this is done by allowing client to pass in  `l1Sssr` to the anchor transaction. A circuit will need to verify the `anchor` transaction's input value for `l1Sssr` pass integrity check with `l1Hash` and `l1SignalServiceAddress` which are part of the ZKP instance input.

- L2->L1, this is done by requesting the block prover to provide the `l2Sssr` value as part of the evidence. The ZKP instance value now incorporates the `l2Sssr` and the L2SignalService address together with `l2Height` and `l2Hash`, which literally demands the main ZKP to do the same Merkle tree verification in circuits.

## The benefit
Currently to prove signal's validity on the destination chain, we need to provide a full merkle proof which has two parts, the account proof and the storage proof. 
After this change, the account proof will be done only once. Users only need to provide the storage merkle proof, which will be less espensive.

### DIFF
--- a/packages/protocol/contracts/L2/TaikoL2.sol
+++ b/packages/protocol/contracts/L2/TaikoL2.sol
@@ -81,17 +81,21 @@ contract TaikoL2 is EssentialContract, IHeaderSync {
      *
      * @param l1Height The latest L1 block height when this block was proposed.
      * @param l1Hash The latest L1 block hash when this block was proposed.
-     * @param l1Sssr The latest value of the L1 "signal service storage root".
+     * @param l1SignalStorageRoot The latest value of the L1 "signal service storage root".
      */
-    function anchor(uint256 l1Height, bytes32 l1Hash, bytes32 l1Sssr) external {
+    function anchor(
+        uint256 l1Height,
+        bytes32 l1Hash,
+        bytes32 l1SignalStorageRoot
+    ) external {
         _checkPublicInputs();
 
         latestSyncedL1Height = l1Height;
-        SyncData memory syncData = SyncData(l1Hash, l1Sssr);
+        SyncData memory syncData = SyncData(l1Hash, l1SignalStorageRoot);
         _l1SyncData[l1Height] = syncData;
 
         // A circuit will verify the integratity among:
-        // l1Hash, l1Sssr, and l1SignalServiceAddress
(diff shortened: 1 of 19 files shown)


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -44,8 +48,16 @@ contract TaikoL2 is IHeaderSync {
...
-    constructor() {
-        if (block.chainid == 0) revert L2_INVALID_CHAIN_ID();
+    function init(
+        address _addressManager,
+        uint256 _l1ChainId
+    ) external initializer {
+        EssentialContract._init(_addressManager);
+        l1ChainId = _l1ChainId;
+
+        if (block.chainid == 0 && block.chainid == _l1ChainId) {
```
cyberhorsey, line 58: Should be || ?

cyberhorsey (changes requested):

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -75,12 +87,26 @@ contract TaikoL2 is IHeaderSync {
      *
      * @param l1Height The latest L1 block height when this block was proposed.
      * @param l1Hash The latest L1 block hash when this block was proposed.
+     * @param l1Sssr The latest value of the L1 "signal service storage root".
+     * @param l1SssrProof Merkel proof for l1Sssr.
```
Brechtpd, line 91: There should be no need for a Merkle proof, let's do that outside the L2 smart contract and in the circuit so it's more efficient and the anchor transaction data remains small.

However not so sure where to best recover the MPT root from the block hash. I guess let's assume we can do that in the circuits already because we kind of have to do the same thing for the L2 block hash as well.

Brechtpd:
> SignalServcie Storage Root (SSSR or Sssr for short).

Bridge root?

dantaik:
> > SignalServcie Storage Root (SSSR or Sssr for short).
> 
> Bridge root?

renamed to singalRoot

PR #13249, opened 2023-03-03
feat(bridge-ui): support both bull and horse tokens on the bridge UI

### LINKED ISSUE #13243: feat(bridge-ui): support both BULL and HORSE tokens on the bridge-ui
Currently, the bridge-ui only supports a single test ERC20 token. We need to:
- support both HORSE and BULL tokens in the token selector
- mint both tokens in one go from the faucet modal.

This is just a temporary feature required till we have the test ERC20 tokens in place.

jscriptcoder:
What's done here:

1. Re-implementation of `TEST_ERC20`, which creates a list of tokens based on a stringified object coming from env var.
2. Added support for listing multiple test tokens in SelectToken dropdown from `TEST_ERC20`.
3. We no longer check a fixed token, what previously was `TEST_ERC20`, but the selected one from token store.
4. Changed wording in FaucetModal informing about the amount of tokens to be minted.
5. Removed unnecessary BaseQueries component.
6. Configured prettier.
7. New env var: `VITE_TEST_ERC20`: `'[{ address, symbol, name }, ...]'`
8. Updated `README.md` and `.default.env` file to reflect changes on the env vars.

Screenshot 2023-03-05 at 22 04 03


shadab-taiko:
Thanks for configuring the prettier @jscriptcoder , much needed 🙏 .

Could you please add the new env var to the README here: https://github.com/taikoxyz/taiko-mono/blob/main/packages/bridge-ui/README.md#environment-variables that way anyone would be able to run the bridge with the required vars. Please also change any other value that you might have changed locally.

shadab-taiko (dismissed):

packages/bridge-ui/src/components/buttons/SelectToken.svelte
```diff
@@ -1,21 +1,28 @@
...
   import { signer } from '../../store/signer';
   import { userTokens, tokenService } from '../../store/userToken';
   import { fromChain, toChain } from '../../store/chain';
   import Erc20 from '../icons/ERC20.svelte';
   import AddCustomErc20 from '../form/AddCustomERC20.svelte';
 
   let dropdownElement: HTMLDivElement;
+  let showAddressField = false;
+  let loading = false;
+  let customTokens = [];
```
cyberhorsey, line 21: missing type annotation, lets avoid using `any` arrays

packages/bridge-ui/src/components/buttons/SelectToken.svelte
```diff
@@ -1,21 +1,28 @@
...
+
+  userTokens.subscribe((value) => {
```
cyberhorsey, line 23: if we are just setting customTokens to userTokens, why do we duplciate the customTokens array?

we can simple use:
```ts
{#each $userTokens as token}

{/each}
```

which will auto subscribe.

packages/bridge-ui/src/components/buttons/SelectToken.svelte
```diff
@@ -34,29 +41,25 @@
...
-
   async function addERC20(event: SubmitEvent) {
     try {
       loading = true;
       const eventTarget = event.target as HTMLFormElement;
       const { customTokenAddress } = eventTarget;
       const tokenAddress = customTokenAddress.value;
-      if(!ethers.utils.isAddress(tokenAddress)) {
+      if (!ethers.utils.isAddress(tokenAddress)) {
         throw new Error();
```
cyberhorsey, line 51: should we throw a valuable error to display to the user?

packages/bridge-ui/src/components/modals/FaucetModal.svelte
```diff
@@ -1,41 +1,35 @@
...
-
   async function isBtnDisabled(signer: Signer) {
     if (!signer) return;
+
     const balance = await signer.getBalance();
+    const address = await signer.getAddress();
+
     const contract = new ethers.Contract(
-      TEST_ERC20.addresses[0].address,
+      $token.addresses[0].address,
```
cyberhorsey, line 27: is $token guaranteed to be set?

cyberhorsey (changes requested):

packages/bridge-ui/src/components/buttons/SelectToken.svelte
jscriptcoder: Good point. Looks like we can simply get rid of it and use the store instead.

packages/bridge-ui/src/components/buttons/SelectToken.svelte
jscriptcoder: Yeah, happens here is that this whole logic is within a try-catch. In the catch we present a toast saying: 'Not a valid ERC20 address'. The problem I see is that many things might go wrong here, not just an invalid ERC20 address. I'll change this error handling a bit.

packages/bridge-ui/src/components/buttons/SelectToken.svelte
jscriptcoder: I actually got rid of customTokens altogether since it wasn't really needed.

packages/bridge-ui/src/components/modals/FaucetModal.svelte
jscriptcoder: I renamed the function `isBtnDisabled()` since it was a bit misleading (does not return a boolean) to `shouldEnableButton` which set the internal state `disabled`, and handled the possibility of not having a token selected

shadab-taiko (dismissed):

dantaik (approved):

cyberhorsey (approved):

PR #13270, opened 2023-03-07
refactor(protocol): remove proofInvalidBlock [step 2]

dantaik:
- change extraData and mixHash
- No longer pass-in 0 values in proposeBlock
- Inline _checkMetadata and delete the method
- Change some 'memory' parameter to 'calldata' to avoid data copying
- Unify proveBlock and proveBlockInvalid

### More about Unify proveBlock and proveBlockInvalid
I realized that RLP handling on L1 is really expensive (@cyberhorsey the bridge code may also have to be optimized), more expensive than the protocol can afford. Then I tried, in this PR, to remove all RLP logics in the core protocol, and ended up with something Brecht has suggested before -- unifying proveBlock and probeBlockInvalid.

Previously, to prove a block is valid, we ask the prover to provide a BlockHeader, then hash the BlockHeader to get the L2 block's blockhash. 

Now the L2 block's blockhash is provided directly by the prover, then we need to hash all non-default BlockHeader fileds (available in the BlockMetadata) into plonk verifier's instance.  To avoid using RPL encoding which is super expensive, all these fields are converted to bytes32 then concatenated into a long byte array to be keccaked.


davidtaikocha (approved):

adaki2004 (approved):

Brechtpd (approved):

packages/protocol/contracts/L1/TaikoConfig.sol
```diff
@@ -34,7 +34,6 @@ library TaikoConfig {
                 maxVerificationsPerTx: 10, //owner:david. Each time one more block is verified, there will be ~20k more gas cost.
                 blockMaxGasLimit: 6000000, // owner:david. Set it to 6M, since its the upper limit of the Alpha-2 testnet's circuits.
                 maxTransactionsPerBlock: 79, //  owner:david. Set it to 79  (+1 TaikoL2.anchor transaction = 80), and 80 is the upper limit of the Alpha-2 testnet's circuits.
-                maxBytesPerTxList: 120000, // owner:david. Set it to 120KB, since 128KB is the upper size limit of a geth transaction, so using 120KB for the proposed transactions list calldata, 8K for the remaining tx fields.
```
davidtaikocha, line 37: Just wondering that, we removed this `maxBytesPerTxList` limit here because of 4884, and after 4844, in theory a user will be able to send a big blob tx which uses up all blob space in a block (so the bytes limit here is really large), so will we support something like allowing multiple txLists in one big tx blob? if so, combined with conditional transactions, maybe we can mitigate those decentralized proposers issues without sacrificing many L2 tps?

Brechtpd: I've thought a bit about this, but I'm really sure yet how this will play out in practice. In any case it seems inevitable that we will have so support at least a custom rub-range inside a single blob so you can indeed create multiple blocks with it.
1. Easiest case is that proposer submits a bundle of the transaction data and then one or more blocks together. This is easy because it's basically a single big transaction that either will get included or not.
2. Another possible case is that submitting the transaction data in a blob is kind of like a separate phase. So there's somebody that just puts that collects all transactions and pushes them in blobs and then proposers can pick and match which transactions, and in which order, to include the transaction in a block. Using a transaction in a block would need a fee payment of some kind to whoever put the data in the blob. Basically it's kind of like an on-chain mempool where there's a more market for transactions. So the block data in the smart contract would contain something like 1 or more blob commitments and a list of indices of transactions available in those blobs
3. Probably some other possibilities as well.

We may even still want to support transactions in the transaction data if 2. is the more realistic one. It won't always make sense to get all transaction in blob data first (because e.g. you want to keep it private as long as possible because it's the transaction that makes MEV possible or things like that, or using blob data is overkill). 

In the circuits we'll have some `maxBlobsPerTxList` as well.

PR #13271, opened 2023-03-07
fix(relayer): estimate gas for tx, set gas to 2.5mil if not estimatable. works now.

cyberhorsey:


RogerLamTd (dismissed):

RogerLamTd:
Just out of curiosity, is there a specific benefit to separating estimate gas and isProfitable? @cyberhorsey

cyberhorsey:
> Just out of curiosity, is there a specific benefit to separating estimate gas and isProfitable? @cyberhorsey

Not only are they two different domain problems that should be unrelated, it lets us handle the errors from them separately. For instance now we don't even run the isProfitable code if the CLI flag for isProfitable isnt set, and now if the estimateGas code doesn't return us a good number, we set a large limit (because it seems unable to estimate cost for deploying an ERC20 when necessary)

RogerLamTd:
> > Just out of curiosity, is there a specific benefit to separating estimate gas and isProfitable? @cyberhorsey
> 
> 
> 
> Not only are they two different domain problems that should be unrelated, it lets us handle the errors from them separately. For instance now we don't even run the isProfitable code if the CLI flag for isProfitable isnt set, and now if the estimateGas code doesn't return us a good number, we set a large limit (because it seems unable to estimate cost for deploying an ERC20 when necessary)

Gotcha, so we're leaving the flag as false for this coming testnet entirely then

RogerLamTd (approved):

davidtaikocha (approved):

PR #13272, opened 2023-03-07
feat(relayer): MessageStatusChanged events

cyberhorsey:
Get all transactions where a msg hash status has been changed, so frontend can use this to show claiming tx's.


packages/relayer/indexer/subscribe.go
```diff
@@ -63,3 +82,32 @@ func (svc *Service) subscribe(ctx context.Context, chainID *big.Int) error {
...
+		return svc.bridge.WatchMessageStatusChanged(&bind.WatchOpts{
+			Context: ctx,
+		}, sink, nil)
+	})
+
+	defer sub.Unsubscribe()
+
+	for {
+		select {
+		case err := <-sub.Err():
```
davidtaikocha, line 103: Do we need to also select `ctx.Done()` here?

packages/relayer/message/process_message.go
```diff
@@ -165,3 +172,48 @@ func (p *Processor) sendProcessMessageCall(
...
+	bridgeAbi, err := abi.JSON(strings.NewReader(bridge.BridgeABI))
+	if err != nil {
+		return errors.Wrap(err, "abi.JSON")
+	}
+
+	m := make(map[string]interface{})
+
+	for _, log := range receipt.Logs {
+		topic := log.Topics[0]
+		if topic == crypto.Keccak256Hash([]byte("MessageStatusChanged(bytes32,uint8,address)")) {
```
davidtaikocha, line 191: Can we re-use the values defined in `bridgeAbi.Events[idx].Name / bridgeAbi.Events[idx].ID` here? instead of maintaining the hard-coded event selectors manually here?

cyberhorsey: Yep, will do!

packages/relayer/indexer/subscribe.go
cyberhorsey: Will do!

davidtaikocha (approved):

dantaik (approved):

PR #13280, opened 2023-03-09
refactor(protocol): unify proveBlock and proveBlockInvalid [step 3]

dantaik:
I realized that RLP handling on L1 is really expensive (@cyberhorsey the bridge code may also have to be optimized), more expensive than the protocol can afford. Then I tried, in this PR, to remove all RLP logics in the core protocol, and ended up with something Brecht has suggested before -- unifying proveBlock and probeBlockInvalid.

Previously, to prove a block is valid, we ask the prover to provide a BlockHeader, then hash the BlockHeader to get the L2 block's blockhash. 

Now the L2 block's blockhash is provided directly by the prover, then we need to hash all non-default BlockHeader fileds (available in the BlockMetadata) into plonk verifier's instance.  To avoid using RPL encoding which is super expensive, all these fields are converted to bytes32 then concatenated into a long byte array to be keccaked.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProving.sol
+++ b/packages/protocol/contracts/L1/libs/LibProving.sol
@@ -16,9 +16,9 @@ import {TaikoData} from "../../L1/TaikoData.sol";
 library LibProving {
     using LibUtils for TaikoData.State;
 
-
-    bytes32 public constant  VERIFIER_OK =  // keccak256("Taiko")
-    0xc6baa0f809cf694efa91aff0be8930f6986b3b4037ae3a94e302aeff2f794039;
+    // keccak256("Taiko")
+    bytes32 public constant VERIFIER_OK =
+        0xc6baa0f809cf694efa91aff0be8930f6986b3b4037ae3a94e302aeff2f794039;
 
     event BlockProven(
         uint256 indexed id,
@@ -139,7 +139,8 @@ library LibProving {
                 )
             );
 
-            if (!verified || ret.length != 32 || bytes32(ret) != VERIFIER_OK) revert L1_INVALID_PROOF();
+            if (!verified || ret.length != 32 || bytes32(ret) != VERIFIER_OK)
+                revert L1_INVALID_PROOF();
         }
 
         emit BlockProven({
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -7,18 +7,19 @@
...
 import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
 import {TaikoData} from "../../L1/TaikoData.sol";
 
 library LibProving {
-    using LibBlockHeader for BlockHeader;
     using LibUtils for TaikoData.State;
 
+    // keccak256("taiko")
+    bytes32 public constant VERIFIER_OK =
+        0x93ac8fdbfc0b0608f9195474a0dd6242f019f5abc3c4e26ad51fefb059cc0177;
```
Brechtpd, line 21: It should be possible to just do `bytes32 public constant VERIFIER_OK =keccak256("taiko");` and the compiler will calculate it at compile time. Though I remember a couple of years ago it didn't always work reliably but should hopefully be working as expected in all cases.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -148,70 +82,71 @@ library LibProving {
...
+                    "signal_service",
+                    false
+                );
+                address l2SignalService = resolver.resolve(
+                    config.chainId,
+                    "signal_service",
+                    false
+                );
+
+                instance = keccak256(
```
Brechtpd, line 115: I think data like `timestamp` and `mixhash` are missing here so the circuit can verify the blockhash is calculated using these expected values. Or could also be I'm overlooking where these are inserted in the circuit.

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: fixed

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: Brecht, since we no longer pass L2 block header to the `proveBlock` function, we don't have access to the `timestamp` and `mixhash` values on L1.

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: fixed

PR #13281, opened 2023-03-09
feat(bridge-ui): support route param to directly open transactions tab

shadab-taiko:
Adds support to directly link to transactions tab.

To directly open transactions tab use the URL: `<bridge-url>/#/transactions`

### DIFF
--- a/packages/bridge-ui/src/App.svelte
+++ b/packages/bridge-ui/src/App.svelte
@@ -251,7 +251,7 @@
   };
 
   const routes = {
-    "/": wrap({
+    "/:tab?": wrap({
       component: Home,
       props: {},
       userData: {},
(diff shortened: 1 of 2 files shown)


davidtaikocha (approved):

packages/bridge-ui/src/App.svelte
```diff
@@ -251,7 +251,7 @@
   };
 
   const routes = {
-    "/": wrap({
+    "/:tab?": wrap({
```
jscriptcoder, line 254: I have a question about the `wrap()` function. Do we really need to wrap the router if we're not using any of the features? I guess it doesn't make any harm, just curiosity.

jscriptcoder (approved):

packages/bridge-ui/src/App.svelte
shadab-taiko: We can still use it as it enables code splitting. Sets a good tone for future routes as well reduces size per request for the current bridge UI.

jscriptcoder: Got it. Thanks

PR #13283, opened 2023-03-09
feat(protocol): Bridge proof should only use storage proof

cyberhorsey:
Not sure if this is the correct way to do this, but I'd take a stab at it regardless. This is a large gas optimization for the bridge.
@dantaik because we sync the storageRoot, we can remove the account proof, and use the synced signal root instead. I'm not sure if I did it correctly, tests pass though. Please let me know if this is the appropriate way or if I missed some additional verification necessary, or removed too much.

`verifyWithAccountProof` is for `isMessageFailed`, which still requires an account proof, since it uses the Bridge's storage tree not the SignalService.

### DIFF
--- a/packages/protocol/contracts/libs/LibTrieProof.sol
+++ b/packages/protocol/contracts/libs/LibTrieProof.sol
@@ -27,18 +27,38 @@ library LibTrieProof {
      *********************/
 
     /**
-     * Verifies that the value of a slot in the storage tree of `addr`
-     * is `value`.
+     * Verifies that the value of a slot in the storage of an account is value.
      *
-     * @param stateRoot The merkle root of state tree.
-     * @param addr The contract address.
      * @param slot The slot in the contract.
      * @param value The value to be verified.
-     * @param mkproof The proof obtained by encoding state proof and storage
-     *        proof.
+     * @param storageProof The proof obtained by encoding storage proof.
+     * @param storageRoot The synced storage root for the given block.
      * @return verified The verification result.
      */
     function verify(
+        bytes32 slot,
+        bytes32 value,
+        bytes calldata storageProof,
+        bytes32 storageRoot
+    ) public pure returns (bool verified) {
+        verified = LibSecureMerkleTrie.verifyInclusionProof(
+            abi.encodePacked(slot),
+            LibRLPWriter.writeBytes32(value),
+            storageProof,
+            storageRoot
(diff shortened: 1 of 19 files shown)


dantaik:
@cyberhorsey what's the gas cost difference?

packages/protocol/contracts/libs/LibTrieProof.sol
```diff
@@ -27,18 +27,38 @@ library LibTrieProof {
...
+            LibRLPWriter.writeBytes32(value),
+            storageProof,
+            storageRoot
+        );
+    }
+
+    /**
+     * Verifies that the value of a slot in the storage of an account is value.
+     *
+     * @param stateRoot The merkle root of state tree..
```
davidtaikocha, line 55:
```suggestion
     * @param stateRoot The merkle root of state tree.
```

davidtaikocha (approved):

Brechtpd (approved):

Brechtpd:
Just a general question, it seems like it's possible to move `isSignalReceived` out of `SignalService` pretty easily by for example making it an external library function instead. That way the the contract gets much simpler and the contract is closer to just being a storage contract with a method to store some data and a method to read some data, so less risky and then there will never be a need to update it. Now with the built in inclusion of the MPT check, the contract gets outdated when for example Ethereum changes something its trie format or a precompile is made available that is more efficient becomes available.

Also a specific reason signals cannot be `0`? Maybe a potential issue with MPT check that always returns true even when the signal is set to `0` because it's the default value? (depends on the MPT verification logic how that is handled).

cyberhorsey:
> Just a general question, it seems like it's possible to move `isSignalReceived` out of `SignalService` pretty easily by for example making it an external library function instead. That way the the contract gets much simpler and the contract is closer to just being a storage contract with a method to store some data and a method to read some data, so less risky and then there will never be a need to update it. Now with the built in inclusion of the MPT check, the contract gets outdated when for example Ethereum changes something its trie format or a precompile is made available that is more efficient becomes available.
> 
> Also a specific reason signals cannot be `0`? Maybe a potential issue with MPT check that always returns true even when the signal is set to `0` because it's the default value? (depends on the MPT verification logic how that is handled).

I can move that logic out of SignalService to an external library method in a separate PR!

And yes, I think that is the reason it cannot be 0.

PR #13285, opened 2023-03-09
fix(protocol): add foundry tests and fix tokenomic bugs

dantaik:
- support foundry tests and  add a few foundry tests
- rename `config.rewardMultiplierPctg` to `config.feeMultiplierPctg` to be consistent with other config vars.
- fix math in `getTimeAdjustedFee`
- validate configs in `TaikoL1.init()`
- baseFee is now a uint64 represent *Twei*, not *Szabo*.
- a few optimization using `unchecked` -- **Need your careful review and feedback**
- allow prover/proposer to have independent configs for dynamic fees
- optimized gas cost for L2's anchor transaction

A natural question to ask is "what exactly do we need to prove using ZKP?", to answer the question, I'd like to share [this file ](https://www.notion.so/taikoxyz/What-need-to-be-zk-proved-in-a-Taiko-block-c93fdd32b67448a3a11e777be30facec)for us to review and improve.

| getBlockFee                                    | 4749            | 5165   | 5334   | 5371   | 471     |
| getConfig                                      | 4194            | 4194   | 4194   | 4194   | 5       |
| getStateVariables                              | 1813            | 1850   | 1813   | 7813   | 810     |
| init                                           | 189224          | 189224 | 189224 | 189224 | 5       |
| keyForName                                     | 2213            | 2355   | 2355   | 2498   | 30      |
| proposeBlock                                   | 16999           | 23102  | 17019  | 64146  | 339     |
| proveBlock                                     | 51095           | 59954  | 51095  | 125395 | 339     |
| resolve(string,bool)(address)                  | 4374            | 4560   | 4599   | 10874  | 689     |
| resolve(uint256,string,bool)(address)          | 4277            | 4306   | 4277   | 6277   | 339     |
| verifyBlocks                                   | 17288           | 33842  | 17288  | 130814 | 331     |
| anchor                                    | 4038            | 69459  | 75150  | 101012 | 1002    |


packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -17,32 +17,67 @@ import {TaikoToken} from "../TaikoToken.sol";
...
+        uint256 amount
     ) internal {
         uint256 balance = state.balances[msg.sender];
-        if (balance <= 1) return;
+        if (balance <= amount) revert L1_INSUFFICIENT_TOKEN();
+
+        uint256 x;
+        {
+            x = balance - amount;
+        }
```
Brechtpd, line 36: Missing `unchecked`? I think this function should also be used quite infrequently so maybe not worth it to complicate the code to save some gas?

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -113,59 +116,60 @@ library LibVerifying {
...
-                        state.balances[fc.prover] = 1;
-                    } else {
-                        state.balances[fc.prover] += reward;
-                    }
+            // reward the prover
+            if (reward > 0) {
+                if (state.balances[fc.prover] == 0) {
+                    // Reduce reward to 1 wei as a penalty if the prover
+                    // has 0 TKO outstanding balance.
+                    state.balances[fc.prover] = 1;
```
Brechtpd, line 132: Whoever is running the block verification still pays the same amount of gas, so maybe better to simply not do anything? This will reduce the worst case gas cost for block verification.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -113,59 +116,60 @@ library LibVerifying {
...
-                }
-                if (refund > 0) {
-                    if (state.balances[proposal.proposer] == 0) {
-                        // Reduce refund to 1 wei as a penalty if the proposer
-                        // has 0 TKO outstanding balance.
-                        state.balances[proposal.proposer] = 1;
-                    } else {
-                        state.balances[proposal.proposer] += refund;
-                    }
+            // refund proposer deposit for valid blocks
```
Brechtpd, line 138: (DOS risk reminder, but I guess makes sense to leave this here for now)

packages/protocol/contracts/common/AddressResolver.sol
```diff
@@ -69,6 +69,16 @@ abstract contract AddressResolver {
...
     }
 
+    function keyForName(
+        uint256 chainId,
+        string memory name
+    ) public pure returns (string memory key) {
+        key = string.concat(Strings.toString(chainId), ".", name);
+        // TODO: the next line is cheaper in gas but will break
+        //       many Hardhat tests.
+        // key = string(bytes.concat(bytes32(chainId), bytes(name)));
```
adaki2004, line 79: Hey @dantaik !

The reason why the hardhat tests are failing (and foundry not) is because the `bytes32(uint256)` kind of conversion does convert the integer into a seemingly empty bytes variable - which only on the surface looks empty.
So basically here, what you get back is the `EMPTY_BYTES + bytes(name)` , within foundry it is passing because the expects also written in Solidity, but the hardhat is different. (+1 for HH :) ).

If you want to double check, here is a demonstration of the issue (run foundry test -vv):

```
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.18;

import "forge-std/Test.sol";
import "forge-std/console.sol";

contract FooBar {
   
    function debugBytesConversion() public {
        address dummyCanonicalTokenAddress = address(this);
        uint256 chainId = 1;
        uint256 chainId2 = 2;

        bytes memory concatenation =
                bytes.concat(
                    bytes32(chainId)
        );

        console.log("Output with chain ID 1 if not keccak hashed:");
        console.log(string(abi.encodePacked(concatenation)));

        bytes memory concatenation2 =
                bytes.concat(
                    bytes32(chainId2)
                );

        console.log("Output with chain ID 2 if not keccak hashed:");
        console.log(string(abi.encodePacked(concatenation2)));


        console.log("Output with chain ID 1 if keccak hashed:");
        console.log(string(abi.encodePacked(keccak256(concatenation))));


        console.log("Output with chain ID 2 if keccak hashed:");
        console.log(string(abi.encodePacked(keccak256(concatenation2))));
    }
}

contract EmptyBytesTest is Test {
    FooBar foobar;

    function setUp() public {
        foobar = new FooBar();
    }

    function testBytes() external {
        foobar.debugBytesConversion();
    } 
}
```

adaki2004: @dantaik 
The difference is coming from how `String.toString(chainId)` works under the hood, and how `bytes32(chainId)` does that.
The former with chainID = 1 will be equal to (hex"0000000000000000000000000000000000000000000000000000000000000061"), the one digit character.
The latter with chainID = 1 will be equal to (hex"0000000000000000000000000000000000000000000000000000000000000001"), the SOH (start of heading) character.

Both could work, but on the other hand we would need to adjust every call of the AddressManager's
```
 function setAddress(
        string memory name,
        address newAddress
  )
```
that using the same mechanism to construct the key (name).

adaki2004: Went deeper debugging and this is how this line:
`key = string(bytes.concat(bytes32(chainId), bytes(name)));`
 interprets it:
 
kép

1. Our chainId (uint256) - passed to the function- is converted to from integer to 32bytes hex number (identical number as decimal):
`0x0000000000000000000000000000000000000000000000000000000000000538`
2. The concatenation is - because the word 'taiko' is represented in ascii nibles.
`0x00000000000000000000000000000000000000000000000000000000000005387461696b6f` --> 74 61 69 6b 6f
3. After string casting it looks (and returns) like this:
`8taiko` but actually it is (with leading NULL characters, this is why editors/IDEs does not visualize them):
kép

So this is why I would not recommend this exact solution. I'll open a branch + PR to propose a solution soon.

packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -17,32 +17,67 @@ import {TaikoToken} from "../TaikoToken.sol";
...
+        if (x == 0) {
```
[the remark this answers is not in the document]
dantaik: fixed

packages/protocol/contracts/common/AddressResolver.sol
dantaik: Thank you for digging into it. I know the problem, I guess we just need to implement a TypeScript method for `string(bytes.concat(bytes32(chainId), bytes(name)));` then use that method in all tests.

adaki2004: For example this is what we have now:
`0x00000000000000000000000000000000000000000000000000000000000005387461696b6f` --> 05 38 74 61 69 6b 6f.
05 will chopped by the string() casting within the contract and the key will be `8taiko`.

Now, we have another chan let’s say these are the created bytes:

0x00000000000000000000000000000000000000000000000000000000000002387461696b6f --> 02 38 74 61 69 6b 6f
This will result in tha same, `8taiko` .

i know that on different chains, most we will not see clashes (in case tho we want to keep track other , ‘B’ chain data on ‘A’ chain we can have) but also the public function keyForName() is queriable and not returns the proper chainId+name concatenation.

If you still want to proceed anyways, i can try to implement on test side to pass. (Today a bit sick tho.)

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -116,50 +115,65 @@ library LibProving {
...
+                inputs[6] = bytes32(uint256(evidence.meta.id));
+                inputs[7] = bytes32(evidence.meta.l1Height);
+                inputs[8] = evidence.meta.l1Hash;
+                inputs[9] = evidence.meta.txListHash;
+
+                // Circuits shall use this value to check anchor gas limit.
+                // Note that this value is not necessary and can be hard-coded
+                // in to the circuit code, but if we upgrade the protocol
+                // and the gas limit changes, then having it here may be handy.
+                inputs[10] = bytes32(config.anchorTxGasLimit);
```
dantaik, line 136: Brecht, since we no longer pass L2 block header to the `proveBlock` function, we don't have access to the `timestamp` and `mixhash` values on L1.

dantaik: On L2, I think we also don't need to put these values into the `inputs` array as they will be part of the block header which is available to the circuits. `chainid` is not part of th block header, thus has to be part of the input array.

dantaik: fix an issue here: https://github.com/taikoxyz/taiko-mono/pull/13336

packages/protocol/contracts/common/AddressResolver.sol
adaki2004: I think it is dangerous especially in the bridge, where eg. in `LibBridgeProcess.sol`:
```
        address srcBridge = resolver.resolve(
            message.srcChainId,
            "bridge",
            false
        );
```

On the destination chain, checking the source chain. And in case for example destination chainId is: 1336 (0x538) and destination is : 568 (0x238). Then, this is a clash.

dantaik:
Guys, I'll merge this PR to the upstream branch so there you guys can give it a thorough look at the whole new implementation.

PR #13307, opened 2023-03-13
fix(bridge-ui): transaction and pendingTransaction refactor

shadab-taiko:


packages/bridge-ui/src/App.svelte
```diff
@@ -174,20 +174,20 @@
   });
 
   pendingTransactions.subscribe((store) => {
-    store.forEach(async (tx) => {
-      await $signer.provider.waitForTransaction(tx.hash, 1);
+    (async () => {
+      const confirmedPendingTxIndex = await Promise.race(
```
shadab-taiko, line 178: This way, we wouldn't be having stale `Promise`s. Earlier stale `Promise`s were resolving and each was triggering a toast which resulted in more than one toast being shown for each pending transaction.

packages/bridge-ui/src/relayer-api/service.ts
```diff
@@ -116,7 +110,9 @@ class RelayerAPIService implements RelayerAPI {
         );
 
         const event = events.find(
-          (e) => e.args.message.owner.toLowerCase() === address.toLowerCase(),
+          (e) =>
+            e.args.message.owner.toLowerCase() === address.toLowerCase() &&
+            e.args.message.depositValue.eq(tx.message.depositValue),
```
shadab-taiko, line 115: For transactions included within the same block we were picking the first event out of many which resulted in wrong data being shown in the transaction list. This fixes it.

jscriptcoder: Maybe we could add that as a comment?

packages/bridge-ui/src/vite-env.d.ts
```diff
@@ -1,2 +1,32 @@
...
+  readonly VITE_MAINNET_HEADER_SYNC_ADDRESS: string;
+  readonly VITE_TAIKO_HEADER_SYNC_ADDRESS: string;
+  readonly VITE_MAINNET_BRIDGE_ADDRESS: string;
+  readonly VITE_TAIKO_BRIDGE_ADDRESS: string;
+  readonly VITE_MAINNET_SIGNAL_SERVICE_ADDRESS: string;
+  readonly VITE_TAIKO_SIGNAL_SERVICE_ADDRESS: string;
+  readonly VITE_TEST_ERC20: string;
+}
+
+interface ImportMeta {
```
jscriptcoder, line 30: Nice one

jscriptcoder (dismissed):

cyberhorsey (dismissed):

packages/bridge-ui/src/relayer-api/service.ts
shadab-taiko: I just realized that even the depositValue could be the exact same. So we need to also compare `msgHash`.

jscriptcoder (approved):

davidtaikocha (approved):

PR #13313, opened 2023-03-14
chore(bridge-ui): reduce logic and remove some store variables

jscriptcoder:
Store is s powerful tool in Svelte for state management. But if we want to store small values that are already available initially and won't ever change, then we are introducing unnecessary complexity here, and also could affect the performance. This might be the case of our providers map. It's used across different components, but its value, which is available when hitting the app, will always be the same.

Found a similar situation with our `wagmiClient` variable, which is only used in `Connect.svelte`, but initialized in `App.svelte`. It could simply be a local variable

### DIFF
--- a/packages/bridge-ui/src/App.svelte
+++ b/packages/bridge-ui/src/App.svelte
@@ -53,7 +53,9 @@
     import.meta.env.VITE_RELAYER_URL,
   );
 
-  const tokenStore: TokenService = new CustomTokenService(window.localStorage);
+  const tokenStore: TokenService = new CustomTokenService(
+    globalThis.localStorage,
+  );
 
   tokenService.set(tokenStore);
 
--- a/packages/bridge-ui/src/App.svelte
+++ b/packages/bridge-ui/src/App.svelte
@@ -7,10 +7,7 @@
 
   import Home from './pages/home/Home.svelte';
   import { setupI18n } from './i18n';
-  import { BridgeType } from './domain/bridge';
-  import ETHBridge from './eth/bridge';
-  import { bridges, chainIdToTokenVaultAddress } from './store/bridge';
-  import ERC20Bridge from './erc20/bridge';
+  import { chainIdToTokenVaultAddress } from './store/bridge';
   import {
     pendingTransactions,
     transactioner,
@@ -24,9 +21,7 @@
   import { chains, CHAIN_MAINNET, CHAIN_TKO } from './domain/chain';
   import { providers } from './domain/provider';
   import SwitchEthereumChainModal from './components/modals/SwitchEthereumChainModal.svelte';
-  import { ProofService } from './proof/service';
   import { ethers } from 'ethers';
-  import type { Prover } from './domain/proof';
   import { successToast } from './utils/toast';
   import { StorageService } from './storage/service';
   import { MessageStatus } from './domain/message';
@@ -38,17 +33,6 @@
   import RelayerAPIService from './relayer-api/service';
   import type { RelayerAPI } from './domain/relayerApi';
   import { relayerApi, relayerBlockInfoMap } from './store/relayerApi';
-
-  const prover: Prover = new ProofService(providers);
-
-  const ethBridge = new ETHBridge(prover);
-  const erc20Bridge = new ERC20Bridge(prover);
-
-  bridges.update((store) => {
-    store.set(BridgeType.ETH, ethBridge);
-    store.set(BridgeType.ERC20, erc20Bridge);
-    return store;
-  });
 
   chainIdToTokenVaultAddress.update((store) => {
     store.set(CHAIN_TKO.id, import.meta.env.VITE_TAIKO_TOKEN_VAULT_ADDRESS);
--- a/packages/bridge-ui/src/App.svelte
+++ b/packages/bridge-ui/src/App.svelte
@@ -4,12 +4,6 @@
   import Router from 'svelte-spa-router';
   import { SvelteToast } from '@zerodevx/svelte-toast';
   import type { SvelteToastOptions } from '@zerodevx/svelte-toast';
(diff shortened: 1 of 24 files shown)


packages/bridge-ui/src/domain/chain.ts
```diff
@@ -137,3 +137,8 @@ export const taiko: WagmiChain = {
   //   address: "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e",
   // },
 };
+
+// Will help us to map from chain id to RPC provider
+export const providers = new Map<number, ethers.providers.JsonRpcProvider>();
```
shadab-taiko, line 142: I am not an expert in domain driven design but I think `providers` should be a in a domain of it's own.

shadab-taiko: @cyberhorsey what do you think?

packages/bridge-ui/src/components/buttons/Connect.svelte
```diff
@@ -11,22 +10,74 @@
...
   } from '../../store/modal';
   import { errorToast, successToast } from '../../utils/toast';
   import Modal from '../modals/Modal.svelte';
-  import { wagmiClient } from '../../store/wagmi';
   import MetaMask from '../icons/MetaMask.svelte';
   import WalletConnect from '../icons/WalletConnect.svelte';
   import CoinbaseWallet from '../icons/CoinbaseWallet.svelte';
   import { transactioner, transactions } from '../../store/transactions';
 
+  const iconMap = {
```
shadab-taiko, line 43: Not sure if initialising `wagmi` in a UI component file makes sense. It should either be in `App.svelte` like it was and I am fine with not writing it to a store, or it could be exposed as a singleton from service.

It makes more semantic sense to have components act as consumers and not initialisers. @jscriptcoder @cyberhorsey  what do you think about this?

packages/bridge-ui/src/domain/chain.ts
d1onys1us, line 142: do we need to use a map? maybe we can use a regular object / dictionary. if we use a regular object with type inference, we can enforce the specific chainId is only accessible from the dictionary.

jscriptcoder: No, we don't, but this is used in some models that take a Map, and I didn't want to refactor these models

packages/bridge-ui/src/components/buttons/Connect.svelte
jscriptcoder: Agree the component should act as a consumer, in which case the client should be injected as a prop. The problem with using Svelte store in order to provide Connect component with wagmi client feels hacky. I prefer the idea of using a service like you mention, but definitely I don't see it as part of the state of the application

packages/bridge-ui/src/domain/chain.ts
jscriptcoder: Let's move it into a separate domain. I wasn't sure where to add this map, which I see as a service, but I think I get the idea of having these domains... I'll make the change

packages/bridge-ui/src/App.svelte
```diff
@@ -54,66 +34,6 @@
...
-      }),
-    ],
-  });
-
-  const prover: Prover = new ProofService(providerMap);
-
-  const ethBridge = new ETHBridge(prover);
-  const erc20Bridge = new ERC20Bridge(prover);
-
-  bridges.update((store) => {
```
jscriptcoder, line 111: We have same situation with this map. It's added into a store variable, when it could simply be a service. Again, I don't think this is part of the state of the application.

cc: @shadab-taiko @cyberhorsey

packages/bridge-ui/src/store/transactions.ts
jscriptcoder, line 1: `showTransactionDetails` and `showMessageStatusTooltip` can perfectly be local variables

packages/bridge-ui/src/store/transactions.ts
```diff
@@ -6,12 +6,5 @@ import type { BridgeTransaction, Transactioner } from '../domain/transactions';
 const pendingTransactions = writable<Transaction[]>([]);
 const transactions = writable<BridgeTransaction[]>([]);
 const transactioner = writable<Transactioner>();
-const showTransactionDetails = writable<BridgeTransaction>();
-const showMessageStatusTooltip = writable<boolean>();
```
jscriptcoder, line 10: `showTransactionDetails` and `showMessageStatusTooltip` can perfectly be local variables within `Transactions` component

cyberhorsey (approved):

cyberhorsey (dismissed):

packages/bridge-ui/src/domain/chain.ts
cyberhorsey: Yep agreed, its own providers file is great!

packages/bridge-ui/src/components/buttons/Connect.svelte
cyberhorsey: I think its fine to have here, but maybe the logic for instantiating the wagmi stuff could be wrapped in a utils method, so the definition is not here, but the consuming of it is? I dont think it merits being its own service, but a simple `utils/connectWallet.ts` where it wraps all the wagmi/connect/init logic could be a lot cleaner.

packages/bridge-ui/src/domain/chain.ts
cyberhorsey: Using the dictionary would require us to make hardcoded changes to the object type unless we inject the expected chainIDs via env var and use them in the typing. IE: not all of our taiko chains are `167001`, so enforcing that as a rule means a code PR to allow `167002` or otherwise.

jscriptcoder:
Closing the PR in favour of splitting it up in smaller chunks

PR #13317, opened 2023-03-14
chore(protocol): Use hardhat loadfixtrue to speed up test case execution

alexshliu:
Use the [hardhat fixture](https://hardhat.org/hardhat-runner/docs/guides/test-contracts#using-fixtures) to speed up test case execution. As stated in the article:
> The first time loadFixture is called, the fixture is executed. But the second time, instead of executing the fixture again, loadFixture will reset the state of the network to the point where it was right after the fixture was executed. This is faster, and it undoes any state changes done by the previous test.

I think it's better than the mocha's beforeEach method, which is used now.


cyberhorsey:
We originally had fixtures but I was asked to remove them.

dantaik:
Lets not merge this PR -- going forward for unit testing, we shall use Foundry.

PR #13323, opened 2023-03-15
docs(website): update presentations page

archie0x:


d1onys1us:
@umededoteth i believe the previous descending order is better for the long run. So the most recent will be at the top. Otherwise looks good to me

umededoteth:
> @umededoteth i believe the previous descending order is better for the long run. So the most recent will be at the top. Otherwise looks good to me

Thanks Dave, updated!

packages/website/pages/docs/resources/presentations.md
```diff
@@ -1,5 +0,0 @@
-# Presentations
```
dantaik, line 1: why are we removing these linkns?

dionysuzx (approved):

dantaik (approved):

PR #13324, opened 2023-03-15
refactor(protocol): optimize anchor transaction to reduce gas cost

dantaik:
Optimization results:

|-------------------------------------------|-----------------|--------|--------|--------|---------|
| anchor                                    | 4038            | 69459  | 75150  | 101012 | 1002    |

This function still cost 101012 gas as for each L2 block, in the anchor function we need to read up to 256 block hashes and hash them into one. 

An alternative solution is to save all L2 hashes into storage then read them out to compare with the blockhashes one by one. Such a solution cost 156244 gas.


dantaik:
@Brechtpd one question I have is: how can we verify that at block n, blochhash(n+10000), for example, does return 0?

davidtaikocha:
> This function still cost 101012 gas as for each L2 block

According to the benchmark in [`generate_genesis.test.ts`](https://github.com/taikoxyz/taiko-mono/blob/optimize_anchor_2/packages/protocol/test/genesis/generate_genesis.test.ts#L170), the latest anchor's gas cost is still `~150000`, not quite sure but seems this one from `receipt.gasUsed` is more accurate?

```
{
  message: 'TaikoL2.anchor gas cost after 256 L2 blocks',
  gasUsed: BigNumber { value: "150333" }
}
```

packages/protocol/contracts/L2/TaikoL2.sol
davidtaikocha, line 1: and we shall also update `LibSharedConfig.anchorTxGasLimit`?

dantaik:
> > This function still cost 101012 gas as for each L2 block
> 
> According to the benchmark in [`generate_genesis.test.ts`](https://github.com/taikoxyz/taiko-mono/blob/optimize_anchor_2/packages/protocol/test/genesis/generate_genesis.test.ts#L170), the latest anchor's gas cost is still `~150000`, not quite sure but seems this one from `receipt.gasUsed` is more accurate?
> 
> ```
> {
>   message: 'TaikoL2.anchor gas cost after 256 L2 blocks',
>   gasUsed: BigNumber { value: "150333" }
> }
> ```

That's interesting...

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: Not really, this config has been removed but the gas limit value is still useful and shall be used by the circuit I believe.

Brechtpd: The circuits will hard-code some gas limit for the transaction, so having the config there makes sense to me. I don't know if it's worth it to pass in the limit so the circuit uses that one or just hard-code it in the circuits.

Brechtpd:
> @Brechtpd one question I have is: how can we verify that at block n, blochhash(n+10000), for example, does return 0?

This is easily handled in the opcode implementation.

How expensive is the block hash checking you think, just looking it at I would estimate around ~30,000 gas? Looks like the biggest cost may be storing 3 values + tx cost, so already at 81,000 gas just for that. If checking the block hashes is expensive makes sense to spend the time to do these checks in circuit instead.

packages/protocol/test2/GasComparison.t.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: UNLICENSED
```
davidtaikocha, line 1:
```suggestion
// SPDX-License-Identifier: MIT
```
MIT like other files in this repo?

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: I made this change in the other PR: https://github.com/taikoxyz/taiko-mono/pull/13285/commits/a6a48abbd425cb7febfef71043827c4a8fb5e18e

I think we can have it in the instance for now, we can remove it if it is indeed uncessary.

dantaik:
> > @Brechtpd one question I have is: how can we verify that at block n, blochhash(n+10000), for example, does return 0?
> 
> This is easily handled in the opcode implementation.
> 
> How expensive is the block hash checking you think, just looking it at I would estimate around ~30,000 gas? Looks like the biggest cost may be storing 3 values + tx cost, so already at 81,000 gas just for that. If checking the block hashes is expensive makes sense to spend the time to do these checks in circuit instead.

Lets do this in steps. The current gas cost for checking public inputs is still ok as it is on L2. Lets optimize this later by having a circuit. I'll add a comment in the current file.

packages/protocol/test2/TaikoL2.t.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: UNLICENSED
```
davidtaikocha, line 1:
```suggestion
// SPDX-License-Identifier: MIT
```
ditto, and other changes LGTM.

davidtaikocha (approved):

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -113,17 +113,19 @@ contract SignalService is ISignalService, EssentialContract {
...
 
             // Calculate the hash of the concatenated arguments using keccak256
-            signalSlot := keccak256(input, 0x40)
+            signalSlot := keccak256(ptr, 64)
 
             // Free the memory allocated for the input
-            mstore(0x40, add(input, 0x60))
+            // Question: since we only used 2 bytes32, should we add 64 to input instead?
+            // Jeff, Brecht?
+            // mstore(0x40, add(ptr, 64))
```
dantaik, line 128: @Brechtpd @cyberhorsey take a look at the question here please

Brechtpd: Replied on Discord: 
> I also think 0x40 is correct. Not updating the free memory pointer is dangerous because uninitialized memory allocated by solidity won't be zero's anymore so any code depending on that will not work correctly anymore. Memory expansion only costs like 3 gas/word so better to do it just to be safe I feel.

Issue #14, opened 2022-07-29
On Tokenomics


dantaik:
# Roles

- Users:
	- paying: `TXS_FEES`L2 transaction fees using the L2's native coins. The native coins can be interpreted as Ether, TAI, or potentially another ERC20 token on L1.
	- receiving: nothing

- Proposers:
	- desc: they build blocks and propose them on L1.
	- paying: 1) `PROPOSING_TX_FEE` ETH as L1 tx fees, 2) `PROVER_FEE ` block proving fees, and 3) small fiat fees for IT infrastructure (ignored in our discussion)
	- receiving: `TXS_FEES `
	- profit = `TXS_FEES	- PROPOSING_TX_FEE	- PROVER_FEE`

- Provers:
	- desc: they generate ZKP proofs to prove blocks are valid or invalid.
	- paying: 1) `PROVING_TX_FEE` ETH as L1 tx fees, and 2) `IT_COST` a considerable amount of ZKP computation hardware cost.
	- receiving: 1) `PROVER_FEE` block proving fees, and potentially 2) `PROVER_REWARD`a performance-based block reward in TAI.
	- profit = `PROVER_FEE + PROVER_REWARD	- PROVING_TX_FEE	- IT_COST `

Here I chose to use *proposers* instead of *sequencers* because in our rollup contract there is nothing related to *ordering transactions in a block*. Given that Ethereum uses PBS (proposer builder separation), we should probably also use *proposers* and *builders* for clarity. The builder-vs-proposer economics is beyond the scope of our tokenomics design.

# Prover Market
The need for ZKPs will inevitably create a market where only the highest bids receive their desired proofs. Such a market will serve as a foundation for almost all ZKP projects that are fulfillable by the prover's services/hardware.

We may choose to or have to build our own prover market, and other projects may also do the same, but eventually, there will be project-independent markets that try to maximize the prover's financial return and minimize the proposer's cost. (It is very similar to DSP, or demand-side platform in web2 ads ecosystem). Such a market will most likely operate off-chain but offer on-chain verifications of off-chain deals.

> INFO: A demand-side platform (DSP) is a type of software that allows an advertiser to buy advertising with the help of automation. Because they allow mobile advertisers to buy high-quality traffic at scale with minimal friction, DSPs are a powerful marketing automation tool.

[...]

PR #13328, opened 2023-03-15
refactor(protocol): update avgBlockTime at verification

dantaik:
Previous we update avgBlockTime when a block is proposed, now I feel it's much reasonable it is updated when a block is proven.  We introduced another variable `avgBlockTimeRT` which will be updated when a block is proposed.

In a future PR, I'll adjust fees based on two comparisons:
1. avgBlocktimeRT <> avgBlockTime
2. numer of pending blocks <> (avgProofTime/avgBlockTime)

| getBlockFee                                    | 5207            | 5627   | 5792   | 5829   | 471     |
| getConfig                                      | 4194            | 4194   | 4194   | 4194   | 5       |
| getStateVariables                              | 5107            | 5156   | 5107   | 13107  | 810     |
| init                                           | 211470          | 211470 | 211470 | 211470 | 5       |
| keyForName                                     | 2213            | 2355   | 2355   | 2498   | 30      |
| proposeBlock                                   | 17537           | 23500  | 17557  | 65422  | 339     |
| proveBlock                                     | 51081           | 59940  | 51081  | 125381 | 339     |
| resolve(string,bool)(address)                  | 4374            | 4560   | 4599   | 10874  | 689     |
| resolve(uint256,string,bool)(address)          | 4277            | 4306   | 4277   | 6277   | 339     |
| verifyBlocks                                   | 18300           | 34880  | 18300  | 141577 | 331     |


davidtaikocha (approved):

Brechtpd:
> Previous we update avgBlockTime when a block is proposed, now I feel it's much reasonable it is updated when a block is proven. We introduced another variable avgBlockTimeRT which will be updated when a block is proposed.

(I'll start with a warning that I think about these fees quite a bit differently than what the current implementation tries to model, so I guess my feedback is not necessarily any good).

What does proving a block have to do with the average block time? Or what is it trying to take into account by doing so? I find the concept of an average block time already quite strange for an L2, but with an extra dependency on when blocks are proven I'm not sure what it is trying to model.

dantaik:
> > Previous we update avgBlockTime when a block is proposed, now I feel it's much reasonable it is updated when a block is proven. We introduced another variable avgBlockTimeRT which will be updated when a block is proposed.
> 
> (I'll start with a warning that I think about these fees quite a bit differently than what the current implementation tries to model, so I guess my feedback is not necessarily any good).
> 
> What does proving a block have to do with the average block time? Or what is it trying to take into account by doing so? I find the concept of an average block time already quite strange for an L2, but with an extra dependency on when blocks are proven I'm not sure what it is trying to model.

`avgBlockTime` and `avgProofTime` are now both updated at the same time when a block is verified, so `avgProofTime/avgBlockTime` is the latest known *level of ZK computation parallelism*. Previously I have the same goal of having this metric but since `avgBlockTime` is updated when a block is proposed, the value of parallelism is not correct.

On the other hand, the `avgBlockTimeRT` reflects the most recent trend of block time as it covers all pending blocks as well. By comparing `avgBlockTime` which lags behind and `avgBlockTimeRT` which is more realtime, we can learn if block proposals is more frequent than earlier or less frequent.


This opens the door for EIP-1559 style adjustment of fees and rewards. EIP1559 tracks a metric agains a target with a cap. If the metric is 0, then the fee goes down to 1-x%, if target is min(2*target, cap), the fee goes up to 1+x%.

Now we can use EIP-1559 in two different ways and may be able to combine them:

### Metric #1
- metric: the number of pending blocks
- target: `avgProofTime/avgBlockTime`
- cap: the ringbuff size - `maxNumBlocks`


### Metric #2
- metric: `avgBlockTimeRT`
- target: `avgBlockTime`
- cap: 30 seconds, for example.

I haven't really figured out all details yet. Still need more time on it.

This is an attempt to get rid of the `slotSmoothingFactor` from the config. It is not straightforward and its value is actually reverse-engineers based on `maxNumBlocks`. Ideally, we can choose a large enough value for `maxNumBlocks` and it should not affect the fee/reward changes, but as of know, it does.

Brechtpd:
These are interesting statistics, but I don't see a direct connection to how these should dictate fees. I still think the only real scarce resource is the amount of available block space, so targeting some X gas/second is the only thing that really makes sense to me on the proposer side. Shameless plug of my very simply proposal, but it's been over 7 months so I think I'm allowed to post this one again: https://github.com/taikoxyz/taiko-mono/issues/14#issuecomment-1207512321. :)

A couple of weeks ago I also wrote this thing about a garbage collection company, I'll paste it here for all garbage company enthusiasts!

### Garbage collection company

There’s a garbage collection responsible for the garbage collection company in a city. 

The city is initially quite small, there’s only a single street with 1000 houses. And so the company hires a single garbage man to collect the trash of each house every day. The garbage man gets paid $100/day. Each house pays $1 to get its trash picked up. Everybody is happy.

The city grows and now there are 2 streets each with 1000 houses. And so the garbage collection company hires another garbage man who also gets paid $100/day. Each house still pays $1 to get its trash picked up. Everybody is happy.

The city grows even larger and now there are 10 streets each with 1000 houses. The company now has 10 garbage men working to pick up the trash each day for $100/day. Each house still pays $1 to get its trash picked up. Because of the larger scale of the operation the company actually makes a larger profit as fixed costs like office space/administration are a smaller part of total pie, and the revenue is more predictable because the amount of trash/day is less variable because of the larger scale. These benefits could allow the company to reduce the price charged to houses to get their trash picked up. Everybody is happy.

The city has grown 10x in size, and so the amount of trash in the street each day has also increased 10x. But this doesn’t matter, the only thing that matters is that the trash gets picked up every day.

However not all remains well and the garbage men go on strike to get paid more. The trash doesn’t get picked up each day anymore. Trash is piling up in the streets. This is a bit inconvenient for people and they are concerned when their trash will get picked up again. But their lives go on largely like usual. People have time sensitive things to do, they have to go to the doctor, they have to buy food, … they can’t stay locked up in their house just because there’s more trash in the street than usual!

Eventually the garbage men get what they want and they get their 10% wage increase. This extra cost to the company is made up by increasing the price that houses pay to get their trash picked up to $1.1. The garbage men work overtime to clear up the backlog of trash which will take a short amount of time. People are reassured that their trash is being picked up again each day, but this didn’t really have any significant impact on their lives so they didn’t care too much to begin with. Everybody is happy.

dantaik:
> These are interesting statistics, but I don't see a direct connection to how these should dictate fees. I still think the only real scarce resource is the amount of available block space, so targeting some X gas/second is the only thing that really makes sense to me on the proposer side. Shameless plug of my very simply proposal, but it's been over 7 months so I think I'm allowed to post this one again: [#14 (comment)](https://github.com/taikoxyz/taiko-mono/issues/14#issuecomment-1207512321). :)
> 
> A couple of weeks ago I also wrote this thing about a garbage collection company, I'll paste it here for all garbage company enthusiasts!
> 
> ### Garbage collection company
> There’s a garbage collection responsible for the garbage collection company in a city.
> 
> The city is initially quite small, there’s only a single street with 1000 houses. And so the company hires a single garbage man to collect the trash of each house every day. The garbage man gets paid $100/day. Each house pays $1 to get its trash picked up. Everybody is happy.
> 
> The city grows and now there are 2 streets each with 1000 houses. And so the garbage collection company hires another garbage man who also gets paid $100/day. Each house still pays $1 to get its trash picked up. Everybody is happy.
> 
> The city grows even larger and now there are 10 streets each with 1000 houses. The company now has 10 garbage men working to pick up the trash each day for $100/day. Each house still pays $1 to get its trash picked up. Because of the larger scale of the operation the company actually makes a larger profit as fixed costs like office space/administration are a smaller part of total pie, and the revenue is more predictable because the amount of trash/day is less variable because of the larger scale. These benefits could allow the company to reduce the price charged to houses to get their trash picked up. Everybody is happy.
> 
> The city has grown 10x in size, and so the amount of trash in the street each day has also increased 10x. But this doesn’t matter, the only thing that matters is that the trash gets picked up every day.
> 
> However not all remains well and the garbage men go on strike to get paid more. The trash doesn’t get picked up each day anymore. Trash is piling up in the streets. This is a bit inconvenient for people and they are concerned when their trash will get picked up again. But their lives go on largely like usual. People have time sensitive things to do, they have to go to the doctor, they have to buy food, … they can’t stay locked up in their house just because there’s more trash in the street than usual!
> 
> Eventually the garbage men get what they want and they get their 10% wage increase. This extra cost to the company is made up by increasing the price that houses pay to get their trash picked up to $1.1. The garbage men work overtime to clear up the backlog of trash which will take a short amount of time. People are reassured that their trash is being picked up again each day, but this didn’t really have any significant impact on their lives so they didn’t care too much to begin with. Everybody is happy.

Thank you for sharing.  It is a very good analogy and I'm totally sharing the same understanding of the problem. So the point is *the garbage company is able to adjust the wage dynamically* to handle the strike and a scenario where more workers now want to be hired as garbage men, under such a scenario the wage shall be lowered a bit).

But, we need to make sure the garbage company can make quick adjustment to wages, otherwise, driveways may be blocked by trash and nobody can move, the city (our L2) will halt. if this happens, everybody cares and will decide to move away from the city.  Well, if the driveways are so wide and nobody get blocked, people will still dislike the existence of a lot of trash and may decide to migrate to somewhere else.

dantaik:
I think we still don't have a good standard to measure what a good tokenomics will look like. I'll put this PR on hold for now.

PR #13336, opened 2023-03-16
fix(protocol): fix missing data from L2 anchor inputs

dantaik:


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -81,17 +82,19 @@ contract TaikoL2 is EssentialContract, IXchainSync {
...
      * @param l1SignalRoot The latest value of the L1 "signal service storage root".
+     * @return parentHash The parent hash. Circuits need to check this hash is
+     *         the same as the block header's parentHash.
      */
     function anchor(
         uint256 l1Height,
         bytes32 l1Hash,
         bytes32 l1SignalRoot
-    ) external {
+    ) external returns (bytes32 parentHash) {
```
dantaik, line 92: @Brechtpd I think before this change, we are not verifying that the `blockhash(block.number-1)` on L2 has the same value as the `parentHash` in the block header.  The circuit needs to verify the return value of `anchor` matches the `parentHash` in the block header. Does it?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -63,7 +63,7 @@ library LibProposing {
         // add salt to this random number as L2 mixHash
         uint256 mixHash;
         unchecked {
-            mixHash = block.prevrandao * state.nextBlockId;
+            mixHash = block.difficulty * state.nextBlockId;
```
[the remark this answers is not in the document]
dantaik: fixed

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -104,16 +107,22 @@ contract TaikoL2 is EssentialContract, IXchainSync {
                     inputs[j % 255] = blockhash(j);
                 }
             }
+
+            // All block properties mentioned in
+            // https://docs.soliditylang.org/en/v0.8.17/units-and-global-variables.html
+            // but not part of a L2 block header shall be added to the list.
             inputs[255] = bytes32(block.chainid);
+            inputs[256] = bytes32(block.basefee);
```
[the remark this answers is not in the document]
dantaik: in that case, we can comment it out but still keep 257 elements.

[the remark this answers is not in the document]
dantaik: comment it out already.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -81,39 +85,53 @@ contract TaikoL2 is EssentialContract, IXchainSync {
...
      * @param l1Hash The latest L1 block hash when this block was proposed.
      * @param l1SignalRoot The latest value of the L1 "signal service storage root".
+     * @param l2parentHash The expected parent hash.
      */
     function anchor(
         uint256 l1Height,
         bytes32 l1Hash,
-        bytes32 l1SignalRoot
+        bytes32 l1SignalRoot,
+        bytes32 l2parentHash
```
dantaik, line 94: @Brechtpd I realized in anchor transaciton, the current block's parent is not verified to be the same value as the parent hash in the block's header. With this extra input, the circuits can do the check.

davidtaikocha (approved):

packages/protocol/contracts/L2/TaikoL2.sol
Brechtpd: Alright. Checking that the previous block hash is correct is trivial because that blockhash is directly available in the circuit so it's just an equality check. It's all the other blockhashes that need to be calculated in the circuit that is the actual circuit work (which is why doing that in the smart contract still makes a lot of sense because that saves quite a bit or work on the circuit side).

dantaik: I ended up removing `l2parentHash` from anchor transaction input and emit an event that captures all block level variables mentioned in https://docs.soliditylang.org/en/v0.8.18/units-and-global-variables.html#block-and-transaction-properties

I'm assuming all these block level variables except `block.chainId`, and `blockhash(n-256)` ... `blockhash(n--2)` will be check by circuits using lookup table data.

dantaik:
@Brechtpd @davidtaikocha please review again.

packages/protocol/test/genesis/generate_genesis.test.ts
```diff
@@ -153,10 +153,13 @@ action("Generate Genesis", function () {
 
             let latestL1Height = 1;
             for (let i = 0; i < 300; i++) {
+                const parent = await provider.getBlock("latest");
+
                 const tx = await TaikoL2.anchor(
                     latestL1Height++,
                     ethers.utils.hexlify(ethers.utils.randomBytes(32)),
                     ethers.utils.hexlify(ethers.utils.randomBytes(32)),
+                    parent.hash,
```
Brechtpd, line 162: Revert?

packages/protocol/utils/generate_genesis/taikoL2.ts
```diff
@@ -272,7 +272,7 @@ async function generateContractConfigs(
                 // TaikoL2
                 // keccak256(abi.encodePacked(block.chainid, basefee, ancestors))
                 _publicInputHash: `${ethers.utils.solidityKeccak256(
-                    ["bytes32[256]"],
+                    ["bytes32[257]"],
```
Brechtpd, line 275: Revert?

PR #13349, opened 2023-03-17
chore(bridge-ui): move maps out of the store

jscriptcoder:
Moving maps into their own domain as initialized services when they are imported. The benefits of doing so is to reduce initialization code in `App.svelte`, which is quite cluttered, as well as getting them out of the stores since they're not part of the state. Less code, less maintenance, easier to reason about and less prone to errors.

I've increased a bit the coverage since it fell a bit bellowed the threshold after these changes.


packages/bridge-ui/src/bridge/birdges.ts
```diff
@@ -0,0 +1,16 @@
+import { ProofService } from '../proof/ProofService';
```
shadab-taiko, line 1: The name of this file has a typo, should be `bridges.ts`

shadab-taiko (dismissed):

dantaik (dismissed):

shadab-taiko (dismissed):

dantaik (dismissed):

cyberhorsey (dismissed):

dionysuzx (dismissed):

shadab-taiko (approved):

adaki2004 (approved):

PR #13356, opened 2023-03-18
feat(protocol): allow multiple L2 blocks to share a txList (blob after 4844)

dantaik:
| contracts/L1/TaikoToken.sol:TaikoToken contract |                 |        |        |        |         |
|-------------------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                                 | Deployment Size |        |        |        |         |
| 1266515                                         | 6358            |        |        |        |         |
| Function Name                                   | min             | avg    | median | max    | # calls |
| burn                                            | 4966            | 8166   | 4966   | 12006  | 11      |
| init                                            | 161314          | 161314 | 161314 | 161314 | 5       |
| mint                                            | 49937           | 49937  | 49937  | 49937  | 5       |
| transfer                                        | 25230           | 27411  | 25230  | 30030  | 11      |

|-------------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                           | Deployment Size |        |        |        |         |
| 770605                                    | 3881            |        |        |        |         |
| anchor                                    | 4038            | 72415  | 78109  | 103971 | 1002    |
| init                                      | 141143          | 141143 | 141143 | 141143 | 2       |

| contracts/signal/SignalService.sol:SignalService contract |                 |       |        |       |         |
|-----------------------------------------------------------|-----------------|-------|--------|-------|---------|
| Deployment Cost                                           | Deployment Size |       |        |       |         |
| 1572045                                                   | 7884            |       |        |       |         |
| Function Name                                             | min             | avg   | median | max   | # calls |
| init                                                      | 93020           | 93020 | 93020  | 93020 | 5       |

| test2/TaikoL1.t.sol:TaikoL1WithConfig contract |                 |        |        |        |         |
| Deployment Cost                                | Deployment Size |        |        |        |         |
| 3328899                                        | 16658           |        |        |        |         |
| getBlockFee                                    | 4873            | 5289   | 5458   | 5495   | 471     |
| getConfig                                      | 4373            | 4373   | 4373   | 4373   | 5       |
| getStateVariables                              | 1813            | 1850   | 1813   | 7813   | 810     |
| init                                           | 189440          | 189440 | 189440 | 189440 | 5       |
| keyForName                                     | 2210            | 2314   | 2355   | 2355   | 35      |
| proposeBlock                                   | 18809           | 24912  | 18829  | 65956  | 339     |
| proveBlock                                     | 56284           | 65173  | 56284  | 132584 | 339     |
| resolve(string,bool)(address)                  | 4374            | 4450   | 4374   | 10874  | 689     |
| resolve(uint256,string,bool)(address)          | 4277            | 4306   | 4277   | 6277   | 678     |
| verifyBlocks                                   | 17410           | 33964  | 17410  | 130911 | 331     |


packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -71,6 +76,8 @@ library TaikoData {
         bytes32 l1Hash;
         bytes32 mixHash;
         bytes32 txListHash;
+        uint32 txStartIdx;
```
dantaik, line 79: Brecht, the start/end index can either be the byte-index or the transaction index. Which one is easier for circuits?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -43,16 +46,41 @@ library LibProposing {
...
+
+            if (
+                info.size == 0 ||
+                info.validSince + config.txListCacheExpiry < _now
+            ) revert L1_TX_LIST_NOT_EXIST();
+
+            if (input.txListEnd > info.size) revert L1_TX_LIST_RANGE();
+        } else {
+            if (_size > config.maxBytesPerTxList) revert L1_TX_LIST();
+            if (input.txListEnd > _size) revert L1_TX_LIST_RANGE();
```
dantaik, line 73: @Brechtpd as of now, the `txListStart` and `txListEnd` are byte-wise indices. Will use them as the transaction indices easier for circuits to implement?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -43,16 +46,41 @@ library LibProposing {
...
             input.gasLimit > config.blockMaxGasLimit
         ) revert L1_INVALID_METADATA();
 
-        // We need txListHash as with EIP-4844, txList is no longer
-        // accssible to EVM.
-        if (input.txListHash != keccak256(txList)) revert L1_TX_LIST_HASH();
+        uint64 _now = uint64(block.timestamp);
+        uint32 _size = uint32(txList.length);
+
+        if (_size == 0) {
```
Brechtpd, line 59: Not sure I understand what this is trying to do. Is this trying to emulate data blobs using tx data?

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: Byte indices will be easier, otherwise we have to be able to decode all data, which may not even be possible because it may contain other data as well which should not impact the given range.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -191,9 +191,13 @@ library LibVerifying {
             config.blockMaxGasLimit == 0 ||
             config.maxTransactionsPerBlock == 0 ||
             config.maxBytesPerTxList == 0 ||
+            // EIP-4844 blob size up to 128K
+            config.maxBytesPerTxList > 128000 ||
             config.minTxGasLimit == 0 ||
             config.slotSmoothingFactor == 0 ||
             config.anchorTxGasLimit == 0 ||
+            // EIP-4844 blob deleted after 30 days
+            config.txListCacheExpiry > 30 * 24 hours ||
```
Brechtpd, line 200: I'm pretty sure we cannot know how old the data is with actual data blobs. I also don't think we should care, we have big problems otherwise. :)

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: Yes, that's the goal. Do you think it make sense?

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: Blob data will be available for 30 days before they are deleted. In the worse case, if a blob is posted to the beacon chain 1 month ago, new block referencing it may not have DA at all. 

With calldata this is not a problem.

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: From a practical point of view, the decentralized sequencer aspect complicates this a bit I think because currently you kind of only want to commit data on-chain when you know when you are also the one that actually uses in a block, otherwise a proposer simply paid the data fee for another proposer to use in the future. It's currently as efficient, or even more efficient because more flexible, to just push the data for each block separately so no real reason to use this system.

But if we just want to test the system as close as possible to EIP-4844 because that's what eventually will be used it does make sense to me to do this. But then I would decouple submitting the tx data from the block proposal completely to more closely resemble blob transactions.

packages/protocol/contracts/L1/libs/LibVerifying.sol
Brechtpd: But I don't think EIP-4844 even exposes a timestamp for when the data was committed, so old commitments will be usable and that's fine.

>  new block referencing it may not have DA at all.

If that would be true we would not be able to use EIP-4844. Ethereum itself doesn't guarantee the data to be there anymore, of course we still depend on ALL data to still be available forever, otherwise the rollup would not be secure anymore!

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: I guess I don't know yet whether with EIP-4844, the blob must be submitted using a separate transaction or any transaction can just attach a blob with it. Do you know?

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: > But I don't think EIP-4844 even exposes a timestamp for when the data was committed, so old commitments will be usable. And that's fine!
I don't think EIP-4844 explicitly exposes a timestamp **after** a blob is submitted, but we know the *current timestamp* in the blob transaction (or the transaction that can attach a blob). In otherword, I'm assuming the blob transaction is not a pure blob, but is capable of running custom contract that verifies the blob.

dantaik: > we still depend on ALL data to still be available forever, otherwise the rollup would not be secure anymore!

https://eips.ethereum.org/EIPS/eip-4844
Screenshot 2023-03-19 at 11 49 48

We need to figure out, post EIP4844, how to dupliciate blob data and make sure they are available for ever.

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: Oh I assumed it would be completely separate because the blob transactions are a separate transaction type and have their own mempool, but they seem to still have a `to` address and I don't know what the purpose of that is. So not sure if blob transactions can execute smart contract code or not.

packages/protocol/contracts/L1/libs/LibVerifying.sol
Brechtpd: > I don't think EIP-4844 explicitly exposes a timestamp after a blob is submitted, but we know the current timestamp in the blob transaction (or the transaction that can attach a blob). In otherword, I'm assuming the blob transaction is not a pure blob, but is capable of running custom contract that verifies the blob.

Maybe so, I could be wrong about this.

> Rollups need data to be available once, long enough to ensure honest actors can construct the rollup state, but not forever.

Yeah exactly, once it's made available by Ethereum for a short time it's up to others to store the data forever, because all users need to be able to get all data from the start of the chain.

> We need to figure out, post EIP4844, how to dupliciate blob data and make sure they are available for ever.

I don't think we really need to do anything special for this. There will be projects like the graph/etherscan that will store all this data, and the data will also be available for everyone running an L2 node, so users should have plenty of options.

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: Okay looks like it's basically a normal transaction with blob data attached, I got this completely wrong. It is possible there could be some small limitations: https://ethereum-magicians.org/t/eip-4844-shard-blob-transactions/8430/34.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -43,16 +49,43 @@ library LibProposing {
             msg.sender != resolver.resolve("solo_proposer", false)
         ) revert L1_NOT_SOLO_PROPOSER();
 
-        if (txList.length > config.maxBytesPerTxList) revert L1_TX_LIST();
+        if (input.txListByteEnd <= input.txListByteStart)
```
Brechtpd, line 52:
```suggestion
        if (input.txListByteEnd < input.txListByteStart)
```
I guess making an empty range still valid to allow creating empty blocks could still be useful to bring the L1 data to L2 for example.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -191,9 +191,13 @@ library LibVerifying {
             config.blockMaxGasLimit == 0 ||
             config.maxTransactionsPerBlock == 0 ||
             config.maxBytesPerTxList == 0 ||
+            // EIP-4844 blob size up to 128K
+            config.maxBytesPerTxList > 128000 ||
```
Brechtpd, line 195:
```suggestion
            config.maxBytesPerTxList > 128*1024 ||
```

Brechtpd (approved):
LGTM (minus tiny suggestions)!

Certainly not suggesting we do this now because quite complicated, but there could be some benefit of being able to post data on-chain without submitting a block because the fee markets for gas, the L2 EIP1559 base fee, and the blob data are independent. So it could make sense to publish tx data when blob data is cheap, and then afterwards include them in L2 blocks with a newly specified order when the L1/L2 base fee are low. So pretty complex because it's an onchain mempool and so probably needs a tx market and more flexible way to order txs.

dantaik:
txListCacheExpiry is set to 0 to disable this feature by default

PR #13365, opened 2023-03-20
discussion only: take snapshot at fixed interval for quicker syncing

dantaik:
We'd like to make sure L2 nodes can quickly sync-up with peers using a snapshot. The idea is simple, the client software will take a snapshot of the L2 world state whenever `blockId % snapshotInterval` is zero. New L2 node can query the `lastSnapshotId` then download the snapshot from the peer and this is very likely to succeed.

@davidtaikocha We need to discuss if node software can actually force the snapshot at a given interval.


Brechtpd:
Seems like this is something for the node to decide depending on what the node is trying to be (just validation, archival, etc....). If you connect to the node I guess you'll find out what that node in question supports? Not sure how a smart contract can help with trying to standardize it. I think most Ethereum node implementations already support fast syncing that doesn't require executing all previous transactions, so we can just re-use that one?

dantaik:
>  I think most Ethereum node implementations already support fast syncing that doesn't require executing all previous transactions, so we can just re-use that one?

Yes. But currently all nodes just take quite random snapshots. I will be great if all L2 nodes can take snapshots at given heights so fetch snapshots from peers is more likely to succeed.

The protocol smart contract can never force a node to do anything, so it may not make sense to merge this PR at all (didn't plan to), but maybe the protocol can provide good enough hints so all honest nodes will honor such a snapshot scheduling for new nodes to quickly catch up.

PR #13380, opened 2023-03-21
feat(bridge-ui): Tabs Component

jscriptcoder:
New Tabs component, reusable and fully accessible

### DIFF
--- a/packages/bridge-ui/src/components/Tabs/Tab.svelte
+++ b/packages/bridge-ui/src/components/Tabs/Tab.svelte
@@ -1,13 +1,14 @@
 <script lang="ts">
   import { link } from 'svelte-spa-router';
-  import { setContext, getContext } from 'svelte';
+  import { getContext } from 'svelte';
+  import type { Writable } from 'svelte/store';
 
   export let href: string = '';
   export let name: string = '';
 
-  const activeTab = getContext('activeTab');
+  const activeTab = getContext<Writable<string>>('activeTab');
 
-  $: selected = name === activeTab;
+  $: selected = name === $activeTab;
   $: tabActive = selected ? 'tab-active' : '';
 </script>
 
@@ -15,7 +16,7 @@
   role="tab"
   aria-selected={selected}
   class="tab tab-bordered {tabActive}"
-  on:click={() => setContext('activeTab', name)}
+  on:click={() => ($activeTab = name)}
   {href}
   use:link>
   <slot />
--- a/packages/bridge-ui/src/components/Tabs/Tab.svelte
+++ b/packages/bridge-ui/src/components/Tabs/Tab.svelte
@@ -9,13 +9,13 @@
   const activeTab = getContext<Writable<string>>('activeTab');
 
   $: selected = name === $activeTab;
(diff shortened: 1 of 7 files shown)


dantaik (dismissed):

packages/bridge-ui/src/pages/home/Home.svelte
```diff
@@ -1,17 +1,21 @@
...
 
   let bridgeWidth: number;
   let bridgeHeight: number;
 
   $: activeTab = $location.replace('/', '').startsWith('transactions')
     ? 'transactions'
     : 'bridge';
+
+  // TODO: do we really need all these tricks to style containers
+  //       Rethink this part: fluid, fixing on small screens
```
jscriptcoder, line 18: Typo here, I meant, fixing on bigger screens

packages/bridge-ui/src/pages/home/Home.svelte
```diff
@@ -1,17 +1,21 @@
...
   import { transactions } from '../../store/transactions';
   import BridgeForm from '../../components/form/BridgeForm.svelte';
   import TaikoBanner from '../../components/TaikoBanner.svelte';
   import Transactions from '../../components/Transactions.svelte';
+  import { Tabs, TabList, Tab, TabPanel } from '../../components/Tabs';
 
   let bridgeWidth: number;
   let bridgeHeight: number;
 
   $: activeTab = $location.replace('/', '').startsWith('transactions')
```
cyberhorsey, line 13: this seems brittle, if we change the route name, or if we add another API route like `/transactionsForChain` and only show one chain's transactions, or something.

packages/bridge-ui/src/components/Tabs/Tab.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
+  import { link } from 'svelte-spa-router';
+  import { getContext } from 'svelte';
+  import type { Writable } from 'svelte/store';
+
+  export let href: string = '';
+  export let name: string = '';
+
+  const activeTab = getContext<Writable<string>>('activeTab');
```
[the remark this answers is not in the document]
cyberhorsey: this is actually a good use case for a `Symbol`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol

packages/bridge-ui/src/pages/home/Home.svelte
jscriptcoder: Agree. Let me change that

packages/bridge-ui/src/components/Tabs/Tab.svelte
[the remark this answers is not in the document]
jscriptcoder: That's actually a great idea, but I'd like to keep the encapsulation so I'm gonna expose the key to be accessible across the children:

```
<-- Tabs.svelte -->
<script lang="ts" context="module">
  export const key = Symbol();
</script>
```

cyberhorsey (approved):

dionysuzx (approved):

PR #13381, opened 2023-03-21
fix(bridge-ui): minor ui fixes for alpha 2

shadab-taiko:
- Changes approve button color
- Changes tx dedup logic to avoid duplication between relayer and localStorage in some cases. In future we will revert back to the commented out logic and will instead try to tweak the relayer response for this.
- Fix address display in transaction detail popup

### DIFF
--- a/packages/bridge-ui/src/app.css
+++ b/packages/bridge-ui/src/app.css
@@ -35,16 +35,6 @@
   border-color: hsla(var(--a) / var(--tw-bg-opacity, 1));
 }
 
-.btn.btn-accent.approve-btn {
-  background-color: #4c1d95;
-  border-color: #4c1d95;
-}
-
-.btn.btn-accent.approve-btn:hover {
-  background-color: #5b21b6;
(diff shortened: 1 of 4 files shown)


davidtaikocha (approved):

packages/bridge-ui/src/app.css
```diff
@@ -35,6 +35,16 @@
   border-color: hsla(var(--a) / var(--tw-bg-opacity, 1));
 }
 
+.btn.btn-accent.approve-btn {
+  background-color: #4c1d95;
+  border-color: #4c1d95;
+}
+
+.btn.btn-accent.approve-btn:hover {
```
jscriptcoder, line 43: Why not to add this specific styling into the component?, feels like it should be scoped

jscriptcoder (approved):

packages/bridge-ui/src/app.css
shadab-taiko: Made the change, thanks for pointing it out.

PR #13387, opened 2023-03-22
fix(protocol): fix fork choice error

dantaik:


Brechtpd (approved):
Looks good to me now! Will go over it again tomorrow to double check.

PR #13389, opened 2023-03-22
feat(protocol): major protocol upgrade

dantaik:
This is a PR rebased on top of PR https://github.com/taikoxyz/taiko-mono/pull/13254. If any comments are not resolved there, we will continue to discuss there but new changes will be applied to this PR.

Major Changes:
- [x] **move as much logic from contracts to circuits as possible. Write [an internal doc ](https://www.notion.so/taikoxyz/What-need-to-be-zk-proved-in-a-Taiko-block-c93fdd32b67448a3a11e777be30facec) regarding *what needs to be proven by a ZKP*.**
- [x] remove commit-and-propose scheme
- [x] enable EIP-1559 on L2
- [x] invalid blocks are proven by using zkp as well, just like valid blocks. (invalidatedBlock function is deleted from TaikoL2.sol)
- [x] the anchor transaction validation is deleted from L1 and shall be done by a new circuit on top of the existing ZKP.
- [x] synchronize not only the block header hashes across chains, but also the SignalServcie Storage Roots (signalRoot) -- Currently to prove signal's validity on the destination chain, we need to provide a full merkle proof which has two parts, the account proof and the storage proof.  After this change, the account proof will be done only once in circuits. Users only need to provide the storage merkle proof, which will be less expensive.
- [x] use uint64 instead of uint256 for feeBase
- [x] do NOT refund the proposer if the block is invalid
- [x] require proposers to deposit TKO token first then burn from its balance before a block can be proposed.
- [x] call plonk verifiers directly from LibProving by removing IProofVerifier contracts
- [x] use `string.concat` and `bytes.concat` instead of abi.encodePacked/abi.encode 
- [x] delete a few unused libraries
- [x] disable a lot of protocol tests based on Dave's comment (probably need rewrite)
- [x] support foundry tests and  add a few foundry tests
- [x] rename `config.rewardMultiplierPctg` to `config.feeMultiplierPctg` to be consistent with other config vars.
- [x] fix math in `getTimeAdjustedFee`
- [x] validate configs in `TaikoL1.init()`
- [x] baseFee is now a uint64 represent *Twei*, not *Szabo*.
- [x] a few optimization using `unchecked` -- **Need your careful review and feedback**
- [x] allow prover/proposer to have independent configs for dynamic fees
- [x] optimized gas cost for L2's anchor transaction


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -6,372 +6,194 @@
...
-                evidence.meta.extraData.length ||
-                keccak256(evidence.header.extraData) !=
-                keccak256(evidence.meta.extraData) ||
-                evidence.header.mixHash != evidence.meta.mixHash
-            ) revert L1_META_MISMATCH();
+            unchecked {
+                ++blk.nextForkChoiceId;
+            }
+        } else if (fcId >= blk.nextForkChoiceId) {
+            revert L1_UNEXPECTED_FORK_CHOICE_ID(); // this shall not happen
```
adaki2004, line 78: If i read the flow - how i see (and that might be totally wrong) - this revert can only be reached in 1 scenario:
1. Proof submitted and `maxVerificationsPerTx > 0` (meaninig the the prover will / might verify in the same TXN). This way the `blk.nextForkChoiceId` will be re-setted to 1 in `verifyBlock()`.
2. Then, some other prover wants to prove the same proposed block (tho is already proven and verified) with `proveBlock() `but this time the fcId will be = 1 while the the blk.nextForkChoiceId will also be 1.

So basically this line says: **'Hey Prover, this proposed block is already proven and verified !'**

Am I getting this right ? Also what i don't see is a scenario, where the fcId could be bigger then blk.nextForkChoiceId. (Only equal, in the above mentioned situation).

Brechtpd: I think everything you said is correct.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -6,109 +6,155 @@
...
+        FeeConfig proposingConfig;
+        FeeConfig provingConfig;
+    }
+
+    struct StateVariables {
+        uint64 feeBaseTwei;
+        uint64 genesisHeight;
+        uint64 genesisTimestamp;
+        uint64 nextBlockId;
+        uint64 lastBlockId;
```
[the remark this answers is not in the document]
dantaik: replaced everywhere.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -6,238 +6,218 @@
...
             ];
-            TaikoData.ProposedBlock storage target = state.getProposedBlock(
-                config.maxNumBlocks,
-                i
-            );
 
-            if (fc.prover == address(0)) {
+            uint256 fcId = state.forkChoiceIds[i][blockHash];
+
+            if (blk.nextForkChoiceId <= fcId) {
```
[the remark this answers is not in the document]
dantaik: Changed as suggested

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -6,109 +6,155 @@
...
+    // 4 slots
     struct ProposedBlock {
         bytes32 metaHash;
         uint256 deposit;
         address proposer;
         uint64 proposedAt;
+        uint24 nextForkChoiceId;
+        // ForkChoice storage are reusable
+        mapping(uint256 forkChoiceId => ForkChoice) forkChoices;
     }
```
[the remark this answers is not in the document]
dantaik: Please see: https://github.com/taikoxyz/taiko-mono/pull/13427/files
The idea is indeed to make sure the ring buffer is bigger than the max number of pending blocks.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -6,238 +6,206 @@
...
-            config.rewardMultiplierPctg < 100
-        ) revert L1_INVALID_CONFIG();
-
-        if (feeBase == 0) revert L1_0_FEE_BASE();
+        _checkConfig(config);
 
-        state.genesisHeight = uint64(block.number);
-        state.genesisTimestamp = uint64(block.timestamp);
+        uint64 timeNow = uint64(block.timestamp);
+        state.genesisHeight = timeNow;
```
davidtaikocha, line 39:
```suggestion
        state.genesisHeight = uint64(block.number);
```

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -6,110 +6,145 @@
...
+        bytes32 txListHash;
+        address beneficiary;
+        uint32 gasLimit;
+        uint24 txListByteStart; // byte-wise start index (inclusive)
+        uint24 txListByteEnd; // byte-wise end index (exclusive)
+        uint8 cacheTxListInfo; // non-zero = True
     }
 
+    // 6 slots
+    // Changing this struct requires chaing LibUtils.hashMetadata accordingly.
```
davidtaikocha, line 63:
```suggestion
    // Changing this struct requires changing LibUtils.hashMetadata accordingly.
```

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -6,110 +6,145 @@
...
     struct BlockMetadata {
-        uint256 id;
-        uint256 l1Height;
+        uint64 id;
+        uint64 timestamp;
+        uint64 l1Height;
+        uint64 l2Basefee;
```
davidtaikocha, line 68: no longer used

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -120,25 +182,53 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, IXchainSync {
...
+            chainid: uint32(block.chainid)
         });
     }
 
     /**********************
      * Public Functions   *
      **********************/
 
+    function getBasefee(
+        uint32 timeSinceNow,
```
adaki2004, line 221: What is purpose of having the `timeSinceNow` ?
The name little bit confusing to me and also the fact that the `_getBasefeeAndPrint()` (which calls `getBasefee()` ) called everywhere with 0.

packages/protocol/contracts/L1/TaikoData.sol
dantaik: this will be deleted in Dani's PR.

dantaik: deleted

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: You can get the basefee for a future time, assuming you will propose a block in 24 seconds, you can use 24 as the value of `timeSinceNow`.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -30,50 +42,95 @@ contract TaikoL2 is AddressResolver, ReentrancyGuard, IHeaderSync {
...
+            (_xscale, yscale) = Lib1559Math.calculateScales({
+                xExcessMax: _param1559.gasExcessMax,
+                price: _param1559.basefee,
+                target: _param1559.gasTarget,
+                ratio2x1x: _param1559.ratio2x1x
+            });
+
+            if (_xscale == 0 || _xscale >= type(uint64).max || yscale == 0)
+                revert L2_INVALID_1559_PARAMS();
+            xscale = uint64(_xscale);
```
Brechtpd, line 118: Maybe no scaling needed: https://github.com/taikoxyz/taiko-mono/pull/13501#issuecomment-1492671010
Seems to work okay on the prover fee side, the accuracy of `exp` also doesn't really matter much because errors are not accumulated over blocks.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -30,50 +42,95 @@ contract TaikoL2 is AddressResolver, ReentrancyGuard, IHeaderSync {
...
+
+            // basefee = _param1559.basefee;
+            gasIssuedPerSecond = _param1559.gasIssuedPerSecond;
+            gasExcess = _param1559.gasExcessMax / 2;
```
Brechtpd, line 122: Not sure how big `gasExcessMax`, but seems like it should be a pretty reasonable small value otherwise `gasExcess` will produce a very high basefee that will only go down very slowly (potentially never). Maybe just using `0` here safer?

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -173,51 +257,53 @@ contract TaikoL2 is AddressResolver, ReentrancyGuard, IHeaderSync {
...
+
+        uint256 __basefee = Lib1559Math.calculatePrice({
+            xscale: xscale,
+            yscale: yscale,
+            xExcess: _gasExcess,
+            xPurchase: gasLimit
+        });
+
+        // Very important to cap basefee uint64
+        _basefee = uint64(__basefee.min(type(uint64).max));
```
Brechtpd, line 307: No reason anymore I think because we don't store this value anymore.

Brechtpd (dismissed):
Just some comments from other PRs that I pasted in here so they don't get lost.

Because the diff of this PR diff is too big, with some pending PRs that will still do some significant changes to some parts. I agree it's a good idea to just merge this in.

cyberhorsey (dismissed):

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -86,77 +143,104 @@ contract TaikoL2 is AddressResolver, ReentrancyGuard, IHeaderSync {
...
-     */
-    function invalidateBlock(
-        bytes calldata txList,
-        LibInvalidTxList.Hint hint,
-        uint256 txIdx
+    function anchor(
+        bytes32 l1Hash,
+        bytes32 l1SignalRoot,
+        uint64 l1Height,
+        uint64 parentGasUsed
```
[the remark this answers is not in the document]
dantaik: Unlike `block.gaslimit`, *current block's total gas used* is not available inside the block.

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: Create an issue for it: https://github.com/taikoxyz/taiko-mono/issues/13556

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: I think maybe these two options are mathematically identical, one is to set the initial basefee to a best-guess value, the other chooses 0 as the value then wait it to go up gradually.

dantaik: After the initial deployment, if not a lot of gas is used, then the price will gradually decrease to zero, your understanding is correct.

packages/protocol/contracts/bridge/libs/LibBridgeProcess.sol
```diff
@@ -129,9 +126,7 @@ library LibBridgeProcess {
                 status = LibBridgeStatus.MessageStatus.DONE;
             } else {
                 status = LibBridgeStatus.MessageStatus.RETRIABLE;
-                if (ethVault != address(0)) {
-                    ethVault.sendEther(message.callValue);
-                }
+                ethVault.sendEther(message.callValue);
```
cyberhorsey, line 129: do we not burn the `message.callValue` now, because ethVault can be 0 address?

PR #13390, opened 2023-03-22
refactor(protocol): check if amount > 0 inside sendEther

dantaik:
It's quite weird that checking `amount>0` inside the sendEther function is actually cheaper...how come


Brechtpd (approved):
> It's quite weird that checking amount>0 inside the sendEther function is actually cheaper...how come

Was not expecting that, I thought it would be the same at best. :) The only possible explanation I can think of is that the function call is inlined, and so the compiler can do some extra tiny optimizations because of that.

PR #13426, opened 2023-03-23
fix(protocol): allow deposit Ether to any address on L2

dantaik:
@Dani | Taiko @Daniel | Taiko i think we have a bug. ourBBridge cant actually allow send ETH to custom recipient.
because 
LibBridgeProcess.sol

        if (message.depositValue > 0) {
            // we use message.owner here, not message.to
            message.owner.sendEther(message.depositValue);
        }

so the workaround, in oru code, is to set no depositValue, and set the message.callValue when sending ETH to an address that is diffeernt than the person who owns the message.

            bool success = LibBridgeInvoke.invokeMessageCall({
                state: state,
                message: message,
                msgHash: msgHash,
                gasLimit: gasLimit
            });

which does:

  (success, ) = message.to.call{value: message.callValue, gas: gasLimit}(
            message.data
        );

All fine and dandy so far, just set callValue and to appropriately.
Except in TokenVault.sol, sendEther method:

        // prevent future PRs from changing the callValue when it must be zero
        if (message.callValue != 0) revert TOKENVAULT_INVALID_CALL_VALUE();

so callValue can never be set for sendEther.

### DIFF
--- a/packages/protocol/contracts/bridge/TokenVault.sol
+++ b/packages/protocol/contracts/bridge/TokenVault.sol
@@ -140,6 +140,10 @@ contract TokenVault is EssentialContract {
     /**
      * Receives Ether and constructs a Bridge message. Sends the Ether and
      * message along to the Bridge.
+     *
+     * @dev This function doesn't' seem to belong here as it has nothing to
+     *      do with ERC20 tokens. It's added here only for convenience.
+     *
      * @param destChainId @custom:see IBridge.Message
      * @param to @custom:see IBridge.Message
      * @param gasLimit @custom:see IBridge.Message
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/bridge/TokenVault.sol
```diff
@@ -168,12 +167,10 @@ contract TokenVault is EssentialContract {
         message.to = to;
         message.gasLimit = gasLimit;
         message.processingFee = processingFee;
-        message.depositValue = msg.value - processingFee;
+        message.callValue = msg.value - processingFee;
         message.refundAddress = refundAddress;
         message.memo = memo;
-
-        // prevent future PRs from changing the callValue when it must be zero
-        if (message.callValue != 0) revert TOKENVAULT_INVALID_CALL_VALUE();
```
dantaik, line 176: I think I added this comment to prevent both `depositValue` and `callVaue` to be set to `msg.value - processingFee` to cause a double-deposit.

cyberhorsey:
Does `.depositValue` serve any purpose at all anymore, if we want to transfer ETH via `LibBridgeInvoke` with the `message.to` and `message.callValue` ?

davidtaikocha (approved):

adaki2004 (approved):

PR #13429, opened 2023-03-23
fix(bridge-ui): alpha-2 hotfixes

shadab-taiko:
- Fixes `eth_chainid` call spam
- Approve button text color in light theme
- Shows `Failed` status if the bridge transaction failed at source chain

### DIFF
--- a/packages/bridge-ui/src/components/Transaction.svelte
+++ b/packages/bridge-ui/src/components/Transaction.svelte
@@ -185,7 +185,7 @@
 
       transaction.status = await contract.getMessageStatus(transaction.msgHash);
 
-      if (transaction.receipt.status !== 1) {
+      if (transaction.receipt && transaction.receipt.status !== 1) {
         clearInterval(interval);
         return;
       }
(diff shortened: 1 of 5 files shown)


davidtaikocha (approved):

packages/bridge-ui/src/components/Transaction.svelte
```diff
@@ -241,7 +246,9 @@
   <td>
     <ButtonWithTooltip onClick={() => onTooltipClick(false)}>
       <span slot="buttonText">
-        {#if !processable}
+        {#if transaction.receipt && transaction.receipt.status !== 1}
```
jscriptcoder, line 249: What status is `1`? don't we have a enum with these status?

packages/bridge-ui/src/components/Transaction.svelte
```diff
@@ -185,6 +185,11 @@
 
       transaction.status = await contract.getMessageStatus(transaction.msgHash);
 
+      if (transaction.receipt.status !== 1) {
```
jscriptcoder, line 188: How about using an enum?

jscriptcoder (approved):

PR #13442, opened 2023-03-24
feat(protocol): change time-based fee/reward calculation

dantaik:
The current proving fee reward is only lowered if the proposers collectively wait longer to express their unwillingness to pay the current fees. It lacks a way to allow for competition among provers -- some provers may just want to get lower reward than others (probably because their proving cost is also smaller).

This PR change the time-based dynamic fee calculation with a new parameter called `startBips`. Assuming it has a value of 4000 bips (40%):

1. For provers: if a proof is submitted with 0 delay, the prover will get `60%*feeBase`, if the proof is submitted at anytime `>= 2*averageProofTime`, the prover will get `140*feeBase`, anytime in the middle is linear. 
2. For proposers: if a block is proposed with 0 delay, the fee will be `140*feeBase`, and if the proposal is at or later than `2*averageBlockTime`, the fee will be `60%*feeBase`.

(_In the future, with auctions, we may want to allow someone to win the block if 1) he is willing to deposit for that block at least 20% higher than the current winner's deposit, or 2) he is willing to give up (burn) at least 5% reward than the current winner. For example, Alice wins the block saying she is willing to get only 95% of rewards (whatever it is), then Bob winds the block saying he is willing to get only 95% *95%=90.25% of the reward, or he can be aggressive enough to give more more as 95% *90%= 85.5%._)

I changes the implementation on top of [major_protocol_upgrade_rebase] to  test the math using foundry.

@Brechtpd @cyberhorsey @davidtaikocha If you feel this is a good fix to the current problem, I'll create another PR aginst the alpha-2 branch.


packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -180,26 +180,23 @@ library LibTokenomics {
...
+
+            newFeeBase = (feeBase * (10000 + (t * p) / a - p)) / 10000;
 
             if (isProposal) {
-                newFeeBase = (feeBase * 10000) / alpha; // fee
-            } else {
-                newFeeBase = (feeBase * alpha) / 10000; // reward
+                newFeeBase = (feeBase * 2) - newFeeBase;
+            } else if (p > 0) {
+                premiumRate = ((t.max(a) - a) * 10000) / a;
```
adaki2004, line 199: I have a question to this formula.
Let's say we have 2 scenarios. (In both cases the `startBips` is `2500` = 25% and the `feeBase` is `10`)

1. Quciker than avg (10s )
Average time is 20s (`timeAverage=20.000 ms`) and elapsed time is 10s (`timeUsed = 10.000 ms`). 
Then:
t = (10.000ms ) MIN (20.000 * 2 = 40.000ms) : 10.000MS
newFeeBase = (10 * (10.000 + (10.000 * 2.500 ) / 20.000 - 2.500 ) / 10.000 =  11.42

And in case of proposal it is: `newFeeBase = (10 * 2 ) - 11.42` -> 8.57

2. Slower than avg (50s )
Average time is 20s (`timeAverage=20.000 ms`) and elapsed time is 50s (`timeUsed = 150.000 ms`). 
Then:
t = (50.000ms ) MIN (20.000 * 2 = 40.000ms) : 40.000MS
newFeeBase = (10 * (10.000 + (40.000 * 2.500 ) / 20.000 - 2.500 ) / 10.000 =  15.72

And in case of proposal it is: `newFeeBase = (10 * 2 ) - 15.72` -> 4.28

If the calcs are correct it means the initial baseFee can only decrease or stagnate.
Are my calculations correct ?

packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -180,26 +180,23 @@ library LibTokenomics {
...
-            uint256 max = (feeConfig.maxPeriodPctg * tAvg) / 100;
-            uint256 grace = (feeConfig.gracePeriodPctg * tAvg) / 100;
-            uint256 t = ((tNow - tLast) * 1000).max(grace).min(max);
-            tRelBp = (10000 * (t - grace)) / (max - grace); // [0-10000]
-            uint256 alpha = 10000 + (tRelBp * feeConfig.multiplerPctg) / 100;
+            uint p = feeConfig.startBips; // [0-10000]
+            uint a = timeAverage;
+            uint t = (timeUsed * 1000).min(a * 2); // millisconds
+
+            newFeeBase = (feeBase * (10000 + (t * p) / a - p)) / 10000;
```
Brechtpd, line 194: I still find this formula very tricky to really understand. If not using one of the other common formula's to calculate something like, why can't this be as simple as something like this (disregarding rounding problems, so needs some additional factors):
```
newFeeBase = feeBase * ((timeUsed / timeAverage) / SMOOTHING_FACTOR);
```

Brechtpd:
> For provers: if a proof is submitted with 0 delay, the prover will get 60%*feeBase, if the proof is submitted at anytime >= 2*averageProofTime, the prover will get 140*feeBase, anytime in the middle is linear.

What if `140*feeBase` is not enough? Having a hard cap seems a bit dangerous.

> For proposers: if a block is proposed with 0 delay, the fee will be 140*feeBase, and if the proposal is at or later than 2*averageBlockTime, the fee will be 60%*feeBase.

Should be easier I think to simply update the base fee first, then the proposer simply pays the fee based on that (or I could be misunderstanding this).

packages/protocol/contracts/L1/libs/LibTokenomics.sol
dantaik: - scenario 1 (at time 10sec):
    - fee for proposer is 11.25, 
    - reward for prover is 8.65.
- scenario 2 (at time 50sec):
    - fee for proposer is 7.5
    - reward for prover is 12.5.

packages/protocol/contracts/L1/libs/LibTokenomics.sol
dantaik: Brecht, I think the implementation is exactly the same as the one you suggested.
`(t * p) / a - p)` = `p * (t/a - 1)`

`(10000 + (t * p) / a - p)) / 10000` = `(1 + p * (t/a - 1))`

p is the DAMPING_FACTOR, t is timUsed, a is timeAvareege.

dantaik:
> What if `140*feeBase` is not enough? Having a hard cap seems a bit dangerous.

Not sure if 140% is good enough. But now somehow I feel a bit reluctant to allow the fees/rewards to have a potential to go up infinitely or down to zero in one block.

> Should be easier I think to simply update the base fee first, then the proposer simply pays the fee based on that (or I could be misunderstanding this).

If we do that, the fee/reward for the next block will only change very slightly depends on the value of the average smoothing factor.

packages/protocol/contracts/L1/libs/LibTokenomics.sol
Brechtpd: Ah haha didn't realize they were that close. Why is it the damping factor called `startBips` (I though it was also used for something else, didn't know why it was also used here)?

Brechtpd:
> Not sure if 140% is good enough. But now somehow I feel a bit reluctant to allow the fees/rewards to have a potential to go up infinitely or down to zero in one block.

Yeah for sure, though what a single prover gets I'm not really that concerned about (except on the infinity side of course :). But I think we now know a couple of ways that should allow us to avoid inconvenient scenario's in a way that doesn't depend on hard to predict fixed values (still remains to be seen of course).

> If we do that, the fee/reward for the next block will only change very slightly depends on the value of the average smoothing factor.

...that's good right?

In any case, this is already a very nice simplification! I believe we can go quite a bit further though (like completely decoupling the proving fee from how blocks are proposed).

packages/protocol/contracts/L1/libs/LibTokenomics.sol
dantaik: startBips is not a good name, `dampingFactor` instead? It is only used here in the time-based fee calculation.

dantaik:
> > Not sure if 140% is good enough. But now somehow I feel a bit reluctant to allow the fees/rewards to have a potential to go up infinitely or down to zero in one block.
> 
> Yeah for sure, though what a single prover gets I'm not really that concerned about (except on the infinity side of course :). But I think we now know a couple of ways that should allow us to avoid inconvenient scenario's in a way that doesn't depend on hard to predict fixed values (still remains to be seen of course).
> 
> > If we do that, the fee/reward for the next block will only change very slightly depends on the value of the average smoothing factor.
> 
> ...that's good right?
> 
> In any case, this is already a very nice simplification! I believe we can go quite a bit further though (like completely decoupling the proving fee from how blocks are proposed).

I think auction will change how we adjust fees/rewards. I'll be working on that with Jeff soon. Hopefully auction is available in two weeks for review.

Brechtpd (approved):
Approving this one as future PRs will address additional concerns.

davidtaikocha (approved):

PR #13454, opened 2023-03-25
feat(protocol): calculate EIP1559 base fee on L1.

dantaik:
- the L2's EIP-1559 `basefee` is now [calculated using the AMM style math](https://ethresear.ch/t/make-eip-1559-more-like-an-amm-curve/9082) on L1 and is made part of the block metadata. Note that `basefee` for the next block only depends on the parent block ,not the block being proposed. This is to ensure `basefee` for next block is deterministic so users can chose their transaction settings correctly. Also note that this is not to be confused with the proposer/prover fees/rewards (`feeBase`) in TaikoToken. `basefee` is an Ether amount.
- Implementation is incomplete and tests are not written, please review first before I spent more time on it.
- @davidtaikocha In taiko-client, we need to make sure the EIP1559 base fees are not burnt, but send to the block beneficiary address together with 1559 tips. 
- After this PR: proposers pays:
    1. Ether as transaction fees to L1 validators to include their transactions
    2. Ether to Taiko DAO to buy block space, the amount to send is `baseCharge1559` in the code, which is `basefee * gasLimit`.
    3. Taiko Token to provers for proving their blocks.

@Brechtpd  please let me know if you want to have a quick chat.

### DIFF
--- a/packages/protocol/contracts/L1/libs/Lib1559.sol
+++ b/packages/protocol/contracts/L1/libs/Lib1559.sol
@@ -8,56 +8,66 @@ pragma solidity ^0.8.18;
 
 import {AddressResolver} from "../../common/AddressResolver.sol";
 import {LibMath} from "../../libs/LibMath.sol";
-// import {
-//     SafeCastUpgradeable
-// } from "@openzeppelin/contracts-upgradeable/utils/math/SafeCastUpgradeable.sol";
 import {TaikoData} from "../TaikoData.sol";
-
-// import {TaikoToken} from "../TaikoToken.sol";
 
 library Lib1559 {
     using LibMath for uint256;
 
-    error L1_BLOCK_GAS_LIMIT_TOO_LARGE();
-
-    function get1559BurnAmountAndBaseFee(
+    function getGasFeeStatus(
+        TaikoData.State storage state,
         TaikoData.Config memory config,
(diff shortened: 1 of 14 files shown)


packages/protocol/contracts/L1/libs/Lib1559.sol
```diff
@@ -0,0 +1,77 @@
...
+    )
+        internal
+        pure
+        returns (
+            uint64 basefeePerGas,
+            uint256 gasPurchaseCost,
+            uint256 newGasExcess
+        )
+    {
+        uint256 t = config.blockGasTarget;
```
Brechtpd, line 61: I think in our case this needs to be `blockGasThrottle` *not the per block gas limit on L2) because that's our actual target. How many L2 blocks it takes to get there (and how much gas used by each) doesn't seem important. L2 blocks don't really have much meaning for us.

Brechtpd (commented):
>  This is to ensure basefee for next block is deterministic so users can chose their transaction settings correctly.

Interesting, could be useful but the dynamics at work could be difficult. Like if everybody sets the base fee to the expected value, I guess it's just a game of highest priority fee winning again? (which would be less convenient for us because that means we can capture less value, a higher worst case basefee and lower priority fee may be preferable).

dantaik:
Actually I will propose something different.

PR #13457, opened 2023-03-25
Add a new Sepolia faucet

StarryNight777:
I see that some community members are missing taps
Please add this site to our faucet
Each person can receive 0.5 ETH per day, but only if they are a previously registered member
https://sepoliafaucet.com/


davidtaikocha:
duplicated to https://github.com/taikoxyz/taiko-mono/pull/13453

PR #13459, opened 2023-03-25
refactor(protocol): various small improvements

dantaik:
Also removed gasUsed and anchorTxGasLimit.

- anchorTxGasLimit: anchor will be a special TX type, the anchorTxGasLimit will be checked in circuits
- gasUsed is not really needed on L1 contracts


packages/protocol/contracts/L1/libs/LibL1Tokenomics.sol
```diff
@@ -14,7 +14,7 @@ import {
 import {TaikoData} from "../TaikoData.sol";
 import {TaikoToken} from "../TaikoToken.sol";
 
-library LibTokenomics {
+library LibL1Tokenomics {
```
dantaik, line 17: Change name so we can also have a LibL2Tokenomics to calculate L2 EIP-1559 fees.

davidtaikocha (approved):

adaki2004 (approved):

PR #13463, opened 2023-03-26
should change "Claimable" to "Claim" in message status? because on transactions, it shows "claim"

nothingholics:
屏幕截图 2023-03-26 202558
屏幕截图 2023-03-24 214422


cyberhorsey:
PR should not be alpha 2 to main, I can make this change separately.

PR #13468, opened 2023-03-27
feat(protocol): calculate L2 EIP-1558 basefee on L1

dantaik:


packages/protocol/contracts/libs/LibRealMath.sol
```diff
@@ -0,0 +1,13 @@
...
+//  _____     _ _         _         _
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+
+pragma solidity ^0.8.18;
+
+library LibRealMath {
+    function exp(uint128 b) internal pure returns (uint256) {
+        // https://github.com/NovakDistributed/macroverse/blob/master/contracts/RealMath.sol
```
Brechtpd, line 11: This one from Remco seems nice: https://github.com/recmo/experiment-solexp/blob/main/src/FixedPointMathLib.sol

I think my previous implementation is probably not good enough because it was only accurate around a single point (but maybe using a lot of rounds would still be okay). I'm not sure what Remco's implementation uses, doesn't really look like Taylor.

packages/protocol/contracts/L1/libs/LibL2Tokenomics.sol
```diff
@@ -0,0 +1,78 @@
...
+            return (gasExcess, 0, 0);
+        }
+        unchecked {
+            uint256 _gasExcess = gasTargetPerSecond * blockTime + gasExcess;
+
+            _gasExcess = _gasExcess.min(type(uint64).max);
+
+            if (gasInBlock >= _gasExcess) revert L1_OUT_OF_BLOCK_SPACE();
+
+            newGasExcess = uint64(_gasExcess - gasInBlock);
```
Brechtpd, line 67: Accumulates gas issued instead of extra gas used like in Vitalik's post. So if the chain is not used a lot for a long time this will keep accumulating which should be fine, but not sure if this is indeed what we want. I'm not sure the math checks out when accumulating the gas like this (I guess that may depend on the curve).

packages/protocol/contracts/L1/libs/LibL2Tokenomics.sol
```diff
@@ -0,0 +1,78 @@
...
+        uint32 gasInBlock,
+        uint256 blockTime
+    )
+        internal
+        pure
+        returns (uint64 newGasExcess, uint64 basefee, uint256 gasPurchaseCost)
+    {
+        if (gasInBlock == 0) {
+            return (gasExcess, 0, 0);
+        }
```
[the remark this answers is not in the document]
dantaik: > Any reason to have this special case?
To avoid divide-by-zero here `uint256 _basefee = gasPurchaseCost / gasInBlock;`

packages/protocol/contracts/libs/LibRealMath.sol
dantaik: Using e-curve will be tried on a different PR.

Brechtpd (approved):

PR #13488, opened 2023-03-28
feat(bridge-ui): more informative processing fee

### LINKED ISSUE #13476: feat(bridge-ui): UI/UX suggestion to better inform the user

I put together a few screens that I think would help to better inform users of what to expect when bridging. It seems that most of them are not using the tool tips and don't realize:

1. relayer will auto-claim if fee is set to Recommended
2. user needs to manually claim (with ETH on receiving chain) if fee is set to None

Descriptions of each screenshot:

Screenshot #1 - When user selects "None" for fee, show a pop-up message

Screenshot #2 - Change the Claim button to fluro background and white text, to match the button style of "Bridge" on the main site. Should make it look more like a button and good to be consistent

Screenshot #3 - If you selected Recommended and they click "Claim" show pop-up to remind them they should wait for the relayer to auto-claim

Bridge

N/A

### Additional context

Happy to help modify them if needed, but I'm sure the UI/UX team will knock it out of the park.

jscriptcoder:
This PR addresses the following suggestions: https://github.com/taikoxyz/taiko-mono/issues/13476. Besides that there a few more things:
- Type improvements:
  - Transaction data was a record of `any`. I added the correct types for this field.
  - Some chain ids were `BigNumber `type, which I found it unnecessary. They can simply be `number`.
- Added a new component, `NoticeModal`, which presents the user with some message and the option to not seeing it again.
- Grouped Transactions part, and done some improvements in the `Transaction` component. Still a lot to do here. I believe it's quite bloated. We could perfectly extract a lot of logic outside and also up in its parent such as claiming and releasing functions.
- ProcessingFee component and related components have been grouped (we could move this to the root folder components)
  - Got rid of processingFee store. This is simply a bound option between parent and child. No need to be part of the global state.
- Created a Button component to be consistent across the bridge, following a more declarative way to build UI. It's quite basic though. We need a design system here.
- Other minor changes


packages/bridge-ui/src/store/fee.ts
jscriptcoder, line 1: We simply don't need this at all. It's a bound prop

jscriptcoder:
Inform about None processing fee:
https://user-images.githubusercontent.com/613724/228312077-d9535aa1-17f6-43d3-bd95-5bf328e35fcc.mov

jscriptcoder:
Inform about Claim when Recommend or Custom fee has been selected
https://user-images.githubusercontent.com/613724/228314037-154c2757-8094-4966-b96b-d7492806ddae.mov

packages/bridge-ui/src/components/Transactions/MessageStatusTooltip.svelte
```diff
@@ -12,14 +13,9 @@
...
           <strong>Pending</strong>: Your asset is not ready to be bridged. Taiko
-          A2 => {import.meta.env
-            ? import.meta.env.VITE_MAINNET_CHAIN_NAME
-            : 'Ethereum A2'} bridging can take several hours before being ready.
-          {import.meta.env
-            ? import.meta.env.VITE_MAINNET_CHAIN_NAME
-            : 'Ethereum A2'} => {import.meta.env
-            ? import.meta.env.VITE_TAIKO_CHAIN_NAME
-            : 'Taiko A2'} should be available to claim within minutes.
+          A2 => {L1_CHAIN_NAME} bridging can take several hours before being ready.
```
cyberhorsey, line 16: Shouldn't this be {L2_CHAIN_NAME}

packages/bridge-ui/src/components/Transactions/Transaction.svelte
```diff
@@ -161,6 +169,7 @@
     }
   }
 
+  // TODO: this could also live in an utility: isTransactionProcessable?
   async function isProcessable() {
     if (!transaction.receipt) return false;
```
cyberhorsey, line 174: Let's move it to utils and write a test

packages/bridge-ui/src/components/Transactions/Transaction.svelte
```diff
@@ -274,29 +283,35 @@
...
+                  We keep on accessing transaction props
+                  over and over again.
+          -->
+          <Button
+            type="accent"
+            size="sm"
+            on:click={onClaimClick}
             disabled={transaction.status === MessageStatus.ClaimInProgress}>
             Claim
-          </button>
```
cyberhorsey, line 282: What is MessageStatus.ClaimInProgress?the MessageStatuses map to MessageStatuses from the contract, and this isn't one?

packages/bridge-ui/src/domain/button.ts
```diff
@@ -0,0 +1,5 @@
+// TODO: still to figure out after design
+
+export type TypeButton = 'primary' | 'secundary' | 'accent' | ''; // TODO: daisyUI?
```
cyberhorsey, line 3: secondary is mispelled

cyberhorsey (dismissed):

packages/bridge-ui/src/components/Transactions/Transaction.svelte
jscriptcoder: Apparently this is for UI, but I'm not very sure. Maybe @shadab-taiko can answer that. I'll investigate how this is used though...

packages/bridge-ui/src/domain/button.ts
jscriptcoder: Corrected. Thanks

packages/bridge-ui/src/components/Transactions/MessageStatusTooltip.svelte
jscriptcoder: Indeed... Corrected. Thanks

dantaik (approved):

packages/bridge-ui/src/components/Transactions/Transaction.svelte
shadab-taiko: @cyberhorsey The main idea was to disable the Claim button if the user has already clicked on it once. This wasn't meant to be the final solution for this since this would change once we start tracking claim transactions.

shadab-taiko (approved):

PR #13489, opened 2023-03-28
feat(protocol): Calculate fee/reward based on proof time w/ 1559 math

adaki2004:
## Proposer fee /prover reward balanced algorithm

This PR introduces an implementation of the proposer fee / prover reward equilibrium calculation, aiming to balance deposits and withdrawals. The implemented (new `baseFeeProof`) calculation depends on the proving time and utilizes the following parameters:

    proofTimeTarget: A preset of the average proving time (shall be overestimated-> cause no harm)
    gasConsumed: The consumed L2 gas per block (currently set to the gasLimit).
    adjustmentQuotient: The quotient used for adjusting base fees (16).

The base fee proof serves as a dynamic fee that adapts to the network conditions, aiming to keep the inflow and outflow of tokens in balance.

See issue discussion: 
https://github.com/taikoxyz/internal/issues/121
(Especially see Brecht's comment [here](https://github.com/taikoxyz/internal/issues/121#issuecomment-1487776906), with fixed proof time target (Target 1) and 'Mint new tokens when needed' approach - in reality the deposits and withdrawals will be balanced.)

This is an experiment where the base idea is to have a simplified fee / reward calculation structure with TKO for provers and proposers and depends on proving time.

#### Notes:
- If `targetProofTime` is far (above) from the actual proof time, the system will need more time to reach equilibrium when the time starts stabilizing - (and need to be stabilized **below** this **target** 
- If proof time is very stable but a tiny bit above average, it will increase the baseFeeProof ( see foundry test `test_reward_and_fee_if_proof_time_stable_but_above_time_target`) -> Setting the `targetProofTime` above the actual edge (where we expect the proof times) does not harm at all, because if proof times are below that, the equilibrium will be reached (deposits ≈ withdrawals) anyhow at some point.

What remains to be done in this PR:
- [x] Implement pooling mechanism to distribute rewards based on deposits - in case `mintAllowed = false`
- [x] Implement test cases when `mintAllowed = false`
- [x] rebase (lots of merge conflicts with the base `L2_1559_rebase`)
- [ ] address failing HH / Typescript tests (and other interface changes)
- [ ] search for more gas optimization


adaki2004:
@Brechtpd @dantaik 
So here is a draft version from what we just talked about [here](https://github.com/taikoxyz/internal/issues/121).

This is the version of **race 2** (moving average is the target) however I'd have some questions after doing some benchmark tests (python code vs. solidity/foundry tests).
Tendencies (python code vs this) seems to be the same(slower time of proof -> higher fee and reward) but for example the reward - ratio is very weird in both.

 '`proof reward`' vs. the '`proposer_fee`' are not on the same 'level'.

Here is an example logs i extracted from my test:
(you can run if you pull this branch and `forge build` &&` forge test -vv`)

### Solidity
```
Proving time increases:

1.
Proposer fee:         0
FeeBase :              0
Prover reward: 3320,31250000 (TKO)

2.
Proposer fee:           425000,00000000 (TKO)
FeeBase:              42500000
Prover reward:             7080,07812500 (TKO)

3.
Proposer fee:           453125,00000000 (TKO)
FeeBase:              45312500
Prover reward:             9624,38648001 (TKO)

4.
Proposer fee:           451704,54000000
FeeBase:              45170454
Prover reward:             10780,03874648 (TKO)


Proving time decreases:

1.
Proposer fee:         0
FeeBase :              0
Prover reward:      21240,23437500 (TKO)

2.
Proposer fee:           543750,00000000 (TKO)
FeeBase:              54375000
Prover reward:            523,43750000 (TKO)

3.
Proposer fee:           83750,00000000 (TKO)
FeeBase:              8375000
Prover reward:             402,66423264 (TKO)

4.
Proposer fee:           83333,33000000 (TKO)
FeeBase:              8333333
Prover reward:             288,94990704 (TKO)
```

So as you see the TKO as fee is much higher, than the reward. 
I do not know if this is within the math, but i somewhat conclude both (so in the math somewhere, and also in my code. The latter also comes from the fact that solidity has no floating point operations, so for example 1.001 - 1.99 exp is always 1. I tried to create a workaround for it, and use higher precision, this is also within this branch - please see `EXPONENTIAL_REWARD_FACTOR`).

So the python script from [here](https://github.com/taikoxyz/internal/issues/121#issuecomment-1487776906) has same kind of an issue when i run them it shows 6-8 x between fee and reward):

### Python
```
print("Quciker proving proving")
propose_block(50000000)
time = 400
prove_block(50000000)
propose_block(50000000)
time = 600
prove_block(50000000)
propose_block(50000000)
time = 650
prove_block(50000000)
propose_block(50000000)
time = 675
prove_block(50000000)
```

The results are:

```
Quciker proving proving
******* propose ********
network fee: 0.3668379411737963
prover fee: 500.00000000000006
basefee_network: 8.542737132336227e-09
******* prove ********
proof_time: 400
prover reward: 526700.2765236852
basefee_proof: 0.0015801008295710556
proof_time_average: 77.00000000000001
******* propose ********
network fee: 0.3668379411737963
prover fee: 79005.04147855278
basefee_network: 8.542737132336227e-09
******* prove ********
proof_time: 200
prover reward: 161161.83912022863
basefee_proof: 0.0012409461612257608
proof_time_average: 83.15000000000002
******* propose ********
network fee: 0.3668379411737963
prover fee: 62047.30806128804
basefee_network: 8.542737132336227e-09
******* prove ********
proof_time: 50
prover reward: 32658.909283771933
basefee_proof: 0.0010862353227782547
proof_time_average: 81.49250000000002
******* propose ********
network fee: 0.3668379411737963
prover fee: 54311.76613891274
basefee_network: 8.542737132336227e-09
******* prove ********
proof_time: 25
prover reward: 16402.00134360839
basefee_proof: 0.0010693120755952057
proof_time_average: 78.66787500000001
```

Quickest way to trace back calculation in the solidity code is searching for 2 debug print keywords:
- `In proposeBlock()` (LibProposing.sol line 120)
-  `In verifyBlock()`  (LibVerifying.sol line 160)

Brechtpd:
The large difference between the prover fee paid by the proposer and the reward is because the base fee is calculated based on `proof_time_issued`. Because the initial value for that is set to `0`, the basefee will be wildly different from the expected value because the base fee needs some time to adjust to a correct value that will result in a reasonable fee. To get around that you can set `proof_time_issued` to an initial value that will result in a fee that makes sense from the start (I also updated my other python code):

```
basefee_proof = 10**18/L2_BLOCK_GAS_LIMIT
proof_time_issued = 
  (PROOF_TIME_TARGET * PROOF_ADJUSTMENT_QUOTIENT) * 
  math.log(basefee_proof * (PROOF_TIME_TARGET * PROOF_ADJUSTMENT_QUOTIENT))
```

This will make sure that the initial proving fee for a block of `L2_BLOCK_GAS_LIMIT` gas will be 1 token (10**18 decimals).

Do make sure that the part in the `ln()` is >= 1 because otherwise the initial value for `proof_time_issued` < 0 which doesn't make sense.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -107,6 +112,23 @@ library LibProposing {
...
+        // debug them. (Curly braces due to stack too deep error)
+        console2.log("----------------------------------------------------");
+        console2.log("------------------In proposeBlock()-----------------");
+        {
+            // Need to scale 'down', because baseFeeProof is scaled 'up'
+            // Todo: Could you please Brecht advise on the math ?
+            // So since in baseFee() function i needed to scale up, otherwise divison with zero
+            // somewhere we need to scale down..
+            uint256 feeNew = (input.gasLimit *
+                state.baseFeeProof); /*  / SCALING_FACTOR; (??) */
```
Brechtpd, line 127: Ideally it's not necessary to do any scaling here, but if the taiko token is used here and that one now only has 8 decimals it may be needed unfortunately.

packages/protocol/contracts/L1/libs/LibL1Tokenomics.sol
```diff
@@ -55,126 +62,99 @@ library LibL1Tokenomics {
...
-        if (timeAverage == 0) {
-            return (feeBase, 0);
-        }
-        unchecked {
-            uint p = feeConfig.dampingFactorBips; // [0-10000]
-            uint a = timeAverage;
-            uint t = (timeUsed * 1000).min(a * 2); // millisconds
+        newBaseFeeProof = uint64(
+            baseFee(cumulativeProofTime, proofTimeTarget, quotient) /
+                SCALING_FACTOR_1E18
```
Brechtpd, line 117: If `baseFee` is in 10*18 fixed point (which it currently isn't, but okay :), I think it would be nice to keep it in that notation. If we do that we should use the full `uin256`, but seems like a safe thing to do for now.

packages/protocol/contracts/L1/libs/LibL1Tokenomics.sol
```diff
@@ -55,126 +62,99 @@ library LibL1Tokenomics {
...
+        );
 
-            newFeeBase = (feeBase * (10000 + (t * p) / a - p)) / 10000;
+        reward =
+            ((newBaseFeeProof * usedGas) *
+                ((actProofTime * SCALING_FACTOR_1E8) / proofTimeTarget)) /
+            SCALING_FACTOR_1E8;
```
Brechtpd, line 123:
```suggestion
        reward = (state.baseFeeProof * usedGas * actProofTime) / (proofTimeTarget * SCALING_FACTOR_1E18);
```
- I think we can do this if the basefee is in 10**18 fixed point notation
- Using the previous basefee because I think it makes a little bit more sense now that I see it like this (newBaseFeeProof already shifted because of this proof, so the actual proof time would be taken into account twice for the reward).
- It's pretty inconvenient that the taiko token currently has a different decimals because the precision needs to be higher for it than for ETH, so not sure if my suggestions actually work because it's harder to think about it because a mix of formats.

Brechtpd (commented):
Looks good, nice work! I have a couple of possible small improvements. Also the diff is a bit confusing because it contains some of Daniel's old code so I may have missed some of your changes because of it.

It would be great if you could also add the time weighted reward thing in as an option, should be easy to do so!

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -87,17 +107,20 @@ library LibProposing {
...
-            if (state.balances[msg.sender] < burnAmount)
+            if (state.balances[msg.sender] < fee)
                 revert L1_INSUFFICIENT_TOKEN();
 
             unchecked {
-                state.balances[msg.sender] -= burnAmount;
+                state.balances[msg.sender] -= fee;
             }
 
             // Update feeBase and avgBlockTime
```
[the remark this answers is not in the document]
adaki2004: Ok, understood, thank you. Indeed was unsure if we ever want to use `avgBlockTime` later maybe, but then if so - i guess - we just update.

packages/protocol/contracts/L1/libs/LibL1Tokenomics.sol
adaki2004: Will look into this, tried quickly and did not work out of the blue, but will debug further.

adaki2004: Okay, i debugged it further the formula should be almost the same except `SCALING_FACTOR_1E18`. Since everything (beside TKO) is in 10 ** 18, but token has 8 decimals we need to use something like that: `SCALING_FROM_18_TO_TKO_DEC = 1e10;`

adaki2004:
@Brechtpd 
Actually I think there is something within the math somewhere.(Not related to any of the PR comments yet because i was about to apply your findings, after that tried to run tests but was failing.. And then i did a deeper investigation. But i reserve the right being noob.)

It seemed it worked on a smaller scale  , but i realized when running the following tests `test_reward_and_fee_if_proof_time_decreasing_then_stabilizes` - if you raise the loop counter for 'stable' proof submission:
```
console2.log("Stable");
for (uint256 blockId = 1; blockId < 60; blockId++) { 
```
The expected outcome is that it stabilizes long term and the ratio (balance of Bon and Alice) will getting closer to 1.

But Alice's balance's shrinking is slower than Bob's balance increasing. So the reward and fee is constant but they are not the same, reward is bigger.

adaki2004:
@Brechtpd @dantaik 
Feel free to review. There are some outstanding items (see PR desc) otherwise functional wise this is the one Brecht put together in a python script.

Some observations from today's testing. I might not like that much the 'targetTime' anymore - as much as i did - because it has the implications we need to set it very precisely.

2 scenarios.:
1. If proof time fluctuates then stabilizes AT proof time we get the equilibrium (long term : deposits = withdrawals)
See foundry test: `test_reward_and_fee_if_proof_time_increasing_then_stabilizes_at_the_proof_time_target`
kép

2. If proof time increases / decreases then stabilizes below AT proof time we will not get to  equilibrium (long term : deposits will be more, than withdrawals). The deposits and withdrawals will be constant, but different from each other obviously. (Withdrawal < Deposits)
Test case name: `test_reward_and_fee_if_proof_time_increasing_then_stabilizes_below_the_proof_time_target`
kép


The last commit (where we put amechanism, (or option)  not allow minting, just distributes what we have in treasury is the one i like the most, tho I need to write some test cases.

packages/protocol/contracts/L1/libs/LibL1Tokenomics.sol
```diff
@@ -53,128 +61,108 @@ library LibL1Tokenomics {
...
+            usedGas
+        );
+    }
+
+    /// @notice Update the baseFee for proofs
+    /// @param state - The actual state data
+    /// @param config - Config data
+    /// @param actProofTime - The actual proof time
+    /// @param usedGas - Gas in the block
+    function calculateBaseProof(
```
Brechtpd, line 92:
```suggestion
    function calculateBaseFeeProof(
```

packages/protocol/contracts/L1/libs/LibL1Tokenomics.sol
```diff
@@ -53,128 +61,108 @@ library LibL1Tokenomics {
...
+        TaikoData.State storage state,
+        TaikoData.Config memory config,
+        uint64 actProofTime,
```
Brechtpd, line 95:
```suggestion
        uint64 proofTime,
```

packages/protocol/contracts/L1/libs/LibL1Tokenomics.sol
```diff
@@ -53,128 +61,108 @@ library LibL1Tokenomics {
...
         view
-        returns (uint256 newFeeBase, uint256 reward, uint256 premiumRate)
+        returns (
+            uint256 reward,
+            uint256 newProofTimeIssued,
+            uint256 newBaseFeeProof
+        )
     {
-        if (proposedAt > provenAt) revert L1_INVALID_PARAM();
+        uint256 proofTime = state.proofTimeIssued;
```
Brechtpd, line 106:
```suggestion
        uint256 proofTimeIssued = state.proofTimeIssued;
```
Otherwise pretty confusing.

Brechtpd (commented):
LGTM! Just some very minor suggestions that don't impact anything.

packages/protocol/contracts/L1/libs/LibL1Tokenomics.sol
```diff
@@ -53,128 +61,108 @@ library LibL1Tokenomics {
...
+    /// @param value - Result of cumulativeProofTime
+    /// @param target - Target proof time
+    /// @param quotient - Quotient
+    function baseFee(
+        uint256 value,
+        uint256 target,
+        uint256 quotient
+    ) internal pure returns (uint256) {
+        uint256 newValue = (expCalculation(value, target, quotient));
+        return ((newValue / (target * quotient)));
```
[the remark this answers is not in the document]
adaki2004: Ah yes! Separated for debug print purposes and forget to rearrange ! Thanks.

dantaik (approved):

packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -52,108 +59,116 @@ library LibTokenomics {
...
 
-        unchecked {
-            reward = uint64((reward * (10000 - config.rewardBurnBips)) / 10000);
+        if (config.allowMinting) {
+            reward = ((state.baseFeeProof * usedGas * proofTime) /
+                (config.proofTimeTarget * SCALING_FROM_18_TO_TKO_DEC));
+        } else {
+            /// TODO: Verify with functional tests
+            uint256 numBlocksBeingProven = state.numBlocks -
+                state.lastVerifiedBlockId;
```
Brechtpd, line 124:
```suggestion
            uint256 numBlocksBeingProven = state.numBlocks - state.lastVerifiedBlockId - 1;
```

I think we need to subtract `1` here.

Brechtpd (approved):

dantaik (changes requested):

packages/protocol/contracts/L1/TaikoConfig.sol
```diff
@@ -34,20 +36,17 @@ library TaikoConfig {
                 slotSmoothingFactor: 946649,
                 // 100 basis points or 1%
                 rewardBurnBips: 100,
-                proposerDepositPctg: 25, // - 25%
                 // Moving average factors
                 feeBaseMAF: 1024,
-                bootstrapDiscountHalvingPeriod: 1 seconds, // owner:daniel
-                constantFeeRewardBlocks: 1024,
                 txListCacheExpiry: 0,
+                proofTimeTarget: 85, // 85s based on A2 testnet status
```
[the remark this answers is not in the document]
adaki2004: Ok, wee need to keep that in mind, but I would leave it as is for the next couple iterations (testnets), because if:
 `proofTime` highly differs from `proofTimeTarget` AND `allowMinting` is `true` we will not reach balance and deposits will be much bigger than withdrawals - which is net good, but definitely an un-balanced state.

dantaik:
@Brechtpd @adaki2004  so if `allowMinting` is true, is it possible that the supply will go down? When can that happen?

I'm not sure if we should support this config, maybe we should simply allow dynamic minting and burning of tokens.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -125,25 +132,31 @@ library TaikoData {
...
         mapping(address account => uint256 balance) balances;
         mapping(bytes32 txListHash => TxListInfo) txListInfo;
+        // Cummulated proofTime for reward calculation - changed in verifyBlock()
+        uint256 proofTimeIssued;
+        // Changing baseFee for proving - changed in verifyBlock()
+        uint256 baseFeeProof;
+        // Changing accumulated time for proposing - changed in proposeBlock() and in verifyBlock()
+        uint256 accProposalTime;
+        // Treasury amount - changed in proposeBlock() and in verifyBlock()
+        uint256 proofFeeTreasury;
```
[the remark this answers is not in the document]
Brechtpd: Should we optimize these things already? It's to test things out with different configurations so some of the will still be removed for sure.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -25,17 +25,24 @@ contract TaikoL1 is EssentialContract, IXchainSync, TaikoEvents, TaikoErrors {
...
+     *
+     * @param _addressManager The AddressManager address.
+     * @param _feeBase The initial value of the proposer-fee/prover-reward feeBase.
+     * @param _genesisBlockHash The block hash of the genesis block.
+     */
     function init(
         address _addressManager,
-        bytes32 _genesisBlockHash,
-        uint64 _feeBase
+        uint64 _feeBase,
```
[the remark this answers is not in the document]
Brechtpd: Still a bit confusing to me with L1 basefee vs Ehereum base fee. I think prover basefee is the most clear one.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -87,45 +93,30 @@ library LibProposing {
         blk.verifiedForkChoiceId = 0;
         blk.metaHash = LibUtils.hashMetadata(meta);
         blk.proposer = msg.sender;
+        // Later on we might need to have actual gas consumed in that L2 block
+        blk.gasConsumed = input.gasLimit;
 
         if (config.enableTokenomics) {
-            (uint256 newFeeBase, uint256 fee, uint256 deposit) = LibL1Tokenomics
-                .getBlockFee(state, config);
+            (, uint256 fee) = LibTokenomics.getProverFee(state, input.gasLimit);
```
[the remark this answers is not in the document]
Brechtpd: I kind of hope that we will never have to store any amounts anywhere, so we can return the taiko token back to 18 decimals so it's the same as ETH and only 1 format to think about.

adaki2004:
> > @Brechtpd @adaki2004 so if `allowMinting` is true, is it possible that the supply will go down? When can that happen?
> > I'm not sure if we should support this config, maybe we should simply allow dynamic minting and burning of tokens.
> 
> `allowMinting == true` is basically the same behavior as we had previously so, burning tokens at the proposing side, minting tokens at the reward side. So pretty sure it should be roughly no change in supply.
> 
> > For useTimeWeightedReward, I'd also prefer that we remove this parameter and use useTimeWeightedReward=true in our design.
> 
> I do also think the non time weighted one is very unlikely to work for us, so makes sense to me to remove it already.

@Brechtpd @dantaik 
Daniel and I had a meeting this morning - and discussed some things, but i'd like to double check with both of you my curent blockers (point 1).

1. Daniel created with proposed changes this branch + PR:
https://github.com/taikoxyz/taiko-mono/pull/13540
Problem with this solution is that in verifyBlock() he uses `fc.gasUsed`, while in proposeBlock() the `input.gasLimit` (gasUsed not available at prroposal time).
I dont know how to resolve this ATM, 
- If we want to adjust basefee with the gas used (tho it depends highly on the proof time, not the demand for block space (?)) we need to have this information at `proposeBlock()` ?
- If we don't want to adjust it according to the demand for blockspace we can use `input.gasLimit` (need to be re-added to `struct Block{}`) or we use a constant ?

2. We agreed to pick 1 algorithm - since we anyhow try one / testnet and i'd go for the time weighted treasury pool one and remove the rest. 
Reason i'd choose that because the `allowMinting` version requires us to be punctual with the proofTimeTarget - otherwise we end up in an unbalanced state and maybe kind of frequent updates.
Since the test results showed: 
- stable above `proofTimeTarget`: ever-increasing baseFee (hance fee and reward)
- stable but below `proofTimeTarget`: there is a surplus long term because rewards are smaller than deposits. (So we would need to do something with that surplus. Burn or lower the `prooTimeTarget`, etc.)
- Only ideal if `proofTimeTarget` is punctual.
Is it fine with you ?

Brechtpd:
>  Daniel created with proposed changes this branch + PR:
>  [refactor(protocol): recommend changes to Dani's PR (part1) #13540](https://github.com/taikoxyz/taiko-mono/pull/13540)
>  Problem with this solution is that in verifyBlock() he uses `fc.gasUsed`, while in proposeBlock() the `input.gasLimit` (gasUsed not available at prroposal time).
>  I dont know how to resolve this ATM,
> 
> 
> If we want to adjust basefee with the gas used (tho it depends highly on the proof time, not the demand for block space (?)) we need to have this information at `proposeBlock()` ?
> 
> If we don't want to adjust it according to the demand for blockspace we can use `input.gasLimit` (need to be re-added to `struct Block{}`) or we use a constant ?
> 

Using `gasUsed` for the reward calculation is a bit problematic I think because it could be 0 (empty block), while the prover obviously still has significant costs (some proof generation, but also L1 gas costs that are independent of how much gas is used in a block). For the same reason the proposer set gas limit also isn't really usable. Best we can do is probably to just use the fixed L2 block gas limit on the proposer side and the reward side.

> 2. We agreed to pick 1 algorithm - since we anyhow try one / testnet and i'd go for the time weighted treasury pool one and remove the rest.

That's also the one I like best, if it works of course. :)

> Reason i'd choose that because the `allowMinting` version requires us to be punctual with the proofTimeTarget - otherwise we end up in an unbalanced state and maybe kind of frequent updates.
> Since the test results showed:
> 
>  stable above `proofTimeTarget`: ever-increasing baseFee (hance fee and reward)

Yeah we'd definitely have to make sure that the target is more than enough for a proof to be generated, otherwise it cannot work.

>  stable but below `proofTimeTarget`: there is a surplus long term because rewards are smaller than deposits. (So we would need to do something with that surplus. Burn or lower the `prooTimeTarget`, etc.)

In this case the basefee would going down so people would stop submitting proofs before the target no? Which would then in turn increase the basefee again because proofs are submitted too late.

> Only ideal if `proofTimeTarget` is punctual.

I agree that finding a good target would be a bit tricky, though I think with limited impact even if the value isn't ideal (as long as it's not too short).

dantaik:
> > > @Brechtpd @adaki2004 so if `allowMinting` is true, is it possible that the supply will go down? When can that happen?
> > > I'm not sure if we should support this config, maybe we should simply allow dynamic minting and burning of tokens.
> > 
> > 
> > `allowMinting == true` is basically the same behavior as we had previously so, burning tokens at the proposing side, minting tokens at the reward side. So pretty sure it should be roughly no change in supply.
> > > For useTimeWeightedReward, I'd also prefer that we remove this parameter and use useTimeWeightedReward=true in our design.
> > 
> > 
> > I do also think the non time weighted one is very unlikely to work for us, so makes sense to me to remove it already.
> 
> @Brechtpd @dantaik Daniel and I had a meeting this morning - and discussed some things, but i'd like to double check with both of you my curent blockers (point 1).
> 
> 1. Daniel created with proposed changes this branch + PR:
>    [refactor(protocol): recommend changes to Dani's PR (part1) #13540](https://github.com/taikoxyz/taiko-mono/pull/13540)
>    Problem with this solution is that in verifyBlock() he uses `fc.gasUsed`, while in proposeBlock() the `input.gasLimit` (gasUsed not available at prroposal time).
>    I dont know how to resolve this ATM,
> 
> * If we want to adjust basefee with the gas used (tho it depends highly on the proof time, not the demand for block space (?)) we need to have this information at `proposeBlock()` ?
> * If we don't want to adjust it according to the demand for blockspace we can use `input.gasLimit` (need to be re-added to `struct Block{}`) or we use a constant ?
> 
> 2. We agreed to pick 1 algorithm - since we anyhow try one / testnet and i'd go for the time weighted treasury pool one and remove the rest.
>    Reason i'd choose that because the `allowMinting` version requires us to be punctual with the proofTimeTarget - otherwise we end up in an unbalanced state and maybe kind of frequent updates.
>    Since the test results showed:
> 
> * stable above `proofTimeTarget`: ever-increasing baseFee (hance fee and reward)
> * stable but below `proofTimeTarget`: there is a surplus long term because rewards are smaller than deposits. (So we would need to do something with that surplus. Burn or lower the `prooTimeTarget`, etc.)
> * Only ideal if `proofTimeTarget` is punctual.
>   Is it fine with you ?

Lets talk about it tomorrow.

packages/protocol/utils/generate_config/main.py
```diff
@@ -39,4 +39,3 @@
 
     print("---------")
     print("maxNumProposedBlocks:", max_num_slots)
-    print("slotSmoothingFactor:", fee_smoothing_factor)
```
dantaik, line 42: How about we delete this file.

packages/tokenomics/main.py
```diff
@@ -96,7 +96,7 @@ def setup(self, config):
 
     def get_time_adjusted_fee(self, is_proposal, t_now, t_last, t_avg, t_cap):
         #  if (tAvg == 0) {
-        #      return s.feeBase;
```
dantaik, line 99: Lets also delete this file.

packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -53,107 +56,118 @@ library LibTokenomics {
...
+        if (result > type(uint64).max) return type(uint64).max;
+
+        return uint64(result);
+    }
+
+    /// @notice Calculating the exponential via LibFixedPointMath.sol
+    /// @param value - Result of cumulativeProofTime
+    /// @param target - Target proof time
+    /// @param quotient - Quotient
+    function _expCalculation(
```
dantaik, line 164: maybe we cal also inline this one?

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -204,20 +192,8 @@ library LibVerifying {
             // EIP-4844 blob size up to 128K
             config.maxBytesPerTxList > 128 * 1024 ||
             config.minTxGasLimit == 0 ||
-            config.slotSmoothingFactor == 0 ||
             // EIP-4844 blob deleted after 30 days
-            config.txListCacheExpiry > 30 * 24 hours ||
-            config.rewardBurnBips >= 10000
+            config.txListCacheExpiry > 30 * 24 hours
```
dantaik, line 196: Shall we check non-zero for `proofTimeTarget` and `adjustmentQuotient`?

packages/protocol/test2/TaikoL1TestBase.t.sol
```diff
@@ -19,6 +27,8 @@ abstract contract TaikoL1TestBase is Test {
     TaikoData.Config conf;
     uint256 internal logCount;
 
+    uint256 private constant SCALING_E18 = 1e18;
```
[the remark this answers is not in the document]
adaki2004: We can put for sure, eauswe it is needed for 'off-chain' calculation, to determine the initProofTimeTarget, tho changing to 1e17 is not recommended, bc. the ln calcl function expects 1e18 fixed notation.
Will put tho this into a common place.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -142,24 +132,26 @@ library TaikoData {
...
+        uint64 lastProposedAt;
         uint64 numBlocks;
-        uint64 lastProposedAt; // Timestamp when the last block is proposed.
-        uint64 avgBlockTime; // miliseconds
-        uint64 __reserved3;
-        // Changed when a block is proven/verified
+        uint64 accProposedAt; // also by verifyBlocks
+        uint64 rewardPool; // also by verifyBlocks
+        // Slot 7: changed by proveBlock
+        // uint64 __reserved71;
```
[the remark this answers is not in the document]
adaki2004: I'm not sure i understand this. Uncomment these 4 used slots ? (accProposedAt, and rewardPool ?)

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -29,20 +29,23 @@ contract TaikoL1 is EssentialContract, IXchainSync, TaikoEvents, TaikoErrors {
...
+     * @param _initBasefee Initial (reasonable) basefee value.
+     * @param _initProofTimeIssued Initial proof time which keeps the inflow/outflow in balance
      */
     function init(
         address _addressManager,
-        uint64 _feeBase,
-        bytes32 _genesisBlockHash
+        bytes32 _genesisBlockHash,
+        uint64 _initBasefee,
+        uint64 _initProofTimeIssued
```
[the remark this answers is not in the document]
adaki2004: Yes, this is what we do !
So basically:
- Determine an ideal proofTimeTarget
- Determine "off-chain" proofTimeIssued variable in order to maintain the same fee - if everything is ideal.

To determine the calculation formula we apply this calculation:
https://github.com/taikoxyz/taiko-mono/pull/13564#discussion_r1160780472
In:
TaikoL1LibTokenomicsMainnet.t
TaikoL1LibTokenomicsTestnet.t
DeployOnL1.s.sol

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -24,29 +19,24 @@ library TaikoData {
...
         bool enableOracleProver;
         bool enableTokenomics;
         bool skipZKPVerification;
-        FeeConfig proposingConfig;
-        FeeConfig provingConfig;
     }
 
     struct StateVariables {
-        uint64 feeBase;
+        uint64 basefee;
```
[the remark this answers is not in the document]
adaki2004: Hmm, represent more what we mean ATM i agree - however if we do not stick to this tokenomics later on (which is definitely probable) it might be just a basefee + some staked/staking fee, or whatever. Currently indeed a blockfee, tho.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -60,7 +63,7 @@ contract TaikoL1 is EssentialContract, IXchainSync, TaikoEvents, TaikoErrors {
     function proposeBlock(
         bytes calldata input,
         bytes calldata txList
-    ) external nonReentrant {
+    ) external payable nonReentrant {
```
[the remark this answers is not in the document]
adaki2004: Most probably it is a leftover from the commit where we wanted to calculate 1559 for L2 on L1. Will remove.
https://github.com/taikoxyz/taiko-mono/commit/0725c7a0fe3c36715f8c9198c89d9c2c4daa010b#diff-8493268db26a2d0769831c83971c1ead8e2c25cf50e582e02749ea8ae157c361

packages/protocol/script/DeployOnL1.s.sol
```diff
@@ -48,6 +50,11 @@ contract DeployOnL1 is Script, AddressResolver {
 
     address public addressManagerProxy;
 
+    // New fee/reward related variables
+    uint256 public constant SCALING_E18 = 1e18;
```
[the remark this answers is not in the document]
adaki2004: Since we need the ln calc, i put it there.

packages/protocol/test2/Ln.sol
```diff
@@ -0,0 +1,207 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.18;
```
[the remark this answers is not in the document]
adaki2004: Same as LibFixedPointMath - but i think you removed the LN functions. Which is totally fine, we dont need them 'on-chain'. (Will comment into the file)

packages/protocol/test2/TaikoL1.t.sol
```diff
@@ -57,14 +39,10 @@ contract TaikoL1Test is TaikoL1TestBase {
...
         TaikoL1TestBase.setUp();
-        _registerAddress(
-            string(abi.encodePacked("verifier_", uint16(100))),
-            address(new Verifier())
-        );
     }
 
     /// @dev Test we can propose, prove, then verify more blocks than 'maxNumProposedBlocks'
-    function test_more_blocks_than_ring_buffer_size() external {
+    function xtest_more_blocks_than_ring_buffer_size() external {
```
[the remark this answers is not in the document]
adaki2004: These were temporary disabled only to not spam debugging.

packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -53,107 +56,118 @@ library LibTokenomics {
...
         }
+
+        newProofTimeIssued = proofTimeIssued;
+    }
+
+    /// @notice Calculating the exponential smoothened with (target/quotient)
+    /// @param value - Result of cumulativeProofTime
+    /// @param target - Target proof time
+    /// @param quotient - Quotient
+    function _calcBasefee(
```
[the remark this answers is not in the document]
adaki2004: Would rather leave is as is and incorporate the _expCalculation into here.

adaki2004:
@dantaik @Brechtpd 

 We definitely need to be reasonable with the proofTimeTarget.
- Better to overestimate a little bit than under. Overestimating might cause fee decreasing, which net good, if not profitable, provers simply refuse to provide proofs within that time (Question if this concern is applicable on testnet..?)  If we underestimate and provers are technically not able to be within the window the fee will be ever-increasing, so that is the worst case.

 **How to test possible main net scenario ?** 
 Remove the ‘x’ prefix from the function names of the TaikoL1LibTokenomicsMainnet.t.sol and use gas setting: 3000M in foundry.toml -> 
    - Currently these tests shall be run locally, but simulates a kind of mainnet scenario where the first 30 minutes there will be some blocks being proposed (not proved) then proving starts. 
    - If this gas setting is enabled to run simulation, the anchor tests will fails, because of baseFee mismatch (!!)

packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -53,107 +56,100 @@ library LibTokenomics {
...
-            // k is `m − n + 1` or `m − n - 1`in the whitepaper
-            uint256 k = isProposal ? m - n - 1000 : m - n + 1000;
-            return uint64((feeBase * (m - 1000) * m) / (m - n) / k);
+        newBasefee = _calcBasefee(
+            newProofTimeIssued,
+            config.proofTimeTarget,
+            config.adjustmentQuotient
+        );
+
+        uint64 numBlocksBeingProven = state.numBlocks -
```
davidtaikocha, line 105: actually I dont think this value (`state.numBlocks - state.lastVerifiedBlockId - 1`) can represent `numBlocksBeingProven`, its just `numBlocksUnverified`, lets say there are 100 unproven blocks (height 101 to height 200), then we submit the proofs for blocks from height 102 to 200, and then this value will be still 100 (since block 101 has not been proven yet).

packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -53,107 +56,100 @@ library LibTokenomics {
...
+            );
+            ///@dev If block timestamp is equal to state.accProposedAt (not really, but theoretically possible)
+            ///@dev there will be division by 0 error
+            if (totalNumProvingSeconds == 0) {
+                totalNumProvingSeconds = 1;
+            }
+
+            reward = uint64(
+                (uint256(state.accBlockFees) * proofTime) /
+                    totalNumProvingSeconds
```
davidtaikocha, line 124: seems that there is a chance that when there are more and more unproven blocks on chain, reward will be less and less? or i missed something?

adaki2004: Theoretically possible, in case proposers stop proposing.
The fee connected with the proof time issued seconds, so the more 'unproven' -> the more the fee -> the bigger the reward pool.

packages/protocol/contracts/L1/libs/LibTokenomics.sol
adaki2004: Fair point !

adaki2004: Hmm, the more I think about it the more I tend to think does not change anything.

It is indeed not the `numBlocksBeingProven`, but we could call it like `numBlocksUnverified` or even `numBlocksUnpaidRewards` which is a valid representation what we would like to achieve. Every update in basefee and reward is done during `verifyBlock`, as well as the payouts.
@Brechtpd  ?

dantaik:
@adaki2004 please also fix all the warnings except those related to `block.prevrandao`.

packages/protocol/test2/TaikoL1.sim.sol
```diff
@@ -44,13 +43,18 @@ contract TaikoL1Simulation is TaikoL1TestBase, FoundryRandom {
...
         TaikoL1TestBase.setUp();
+        // TODO(daniel): update string key generation using bytes.concat
         _registerAddress(
             string(abi.encodePacked("verifier_", uint16(100))),
             address(new Verifier())
         );
     }
 
-    function testGeneratingManyRandomBlocks() external {
+    // Disabling this test for now. (We have someshting similar
```
dantaik, line 53: This test is where we want to simulate the real world scenarios by randomizing as many parameters as possible without many detailed assertions. Then later we can export some stats and analysis the data by visualizing them, that's why CSV data is exported, one line per block.

packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -53,107 +56,99 @@ library LibTokenomics {
     }
 
     function getBlockFee(
```
dantaik, line 58: Lets delete this method and use `state.basefee` instead

dantaik (approved):

PR #13490, opened 2023-03-28
feat(website): add index sections to reference and resources

developerkunal:
This commit adds new sections to the File Description Page for Taiko Protocol resources, including guidelines for contributing to Taiko Protocol on GitHub, official links, presentations, a Taiko glossary, and the official whitepaper. These additions should make it easier for developers and users to navigate and access relevant information about Taiko Protocol.

@jscriptcoder

PR #13502, opened 2023-03-30
docs(website): improve UX on Contract Documentation references

fzn0x:
Not really a big PR, just a little improvements on the UX for contract documentation references.

Now everytime users go to https://taiko.xyz/docs/reference/contract-documentation they know where to go next and avoid double click on the sidebar.

Preview: https://website-nbrvfw3y1-taikoxyz.vercel.app/docs/reference/contract-documentation

### DIFF
--- a/packages/website/pages/docs/reference/contract-documentation.md
+++ b/packages/website/pages/docs/reference/contract-documentation.md
@@ -1,1 +1,24 @@
-Here you will find the documentation for the Taiko smart contracts.
+Here you will find the documentation for the Taiko smart contracts:
+
+- L1
+  - [IProofVerifier](/docs/reference/contract-documentation/L1/TaikoL1)
+  - [TaikoCustomErrors](/docs/reference/contract-documentation/L1/TaikoCustomErrors)
+  - [TaikoData](/docs/reference/contract-documentation/L1/TaikoData)
+  - [TaikoEvents](/docs/reference/contract-documentation/L1/TaikoEvents)
+  - [TaikoL1](/docs/reference/contract-documentation/L1/TaikoL1)
+  - [TaikoToken](/docs/reference/contract-documentation/L1/TaikoToken)
+- L2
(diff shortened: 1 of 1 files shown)


dantaik (approved):

d1onys1us:
this has the same maintainability issue as here: https://github.com/taikoxyz/taiko-mono/pull/13490#issuecomment-1490621145

it will become stale over time and this one will be particularly difficult to maintain. but thank you for the PR, the idea is definitely worth it, i'm just concerned about the maintainability.. do you agree?

fzn0x:
> this has the same maintainability issue as here: [#13490 (comment)](https://github.com/taikoxyz/taiko-mono/pull/13490#issuecomment-1490621145)
> 
> it will become stale over time and this one will be particularly difficult to maintain. but thank you for the PR, the idea is definitely worth it, i'm just concerned about the maintainability.. do you agree?

I definitely agree, so we are preferring another approach which automically indexes through all the files and the output is indexing the pages to become references in the main page automically?

d1onys1us:
> > this has the same maintainability issue as here: [#13490 (comment)](https://github.com/taikoxyz/taiko-mono/pull/13490#issuecomment-1490621145)
> > it will become stale over time and this one will be particularly difficult to maintain. but thank you for the PR, the idea is definitely worth it, i'm just concerned about the maintainability.. do you agree?
> 
> I definitely agree, so we are preferring another approach which automically indexes through all the files and the output is indexing the pages to become references in the main page automically?

in the long run, yes, but we have not created that solution yet. basically 1) we would need to create a PR towards nextra to improve or write our own script in order to achieve this and dynamically create these links or 2) move to another framework which supports it natively. i lean slightly towards 1), because i don't think it would be *too difficult* to do this, but i could be wrong.

in any case i think that this particular one is a little bit too high risk of getting out of date and being too much maintenance overhead because of the constant changing nature of the smart contract documentation.

PR #13510, opened 2023-03-31
refactor(protocol): remove prefix in hashMessage()

dantaik:


packages/protocol/contracts/bridge/libs/LibBridgeData.sol
```diff
@@ -44,6 +44,6 @@ library LibBridgeData {
     function hashMessage(
```
davidtaikocha, line 44: Shall we also update the comment of this function?

dantaik: fixed

davidtaikocha (approved):

PR #13516, opened 2023-03-31
docs(starter-dapp): update README.md with correct env var values

LouisWT:
Currently README.md's env config can't be used. Because https://l1rpc.internal.taiko.xyz is fake url, and metamask can't add this as a new rpc, so it will fail when request connecting.
A out of box env config will be more friendly for developer.

### DIFF
--- a/packages/starter-dapp/README.md
+++ b/packages/starter-dapp/README.md
@@ -13,15 +13,14 @@
 You can use the following values in the `.env` file to spin up the Starter dapp locally.
 
 ```
-VITE_NODE_ENV=dev
-VITE_L1_RPC_URL="https://l1rpc.internal.taiko.xyz/"
-VITE_L2_RPC_URL="https://l2rpc.internal.taiko.xyz/"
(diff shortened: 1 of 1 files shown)


packages/starter-dapp/README.md
```diff
@@ -13,15 +13,14 @@
 You can use the following values in the `.env` file to spin up the Starter dapp locally.
 
 ```
-VITE_NODE_ENV=dev
```
d1onys1us, line 16: why are we removing this VITE_NODE_ENV=dev var? might be useful to keep.

d1onys1us:
@web3Liuwentong thanks for your contribution. i would recommend just replacing the "internal" keyword with "a2" to make sure this doesn't break anything

d1onys1us:
just waiting on this to merge and i will fix the build: https://github.com/taikoxyz/taiko-mono/pull/13551

dionysuzx (approved):

dantaik (approved):

PR #13519, opened 2023-03-31
fix(bridge-ui): only request connect if require

LouisWT:
Without this check, wagmi will throw ConnectorAlreadyConnectedError when Metamask is already available .


web3Liuwentong:
fix #13520 by this PR.

cyberhorsey (approved):

dionysuzx (approved):

PR #13521, opened 2023-03-31
update docker-compose commands

Lesss1:


davidtaikocha:
`docker compose`(a newer golang implementation) is preferred than `docker-compose`(an older python implementation, deprecated now): https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose

PR #13525, opened 2023-04-03
feat(protocol): L2 1559 use gasUsed to adjust basefee

dantaik:
@Brechtpd  this PR enables L2 1559 gas fee to be calculated and verified when a block is to be verified. When a block is proven, `gasfee` and `gasUsed` must be provided by the prover, `gasUsed` will be part of the fork choice, and `gasfee` is used as a link to this fork choice.

Something to notice: now for a block, its basefee does not depends on its own actual gas used, instead, the basefee is the one updated after its parent block. In other words, the basefee is the spot price of the curve, not the average price of a gas sale. The major reason for such a change is to remove basefee calculate's self-dependency. **If we keep using the average price of a gas sale, then we have to iterate through all fork choices to find the correct one**, as basefee depends on gasUsed, while gasUsed maybe affected by the value of basefee. But I'm ok to implementation suggestions so we can still use the average price, not the spot price.

It seems Ethereum also uses the basefee calculated on the previous blocks, not the current block's gas used.

TODOs:
- [ ] add more testing
- [ ] Talk to client team so they know how to implement this and how to tell users the basefee value at any given moment.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -48,6 +48,8 @@ library TaikoData {
         uint64 avgBlockTime;
         uint64 avgProofTime;
         uint64 lastProposedAt;
+        uint64 l2Basefee; // L2 1559
+        uint64 l2GasExcess; // L2 1559
     }
 
     // 3 slots
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -141,13 +141,13 @@ library TaikoData {
         uint64 numBlocks;
         uint64 lastProposedAt; // Timestamp when the last block is proposed.
         uint64 avgBlockTime; // miliseconds
-        uint64 l2GasExcess;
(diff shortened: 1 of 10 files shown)


Brechtpd (commented):
Looks pretty nice! The only downside of this approach is a bit messier fork choices and still having to store some state on L1 with some extra cross-chain data. Doing things in the anchor function keeps the fork choice logic simpler and moves everything to L2 (nothing visible on L1, no changes required to L1 contracts) so potentially a bit cheaper, but it would be doing the exact same thing using practically the same code, just in a different place, so both fine with me.

> But I'm ok to implementation suggestions so we can still use the average price, not the spot price.

I don't think we can do it without having to implement refunds which seems a lot of work for basically no real benefit, so totally agree we should not do it.

dantaik:
> Doing things in the anchor function keeps the fork choice logic simpler and moves everything to L2

I'll think about it and if it is possible, we should definitely move it inside L2 anchor.

dantaik:
Closed with a new PR: https://github.com/taikoxyz/taiko-mono/pull/13533

PR #13527, opened 2023-04-03
docs(website): improve taiko nodes documentation

archie0x:
Hey @d1onys1us, here's a revamped version of the Taiko nodes docs.

A few comments:

1. I replaced the Engine API link with another one because the old one was broken. Need to check if it's the correct one.
2. I added a few links to terms from the rollup glossary. I'm not sure if it's the right move but I thought since we have some definitions there that we don't have on the website, why not use it?
3. I used the callout components as you said but at least in the preview, the code doesn't seem to work for some reason. Would appreciate it if you could look into it.
4. Do we have anything unique to say in the `Engine API` section? We kind of repeat ourselves there, I wonder if we even need it?

Let me know what you think!

### DIFF
--- a/packages/website/pages/docs/concepts/taiko-nodes.mdx
+++ b/packages/website/pages/docs/concepts/taiko-nodes.mdx
@@ -1,98 +1,119 @@
-There are two parts to a Taiko node, which are connected over the engine API:
-- `taiko-geth`
-- `taiko-client`
+## Overview
+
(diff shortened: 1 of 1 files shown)


umededoteth:
Also, @d1onys1us, I can't seem to link this PR to the issue you created.

davidtaikocha (dismissed):

d1onys1us:
hey umede, thanks for the PR! looks great to me, nice work 💯 

i made a small fix, if you're curious: 453966e

dionysuzx (approved):

davidtaikocha (approved):

PR #13533, opened 2023-04-04
feat(protocol): calculate and verify 1559 basefee in anchor

dantaik:
- [x] add more uint tests (in another PR)
- [x] add an getBasefee() function in TaikoL2.sol

| test2/TaikoL1.t.sol:TaikoL1_a contract |                 |        |        |        |         |
| Deployment Cost                        | Deployment Size |        |        |        |         |
| 3202739                                | 16028           |        |        |        |         |
| proposeBlock                           | 22760           | 32881  | 22780  | 92779  | 309     |
| proveBlock                             | 55431           | 65485  | 55431  | 131831 | 309     |
| verifyBlocks                           | 13575           | 14588  | 14293  | 80372  | 301     |
|   |  |  |  |  |  | 
| Deployment Cost                           | Deployment Size |        |        |        |         |
| 1535403                                   | 7701            |        |        |        |         |
| anchor                                    | 539             | 44802  | 42437  | 89258  | 165     |


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -191,4 +273,28 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, IXchainSync {
...
+            xPurchase: gasLimit
+        });
+
+        // Very important to cap basefee uint64
+        _basefee = uint64(__basefee.min(type(uint64).max));
+
+        // Very important to cap _gasExcess uint64
+        _gasExcess = uint64(
+            (uint256(_gasExcess) + gasLimit).min(type(uint64).max)
+        );
```
Brechtpd, line 298: Using gas used here I think would be a nice improvement so the basefee is less gameable. But that does make things a little bit more complex because gas used would need to be an input of the anchor function that is verified by the circuits (the same is true about the other inputs so that shouldn't really be any extra work), the gas used value will have to be the total gas used by the block excluding the anchor tx itself to not create a circular dependency. I think both total gas used in the block and gas used for the anchor tx will already be known by the circuit so should also be an easy enough value to use.

Brechtpd (approved):
LGTM! I guess the gas used suggestion we can always do later if we want to do so if we want to avoid doing it for now.

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: Using gasUsed of the current block will create a  circular dependency for client: the client has to run transactions in a block with some initial basefee, then get the total gasUsed, then verify the basefee is good in the anchor transaction, if the basefee value is not good, the client must adjust it, rerun all txs to get a new gasUsed, then check again the gasUsed value and the basefee can pass the anchor's check.

Brechtpd: I'm only proposing to use gas_used to update the gasExcess value here because that value will be used for future base fee calculations, not the base fee for the current block. So the current block base fee is still an approximation using the gas limit (deterministic, unrelated to the current block gas_used, only impacts the base fee for the current block so doesn't really matter). The most important value is `gasExcess` because that's the one that is brought over for all future base fee calculations, so important that this value cannot be manipulated somehow.

dantaik: Updated, please take a look.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -8,13 +8,31 @@ pragma solidity ^0.8.18;
...
 import {TaikoL2Signer} from "./TaikoL2Signer.sol";
+import {
+    SafeCastUpgradeable
+} from "@openzeppelin/contracts-upgradeable/utils/math/SafeCastUpgradeable.sol";
 
 contract TaikoL2 is EssentialContract, TaikoL2Signer, IXchainSync {
+    using SafeCastUpgradeable for uint256;
+    using LibMath for uint256;
+
+    uint64 public ANCHOR_GAS_COST = 47000;
```
dantaik, line 22: @davidtaikocha please confirm this gas cost value.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -191,4 +281,31 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, IXchainSync {
             currPIH := keccak256(inputs, mul(256, 32))
         }
     }
+
+    function _calcBasefee(
+        uint256 timeSinceParent,
+        uint64 gasLimit,
+        uint64 parentGasUsed
+    ) private view returns (uint64 _basefee, uint64 _gasExcess) {
+        uint256 gasIssued = gasIssuedPerSecond * timeSinceParent;
```
davidtaikocha, line 290:
```suggestion
```

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -89,15 +150,17 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, IXchainSync {
...
+     * @param parentGasUsed the gas used in the parent block.
      */
 
     function anchor(
-        uint256 l1Height,
         bytes32 l1Hash,
-        bytes32 l1SignalRoot
+        bytes32 l1SignalRoot,
+        uint64 l1Height,
+        uint64 parentGasUsed
```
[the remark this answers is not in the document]
dantaik: Unlike `block.gaslimit`, "current block gas used" is not available.

PR #13537, opened 2023-04-04
feat(protocol): auction implementation

cyberhorsey:
VERY WIP!

- you bid for blocks in terms of batches that have a batchSize, ie, 100
- you can bid for blocks up to auctionBlockGap away from lastVerifiedID, ie: 2000
- this means 20 batches can be bid on it ahead of lastVerifiedId

first bid opens an auction for `auctionLength` length after first bid.

Bidding works by offering a minFeePerGas, which is the lowest amount of tko token you would be paid per gasUsed, and now we submit gasUsed with the proof as well.

This means: the proof reward is simply minFeePerGas * gasUsed.

we also have a deposit which is, your confidence in winnign the auction.
We create a weight of the minFeePerGas (more heavily weighted) + the gasUsed (lower weighted). to determine if your bid outbids the other.

Ideally, we would not keep track of the `proofReward` with math at all, and just let demand for block space and proving rights determine the payout.

Many TODOs:
- add config for `avgProofTime + delay` to let anyone submit proof, which is a scenario where the prover has faile to submit proof in time, and then they lose their deposit.
- fix weighting (just placeholder right now)
- add some mechanism for the prover client to be able to determine which batchs are available for bidding, current bid, etc.


packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,190 @@
...
+        TaikoData.Config memory config,
+        TaikoData.State storage state,
+        uint256 batchId
+    ) internal view returns (bool) {
+        // auction is always open if there is no winner.
+        if (state.blockAuctionBids[batchId].account == address(0)) return true;
+
+        // if auctionLengthInSeconds hasnt passed, the auction is still open
+        if (
+            state.blockAuctionBids[batchId].auctionStartedAt +
```
dantaik, line 150: If we use the following check for aunctions:

```
uint firstBatch = (nextBlockId << 8) + 1;
uint lastBatch = firstBatch + config.auctionBatchGap
require(batchId >=firstBatch && batchId <= lastBatch);
```

Nobody can win a batch the (in the same L1 tx) prove the blocks immediately, which create outbid opportunities for other bidders. If the delay between bidding and proving should be larger, then we can change the above to:


```
uint firstBatch = (nextBlockId << 8) + 2; // changed from 1 to 2.
uint lastBatch = firstBatch + config.auctionBatchGap
require(batchId >=firstBatch && batchId <= lastBatch);
```

Now maybe we can get rid of auctionLengthInSeconds.

cyberhorsey:
@dantaik because we dont want auction in the next testnet, I am going to ignore this PR, because we are still making constant changes elsewhere, and keeping it up-to-date will be time-consuming but also not a good use of time before `alpha-3`.

When your `major_protocol_upgrade_rebase` branch is merged to main, with all relevant changes, I will revisit the auction mechanism.

dantaik:
Not sure we covered this:
We have to take into consideration of the different circuits (they have different proving time) used by different blocks.

PR #13541, opened 2023-04-05
feat(protocol): refactor L2 genesis generation scripts / tests by foundry && update `ANCHOR_GAS_COST`

davidtaikocha:
Refactor L2 genesis generation scripts / tests by foundry.

Test output:
```
Logs:
  TaikoL2.anchor gas cost after 256 L2 blocks: 127202
```
I feel that this output value is little bit lower than my expectation, will double check it later.


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -19,8 +19,7 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, IXchainSync {
     using SafeCastUpgradeable for uint256;
     using LibMath for uint256;
 
-    // TODO(david): need you to confirm this value.
-    uint64 public ANCHOR_GAS_COST = 47000;
+    uint64 public ANCHOR_GAS_COST = 150000; // owner:david
```
dantaik, line 22: this is way more than I thought :(

packages/protocol/test2/LibFixedPointMath.t.sol
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: UNLICENSED
+// SPDX-License-Identifier: MIT
```
dantaik, line 1: This belongs to a third party, so maybe keep it UNLICENSED

dantaik (approved):

packages/protocol/test2/LibFixedPointMath.t.sol
davidtaikocha: Changed.

PR #13546, opened 2023-04-05
feat(protocol): Prover rewards (proposer fees) discussed changes

adaki2004:
Discussed changes:
- Rearrange state vars to fitting into slots based on where are they modified
- Instead of gasLimit (which was anyhow a constant) use a constant with which we can tune the TKO precision

Note:
This PR does not contain the deletion of the 2 algos (`allow minting` + `non-minting & non time-weighted)`. This will be done once the functionality of the `non-minting & time weighted` mechanism is verified with test.

UPDATE:
Went for the time weighted mechanism, and implemented 3 different rewarding calculations. (2 is commented out, needs to be tested, which formula could fit best.)
See additional calculation formulas: https://github.com/taikoxyz/taiko-mono/pull/13564#discussion_r1160769597

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -120,27 +120,27 @@ library TaikoData {
         // mapping(address account => uint64 balance) balances;
         mapping(address account => uint256 balance) balances;
         mapping(bytes32 txListHash => TxListInfo) txListInfo;
-        // Cummulated proofTime for reward calculation - changed in verifyBlock()
-        uint64 proofTimeIssued;
-        // Changing baseFee for proving - changed in verifyBlock()
-        uint64 basefee;
-        // Changing last verified block to a new id - changed in verifyBlock()
-        uint64 lastVerifiedBlockId;
-        uint64 __reserved1;
-        // Changing accumulated time for proposing - changed in proposeBlock() and in verifyBlock()
-        uint64 accProposedAt;
-        // Treasury amount - changed in proposeBlock() and in verifyBlock()
-        uint64 rewardPool;
-        uint128 __reserved2;
-        // Never or rarely changed
+        // Slot 5: never or rarely changed
         uint64 genesisHeight;
         uint64 genesisTimestamp;
-        uint128 __reserved3;
-        // Changing when a block is proposed - changed in proposeBlock()
-        uint64 numBlocks;
-        // Changing timestamp when a block is proposed - changed in proposeBlock()
+        uint64 __reserved51;
+        uint64 __reserved52;
+        // Slot 6: changed by proposeBlock
         uint64 lastProposedAt;
+        uint64 numBlocks;
+        uint64 accProposedAt; // also by verifyBlocks
+        uint64 rewardPool; // also by verifyBlocks
+        // Slot 7: changed by proveBlock
+        // uint64 __reserved71;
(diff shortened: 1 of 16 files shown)


packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -23,6 +23,21 @@ library LibTokenomics {
...
+    ///
+    /// @dev To give more context, the fee and reward: (if the proof arrives at around target time)
+    /// IF DECIMAL_FACTOR = 1_000_000;
+    /// with proofTImeTarget 85 sec (current testnet) apprx.: 782.7 TKO is the reward (both with and without minting)
+    /// with proofTimeTarget is 20 mins (on mainnet): apprx.: 55.4 TKO (approx. both with and without minting)
+
+    /// IF DECIMAL_FACTOR = 100_000;
+    /// with proofTImeTarget 85 sec (current testnet) apprx.: 78.27 TKO is the reward
+    /// with proofTimeTarget is 20 mins (on mainnet): apprx.: 5.54 TKO
+    uint32 private constant DECIMAL_FACTOR = 100_000;
```
Brechtpd, line 39: To set reasonable starting values I think better to set some rough expected target values for the initial fee and calculate the corresponding `proofTimeIssued` value, that way there are fewer magic constants. From the python code:
```
basefee_proof = 10**8
proof_time_issued = (PROOF_TIME_TARGET * PROOF_ADJUSTMENT_QUOTIENT) * math.log(basefee_proof * (PROOF_TIME_TARGET * PROOF_ADJUSTMENT_QUOTIENT))
```

So if we're not interested into taking gas into account anymore (which I generally agree with because some of the costs are completely unrelated to L2 gas so not really a good estimate), we don't really _have_ to multiply with any factor, `basefee == fee` directly.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -119,30 +120,27 @@ library TaikoData {
...
+        // Slot 6: changed by proposeBlock
+        uint64 lastProposedAt;
         uint64 numBlocks;
-        uint64 lastProposedAt; // Timestamp when the last block is proposed.
-        uint64 __reserved3;
-        uint64 __reserved4;
-        // Changed when a block is proven/finalized
+        uint64 accProposedAt; // also by verifyBlocks
+        uint64 rewardPool; // also by verifyBlocks
+        // Slot 7: changed by proveBlock
```
dantaik, line 133: @adaki2004 I pushed a commit to re-org the data layout here.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -130,23 +130,20 @@ contract TaikoL1 is EssentialContract, IXchainSync, TaikoEvents, TaikoErrors {
...
-        (, feeAmount) = LibTokenomics.getProverFee(state, gasUsed);
+    function getProverFee() public view returns (uint64 feeAmount) {
+        (, feeAmount) = LibTokenomics.getProverFee(state);
     }
 
     function getProofReward(
         uint64 provenAt,
         uint64 proposedAt,
-        uint32 gasUsed
+        uint32 gasUsed // TODO(daniel): remove this one?
```
dantaik, line 140: `gasUsed` is not used, delete it please

Brechtpd (approved):

PR #13552, opened 2023-04-06
refactor(protocol): split oracle proving into a function

dong77:
This PR allows the oracle prover to batch-prove multiple blocks in one L1 transaction to save cost. These are the following benefits:
- Now the call data is minimized to reduce cost.
- A block can be oracle-proven multiple times. In case the oracle prover makes a mistake, he can redo the proving with different hashes.


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -22,23 +22,91 @@ library LibProving {
...
+                unchecked {
+                    ++blk.nextForkChoiceId;
+                }
+                assert(fcId > 0);
+
+                fc = blk.forkChoices[fcId];
+                state.forkChoiceIds[id][oracle.parentHash] = fcId;
+            } else {
+                fc = blk.forkChoices[fcId];
+                if (fc.prover != address(0)) revert L1_ALREADY_PROVEN();
```
Brechtpd, line 87: Maybe still allow overwriting even when already proven by another prover? Only possible when ZKPs are incomplete/buggy, but we're likely to need something that'll allow us to do this somehow.

Brechtpd (approved):

PR #13560, opened 2023-04-06
feat(protocol): make parentGasUsed verifiable on L2

dantaik:
A still work-in-progress PR to show how to make parent gas used verifiable to circuits.

The idea is to change:
```mapping(uint256 blockId => mapping(bytes32 parentHash => uint256 forkChoiceId)) forkChoiceIds;``` to
```mapping(uint256 blockId => mapping(bytes32 parentHash => mapping(uint32 parentGasUsed => uint256 forkChoiceId))) forkChoiceIds;```

TODO:
- [x] make sure storage layout is optimized

#### Previously
| proposeBlock                           | 22420           | 32541  | 22440  | 92439  | 309     |
| proveBlock                             | 55427           | 65481  | 55427  | 131827 | 309     |
| verifyBlocks                           | 13576           | 14589  | 14294  | 80373  | 301     |

#### Now
| proposeBlock                           | 22420           | 32541  | 22440  | 92439  | 309     |
| proveBlock                             | 56108           | 66142  | 56108  | 132508 | 309     |
| verifyBlocks                           | 14086           | 15142  | 14804  | 85883  | 301     |


Brechtpd:
I can  think of 3 options to use gas used:
1. Make the parent gas used available on L1: Implemented in this PR. Pretty easy, the downsides are pretty small: slightly more complicated fork choice logic, a bit higher gas costs on L1.
2. Use the current block gas used (without anchor tx gas used): Only uses data from the current block so no need to store any data anywhere from a previous block. The most important downside I think is that the node kind of has to execute the transactions in the block twice: Once with a dummy gas used value in the anchor tx, then do all transactions to find out how much gas those transactions use. Then the transactions are executed again with the correct gas used in the anchor tx. (need to make sure none of the of the variables impacted by gas_used are exposed to other smart contracts because then this wouldn't work!).
3. Instead of a single pre-block anchor transaction, we have a pre-block and post-block transaction. This way we solve the problem in 2. where the transactions need to be executed twice, we just execute the transactions once and then use the gas_used of those transactions in the post-block tx.

I think all fairly simple, though none are without downsides. 3. could in theory be useful for other things as well.

dantaik:
> I can think of 3 options to use gas used:
> 
> 1. Make the parent gas used available on L1: Implemented in this PR. Pretty easy, the downsides are pretty small: slightly more complicated fork choice logic, a bit higher gas costs on L1.
> 2. Use the current block gas used (without anchor tx gas used): Only uses data from the current block so no need to store any data anywhere from a previous block. The most important downside I think is that the node kind of has to execute the transactions in the block twice: Once with a dummy gas used value in the anchor tx, then do all transactions to find out how much gas those transactions use. Then the transactions are executed again with the correct gas used in the anchor tx. (need to make sure none of the of the variables impacted by gas_used are exposed to other smart contracts because then this wouldn't work!).
> 3. Instead of a single pre-block anchor transaction, we have a pre-block and post-block transaction. This way we solve the problem in 2. where the transactions need to be executed twice, we just execute the transactions once and then use the gas_used of those transactions in the post-block tx.
> 
> I think all fairly simple, though none are without downsides. 3. could in theory be useful for other things as well.

By comparison, I prefer the first approach :)

Brechtpd (approved):

PR #13564, opened 2023-04-06
feat(protocol): Test time weighted solution and cut rest if good

adaki2004:
Algo of the time weighted mechanism is validated but fine tuning of param might be necessary.
@Brechtpd knows every bits and pieces of the small issues in question.

To be discussed / determined:
- initBasefee proper init value (?)
- proofTimeIssued proper init value (?)

Todos:
- Cutting out the 'allow minting' version from the code.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibVerifying.sol
+++ b/packages/protocol/contracts/L1/libs/LibVerifying.sol
@@ -134,7 +134,6 @@ library LibVerifying {
                     uint64(proofTime)
                 );
 
-            // Todo (dani) kiiratast hogy a reward meg a basefee ugyan az-e !?
             state.basefee = newBasefee;
             state.proofTimeIssued = proofTimeIssued;
 
--- a/packages/protocol/contracts/L1/libs/LibVerifying.sol
+++ b/packages/protocol/contracts/L1/libs/LibVerifying.sol
@@ -31,7 +31,8 @@ library LibVerifying {
         TaikoData.State storage state,
         TaikoData.Config memory config,
         bytes32 genesisBlockHash,
-        uint64 initBasefee
+        uint64 initBasefee,
+        uint64 initProofTimeIssued
     ) internal {
         _checkConfig(config);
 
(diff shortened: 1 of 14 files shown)


packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -39,8 +39,7 @@ library LibVerifying {
         state.genesisHeight = uint64(block.number);
         state.genesisTimestamp = timeNow;
 
-        // TODO(dani): should be a parameter in init().
-        state.basefee = initBasefee; // 1 Taiko Token (1**8 * 1**6 = tested starter fee - eliminating magic values like decimal_factor)
+        state.basefee = initBasefee;
```
Brechtpd, line 42: Also set `state.proofTimeIssued` here so it doesn't start at `0`, but starts at the value that will give `initBasefee`. Probably easiest to just pass it in as a function input because otherwise we need `ln` in the smart contract to calculate it, though that may be a pain to keep up-to-date and with tests written in solidity maybe easier to just have some test-only code for `ln` from here: https://github.com/recmo/experiment-solexp/blob/main/src/FixedPointMathLib.sol

The way to calculate it is like this (also don't forget to take into account `SCALING_FROM_18_FIXED_EXP_TO_TKO_AMOUNT`):

```
scale = config.proofTimeTarget * config.adjustmentQuotient
proof_time_issued = scale * math.log(initBasefee * scale)
```

Make sure `proof_time_issued` is a reasonable value, at least as big as ~`1000 * config.proofTimeTarget` or something like that so we have enough room to go up and down with `proof_time_issued`. You may want to adjust `SCALING_FROM_18_FIXED_EXP_TO_TKO_AMOUNT` if the `proof_time_issued` you get is too small.

adaki2004: @Brechtpd Could you please confirm or deny something ? :)

In the python script it is like that:
```
def eth_amount(value, target, quotient):
    return math.exp(value / target / quotient)

def base_fee(value, target, quotient):
    return eth_amount(value, target, quotient) / (target * quotient)
```

It seems to me we divide with `(target * quotient)` twice. If so, i think the math here does not work - i think if we (in the Solidity code) we divide once, that will give the same value then in this math.
```
scale = config.proofTimeTarget * config.adjustmentQuotient
proof_time_issued = scale * math.log(initBasefee * scale)
```
 (Not 100% sure,tho).

PR #13565, opened 2023-04-07
feat(protocol): allow multiple Taiko deployments to sync signals

dantaik:
Imagine two Taiko layer 3s (Taiko-L3A and Taiko-L3B) are deployed on a Taiko-L2  and the layer 3s share the same SignalService instance on L2. The demand to transfer funds (or send signals in general) across these layer 3s will be common. 

This change makes sure when Taiko-L3A's signalRoot is sent over to L2, the L2 contract will relay the signalRoot to the shared SignalService as a signal, then when Taiko-L3B gets the L2's signalRoot, that signalRoot carries info about Taiko-L3A's signalRoot. Verifying the signal on Taiko-L3B will require two (partial) Merkle proofs.

This change will increase `verifyBlocks`'s gas cost from 14086 to 41325, that's because we always write to new signal slots. But since this feature will only be turned for Taiko layer 3 deployments (their *TaikoL1.sol* contracts will be on Taiko L2), the extra gas cost shall be fine.


packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -106,6 +108,14 @@ library LibVerifying {
             unchecked {
                 state.lastVerifiedBlockId += processed;
             }
+
+            if (config.relaySignalRoot) {
+                // Send the L2's signal root to the signal service so other TaikoL1
+                // deployments, if they share the same signal service, can relay the
+                // signal to their corresponding TaikoL2 contract.
+                ISignalService(resolver.resolve("signal_service", false))
+                    .sendSignal(signalRoot);
```
Brechtpd, line 117: This will always store to the same slot I think no? Just checking because I think that is also how it should work, but you wrote "because we always write to new signal slots" so just want to make sure I got this right.

Brechtpd (approved):
Just two storage proofs is a nice reduction in costs!

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: The signal service is designed to never rewrite any slots. For each new value of `signalRoot`, it always ends up with a different slot and the actual data stored in the slot is `1`.

Brechtpd: Ah right I see now, forgot how it worked.

PR #13570, opened 2023-04-07
feat(protocol): update `TaikoL2.getBasefee` parameters

davidtaikocha:


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -215,14 +215,10 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, IXchainSync {
      **********************/
 
     function getBasefee(
-        uint32 timeSinceNow,
+        uint32 timeSinceParent,
```
dantaik, line 218: @davidtaikocha Lets talk about it, I think you may missed something here.

davidtaikocha: lets talk about it tomorrow.

dantaik (approved):

PR #13571, opened 2023-04-07
refactor(protocol): optimize proveBlock gas cost

dantaik:
For the first forkchoice, we do not write the `forkChoiceIds` mapping but write a `key` value in the ForkChoice itself. Then when a block is to be verified, when we cannot find a non-zero value in the `forkChoiceIds` mapping lookup, we assume it's the first fork choice, then verify the key matches. This ends up with the same numbers of SSTOREs, but the `key` of the first fork choice is always reused thus will save gas. I assume the first fork choice for most blocks are the correct fork choise.
 

#### Before
| proveBlock                             | 53045           | 60752  | 53045  | 129445 | 99      |
| verifyBlocks                           | 14804           | 14885  | 14804  | 22886  | 99      |

#### Now
| proveBlock                             | 34461           | 44822  | 34461  | 132761 | 99      |
| verifyBlocks                           | 14872           | 14953  | 14872  | 22960  | 99      |


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -155,9 +174,16 @@ library LibProving {
...
+            ] = fcId == 1 ? 0 : fcId;
 
             TaikoData.ForkChoice storage fc = blk.forkChoices[fcId];
+            if (fcId == 1) {
+                // We only write the key when fcId is 1.
+                fc.key = LibUtils.keyForForkChoice(
+                    evidence.parentHash,
+                    evidence.parentGasUsed
+                );
+            }
```
Brechtpd, line 186:
```suggestion
            TaikoData.ForkChoice storage fc = blk.forkChoices[fcId];
             // Only write the key for the first fork choice, 
             // use the mapping when there are multiple fork choices
            if (fcId == 1) {
                fc.key = LibUtils.keyForForkChoice(
                    evidence.parentHash,
                    evidence.parentGasUsed
                );
            } else {
              state.forkChoiceIds[blockId][evidence.parentHash][evidence.parentGasUsed] = fcId;
            }
```
Same as above, maybe could be its own internal function.

packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -28,6 +28,23 @@ library LibUtils {
...
 
+    function getForkChoiceId(
+        TaikoData.State storage state,
+        TaikoData.Block storage blk,
+        uint256 blockId,
+        bytes32 parentHash,
+        uint32 parentGasUsed
+    ) internal view returns (uint256) {
+        uint256 fcId = state.forkChoiceIds[blockId][parentHash][parentGasUsed];
+        if (fcId >= blk.nextForkChoiceId) return 0;
```
Brechtpd, line 39:
```suggestion
        bytes32 parentHash,
        uint32 parentGasUsed
    ) internal view returns (uint256) {
        uint256 fcId = state.forkChoiceIds[blk.blockId][parentHash][parentGasUsed];
```
Maybe useful to do it like this because this automatically rules out some cases that should not happen and makes sure the input data is consistent.

Brechtpd (approved):

PR #13579, opened 2023-04-09
refactor(protocol): minor optimize signal service merkle proof

dantaik:
In theory, the gas cost shall become smaller, but it is not confirmed by tests.


packages/protocol/contracts/thirdparty/LibSecureMerkleTrie.sol
```diff
@@ -88,6 +88,6 @@ library LibSecureMerkleTrie {
     function _getSecureKey(
         bytes memory _key
     ) private pure returns (bytes memory _secureKey) {
-        return abi.encodePacked(keccak256(_key));
+        return bytes.concat(keccak256(_key));
```
dantaik, line 91: @cyberhorsey do we really need this extra `keccak256` here?

cyberhorsey (approved):

PR #13598, opened 2023-04-14
refactor(bridge-ui): storage services as modules

jscriptcoder:
Storage and Token services have been turned into modules. There is no need to add them into stores nor initialize them in the App.svelte component. As soon as they're imported, they're instantiated. 

What are we gaining here?, cleaner code and less initialization in the App.svelte. Unnecessary use of svelte stores.

TODO: we can do the same with the RelayerAPI service


dantaik:
Ready for review?

shadab-taiko (approved):

cyberhorsey (approved):

PR #13611, opened 2023-04-15
refactor(protocol): optimize proving logics [3]

dantaik:
- Do not need a BlockMetadata to prove a block, which reduces data cost
- *Cache* a staticRefs hash so we do not to need resolve addresses for some addresses each time we prove a block. This also saves gas.

## Previous

| Function Name                                   | min             | avg    | median | max    | # calls |
|-------------------------------------------------|-----------------|--------|--------|--------|---------|
| proposeBlock                                    | 21894           | 32015  | 21914  | 91913  | 309     |
| proveBlock                                      | 33700           | 47151  | 33700  | 132100 | 309     |
| verifyBlocks                                    | 14773           | 15842  | 15491  | 90328  | 301     |

## Now

| Function Name                                   | min             | avg    | median | max    | # calls |
|-------------------------------------------------|-----------------|--------|--------|--------|---------|
| proposeBlock                                    | 21762           | 31883  | 21782  | 91781  | 309     |
| proveBlock                                      | 19095           | 32481  | 19095  | 113495 | 309     |
| verifyBlocks                                    | 14773           | 15842  | 15491  | 90328  | 301     |


packages/protocol/script/deploy_on_l1.sh
```diff
@@ -9,6 +9,7 @@ ORACLE_PROVER=0xa0Ee7A142d267C1f36714E4a8F75612F20a79720 \
 SOLO_PROPOSER=0xa0Ee7A142d267C1f36714E4a8F75612F20a79720 \
 OWNER=0x70997970C51812dc3A010C7d01b50e0d17dc79C8 \
 TAIKO_L2_ADDRESS=0xa0Ee7A142d267C1f36714E4a8F75612F20a79720 \
+L2_SIGNAL_SERVICE=0x0000777700000000000000000000000000000007 \
```
dantaik, line 12: @davidtaikocha , how to figure out these values? I ran `pnpm test:genesis` and it fails.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -38,63 +38,56 @@ library LibProving {
...
 
         if (isOracleProof) {
             address oracleProver = resolver.resolve("oracle_prover", true);
             if (oracleProver == address(0)) revert L1_ORACLE_DISABLED();
-
             if (msg.sender != oracleProver) {
-                if (evidence.zkproof.data.length == 64) {
-                    uint8 v = uint8(evidence.zkproof.verifierId);
+                if (evidence.proof.length == 64) {
+                    uint8 v = uint8(evidence.verifierId);
```
Brechtpd, line 68: Why not put `v` in `proof` as well?

Brechtpd (commented):
> Cache a staticRefs hash so we do not to need resolve addresses for some addresses each time we prove a block. This also saves gas.

I kind of assumed that for an actual release most/all addresses returned by resolve would be hardcoded constant values so very cheap to read. Is that the plan or is the idea that they will remain storage values?

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: no perticular reason, just want to reuse the verifiedId field. Otherwise, we'll have a zero there too.

packages/protocol/script/deploy_on_l1.sh
davidtaikocha: they are defined in https://github.com/taikoxyz/taiko-mono/blob/opimize_proving/packages/protocol/test2/genesis/test_config.json#L36, to make the L2 pre-deployed contract addresses deterministic

dantaik:
> > Cache a staticRefs hash so we do not to need resolve addresses for some addresses each time we prove a block. This also saves gas.
> 
> I kind of assumed that for an actual release most/all addresses returned by resolve would be hardcoded constant values so very cheap to read. Is that the plan or is the idea that they will remain storage values?

I think this requires us to compose some code dynamically during deployment.

packages/protocol/script/deploy_on_l1.sh
dantaik: `TAIKO_L2_ADDRESS= 0xa0Ee7A142d267C1f36714E4a8F75612F20a79720 ` is different from the one in test_config.json

Brechtpd (approved):

davidtaikocha (approved):

PR #13632, opened 2023-04-21
test(protocol): Signal service foundry tests

cyberhorsey:
the `major_protocol_upgrade_rebase` branch had a `.only` on a SignalService test, so the tests werent actually running. Only a single test was
. Now that I've deleted that test, this shows that 6 of them are now failing, but they were always failing on your branch @dantaik , just hidden.

Should we just scrap the hardhat tests, or add another `.only` so they "appear" to pass? I'm in favor of just deleting them all rather than putting it any work to make them pass.


dantaik:
@cyberhorsey  There are testing failures.

cyberhorsey:
> @cyberhorsey There are testing failures.

Please read the main comment of this PR

dantaik:
> 

Lets delete them.

cyberhorsey:
> > 
> 
> Lets delete them.

Done, plus I changed the package.json commands, and Github Actions.

dantaik (approved):

PR #13640, opened 2023-04-23
feat(protocol): major protocol upgrade for alpha-3 testnet

### LINKED ISSUE #13662: Merge big protocol PR changes
- [ ] Merge the big PR #13640 
- [ ] How to verify L2 EIP1559 fee recipient in protocol or ZK? [The circuits will verify the ETH sending to a designated address, we need to document this (Dani)]
- [ ] #13661

dantaik:
## test/TaikoL1LibTokenomicsTestnet.t.sol:TaikoL1WithTestnetConfig

| Function Name                                                            | min             | avg    | median | max    | # calls |
|--------------------------------------------------------------------------|-----------------|--------|--------|--------|---------|
| proposeBlock                                                             | 24976           | 70459  | 90776  | 104876 | 325     |
| proveBlock                                                               | 20066           | 84091  | 107966 | 155305 | 315     |
| verifyBlocks                                                             | 6026            | 14783  | 14748  | 23548  | 325     |


packages/protocol/contracts/L1/TaikoConfig.sol
```diff
@@ -0,0 +1,47 @@
...
+                minEthDepositsPerBlock: 8,
+                maxEthDepositsPerBlock: 32,
+                maxEthDepositAmount: 10000 ether,
+                minEthDepositAmount: 1 ether,
+                // Set it to 120KB, since 128KB is the upper size limit
+                // of a geth transaction, so using 120KB for the proposed
+                // transactions list calldata, 8K for the remaining tx fields.
+                maxBytesPerTxList: 120000,
+                minTxGasLimit: 21000,
+                proofCooldownPeriod: 5 minutes,
```
Brechtpd, line 37: For mainnet I think this will need to be much higher because this works in the same way as a fraudproof, so the same censorship concerns.

dantaik: Change it to 30 minutes?

Brechtpd: Still seems low to me, but I think the balance is the amount of money that can be stolen (up to the TVL of the rollup) and the cost to censor the oracle proof transaction on L1 for a certain period of time. So if the TVL is low the challenge period can be quite short, but if the TVL is high then probably needs a long window (like optimistic rollups).

dantaik: I see, but if this value is large, we need to find a liquidity bridge partner.

dantaik: I think for now, we can use 30 minutes for testnets.

packages/protocol/test/Taiko1559Params.t.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: UNLICENSED
```
d1onys1us, line 1:
```suggestion
// SPDX-License-Identifier: MIT
```

davidtaikocha (dismissed):

packages/protocol/contracts/thirdparty/LibFixedPointMath.sol
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: UNLICENSED
```
d1onys1us, line 1: not to be nitpicky but, i believe UNLICENSED is an issue. i think this breaks our ability to be MIT licensed technically, even though I doubt it will be a real issue with the maintainer. i reached out to recmo on twitter to see if we can use the MIT license.

Brechtpd (dismissed):
Only some nitpicks and suggestions.

cyberhorsey (dismissed):

cyberhorsey:
I approved this PR because it changed 296 files, and is so big, we will never go back to what is in `main` over this. It needs to be merged.

packages/protocol/contracts/thirdparty/LibFixedPointMath.sol
dantaik: Can we create an issue and fix it later?

dantaik:
Will fix all comments today for another review.

dantaik:
READY for review again.

davidtaikocha (approved):

Brechtpd (approved):

PR #13656, opened 2023-04-26
fix(bridge-ui): fix claim eth message processed

jscriptcoder:
Fixed message already processed when claiming ETH

### DIFF
--- a/packages/bridge-ui/src/bridge/ETHBridge.ts
+++ b/packages/bridge-ui/src/bridge/ETHBridge.ts
@@ -100,8 +100,12 @@ export class ETHBridge implements Bridge {
       opts.msgHash,
     );
 
-    if ([MessageStatus.Done, MessageStatus.Failed].includes(messageStatus)) {
+    if (messageStatus === MessageStatus.Done) {
       throw Error('message already processed');
+    }
+
+    if (messageStatus === MessageStatus.Failed) {
+      throw Error('user can not process this, message has failed');
     }
 
     const signerAddress = await opts.signer.getAddress();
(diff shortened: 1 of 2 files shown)


packages/bridge-ui/src/bridge/ETHBridge.ts
```diff
@@ -100,7 +100,7 @@ export class ETHBridge implements Bridge {
       opts.msgHash,
     );
 
-    if (messageStatus === MessageStatus.Done) {
+    if ([MessageStatus.Done, MessageStatus.Failed].includes(messageStatus)) {
```
adaki2004, line 103: If not mistaken when `MessageStatus.Failed` is the status, then it is not processed yet. There need to be a 'releaseEther()'

dantaik (approved):

packages/bridge-ui/src/bridge/ETHBridge.ts
adaki2004: If Claimed called after releaseEther it's fine i guess - donno when / how automatized it is on the relayer side.

jscriptcoder: We should then throw a different error message here?, In the UI when the status is Failed we show the Release button, claiming will no longer be available, so theoretically the user shouldn't be able to call this function when messageStatus is Failed... Let me make that change

adaki2004: Ok, yes, so basically its kinda not possible for regular users, so then has not too much effect anyways.

jscriptcoder: I made a change to throw different error when the message has failed. One could argue though that the message has failed and therefore already processed. @cyberhorsey, what do you think?

In V2 I'd like to use error codes instead of messages

cyberhorsey (approved):

PR #13671, opened 2023-04-27
docs(website): add new presentations and update blog posts

mfinestone:
quick notes on the presentations page:

- thinking if we should specify if each item is a deck or a video in the description with a [video] or [deck] tag.
- if there's a video of the pres, do we only share video, or add the deck as well? (assuming video > deck in terms of value to share, which may not be true)


dionysuzx (approved):

d1onys1us:
@mfinestone i think we should add both the deck and the video, let me see if i can come up with a new formatting.

dantaik (approved):

PR #13672, opened 2023-04-28
feat(protocol): ensure non-zero 1559 basefee

dantaik:


davidtaikocha (approved):

Brechtpd (approved):

Brechtpd:
Do we think there are any potential problems with 0 basefee or this is just to make sure that the behavior is the same?

dantaik:
> Do we think there are any potential problems with 0 basefee or this is just to make sure that the behavior is the same?

I actually don't think there is any problem with 0 basefee. So mostly just to make sure the behaviour is the same.

PR #13681, opened 2023-04-29
feat(protocol): prevent oracle prover overriding the same proof

dantaik:
1. I think we should prevent oracle prover to override the same fork choice as implemented.
2. Cooldown period is disabled if:
    - The oracle prover is address(0)
    - The oracle prover proves blocks that doesn't need real ZKPs.

Tests will need to be added (tracked by https://github.com/taikoxyz/taiko-mono/issues/13686)


dantaik:
@adaki2004 maybe we should ignore out the simulation out file files (png and other files)?

adaki2004:
> @adaki2004 maybe we should ignore out the simulation out file files (png and other files)?

Hmm, def. not include all, just maybe having some there as a demonstration/benchmark , or at least ppl who try to run, knows how the output looks like. (?).
1 sample shall be enough i guess.

davidtaikocha (approved):

Brechtpd (approved):

PR #13704, opened 2023-05-03
feat(protocol): add parentGasUsed to blockproven

cyberhorsey:
We also could use the `parentGasUsed` in the taiko-client on `BlockProven` events, but LibProving is at maximum stack, so I had to remove one variable.

Correspondign taiko-client pr: https://github.com/taikoxyz/taiko-client/pull/207


packages/protocol/contracts/L1/TaikoEvents.sol
```diff
@@ -17,7 +17,8 @@ abstract contract TaikoEvents {
         bytes32 parentHash,
         bytes32 blockHash,
         bytes32 signalRoot,
-        address prover
+        address prover,
+        uint32 parentGasUsed
```
davidtaikocha, line 21: hmmm since `parentHash` field is in this event, not sure if its necessary to compare the `parentGasUsed` in client again?

cyberhorsey: @Brechtpd what do you think? you mentioned in the taiko-client PR to compare both

dantaik: I think it's OK to have `parentGasUsed` in the event.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -246,4 +246,8 @@ contract TaikoL1 is
     function getVerifierName(uint16 id) public pure returns (bytes32) {
         return LibUtils.getVerifierName(id);
     }
+
+    function getOracleProver() public view returns (address) {
```
dantaik, line 250: If we add this function, shall we also add more for other named addresses?

cyberhorsey: I don't necessarily need them, but we could actually just do a generic function.
Ie:

function resolveAddressForName(string name) returns (address)

And then we can just pass in "oracle_prover", etc

packages/protocol/contracts/L1/TaikoEvents.sol
Brechtpd: > hmmm since `parentHash` field is in this event, not sure if its necessary to compare the `parentGasUsed` in client again?

The prover can submit any `parentGasUsed` value, so if the prover did not submit the correct value for it to the smart contract the proof that was submitted will be in the wrong fork choice so we still need people to submit a valid proof with both the correct `parentHash` and `parentGasUsed` for the block to be able to get verified. Or do I miss something?

davidtaikocha (dismissed):

dantaik (approved):

davidtaikocha (approved):

PR #13705, opened 2023-05-04
fix(bridge-ui): update abis

jscriptcoder:
- Updated ABIs
- Using wagmi/cli to copy ABIs from protocol/abi (`pnpm export:abi`)
  - Generates `index.ts` exporting ABIs
- Increase coverage on recommendedProcessingFee

TODO: run exporting ABIs (`pnpm export:abi`) in the script before copying the json files?


dantaik (approved):

dantaik (dismissed):

packages/bridge-ui/scripts/copy-abi.js
```diff
@@ -0,0 +1,39 @@
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
dantaik, line 6: should we deleted all old files before copying?

jscriptcoder: They'll be overwritten, but yes, that's a good idea in case of change of name

dantaik:
> * Updated ABIs
> * Script to copy ABIs from protocol/abi (`pnpm export:abi`)
>   
>   * Generates `index.ts` exporting ABIs
> * Increase coverage on recommendedProcessingFee
> 
> TODO: run exporting ABIs (`pnpm export:abi`) in the script before copying the json files?

Maybe we should add a github hook to automate the abi export?

packages/bridge-ui/scripts/copy-abi.js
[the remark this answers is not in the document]
d1onys1us: if it's a low hanging fruit like short task then i'd say we can try it. but if it takes too much time i agree let's just go with this custom script.

dionysuzx (dismissed):

dantaik (approved):

dionysuzx (approved):

PR #13707, opened 2023-05-04
feat(protocol): Change back token decimals

adaki2004:
Change back token decimal settings to 8 from 18, because it affects not only tests but the code itself.
`uint64` range of the baseFee currently implies some limitations, so decimal 8 would perfectly fit in but 18 needs to involve some other changes.


adaki2004:
We can change to 18 and uint256 later, but requires some changes and in case we need to deploy testnet soon, better to stick to it.

Brechtpd (approved):
Makes sense not to break things for a3, but let's still do the change afterwards!

davidtaikocha (approved):

PR #13729, opened 2023-05-09
feat(protocol): Introduce oracle and system prover concept

### LINKED ISSUE #13727: feat(protocol): add a shorter `proofCooldownPeriod` for oracle proofs

adaki2004:
Modifications:
- Separated Oracle proofs and System proofs (System proofs was the skipped ones, usually called it sometimes "fake proofs", but decided not a good wording.)

Oracle proofs: 
- They have to be overwritten (by regular proofs) - otherwise cannot be verified
- They cannot overwrite the existing proof if it's the same
- They can overwrite if FC's data differs

System proofs:
- They are overwriteable and also verifiable as well (UPDATE: was not overwriteable first, we changed)

Related to : https://github.com/taikoxyz/taiko-mono/pull/13728


dantaik:
> 

Lets create an issue for it.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -129,15 +130,30 @@ library LibProving {
...
                 fc.blockHash == evidence.blockHash &&
                 fc.signalRoot == evidence.signalRoot &&
                 fc.gasUsed == evidence.gasUsed
             ) revert L1_SAME_PROOF();
         } else {
-            revert L1_ALREADY_PROVEN();
+            // This is the branch provers trying to overwrite
+            fc = blk.forkChoices[fcId];
+            // Only allow overwrite in case:
+            // - previous prover is oracle prover or system proof
```
dantaik, line 144: I think we can now remove these comments as they are exactly the code below.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -146,7 +162,8 @@ library LibProving {
         fc.provenAt = uint64(block.timestamp);
         fc.prover = evidence.prover;
 
-        if (evidence.prover != address(0)) {
+        // evidence.prover cannot be address(0) at this point - either oracle, system, or a regular prover
```
dantaik, line 165: Update this comment please

dantaik (dismissed):

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -60,42 +63,40 @@ library LibProving {
...
+                }
+
+                // clear the proof before hashing evidence
+                evidence.verifierId = 0;
+                evidence.proof = new bytes(0);
+
+                if (
+                    specialProver !=
+                    ecrecover(keccak256(abi.encode(evidence)), v, r, s)
+                ) revert L1_NOT_ORACLE_PROVER();
```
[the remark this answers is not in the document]
cyberhorsey: If we change this we need to change in Taiko client as well as we check for this custom error, I will monitor

dantaik (dismissed):

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -146,7 +163,9 @@ library LibProving {
         fc.provenAt = uint64(block.timestamp);
         fc.prover = evidence.prover;
 
-        if (evidence.prover != address(0)) {
+        // evidence.prover cannot be address(0) at this point -
+        // either system or a regular prover
```
Brechtpd, line 167: Why not? Seems to get there if the fc is being overwritten by the oracle prover or do I misunderstand the comment?

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -111,9 +108,8 @@ library LibVerifying {
 
             if (fc.prover == address(0)) break;
 
-            uint256 proofCooldownPeriod = oracleProver == address(0) ||
-                fc.prover == oracleProver
-                ? 0
+            uint256 proofCooldownPeriod = fc.prover == address(1)
```
Brechtpd, line 111: Rewards will get sent to `address(1)`, may be problematic if we want to give these tokens value?

packages/protocol/contracts/L1/libs/LibProving.sol
adaki2004: Old commenting from prev. version, thanks for the note!

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -129,15 +137,24 @@ library LibProving {
...
             if (
                 fc.blockHash == evidence.blockHash &&
                 fc.signalRoot == evidence.signalRoot &&
                 fc.gasUsed == evidence.gasUsed
             ) revert L1_SAME_PROOF();
         } else {
-            revert L1_ALREADY_PROVEN();
+            // This is the branch provers trying to overwrite
+            fc = blk.forkChoices[fcId];
+            if (fc.prover != address(0) && fc.prover != address(1))
```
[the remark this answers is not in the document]
adaki2004: Meant to be overwriteable, but after discussion, we changed it, so regular provers can overwrite it now.

packages/protocol/contracts/L1/libs/LibVerifying.sol
adaki2004: Well, yes we can / should figure out something. Currently rewards (for the system prover) will be held in the protocol (`state.taikoTokenBalances[address(1)]`).
Some notes:
- It will have a 15 min cool down period, so people might overwrite this block so they will be beneficiaries.
- We can add a specific role/withdraw function to claim this reward on behalf of X.
- Or we can even set a specific address at reward amount accounting if `address(1)`

Let me know which one would be best.

packages/protocol/contracts/L1/libs/LibProving.sol
Brechtpd: Still gotta skip the proof verification below if `&& evidence.prover != address(0)` as well no?

packages/protocol/contracts/L1/libs/LibVerifying.sol
Brechtpd: Last option seems simplest to me, but no strong opinion.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -129,15 +137,24 @@ library LibProving {
...
+                revert L1_ALREADY_PROVEN();
```
davidtaikocha, line 151: lets say `realProofSkipSize` is `10`, and right now we have a new block whose id is `10`. Then we submit an oracle proof to the protocol, after that, we need to submit a real proof for that block, but the real proof submission will run into this `else` logic and get a `L1_ALREADY_PROVEN` custom error?

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: If prover == address(1), simple send to "system_prover".

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: With the current code,  the error will not occur. As `fc.prover == address(0)`.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -146,7 +163,7 @@ library LibProving {
         fc.provenAt = uint64(block.timestamp);
         fc.prover = evidence.prover;
 
-        if (evidence.prover != address(0)) {
+        if (evidence.prover != address(1)) {
```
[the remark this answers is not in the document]
dantaik: You are right, fixed.

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: Yes, fixed.

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: Fixed

dantaik (approved):

davidtaikocha (approved):

PR #13742, opened 2023-05-11
feat(protocol): make sure system proof delay is proofTimeTarget

dantaik:
@davidtaikocha is it going to make your client/prover logic simpler?


davidtaikocha (approved):
@davidtaikocha is it going to make your client/prover logic simpler?

it will!

adaki2004 (approved):

PR #13743, opened 2023-05-11
feat(protocol): send system-proof reward back to proposer

dantaik:
Is it a good idea for our testnet?

If we do this, proposers will get their tokens back very soon then they can propose more blocks with their tokens. But this may also create spamming issues.


Brechtpd:
Seems to open up the possibility to spam indeed. Maybe a better way is to simply reduce the payment by the proposer as `prover_fee / realProofSkipSize`?

Though the prover fee on its own not the best way to prevent spam anyway because the prover fee can be made very low. As long as the testnet ETH is scarce enough eip-1559 should do its work for preventing spam I think.

PR #13745, opened 2023-05-12
feat(protocol): Enable protocol to handle multiple proof types

### LINKED ISSUE #13731: feat(protocol): Even more basic multi-prover with PoA
Simplify oracle/system prover system, in a way that will be compatible with https://github.com/taikoxyz/taiko-mono/issues/13693:

- Blocks need a zk proof + a signature (currently just one we generate and share somehow)
- For proofs that need to be skipped only a valid signature is required, the zk proof verification is skipped. Here we force the proof time to be the target proof time to not impact the base fee calculation.

Then, currently optional because not needed, we just have an additional fail safe mechanism to overwrite and lock a fork choice directly, not going through the normal proving code in proveBlock.

adaki2004:
Related to: https://github.com/taikoxyz/taiko-mono/issues/13731

- [x] Code changes required to verify multiple proofs
- [x] Code changes required to overwrite unverified fork choice
- [x] Test changes of multiple proofs (ZKP & SGX) + overwrite unverified FK
- [x] Adjust other tests to changes


packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -30,6 +30,14 @@ library TaikoData {
...
         uint64 proofTimeTarget;
         uint8 adjustmentQuotient;
+        // // BITMAP for efficient iteration and flexible additions later
+        // ZKP_ONLY,            // 0000 0001
+        // SGX_ONLY,            // 0000 0010
+        // RESERVED_X_ONLY,     // 0000 0100
+        // RESERVED_Y_ONLY,     // 0000 1000
+        // ZKP_AND_SGX,         // 0000 0011
+        // X_ZKP_SGX,           // 0000 0111
+        uint8 proofTypeEnabled;
```
Brechtpd, line 40: Nice! Maybe a more clear name would be `proofToggleMask` or something.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -138,87 +89,106 @@ library LibProving {
...
-                uint160(
-                    address(
-                        resolver.resolve(
-                            config.chainId,
-                            "signal_service",
-                            false
-                        )
+        // config.proofTypeEnabled can support 8 different proofs on 8 bits
+        for (uint8 index; index < 8; ) {
+            if (index == 0 && ((config.proofTypeEnabled >> index) & 1) == 1) {
```
Brechtpd, line 136:
```suggestion
            if (((config.proofTypeEnabled >> index) & 1) == 1) {
               if (index == 0) {
                 // ...
               } else if (index == 1) {
                 // ...
               }
```

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -138,87 +89,106 @@ library LibProving {
...
-            );
-            inputs[1] = uint256(
-                uint160(
-                    address(
-                        resolver.resolve(
-                            config.chainId,
-                            "signal_service",
-                            false
-                        )
+        // config.proofTypeEnabled can support 8 different proofs on 8 bits
```
Brechtpd, line 134: The plan is still to abstract the actual implementations away in the verifier, or do you think a simple for loop with some ifs would also work? Seems okay to me like this, but I'm mostly concerned about the additional data needed for the different proof types, having a separate contract for each allows things to be very flexible with what a new proof type needs

Brechtpd (commented):
Looking good!

packages/protocol/contracts/L1/libs/LibProving.sol
adaki2004: Yep, agree, thanks for the remark! Outsourced to new files now.

packages/protocol/contracts/L1/libs/proofTypes/LibVerifySGX.sol
```diff
@@ -0,0 +1,44 @@
...
+pragma solidity ^0.8.18;
+
+import {AddressResolver} from "../../../common/AddressResolver.sol";
+import {LibUtils} from "../LibUtils.sol";
+import {TaikoData} from "../../TaikoData.sol";
+
+library LibVerifySGX {
+    error L1_INVALID_SGX_SIGNATURE();
+
+    function verifySgxProof(
```
dantaik, line 16: I think we can also use this library for non-sgx signers, so I suggested to change the name to LibVerifyTrusted

dantaik, line 16: rename to verifyProof?

packages/protocol/contracts/L1/libs/proofTypes/LibVerifyZKP.sol
```diff
@@ -0,0 +1,29 @@
...
+pragma solidity ^0.8.18;
+
+import {AddressResolver} from "../../../common/AddressResolver.sol";
+import {LibUtils} from "../LibUtils.sol";
+import {TaikoData} from "../../TaikoData.sol";
+
+library LibVerifyZKP {
+    error L1_INVALID_PROOF();
+
+    function verifyZkProof(
```
dantaik, line 16: rename to verifyProof?

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -74,6 +81,11 @@ library TaikoData {
         TaikoData.EthDeposit[] depositsProcessed;
     }
 
+    struct BlockProof {
```
dantaik, line 84: How about we added a verifierType here?
```uint16 proofType```,

File: packages/protocol/contracts/L1/libs/LibProving.sol [hunk 7]
```diff
@@ -138,87 +89,75 @@ library LibProving {
...
+            (uint256(evidence.parentGasUsed) << 64) |
+            (uint256(evidence.gasUsed) << 32);
+
+        bytes32 instance;
+        assembly {
+            instance := keccak256(inputs, mul(32, 9))
+        }
+
+        // config.proofToggleMask can support 8 different proof types on 8 bits
+        for (uint8 index; index < 8; ) {
```
dantaik, line 135: Should we do something like this?

```solidity
uint8 mask = config.proofToggleMask;

for (uint i = 0; i < blockProofs.length; i++) {
  TypedProof memory proof = blockProofs[i];
  uint8 bitMask = 1 << (proof.proofType - 1);
  require(mask & bitMask) != 0, "type_not_enabled_or_duplicated");
  _verifyTypedProof(proof);
  mask &= ~bitMask;
}
require(mask == 0, "not_all_requierd_proof_verified");
```

packages/protocol/contracts/L1/TaikoData.sol
dantaik, line 84: BlockProof => TypedProof

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -30,6 +29,14 @@ library TaikoData {
...
         uint64 proofTimeTarget;
         uint8 adjustmentQuotient;
+        // // BITMAP for efficient iteration and flexible additions later
+        // ZKP_ONLY,            // 0000 0001
+        // SGX_ONLY,            // 0000 0010
+        // RESERVED_X_ONLY,     // 0000 0100
+        // RESERVED_Y_ONLY,     // 0000 1000
+        // ZKP_AND_SGX,         // 0000 0011
+        // X_ZKP_SGX,           // 0000 0111
+        uint8 proofToggleMask;
```
dantaik, line 39: uint16?

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: I also suggest that the cooldown period is a value returned from `_verifyTypedProof`:

```solidity
uint8 mask = config.proofToggleMask;
uint64 maxCoolDownPeriod;
for (uint i = 0; i < blockProofs.length; i++) {
  TypedProof memory proof = blockProofs[i];
  uint8 bitMask = 1 << (proof.proofType - 1);
  require(mask & bitMask) != 0, "type_not_enabled_or_duplicated");
  uint64 coolDownPeriod = _verifyTypedProof(proof);
  maxCoolDownPeriod = maxCoolDownPeriod.max(coolDownPeriod);
  mask &= ~bitMask;
}
require(mask == 0, "not_all_requierd_proof_verified");
fc.provenAt = uint64(block.timestamp +maxCoolDownPeriod );
```

dantaik: Inside, LibVerifyerSGX and LibVerifyingZKP, in the verify function, we also return a cooldown period. This allows for different cooldown period for different verifiers, then we just take the max.

dantaik:
I think this PR still has a high level design issue: the `proofToggleMask` can only specify an AND logics, but not something like: *if a valid SGX proof is given, then ZKP is not necessary; *.

dantaik:
Need to remvoe ""oracle_prover" and "system_prover" everywhere.

Brechtpd:
> I think this PR still has a high level design issue: the `proofToggleMask` can only specify an AND logics, but not something like: *if a valid SGX proof is given, then ZKP is not necessary; *.

Could you give a scenario where we would use conditional logic like that? I saw this only really as an all proofs that are enabled are required thing.

I could see some benefit of the different proofs coming in at different times (like SGX proof first, which would set the cooldown period still pretty high, then a zkp proof that lowers it) but in practice I think probably doesn't really make a difference (we'll always want the zkp anyway, which is the slowest, and all of them being required all the time is best for security), and complicates things quite a bit like the fee payments.

We could still implement something like this in a verifier contract, the main protocol would would just require a single proof related to that verifier contract, but internally the verifier contract can do whatever it wants so could contain multiple sub-proofs with some special logic. This would actually be a pretty nice way to implement the testnet scenario where we want to skip some ZKP proofs, then there's no need to build this into the core protocol.

So maybe let's simplify the core protocol as much as possible and only let it support a single "proof" (as in some data and a verifier contract to call), then let the verifier contract handle the rest.

packages/protocol/contracts/L1/libs/LibProving.sol
Brechtpd: > Inside, LibVerifyerSGX and LibVerifyingZKP, in the verify function, we also return a cooldown period. This allows for different cooldown period for different verifiers, then we just take the max.

Maybe min makes more sense otherwise adding a proof type can increase the cooldown, although cooldown period length will always a bit tricky and handwavy and not sure if something that can be decided per proof type.

> Question to this cooldown period: Since this affects `proofTime = provenAt - proposedAt` and cooldown only delays verification, why should not we let the the `LibVerifying.sol` do this job and revert ?

What do you mean with revert?

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -183,6 +189,34 @@ library LibProving {
...
+        TaikoData.Config memory config,
+        AddressResolver resolver,
+        uint256 blockId,
+        TaikoData.BlockEvidence memory evidence
+    ) internal {
+        if (msg.sender != resolver.resolve("forkchoice_failsafe", true)) {
+            revert L1_NO_AUTH_TO_OVERWRITE_FK();
+        }
+
+        if (blockId <= state.lastVerifiedBlockId || blockId >= state.numBlocks) {
```
Brechtpd, line 203: Small but not completely trivial logic that is shared with the prover code, maybe a bit better to move this to a util function called `isBlockBeingProven` or something.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -204,6 +189,34 @@ library LibProving {
...
+            revert L1_BLOCK_ID();
+        }
+
+        TaikoData.Block storage blk = state.blocks[blockId % config.ringBufferSize];
+        //fcId in such case shall be always at array pos 1
```
Brechtpd, line 208:
```suggestion
        // We make it so this will always be the first fork choice
```

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -204,6 +189,34 @@ library LibProving {
...
+        TaikoData.Config memory config,
+        AddressResolver resolver,
+        uint256 blockId,
+        TaikoData.BlockEvidence memory evidence
+    ) internal {
+        if (msg.sender != resolver.resolve("forkchoice_failsafe", true)) {
+            revert L1_NO_AUTH_TO_OVERWRITE_FK();
+        }
+
+        if (blockId <= state.lastVerifiedBlockId || blockId >= state.numBlocks) {
```
Brechtpd, line 203: Small but not completely trivial logic that is shared with the prover code, maybe a bit better to move this to a util function called isBlockBeingProven or something.

Brechtpd: Hmm I would expect the function to get inlined. I see the number of runs is also set pretty low: `optimizer_runs = 200` which may impact this. Maybe measure again after increasing that value.

In any case the number of runs used by the optimizer probably should be set much higher in general, it used to make quite a bit of a difference if set to a high value (10,000 or something) for some free gas savings.

Brechtpd:
Probably makes more sense to do the opposite which I think is what we already do, make a release for alpha-n, and keep working on main.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -204,6 +159,41 @@ library LibProving {
         });
     }
 
+    function setForkChoice(
+        TaikoData.State storage state,
+        TaikoData.Config memory config,
+        AddressResolver resolver,
+        uint256 blockId,
+        TaikoData.BlockEvidence memory evidence
```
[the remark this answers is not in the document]
Brechtpd: Hmmm why? The functionality of this is significantly different from `proveBlock`, which made `proveBlock` really complicated before.

dantaik:
@adaki2004  given that our core protocol and tokenomics have changed a lot, shall we close this PR then re-implement the idea later?

PR #13750, opened 2023-05-13
docs(protocol): add a markdown to document how blocks are proven

dantaik:
Dani, please view the file on github using this link:https://github.com/taikoxyz/taiko-mono/blob/2ab1dd16fc708b2559ae400f0a7567dfb147bd5d/packages/protocol/docs/ZKReq.md (the link may change if you edit the file).

Please migrate our internal doc into this file and update the diagram accordingly. The diagram is partially incorrect as many suggestions in your commented PDF file are not incorporated. I'll leave this PR to you.

When this PR is ready, please ask Brecht, Yue, David, and Jeff for review.


packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,294 @@
...
+In addition, ZKP must also prove the following:
+
+- `parentHash` must be the same as `evidence.parentHash`.
+- `ommersHash` must be `0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347`.
+- `beneficiary` must be `meta.beneficiary` (duplicated, as stated above).
+- `logsBloom` must be a bytes32[8] with all zeros.
+- `difficulty` must be `meta.mixHash`.
+- `height` must be `meta.id`.
+- `gasLimit` == `meta.gasLimit`.
+- `gasUsed` ≤ `gasLimit` (we are assuming the EVM proof will verify that gasLimit has the right value), and `gasUsed` == `evidence.gasUsed`
```
Brechtpd, line 127:
```suggestion
- `gasUsed` ≤ `gasLimit` (we are assuming the ZKP will verify that gasLimit has the right value), and `gasUsed` == `evidence.gasUsed`
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,294 @@
...
+- `timestamp` == `meta.proposedAt`.
+- `extraData` == "".
+- `mixHash` == `meta.mixHash`.
+- `nonce` == 0.
+- `baseFeePerGas` == 0.
```
Brechtpd, line 132:
```suggestion
- `baseFeePerGas` == the calculated EIP-1559 style base fee.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,294 @@
...
+- `withdrawalsRoot` == `0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421`  according to [https://github.com/ethereum/go-ethereum/blob/master/core/types/hashes.go#L41](https://github.com/ethereum/go-ethereum/blob/master/core/types/hashes.go#L41)
+
+Note that some of the above header field checks are duplicates of checks from the Block Variable section.
+
+## Signal Storage Root
+
+ZKP also needs to prove cross chain signal service’s storage roots have the right value.
+
+- **For L2 Signal Service**:  the storage root is the second parameter in the anchor transaction. ZKP shall verify that the storage root of the L2 Signal Service address indeed has this given value by using zkEVM storage proof.
```
Brechtpd, line 141:
```suggestion
- **For L2 Signal Service**:  the storage root is the second parameter in the anchor transaction. The ZKP shall verify that the storage root of the L2 Signal Service address has this given value by using a standard zkEVM storage proof.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,294 @@
...
+- **For L1 Signal Service**: the storage root verification shall be done in circuits by using a Merkle proof as witness against `meta.l1Hash`. The `l1SignalRoot` value is not directly provided by L1 smart contracts as it is not available to EVM and must be queried by L2 client from a L1 node.
```
Brechtpd, line 142:
```suggestion
- **For L1 Signal Service**: the storage root verification will be done in the circuits by using a Merkle proof against the `meta.l1Hash`, the `l1SignalRoot` value is not directly available in the L1 smart contracts. This MPT proof must be queried from an L1 node by the L2 client.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,294 @@
...
+a_h1_hash(h1Hash)
+a_h1_signal_root[h1SignalRoot]
+a_parent_gas_used[parentGasUsed]
+end
+
+subgraph L1Storage
+b_signal_root[signalRoot]
+b_l1_taiko_addr[taikoL1Address]
+b_l1_signal_service_addr[L1 signalServiceAddress]
+b_l2_signal_service_addr[L2 signalServiceAddress]
```
Brechtpd, line 258: I guess it depends a bit what you see under `L1Storage`, but I wouldn't put the `L2 signalServiceAddress` here.

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,294 @@
...
+# The *BlockMetadata*
+
+This struct represents the metadata of a block. The data will be hashed and be part of the public input. So anything within this struct, will be hashed and put into ZKP `instance`.
+
+- `id`: Represents the block height in L2.
+- `timestamp`: The block timestamp in L2.
+- `l1Height`: The actual block height in L1.
+- `l1Hash`: The actual block hash in L1.
+- `mixHash`: Salted random number to accommodate multiple L2 blocks fitting into one L1 block.
+- `depositsRoot`: Matches the withdrawalsRoot in the L2 header.
```
Brechtpd, line 76:
```suggestion
- `depositsRoot`: Hash of the list of deposits.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,294 @@
...
+- `difficulty` must be `meta.mixHash`.
+- `height` must be `meta.id`.
+- `gasLimit` == `meta.gasLimit`.
+- `gasUsed` ≤ `gasLimit` (we are assuming the EVM proof will verify that gasLimit has the right value), and `gasUsed` == `evidence.gasUsed`
+- `timestamp` == `meta.proposedAt`.
+- `extraData` == "".
+- `mixHash` == `meta.mixHash`.
+- `nonce` == 0.
+- `baseFeePerGas` == 0.
+- `withdrawalsRoot` == `0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421`  according to [https://github.com/ethereum/go-ethereum/blob/master/core/types/hashes.go#L41](https://github.com/ethereum/go-ethereum/blob/master/core/types/hashes.go#L41)
```
Brechtpd, line 133:
```suggestion
- `withdrawalsRoot` == The withdrawals MPT correspoding to the deposits list.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,294 @@
...
+v_blockhash_others -.-> dot1 -.->|keccak| s_public_input_hash;
+
+v_block_basefee -.-> h_basefee;
+
+v_block_gaslimit -.-> dot2;
+v_block_timestamp -.-> dot2;
+s_parent_timestamp -.-> dot2;
+s_gas_excess -.-> dot2 ---|calcBasefee| v_block_basefee;
+
+processed_deposits -.->|keccak| m_deposits_root --- h_withdrawals_root;
```
Brechtpd, line 178: It's not the same value, but it's deduced from it so needs to be a dotted line?

Brechtpd (commented):
Couple of comments, overall looks good, I haven't checked everything yet.

adaki2004 (dismissed):

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+  - `blockhash(block.number - 1)`, has the same value as in the block header and is also the same value as the parent block's hash on L2.
+  - The other 255 hashes, `blockhash(block.number - 256)` to `blockhash(block.number - 2)` are checked in the anchor transaction to simplify circuits. Therefore, as long as the anchor transaction is zk-proven, these 255 ancestor hashes are proven indirectly.
+- `block.basefee`: verified to be the correct value in the anchor transaction.
+- `block.chainid`: this field is also checked by the anchor transaction, so no extra ZKP circuits are required.
+- `block.coinbase`: ZKP must verify the value must be the same as `meta.beneficiary`. Again, the metadata hash is part of the ZK instance.
+- `block.difficulty`: this is now the same as `block.prevrandao`, so we only check `block.prevrandao`.
+- `block.gaslimit`: ZKP must verify this value must equal `meta.gasLimit`.
+- `block.gasUsed`: ZKP must verify this value must be smaller than `meta.gasLimit`.
```
Brechtpd, line 131: `gasUsed` is not accessible in the EVM so should be removed here.

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+- `block.number`: this must be checked against the block header and `meta.id`.
+- `block.prevrandao`: this must be checked against the `mixHash` field in the L2 block header and `meta.mixHash`.
+- `block.timestamp`: this must be checked against the `timestamp` field in the L2 block header and `meta.proposedAt`.
+
+### Block Header
+
+The block header is no longer presented to L1 contracts; therefore, ZKP must verify the integrity of the block header and ensure the block header can hash into the same value as `evidence.blockHash`, which is part of the ZK instance input.
```
Brechtpd, line 138:
```suggestion
Not all block header data is available in the L1 contracts; therefore, the ZKP must verify the integrity of the block header and ensure the block header hashes into the same value as `evidence.blockHash`, which is part of the ZK instance input.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+}
+```
+
+In addition, ZKP must also prove the following:
+
+- `parentHash` must be the same as `evidence.parentHash`.
+- `ommersHash` must be the keccak256 of `[]`,or `0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347`.
+- `beneficiary` must be `meta.beneficiary` (duplicated, as stated above).
+- `logsBloom` must be a `bytes32[8]` with all zeros.
+- `difficulty` must be `meta.mixHash`.
```
Brechtpd, line 168:
```suggestion
- `difficulty` == 0.
```

As @ggkitsas has found, this should not be the `mixHash`, the `DIFFICULTY` opcode just returns the `mixHash/prevRandao`. Its value in the block header is simply `0`: https://blog.ethereum.org/2021/11/29/how-the-merge-impacts-app-layer. In the diagram below it is correct.

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+- `height` must be `meta.id`.
+- `gasLimit` == `meta.gasLimit`.
+- `gasUsed` ≤ `gasLimit` (we are assuming the ZKP will verify that gasLimit has the right value), and `gasUsed` == `evidence.gasUsed`
```
Brechtpd, line 171:
```suggestion
- `gasUsed` must be the sum of all gas used by all valid transactions in the block.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+- `timestamp` == `meta.proposedAt`.
+- `extraData` == "".
+- `mixHash` == `meta.mixHash`.
+- `nonce` == 0.
+- `baseFeePerGas` == the calculated EIP-1559 style base fee.
+- `withdrawalsRoot` == The withdrawals MPT correspoding to the deposits list.
```
Brechtpd, line 177:
```suggestion
- `withdrawalsRoot` == The withdrawals MPT root corresponding to the deposits list.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+
+Note that some of the above header field checks are duplicates of checks from the Global Variable section.
```
Brechtpd, line 179:
```suggestion
Note that some of the header field checks above are duplicates of checks done in the Global Variable section.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+
+### Signal Storage
+
+ZKP also needs to prove cross chain signal service’s storage roots have the right values.
```
Brechtpd, line 183:
```suggestion
The ZKP also needs to prove that the cross chain signal service’s storage roots have the correct values.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+
+- **For L2 Signal Service**: the storage root is the second parameter in the anchor transaction. ZKP shall verify that the storage root of the L2 Signal Service address indeed has this given value by using zkEVM storage proof.
```
Brechtpd, line 185:
```suggestion
- **For L2 Signal Service**: the L1 storage root of the signal service is the second parameter in the anchor transaction. The ZKP shall verify that the storage root of the L1 Signal Service address has the given value by using an MPT proof against the state root stored in `meta.l1Hash` for the `l1SignalServiceAddress` account. This MPT proof must be queried by the L2 client from an L1 node .
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+
+- **For L1 Signal Service**: the storage root verification will be done in the circuits by using a Merkle proof against the `meta.l1Hash`, the `l1SignalRoot` value is not directly available in the L1 smart contracts. This MPT proof must be queried from an L1 node by the L2 client.
```
Brechtpd, line 187:
```suggestion
- **For L1 Signal Service**: the L2 storage root verification will be done in the circuits by using an MPT proof against the post block state root for the `l2SignalServiceAddress` account.
```

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Data Consistency
+
+Note that `txList` is not directly know to L1 contracts, therefore, ZKP shall further prove that the chosen `txList` is a slice of the given blob data (from `meta.txListByteStart` to `meta.txListByteEnd`) and the blob’s hash or polynomial commitment (PC) is `meta.blobHash` — not the `txList`’s hash or PC). ( @todo/question: blobHash is not part of the contracts anymore.)
+
+## Anchor Transactions
+
+Every L2 block has exactly one anchor function call as its first transaction.
+
+ZKP must prove that *TaikoL2.anchor(...)* is the first transaction in the block, with the correct input parameters and gas limit, signed by the gold-touch address, and did not throw.
```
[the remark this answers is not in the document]
dantaik: It's still a regular transaction based on what I know.

packages/protocol/docs/ZKReq.md
```diff
@@ -0,0 +1,363 @@
...
+- `l1Height`: The actual block height in L1.
+- `l1Hash`: The actual block hash in L1.
+- `mixHash`: Salted random number to accommodate multiple L2 blocks fitting into one L1 block.
+- `depositsRoot`: Hash of the list of deposits.
+- `txListHash`: Hash of the transaction list in L2.
+- `txListByteStart`: Byte start of the transaction list in L2.
+- `txListByteEnd`: Byte end of the transaction list in L2.
+- `gasLimit`: Gas limit for the L2 block.
+- `beneficiary`: The address of the beneficiary in L2.
+- `cacheTxListInfo`: Indicates whether the transaction list info should be cached or not.
```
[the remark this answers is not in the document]
dantaik: This `cacheTxListInfo` indicates whether or not to cache the current `txList` in the transaction if the  `txList` is not empty. 

To use an cached txList (regardless if it is a blob or a calldata), the current block shall provide an empty `txList` to indicate to reuse the cached txList data.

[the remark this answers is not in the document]
Brechtpd: Sure, but that has no impact on the circuit which always just gets the `txListHash` (or blob commitment in the future) as the input no?

packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -0,0 +1,360 @@
...
+- Has no contract code deployed on the sender account (see EIP-3607 by Feist et al. [2021]).
+- Has a gas limit no smaller than the intrinsic gas, _`g0`_, used by the transaction; and the sender account balance contains at least the cost, _`v0`_, required in up-front payment.
+
+#### Slicing and Consistency
+
+Note that the data of `txList` cannot be assumed to be directly accessible to L1 contracts, and so the ZKP shall further prove that the chosen `txList` is a slice of the given blob data. The `blobHash` opcode will be available when EIP-4844 is merged into Ethereum.
+
+### Anchor Transactions
+
+Every L2 block has exactly one anchor function call as its first transaction.
```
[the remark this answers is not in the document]
Brechtpd: Yep that's the idea. But quickly looking at the anchor code I'm not sure if that's currently the case! I don't know if we have a test for that, but maybe making it very explicit by making sure the L2 block number is unique would be safer.

packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -0,0 +1,360 @@
...
+- `difficulty` == 0.
+- `height` must be `meta.id`.
+- `gasLimit` == `meta.gasLimit`.
+- `gasUsed` must be the sum of all gas used by all valid transactions in the block.
+- `timestamp` == `meta.proposedAt`.
+- `extraData` == "".
+- `mixHash` == `meta.mixHash`.
+- `nonce` == 0.
+- `baseFeePerGas` == the calculated EIP-1559 style base fee.
+- `withdrawalsRoot` == The withdrawals MPT root corresponding to the deposits list.
```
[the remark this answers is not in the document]
Brechtpd: Oh we store that hash directly in our L2 block header? Maybe better to then just store the empty MPT root hash instead for now so we are at least compatible with standard Ethereum block headers?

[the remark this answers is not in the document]
davidtaikocha: I think in current codebase, yes. Here is the calculation of the `depositRoot`: https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/L1/libs/LibEthDepositing.sol#L107 which the corresponding `header.withdrawRoot` needs to equal to.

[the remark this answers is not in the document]
Brechtpd: Fair enough for now, but something I think we should change in the future.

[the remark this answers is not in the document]
dantaik: One reason we store the hash of a list of `EthDeposit` is to reduce cost -- hashing the MPT root requires more `keccek` ops.

packages/protocol/docs/how_taiko_proves_blocks.md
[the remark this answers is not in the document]
dantaik: We have this test: testAnchorTxsFailInTheSameBlock already.

packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -0,0 +1,360 @@
...
+    uint64 parentGasUsed
+  ) external;
+  ```
+- A circuit will verify the integrity among: `l1Hash`, `l1SignalRoot`, and `l1SignalServiceAddress`
+- `l1SignalServiceAddress`, `l2SignalServiceAddress` and `parentGasUsed` are directly hashed into the ZKP's instance
+- `l1Height` and `l1Hash` are both part of the block metadata (`meta.l1Height` and `meta.l1Hash`), the `metaHash` is used to calculate the ZKP instance.
+- `l1SignalRoot` is part of the evidence and is also used to calculate the ZKP instance.
+- The transaction's status code is 0 (success).
+- The transaction's `tx.origin` and `msg.sender` must be _`LibAnchorSignature.K_GOLDEN_TOUCH_ADDRESS`_.
+- The transaction's signature must be the same as `LibAnchorSignature.signTransaction(...)`.
```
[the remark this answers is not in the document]
dantaik: Added

packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -0,0 +1,361 @@
...
+- `logsBloom` must be a `bytes32[8]` with all zeros.
+- `difficulty` == 0.
+- `height` must be `meta.id`.
+- `gasLimit` == `meta.gasLimit`.
+- `gasUsed` must be the sum of all gas used by all valid transactions in the block.
+- `timestamp` == `meta.proposedAt`.
+- `extraData` == "".
+- `mixHash` == `meta.mixHash`.
+- `nonce` == 0.
+- `baseFeePerGas` == the calculated EIP-1559 style base fee.
```
davidtaikocha, line 174: Hmm im not sure if ZKP must *also* prove this, since if the anchor transaction's status is `0`, which means this has already been proved?

davidtaikocha, line 174: And it needs to also prove that all base fee was transferred to `treasury` address?

dantaik: I think you are right: the ZKP doesn't have to prove the base fee calculation, but it should verify that `header.baseFeePerGas === block.basefee`

packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -0,0 +1,361 @@
...
+    bytes32 l1SignalRoot,
+    uint64 l1Height,
+    uint64 parentGasUsed
+  ) external;
+  ```
+- A circuit will verify the integrity among: `l1Hash`, `l1SignalRoot`, and `l1SignalServiceAddress`
+- `l1SignalServiceAddress`, `l2SignalServiceAddress` and `parentGasUsed` are directly hashed into the ZKP's instance
+- `l1Height` and `l1Hash` are both part of the block metadata (`meta.l1Height` and `meta.l1Hash`), the `metaHash` is used to calculate the ZKP instance.
+- `l1SignalRoot` is part of the evidence and is also used to calculate the ZKP instance.
+- The transaction's status code is 0 (success).
```
davidtaikocha, line 54:
```suggestion
- The transaction's status code is 1 (success).
```

packages/protocol/docs/how_taiko_proves_blocks.md
dantaik: Added a small section for it.

packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -0,0 +1,361 @@
...
+
+The transaction list, `txList`, of a Layer 2 (L2) block will eventually be part of a blob in the Layer 1 (L1) Consensus Layer (CL) -- right now it is within the calldata.
+It is meant to be a list of RLP-encoded L2 transactions, with only its length and commitment available to the L1 Execution Layer (EL). The length is checked when a block is proposed, but we need to prove whether the `txList` is valid or not using a ZKP.
+
+A valid `txList` (until [issue #13724](https://github.com/taikoxyz/taiko-mono/issues/13724) is enabled and implemented):
+
+- Has a byte-size smaller than the protocol constant _`maxBytesPerTxList`_ (also enforced in contracts);
+- Can be RLP-decoded into a list of transactions without trailing space;
+- Contains no more transactions (valid and invalid) than the protocol constant _`maxTransactionsPerBlock`_;
+- Has a total gas limit for all valid transactions not exceeding the protocol constant _`blockMaxGasLimit`_.
```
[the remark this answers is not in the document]
dantaik: Added

packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -0,0 +1,366 @@
...
+
+For a valid `txList`, ZKP must also prove whether each enclosed transaction is valid or invalid. All valid transactions will be part of the actual L2 block with an anchor transaction as the very first one.
+
+A valid transaction (defined in the Ethereum Yellow Paper):
+
+- Has a valid transaction signature.
+- Has a valid transaction nonce (equivalent to the sender account's current nonce).
+- Has no contract code deployed on the sender account (see EIP-3607 by Feist et al. [2021]).
+- Has a gas limit no smaller than the intrinsic gas, _`g0`_, used by the transaction; and the sender account balance contains at least the cost, _`v0`_, required in up-front payment.
+- Has a gas limit shall be no smaller than the protocol constant `_minTxGasLimit_`.
```
dantaik, line 29: Actually, @dantaik  _minTxGasLimit_ is g0

packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -0,0 +1,365 @@
...
+dot2((" "))
+end
+
+GlobalVariables:::group
+
+
+subgraph Anchor [Anchor Tx]
+a_h1_height(h1Height)
+a_h1_hash(h1Hash)
+a_h1_signal_root(h1SignalRoot)
```
davidtaikocha, line 319:
```suggestion
a_l1_height(l1Height)
a_l1_hash(l1Hash)
a_l1_signal_root(l1SignalRoot)
```

davidtaikocha (dismissed):

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -172,9 +172,14 @@ library LibProving {
             inputs[8] = (uint256(uint160(evidence.prover)) << 96)
                 | (uint256(evidence.parentGasUsed) << 64) | (uint256(evidence.gasUsed) << 32);
 
+            // Also hash configs that will be used by circuits
```
dantaik, line 175: @davidtaikocha  please take a look at this.

davidtaikocha: yah i believe they are also necessary. @smtmfft @johntaiko

davidtaikocha (dismissed):

adaki2004 (dismissed):

packages/protocol/docs/how_taiko_proves_blocks.md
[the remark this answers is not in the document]
Brechtpd: Oh yeah I'm not saying we should modify the solidity code, the `depositsRoot` of that data is just a circuit input so we can do whatever we want there, no problem. It's just that the value stored in the L2 block header for this should ideally be the MPT root of this data, and that MPT root would be calculated in the circuits (or we could just leave it be the empty MPT tree hash for now).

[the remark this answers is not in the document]
dantaik: so `depositsRoot` is still the flat hash of the _deposits_, and `withdrawalsRoot` (which is in the header) is the MPT root of the same deposit list?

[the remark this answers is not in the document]
davidtaikocha: Not sure if its hard for circuits? but it will reduce lots of changes which we've made in go-ethereum.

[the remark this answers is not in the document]
Brechtpd: Should be very similar to creating the tx/log trie, so should be fine. I honestly also think an acceptable (perhaps intermediate) solution is just setting the withdrawals root to simply the empty MPT root in the block header, it's not like the value is that useful. And there are ways to make depositing ETH even cheaper than this mechanism, so maybe won't even be used in the future. Is having the actual data in the MPT trie root important for the node?

> but it will reduce lots of changes which we've made in go-ethereum.

Can you explain a bit more why it required a lot of changes?

packages/protocol/contracts/L1/libs/LibProving.sol
johntaiko: Is this change will in a3 testnet?

johntaiko: Can we remove the `parentHash` and `gasUsed`, because these are contained in the `blockHash`?

packages/protocol/docs/how_taiko_proves_blocks.md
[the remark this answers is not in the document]
dantaik: Brecht, it would be ideal to utilize the `withdrawalsRoot` as it is because the go-ethereum code already supports this feature, requiring no additional action on our part. When I say *use the withdrawalsRoot as is*, I mean:

- The structure of an _EthDeposit_ should be defined in the same way as an ETH2 withdrawal. (@davidtaikocha  Could you please provide the definition)?

- We should hash these structures in the same manner as in ETH2.

Currently, our _EthDeposit_ definition is independent of the ETH2 withdrawal data structure, and we hash all deposits as a simple array instead of a tree. That's why David mentioned that we had to modify the go-ethereum codebase with additional code. In other words, we only re-use the `withdrawalsRoot`, not the corresponding go-ethereum code.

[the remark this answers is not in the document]
davidtaikocha: here is the definition:
image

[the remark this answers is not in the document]
davidtaikocha: > Brecht, it would be ideal to utilize the `withdrawalsRoot` as it is because the go-ethereum code already supports this feature, requiring no additional action on our part. When I say _use the withdrawalsRoot as is_, I mean:
> 
> * The structure of an _EthDeposit_ should be defined in the same way as an ETH2 withdrawal. (@davidtaikocha  Could you please provide the definition)?
> * We should hash these structures in the same manner as in ETH2.
> 
> Currently, our _EthDeposit_ definition is independent of the ETH2 withdrawal data structure, and we hash all deposits as a simple array instead of a tree. That's why David mentioned that we had to modify the go-ethereum codebase with additional code. In other words, we only re-use the `withdrawalsRoot` field in the block header, not the corresponding go-ethereum code.

and i think right now there is at least one different between the protocol's `EthDeposit` and eth2's `Withdrawal`, which is the unit of `EthDeposit.Amount` in protocol is `wei`, but when it comes to eth2's `Withdrawal.Amount`, the unit is `gwei`.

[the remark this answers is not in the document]
Brechtpd: My main question is still if `withdrawalsRoot` is actually used in the node to do anything? If you change it to the root of an empty tree would everything still work on the node side (except not being able to query what withdrawals happened in a block, which seems okay to me)?

Of course somewhere on the node side we'll still have to get the ETH deposits from the smart contract (this could be easy just using data from events or something). Then I would think it takes that data and translates it into whatever the data structure is already used in the node to actually execute these withdrawals so I would expect the actual node changes to be quite small. I assume here that that data structure is actually totally unrelated to the withdrawals MPT, which I would think is just be there to make the data accessible in an efficient way (like the tx tree), and is otherwise unused in the node while actually executing the block.

packages/protocol/contracts/L1/libs/LibProving.sol
Brechtpd: We need `parentHash` to make sure the blocks are correctly linked with each other. `gasUsed` we need in the smart contract, it's not directly available in the block hash as far as I know.

packages/protocol/docs/how_taiko_proves_blocks.md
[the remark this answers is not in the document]
dantaik: > My main question is still if withdrawalsRoot is actually used in the node to do anything?
I think it is actually used by the client to perform minting Ether out of thin air. @davidtaikocha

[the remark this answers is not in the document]
davidtaikocha: > > My main question is still if withdrawalsRoot is actually used in the node to do anything?
> 
> I think it is actually used by the client to perform minting Ether out of thin air. @davidtaikocha

right, its used to mint real L2 ETHs, if there are some `depositsProcessed` in the corresponding `blockMetaData`

packages/protocol/docs/ZKReq.md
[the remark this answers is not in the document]
Brechtpd: If the purpose of the value is to just let the circuits know if a tx list is "cached" or not, then yes it should be removed. That information is not useful for the circuits.

But if the meaning of the variable is to decode the data from the tx list or blob data then that would be useful to put in the circuit.

packages/protocol/docs/how_taiko_proves_blocks.md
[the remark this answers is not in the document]
Brechtpd: So the actual data is stored as a simple list in `depositsProcessed` in `blockMetaData`? So the node does not use the withdrawals MPT data directly then?

I'm just trying to figure out what it's actually used for (except making it efficient to check what withdrawals were done after the block has already been executed, so not important for the execution itself). The transactions in a block are also made available in the tx tree, but I don't think the node actually ever uses that tree (I think it's specifically made to just put in the block header).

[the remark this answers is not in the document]
davidtaikocha: Yes, the actual data is stored in protocol, when mining a new block, the node need to create a ETH2 withdrawals list based on the corresponding `depositsProcessed` and process them, after that the node will use the created `ETH2 withdrawals list` to calculate a `block.withdrawalsRoot`.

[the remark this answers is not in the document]
Brechtpd: Okay thanks! So I would think node changes can be kept to a minimum then because we do indeed only have to do the `depositsProcessed` -> `ETH2 withdrawals list` translation somewhere and then everything else can remain the same for the node as far as I understand.

Then I would also think that we have a lot of freedom for the value we put in the block header for `withdrawalsRoot`.

Ideally of course it's the standard MPT root value, and I would say let's jet set it to that value for now to minimize differences.

Forcing it to the empty MPT root for each block seems also okay for me, the only thing we break by doing that is that it doesn't allow using Merkle proofs anymore to check which withdrawal was done in a block. So if we want to save the work building this MPT tree in the circuit that sounds fine to me. This also requires almost no node changes.

packages/protocol/docs/ZKReq.md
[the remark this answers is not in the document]
dantaik: Please see this PR: https://github.com/taikoxyz/taiko-mono/pull/13793/files

davidtaikocha (dismissed):

packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -76,6 +77,7 @@ library LibUtils {
         return _ma > 0 ? _ma : maValue;
     }
 
+    /// @dev Hasing the block metadata, ignoring the `cacheTxListInfo` field.
```
davidtaikocha, line 80:
```suggestion
    /// @dev Hashing the block metadata, ignoring the `cacheTxListInfo` field.
```

adaki2004 (dismissed):

Brechtpd (dismissed):

davidtaikocha (approved):

Brechtpd (approved):

PR #13775, opened 2023-05-17
feat(branding): tailwindcss + daisyui

jscriptcoder:
Tailwind + DaisyUI configuration that follows our design guidelines. This repo could potentially contain a showcase and UI components that could be shared across multiple packages and products


.github/workflows/ui.yml
```diff
@@ -0,0 +1,25 @@
...
+      - "packages/ui/**"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Cancel previous runs
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
+          access_token: ${{ github.token }}
```
cyberhorsey, line 19: This CICD file does nothing? It just checks out the repo, there is no tests or anything, I don't think it's needed

jscriptcoder: There is nothing yet to test. I just got it ready... but I can remove it if you want

cyberhorsey: It will add one minute to our CICD times with no benefit right now, but I don't think that's the end of the world.

d1onys1us:
instead of a new repo would it make sense to just put this under `/branding`? like in some folder: `/branding/tailwind/` with a README.md describing how to utilize the config? and out of curiosity what kind of re-usable components do you imagine we'd have? if we are using daisyUI isn't that already the component library and all you need is the tailwind config file? if we were using flowbite i could see it making more sense.

.github/workflows/lint-pr.yml
```diff
@@ -29,4 +29,5 @@ jobs:
             tokenomics
             website
             whitepaper
+            ui
```
d1onys1us, line 32:
```suggestion
            ui
            website
            whitepaper
```

.github/workflows/ui.yml
d1onys1us: i'd say we can remove if we don't need it now, or comment it out.

packages/ui/package.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "keywords": [
+    "Ethereum",
+    "ZKP",
+    "zkRollup",
+    "Layer2",
+    "Tailwindcss",
+    "Design",
+    "Guidelines"
+  ],
```
d1onys1us, line 18:
```suggestion
```
reason for removing keywords: it is a private package so we don't need any SEO keywords here
reason for removing description: duplication of README.md

also would we need a README.md if we decide to use this package instead of `/branding`

jscriptcoder:
Sounds good, `branding` is a good name. I'll change that... Not sure I understood when you say "instead of a new repo", you mean there is no need for a package?, but instead a simple `branding/tailwind` folder under mono-repo root? Ideally this configuration should be our base one, and each product could extend it. It should have its own version and dependencies, so I see it as a package.

Regarding having re-usable components, this is just an idea. We might want to have some base components, which styling is based on props rather than classes, but the downside of that is that they'll be coupled to a framework such as Svelte. E.g. `<Button type="primary" size="small" />` instead of `<button class="btn btn-primary btn-sm" />`. This is more declarative... but again, since they would be tight to the framework, this might not be something we really want to do.

jscriptcoder:
Oh, I just realized there is already a package called `branding`. Ok, in that case I can add the configuration under `Tailwind` folder. What do you think @d1onys1us ?

packages/branding/Tailwind/tailwind.config.js
```diff
@@ -0,0 +1,32 @@
...
+
+    // https://daisyui.com/docs/themes/
+    themes: [
+      {
+        dark: {},
+        light: {},
+      },
+    ],
+  },
+};
```
d1onys1us, line 32: can we name the folder `tailwind` (lowercase) instead of `Tailwind`?

packages/branding/package.json
```diff
@@ -1,5 +1,11 @@
...
   "name": "@taiko/branding",
   "version": "0.1.1",
-  "private": true
+  "private": true,
+  "devDependencies": {
+    "tailwindcss": "^3.3.2"
+  },
+  "dependencies": {
+    "daisyui": "1.16.6"
+  }
```
d1onys1us, line 10: do we need these now? if it's just for later, i feel like we can not include it now and include it if/when we need it. so just the `tailwind.config.js` could be sufficient to keep track of now, and we will update that `tailwind.config.js` with the correct configuration as per the new design and use it where we need it.

but if we do include it, `daisyUI` can be moved to`devDependencies`

d1onys1us: bump @jscriptcoder

jscriptcoder: bumped... regarding whether we need or not the dependencies, at the moment we don't. DaisyUI should be though a peerDependency, since it's getting imported in tailwind.config.js, and if we don't want the IDE to complain we should add it... I'll remove them

dantaik:
@jscriptcoder do you think this PR is still needed? It's been here for quite some time.

jscriptcoder:
Closing it as we can base the configuration on the work done in the Bridge UI v2

PR #13780, opened 2023-05-18
fix(protocol): rename treasure to treasury

dantaik:


packages/protocol/script/DeployOnL1.s.sol
```diff
@@ -38,7 +38,7 @@ contract DeployOnL1 is Script {
 
     address public sharedSignalService = vm.envAddress("SHARED_SIGNAL_SERVICE");
 
-    address public treasure = vm.envAddress("TREASURE");
+    address public treasury = vm.envAddress("TREASURE");
```
davidtaikocha, line 41:
```suggestion
    address public treasury = vm.envAddress("TREASURY");
```

davidtaikocha: and after that, seems we also need to change the env var name in `test_deploy_on_l1.sh`

davidtaikocha (approved):

cyberhorsey (approved):

PR #13781, opened 2023-05-18
fix(protocol): Replace LibEthDeposit assembly

adaki2004:
David asked me today to help investigate why the `depositsRoot` he calculates (seemingly from same test data) within the golang code does not match the one with the solidity calculates.

After investigation / debugging, the assembly code turned out to be the issue, because the underlying array of struct `EthDeposit` within a struct `State` is not stored on a continuous memory location but with the help of pointers rather. 

(If it would have been stored on a continuous location, the assembly would have been wrong anyways, because the first 32 bytes is reserved for the length of the array which should have been chopped).

(If anyone wants to debug how the continuous memory area of the 33-byte-long depossitsProcessed variable look like feel free to put this into `LibEthDepositing.sol` and debug print the value).

```
bytes memory result = new bytes(33 * 32); // allocate 33*32 bytes
assembly {
            for {let i := 0} lt(i, 34) {i := add(i, 1)} {
                let memLocation := add(depositsProcessed, mul(i, 32)) // calculate the memory location
                let value := mload(memLocation) // load the data from memory which is the memory location of the address (but not the deposit amount)
                mstore(add(result, add(32, mul(i, 32))), value) // store the data in the result variable
             }
        }
```

@davidtaikocha :
To help you visualizing the encoding this is how the test data looks (in `test_deposit_hash_creation` test):

0xa9bcf99f5eb19277f48b71f9b14f5960aea58a89000000000de0b6b3a7640000200708d76eb1b69761c23821809d53f65049939e000000001bc16d674ec80000300c9b60e19634e12fc6d68b7fea7bfb26c2e4190000000029a2241af62c0000400147c0eb43d8d71b2b03037bb7b31f8f78ef5f000000003782dace9d90000050081b12838240b1ba02b3177153bca678a86078000000004563918244f40000430c9b60e19634e12fc6d68b7fea7bfb26c2e4190000000053444835ec580000520147c0eb43d8d71b2b03037bb7b31f8f78ef5f000000006124fee993bc000061081b12838240b1ba02b3177153bca678a86078000000006f05b59d3b200000

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibEthDepositing.sol
+++ b/packages/protocol/contracts/L1/libs/LibEthDepositing.sol
@@ -101,8 +101,8 @@ library LibEthDepositing {
             }
         }
 
-        if (depositsProcessed[0].recipient == address(0)) {
-            depositsProcessed = new TaikoData.EthDeposit[](0);
+        assembly {
+            mstore(depositsProcessed, j)
         }
         depositsRoot = hashDeposits(depositsProcessed);
     }
--- a/packages/protocol/contracts/L1/libs/LibEthDepositing.sol
+++ b/packages/protocol/contracts/L1/libs/LibEthDepositing.sol
@@ -101,10 +101,19 @@ library LibEthDepositing {
             }
         }
 
-        // resize the length of depositsProcessed to j.
-        assembly {
-            mstore(depositsProcessed, j)
-            depositsRoot := keccak256(depositsProcessed, mul(j, 32))
+        if (depositsProcessed[0].recipient == address(0)) {
+            depositsProcessed = new TaikoData.EthDeposit[](0);
         }
+        depositsRoot = hashDeposits(depositsProcessed);
+    }
+
+    function hashDeposits(TaikoData.EthDeposit[] memory deposits) internal pure returns (bytes32) {
+        bytes memory buffer;
+
+        for (uint256 i = 0; i < deposits.length; i++) {
+            buffer = abi.encodePacked(buffer, deposits[i].recipient, deposits[i].amount);
+        }
+
+        return keccak256(buffer);
     }
 }
--- a/packages/protocol/test/TaikoL1.t.sol
+++ b/packages/protocol/test/TaikoL1.t.sol
@@ -247,4 +247,49 @@ contract TaikoL1Test is TaikoL1TestBase {
         // Not found
         assertEq(bytes32(0), L1.getCrossChainSignalRoot((iterationCnt + 1)));
     }
+
+    function test_deposit_hash_creation() external {
+        uint96 minAmount = conf.minEthDepositAmount;
+        uint96 maxAmount = conf.maxEthDepositAmount;
+
+        // We need 8 depostis otherwise we are not processing them !
+        depositTaikoToken(Alice, 1e6 * 1e8, maxAmount + 1 ether);
+        depositTaikoToken(Bob, 0, maxAmount + 1 ether);
+        depositTaikoToken(Carol, 0, maxAmount + 1 ether);
+        depositTaikoToken(Dave, 0, maxAmount + 1 ether);
+        depositTaikoToken(Eve, 0, maxAmount + 1 ether);
+        depositTaikoToken(Frank, 0, maxAmount + 1 ether);
+        depositTaikoToken(George, 0, maxAmount + 1 ether);
+        depositTaikoToken(Hilbert, 0, maxAmount + 1 ether);
+
+        // So after this point we have 8 deposits
+        vm.prank(Alice, Alice);
+        L1.depositEtherToL2{value: 1 ether}();
+        vm.prank(Bob, Bob);
+        L1.depositEtherToL2{value: 2 ether}();
+        vm.prank(Carol, Carol);
+        L1.depositEtherToL2{value: 3 ether}();
+        vm.prank(Dave, Dave);
+        L1.depositEtherToL2{value: 4 ether}();
+        vm.prank(Eve, Eve);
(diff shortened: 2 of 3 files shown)


Brechtpd (commented):
> After investigation / debugging, the assembly code turned out to be the issue, because the underlying array of struct EthDeposit within a struct State is not stored on a continuous memory location but with the help of pointers rather.

Huh I think it should be stored as a continuous chunk of memory! Or at least I remember it always being like that. But yeah, `EthDeposit` is only stored compactly in storage, never in memory so that's indeed something that needed to be taken into account with the previous solution (64 bytes in memory vs 32 bytes in storage). Also the off by one error on `j`.

This is the layout in memory I would have expected (in continuous chunk):
`| length (32 bytes) | recipient_0 (32 bytes) | amount_0 (32 bytes) | recipient_1 (32 bytes) | amount_1 (32 bytes) | ....` 

> So in case there is at least 8 deposits to be processed (in our case it's indeed 8 person's deposits), the length of this array is 32+1, encoded as you could see above, 1 struct element = 32 bytes and continuously packed.

Length of the array is 8+1?

Brechtpd (approved):

Brechtpd:
> [Solidity](https://docs.soliditylang.org/en/v0.8.19/types.html) `Note how in all the functions, a struct type is assigned to a local variable with data location storage. This does not copy the struct but only stores a reference`

This seems to be talking about storage, not memory?

cyberhorsey:
I am not sure this is producing the correct hash?
https://emn178.github.io/online-tools/keccak_256.html
paste in the data:
```ts
a9bcf99f5eb19277f48b71f9b14f5960aea58a89000000000de0b6b3a7640000200708d76eb1b69761c23821809d53f65049939e000000001bc16d674ec80000300c9b60e19634e12fc6d68b7fea7bfb26c2e4190000000029a2241af62c0000400147c0eb43d8d71b2b03037bb7b31f8f78ef5f000000003782dace9d90000050081b12838240b1ba02b3177153bca678a86078000000004563918244f40000430c9b60e19634e12fc6d68b7fea7bfb26c2e4190000000053444835ec580000520147c0eb43d8d71b2b03037bb7b31f8f78ef5f000000006124fee993bc000061081b12838240b1ba02b3177153bca678a86078000000006f05b59d3b200000
```
and select "hex", to receive hash:
`8117066d69ff650d78f0d7383a10cc802c2b8c0eedd932d70994252e2438c636`

which is the same as this Go code produces:
```go
func CalcWithdrawalsRootTaiko(withdrawals []*Withdrawal) common.Hash {
	// only process withdrawals/deposits of 8 minimum
	if len(withdrawals) < 8 {
		return EmptyWithdrawalsHash
	}

	var b []byte

	for _, withdrawal := range withdrawals {
		// uint96 solidity type needs us to make 12 length byte slice and put
		// the uint64 into the last 8 bytes. solidity is also bigendian by
		// default so we need to be specific.
		amountBytes := make([]byte, 12)
		binary.BigEndian.PutUint64(amountBytes[4:], withdrawal.Amount)

		b = bytes.Join([][]byte{b, withdrawal.Address.Bytes(), amountBytes}, nil)
	}

	return crypto.Keccak256Hash(b)
}
```

output:
actual  : 0x8117066d69ff650d78f0d7383a10cc802c2b8c0eedd932d70994252e2438c636
```

I'm not too well versed in the under the hood mechanism of solidity's keccak256, but is it hashing it as a non-hex string, and non-text, and thats the issue?

@Brechtpd you probably know the answer here : ).

**Comment by cyberhorsey** [comment]
corresponding taiko-geth PR:
https://github.com/taikoxyz/taiko-geth/pull/76

**Review by davidtaikocha — APPROVED** [review]

File: packages/protocol/contracts/L1/libs/LibEthDepositing.sol [hunk 1]
```diff
@@ -101,10 +101,19 @@ library LibEthDepositing {
             }
         }
 
-        // resize the length of depositsProcessed to j.
         assembly {
             mstore(depositsProcessed, j)
-            depositsRoot := keccak256(depositsProcessed, mul(j, 32))
         }
+        depositsRoot = hashDeposits(depositsProcessed);
```
dantaik, line 107: The previous hashing of depositsProcessed is buggy, but I think it can get fixed simply using:
```solidity
 depositsRoot := keccak256(add(depositsProcessed, 32), mul(j, 32))
```

dantaik: I think maybe the expected value in the test (`0x8117066d69ff650d78f0d7383a10cc802c2b8c0eedd932d70994252e2438c636`) is also calculated wrong. How is it calculated exactly, better to calculate the hash inside the test.

cyberhorsey: There is a PR for the looping being done in assembly and keccak256 being done outside here Dani: https://github.com/taikoxyz/taiko-mono/pull/13782/files

Much less readable but saves some gas

Brechtpd:
Seems like the memory array does indeed store relative offsets to the struct locations! Looks like I remembered wrongly or I just never did this with structs which are different (or solidity now does it differently, I remember them not wanting people to depend on how things are stored in memory because it could change).

I don't think anything storage is related to what we do here because we copy the basic types over to the memory array, which are actual copies.

We could simply skip over the offsets because solidity should always just put the array elements in a continuous chunk after the relative offsets (which is why I don't really understand why it's done this way for memory, maybe it helps minimize the difference with abi encoded calldata). But for what we do here that doesn't solve the issue completely anyway because we want to pack the data tighter than it is stored in memory.

PR #13782, opened 2023-05-19
chore(protocol): optimize _hashEthDeposits

dong77:
Reduce `proposeblock` gas cost from 97908 to 94780

It turns out for a dynamic array, each element's location is stored in side a continuous region in memory, but the actual elements are stored somewhere else.


davidtaikocha (dismissed):
> It turns out for a dynamic array, each element's location is stored in side a continuous region in memory, but the actual elements are stored somewhere else.

thats also what i thought, the bytes coming after the `length` pointer are not the actual struct data bytes, but seems like the slot keys of those slots which save the actual struct data

davidtaikocha (approved):

adaki2004 (approved):

Issue #13791, opened 2023-05-22
feat: Separate Tokenomics from Core Protocol to a Controller Contract


dantaik:
Currently, tokenomics is tightly integrated with the core protocol. I suggest decoupling the tokenomics code into a distinct contract. This division would:

1. Streamline the core protocol, improving testability.
2. Promote codebase reusability, enabling others to employ it independently of our tokenomics.
3. Simplify tokenomics testing by eliminating the need for proposing/proving actual blocks.
4. Allow concurrent tokenomics deployment across different L3 testnets without altering the core protocol.
5. Facilitate safer upgradability, leaving the core protocol undisturbed when tokenomics updates are necessary.

This decoupling, however, might inflate gas costs.

Certain implementation precautions are necessary:

1. The tokenomics contract should not hold custody of user-deposited TaikoTokens; this role should remain with the core protocol to prevent token migration during a tokenomics switch or we should simply avoid user-deposited TaikoTokens (see below). 
2. To minimize gas costs, a single call to the tokenomics contract should be implemented when multiple blocks are verified.
3. If **_tokenomics_** lookup resolves to '0x0', it signifies the absence of tokenomics (the core protocol shall still work as a permissionless rollup)

It may be beneficial to rename 'tokenomics' to a more representative term like **_controller_**. This contract could whitelist proposer/prover addresses without token or tokenomics involvement. For example, a company could create a simple **_controller_** contract for deploying a centralized, permissioned Taiko rollup without a token.

Considering the core protocol may not require a token, we might want to remove `taikoTokenBalances` and associated functions from the core protocol. The controller contract could then undertake token minting/burning operations if tokens are introduced.

Brechtpd:
With tokenomics you mean the proposer selection/prover selection/fee system right? (I think that goes well above what people would normally understand under tokenomics so just making sure).

Yes I think we should do this, same with the verifier contracts this will otherwise get too messy if we want to make this more flexible/modular.

> This decoupling, however, might inflate gas costs.
> To minimize gas costs, a single call to the tokenomics contract should be implemented when multiple blocks are verified.

If it makes things significantly simpler I think it should be fine to call for each block. With how current gas costs are set up to make warm storage reads/writes very cheap compared to cold ones (even true for calls), the gas savings probably won't be that significant while the complexity of batching things could be high (and also costs gas).

[...]

PR #13794, opened 2023-05-22
feat(protocol): Separate proof verification to a standalone, deployed contract

adaki2004:
Base branch of this introduced multi-prover possibilities. 
This one outsourced proof verification into a separate, deployed contract.

Gas comparison:

Base: (42.074 gwei)
kép

This one (47.418 gwei)
kép

**IMPORTANT NOTE**

Code to calculate the public _`instance`_  below stayed in LibProving.
```
// Put together the input for proof and signature verification
        uint256[9] memory inputs;
...

        bytes32 instance;
        assembly {
            instance := keccak256(inputs, mul(32, 9))
        }
```

I decided to leave it there, because:
1. tested a version where i pass the `Evidence` towards the verifier contract and calculate it there, but it adds additional data (overhead) which costs roughly +1K gas. 
2. In case there is an additional protocol upgrade which changes the `Evidence` we need to upgrade this new Verifier anyways as well..
 
Only issue I see with this is: there will be a completely new proof method which will require only some portion of the `Evidence` so would be useful having that extra data at the verifier contract.

With current plans (ZK, SGX and Failsafe signature overwrite) i see it a no issue, but who knows, maybe worth to pass the whole `Evidence` data ? 
Please let me know what you think @dantaik @Brechtpd

### DIFF
--- a/packages/protocol/contracts/L1/IProofVerifier.sol
+++ b/packages/protocol/contracts/L1/IProofVerifier.sol
@@ -17,7 +17,7 @@ interface IProofVerifier {
      * @param blockProofs Proof array
      * @param resolver Current (up-to-date) address resolver
      */
-    function verifyProof(
+    function verifyProofs(
         bytes32 instance, 
         TaikoData.TypedProof[] calldata blockProofs,
         AddressResolver resolver
--- a/packages/protocol/contracts/L1/ITaikoProofVerifier.sol
+++ b/packages/protocol/contracts/L1/ITaikoProofVerifier.sol
@@ -19,7 +19,7 @@ interface IProofVerifier {
      */
     function verifyProof(
         bytes32 instance, 
-        TaikoData.TypedProof[] memory blockProofs,
+        TaikoData.TypedProof[] calldata blockProofs,
         AddressResolver resolver
     ) external;
 }
--- a/packages/protocol/contracts/L1/ITaikoProofVerifier.sol
+++ b/packages/protocol/contracts/L1/ITaikoProofVerifier.sol
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+//  _____     _ _         _         _
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+
+pragma solidity ^0.8.18;
+
+import {TaikoData} from "./TaikoData.sol";
+import {AddressResolver} from "../common/AddressResolver.sol";
+
+interface IProofVerifier {
+    /**
+     * Verifying proof via the ProofVerifier contract
+     *
+     * @param instance Hashed public input
+     * @param blockProofs Proof array
+     * @param resolver Current (up-to-date) address resolver
+     */
+    function verifyProof(
+        bytes32 instance, 
+        TaikoData.TypedProof[] memory blockProofs,
+        AddressResolver resolver
+    ) external;
+}
--- a/packages/protocol/contracts/L1/ProofVerifier.sol
+++ b/packages/protocol/contracts/L1/ProofVerifier.sol
@@ -27,14 +27,14 @@ library TaikoProofToggleMask {
 }
 
 /// @custom:security-contact hello@taiko.xyz
-contract TaikoProofVerifier is TaikoErrors {
+contract ProofVerifier is TaikoErrors {
     uint256[50] private __gap;
 
     function getToggleMask() public pure virtual returns (uint16) {
         return TaikoProofToggleMask.getToggleMask();
     }
 
-    function verifyProof(
+    function verifyProofs(
         bytes32 instance, 
         TaikoData.TypedProof[] calldata blockProofs,
         AddressResolver resolver
@@ -94,4 +94,4 @@ contract TaikoProofVerifier is TaikoErrors {
     }
 }
 
-contract ProxiedTaikoProofVerifier is Proxied, TaikoProofVerifier {}
+contract ProxiedProofVerifier is Proxied, ProofVerifier {}
--- a/packages/protocol/contracts/L1/TaikoProofVerifier.sol
+++ b/packages/protocol/contracts/L1/TaikoProofVerifier.sol
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: MIT
+//  _____     _ _         _         _
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+
+pragma solidity ^0.8.18;
+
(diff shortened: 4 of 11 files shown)


packages/protocol/contracts/L1/TaikoProofVerifier.sol
```diff
@@ -0,0 +1,97 @@
...
+    uint256[50] private __gap;
+
+    function getToggleMask() public pure virtual returns (uint16) {
+        return TaikoProofToggleMask.getToggleMask();
+    }
+
+    function verifyProof(
+        bytes32 instance, 
+        TaikoData.TypedProof[] calldata blockProofs,
+        AddressResolver resolver
```
dantaik, line 40: We should not pass in the resolver as this contract shall be a resolver itself.

packages/protocol/contracts/L1/ITaikoProofVerifier.sol
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Verifying proof via the ProofVerifier contract
+     *
+     * @param instance Hashed public input
+     * @param blockProofs Proof array
+     * @param resolver Current (up-to-date) address resolver
+     */
+    function verifyProof(
+        bytes32 instance, 
+        TaikoData.TypedProof[] calldata blockProofs,
```
dantaik, line 22: What if you change `calldata` to `memory`? Will the cost be lower?

packages/protocol/contracts/L1/ITaikoProofVerifier.sol
```diff
@@ -0,0 +1,25 @@
...
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+
+pragma solidity ^0.8.18;
+
+import {TaikoData} from "./TaikoData.sol";
+import {AddressResolver} from "../common/AddressResolver.sol";
+
+interface IProofVerifier {
```
dantaik, line 12: rename the file to IProofVerifier.sol

packages/protocol/contracts/L1/TaikoProofVerifier.sol
```diff
@@ -0,0 +1,97 @@
...
+        // RESERVED_X_ONLY,     // 0000 0100
+        // RESERVED_Y_ONLY,     // 0000 1000
+        // ZKP_AND_SGX,         // 0000 0011
+        // X_ZKP_SGX,           // 0000 0111
+        return uint16(1); // ZKP ONLY by default
+    }
+}
+
+/// @custom:security-contact hello@taiko.xyz
+contract TaikoProofVerifier is TaikoErrors {
```
dantaik, line 30: Rename to `ProofVerifier`? (and the filename)

packages/protocol/test/TaikoL1MultiProving.t.sol
```diff
@@ -13,6 +13,7 @@ import {TaikoToken} from "../contracts/L1/TaikoToken.sol";
 import {SignalService} from "../contracts/signal/SignalService.sol";
 import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
 import {TaikoL1TestBase} from "./TaikoL1TestBase.t.sol";
+import {TaikoProofToggleMask, TaikoProofVerifier} from "../contracts/L1/TaikoProofVerifier.sol";
```
dantaik, line 16: TaikoProofToggleMask not used at all.

packages/protocol/contracts/L1/TaikoProofVerifier.sol
```diff
@@ -0,0 +1,97 @@
...
+pragma solidity ^0.8.18;
+
+import {AddressResolver} from "../common/AddressResolver.sol";
+import {Proxied} from "../common/Proxied.sol";
+import {TaikoErrors} from "./TaikoErrors.sol";
+import {TaikoData} from "./TaikoData.sol";
+import {LibVerifyTrusted} from "./libs/proofTypes/LibVerifyTrusted.sol";
+import {LibVerifyZKP} from "./libs/proofTypes/LibVerifyZKP.sol";
+
+library TaikoProofToggleMask {
```
dantaik, line 16: Why do we need the code in a library?

packages/protocol/contracts/L1/ITaikoProofVerifier.sol
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Verifying proof via the ProofVerifier contract
+     *
+     * @param instance Hashed public input
+     * @param blockProofs Proof array
+     * @param resolver Current (up-to-date) address resolver
+     */
+    function verifyProof(
```
dantaik, line 20: rename to `verifyProofs`?

dantaik (changes requested):

packages/protocol/contracts/L1/TaikoProofVerifier.sol
```diff
@@ -0,0 +1,97 @@
...
+            verifyTypedProof(proof, instance, resolver);
+            mask &= ~bitMask;
+
+            unchecked {
+                ++i;
+            }
+        }
+
+        if(mask != 0) {
+            revert L1_NOT_ALL_REQ_PROOF_VERIFIED();
```
dantaik, line 66: @adaki2004 Lets talk about it tomorrow. But i feel like this current code can only express AND logic, not the OR logic.

packages/protocol/contracts/L1/IProofVerifier.sol
```diff
@@ -0,0 +1,25 @@
...
+     *
+     * @param instance Hashed public input
+     * @param blockProofs Proof array
+     * @param resolver Current (up-to-date) address resolver
+     */
+    function verifyProofs(
+        bytes32 instance, 
+        TaikoData.TypedProof[] calldata blockProofs,
+        AddressResolver resolver
+    ) external;
```
Brechtpd, line 24:
```suggestion
    function verifyProof(bytes calldata proof) external;
```

What do you think of this API? This way we don't have any dependencies and the proof can really be whatever, and proofs don't have to confirm to `TypedProof` which may not always fit.

packages/protocol/contracts/L1/ProofVerifier.sol
```diff
@@ -0,0 +1,97 @@
...
+
+    function verifyProofs(
+        bytes32 instance, 
+        TaikoData.TypedProof[] calldata blockProofs,
+        AddressResolver resolver
+    )
+    external
+    view
+    {
+        uint16 mask = getToggleMask();
```
Brechtpd, line 45: Maybe we should simply remove the mask? A verifier just always hard-codes the logic that needs to be followed. If this logic changes, a different verifier contract should be deployed.

Well, it is maybe useful for testing still, so this could be the testnet verifier contract or something like that. But on mainnet, probably just want to keep it as simple as possible.

dantaik:
@adaki2004 , as mentioned https://github.com/taikoxyz/taiko-mono/issues/13791#issuecomment-1558411729, I now perfer not splitting verification logics into their own contracts.

PR #13796, opened 2023-05-22
feat(protocol): document skipping withdrawalsRoot check in circuits and client

### LINKED ISSUE #13795: feat: skip checking withdrawalsRoot in circuits and client
As of now,  the `withdrawalsRoot` in Taiko blockheader is the hash of a list of EthDeposits from L1 to L2. In both circuits and client, we check this value is correctly calculated and matches the actual EthDepoisits.

To simpify the code is to enforce `withdrawalsRoot` to have the value of the **empty MPT withdrawals root** (not `0x0`) . In the circuits, verify the hash of the list of EthDeposits actually match the value used by the protocol to calculate the right metahash.

Screenshot 2023-05-22 at 22 41 52

dantaik:
See https://github.com/taikoxyz/taiko-mono/issues/13795


packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -278,16 +275,16 @@ h_timestamp(timestamp)
...
-h_ommers_hash(ommersHash)
+h_ommers_hash("ommersHash = keccek([])")
 h_state_root(stateRoot)
 h_transactions_root(transactionsRoot)
 h_receipts_root(receiptsRoot)
-h_logs_bloom(logsBloom)
-h_difficulty(difficulty)
-h_extra_data(extraData)
-h_nonce(nonce)
+h_logs_bloom("logsBloom = []")
```
davidtaikocha, line 282: not sure the motivation of make `header.logsBloom` to equal to `[]`?

packages/protocol/docs/how_taiko_proves_blocks.md
```diff
@@ -278,16 +275,16 @@ h_timestamp(timestamp)
 h_mix_hash(mixHash)
 h_beneficiary(beneficiary)
 h_parent_hash(parentHash)
-h_ommers_hash(ommersHash)
+h_ommers_hash("ommersHash = keccek([])")
```
Brechtpd, line 278:
```suggestion
h_ommers_hash("ommersHash = keccak([])")
```

packages/protocol/docs/how_taiko_proves_blocks.md
dantaik: Is it correct to say: "`logsBloom` must be a `bytes32[8]` with all zeros"?

Brechtpd: It seems like the common representation is `bytes[256]` where I look, but yes, 256 `0` bytes.

PR #13799, opened 2023-05-22
feat(protocol): Add setter to IAddressManager of AddressResolver

### LINKED ISSUE #13798: AddressRresolver's IAddressManager setter
As per https://discord.com/channels/984015101017346058/994824826550231183/1110248017434509352

adaki2004:
Adding a setter to IAddressManager.

Because contracts connected to AddressManager (be it static or not) via an interface, a setter is necessary.

See issue : https://github.com/taikoxyz/taiko-mono/issues/13798

Also I feel there is a need for the setter, because both @dantaik and @Brechtpd indicated, that each contract (i.e. ProofVerifier, TaikoL1, Bridge) shall have it's own AddressResolver (= IAddressManager interface) - tho currently the TaikoL1, Bridge, TaikoToken sharing the same AddressManager (as we can see within the tests and also within the deployment script).

There could be situations (as described in discord convo) which implies security issues for not having a setter OR we indeed need separate instances for those contracts. For an example like: TaikoL1 needs BridgeV2, although TaikoNFTBridge still needs BridgeV1, therefore TaikoL1 and TaikoNFTBridge cannot share the same.

### DIFF
--- a/packages/protocol/contracts/common/AddressResolver.sol
+++ b/packages/protocol/contracts/common/AddressResolver.sol
@@ -29,7 +29,7 @@ abstract contract AddressResolver {
         _;
     }
 
-    event IAddressManagerChanged(address changedOn, address newAddress);
+    event AddressManagerChanged(address addressManager);
 
     /**
      * Resolves a name to an address on the current chain.
--- a/packages/protocol/contracts/common/AddressResolver.sol
+++ b/packages/protocol/contracts/common/AddressResolver.sol
@@ -29,6 +29,8 @@ abstract contract AddressResolver {
         _;
     }
 
+    event IAddressManagerChanged(address changedOn, address newAddress);
+
     /**
      * Resolves a name to an address on the current chain.
      *
--- a/packages/protocol/contracts/common/EssentialContract.sol
+++ b/packages/protocol/contracts/common/EssentialContract.sol
@@ -6,6 +6,7 @@
 
 pragma solidity ^0.8.18;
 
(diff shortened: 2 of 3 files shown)


Brechtpd (commented):
I think this makes sense in general.

dantaik:
> AddressResolver



> Adding a setter to IAddressManager.
> 
> Because contracts connected to AddressManager (be it static or not) via an interface, a setter is necessary.
> 
> See issue : #13798
> 
> Also I feel there is a need for the setter, because both @dantaik and @Brechtpd indicated, that each contract (i.e. ProofVerifier, TaikoL1, Bridge) shall have it's own AddressResolver (= IAddressManager interface) - tho currently the TaikoL1, Bridge, TaikoToken sharing the same AddressManager (as we can see within the tests and also within the deployment script).
> 
> There could be situations (as described in discord convo) which implies security issues for not having a setter OR we indeed need separate instances for those contracts. For an example like: TaikoL1 needs BridgeV2, although TaikoNFTBridge still needs BridgeV1, therefore TaikoL1 and TaikoNFTBridge cannot share the same.

Hmmm. I'm not sure I agree. The idea is that each contract that needs address lookup is an `IAddressResover` itself. These address lookup contracts do not expose any function to *set  addresses* as this is done by the admin interacting directly with the underlying/shared `AddressManager`.

dantaik (approved):

davidtaikocha (approved):

PR #13804, opened 2023-05-23
refactor(protocol): add typed proof

dantaik:
Do not merge


davidtaikocha (approved):

dantaik:
> @dantaik This is the same one we discussed this morning. I also started implementing it on my [branch](https://github.com/taikoxyz/taiko-mono/pull/13745). Let me finish it and we can compare (readability, gas consumption, etc.)

Use mine as a reference. Do not merge any breaking changes until @david makes a alpha-3 release.

Brechtpd:
> If SGX is enabled, are those proofs having AND / OR relation ?
Only thing that makes sense to me is AND

> If SGX is enabled, are they coming separate (txns) or together ?
I would say only together. Having one or the other on its own isn't very useful and complicates things quite a bit (multiple phases, fee payment,...).

I don't really understand the benefits of the approach in this PR compared to the one Dani was working on, so would be great to learn more about the thinking behind it.

Issue #13556, opened 2023-04-06
Remove x/y scaling in EIP1559 calculation


dantaik:
### Describe the feature request

From Brecht:
> The `exp` implementation has as input and output values in 10**18 fixed format notation, so `exp(0) == 10**18`. Let's say that's 1 wei (so plenty of precision). The `exp` function also works with inputs up till `~135`, which will return an enormous number: `log10(exp(135)) == 58`. The total ETH supply is ~100,000,000 ETH, and 10*18 wei/ETH, so `log10(100,000,000*(10**18)) == 26`, much smaller than the max value returned by `exp`.

dantaik:
I don't think it has something to do with value boundaries or accuracy, it's about finding the parameters to shape the curve's slope to ensure that at the initial best-guess price, the 2x/1x basefee ratio is as expected.

I tried but didn't find a way to just use one scaling factor to make it work.

Brechtpd:
I think the y scaling can be avoided (I haven't done the math, so take with a grain of salt). To target `p` you would pick some `ADJUSTMENT_QUOTIENT` (total x scaling is done with `TARGET * ADJUSTMENT_QUOTIENT`). Then to get `gasExcess` to target some initial fee, you would just do the inverse basefee calculation:

```
initial_basefee = 10**8
gasExcess = (TARGET * ADJUSTMENT_QUOTIENT) * math.log(initial_basefee * (TARGET * ADJUSTMENT_QUOTIENT))
```

dantaik:
I suggest that we keep it as-is to reduce risk. Removing y-scale still ends up with the same number of storage slots that we have to write (1 for now), the only small downside is that we have an extra DIV opcode to be optimized away, it's too small of a gain I think.

PR #13483, opened 2023-03-28
test(bridge-ui): add  unit test for checkIfTokenIsDeployedCrossChain util

murrayee:

PR #13806, opened 2023-05-24
chore(protocol): update Alpha-3 protocol configurations

### LINKED ISSUE #13802: Update configs in `TaikoConfig.sol` \ `EIP1559Params` for Alpha 3 testnet
Current config (we will change `chainId` for sure, and what else?):
```solidity
library TaikoConfig {
    function getConfig() internal pure returns (TaikoData.Config memory) {
        return TaikoData.Config({
            chainId: 167,
            // Two weeks if avg block time is 10 seconds
            maxNumProposedBlocks: 120960,
            ringBufferSize: 120960 + 10,
            // Each time one more block is verified, there will be ~20k
            // more gas cost.
            maxVerificationsPerTx: 10,
            // Set it to 6M, since its the upper limit of the Alpha-2
            // testnet's circuits.
            blockMaxGasLimit: 6000000,
            // Set it to 79  (+1 TaikoL2.anchor transaction = 80),
            // and 80 is the upper limit of the Alpha-2 testnet's circuits.
            maxTransactionsPerBlock: 79,
            minEthDepositsPerBlock: 8,
            maxEthDepositsPerBlock: 32,
            maxEthDepositAmount: 10000 ether,
            minEthDepositAmount: 1 ether,
            // Set it to 120KB, since 128KB is the upper size limit
            // of a geth transaction, so using 120KB for the proposed
            // transactions list calldata, 8K for the remaining tx fields.
            maxBytesPerTxList: 120000,
            proofCooldownPeriod: 30 minutes,
            systemProofCooldownPeriod: 15 minutes,
            // Only need 1 real zkp per 10 blocks.
            // If block number is N, then only when N % 10 == 0, the real ZKP
            // is needed. For mainnet, this must be 0 or 1.
            realProofSkipSize: 10,
            ethDepositGas: 21000,
            ethDepositMaxFee: 1 ether / 10,
            txListCacheExpiry: 0,
            adjustmentQuotient: 16,
            relaySignalRoot: false
        });
    }
}
```

And we also need to create a `EIP1559Params` for the testnet:

```solidity
struct EIP1559Params {
        uint64 basefee;
        uint64 gasIssuedPerSecond;
        uint64 gasExcessMax;
        uint64 gasTarget;
        uint64 ratio2x1x;
}
```

davidtaikocha:
Questions to discuss:

- [x] What average block time will we use / assume for A3? (3 seconds?) => 3s seems to be reasonable
- [x] Shall we change `minEthDepositAmount` to 0.1 ETHs?  ([some concerns here](https://github.com/taikoxyz/taiko-mono/issues/13802#issuecomment-1559036054)) => [should be fine](https://github.com/taikoxyz/taiko-mono/issues/13802#issuecomment-1562203757)
- [x] Do we need to increase the cooldown period for A3 network? (`proofCooldownPeriod` + `systemProofCooldownPeriod`) => if Sepolia generally work well and we're confident we can get these overwrite transactions in fast enough then seems ok.
- [x] Do we need to change `realProofSkipSize`? (i think it depends on the avg block time that we predict) => start with realProofSkipSize: 10 for the initial wave, we can reconsider afterwards
- [x] What `treasury` address will we use? => `0xdf09A0afD09a63fb04ab3573922437e1e637dE8b`
- [x] What `INITIAL_PROOF_TIME_TARGET` will we use? (in current internal devnet, avg proof submission time is ~80s, so shall we use `120s`? if we use `120s`, then `initProofTimeIssued` will be `49883`) => `2 * 80 (avg proof submission time) = 160`
- [x] What `param1559.gasIssuedPerSecond` will we use? (i think it depends on the avg block time that we predict) => param1559.gasIssuedPerSecond := 2,000,000 gas/s == 4 L2 blocks of 6M gas/L1 block
- [x] Shall we change `param1559.yscale` / `param1559.xscale` / `param1559.gasExcess`? => lets calculate them in TaikoL2's init() function
- [x] Shall we make sure `state.blockFee` always `>=1`? (to prevent proposing becomes "totally free"?) => If the proposing EIP-1559 parameters are set well, seems this will be unnecessary (could be a big IF)


packages/protocol/contracts/L1/TaikoConfig.sol
```diff
@@ -11,23 +11,23 @@ import {TaikoData} from "../L1/TaikoData.sol";
 library TaikoConfig {
     function getConfig() internal pure returns (TaikoData.Config memory) {
         return TaikoData.Config({
-            chainId: 167,
-            // Two weeks if avg block time is 10 seconds
-            maxNumProposedBlocks: 120960,
-            ringBufferSize: 120960 + 10,
+            chainId: 167005,
+            // Two weeks if avg block time is 3 seconds (is it reasonable?)
```
dantaik, line 15: I think it is reasonable. How is 403200 calculated?

packages/protocol/script/DeployOnL1.s.sol
```diff
@@ -38,7 +38,7 @@ contract DeployOnL1 is Script {
 
     address public sharedSignalService = vm.envAddress("SHARED_SIGNAL_SERVICE");
 
-    address public treasury = vm.envAddress("TREASURY");
+    address public treasury = vm.envAddress("TREASURY"); // Q: which address will we use?
```
dantaik, line 41: You can use this value `0xdf09A0afD09a63fb04ab3573922437e1e637dE8b`, it's one of my hardware wallets.

packages/protocol/script/DeployOnL1.s.sol
```diff
@@ -49,7 +49,7 @@ contract DeployOnL1 is Script {
     // For testnet it could be somewhere 85-100s
     // For mainnet it could be around 1800 s (30mins)
     // Can be adjusted later with setters
-    uint64 public INITIAL_PROOF_TIME_TARGET = uint64(vm.envUint("INITIAL_PROOF_TIME_TARGET"));
+    uint64 public INITIAL_PROOF_TIME_TARGET = uint64(vm.envUint("INITIAL_PROOF_TIME_TARGET")); // Q: INITIAL_PROOF_TIME_TARGET=`120s`? => initProofTimeIssued will be `49883`
```
dantaik, line 52: @smtmfft What's the current average time required per proof? 

@Brechtpd is twice the above value a good start?

packages/protocol/test/genesis/test_config.json
```diff
@@ -80,10 +80,10 @@
     "SignalServiceProxy": "0x1000777700000000000000000000000000000007"
   },
   "param1559": {
-    "yscale": "358298803609133338137582400989",
-    "xscale": "1488514844",
-    "gasIssuedPerSecond": "12500000",
-    "gasExcess": "45450000000"
+    "yscale": "358298803609133338137582400989", // Q: do we need to change this?
```
dantaik, line 83: `xscale`, `yscale`, and `gasExcess` are all calculated in _TaikoL2_'s `init()` function, I don't think we need to provide that value here, do we?

packages/protocol/test/genesis/test_config.json
```diff
@@ -80,10 +80,10 @@
...
+    "xscale": "1488514844", // Q: do we need to change this?
+    "gasIssuedPerSecond": "2000000", // If we assume the avg block time is 3s (is it reasonable?) , and each block is full (block.gasLimit=6m)
+    "gasExcess": "45450000000" // Q: do we need to change this?
```
dantaik, line 86: I think we need `gasExcessMax` here as it is part of the params:

```solidity

    struct EIP1559Params {
        uint64 basefee;
        uint64 gasIssuedPerSecond;
        uint64 gasExcessMax;
        uint64 gasTarget;
        uint64 ratio2x1x;
    }

```

dantaik: In general, we should/can use [TestTaiko1559Params.t.sol](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/test/Taiko1559Params.t.sol)  to tune the `gasExcessMax` and other parameters.

packages/protocol/contracts/L1/TaikoConfig.sol
davidtaikocha: its calculated  by `20 * 60 * 24 * 14` (avg block time is 3 seconds => 20 blocks per minute).

Brechtpd (commented):
Because each proof has the exact same cost currently, I think we should make it so the prover fee cannot change quickly (tweak `adjustmentQuotient`).

>  What average block time will we use / assume for A3? (3 seconds?)

Very important this matches the proposing EIP-1559 target!

> Do we need to increase the cooldown period for A3 network? (

I don't know how stable Sepolia is, but if things generally work well and we're confident we can get these overwrite transactions in fast enough then seems ok to me.

> What param1559.gasIssuedPerSecond will we use? (i think it depends on the avg block time that we predict)

EDIT: I confused this value with `gasExcess`.

> Shall we change param1559.yscale / param1559.xscale / param1559.gasExcess?

See comment on this above.

> Shall we make sure state.blockFee always >=1? (to prevent proposing becomes "totally free"?)

If the proposing EIP-1559 parameters are set well, I don't think this will be necessary (could be a big IF)

Brechtpd:
- Maybe we should do `minEthDepositsPerBlock: 1`? Feels like that could be a big UX improvement.
- I think smart to start with  `realProofSkipSize: 10` for the initial wave, we can reconsider afterwards.
- Regardless of the value of `minEthDepositAmount`, I think there's definitely spam risk on this queue. Anybody with a lot of ETH can make the queue extremely long and we can only process a couple/block, so users may have to wait a long time to get their ETH deposit if abused.
- `INITIAL_PROOF_TIME_TARGET = 80s * 2`
- prover fee `adjustmentQuotient = TBD`
- `param1559.gasIssuedPerSecond := 2,000,000 gas/s` == 4 L2 blocks of 6M gas/L1 block. Seems reasonable to me and is the proposed 3s block times. Other values:
```
struct EIP1559Params {
        basefee = ? (could just be 0 to start);
        gasIssuedPerSecond = 2000000;
        gasExcessMax = ?;
        gasTarget = 2000000 * 12;
        ratio2x1x = 11250;
    }
```
Unfortunately not so easy to find all the correct values here. Something like https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/script/DetermineNewProofTimeIssued.s.sol + some input parameter simplification https://github.com/taikoxyz/taiko-mono/issues/13556 like with the proving fee would certainly help to make this easier to configure!

packages/protocol/test/genesis/test_config.json
davidtaikocha: > I think we need `gasExcessMax` here as it is part of the params:
> 
> ```solidity
>     struct EIP1559Params {
>         uint64 basefee;
>         uint64 gasIssuedPerSecond;
>         uint64 gasExcessMax;
>         uint64 gasTarget;
>         uint64 ratio2x1x;
>     }
> ```

Hmm i think we only these values which will be written into TaikoL2's storage slots here, so here should be those results which are calculated by the `TaikoL2.init(EIP1559Params)` method.

davidtaikocha:
Update the configs based on PR feedbacks: https://github.com/taikoxyz/taiko-mono/pull/13806/commits/1f6eed59db8e90e4554c539c5cb9a727b4453bc3 , plz review

packages/protocol/test/genesis/test_config.json
dantaik: https://github.com/taikoxyz/taiko-mono/pull/13815 We can upgrade the settings by upgrading the contract like this.

Brechtpd (dismissed):
Have we seen the proposal basefee work within reasonable ranges on our internal testnet?

davidtaikocha:
> Have we seen the proposal basefee work within reasonable ranges on our internal testnet?

In current internal devnet, all provers will at least wait `proofTargetTime` before submission, but right now the baseFee is still `0`, and i think @adaki2004 is working on tuning the parameters.

Brechtpd:
Ah yeah I heard about that! 

It looks like how I name things is a bit confusing, what about the L2 basefee that the proposer pays on L2? That one also depends on proposing blocks at fixed intervals (same as with the proofs) to keep the basefee steady. So was wondering if the behavior there was as expected and we can for sure push that L2 basefee pretty high in a small amount of time (we need that high L2 basefee to prevent spam).

dantaik:
@davidtaikocha @Brechtpd @adaki2004 Do you think we should get onto a call to quickly resolve this issue?

davidtaikocha:
> @davidtaikocha @Brechtpd @adaki2004 Do you think we should get onto a call to quickly resolve this issue?

image

after upgrade the proxy implementation and use Dani's new params:
```
ProofTimeTarget: 150
ProofTimeIssued: 162263074
BlockFee: 1e8 ( 1 TKO was set so far so i guess we stick to it.. if you disagree i need to recalculate the proofTimeIssued)
AdjustmentQuotient: 32000
```
seems the fee backs to stable again. and before that, our client sometimes will submit the proofs asap, and zkpool's client also always submits proof asap, i think this is also part of the reason.

davidtaikocha:
> @davidtaikocha @Brechtpd @adaki2004 Do you think we should get onto a call to quickly resolve this issue?

And im ok for a call to the final overall checking of this PR.

Brechtpd (approved):

dantaik (approved):

Mehrabani61:
#13483

Mehrabani61:
> 
Duplicate of #

Mehrabani61:
update-alpha-3-configs

PR #13810, opened 2023-05-24
feat(eventindexer): add stats tracking

cyberhorsey:
this PR now tracks blockVerified events, and uses the proofTime and proofReward arguments to maintain an average, and sets us up to calculate other stats we want in the future as well. It exposes them via a /stats endpoint for the status page.

this also updates the status page with the new ABI.

note: this PR relies on code that has not been deployed to alpha-3, Dani's blockVerified event reward and fee changes. so we cant deploy it until we deploy those updates.


packages/eventindexer/indexer/save_block_proven_event.go
```diff
@@ -47,6 +55,8 @@ func (svc *Service) saveBlockProvenEvent(
 	chainID *big.Int,
 	event *taikol1.TaikoL1BlockProven,
 ) error {
+	fmt.Println("blockProven event found")
```
davidtaikocha, line 58: Shall we use `log.Infof`here?

packages/eventindexer/indexer/save_block_verified_event.go
```diff
@@ -0,0 +1,91 @@
...
+	if !events.Next() || events.Event == nil {
+		log.Infof("no BlockVerified events")
+		return nil
+	}
+
+	for {
+		event := events.Event
+
+		if event.Raw.Removed {
+			continue
```
davidtaikocha, line 28: Do we need to remove the corresponding saved event data in database? or simply ignoring it is fine?

packages/eventindexer/indexer/subscribe.go
```diff
@@ -154,3 +155,57 @@ func (svc *Service) subscribeBlockProposed(ctx context.Context, chainID *big.Int
...
+			return
+		case err := <-sub.Err():
+			errChan <- errors.Wrap(err, "sub.Err()")
+		case event := <-sink:
+			log.Infof("blockVerifiedEvent from subscription")
+
+			if err := svc.saveBlockVerifiedEvent(ctx, chainID, event); err != nil {
+				eventindexer.BlockVerifiedEventsProcessedError.Inc()
+
+				log.Errorf("svc.subscribe, svc.saveBlockProvenEvent: %v", err)
```
davidtaikocha, line 188: shall we add a `continue` here?

packages/eventindexer/indexer/save_block_verified_event.go
cyberhorsey: Yeah, none of the event handling removes from DB yet (same with block proposed and block proven still), no reorg handling. I can do it for all 3 in this PR.

packages/eventindexer/indexer/save_block_proven_event.go
cyberhorsey: Yes

packages/eventindexer/indexer/save_block_verified_event.go
cyberhorsey: actually, I think i will do a separate PR to handle reorg entirely, similiar to the taiko-client one, since it will make this PR quite large and take some time.

davidtaikocha (dismissed):

davidtaikocha (approved):

RogerLamTd (approved):

PR #13812, opened 2023-05-25
add prover concepts

dionysuzx:


packages/website/pages/docs/concepts/prover-dynamics.md
```diff
@@ -0,0 +1,17 @@
+# Prover dynamics
+
+## How the proof reward is determined
+
+A proof reward in TTKO is rewarded for successfully proving a block. This reward is dependent on the `proofTimeTarget`. You can see the current proof reward by calling `getProofReward` on the TaikoL1 contract.
```
davidtaikocha, line 5: Shall we mention the `realProofSkipSize` config here? https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/L1/TaikoConfig.sol#LL40C13-L40C30

PR #13821, opened 2023-05-26
feat(status-page): updates for a3

cyberhorsey:
- 8 decimals now, not 18 for tko token. read from contract directly
- fixes bug in calculating proof time in eventindexer
- fixes some issues with block indexing in eventindexer
- add proof time target to status page

stats endpoint returns correct values now:
```ts
{
    "id": 1,
    "averageProofTime": 143,
    "averageProofReward": 111418956,
    "numProofs": 741,
    "numVerifiedBlocks": 712
}
```

keeping in mind we cant index events before today, since the event signature changed.

our current `proofTimeTarget` is 150, so `143` is on point, and average proof reward for today is a more sensible `1.1` TKO.

Screenshot from 2023-05-26 01-44-45


packages/eventindexer/indexer/save_block_proven_event.go
```diff
@@ -75,30 +74,35 @@ func (svc *Service) saveBlockProvenEvent(
 	eventindexer.BlockProvenEventsProcessed.Inc()
 
 	if event.Prover.Hex() != systemProver.Hex() && event.Prover.Hex() != oracleProver.Hex() {
-		if err := svc.updateAverageBlockTime(ctx, event); err != nil {
+		if err := svc.updateAverageProofTime(ctx, event); err != nil {
 			return errors.Wrap(err, "svc.updateAverageBlockTime")
```
davidtaikocha, line 78:
```suggestion
			return errors.Wrap(err, "svc.updateAverageProofTime")
```

packages/eventindexer/indexer/save_block_proven_event.go
```diff
@@ -75,30 +74,35 @@ func (svc *Service) saveBlockProvenEvent(
...
+
 	stat, err := svc.statRepo.Find(ctx)
 	if err != nil {
 		return errors.Wrap(err, "svc.statRepo.Find")
 	}
 
 	proposedAt := block.ProposedAt
 
-	provenAt := time.Now().Unix()
+	provenAt := eventBlock.Time()
```
davidtaikocha, line 103: hmmm seems `eventBlock.Time()` should equal to `block.ProposedAt`

packages/eventindexer/indexer/subscribe.go
```diff
@@ -58,36 +58,40 @@ func (svc *Service) subscribeBlockProven(ctx context.Context, chainID *big.Int,
...
 
-				continue
-			}
-
-			block, err := svc.blockRepo.GetLatestBlockProcessed(chainID)
-			if err != nil {
-				log.Errorf("svc.subscribe, blockRepo.GetLatestBlockProcessed: %v", err)
-				continue
-			}
+					return
```
davidtaikocha, line 72: seems the deleted error handling logic is `continue` here?

packages/eventindexer/indexer/save_block_proven_event.go
cyberhorsey: i dont think so, `eventBlock.TIme()` is the time at which the L1 block was made with the `BlockProven` event with this Id inside of it.

`block.ProposedAt` is when the original block was proposed. 

eventBlock.Time() is greater than block.proposedAt.

cyberhorsey: i dont think so, `eventBlock.TIme()` is the time at which the L1 block was made with the `BlockProven` event with this Id inside of it.

`block.ProposedAt` is when the original block was proposed. 

eventBlock.Time() is greater than block.proposedAt.

packages/eventindexer/indexer/subscribe.go
```diff
@@ -182,35 +197,39 @@ func (svc *Service) subscribeBlockVerified(ctx context.Context, chainID *big.Int
...
-					Height:  event.Raw.BlockNumber,
-					Hash:    event.Raw.BlockHash,
-					ChainID: chainID,
-				})
+			go func() {
+				log.Infof("blockVerifiedEvent from subscription")
+
+				if err := svc.saveBlockVerifiedEvent(ctx, chainID, event); err != nil {
+					eventindexer.BlockVerifiedEventsProcessedError.Inc()
+					log.Errorf("svc.subscribe, svc.saveBlockProvenEvent: %v", err)
```
RogerLamTd, line 208:
```suggestion
					log.Errorf("svc.subscribe, svc.saveBlockVerifiedEvent: %v", err)
```

RogerLamTd (dismissed):

dionysuzx (approved):

RogerLamTd (approved):

davidtaikocha (approved):

dionysuzx (approved):

PR #13825, opened 2023-05-27
chore(protocol): remove prettier-plugin-solidity in favor of forge fmt

dionysuzx:
@dantaik we do not use prettier anymore to format solidity files. Instead, we are using `forge fmt`. 

`forge fmt` is already configured to run on git hook based on the config here: https://github.com/taikoxyz/taiko-mono/blob/main/lefthook.yml#L10. `lint:sol` will call `forge fmt`: https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/package.json#L11. 

The configuration for VSCode is found here: https://github.com/taikoxyz/taiko-mono/blob/main/.vscode/settings.json.

If for some reason formatting is not happening on your staged files locally, please do a `pnpm lefthook install`.

Question though: do we want print width set as 80 or 100 columns? I would suggest we keep 100 columns, and keep `solhint.json` configured to warn at a max-len above 120. If we move to 80 columns, we should configure `solhint.json` to warn at a max-len above 100.


dantaik:
Screenshot 2023-05-28 at 13 19 07

 @d1onys1us I really don't like the new styles, as an example above, the old formats looks much better by breaking lines.

d1onys1us:
> Screenshot 2023-05-28 at 13 19 07 Screenshot 2023-05-28 at 13 24 25 Screenshot 2023-05-28 at 13 24 15 Screenshot 2023-05-28 at 13 24 04
> @d1onys1us I really don't like the new styles, as examples above, the old format looks much better by breaking lines.
> 
> Let me know if you can config "forge fmt" to output similar format.

@dantaik please check [e5125c3](https://github.com/taikoxyz/taiko-mono/pull/13825/commits/e5125c3339df84b5a71e21718bbc10779fa0c239) to see what you think of it. the reason why the difference between `prettier-plugin-solidity` and `forge fmt` were so fundamentally difference was the print width. I updated the print width to be `80` columns wide.

however, i believe 80 column width might be slightly too aggressive. however, i understand many people enjoy a 80 column width for various reasons, we can definitely keep it at 80.

dantaik:
@d1onys1us I like the new tweak, I think I still prefer width to be 80 as the diff will be displayed better in GitHub split view.

dantaik:
Can we also remove ".prettierrc.json"?

dantaik:
May I also suggest adding the following:

multiline_func_header = 'all'
number_underscore = 'thousands'
wrap_comments = true

PR #13831, opened 2023-05-28
feat(protocol): add batch auction design doc

dantaik:


packages/protocol/docs/batch_auction_design.md
```diff
@@ -0,0 +1,162 @@
...
+
+```python
+# This is just to demostrate an idea, the actual implementation needs
+# a more careful algo design.
+#
+# proofTime shall also be considered, but is missing below.
+def will_new_bid_win(new_bid, old_bid):
+
+    # return False immediately if new_bid does not meet these conditions
+    if new_bid.bid_per_gas < old_bid.bid_per_gas * 0.9:
```
davidtaikocha, line 58:
```suggestion
    if new_bid.bid_per_gas > old_bid.bid_per_gas * 0.9:
```

packages/protocol/docs/batch_auction_design.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+### Batch-Based Approach
+In light of the high gas fees associated with Ethereum, a batch-based strategy for conducting auctions is recommended. This strategy grants the winning bidder the right to verify a batch of blocks, thus reducing per-block gas costs. During the testnet phases, we plan to begin with a smaller batch size, eventually scaling up to 256 blocks by the mainnet launch.
+
+### Provisions for Upcoming Blocks and Gas Costs
+To offset potential delays in ZKP, it's recommended to conduct auctions for forthcoming blocks even before they are proposed. This introduces a certain level of uncertainty for provers due to the unknown block gas used and data size at the beginning of the auction. To counter this, an auction pricing model based on the gas/data usage of the auctioned block is proposed. Here, the block reward would be calculated as `b*g`, where `b` is the winning bid in TKO tokens per gas and `g` is the actual gas used (or gas limit) by the block. In this context, `b` will be referred to as the *bid per gas*, or simply the *bid*.
```
davidtaikocha, line 37: > `g` is the actual gas used (or gas limit)

i would say gasUsed only since gasLimit can be manipulated by proposers?

packages/protocol/docs/batch_auction_design.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+### Bid Period, Proofing Window, and Managing Multiple Auctions
+The bid period for an auction begins with the first bid and lasts for 5 minutes. Blocks can only be proven after the auction has ended entirely. To ensure fairness, particularly for late participants, it is crucial not to end auctions prematurely. It is recommended that an auction concludes when the following conditions are met:
+
+1. The 5-minute bid period has elapsed.
+2. The most recent proposed block is within `min_blocks_to_end_auctions` blocks of the first block in the batch.
+
+Auctions are conducted in increasing order of block batches, and the next batch's auction can start as soon as the previous batch's auction has started.
```
davidtaikocha, line 116:
```suggestion
Auctions are conducted in increasing order of block batches, and the next batch's auction can start as soon as the previous batch's auction has ended.
```

dantaik: Actually if auction for block N started, then auction for block N+1 can start

packages/protocol/docs/batch_auction_design.md
dantaik: We charge `fee_per_gas * gas_limit` then refund `fee_per_gas * (gas_limit - gas_used)` when block is verified.

PR #13836, opened 2023-05-29
feat(protocol): implement batch block auction

adaki2004:
So this is the implementation of the Auction mechanism. Not tested and by any means it is ready !
I'd like to first challenge the design by you, then create tests when we think it is good to go.

You can see interfaces introduced in `TaikoL1.sol`, but some desc.:
- `bidForBatch()` : Bids for a batch with the following struct:
```
    struct Bid {
        uint256 batchId; 
        address prover;
        uint64 deposit;
        uint64 feePerGas;
    }
```

- `batchId` can be determined for a block by calling: `blockIdToBatchId(uint256 blockId)`
- Batches currently a range of 100 and this: `id(0):` 1..100, `id(1):` 101..200, etc.
- `isBlockProvableBy(uint256 blockId, address prover)` : It return if a block is proveable by that specified address. It can return true in 2 cases : 
1: address owns the winning bid (auction bid window elapsed, but proofTime window not yet), 
2: an address tho does not own the winning bid, but the proof window (currently set to 2 horus) elapsed

- `isBidAcceptable(TaikoData.Bid calldata bid)` : A helper for the clients, to know if their wannabe bids can be accepted based on the criteria (either bc there is not yet a bid, or 90% better then the current winning bid)
-` isBatchAuctionable(uint256 batchId)`: Again a helper, to see if auction is ongoing (either bc. there is no bid, or there is at least one bid but the time for auction window not yet elapsed)

Open questions:
1. Reward distribution mechanism is:  (see `_markBlockVerified()`)
- If the winning bidder submits a proof per given block, it gets back his deposit / block + the reward.
- If 'anoyne' submitted a block (so outside of proof window), it gets the reward + the half of the deposit/ block.
Is it OK ? Shall i modify something ?

2. Currently the proposer fee is tied to averageRewardPerBlock() (because i cannot tie it to average fee per gas, because at proposal only the gasLimit is available).
- Is it OK ? If so, what should happen when there is no data available for the average ?
- If not ok, what would be the preferred way ?


packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -114,6 +123,13 @@ library TaikoData {
         uint96 amount;
     }
 
+    struct Bid {
+        uint256 minFeePerGasInWei;
```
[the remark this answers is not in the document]
dantaik: I think we should have a Auction struct and a Bid struct

```solidity
// 2 bytes32
struct Bid {
   uint64 batchId; 
   address prover;
   uint16 proofWindow; // allow bider to use a smaller window to win a bid?
   uint64 deposit;
   uint64 feePerGas;
}

// still 2 bytes32
struct Auction {
   Bid bid;
   uint64 startedAt;
}

```

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -126,6 +142,7 @@ library TaikoData {
             ) forkChoiceIds;
         mapping(address account => uint256 balance) taikoTokenBalances;
         mapping(bytes32 txListHash => TxListInfo) txListInfo;
+        mapping(uint256 batchStartBlockId => Bid bid) bids;
```
dantaik, line 145: Use a batch ID here is better I think, `batchId = blockId / batch_size`

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -129,6 +129,28 @@ contract TaikoL1 is
...
 
+    /**
+     * Bid for proving rights of a batch.
+     *
+     * @param startBlockId The start blockId of the given batch
+     * @param minFeePerGasInWei The minimum fee, in wei, per gas you will accept
+     * being paid as a reward for proving the block.
+     */
+
+    function bidForBatch(uint256 startBlockId, uint256 minFeePerGasInWei)
```
dantaik, line 140: function bidForBlocks(Bid calldata bid) {

}

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,149 @@
...
+    error L1_ID_NOT_START_OF_A_BATCH();
+
+    function bidForBatch(
+        TaikoData.State storage state,
+        AddressResolver resolver,
+        TaikoData.Config memory config,
+        uint256 batchStartBlockId,
+        uint256 minFeePerGasInWei
+    ) internal {
+        // If it verified already -> Revert, otherwise it would be possible
```
dantaik, line 42: The block ID => batch mapping should consider the fact the genesis block is at 0, and the first batch will cover block #1 to block #100, assuming batch size is 100.

```solidity
funciton blockIdToBatch(uint blockId) {
   return (blockId - 1)/batch_size;
}
```

dantaik: Block proposal and batch auction can be entirely independent: we can have a million blocks proposed and we are still auctioning the first 100K blocks, or we can have more block batch auctioned but a few blocks proposed. Decoupling these two will make things easier.

Since proofs cannot be submitted without its batch auction complete. When someone bid for a batch, we do not need to check if lastVerifiedBlockId is smaller/larger than the first block in the batch. We just need to check if the auction has started/ended.

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,149 @@
...
+        if (!isBiddingOpenForBlock(config, currentBid)) {
+            revert L1_BID_AUCTION_CLOSED();
+        }
+
+        TaikoToken tkoToken = TaikoToken(resolver.resolve("tko_token", false));
+
+        // If there is an existing bid already
+        // we compare this bid to the existing one first, to see if its a
+        // lower fee accepted.
+        if (currentBid.bidder != address(0)) {
```
dantaik, line 66: Always use a ring buffer please, this check is then incorrect -- we need to check if batchId matches, bidder (call it prover please) will not be reset to address(0).

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,149 @@
...
+
+        TaikoData.Bid memory currentBid = state.bids[batchStartBlockId];
+
+        // if a current bid exists, and the delay after the current bid exists
+        // has passed, bidding is over.
+        if (!isBiddingOpenForBlock(config, currentBid)) {
+            revert L1_BID_AUCTION_CLOSED();
+        }
+
+        TaikoToken tkoToken = TaikoToken(resolver.resolve("tko_token", false));
```
dantaik, line 61: Lookup the token address only when necessary)

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,149 @@
...
+            if (!isBidAcceptable(config, currentBid, minFeePerGasInWei, msg.value)) {
+                revert L1_BID_NOT_ACCEPTABLE();
+            } else {
+                // otherwise we have a new high bid, and can refund the previous claimer deposit
+                try tkoToken.transfer(currentBid.bidder, currentBid.deposit) {}
+                catch {
+                    // allow to fail in case they have a bad onTokenReceived
+                    // so they cant be outbid
+                }
+                emit BidDepositRefunded(
```
dantaik, line 76: I don't think we need to emit this event.

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,149 @@
...
+
+        // If there is an existing bid already
+        // we compare this bid to the existing one first, to see if its a
+        // lower fee accepted.
+        if (currentBid.bidder != address(0)) {
+            if (!isBidAcceptable(config, currentBid, minFeePerGasInWei, msg.value)) {
+                revert L1_BID_NOT_ACCEPTABLE();
+            } else {
+                // otherwise we have a new high bid, and can refund the previous claimer deposit
+                try tkoToken.transfer(currentBid.bidder, currentBid.deposit) {}
```
dantaik, line 71: We should use mint/burn here...

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    // isBidAcceptable determines is checking if the bid is acceptable based on the defined
+    // criteria
+    function isBidAcceptable(
+        TaikoData.Config memory config,
+        TaikoData.Bid memory bid,
+        uint256 minFeePerGasInWei,
+        uint256 depositAmount
+    )
```
dantaik, line 119: Make it virtual please

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,149 @@
...
+        // We also need to be sure, that the blockId here is indeed a start blockId
+        if (batchStartBlockId != getBatchStartBlockId(config, batchStartBlockId)) {
+            revert L1_ID_NOT_START_OF_A_BATCH();
+        }
+
+        TaikoData.Bid memory currentBid = state.bids[batchStartBlockId];
+
+        // if a current bid exists, and the delay after the current bid exists
+        // has passed, bidding is over.
+        if (!isBiddingOpenForBlock(config, currentBid)) {
```
dantaik, line 57: `isBatchAuctionable(uint64 batchId)`?

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,149 @@
...
+    function getBatchStartBlockId(
+        TaikoData.Config memory config,
+        uint256 blockId
+    ) internal pure returns (uint256 startBlockId){
+        return blockId - blockId % config.auctionBatchModulo;
+    }
+
+    // isBidAcceptable determines is checking if the bid is acceptable based on the defined
+    // criteria
+    function isBidAcceptable(
```
dantaik, line 114: We also need to expose a public function

```solidity
isBidAcceptable(Bid) {}
```

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -116,11 +120,9 @@ library LibVerifying {
             signalRoot = fc.signalRoot;
 
             _markBlockVerified({
-                state: state,
                 blk: blk,
                 fcId: uint24(fcId),
-                fc: fc,
-                systemProver: systemProver
+                fc: fc
```
dantaik, line 125: in LibProving, we also need to expose a function:

```solidity
isBlockProvableBy(uint blockId, address prover) returns bool
```

dantaik:
@adaki2004 Do you think we can get this done by the end of this week?

packages/protocol/contracts/L1/TaikoData.sol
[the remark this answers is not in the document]
dantaik: Two reasons to have the proving window there:
- the proving window should be adjusted automatically after a block is verified, but we do not to want to proving window for a specific block or a batch to be dynamic, so better to keep a copy there when the auction started.
- The proving window is another parameter to the bid comparison function, this allows us to favor bids that promises with earlier proofs to win if we want.

packages/protocol/contracts/L1/libs/LibAuction.sol
dantaik: Lets expect provers to have their own ways of distributing blocks of a batch to its workers. They may have their own auction offchain.

packages/protocol/contracts/L1/libs/LibAuction.sol
dantaik: They are not sharing the same ring buffer, so why it is not possible?

packages/protocol/contracts/L1/libs/LibAuction.sol
dantaik: Fine for bids, but for reward and penalty, we should use mint/burn.

dantaik:
> > @adaki2004 Do you think we can get this done by the end of this week?
> 
> I'm almost sure with testing + simulation it is a no. Excluding test + simulaiton, yes.

Lets get to core impl done, and worry about testing later

dantaik:
> If 'anoyne' submitted a block (so outside of proof window), it gets the reward + the half of the deposit/ block.
Is it OK ? Shall i modify something ?

The other half gets burnt

dantaik:
> Currently the proposer fee is tied to averageRewardPerBlock() (because i cannot tie it to average fee per gas, because at proposal only the gasLimit is available).

Is it a good idea to charge `avgFeePerGas * gasLimit` then refund `avgFeePerGas * (gasLimit - gasUsed)` when the block is verified?

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -163,10 +173,42 @@ library LibVerifying {
...
+
+        // Check if prover is equal to auction winner. If so, just distribute normall if not, distribute reward normall
+        // + half of the deposit / block
+        uint256 batchId = LibAuction.blockIdToBatchId(config, blk.blockId);
+        TaikoData.Auction memory auction = state.auctions[batchId];
+
+        TaikoToken tkoToken = TaikoToken(resolver.resolve("tko_token", false));
+        // Prover indeed the one who won the auction
+        if(auction.bid.prover == fc.prover) {
+            try tkoToken.transfer(auction.bid.prover, auction.bid.deposit / config.auctionBatchSize + rewardPerBlock) {}
```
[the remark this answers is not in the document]
dantaik: Even if you want to use `trasnfer`, you should use `transferFrom`. Burning and minting bear the same cost as transfer/transferFrom.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -163,10 +173,42 @@ library LibVerifying {
...
+            catch {
+                // allow to fail in case they have a bad onTokenReceived
```
dantaik, line 192: I don't think it's a good idea to allow failure here. We controls the implementation of `onTokenReceived`, so no failure should be tolerated.

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,283 @@
...
+            } else {
+                // We have a new high bid, refunding back the deposit credited
+                // with taikoTokenBalances
+                state.taikoTokenBalances[auction.bid.prover] +=
+                    auction.bid.deposit;
+            }
+        }
+
+        // New winner, so deduct deposit
+        if (state.taikoTokenBalances[newBid.prover] < auction.bid.deposit) {
```
dantaik, line 76: `< auction.bid.deposit` or `< newBid.deposit`?

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,283 @@
...
+            revert L1_INSUFFICIENT_TOKEN();
+        }
+        state.taikoTokenBalances[newBid.prover] -= newBid.deposit;
```
dantaik, line 79: Where is the code to compare two bid, the old and the new?

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,283 @@
...
+
+        // then we can update the bid for the blockID to the new bidder (prover)
+        state.auctions[newBid.batchId] = TaikoData.Auction({
```
dantaik, line 82: `state.auctions[newBid.batchId]` => `state.auctions[newBid.batchId % batchSize]`?

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,283 @@
...
+        TaikoData.Config memory config,
+        TaikoData.Bid calldata newBid
+    )
+        internal
+    {
+        if (!isBatchAuctionable(state, config, newBid.batchId)) {
+            revert L1_BID_CANNOT_BE_SUBMITTED();
+        }
+
+        TaikoData.Auction memory auction = state.auctions[newBid.batchId];
```
dantaik, line 45: TaikoData.Auction storage auction = state.auctions[newBid.batchId % config.auctionBufferSize];

dantaik (changes requested):

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,268 @@
...
+        internal
+        view
+        returns (bool result)
+    {
+        if(batchId == 0) {
+            return true;
+        }
+
+        //Otherwise batchId is above 0, so we can deduct 1 safely
+        TaikoData.Auction memory auction = state.auctions[(batchId-1) % config.auctionRingBufferSize];
```
dantaik, line 121: return false if `auction.batchId != batchId`

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,268 @@
...
+        view
+        returns (bool result)
+    {
+        // We should expose this function so that clients could query.
+        // We also need to be sure, that the batchId of bid is indeed valid
+        if (batchId != blockIdToBatchId(config, batchId)) {
+            revert L1_ID_NOT_BATCH_ID();
+        }
+
+        if (!isPreviousAuctionEnded(state, config, batchId)){
```
dantaik, line 182: This batch can be auctioned if the previous batch's action has started (not ended)

cyberhorsey:
> 

Would another public helper, to extend `isBidBetter`, be helpful, like `minimumNextBidForBatch`?

PR #13839, opened 2023-05-30
feat(protocol): (Alpha-3) Cap fee within 0.1 and 10 TKO

adaki2004:
Capping the fee and do not allow accumulation of `proofTimeIssued` (allowing faster recovery) in such case it is below/above 0.1/10 TKO. (We could have done this by calculating the corresponding proofTImeIssued reverse, but it would cost more gas.)

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibTokenomics.sol
+++ b/packages/protocol/contracts/L1/libs/LibTokenomics.sol
@@ -107,5 +107,15 @@ library LibTokenomics {
             / (state.proofTimeTarget * state.adjustmentQuotient);
 
         blockFee = uint64(result.min(type(uint64).max));
+
+        // Keep it within 0.1 and 10 TKO and not allow proofTimeIssued accumulated
+        // so that fee could recover quicker when waiting is applied from provers.
+        if (blockFee < 1e7) {
+            blockFee = 1e7;
+            newProofTimeIssued -= proofTime;
+        } else if (blockFee > 1e9) {
+            blockFee = 1e9;
+            newProofTimeIssued -= proofTime;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


Brechtpd (approved):

packages/protocol/contracts/L1/libs/LibTokenomics.sol
```diff
@@ -107,5 +107,15 @@ library LibTokenomics {
             / (state.proofTimeTarget * state.adjustmentQuotient);
 
         blockFee = uint64(result.min(type(uint64).max));
+
+        // Keep it within 0.1 and 10 TKO and not allow proofTimeIssued accumulated
+        // so that fee could recover quicker when waiting is applied from provers.
+        if (blockFee < 1e7) {
```
davidtaikocha, line 113: Shall we also cap the reward?

davidtaikocha (approved):

packages/protocol/contracts/L1/libs/LibTokenomics.sol
dantaik: Will this prevent us from identifying issues?

Brechtpd: I don't think so, it will just help stabilize things when proofs aren't being submitted at the normal pace. So either too fast or not at all when there's a network issue (problem with the prover, etc...). If that happens without the cap it would just take a longer time for things to get back to normal again.

Brechtpd: > Will this prevent us from identifying issues?

Or wait do you mean a cap on the fee or a cap on the reward? If on the reward, yes I think that could hide some issues.

PR #13843, opened 2023-05-31
feat(repo): use label to trigger chatGPT based auto review

dantaik:
Going forward, please add "option.aireview" to trigger the open AI review.


d1onys1us:
yes, the reason why is those default organization labels actually don't sync across our repos. they are only the defaults that are used when initializing a repo.

so we would need to add this label manually to taiko-client, taiko-mono, etc.

side note: there are some github actions to achieve label sync cross repo, but not provided by github by default.

dionysuzx (approved):

davidtaikocha (approved):

Issue #13783, opened 2023-05-19
Use ring buffer for pending Ether Deposits


dantaik:
### Describe the feature request

As of now, the `EthDeposit[] ethDeposits` is an ever growing array. It can be turned into a ring buffer for reducing gas cost.

### Describe alternatives you've considered

Description of the alternatives you've considered here.

### Additional context

Additional context here.

Venoox:
Do you think it's a good idea to implement this with a fixed size array and having two variables to store the index of the start and the end of the buffer?

cyberhorsey:
> I'd like to try this Do you think it's a good idea to implement this with a fixed size array and having two variables to store the index of the start and the end of the buffer?

I think we can just modulo ID by the ringBufferSize? Haven't looked at the code much for this but I think that's how our other ring buffers are implemented.

Venoox:
When the buffer is full and a new deposit is made, we overwrite the oldest deposit, is that correct?

dantaik:
@Venoox I assigned this feature to you. Do you think you can give us an ETA?

Venoox:
A few days would be my best guess

dantaik:
> A few days would be my best guess

I got it done here: https://github.com/taikoxyz/taiko-mono/pull/13868

Venoox:
Oh I've just finished working on it :/

RogerRabbitTH:
Good

PR #13853, opened 2023-06-02
fix(protocol): hash deposit IDs

dantaik:
With this fix, Blockscout can display ETH deposits correctlhy.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibEthDepositing.sol
+++ b/packages/protocol/contracts/L1/libs/LibEthDepositing.sol
@@ -37,6 +37,7 @@ library LibEthDepositing {
         }
 
         TaikoData.EthDeposit memory deposit = TaikoData.EthDeposit({
+            id: uint64(state.ethDeposits.length),
             recipient: msg.sender,
             amount: uint96(msg.value)
         });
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/L1/libs/LibEthDepositing.sol
```diff
@@ -38,7 +38,8 @@ library LibEthDepositing {
 
         TaikoData.EthDeposit memory deposit = TaikoData.EthDeposit({
             recipient: msg.sender,
-            amount: uint96(msg.value)
+            amount: uint96(msg.value),
+            id: uint64(state.ethDeposits.length)
```
dantaik, line 42: `uint64(state.ethDeposits.length + 1)` is 0 allow here by the client?

davidtaikocha: `id == 0` is allowed, since right now all withdrawal's id is `0`. but i believe the potential issue of this approach is we can not do that optimization any more? https://github.com/taikoxyz/taiko-mono/issues/13783 , which seems is quite necessary?

davidtaikocha, line 42: and shall we use state.nextEthDepositToProcess?

dong77: We can still do it, but when it's done, we need some changes, but it's OK.

davidtaikocha (approved):

smtmfft (approved):

PR #13856, opened 2023-06-02
chore(bridge-ui): improve type checking

jscriptcoder:
Using `Address` type to improve type checking here and prevent potential bugs from working with something that's not an address

### DIFF
--- a/packages/bridge-ui/src/components/BridgeForm/BridgeForm.svelte
+++ b/packages/bridge-ui/src/components/BridgeForm/BridgeForm.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { fetchFeeData } from '@wagmi/core';
+  import { type Address,fetchFeeData } from '@wagmi/core';
   import { BigNumber, Contract, ethers, type Signer } from 'ethers';
   import { _ } from 'svelte-i18n';
 
@@ -357,7 +357,7 @@
       // Set it manually.
       tx.chainId = $srcChain.id;
 
-      const userAddress = await $signer.getAddress();
+      const userAddress = (await $signer.getAddress()) as Address;
 
       let transactions: BridgeTransaction[] =
         await storageService.getAllByAddress(userAddress);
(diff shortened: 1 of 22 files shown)


packages/bridge-ui/src/components/BridgeForm/BridgeForm.svelte
```diff
@@ -357,7 +357,7 @@
       // Set it manually.
       tx.chainId = $srcChain.id;
 
-      const userAddress = await $signer.getAddress();
+      const userAddress = (await $signer.getAddress()) as Address;
```
d1onys1us, line 360:
```suggestion
      const userAddress: Address = (await $signer.getAddress());
```
nit

d1onys1us: can we use this? it's safer if possible (check at compile time instead of runtime error)

packages/bridge-ui/src/signer/subscriber.ts
```diff
@@ -1,4 +1,5 @@
-import type { Signer } from 'ethers';
+import type { Address } from '@wagmi/core';
+import { constants,type Signer } from 'ethers';
```
d1onys1us, line 2:
```suggestion
import { constants, type Signer } from 'ethers';
```

packages/bridge-ui/src/vault/tokenVaults.ts
```diff
@@ -1,10 +1,12 @@
+import type { Address } from '@wagmi/core';
+
```
d1onys1us, line 2:
```suggestion
```

packages/bridge-ui/src/components/BridgeForm/BridgeForm.svelte
jscriptcoder: That doesn't work. You cannot assign a `string` to an `Address`. You need to cast the `string`

packages/bridge-ui/src/vault/tokenVaults.ts
jscriptcoder: Nope. I'm using this plugging to sort and organize the imports, https://github.com/lydell/eslint-plugin-simple-import-sort

dantaik (approved):

dionysuzx (approved):

PR #13900, opened 2023-06-06
fix(bridge-ui): inform connect when adding erc20

jscriptcoder:
No need to waste user's time showing them something they can't use. Show a message if they try to add an ERC20 token without being connected


packages/bridge-ui/src/components/BridgeForm/SelectToken.svelte
```diff
@@ -33,6 +37,17 @@
     closeDropdown();
   }
 
+  function showAddERC20() {
+    if (!$signer) {
+      // No need to waste user's time showing them something
+      // they can't use
+      warningToast('Please, connect your wallet.');
```
cyberhorsey, line 44: not sure this comma is necessary/granmatiicaly correct/typical?

jscriptcoder: Ok, let me get rid of it

dionysuzx (approved):

cyberhorsey (approved):

PR #13925, opened 2023-06-08
feat(protocol): Auction testing

adaki2004:
Test auction mechanism.


dantaik:
> Just as a benchmark, auction vs. current economics gas consumption: kép

The first image corresponds to the auction impl?

PR #13956, opened 2023-06-11
feat(protocol): implement batch block auction (#13836)

dantaik:
Dani's initial implementation of a batch auction based on [this design doc](https://github.com/taikoxyz/taiko-mono/blob/alpha-4-base/packages/protocol/docs/batch_auction_design.md).


packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,275 @@
...
+            unchecked {
+                state.numAuctions += 1;
+            }
+        } else {
+            // An ongoing one
+            if (!isBidBetter(bid, auction.bid)) {
+                revert L1_NOT_THE_BEST_BID();
+            }
+            //'Refund' current
+            state.taikoTokenBalances[auction.bid.prover] += totalDeposit;
```
davidtaikocha, line 78: here we need to refund `auction.bid.deposit * config.auctionBatchSize` but not `bid.deposit * config.auctionBatchSize`?

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,275 @@
...
+        TaikoData.Auction memory auction =
+            state.auctions[batchId % config.auctionRingBufferSize];
+
+        // Deposit amount is per block, not per block * auctionBatchSize
+        uint64 totalDeposit = bid.deposit * config.auctionBatchSize;
+
+        if (batchId != auction.batchId) {
+            // It is a new auction
+            auction.startedAt = uint64(block.timestamp);
+            auction.bid = bid;
```
davidtaikocha, line 67: seems this line of code is unnecessary since we will do `auction.bid = bid;` again in line 87?

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -115,9 +129,28 @@ library TaikoData {
...
+    struct Auction {
+        Bid bid;
+        uint64 batchId;
+        uint64 startedAt;
+    }
+
     struct State {
         // Ring buffer for proposed blocks and a some recent verified blocks.
-        mapping(uint256 blockId_mode_ringBufferSize => Block) blocks;
+        mapping(uint256 blockId_mode_blockRingBufferSize => Block) blocks;
```
davidtaikocha, line 153:
```suggestion
        mapping(uint256 blockId_mod_blockRingBufferSize => Block) blocks;
```

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,275 @@
...
+        TaikoData.State storage state,
+        TaikoData.Config memory config,
+        uint256 startBatchId,
+        uint256 count
+    )
+        internal
+        view
+        returns (uint256 currentTime, TaikoData.Auction[] memory auctions)
+    {
+        if (startBatchId == 0 || count == 0) {
```
davidtaikocha, line 104: Do we need to also check `startBatchId` is not overflow / underflow the current ringbuffer?

packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -0,0 +1,275 @@
...
+    // Determines if bid a is better than bid b
+    function isBidBetter(
+        TaikoData.Bid memory a,
+        TaikoData.Bid memory b
+    )
+        internal
+        pure
+        returns (bool)
+    {
+        // Normalize both feePerGas and feePerGas to a comparable scale.
```
davidtaikocha, line 185:
```suggestion
        // Normalize both feePerGas and deposit to a comparable scale.
```

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -105,10 +116,28 @@ library LibProposing {
...
+        TaikoData.Config memory config,
+        uint32 gasLimit
+    )
+        internal
+        view
+        returns (uint64)
+    {
+        // The diff between gasLimit and gasUsed will be redistributed back to
+        // the balance of proposer
+        return state.feePerGas * (gasLimit + config.blockFeeBaseGas);
```
davidtaikocha, line 138: Shall we take https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/L2/LibL2Consts.sol#L10 into consideration?

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -149,22 +171,113 @@ library LibVerifying {
...
+
+            // For system prover, maybe the auction does not exist.
+            if (auction.batchId != batchId) {
+                assert(fc.prover == address(1));
+                auction.batchId = 0; // indicating auction yet to start
+            }
+        }
+
+        // Refund the diff to the proposer
+        assert(fc.gasUsed <= blk.gasLimit);
```
davidtaikocha, line 194: the actually mined L2 block's gasLimit is `blk.gasLimit + LibL2Consts.ANCHOR_GAS_COST`, so `fc.gasUsed` may greater than `blk.gasLimit` here.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -143,6 +171,19 @@ contract TaikoL1 is
         });
     }
 
+    // From proposer side - same way paying the fees - and saving gas.
+    function depositTaikoToken(uint256 amount) external nonReentrant {
```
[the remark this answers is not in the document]
dantaik: totalSupply must be smaller than `type(uint64).max`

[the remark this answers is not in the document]
davidtaikocha: yeah so the type of `amount` should not be uint256?

packages/protocol/contracts/L1/libs/LibAuction.sol
dantaik: It's a bug, fixed.

packages/protocol/contracts/L1/libs/LibAuction.sol
dantaik: Removed

packages/protocol/contracts/L1/libs/LibAuction.sol
dantaik: We don't need to. This function provides auctions for the client to verify (such as whether the auction is empty or not).

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: Added

packages/protocol/contracts/L1/TaikoL1.sol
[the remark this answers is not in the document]
dantaik: Fixed

dantaik:
> * Taking the deposit amount into account for the bid is a bit strange, especially given the goals (efficiency/decentralization). Does this allow the prover with the most TKO to win all bids and set the price almost as it wishes?
The idea is that we can give `feePerGas`, `deposit`, `proofTime` different weights in the `isBidBetter` function. Currently, we use:

```solidity
 return _adjustBidPropertyScore(ONE * b.feePerGas / a.feePerGas, 1)
            * _adjustBidPropertyScore(ONE * a.deposit / b.deposit, 2)
            * _adjustBidPropertyScore(ONE * b.proofWindow / a.proofWindow, 4)
            >= ONE * ONE * ONE * 110 / 100;
```

Which means we give fee the most weight, give deposit 50% of the weight, and proof window 25% of the weight. If we simply want to cheapest proof, then we just comment out the lines for deposit and proofWindow.

> * As the doc also says, do we actually believe different provers will win different auctions? If efficiency is the goal then surely max centralization (at least from the on-chain point of view) is the most efficient and will be how it will actually go. Changing the score to take into account other measurable things on-chain (deposit amount, historical data,...) seem to increase centralization even more by making it harder for new provers to join.

I think if we choose to priority to fastest provers, it will also have the same centralization issue. Maybe we use different weights for different batches, i.e., for `batchId% 3 == 0`. we give `proofWindow` more weight,  `batchId% 3 == 1`. we give `deposit` more weight,  ...
But regardless, I think eventually it will be quite centralized.

I also feel like people won't compete on-chain, they will probably join a off-chain auction pool and bid off-chain, their agent will bid for them on-chain.

> * The winning bid should be extremely stable. So does it make sense to have the prover bid for each batch of blocks if 99.9% of time the same bid (and extremely likely, same prover) will win anyway? Assuming the most efficient prover is pretty stable, and as long as it's possible for the more efficient prover to start proving the blocks instead, should these explicit auctions be held? What do we gain by doing this vs just making provers auto-bid with their required reward without having to do anything.

We cannot assume the winner of auctions is willing to prove the blocks in the next batch. The bidding process is really inexpensive once the auction starts to re-use ring buffers.


> * Because the score is fully deterministic, how will the auction mechanism ensure prover redundancy? I already briefly touched this subject above, but I don't think this is measurable on-chain (when the prover missed its expected proof time it's already too late and liveness was already lost, and if this is taken into account in the score in the future then surely the prover will just switch its address as well).

The worse case scenario is a block to have its proof twice the time of average proof delay if the auction winner didn't provide the proof in time. That's why we need the bidder to deposit a large deposit to win the batch.

> * The bidding mechanism does make it easy for the prover that normally wins all auctions to slowly creep up the price. When someone would place a lower bid, but because the most efficient prover is the most efficient he could still outbid that bid again. This makes sure that someone acting in "good faith", bidding reasonable fees, would in the future not do so anymore because it's a lose-lose situation because the L1 transaction cost is a pure loss. This makes it quite likely the sole prover will be able to increase the block fee substantially without anyone doing something about it.

I think it's OK for the sole prover to gradually increase the fee because once the fee is high enough, another prover who works on other zkRollup will start to prove Taiko's block; if other provers are still not interested in proving Taiko's blocks, the reward is not good enough and it should probably go up. I think the reward is only too high if there are at least two provers to compete with each other.

dantaik:
> 

If provers have the option to submit either a proof per block or a proof per multiple blocks, the decision-making factor will primarily depend on the comparative cost between one-proof-per-block and one-proof-per-multiple-blocks approaches, considering both offchain aggregation cost and onchain proof verification cost.

The protocol's tokenomics should solely care whether a block is proven within its designated proof window, without distinguishing between standalone proofs or aggregated proofs. The protocol, therefore, should incentivize provers who can prove blocks ASAP regardless of the actually proof type. *I'm not sure this is the case in our current implementation though*.

adaki2004 (approved):

cyberhorsey:
> 

Dani, I closed my PR. Ring buffer size can be changed in this PR. You can also fix the typo from `aucton` => `auction` here.

PR #13981, opened 2023-06-13
feat(protocol): alpha-4 with staking-based tokenomics

dantaik:


cyberhorsey:
Working client impl is here: https://github.com/taikoxyz/taiko-client/pull/292
it uses `ProverPool` for test setup, not `ProverPool2`, but ill follow here to see if a switch happens.

PR #13987, opened 2023-06-14
fix(protocol): Fix deploy script, and typo

cyberhorsey:
this small PR fixes a typo, and also fixes the deploy script, which had a mathematical error rendering it undeployable with `L1_INVALID_CONFIG`.

I seem to have differences in formatting on save, @Daniel  W | Taiko is there something new i need to install to work on the Solidity files? The only real change in this file is adding `( )` around the demoninator in this math in `LibVerifying`, and a typo from `aucton` => `auction`

```ts
  (config.maxNumProposedBlocks /
                (config.auctionBatchSize +
                    1 +
                    config.auctionMaxAheadOfProposals)) || //
  ```

### DIFF
--- a/packages/protocol/contracts/L1/TaikoConfig.sol
+++ b/packages/protocol/contracts/L1/TaikoConfig.sol
@@ -6,52 +6,51 @@
 
 pragma solidity ^0.8.20;
 
-import {TaikoData} from "../L1/TaikoData.sol";
+import { TaikoData } from "../L1/TaikoData.sol";
 
 library TaikoConfig {
     function getConfig() internal pure returns (TaikoData.Config memory) {
-        return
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/protocol/contracts/L1/TaikoConfig.sol [hunk 1]
```diff
@@ -47,7 +47,7 @@ library TaikoConfig {
             auctionProofWindowMultiplier: 2,
             auctionDepositMultipler: 10,
             auctionMaxFeePerGasMultipler: 5,
-            auctonMaxAheadOfProposals: 10,
```
dantaik, line 50: I ran `pnpm fmt:sol`

dantaik (changes requested):

PR #13991, opened 2023-06-15
refactor(protocol): simplify oracle proof and system proof

dantaik:
Simplify oracle/system proof by merging them into a one

|  type | Can override existing | Can be overridden | Can prove without auction ended | Can verify block |
|----------|----------|----------|----------|----------|
|   oracle proof  |   Yes  |   Yes  |   Yes  |   No |
|   system proof |   No  |   Yes - only by oracle and regular proof  |   No  |   Yes  |
|   **new oracle proof**  |   Yes  |   Yes  |   Yes  |   Yes |

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibAuction.sol
+++ b/packages/protocol/contracts/L1/libs/LibAuction.sol
@@ -132,29 +132,23 @@ library LibAuction {
         view
         returns (bool provable, TaikoData.Auction memory auction)
     {
-        if (blockId != 0) {
-            if (prover == address(0)) {
-                // Note that auction may not exist at all.
+        // Nobody can prove a block before the auction ended,
+        // including the oracle prover
+        bool ended;
+        (ended, auction) = _hasAuctionEnded({
+            state: state,
+            config: config,
+            batchId: batchForBlock(config, blockId)
+        });
+
+        if (ended) {
+            if (prover == address(1) || prover == auction.bid.prover) {
                 provable = true;
             } else {
-                // Nobody can prove a block before the auction ended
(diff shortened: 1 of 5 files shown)


packages/protocol/contracts/L1/libs/LibAuction.sol
```diff
@@ -132,29 +132,23 @@ library LibAuction {
         view
         returns (bool provable, TaikoData.Auction memory auction)
     {
-        if (blockId != 0) {
```
dantaik, line 135: block != 0 and prover != address(0) are guaranteed.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -145,7 +139,7 @@ library LibProving {
                 state.forkChoiceIds[blk.blockId][evidence.parentHash][evidence
                     .parentGasUsed] = fcId;
             }
-        } else if (evidence.prover == address(0)) {
+        } else if (evidence.prover == address(1)) {
```
dantaik, line 142: the new oracle proof use address(1) not address(0)

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -234,42 +234,36 @@ library LibVerifying {
 
         if (refundBidder) {
             state.taikoTokenBalances[auction.bid.prover] += auction.bid.deposit;
-        } else {
```
dantaik, line 237: I think it's buggy here.

adaki2004 (approved):

PR #13992, opened 2023-06-15
feat(protocol): Fix that after proof window elapses anyone is authorized

adaki2004:
Fix that after proof window elapses anyone is authorized not only `auction.bid.prover`

### DIFF
--- a/packages/protocol/contracts/L1/TaikoL1.sol
+++ b/packages/protocol/contracts/L1/TaikoL1.sol
@@ -318,7 +318,7 @@ contract TaikoL1 is
     )
         public
         view
-        returns (bool, bool, TaikoData.Auction memory)
+        returns (bool provable, bool proofWindowElapsed, TaikoData.Auction memory auction)
     {
         return LibAuction.isBlockProvableBy(state, getConfig(), blockId, prover);
     }
--- a/packages/protocol/contracts/L1/TaikoL1.sol
+++ b/packages/protocol/contracts/L1/TaikoL1.sol
@@ -318,7 +318,7 @@ contract TaikoL1 is
     )
         public
         view
-        returns (bool, TaikoData.Auction memory)
+        returns (bool, bool, TaikoData.Auction memory)
     {
         return LibAuction.isBlockProvableBy(state, getConfig(), blockId, prover);
     }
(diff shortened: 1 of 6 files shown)


packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -318,7 +318,7 @@ contract TaikoL1 is
     )
         public
         view
-        returns (bool, TaikoData.Auction memory)
+        returns (bool, bool, TaikoData.Auction memory)
```
dantaik, line 321: better name the return values otherwise we don't know what to return :)

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -67,7 +67,7 @@ library LibProving {
 
         // We also should know who can prove this block:
         // the one who bid or everyone if it is above the window
-        (bool provable, TaikoData.Auction memory auction) = LibAuction
+        (bool provable, bool windowEnded, TaikoData.Auction memory auction) = LibAuction
```
dantaik, line 70:
```suggestion
        (bool provable, bool proofWindowEnded, TaikoData.Auction memory auction) = LibAuction
```

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -95,6 +95,17 @@ library LibProving {
             if (evidence.verifierId != 0 || evidence.proof.length != 0) {
                 revert L1_INVALID_PROOF();
             }
+        } else if(windowEnded) {
+            // If window ended - everyone could prove
+            authorized = msg.sender;
+
+            // If this is a signature mint, also everyone is authorized
```
dantaik, line 102: I think we can simply remove lines from 102 to 108, right?

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -95,6 +95,17 @@ library LibProving {
             if (evidence.verifierId != 0 || evidence.proof.length != 0) {
                 revert L1_INVALID_PROOF();
             }
+        } else if(windowEnded) {
```
dantaik, line 98:
```suggestion
        } else if(proofWindowEnded) {
```

dantaik (changes requested):

PR #14002, opened 2023-06-15
chore(website): Websocket Endpoint RPC Network Configuration 

MarcusWentz:


davidtaikocha (approved):

dantaik (approved):

d1onys1us:
@MarcusWentz i would suggest we extend the current table by adding an additional column instead of creating a new table, this would only be a one-line diff (a minimum viable documentation standard)

for sepolia where we don't have a web socket we can just put a "-" as a placeholder. for the column names i suggest we do: "RPC (https)" and "RPC (websocket)"

PR #14003, opened 2023-06-16
feat(protocol): allow oracle prover to indicate block not zk-provable

dantaik:
@davidtaikocha the client shall respect the new zkUnprovable field and treat blocks with this flag as empty blocks immediately, which may involve chain reorg on the client side.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoL1.sol
+++ b/packages/protocol/contracts/L1/TaikoL1.sol
@@ -318,7 +318,11 @@ contract TaikoL1 is
     )
         public
         view
-        returns (bool provable, bool proofWindowElapsed, TaikoData.Auction memory auction)
+        returns (
+            bool provable,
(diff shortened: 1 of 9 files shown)


packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -318,7 +318,11 @@ contract TaikoL1 is
     )
         public
         view
-        returns (bool provable, bool proofWindowElapsed, TaikoData.Auction memory auction)
+        returns (
```
dantaik, line 321: @adaki2004 please run `pnpm fmt:sol` when you submit code.

davidtaikocha (approved):

cyberhorsey (approved):

davidtaikocha:
after the discussion, closed this due to we may still want to mine the block with proposed transactions.

PR #14010, opened 2023-06-17
Staking improvement 1

dantaik:
Dani, based on your PR, I tried to use the staking prover pool in the core protocol. There are still many bugs in the code (other than your pool impl which I have not reviewed yet).


cyberhorsey:
This is just an opinion, but I think we try to optimize code too much too early, and we end up throwing a lot away. Maybe for this tokenomics model we can utilize Dani's user friendly public API and some unoptimized code, run it on a testnet, and see if we like the tokenomics model without worrying about how many slots something uses, or how we store the prover addresses?

packages/protocol/contracts/L1/StakingProverPool3.sol
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        TaikoToken(AddressResolver(this).resolve("taiko_token", false)).burn(
+            prover, totalAmount
+        );
+
+        emit ProverEntered(prover, totalAmount, rewardPerGas, capacity);
+    }
+
+    function stakePositionAdjustments(
```
cyberhorsey, line 153: anyone can use this function to adjust a stake position for another person, we need to keep track of a delagate on the Prover struct and set it to the original msg.sender, then check this on the `onlyProver` modifier?

packages/protocol/contracts/L1/StakingProverPool3.sol
```diff
@@ -0,0 +1,316 @@
...
+
+        stakers[proverToId[prover]] = Staker(
+            minimumStakerId, prover, newTotalAmount, rewardPerGas, capacity
+        );
+        provers[proverToId[prover]] = Prover(newTotalAmount, rewardPerGas);
+
+        // Need to determine again who is the last / lowest staker
+        (minimumStakerId, minimumStake) = _determineLowestStaker();
+
+        emit ProverModified(prover, newTotalAmount, rewardPerGas, capacity);
```
cyberhorsey, line 196: IMO, the original design of the API was much better, as this design relies on the user providing their old values as defaults to this all-encompassing function, rather than the other design only providing the value you want to update? Each function would also be much less gas intensive, IE I'm paying for a lot of gas to do other stuff when I just want to adjust capacity?

packages/protocol/contracts/L1/StakingProverPool3.sol
```diff
@@ -0,0 +1,316 @@
...
+        address prover;
+        uint32 stakedAmount; // unit is 10^8, this means a max of
+        // 429496729500000000 tokens, 2.3283%  of total supply
+        uint16 rewardPerGas; // expected rewardPerGas
+        uint16 capacity; // maximum capacity (parallel blocks)
+    }
+
+    // Then we define a mapping from id => Staker
+    mapping(uint256 id => Staker) stakers;
+    // We need an address to id mapping
```
cyberhorsey, line 25: since we overwrite entries here, how does one get their staked info? Dont we need a mapping that just has ALL stakers in it, ie: `mapping(address staker => Staker stakerInfo) stakerInfos` so a user can pull down their current staked info?

packages/protocol/contracts/L1/StakingProverPool3.sol
```diff
@@ -0,0 +1,316 @@
...
+        for (uint8 i = 1; i < mProvers.length; i++) {
+            if (mProvers[i].stakedAmount < minStake) {
+                stakerId = i;
+                minStake = mProvers[i].stakedAmount;
+            }
+        }
+    }
+
+    // Pick a random prover
+    function _pickProverIdx(
```
cyberhorsey, line 303: missing a public `getProverForBlock` function, which previous implementation had?

packages/protocol/contracts/L1/ProverPool.sol
```diff
@@ -0,0 +1,484 @@
...
+            if (provers[31].stakedAmount != 0) {
+                (minimumStakerId, minimumStake) = _determineLowestStaker();
+            }
+        } else {
+            // List is full, we need to check which position we need to put them
+            // into
+            (minimumStakerId,) = _determineLowestStaker();
+
+            // We need to overwrite it's place
+            proverData[minimumStakerId] = ProverData(
```
cyberhorsey, line 294: Re-opening this question now since the file changed, which was:
- if i am the 32nd staker, and someone else comes in to replace me, you overwrite my place in this array. I am no longer in this array. I can not see what I have staked, nor my capacity/etc, nor can I even exit staking, since the `exit` functions look this up in the array.

My stake, since I have been replaced, is lost forever and I am unable to ever get it back. Even the record of what I staked is now gone. You have overwritten me.

There needs to be a way to keep track of ALL stakers, not just the top 32, or when replacing the 32nd person, you need to give them their stake back (which I like less).

As well, there is no way for me, as that now 33rd-infinite staker, to even see what my current stake IS. Maybe I only need to add some "X" TTKO to re-claim my spot in the top 32, but I am unable to fetch, from this contract, any info about my stake, as soon as I'm overwritten in this array.  

I hope this issue is explained better now than it was before. To put it simply, because you overwrite the entry in the array without respecting the original staker, that information and the ability to control the stake is lost forever, because you use `proverData[proverToId[_msgSender()]]` to control the stake in the future, but now, `proverData[myOriginalId]` is now a different staker.

PR #14034, opened 2023-06-22
feat(bridge-ui-v2): env vars

jscriptcoder:
Added `.env.example` and fixed build due to Typescript's path mapping.


packages/bridge-ui-v2/.env.example
```diff
@@ -0,0 +1,66 @@
+# Chain Names
```
d1onys1us, line 1: @jscriptcoder can we do all these with `export`? I'm not sure if it still works fine, if you could please check. 

why? because otherwise this does not support the sort of popular `fish` shell (okay I use it). see here: https://github.com/d1onys1us/dapp-slaps/blob/main/.env.example#L3

.github/workflows/bridge-ui-v2.yml
```diff
@@ -24,6 +24,13 @@ jobs:
       - name: Install pnpm dependencies
         uses: ./.github/actions/install-pnpm-dependencies
 
+      # This step is needed because the .env file is not committed to the repository
+      # and the Svelte check complains about not finding the exports from $env/static/public,
+      # which will be generated based on the .env file when running `svelte-kit sync`
+      - name: Make dummy envfile
```
d1onys1us, line 30: nit: rename to to "Copy .env.example to .env"

jscriptcoder: Done

packages/bridge-ui-v2/.env.example
jscriptcoder: Seems to work

dionysuzx (approved):

davidtaikocha (approved):

PR #14062, opened 2023-06-27
feat(protocol): ProverPool gas optimizations

adaki2004:
Before optimization, gas test of the ProoverPool:
kép

### DIFF
--- a/packages/protocol/contracts/L1/ProverPool.sol
+++ b/packages/protocol/contracts/L1/ProverPool.sol
@@ -41,6 +41,8 @@ contract ProverPool is EssentialContract, IProverPool {
     // reserve more slots than necessary
     uint256[10_000] private proverData;
     mapping(uint256 id => address prover) public idToProver;
+    // Save the weights only when: stake / unstaked / slashed
+    mapping(uint256 id => uint256 weights) public idToWeights;
     mapping(address staker => Staker) public stakers;
 
     uint256[48] private __gap;
@@ -153,6 +155,9 @@ contract ProverPool is EssentialContract, IProverPool {
             _saveProver(staker.proverId, prover);
         }
 
+        uint256 proverWeight = _calcWeight2(staker.maxCapacity, prover.stakedAmount * ONE_TKO, prover.rewardPerGas);
+        idToWeights[staker.proverId] = proverWeight;
+
         emit Slashed(addr, amountToSlash);
     }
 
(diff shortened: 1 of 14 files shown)


dantaik:
Please run `pnpm install` to trigger lints, or manually run `pnpm fmt:sol`, and maybe also run `pnpm lint:sol`, to fix these issues:


```
❯ pnpm lint:sol

> @taiko/protocol@0.10.0 lint:sol /Users/d/Projects/taiko/taiko-mono/packages/protocol
> forge fmt && pnpm solhint 'contracts/**/*.sol' --fix


contracts/L1/ProverPool.sol
  376:5  warning  Function order is incorrect, private view function can not go after private pure function (line 357)  ordering

contracts/L1/ProverPool2.sol
   26:5   warning  Function order is incorrect, struct definition can not go after custom error definition (line 24)  ordering
   36:5   warning  Explicitly mark visibility of state                                                                state-visibility
   37:5   warning  Explicitly mark visibility of state                                                                state-visibility
   58:9   warning  Variable name must be in mixedCase                                                                 var-name-mixedcase
  126:9   warning  Provide an error message for require                                                               reason-string
  137:9   warning  Provide an error message for require                                                               reason-string
  144:13  warning  Provide an error message for require                                                               reason-string
  183:9   warning  Provide an error message for require                                                               reason-string

contracts/L1/TaikoL1.sol
  160:5  warning  Function order is incorrect, external  function can not go after public payable function (line 150)  ordering

✖ 10 problems (0 errors, 10 warnings)
```

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -147,6 +147,7 @@ library TaikoData {
         mapping(bytes32 txListHash => TxListInfo) txListInfo;
         mapping(uint256 depositId_mode_ethDepositRingBufferSize => uint256)
             ethDeposits;
+        mapping(address account => uint256 balance) taikoTokenBalances;
```
dantaik, line 150: __gap should be changed to 42

packages/protocol/contracts/L1/ProverPool.sol
```diff
@@ -373,6 +391,23 @@ contract ProverPool is EssentialContract, IProverPool {
...
+        uint16 rewardPerGas
+    )
+        private
+        pure
+        returns (uint256)
+    {
+        if (currentCapacity == 0 || stakedAmount == 0 || rewardPerGas == 0) {
+            return 0;
+        } else {
+            return uint256(stakedAmount) / rewardPerGas / rewardPerGas;
```
dantaik, line 407: stakedAmount is uint32, rewardPerGas is also uint32, maybe we want to multiply stakedAmount with a const value (or `stakedAmount  << 64`) so the result is not likely to be 0?

PR #14065, opened 2023-06-28
feat(protocol): alpha-4 with staking-based tokenomics

dantaik:
Initial implementation of a token-staking based tokenomics.


packages/protocol/contracts/L1/ProverPool.sol
```diff
@@ -0,0 +1,364 @@
...
+        uint64 exitAmount;
+        uint16 maxCapacity;
+        uint8 proverId; // 0 to indicate the staker is not a top prover
+    }
+
+    // Given that we only have 32 slots for the top provers, if the protocol
+    // can support 1 block per second with an average proof time of 1 hour,
+    // then we need a min capacity of 3600, which means each prover shall
+    // provide a capacity of at least 3600/32=112.
+    uint32 public constant MAX_CAPACITY_LOWER_BOUND = 128;
```
hugo-blue, line 38: In A4, I think the proof won't cost 1 hour so the lower bound may be too big.

packages/protocol/contracts/L1/ProverPool.sol
```diff
@@ -0,0 +1,364 @@
...
+            if (provers[proverId].stakedAmount > provers[i].stakedAmount) {
+                proverId = i;
+            }
+            unchecked {
+                ++i;
+            }
+        }
+
+        if (provers[proverId].stakedAmount >= amount) {
+            revert PROVER_NOT_GOOD_ENOUGH();
```
hugo-blue, line 285: If it's not good enough, can we have the amount not burned so that the staker can withdraw soon?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -71,28 +99,57 @@ library LibProposing {
...
+            blk.assignedProver = assignedProver;
+
+            // Cap the reward to a range of [95%, 105%] * blk.feePerGas, if
+            // rewardPerGasRange is set to 5% (500 bp)
+            uint32 diff = blk.feePerGas * config.rewardPerGasRange / 10_000;
+            blk.rewardPerGas = uint32(
+                uint256(rewardPerGas).min(state.feePerGas + diff).max(
+                    state.feePerGas - diff
+                )
+            );
```
hugo-blue, line 133: As the prover has at least 95% reward, the prover may use a very small rewardpergas value to gain bigger weight?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -71,28 +99,57 @@ library LibProposing {
...
+
+            blk.proofWindow = uint16(
+                uint256(state.avgProofDelay * 3).min(config.proofMaxWindow).max(
```
hugo-blue, line 136: *3 maybe too long. The prover with a high staking amount may use a less efficient machine to reduce cost, and the average proof delay may increase. This design has no encouragement to prover of faster speed. There may be some concern that all provers use cheap machines and have long proof time. That means if the provers found that the total cost of more but less efficient machines is less than less but highly efficient machines, the provers will choose the former, and the average proof time won't be reduced gradually. Though we are not clear about this yet, and it requires more investigation about the price with different machines.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -71,28 +99,57 @@ library LibProposing {
...
+        blk.feePerGas = state.feePerGas;
+        blk.proposedAt = meta.timestamp;
 
-        uint64 blockFee;
+        if (assignedProver == address(0)) {
+            if (state.numOpenBlocks >= config.rewardOpenMaxCount) {
+                revert L1_TOO_MANY_OPEN_BLOCKS();
+            }
+            blk.rewardPerGas = state.feePerGas;
+            ++state.numOpenBlocks;
```
hugo-blue, line 122: Where is the code to open this block?

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -151,24 +172,76 @@ library LibVerifying {
...
+
+            state.feePerGas = uint32(
+                LibUtils.movingAverage({
+                    maValue: state.feePerGas,
+                    newValue: blk.rewardPerGas,
+                    maf: 7200
+                })
+            );
+        } else {
+            // proving out side of the proof window
```
hugo-blue, line 232: If the prover never returns the proof, how do we know it needs to open?

hugo-blue (commented):
I have some questions about this PR, and I may omit some of the logic though.

packages/protocol/contracts/L1/ProverPool.sol
hugo-blue: Though, it's not big issue.

dantaik: We assume eventually each prove may take ~1 hour.

packages/protocol/contracts/L1/ProverPool.sol
```diff
@@ -0,0 +1,364 @@
...
+    // then we need a min capacity of 3600, which means each prover shall
+    // provide a capacity of at least 3600/32=112.
+    uint32 public constant MAX_CAPACITY_LOWER_BOUND = 128;
+    uint64 public constant EXIT_PERIOD = 1 weeks;
+    uint32 public constant SLASH_POINTS = 500; // basis points
+    uint64 public constant MIN_STAKE_PER_CAPACITY = 10_000;
+    uint64 public constant MIN_SLASH_AMOUNT = 1e8; // 1 token
+    uint256 public constant MAX_NUM_PROVERS = 32;
+
+    // Reverse more slots than necessary
```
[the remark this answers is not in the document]
dantaik: Fixed

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: We can adjust the value to 90% or even lower but we do not know what's the best value, so now 95% is used to start testing it.

There is a feature not implemented yet: https://github.com/taikoxyz/taiko-mono/issues/14090

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: Right, we are not encouraging fast proofs now, which is a pending issue. Will work on it later. This is also mentioned in https://github.com/taikoxyz/taiko-mono/issues/14090.

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: If the `assignProver` returns `address(0)`, then the block is considered as an open block.

dantaik: Inside LibProposing:

```solidity
 if (assignedProver == address(0)) {
            if (state.numOpenBlocks >= config.rewardOpenMaxCount) {
                revert L1_TOO_MANY_OPEN_BLOCKS();
            }
            blk.rewardPerGas = state.feePerGas;
            ++state.numOpenBlocks;
        } 
```

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: If the assigned prover didn't submit a proof within the proving window, then anyone else can submit a proof.

hugo-blue:
> > I have some questions about this PR, and I may omit some of the logic though.
> 
> I'm ok to have a chat, @hugo-blue .

Sure, let's have a chat tomorrow.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -42,26 +49,47 @@ library LibProposing {
...
-        });
-
-        if (cacheTxListInfo != 0) {
-            state.txListInfo[input.txListHash] = TaikoData.TxListInfo({
-                validSince: uint64(block.timestamp),
-                size: uint24(txList.length)
+        // Try to select a prover first to revert as earlier as possible
+        (address assignedProver, uint32 rewardPerGas) = IProverPool(
+            resolver.resolve("prover_pool", false)
+        ).assignProver(state.numBlocks, state.feePerGas);
```
hugo-blue, line 55: To handle the max block speed, the prover has to set high capacity, which means large computation power should be prepared. And then, the cost might be high, so the prover needs to be able to adjust the computation power dynamically to save cost.

packages/protocol/contracts/L1/TaikoConfig.sol
```diff
@@ -11,38 +11,44 @@ import { TaikoData } from "../L1/TaikoData.sol";
...
-            // Set it to 79  (+1 TaikoL2.anchor transaction = 80),
-            // and 80 is the upper limit of the Alpha-2 testnet's circuits.
-            maxTransactionsPerBlock: 79,
+            blockMaxTxListBytes: 120_000,
+            blockTxListExpiry: 0,
+            // Group 3: proof related configs
+            proofRegularCooldown: 30 minutes,
+            proofOracleCooldown: 15 minutes,
+            proofMinWindow: 10 minutes,
+            proofMaxWindow: 90 minutes,
```
hugo-blue, line 40: The two parameters are related to circuit size. If this circuit is tiny, we don't have to set a min window of 10min. The max window is similar.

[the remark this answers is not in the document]
dantaik: The `proofMinWindow` will make sure we always allow certain slow proofs to be submitted. But we can tune this parameter value later. Giving it a very small value is equivalent of removing it.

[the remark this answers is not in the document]
hugo-blue: Assume A4 testnet will have a CPU perf of 400s, and the GPU might be less than 40s. It means, average proof time is 40s. Then 40*2=80s < 10 min, it will always use 10mins. Thus, in coming A4 testnet, the min window may be too big.

packages/protocol/contracts/L1/libs/LibProposing.sol
hugo-blue: anyway, it's not big issue, because the required staking token amount for each prover is bigger than 128*10000*10e-8 TTKO.

packages/protocol/contracts/L1/ProverPool.sol
```diff
@@ -0,0 +1,409 @@
...
+        Prover memory _prover;
+        unchecked {
+            for (uint32 i; i < MAX_NUM_PROVERS; ++i) {
+                _prover = provers[i + 1];
+                if (_prover.currentCapacity != 0) {
+                    // Keep the effective rewardPerGas in [75-125%] of feePerGas
+                    if (_prover.rewardPerGas > feePerGas * 125 / 100) {
+                        erpg[i] = feePerGas * 125 / 100;
+                    } else if (_prover.rewardPerGas < feePerGas * 75 / 100) {
+                        erpg[i] = feePerGas * 75 / 100;
```
hugo-blue, line 236: As we previously discussed, it may be preferable for the prover to set the rewardpergas to 1 in order to achieve the highest weight, but with a 75% reward. It's worth noting that the weight of rewardpergas 1 is four times that of reward gas 2 for the same staking amount. Perhaps we should consider establishing a permitted range for rewardpergas?

hugo-blue (commented):
some comments.

packages/protocol/contracts/L1/TaikoConfig.sol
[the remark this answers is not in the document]
hugo-blue: Based on the expected CPU performance of 400s for the A4 testnet, and assuming the GPU performance is less than 40s, the average proof time would be around 40s. This indicates that the minimum window for the upcoming A4 testnet may be too large as even 80s (which is double the average proof time) is less than 10 minutes.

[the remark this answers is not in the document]
hugo-blue: Based on the performance estimates for the A4 testnet, the CPU is expected to take around 400 seconds, while the GPU may take less than 40 seconds. This indicates that the average proof time will be 40 seconds. If we multiply this by 2, we get 80 seconds, which is less than 10 minutes. Therefore, the minimum window for the upcoming A4 testnet may be too large.

packages/protocol/contracts/L1/ProverPool.sol
hugo-blue: Oh, I see this:
 weights[i] = _calcWeight(_prover.stakedAmount, erpg[I]);

The input parameter is ergp[I], not original rewardpergas.
So, no issue here.

packages/protocol/contracts/L1/ProverPool.sol
```diff
@@ -0,0 +1,409 @@
...
+                    } else {
+                        erpg[i] = _prover.rewardPerGas;
+                    }
+                    weights[i] = _calcWeight(_prover.stakedAmount, erpg[i]);
+                }
```
hugo-blue, line 241: A proposal for this item:
"Another desired feature is to somehow encourage provers to submit proofs ASAP. This feature remains to be discussed and potential implemented."  https://github.com/taikoxyz/taiko-mono/issues/14090 

The final reward per gas is also related to the proof time. The faster the proof is provided, the higher the reward. This means that the weight is not related to proof time, but once a prover is selected, we encourage it to submit their proof as soon as possible in order to receive a greater reward.

dantaik: We'll work on this one later :)

packages/protocol/contracts/L1/TaikoConfig.sol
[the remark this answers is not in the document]
dantaik: Noted, but these values are still very temporary and we expect more circuits to be integrated, so sooner, it's going to be much bigger than 400 seconds.

dantaik:
@Brechtpd @cyberhorsey @davidtaikocha @adaki2004 This branch is ready to be reviewed again and merge to the main branch.

dantaik:
@jscriptcoder I disabled the check here for bridge-ui-v2, could you please take a look and fix it?

davidtaikocha (dismissed):

cyberhorsey:
> 

yeah we will switch to revive soon when i get a minute, golint still works for the time being. dosent need to be done in this PR imo, already large and the linting will be different so it can be its own PR.

dantaik:
@Brechtpd @davidtaikocha @adaki2004 could you please approve this PR again for merging before too many changes on the main branch?

davidtaikocha (dismissed):

cyberhorsey:
> @Brechtpd @davidtaikocha @adaki2004 could you please approve this PR again for merging before too many changes on the main branch?

I can approve, but should the protocol build pass first?

davidtaikocha (approved):

adaki2004 (approved):

PR #14067, opened 2023-06-28
feat(protocol): ProverPool2 design

adaki2004:
Core functions of ProverPool2 is ready for review.   Since protocol level interfaces are identical with ProverPool.sol (client differs a bit) the exact same unit/integration tests could be run for them and it is proven the main (protocol) aspects works the same as ProverPool.sol.

ProverPool:
kép

 ProverPool2: 
kép

 proposeBlock: saves 46K gas on average
proveBlock: saves 6K gas on average
verifyBlock: 0.4K gas on average.  

Tho - as stated previously - it has a different design approach and client facing interface!

   In order to speed up the ‘learning curve’ -  how it works under the hood, - please find a document attached with visual explanation (differences, advantages, drawbacks, etc.).   Click [here](https://docs.google.com/presentation/d/1GeyveCCm_RUPCxhJcYTaVln2ibdab8dpMCsccTSdyRk/edit#slide=id.p).

Please review the code thoroughly as well after checking the docs ! 
In case we decide to go with it, we need to tailor it to the code (like renames, etc.) and client needs as well. (Maybe some extra getters, (?) etc.) 

**NOTE**: A linked list would make it possible to keep everything exactly the same for the client, but not implemented currently to keep things simple for now it's not done that way. (+also not guaranteed it would be cheaper than this.)


cyberhorsey:
Gas savings look great. I already ahve working client implementation and I dont think this PR changes that at all (except tests where I need to stake in a different way) since we still only care about certain things. Seems this PR also has the `block.assignedProver being 0 address means anyone can prove it` logic, and the same `proofWindow` calculation checks?

dantaik:
Could you share your analysis and comparison between this one and the other one here in this PR instead of in the Google Doc?

dantaik:
@adaki2004 if you are OK, I'd like to close this PR.

PR #14068, opened 2023-06-28
feat(website): RPC connect Metamask Networks

MarcusWentz:

PR #14070, opened 2023-06-28
feat(website): RPC connect Metamask Networks

MarcusWentz:
Modified:
https://github.com/taikoxyz/taiko-mono/pull/14068
to not have the Metamask pnpm package to handle the
```js
window.ethereum
``` 
object.

Instead
```js
(window as any).ethereum
```
is used to keep things simple for package dependencies.


packages/website/components/MetamaskNetworkButton/ConnectToMetamask.tsx
```diff
@@ -0,0 +1,87 @@
...
+    chainId: "0x28c5d",
+    chainName: "Taiko (Alpha-3 Testnet)",
+    nativeCurrency: {
+      name: "ETH",
+      symbol: "eth",
+      decimals: 18,
+    },
+    rpcUrls: ["https://rpc.test.taiko.xyz"],
+    blockExplorerUrls: ["https://explorer.test.taiko.xyz/"],
+    iconUrls: [],
```
d1onys1us, line 66: can we make sure it matches this exactly: https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-167005.json

otherwise, that alert error will through. i already added taiko from chainlist, and if i click it, it says the symbol is mismatched (i think because of eth vs ETH)

packages/website/components/MetamaskNetworkButton/ConnectToMetamask.tsx
```diff
@@ -0,0 +1,87 @@
...
+  }
+  if ((window as any).ethereum.networkVersion == "11155111") {
+    alert("You are already connected to Sepolia (chainId 11155111).", )
+    return;
+  }
+  try{
+    await (window as any).ethereum.request({
+      method: "wallet_switchEthereumChain",
+      params: [{
+         chainId: "0xaa36a7"
```
d1onys1us, line 31: curious will this work if they have not enabled the test networks in metamask?

d1onys1us:
final comment. i see this was added to RPC configuration, which is great. however, i can't help but to feel, is RPC configuration a duplicate of: https://website-lyc4taqy3-taikoxyz.vercel.app/docs/guides/configure-your-wallet ? Should we add it to configure your wallet as well or somehow consolidate the two? Should configure your wallet guide just be a link to RPC configuration? i think any of these are fine, at the least maybe we just add it to configure-your-wallet

packages/website/components/MetamaskNetworkButton/ConnectToMetamask.tsx
MarcusWentz: You switch networks instead of adding them if they already exist in Metamask by default for security purposes:
https://ethereum.stackexchange.com/questions/127499/metamask-not-switching-chain-to-rinkeby-testnet

MarcusWentz: This also works in Metamask when testnets like Sepolia are hidden based on a test I did now.

packages/website/components/MetamaskNetworkButton/ConnectToMetamask.tsx
MarcusWentz: Updated currency name and symbol.

With 
```js
chainId: "167005",
```
I see error
```js
Expected 0x-prefixed, unpadded, non-zero hexadecimal string 'chainId'. Received: 167005
```
therefore we should keep the chainId in hexadecimal format
```js
chainId: "0x28c5d",
```

MarcusWentz:
I think we should redirect the wallet connection page to the RPC connection page based on links on the wallet connection page.

dantaik (dismissed):

dionysuzx (approved):

dantaik (approved):

PR #14117, opened 2023-07-06
feat(bridge-ui-v2): switch chain

jscriptcoder:
Switching chains and updating state


dantaik:
Is ERC1155 an superset of ERC721? If so, maybe for all ERC721 tokens, we just treat them as ERC1155?

packages/bridge-ui-v2/src/components/ChainSelector/ChainSelector.svelte
```diff
@@ -1,15 +1,20 @@
...
+  import { warningToast } from '$components/NotificationToast';
   import { PUBLIC_L1_CHAIN_ID, PUBLIC_L2_CHAIN_ID } from '$env/static/public';
   import { chains } from '$libs/chain';
   import { uid } from '$libs/util/uid';
+  import { account } from '$stores/account';
+
+  // TODO: think about updating the state for the network here
 
   export let label: string;
+  export let value: Maybe<GetNetworkResult['chain']> = null;
```
d1onys1us, line 17: nit: can we do a better name than `value`? perhaps `selectedChain`

dionysuzx (dismissed):

dantaik (dismissed):

packages/bridge-ui-v2/src/components/ChainSelector/ChainSelector.svelte
jscriptcoder: We could, but this is the value of this control, same as Input, same as Select, same as Textarea, etc... I find it a bit redundant to use `selectedChain` in a ChainSelected component:
`<ChainSelector selectedChain={chain} />` vs `<ChainSelector value={chain} />`

davidtaikocha (approved):

dantaik (approved):

PR #14118, opened 2023-07-06
feat(protocol): NFT (ERC721/ERC1155) bridge

adaki2004:
Create a trustless NFT (ERC721 first, next 1155) bridge with a similar API as the regular bridge.


packages/website/pages/docs/reference/contract-documentation/nft_bridge/Erc721Bridge.md
```diff
@@ -0,0 +1,245 @@
+---
+title: Erc721Bridge
+---
+
+## Erc721Bridge
```
dantaik, line 5: For now, lets use gitignore to ignore all docs generated for these new contracts, then we add them later once and for all.

packages/protocol/contracts/nft_bridge/Erc721Bridge.sol
```diff
@@ -0,0 +1,283 @@
...
+import { LibErc721BridgeSend } from "./erc721/libs/LibErc721BridgeSend.sol";
+import { LibErc721BridgeStatus } from "./erc721/libs/LibErc721BridgeStatus.sol";
+
+/**
+ * This contract is an ERC-721 token bridge contract which is deployed on both L1 and L2.
+ * which calls the library implementations. See _IErc721Bridge_ for more details.
+ * @dev The code hash for the same address on L1 and L2 may be different.
+ * @custom:security-contact hello@taiko.xyz
+ */
+contract Erc721Bridge is EssentialContract, IErc721Bridge, NftBridgeErrors {
```
dantaik, line 26: There should just be one Bridge, Bridge is a common infrastructure for messaging, it's agnostic to message types.

packages/protocol/contracts/nft_bridge/erc721/IErc721Bridge.sol
```diff
@@ -0,0 +1,105 @@
...
+
+pragma solidity ^0.8.20;
+
+/**
+ * Bridge interface for NFT contracts.
+ * @dev Ether is held by Bridges on L1 and by the EtherVault on L2,
+ * not TokenVaults.
+ */
+interface IErc721Bridge {
+    struct Message {
```
dantaik, line 15: In Bridge.sol, the Message is generic, it's not designed for ERC20, and we shall not design a special Message for NFTs either.

packages/protocol/contracts/bridge/BridgeErrors.sol
```diff
@@ -10,6 +10,12 @@ abstract contract BridgeErrors {
     /// Emitted when the contract is not intended to receive Ether
     error B_CANNOT_RECEIVE();
 
+    /// Emitted when the contract is not intended to receive ERC721
+    error B_ERC721_CANNOT_RECEIVE();
```
dantaik, line 14: Dani, I think we can just use B_CANNOT_RECEIVE

packages/protocol/contracts/bridge/NFTVault.sol
```diff
@@ -0,0 +1,326 @@
...
+import {IBridge} from "./IBridge.sol";
+import {LibERC721} from "./erc721/libs/LibERC721.sol";
+import {LibERC1155} from "./erc1155/libs/LibERC1155.sol";
+
+/**
+ * This vault holds all ERC721 and ERC1155 tokens that users have deposited.
+ * It also manages the mapping between canonical ERC721/1155 tokens and their bridged
+ * tokens.
+ */
+contract NFTVault is
```
dantaik, line 29: Is it better and simpler to have separate ERC721Vault and ERC1155Vault?

packages/protocol/contracts/bridge/erc721/libs/LibERC721.sol
```diff
@@ -0,0 +1,208 @@
...
+        uint256 srcChainId,
+        address token,
+        uint256 tokenId
+    );
+
+    error NFTVAULT_CANONICAL_TOKEN_NOT_FOUND();
+    error NFTVAULT_INVALID_OWNER();
+    error NFTVAULT_INVALID_SENDER();
+
+    function sendErc721(
```
dantaik, line 63: We have `sendErc721`, `sendErc1155`, `sendERC20`, I think we can simply call them the same name `sendToken`...  same for `releaseToken`, `receiveToken`.

packages/protocol/script/upgrade/UpgradeTokenVault.s.sol
```diff
@@ -11,16 +11,16 @@ import "forge-std/console2.sol";
 import
     "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 import "@openzeppelin/contracts-upgradeable/utils/math/SafeCastUpgradeable.sol";
-import "../../contracts/bridge/TokenVault.sol";
+import "../../contracts/bridge/erc20/ERC20Vault.sol";
 import "./UpgradeScript.s.sol";
 
-contract UpgradeTokenVault is UpgradeScript {
+contract UpgradeERC20Vault is UpgradeScript {
```
dantaik, line 17: rename the file to `UpgradeERC20Vault.sol` please

packages/protocol/contracts/bridge/erc721/libs/LibERC721.sol
```diff
@@ -0,0 +1,208 @@
...
+} from "@openzeppelin/contracts-upgradeable/utils/Create2Upgradeable.sol";
+import {
+    ERC721Upgradeable
+} from "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721Upgradeable.sol";
+import {NFTVault} from "../../NFTVault.sol";
+import {BridgedERC721} from "../BridgedERC721.sol";
+import {IBridge} from "../../IBridge.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+
+library LibERC721 {
```
dantaik, line 21: Why we have to have a library for ERC721? Otherwise some contracts are too large?

packages/protocol/contracts/bridge/NFTVault.sol
```diff
@@ -0,0 +1,326 @@
...
+     struct CanonicalNFT {
+        uint256 srcChainId;
+        address tokenAddr;
+        string symbol;
+        string name;
+        string uri;
+        NFTType nftType;
+     }
+
+    struct SendNFTOpts {
```
dantaik, line 59: I think for ERC20, ERC712, and ERC1155, we should define different structs but we can name the all the same `BridgeTransferOp` (change ERC20Vault's `MessageDeposit` to it as well).

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -38,7 +38,7 @@ contract Bridge is EssentialContract, IBridge, BridgeErrors {
 
     event DestChainEnabled(uint256 indexed chainId, bool enabled);
 
-    /// Allow Bridge to receive ETH from the TaikoL1, TokenVault or EtherVault.
```
dantaik, line 41: Please search for "token_vault" and replace them with "erc20_vault". 
Also, in this file, you will find:

```solidity
 /// Allow Bridge to receive ETH from the TaikoL1, ERC20Vault or EtherVault.
    receive() external payable {
        if (
            msg.sender != resolve("token_vault", true)
                && msg.sender != resolve("ether_vault", true)
                && msg.sender != resolve("taiko", true) && msg.sender != owner()
        ) {
            revert B_CANNOT_RECEIVE();
        }
    }

```

You need to add "erc721_vault", "erc1155_vault" to the list.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -38,11 +38,13 @@ contract Bridge is EssentialContract, IBridge, BridgeErrors {
...
+    /// Allow Bridge to receive ETH from the TaikoL1, ERC20Vault, EtherVault or any of the NFT vaults.
     receive() external payable {
         if (
-            msg.sender != resolve("token_vault", true)
+            msg.sender != resolve("erc20_vault", true)
```
dantaik, line 44: After a second thought, we may simply just remove this method...

packages/protocol/contracts/bridge/BridgeErrors.sol
```diff
@@ -73,4 +73,36 @@ abstract contract BridgeErrors {
...
     /// Emitted when a signal of zero is used
     error B_ZERO_SIGNAL();
+
+    // NFTVault related errors
+
+    /// Emitted when a canonical token counterpart is not found for the bridged one
+    error NFTVAULT_CANONICAL_TOKEN_NOT_FOUND();
+
+    /// Emitted when opts.amount is 0
+    error NFTVAULT_INVALID_AMOUNT();
```
dantaik, line 83: Bridge is independent from NFT/ERC20 vaults, We should move all Vault errors to vaults.

packages/protocol/contracts/bridge/erc1155/ERC1155Vault.sol
```diff
@@ -0,0 +1,207 @@
...
+        uint256 amount;
+        (nft, owner,,tokenId, amount) = 
+            LibERC1155.decodeTokenData(message.data);
+
+        IBridge bridge = IBridge(resolve("bridge", false));
+        bytes32 msgHash = bridge.hashMessage(message);
+
+        if (nft.tokenAddr == address(0)) revert NFTVAULT_INVALID_TOKEN();
+
+        if (!bridge.isMessageFailed(msgHash, message.destChainId, proof)) {
```
dantaik, line 155: Do you find common logics in this method and the same method in other vaults?

packages/protocol/contracts/bridge/erc1155/libs/LibERC1155.sol
```diff
@@ -0,0 +1,224 @@
...
+import {
+    ERC1155Upgradeable
+} from "@openzeppelin/contracts-upgradeable/token/ERC1155/ERC1155Upgradeable.sol";
+import {NFTVaultParent} from "../../NFTVaultParent.sol";
+import {BridgedERC1155} from "../BridgedERC1155.sol";
+import {IBridge} from "../../IBridge.sol";
+import {LibExtractCalldata} from "../../libs/LibExtractCalldata.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+
+library LibERC1155 {
```
dantaik, line 22: Is it possible that you don't have this lib and merge the code into the Vault?

packages/protocol/contracts/bridge/erc721/ERC721Vault.sol
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /**
+     * Transfers ERC721 tokens to this vault and sends a message to the
+     * destination chain so the user can receive the same (bridged) tokens
+     * by invoking the message call.
+     *
+     * @param opts Options for sending the ERC721/ERC1155 token.
+     */
+    function sendToken(BridgeTransferOp calldata opts) external payable nonReentrant {
```
dantaik, line 46: opts => op?

packages/protocol/contracts/bridge/erc721/libs/LibERC721.sol
```diff
@@ -0,0 +1,227 @@
...
+        AddressResolver resolver,
+        address addressManager,
+        NFTVaultParent.CanonicalNFT memory canonicalToken,
+        address from,
+        address to,
+        uint256 tokenId,
+        address canonicalToBridged
+    ) public returns (bool bridged, address token) {
+        IBridge.Context memory ctx = IBridge(msg.sender).context();
+        if (ctx.sender != resolver.resolve(ctx.srcChainId, "erc721_vault", false))
```
dantaik, line 126: Line 125 to 127 are common logics to all vaults, we should move it to the base class (parent). Using the lib won't allow us to do it, I suggest that we remove the Libs.

dantaik (changes requested):

packages/protocol/contracts/bridge/erc20/ERC20Vault.sol
```diff
@@ -199,7 +192,7 @@ contract TokenVault is EssentialContract {
      * @param memo Any additional data or notes
      */
 
-    function sendERC20(
+    function sendToken(
```
dantaik, line 195: Shall we also define a BridgeTransferOp here?

packages/protocol/contracts/bridge/erc721/ERC721Vault.sol
```diff
@@ -0,0 +1,379 @@
...
+        if (
+            opt.to == address(0) ||
+            opt.to == resolve(opt.destChainId, "erc721_vault", false)
+        ) revert NFTVAULT_INVALID_TO();
+
+        if (opt.token == address(0)) revert NFTVAULT_INVALID_TOKEN();
+
+        if (opt.amount != 1) revert NFTVAULT_INVALID_AMOUNT();
+
+        bytes memory data = _sendToken(
```
dantaik, line 101: `message.data  = _sendToken(`

packages/protocol/contracts/bridge/erc721/ERC721Vault.sol
```diff
@@ -0,0 +1,379 @@
...
+
+    /**
+     * Transfers ERC721 tokens to this vault and sends a message to the
+     * destination chain so the user can receive the same (bridged) tokens
+     * by invoking the message call.
+     *
+     * @param opt Option for sending the ERC721/ERC1155 token.
+     */
+    function sendToken(BridgeTransferOp calldata opt) external payable nonReentrant {
+        if (
```
dantaik, line 92: Line 92 to 95 can be extracted as a modifier

packages/protocol/contracts/bridge/erc721/ERC721Vault.sol
```diff
@@ -0,0 +1,379 @@
...
+                tokenId,
+                bridgedAddress
+            );
+
+        if (bridged) {
+            setBridgedToken(bridgedToken, canonicalToken);
+        }
+    }
+
+    function releaseToken(
```
dantaik, line 172: I noticed you don't have a `messageDeposits` defined in this contract, will this releaseToken function work without it? Please read ERC20 vault.

dantaik (changes requested):

packages/protocol/contracts/tokenvault/BridgedERC721.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    /// @dev Initializer to be called after being deployed behind a proxy.
+    // Intention is for a different BridgedERC721 Contract to be deployed
+    // per unique _srcToken i.e. one for Loopheads, one for CryptoPunks, etc.
+    function init(
+        address _addressManager,
+        address _srcToken,
+        uint256 _srcChainId,
+        string memory _symbol,
+        string memory _name
```
cyberhorsey, line 32: I really think we need the URI, ignoring the URI goes completely against "Ethereum Equivilance" and anyone who deploys an NFT marketplace/etc on Taiko will have an extra hurdle to support Bridged vs Native tokens. Not sure what the intent here is @dantaik @dantaik I'm sure this conversation has already been hashed out but to me it seems like ignoring the decimals in ERC20.

cyberhorsey (changes requested):

packages/protocol/contracts/bridge/Bridge.sol
dantaik: oh, then simply do not check for msg.senders and have a empty receive()

packages/protocol/contracts/tokenvault/BridgedERC721.sol
dantaik: I have two reason not to support it:
- it's more expensive. Usually in an ERC721 token contract, the per-token URL is `baseURL + tokenId.toString()`, but the baseURI is not exposed through an interface function, therefore, we have to query each tokenID's URL and write them in the Bridged contract, so we need one URL write per tokenID.

- the URI may change and it will create data discrepancy.  This will probably more problematic.

dantaik: Lets do it when the UI is absolutely need the URIs on the dest chain.

PR #14160, opened 2023-07-13
test(bridge-ui-v2): increase coverage

jscriptcoder:
- Mocking packages and others
- Increasing coverage


dantaik (approved):

packages/bridge-ui-v2/src/libs/chain/chains.ts
```diff
@@ -21,7 +21,7 @@ import {
 } from '$env/static/public';
 
 export const mainnetChain: Chain = {
-  id: parseInt(PUBLIC_L1_CHAIN_ID),
+  id: +PUBLIC_L1_CHAIN_ID,
```
KorbinianK, line 24: Everytime I see this, it makes me love and hate Javascript at the same time :D

KorbinianK (approved):

packages/bridge-ui-v2/src/libs/chain/chains.ts
jscriptcoder: If you prefer I can remove this magic. Lucky you I haven't yet use double tilde `~~` 😉

I was parsing the int, but since we get this from an env var as a string, and `Chain` from `@wagmi/core` type requires the chain.id to be int, I decided to make it quicker by converting using this trick

PR #14164, opened 2023-07-14
feat(protocol): apply slight reward penalty based on timing

dantaik:
Introduce a new config rewardMaxDelayPenalty, if it is set to 250 bps (2.5%), then when the proof is submitted with a delay == provingWindow, then the reward to the prover is 97.5%*origianlProofReward, if the proof is submitted with a 0 delay, then the reward to prover is 100%*origianlProofReward.

This will encourage assigned provers to submit proofs ASAP.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibVerifying.sol
+++ b/packages/protocol/contracts/L1/libs/LibVerifying.sol
@@ -222,7 +222,7 @@ library LibVerifying {
             proofReward = proofReward * config.rewardOpenMultipler / 100;
             proverPool.slashProver(blk.assignedProver);
         } else if (fc.provenAt <= blk.proposedAt + blk.proofWindow) {
-            // proving inside the window
+            // proving inside the window, by the assigned prover
             uint64 proofDelay;
             unchecked {
                 proofDelay = fc.provenAt - blk.proposedAt;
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -219,13 +220,23 @@ library LibVerifying {
...
+            uint64 proofDelay;
+            unchecked {
+                proofDelay = fc.provenAt - blk.proposedAt;
+
+                if (config.rewardMaxDelayPenalty > 0) {
+                    // Give the reward a penalty up to a small percentage.
+                    // This will encourage prover to submit proof ASAP.
+                    proofReward -= proofReward * proofDelay
+                        * config.rewardMaxDelayPenalty / 10_000 / blk.proofWindow;
+                }
```
cyberhorsey, line 232: Should we have a larger penalty (or no reward, sent to Treasury or burnt?) if the proof was submitted after the proof window expired, or is slashing tokens enough? Because I noticed prover still gets reward for block, even if they are late, and also they get the rewardOpenMultiplier so actually it's a little bonus for being late (aside from slashing).

dantaik: Changed the code based on your suggestion. Please take another look.

cyberhorsey: This makes more sense to me.

cyberhorsey (approved):

davidtaikocha (approved):

PR #14183, opened 2023-07-17
fix(protocol): fix issues in NFT vaults

dantaik:


adaki2004 (approved):
Just a note/remark on non-existing tokenURI():
- It will make the bridged assets just 'plain numbers' and noone - except the Taiko Bridge contract - knows what those assets mean
- not possible to access the visuals (traits) automatically (e.g.: Rarible/Blur/Opensea cannot be able to query tokenURI() and show the images) 

Independently, i'll accept this PR and merge it into the base. In case Daniel changes his decision about the direction, there is an existing [PR](https://github.com/taikoxyz/taiko-mono/pull/14188) - which bridging the tokenURIs.

PR #14188, opened 2023-07-17
feat(protocol): Implement token uri differently

adaki2004:
@dantaik Please see. This is the Nr.3 option i was talking about. (Basically have a direct mapping and use it over on the 'bridged' chain).

Feel free to merge it if you like it. Nr. 1 and Nr.3 both could fit, but with this we eliminate the need of having to input by the FrontEnd a base token uri.

### DIFF
--- a/packages/protocol/contracts/tokenvault/BridgedERC1155.sol
+++ b/packages/protocol/contracts/tokenvault/BridgedERC1155.sol
@@ -26,11 +26,13 @@ contract BridgedERC1155 is
 {
     address public srcToken;
     uint256 public srcChainId;
-    string public srcUri;
     string public name;
     string public symbol;
 
-    uint256[47] private __gap;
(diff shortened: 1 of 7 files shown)


packages/protocol/contracts/tokenvault/BridgedERC1155.sol
```diff
@@ -77,12 +77,14 @@ contract BridgedERC1155 is
...
         uint256 tokenId,
-        uint256 amount
+        uint256 amount,
+        string memory tokenUri
     )
         public
         onlyFromNamed("erc1155_vault")
     {
         _mint(account, tokenId, amount, "");
+        _uris[tokenId] = tokenUri;
```
dantaik, line 87: Dani, I still prefer not supporting URL at all. Firs of all, the data should be available on the source chain, secondly, this will make bridging cheaper and consistent.

PR #14199, opened 2023-07-18
feat(protocol): Alpha 4 fix inception bridge embedded naming

adaki2004:
Currently, we embedded the previous layer's name into the bridged asset, tho it would create an 'ever-embedding' effect seen in the screenshot below.

kép

This PR aim's to fix it with the following way:

Assume we have 2 Networks, `A` and **B.**

`A:` is a L2 with a settlement layer (TaikoL1 contracts) on L1.
`B:` is a L3 with a settlement layer (TaikoL1 contracts) on L2.

It would require 4 TokenVault deployments.
1-1 for network A on both chain.
1-1 for network B on both chains

On Layer2, there will be 2 deployments. 

With this solution they can communicate with each other, by calling `_checkIfInceptedBridging` if `B`'s TokenVault (on L2) knows the other TokenVault (which is the destination token vault of network `A`).

SO how it works (technically):

When someone wants to bridge a token from L2 to L3, during sendErc20() it invokes _checkIfInceptedBridging(). It will query the other (registered) TokenVault instance (which is the destination TokenVault of network A) - if it is indeed an already bridged token, it passes a boolean flag within the `message.data` towards L3.

### DIFF
--- a/packages/protocol/contracts/bridge/BridgedERC20.sol
+++ b/packages/protocol/contracts/bridge/BridgedERC20.sol
@@ -33,7 +33,8 @@ contract BridgedERC20 is
     address public srcToken;
     uint256 public srcChainId;
     uint8 private srcDecimals;
-    uint256[47] private __gap;
+    bool private inceptionBridged;
+    uint256[46] private __gap;
 
     /**
      * Initializes the contract.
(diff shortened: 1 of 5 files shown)


dantaik:
@adaki2004 I reviewed the code, it seems we do not have this bug any more in our current codebase. The reason we are seeing this is that we had a small bug in alpha-3. No need to fix it.

PR #14211, opened 2023-07-19
feat(bridge-ui-v2): NFT bridge UI

KorbinianK:
Initial barebones setup for the NFT bridge UI. 

Split the bridge components, prepared check for contract type and fetching of images


jscriptcoder:
In terms of routing, no need to create deeper path, why not to simply:
`/` => general bridge (it's not only erc20)
`/nft` => nft bridge?

Same regarding components, and if the NFT one is reusing something from the other, we simply move those components to the root level, and state is passed as props.

dantaik (dismissed):

PR #14221, opened 2023-07-20
feat(protocol): decouple proof(s) verification from core protocol

adaki2004:
This PR separates the proof verification logic.
SGX and other support can come late, and core protocol only sees `bytes`, no `verifierId`.


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -206,19 +207,15 @@ library LibProving {
...
-                //
-                || bytes32(ret) != keccak256("taiko")
-            ) {
-                revert L1_INVALID_PROOF();
-            }
+            // Later on it can be like verifyProofs(bytes) where the bytes would
+            // include all of the proofs (ZK, SGX, Oracle, etc. ).
+            IProofVerifier(resolver.resolve("proof_verifier", false)).verifyProofs(
+                abi.encode(
+                    evidence.verifierId,
```
dantaik, line 214: we should use delete `evidence.verifierId` and `evidence.proof` from evidence and add a new field called `bytes proofs`. By doing this, the protocol will be agnostic to verifierId which is ZK related. Then you just send `proofs` to the prover contract. 

In side the prover contact, you decode `proofs` into a struct and then further use the `verifiedId` to dispatch zkProofs to zkVerifier.

dantaik: In core protocol, no ZK proof info should be exposed.

dantaik: Anything else you want to know? I thought it's pretty clear: making sure the core protocol does not know anything about the content of a block proof and how they shall be verified. Move these logics to the Verifier.

dantaik:
For the core protocol, the proof shall be just a byte array, nothing else. Then the IProofVerifier contract should look like:

```solidity
function verifyBlockProof(uint blockId, bytes calldata blockProof) external;
```

Later we can upgrade the implementation of IProofVerifier and use blockId to perform different verification tasks.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -186,9 +187,12 @@ library LibProving {
                 }
             }
 
+            // Formal check of the instance equality. Shifted with
+            // 2 bytes because of the verifierId is also part of
+            // the evidence.proofs on a uint16 (2 bytes)
             if (
                 !LibBytesUtils.equal(
-                    LibBytesUtils.slice(evidence.proof, 0, 32),
+                    LibBytesUtils.slice(evidence.proofs, 2, 32),
```
dantaik, line 195: Can we also make `instance` one input variable to `verifyProofs`, then this logic can also be moved to the Verifier. `verifierId` should not be known here at all.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -197,7 +201,7 @@ library LibProving {
 
             if (
                 !LibBytesUtils.equal(
-                    LibBytesUtils.slice(evidence.proof, 32, 32),
+                    LibBytesUtils.slice(evidence.proofs, 34, 32),
```
dantaik, line 204: This slice operation also assumes some knowledge of the `proofs` structure. We should move it to the Verifier.

dantaik (changes requested):

dantaik:
Please also see https://github.com/taikoxyz/taiko-mono/pull/14221/commits/6e2fb7b23f192a787b7e5f77b7dc98f04a00a246

dantaik:
> > For the core protocol, the proof shall be just a byte array, nothing else. Then the IProofVerifier contract should look like:
> 
> @dantaik Should not we pass on the instance - within the params - as well? Because i guess this code section also needs to be moved into the ProofVerifier:
> 
> ```
>  if (
>                !LibBytesUtils.equal(
>                    LibBytesUtils.slice(evidence.proof, 0, 32),
>                    bytes.concat(bytes16(0), bytes16(instance))
>                )
>            ) {
>                revert L1_INVALID_PROOF();
>            }
> 
>            if (
>                !LibBytesUtils.equal(
>                    LibBytesUtils.slice(evidence.proof, 32, 32),
>                    bytes.concat(
>                        bytes16(0), bytes16(uint128(uint256(instance)))
>                    )
>                )
>            ) {
>                revert L1_INVALID_PROOF();
>            }
> ```

The idea is that the core protocol should know nothing about the construction of "evidence.proofs" at all.

dantaik (approved):

davidtaikocha (approved):

PR #14225, opened 2023-07-20
feat(bridge-ui-v2): bridging ETH and ERC20

jscriptcoder:
- Bridging in place
- Recipient component
- Error handling


dantaik (dismissed):

packages/bridge-ui-v2/src/components/Bridge/Recipient.svelte
```diff
@@ -0,0 +1,123 @@
+<script lang="ts">
```
KorbinianK, line 1: I saw you moved a couple of components out of their folders, but left others. Is there a reason? Now it seems a bit inconsistent, some have their export through an `index.ts`, some don't. Some are in a folder, some are not. I think we should be consistent.

packages/bridge-ui-v2/src/libs/bridge/checkBalanceToBridge.ts
```diff
@@ -49,9 +52,14 @@ export async function checkBalanceToBridge({
     } catch (err) {
       console.error(err);
 
-      if (`${err}`.includes('transaction exceeds the balance of the account')) {
+      // TODO: rely on error code, or instance, instead of string matching
```
KorbinianK, line 55: General question, do we have a backlog of these todos? If it's something important we should create an issue for it and add the issue # into the todo. Let's talk about it tomorrow and discuss this. :)

packages/bridge-ui-v2/src/components/Bridge/Bridge.svelte
```diff
@@ -35,35 +49,205 @@
...
+        to: $recipientAddress || $account.address,
+        wallet: walletClient,
+        srcChainId: $network.id,
+        destChainId: $destNetwork.id,
+        amount: $enteredAmount,
+        processingFee: $processingFee,
+      } as BridgeArgs;
+
+      if (isETH($selectedToken)) {
+        bridge = bridges.ETH as ETHBridge;
```
cyberhorsey, line 140: this kind of logic looks like its getting a bit hairy, should this not be a reactive statement above, and Token simply have a domain `TokenType` property?


and the domain token type can be of type `ETH, ERC721, ERC1155, ERC20`?

```ts
$ : bridge = (token) => {
    return tokenTypeToBridgeType(token.tokenType);
}($selectedToken)

function tokenTypeToBridgeType(t: TokenType) {
   switch(t) {
      case TokenType.ETH:
      return bridges.ETH;
     case TokenType.ERC20:
     return bridges.ERC20;
    case TokenType.ERC721:
    return bridges.ERC721;
     case TokenType.ERC1155:
   return bridges.ERC1155
   }
}
```

and then similarily:

```ts
function tokenTypeToBridgeArgs(t: TokenType) {
  let args: BridgeArgs =  {
        to: $recipientAddress || $account.address,
        wallet: walletClient,
        srcChainId: $network.id,
        destChainId: $destNetwork.id,
        amount: $enteredAmount,
        processingFee: $processingFee,
      };
   switch(t) {
    case TokenType.ETH:
      return { ...bridgeArgs, bridgeAddress };
     case TokenType.ERC20:
     return  const tokenAddress = await getAddress({
          token: $selectedToken,
          srcChainId: $network.id,
          destChainId: $destNetwork.id,
        });
        if (!tokenAddress) {
          throw new Error('token address not found');
        }
        const tokenVaultAddress = chainContractsMap[$network.id].tokenVaultAddress;
        const isTokenAlreadyDeployed = await isDeployedCrossChain({
          token: $selectedToken,
          srcChainId: $network.id,
          destChainId: $destNetwork.id,
        });
        bridgeArgs = {
          ...bridgeArgs,
          tokenAddress,
          tokenVaultAddress,
          isTokenAlreadyDeployed,
        };
        
      // etc, for erc721/1155
   }
}
```

and then your in-form bridge function becomes, without changing for additioonal future bridge functions:

```ts
     bridge.bridge(tokenTypeToBridgeArgs($selectedToken.tokenType));
```

and does not change for any future iterations of ERC1155,ERC721, or anything else we add in the future, the logic is encapsulated within domain helpers.

packages/bridge-ui-v2/src/components/Bridge/Recipient.svelte
```diff
@@ -0,0 +1,123 @@
...
+  import { shortenAddress } from '$libs/util/shortenAddress';
+  import { uid } from '$libs/util/uid';
+  import { account } from '$stores/account';
+
+  import { recipientAddress } from './state';
+
+  let dialogId = `dialog-${uid()}`;
+
+  let modalOpen = false;
+  let invalidAddress = false;
```
cyberhorsey, line 18: Invalid address is unused

packages/bridge-ui-v2/src/components/Bridge/Recipient.svelte
jscriptcoder: Which components do you refer to?

packages/bridge-ui-v2/src/components/Bridge/Recipient.svelte
jscriptcoder: It's unused at the moment. I'm waiting for design for this

packages/bridge-ui-v2/src/libs/bridge/checkBalanceToBridge.ts
jscriptcoder: At the moment these TODOs live in the code because I'm the only one working on this codebase. The plan is actually to create tickets out of them as you mentioned

packages/bridge-ui-v2/src/components/Bridge/Bridge.svelte
jscriptcoder: Agree, was a temporal approach  when I started until I figure out NFT part, then I never changed it. I'll include the type in to token.

packages/bridge-ui-v2/src/components/Bridge/Recipient.svelte
KorbinianK: Let's sync offline

packages/bridge-ui-v2/src/components/Bridge/Actions.svelte
```diff
@@ -0,0 +1,96 @@
...
+  {/if}
+
+  <Button
+    type="primary"
+    class="px-[28px] py-[14px] rounded-full flex-1"
+    disabled={disableBridge}
+    loading={bridging}
+    on:click={onBridgeClick}>
+    {#if bridging}
+      <span class="body-bold">{$t('bridge.button.bridging')}</span>
```
KorbinianK, line 91: Nice: Would be cool to see why the button is disabled, e.g. "insufficient funds". We have all the states for it anyway.

packages/bridge-ui-v2/src/components/Bridge/Bridge.svelte
jscriptcoder: Change made. Token contains now the type

dantaik (dismissed):

packages/bridge-ui-v2/src/components/Bridge/Actions.svelte
jscriptcoder: It does, it says that under the entered amount, which is responsible for it. There are a few reasons why the button can be disabled:
1. Token requires allowance (only for ERC20), in which case this is informed to the user close to what's responsible, in this case the entered amount.
2. Entered amount is too high. The user has not enough balance to cover the cost of bridging. User is informed about this, same as previous one.
3. Base conditions are not  fulfilled: no token or chains has been selected, etc...

KorbinianK: What I meant display the reason inside the button text, I saw it was disabled when I switched to an account with 0 eth connected but it didn't specify why. I guess this is an edge case though as it did not re-trigger the input validation for amount that would display the correct message. Button was disabled, that's good enough.

KorbinianK (dismissed):

packages/bridge-ui-v2/src/libs/bridge/ERC20Bridge.ts
```diff
@@ -60,8 +60,106 @@ export class ERC20Bridge implements Bridge {
 
     const value = processingFee;
 
-    log('Estimating gas for sendERC20 call. Sending value', value);
+    log('Estimating gas for sendERC20 call with value', value);
+
+    const estimatedGas = tokenVaultContract.estimateGas.sendERC20([...sendERC20Args], { value });
```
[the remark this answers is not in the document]
jscriptcoder: It's the processing fee, but it's been renamed

packages/bridge-ui-v2/src/components/Bridge/Actions.svelte
jscriptcoder: Yeah, we can ask Jane about this to include it in the design

dantaik (approved):

KorbinianK (approved):

PR #14230, opened 2023-07-21
feat(protocol): Reduce nft bridge code size

adaki2004:
On ERC721_bridging branch the build CICD step is switched off, but contract code sizes were above the limit.

Before PR:
kép

After PR:
kép


packages/protocol/contracts/tokenvault/BaseVault.sol
```diff
@@ -7,10 +7,7 @@
 pragma solidity ^0.8.20;
 
 import { EssentialContract } from "../common/EssentialContract.sol";
-import { IBridge } from "../bridge/IBridge.sol";
 import { AddressResolver } from "../common/AddressResolver.sol";
-import
-    "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 
 abstract contract BaseVault is EssentialContract {
```
dantaik, line 12: Please remove this contract and use `EssentialContract` to replace `BaseVault` elsewhere.

packages/protocol/contracts/tokenvault/libs/LibVaultUtils.sol
```diff
@@ -0,0 +1,108 @@
...
+import { IBridge } from "../../bridge/IBridge.sol";
+import { AddressResolver } from "../../common/AddressResolver.sol";
+
+library LibVaultUtils {
+    /**
+     * Thrown when the sender in a message context is invalid.
+     * This could happen if the sender isn't the expected token vault on the
+     * source chain.
+     */
+    error VAULT_INVALID_SENDER();
```
dantaik, line 20: You also need to re-declare the same error in all vault contract.

packages/protocol/contracts/tokenvault/libs/LibVaultUtils.sol
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * @dev Checks if context is valid
+     * @param validSender The valid sender to be allowed
+     * @param resolver The address of the resolver
+     */
+    function checkValidContext(
+        bytes32 validSender,
+        address resolver
+    )
+        external
```
dantaik, line 51: this method is used often, maybe make it 'internal' so the gas cost is smaller?

packages/protocol/contracts/tokenvault/libs/LibVaultUtils.sol
```diff
@@ -0,0 +1,108 @@
...
+                    ctx.srcChainId, validSender, false
+                )
+        ) {
+            revert VAULT_INVALID_SENDER();
+        }
+    }
+
+    /**
+     * @dev Returns the decoded data without selector
+     * @param calldataWithSelector Encoded data with selector
```
dantaik, line 68: document return value?

dantaik: To be honest, I still don't know what this function does!
If you just want to remove the selector, why not use https://docs.soliditylang.org/en/v0.6.0/types.html#array-slices

dantaik (changes requested):

packages/protocol/contracts/tokenvault/BaseNFTVault.sol
```diff
@@ -55,11 +56,14 @@ abstract contract BaseNFTVault is BaseVault {
...
 
+    // Released message hashes
+    mapping(bytes32 msgHash => bool released) public releasedMessages;
+
     // In order not to gas-out we need to hard cap the nr. of max
     // tokens (iterations)
     uint256 public constant MAX_TOKEN_PER_TXN = 10;
 
-    uint256[46] private __gap;
+    uint256[44] private __gap;
```
dantaik, line 66: 50-3=47?

packages/protocol/contracts/tokenvault/BaseNFTVault.sol
```diff
@@ -55,11 +56,14 @@ abstract contract BaseNFTVault is BaseVault {
     mapping(uint256 chainId => mapping(address ctokenAddress => address btoken))
         public canonicalToBridged;
 
+    // Released message hashes
+    mapping(bytes32 msgHash => bool released) public releasedMessages;
+
     // In order not to gas-out we need to hard cap the nr. of max
     // tokens (iterations)
     uint256 public constant MAX_TOKEN_PER_TXN = 10;
```
dantaik, line 64: please run pnpm lint:sol

dantaik, line 64: move this to top with other constants

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
```diff
@@ -239,7 +251,7 @@ contract ERC1155Vault is BaseNFTVault, ERC1155ReceiverUpgradeable {
      * @return tokenIds The tokenIds
      * @return amounts The amount per respective ERC1155 tokenid
      */
-    function decodeMessageData(bytes memory dataWithSelector)
+    function decodeMessageData(bytes calldata dataWithSelector)
```
dantaik, line 254: how about "msgData"?

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
```diff
@@ -251,11 +263,14 @@ contract ERC1155Vault is BaseNFTVault, ERC1155ReceiverUpgradeable {
         )
     {
         return abi.decode(
-            _extractCalldata(dataWithSelector),
+            dataWithSelector[4:],
```
dantaik, line 266: I think you can even remove this function, and simply use the following in other place(s):
```
(, CanonicalNFT memory nft,
            address owner,
            address to,
            uint256[] memory tokenIds,
            uint256[] memory amounts) = abi.decode(bytes4, anonicalNFT, address, address, uint256[], uint256[])

```

packages/protocol/contracts/tokenvault/libs/LibVaultUtils.sol
dantaik: Yes, but if this function being inlined will not increase the contract by too much, and gas cost for the transaction is also decreased, we should inline this function.

The other functions in this lib are less frequently used, so they can be external.

dantaik (changes requested):

PR #14246, opened 2023-07-22
chore(protocol): do not check sender in bridge's receive function [test failing]

dantaik:
This allows us to deploy more vaults without changing/upgrading the bridge contract.


packages/protocol/contracts/tokenvault/ERC721Vault.sol
```diff
@@ -145,11 +145,14 @@ contract ERC721Vault is BaseNFTVault, IERC721Receiver {
         }
 
         (
+            ,
             CanonicalNFT memory nft, //
             ,
             ,
             uint256[] memory tokenIds
-        ) = decodeMessageData(message.data);
+        ) = abi.decode(
```
dantaik, line 153: @adaki2004 this is the change I want, but I will lose connection soon, so please continue this change (remove all decodeMessageData)

dantaik:
> @dantaik I think you branched-off from an older state of the `ERC721_Bridging` branch, because that one does not have `BaseVault.sol` anymore. (Will arrange it Monday - will have more time.)

I updated the code, but there is a test failure. Please fix it.

adaki2004 (approved):

dantaik:
> > > @dantaik I think you branched-off from an older state of the `ERC721_Bridging` branch, because that one does not have `BaseVault.sol` anymore. (Will arrange it Monday - will have more time.)
> > 
> > 
> > I updated the code, but there is a test failure. Please fix it.
> 
> Updated the code, It was not a test issue, but a limitation to` abi.decode`. (bc. we encoded with `encodeWithSelector`, which prepends a 4bytes function selector) If we want to strip the first 4 bytes, we need to slice within the array, because for example: `abi.decode(msg.data, (bytes4,uint256)); `will not only skip the first 4 bytes, but assume that the bytes4 part is padded to 32 bytes - that's why it reverts due to the payload being too small. (This happens in our case too.)

Thank you, learned something :)

davidtaikocha (approved):

PR #14258, opened 2023-07-23
feat(eventindexer): speed up sync

cyberhorsey:
We need to resync eventindexers today, this will massively speed up their sync time by running each event task in parallel.


dionysuzx (approved):

RogerLamTd (approved):
lgtm

PR #14272, opened 2023-07-25
fix(protocol): fix issue in slash amount calc

### LINKED ISSUE #14216: In the L3's economic, the slash amount is too big compared with proof reward

At present, the proof reward is quite low, around 0.01 ttko. The slashing percentage stands at 2.5%, and as the staking amount increases, the slashing amount becomes disproportionately large compared to the proof reward.

It might be better to link the slashing amount to the reward per gas, rather than the percentage of staking amount, to make it more reasonable.

### Additional context

Additional context here.

dantaik:
https://github.com/taikoxyz/taiko-mono/issues/14216

### DIFF
--- a/packages/protocol/contracts/L1/IProverPool.sol
+++ b/packages/protocol/contracts/L1/IProverPool.sol
@@ -15,5 +15,5 @@ interface IProverPool {
 
     function releaseProver(address prover) external;
 
-    function slashProver(address prover) external;
+    function slashProver(address prover, uint64 proofReward) external;
 }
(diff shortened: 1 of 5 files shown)


davidtaikocha (dismissed):

adaki2004 (dismissed):

dantaik:
> Copying here what i wrote to Discord - about the root cause and possible fix:
> 
> ```
> We are having a discussion with ZK Pool.
> They raised the topic - and this is why we think we see lots of slashes :
> They need to stake with MAX_CAPACITY_LOWER_BOUND 128. Which means they have the capacity to prove 128 blcosk in parallel. They say:
> 128 is the mininum capacity, but in reality it is very unlikely we run 128 provers 🙂
> calculated the current block speed, with a few clusters from David  running on parallel, I thought 20 + 20 provers should be enough  for the amount we staked.
> Would it make sense to reduce this 128 to 20 for example.. ? I know it might result in 'free market proving' (to not stakers but anyonw) but less slashing.
> ```

I changed it to 32.

davidtaikocha (approved):

adaki2004 (approved):

PR #14319, opened 2023-07-30
Update official-links.mdx

pucedoteth:
This change helps users to know the up to date of official sources for taiko so they don't fall for a scam which many in this space. Also it lets them involved in taiko activities and events.


dantaik (commented):
I'm not sure this is a good idea :) otherwise, there will be a lot more links to be added to our home page.

MarcusWentz:
I think we should keep Taiko core links on the homepage 
and put all other integrations on the ecosystem page to keep the homepage lightweight.

PR #14320, opened 2023-07-31
fix(bridge-ui): alpha 4 fix bridge UI

jscriptcoder:
TODO


jscriptcoder:
Closing as it's not longer needed

PR #14331, opened 2023-08-01
feat(bridge-ui-v2): Fixed input validations and catching additional errors

KorbinianK:
Bridging ERC20 Tokens was not allowing to send all tokens, as it calculated the fees as if we were paying with the token

Also reordered some validations to stop other validations if we already hit an "error"

### DIFF
--- a/packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
+++ b/packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import type { FetchBalanceResult } from '@wagmi/core';
+  import { onMount } from 'svelte';
   import { t } from 'svelte-i18n';
   import { formatUnits, parseUnits } from 'viem';
 
@@ -8,14 +9,16 @@
   import { LoadingText } from '$components/LoadingText';
   import { warningToast } from '$components/NotificationToast';
   import { bridges, checkBalanceToBridge, getMaxAmountToBridge, type RequireAllowanceArgs } from '$libs/bridge';
+  import type { ERC20Bridge } from '$libs/bridge/ERC20Bridge';
+  import { chainContractsMap } from '$libs/chain';
   import { InsufficientAllowanceError, InsufficientBalanceError, RevertedWithFailedError } from '$libs/error';
-  import { getBalance as getTokenBalance, getAddress } from '$libs/token';
+  import { getAddress,getBalance as getTokenBalance } from '$libs/token';
   import { debounce } from '$libs/util/debounce';
+  import { getConnectedWallet } from '$libs/util/getConnectedWallet';
   import { truncateString } from '$libs/util/truncateString';
   import { uid } from '$libs/util/uid';
   import { account } from '$stores/account';
   import { network } from '$stores/network';
-  import type { ERC20Bridge } from '$libs/bridge/ERC20Bridge';
 
   import {
     computingBalance,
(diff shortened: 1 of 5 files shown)


packages/bridge-ui-v2/src/components/Bridge/Bridge.svelte
```diff
@@ -221,6 +230,14 @@
           // TODO: see contract for all possible errors
           errorToast($t('bridge.errors.send_erc20_error'));
           break;
+        case err instanceof UserRejectedRequestError:
+          // Todo: viem does not seem to detect UserRejectError
+          warningToast($t('bridge.errors.rejected'));
+          break;
+        case err instanceof TransactionExecutionError && err.shortMessage === 'User rejected the request.':
```
jscriptcoder, line 237: Not quite understand why this is needed?

packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
```diff
@@ -33,17 +37,19 @@
   let inputBox: InputBox;
   let computingMaxAmount = false;
 
+  onMount(() => {
```
jscriptcoder, line 40: Not sure why this is needed

packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
```diff
@@ -33,17 +37,19 @@
...
+    clearAmount();
```
jscriptcoder, line 41: Why do we want to clear on mount? If the reason is to clear the state, before we actually wanted to leave the amount on purpose when switching pages

packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
```diff
@@ -33,17 +37,19 @@
...
+
   // Public API
   export function clearAmount() {
     inputBox.clear();
     $enteredAmount = BigInt(0);
   }
 
   export async function validateAmount(token = $selectedToken, fee = $processingFee) {
-    $insufficientBalance = false;
-    $insufficientAllowance = false;
```
jscriptcoder, line 44: Is this fixing a problem? We're gonna validate these two, we start assuming there is no issue and then validate

packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
```diff
@@ -53,8 +59,38 @@
...
+    const tokenAddress = await getAddress({
+      token,
+      srcChainId: $network.id,
+      destChainId: $destNetwork.id,
+    });
+
+    if (!tokenAddress) return;
+
+    // Check for allowance
+    $insufficientAllowance = await erc20Bridge.requireAllowance({
```
jscriptcoder, line 86: Why do we need to do this here?, we already get the error when running the estimation, don't we?, is this not redundant?

jscriptcoder (approved):

packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
KorbinianK: It is not persisting (visibly) right now. We either clear it here or we make sure it is also displayed.

packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
KorbinianK: yes, it helps with the flickering of the buttons. In the future we could introduce a loading or disabled status

packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
KorbinianK: As discussed in the call, the gas estimation does not throw an error, so we need to call it.

dantaik (approved):

PR #14332, opened 2023-08-01
feat(protocol): Testing gas consumption maximums (DO NOT MERGE)

adaki2004:
As discussed, here is a test which provides information on gas consumption when:
- proposeBlock() with 10 verification, 120K bytes of non-zero `TxList` input calldata + `cacheTxListInfo` `true` (Tho the latter has no significant addition).
- proveBlock() with 10 verifications

Gas used by proveBlock() with 10 verifications: **263.414**
Gas used by proposeBlock() with 10 verifications: **2.012.363**

_DISCLAIMER_: this is a worst-case scenario, in theory can be that much but proposers/provers will independently verify blocks in practice. (See current testnet for behavior).

You can run the tests yourselves by :
`forge test -vv --match-test "test_gasConsumption"`

(We should not merge this PR because the foundry.toml had to be changed in order to these tests not gassing-out)


packages/protocol/test/MaxGasTest.t.sol
```diff
@@ -0,0 +1,269 @@
...
+        }
+        else{
+            vm.prank(msgSender, msgSender);
+            L1.proveBlock(meta.id, abi.encode(evidence));
+        }
+    }
+
+    function test_gasConsumption_with10verifications_in_a_row_with_prove_and_propose() external {
+
+        bytes memory txList2 = NonEmptyBytes121K.getNonEmpty120KBytes();
```
Brechtpd, line 169: Not possible to just generate a list of non-zero bytes on the fly? Or is the data inside the txList important somehow for the test to succeed?

Brechtpd: Sure but why not simply generate a list of 120k x 0x01 in code vs having a very long constant value that's pasted in the test code?

dantaik:
As of now, verifying a block has no token reward (all reward for that block goes to the prover, not the transactor). Maybe we should split block reward into 2 parts, for example, 97.5% to the prover, and 2.5% to the transactor?

PR #14339, opened 2023-08-01
feat(protocol): Gas limit behavior changes

Brechtpd:
Follow up on https://github.com/taikoxyz/taiko-mono/pull/14087

- Updates the how to prove blocks document 
- Removes the `input.gasLimit > config.blockMaxGasLimit` check in the smart contract

Also changes `meta.gasLimit` to simply always equal `config.blockMaxGasLimit`, this change could be reverted if that requires too many changes in other places as well, the circuits can simply ignore. `input.gasLimit` is not used anymore and could also be removed.

There seem to be two tests failing, but unrelated to my changes I think.


dantaik (dismissed):

davidtaikocha (dismissed):

dantaik:
> Also changes `meta.gasLimit` to simply always equal `config.blockMaxGasLimit`, this change could be reverted if that requires too many changes in other places as well, the circuits can simply ignore. `input.gasLimit` is not used anymore and could also be removed.

Brecht, will this force all L2 blocks to be large and small blocks won't be proposed, as on L2, the 1559 gas price is calculated using the block's gas limit:
Screenshot 2023-08-02 at 10 53 07


If this is true, then the confirmation latency will be larger if there are not enough transactions.

dantaik (dismissed):

dantaik:
@Brechtpd I removed "gasLimit" from input.

dantaik:
@adaki2004  @davidtaikocha @cyberhorsey  please take a look at this PR.

dantaik:
> We don't want the basefee to change that fast, so just using the worst case scenario for the new block should have practically no impact on the calculated basefee.

Brecht, after a second thought, I feel like there may still be an issue. With the current change, we forces L2 proposer to purchase `blockMaxGasLimit` gas through EIP1559. But ideally, the block proposer may want to dry-run the transactions and only purchase enough gas, for example, 10% more than try-run result. Then on L2, the base fee will be smaller compared to purchasing `blockMaxGasLimit` gas, which will allow the proposer to earn more transaction fees.

> As of [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md), the overall fee a transaction creator pays is calculated as: ( (base fee + priority fee) x units of gas used).

Brechtpd:
Making the basefee dependent on what is done in the current block makes block building very tricky, because that would mean that depending on what transactions are included in a block, will also impact what transactions can actually be included and how those transactions are executed. This is also different than how blocks are built for Ethereum.

It's also only additional gas purchased used for the basefee calculation for the current block, so the difference will be really tiny no matter what value is used and have no impact on the basefee of future blocks. The assumption is also that all blocks will be be full when there's congestion (basefee > 0), because if there is no congestion and there aren't enough transactions the basefee == 0 and so purchasing the full block gas limit will have no noticeable impact on the calculated basefee.

It's possible to just purchase 0 gas for the current block instead. I just think it will be more accurate (and a bit safer) to just purchase the max amount because it will be a much better estimate because full blocks are expected. But in the end I think purchasing 0 gas or the block gas limit is only a theoretical difference here.

dantaik:
> Making the basefee dependent on what is done in the current block makes block building very tricky, because that would mean that depending on what transactions are included in a block, will also impact what transactions can actually be included and how those transactions are executed. This is also different than how blocks are built for Ethereum.
> 
> It's also only additional gas purchased used for the basefee calculation for the current block, so the difference will be really tiny no matter what value is used and have no impact on the basefee of future blocks. The assumption is also that all blocks will be be full when there's congestion (basefee > 0), because if there is no congestion and there aren't enough transactions the basefee == 0 and so purchasing the full block gas limit will have no noticeable impact on the calculated basefee.
> 
> It's possible to just purchase 0 gas for the current block instead. I just think it will be more accurate (and a bit safer) to just purchase the max amount because it will be a much better estimate because full blocks are expected. But in the end I think purchasing 0 gas or the block gas limit is only a theoretical difference here.

Created a PR https://github.com/taikoxyz/taiko-mono/pull/14388/files to make sure L2 1559 basefee calculation is independent of gas limit, bur depend on gas used.

dantaik (dismissed):

davidtaikocha (dismissed):

dantaik (approved):

davidtaikocha (approved):

PR #14346, opened 2023-08-02
docs(website): add guidance for changing staked TTKOe Amount in Eldfell L3 Provers

rakkooo:


TIGGERLAND:
Many changes are being shown in the 'Files changed' section, but this might be due to the timing of when I cloned the repository. In reality, the only addition I am recommending is the following text.

### Changing the Staked TTKOe Amount (Only for Eldfell L3 Provers)

In the event that you have staked TTKOe and made an error in the amount, you can change the staked amount at any time [here](https://explorer.test.taiko.xyz/address/0xC9580414A4372BDdBd8e19e01854DC0B2b1390Cf/write-proxy#address-tabs).
However, there are some considerations.
For example, if you've staked 2 TTKOe and want to change the staked amount to 1 TTKOe, the difference of 1 TTKOe will be exited and you will not be able to withdraw it for a week. Be careful when entering the amount.

<Callout type="warning">
  Always manually add eight zeros for the decimal '10^8`.
</Callout>

Enter an arbitrary number below. Unlike the UI's Amount per Capacity, the entered amount will be directly reflected.

8. stake →
amount
rewardPerGas (uint32)
maxCapacity (uint32) (minimum:32)

In the example above, if you enter 1 in the amount and click Write, the amount will change from 2 to 1, and 1 TTKOe will be exited.

PR #14350, opened 2023-08-02
feat(bridge-ui-v2): Style adjustments

KorbinianK:
Inputs and paddings are now closer to our design, more changes will come with the activities PR soon.
image

External Links now have an arrow, indicating that you leave the page 
image


packages/bridge-ui-v2/src/components/Bridge/BridgeTabs.svelte
```diff
@@ -3,17 +3,20 @@
...
   import { page } from '$app/stores';
   import { LinkButton } from '$components/LinkButton';
-  $: isERC20Bridge = $page.route.id === '/' ? true : false;
-  $: isNFTBridge = $page.route.id === '/nft' ? true : false;
-  $: classes = `flex-basis-0 mr-2 p-3 rounded-full flex justify-center items-center  w-auto sm:w-20 2xl:w-32`;
+  import { classNames } from '$libs/util/classNames';
+
+  let classes = classNames('space-x-2', $$props.class);
+
+  $: isERC20Bridge = $page.route.id === '/';
```
cyberhorsey, line 10: These reactive props to determine if it's an erc20 or nft20 bridge by route seem unnecessary and flimsy if the route name changes? Is there no way to determine the bridge type based off of the actual selected domain bridge type? Also, since the components are so largely similiar but only certain core things should change, should they be different routes?

KorbinianK: It was a very quick PoC in my first days to get started. This will change once we properly begin the NFT Bridge feature. Essentially was an easy way to avoid touching the bridge component before previous work was done.

dantaik (dismissed):

dantaik (dismissed):

dantaik (approved):

jscriptcoder (approved):

PR #14356, opened 2023-08-03
docs(website): Run Sepolia Node eth-docker curl request error with fix

MarcusWentz:
Will make it simple for users to quickly debug a specific curl connection error message for setting up their Sepolia node with eth-docker.


packages/website/pages/docs/guides/run-a-node/run-a-sepolia-node.mdx
```diff
@@ -55,6 +55,23 @@ You can do a simple curl command to check if the node is synced (change port fro
...
+```
+check that your Execution Layer client ports 
+are setup by adding `el-shared.yml` to the list of files set with the `COMPOSE_FILE` variable in the `.env` file as mentioned in step `Expose RPC ports`.
+
+For example, your `COMPOSE_FILE` flag in your `.env` file should look something like 
+```bash
+COMPOSE_FILE=lighthouse-cl-only.yml:geth.yml:el-shared.yml
+```
+
+</Callout>
```
d1onys1us, line 73: this would be necessary to include, so i would do this instead as a step nearby "enable archive_mode". perhaps this step can be included to say "update .env". where it gives steps for using "vim" etc. tools to open the .env file, and modify these two keys. instead for a callout, we can use the same/similar warning from eth-docker, from here: https://eth-docker.net/Usage/Advanced/ that it's not encrypted, so do not expose this to internet.

dionysuzx (approved):

davidtaikocha (approved):

PR #14376, opened 2023-08-04
fix(bridge-ui-v2): complying with design

jscriptcoder:


packages/bridge-ui-v2/src/components/Header/Header.svelte
```diff
@@ -27,13 +25,11 @@
...
     md:py-7
  ">
-  <LogoWithText width={150} height={50} class="md:hidden" />
+  <LogoWithText width={77} height={21} class="md:hidden" />
 
   <div class="flex justify-end md:f-between-center w-full">
     {#if isBridgePage}
       <BridgeTabs class="hidden md:flex" />
-    {:else}
-      <div></div>
```
KorbinianK, line 36: This is here on purpose, otherwise it looks like this on every other page:
image

packages/bridge-ui-v2/src/styles/base.css
```diff
@@ -17,9 +17,24 @@
...
+  @media (min-width: 768px) {
+    body {
+      /* Desktop of larger screens (md) have grid */
+      /* TODO: Ask Jane if we have the right svgs here */
+      background-image:
+        url(/bg/grid.svg),
+        url(/bg/spotlights.svg),
+        /* grey-900/70 => grey-900/0 */ linear-gradient(180deg, rgb(11 16 27 / 70%) 0%, rgb(11 16 27 / 0%) 100%);
+    }
+  }
```
KorbinianK:
```suggestion
  body {
    /* No grid background by default */
    height: 100%;
    background-image: url(/bg/grid.svg), url(/bg/spotlights.svg), linear-gradient(270deg, #1f1f1f 0%, #000000 100%);
    background-size: 0, cover, cover; /* by setting the first image size to 0 for mobile, we hide the grid but keep the blend effect */
    background-blend-mode: color-dodge, lighten, multiply;
    background-attachment: fixed;
  }

  @media (min-width: 768px) {
    body{
      background-size: 840px, cover, cover;
    }
  }
```

packages/bridge-ui-v2/src/components/Header/Header.svelte
jscriptcoder: No no, adding an empty div to make something work is a hack, we can do this with CSS, no empty divs. I missed other pages. I'll fix it

jscriptcoder: Fixed. We just need to justify-end for all the children, and the tabs would take up the rest of the space. By default the alignment inside is left, so it'll go all the way to the left

packages/bridge-ui-v2/src/styles/base.css
jscriptcoder: We still have the problem of hiding the grid for mobile. We need to backgrounds here then

KorbinianK (approved):

PR #14377, opened 2023-08-04
feat(website): update FAQ appearance

2manslkh:


d1onys1us:
@2manslkh only have two comments with the PR:
1. this removes the sidebar rendering that is derived from the markdown style headings
2. (i believe) this reduces the indexing for search in the search bar

do you still think it's worth it considering those?

2manslkh:
You are right. The questions will not be indexed as they will not be found in the md files. I think we should not go with these changes

d1onys1us:
okay i will close the PR for now, @2manslkh i will not delete the branch though (will leave that to you)

PR #14386, opened 2023-08-05
chore(protocol): allow tests to override isMessageFailed.

dantaik:


dantaik:
How about this


```solidity

bool constant internal CHECK_MSG_FAILURE_USING_LIB = true;

if (CHECK_MSG_FAILURE_USING_LIB) {
  LIb...
} else {
   // static call
}

```

```solidity
contract TestBridge {
bool constant internal CHECK_MSG_FAILURE_USING_LIB = false;
// also override isMessageFailed() in this contract
```

PR #14388, opened 2023-08-06
feat(protocol): L2 1559 no longer use block gas limit

dantaik:
With this PR, the current block's gasLimit is no longer used in L2's EIP1559 base fee calculation. Instead, in side block N's anchor transaction, we first purchase block N-1's gasUsed, then issue new gas and adjust gasExcess.

### DIFF
--- a/packages/protocol/contracts/L2/TaikoL2.sol
+++ b/packages/protocol/contracts/L2/TaikoL2.sol
@@ -199,9 +199,11 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
         uint256 basefee;
         EIP1559Config memory config = getEIP1559Config();
         if (config.gasIssuedPerSecond != 0) {
-            (basefee, gasExcess) = _calcBasefee(
-                config, block.timestamp - parentTimestamp, parentGasUsed
-            );
+            (basefee, gasExcess) = _calcBasefee({
+                config: config,
+                timeSinceParent: block.timestamp - parentTimestamp,
+                parentGasUsed: parentGasUsed
+            });
         }
 
         // On L2, basefee is not burnt, but sent to a treasury instead.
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -321,36 +321,38 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
-            _gasExcess = uint64((a.max(b) - b).min(type(uint64).max));
         }
 
         _basefee = Lib1559Math.calculatePrice({
             xscale: config.xscale,
             yscale: config.yscale,
-            xExcess: _gasExcess,
-            xPurchase: gasLimit
+            xExcess: gasExcess,
+            xPurchase: parentGasUsedNet
```
Brechtpd, line 342: I think this adds an additional delay on the actual basefee that isn't really required, because this will set the basefee between the two points, but the parent block has already happened so the basefee needs to be at the end point `uint256(gasExcess) + parentGasUsedNet` already. I think the more accurate way to do this is to purchase 0 gas after already taking into account the parent gas ussed, which means calculating the derivative at the end point which is also easier/cheaper to calculate: `basefee = eth_qty(_gasExcess) / (TARGET * ADJUSTMENT_QUOTIENT)`.

The current implementation additionally followed an approach very similar to the one proposed in https://ethresear.ch/t/make-eip-1559-more-like-an-amm-curve/9082 in option 2, purchasing an extra amount of gas for the max amount of gas possible in the block for safety and accuracy, but without implementing the refund because that would be very complex to do.

dantaik: Check out this [new commit](https://github.com/taikoxyz/taiko-mono/pull/14388/commits/d9fbef87a145aab2fad84c1873dbd150f1568fad), let me know if it addresses your concerns.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -321,31 +321,31 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
+            // Very important to cap _gasExcess uint64
+            _gasExcess = uint64((excess - issued).min(type(uint64).max));
         }
 
         _basefee = Lib1559Math.calculatePrice({
             xscale: config.xscale,
             yscale: config.yscale,
             xExcess: _gasExcess,
-            xPurchase: gasLimit
+            xPurchase: 0
```
Brechtpd, line 346: This will now approximate the derivative by calculating the difference between two points with a step of `1`. Theoretically should be okay, but I don't know how this will effect the accuracy because 1 is a tiny difference and there are approximations being done.

dantaik: Feel free to optimize it further.

davidtaikocha (approved):

PR #14446, opened 2023-08-10
fix(bridge-ui): return true if the token address is found on dest chain to send correct gas limit

cyberhorsey:
- fixes network switching by removing the extraneous deffered / switching refactor logic, which doesnt seem to be necessary, and was preventing the page from actually switching the network. This allowed for users to send transactions on wrong networks, and the page to be out of sync with the injected wallet provider.
- properly detects if a token is bridged already or not, reducing gas limit, which reduces processing fee.


KorbinianK (dismissed):

KorbinianK (dismissed):

dantaik:
@xiaodino please also review.

dionysuzx (approved):

KorbinianK (approved):

PR #14448, opened 2023-08-10
feat(bridge-ui-v2): destination token bridged

KorbinianK:
When trying to find the canonical address, we now correctly determine the cross chain address for all cases.


dantaik:
@KorbinianK please make sure you also assign @xiaodino as a reviewer so he can catch up.

jscriptcoder (dismissed):

xiaodino (dismissed):

packages/bridge-ui-v2/src/libs/token/getCrossChainAddress.ts
```diff
@@ -11,22 +12,51 @@ type GetCrossChainAddressArgs = {
...
+
+  const { tokenVaultAddress: srcChainTokenVaultAddress } = chainContractsMap[srcChainId];
+  const { tokenVaultAddress: destChainTokenVaultAddress } = chainContractsMap[destChainId];
 
   const srcChainTokenAddress = token.addresses[srcChainId];
 
   // We cannot find the address if we don't have
   // the token address on the source chain
-  if (!srcChainTokenAddress) return;
+  if (!srcChainTokenAddress) return null;
```
cyberhorsey, line 34: i dont think 
```ts
  // We cannot find the address if we don't have
  // the token address on the source chain
  ```

is true, to return null here. `srcChainId` is the source chain you are on, right? not the actual source chain of the token, thats still be determined.

So if you are:
- on Eldfell (srcChain), bridging to Grimsvotn (destChain), 
- have added a custom token while on Grimsvotn (now the destChain) 
- and dont have a `srcChainTokenAddresses`, shouldnt you look up a `destChainTokenAddress` (ie: token.addresses[destChainId]) and then see if that is a bridged token or not, etc?

KorbinianK: When you add a custom token, the address and chainid are stored and verified. If you now change the chain, we still have that information and don't need to figure it out again. (What for?)

On L3 we show a message that you can't bridge that token back to L2 as it doesn't exist on L3. 

When the user selects a custom token on the native chain, we try to find the bridged address and store that as well. So when the user switched chains, we have the bridged address already. Come to think of it, we should do that as soon as the user adds the custom token. I'll add that.

Now there is the hypothetical case that someone else has bridged the token to L3 while the user is already connected to that chain, then the current source is empty but technically exists. I think only then an additional check actually does anything, right? Let me think about it.

KorbinianK: Made the change now. Assuming the user adds a custom token to Grimsvotn, we check if exists on Eldfell, if it does, we store that address, if not it will be 0x0, same the other way around.

So as soon as we select a token, we always have the addresses. We actually should never be able to reach that if unless something somewhere went terrible wrong and we have neither destination nor source token Address. In that case we now throw an error.

dantaik (approved):

xiaodino (approved):

PR #14451, opened 2023-08-11
chore(protocol): cleanup autogen docs

dionysuzx:
@dantaik @adaki2004 I reviewed each file in the previous PR and applied a consistent standard. Could you please:
1. Check for correctness (most important)
2. Look at the three `TODO(docs)` that I left

I will merge this PR after review.


packages/protocol/contracts/L1/IProofVerifier.sol
```diff
@@ -7,18 +7,17 @@
...
 /**
- * @title IProofVerifier Interface
- * @dev Interface for the ProofVerifier contract which is responsible for
- * verifying proofs.
+ * Contract that is responsible for verifying proofs.
+ * @title IProofVerifier
  */
 interface IProofVerifier {
     /**
-     * @notice Verify proof(s) for the given block.
```
dantaik, line 16: @d1onys1us so we are not using @notice anywhere? If you look at https://docs.soliditylang.org/en/v0.8.17/natspec-format.html, I think @notice is used many places.

packages/protocol/contracts/L1/ProverPool.sol
```diff
@@ -431,7 +434,10 @@ contract ProverPool is EssentialContract, IProverPool {
         return true;
     }
 
-    // Calculates the user weight's when it stakes/unstakes/slashed
+    /// @dev See the documentation for the callers of this internal function.
```
dantaik, line 437: Here in this doc, we used /** **/ and ///, shall we stick to one format across all files?

packages/protocol/contracts/common/AddressResolver.sol
```diff
@@ -132,7 +132,7 @@ abstract contract AddressResolver {
     }
 
     /**
-     * @notice Fetch the AddressManager's address.
+     * Fetch the AddressManager's address.
```
dantaik, line 135: Shall we force a rule for verb so here we use "Fetches" instead of "Fetch"?

packages/protocol/contracts/common/EssentialContract.sol
```diff
@@ -34,6 +32,10 @@ abstract contract EssentialContract is
         emit AddressManagerChanged(newAddressManager);
     }
 
+    /**
+     * Initialize the contract with an address manager.
```
dantaik, line 36: Initializes?

packages/protocol/contracts/signal/ISignalService.sol
```diff
@@ -8,25 +8,28 @@
 pragma solidity ^0.8.20;
 
 /**
+ * The SignalService contract serves as a secure cross-chain message passing
+ * system. It defines methods for sending and verifying signals with merkle
+ * proofs. The trust assumption is that the target chain has secure access
+ * to the merkle root (such as Taiko injects it in the anchor transaction).
+ * With this, verifying a signal is reduced to simply verifying a merkle proof.
```
dantaik, line 15: will put @title as the first lien better?

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -76,6 +46,9 @@ contract SignalService is ISignalService, EssentialContract {
         EssentialContract._init(_addressManager);
     }
 
+    /**
+     * @inheritdoc ISignalService
```
dantaik, line 50: 👍

packages/protocol/contracts/L1/ProverPool.sol
```diff
@@ -93,8 +91,7 @@ contract ProverPool is EssentialContract, IProverPool {
     }
 
     /**
-     * @notice Protocol specifies the current feePerGas and assigns a prover to
-     * a block.
+     * Protocol specifies the current feePerGas and assigns a prover to a block.
```
dantaik, line 94: use inheritdoc here as well?

dantaik:
> @dantaik @adaki2004 I reviewed each file in the previous PR and applied a consistent standard. Could you please:
> 
> 1. Check for correctness (most important)
> 2. Look at the three `TODO(docs)` that I left
> 
> I will merge this PR after review.

I removed the 3 TODOs

packages/protocol/contracts/signal/ISignalService.sol
d1onys1us: overall, i agree with this. in fact, there are two changes i want to make:
- @title before description
- verbosely include "@notice" even though it is not required by solidity compiler
- move multi-line comment blocks to `///` instead of `/** */` for simplicity

however, i would like to introduce that later, after we cleanup the bridge documentation as well and get consistency throughout the codebase. i already commented the bridge/signal service previously using the standards here: https://taiko.xyz/docs/manuals/contributing-manual#source-code-comments

so don't want to bikeshed over this. once we document to the spec, we can update the spec and use chatGPT to fixup all the files quickly.

packages/protocol/contracts/common/AddressResolver.sol
d1onys1us: in the future yes. unless you are aware of a out of the box solution that we can quickly use for this now. otherwise we will integrate an AI based solution or use vale (https://github.com/errata-ai/vale-action) in our CICD after cleaning up protocol docs.

d1onys1us: actually fetch is a verb though lol

d1onys1us: actually i agree with you. i resorted to google's javadoc writing style guide. it is preferred to use a third person singular verb. i updated this.

packages/protocol/contracts/L1/ProverPool.sol
d1onys1us: inheritdoc is not valid for contracts. i am following the same standard as sablier-v2 here: https://github.com/sablier-labs/v2-core/blob/main/src/SablierV2Comptroller.sol#L30

i added `@dev` to describe implementation specific details to the interface. i will add that as a standard in our documentation spec.

d1onys1us:
closing this PR in favor of https://github.com/taikoxyz/taiko-mono/pull/14462

PR #14468, opened 2023-08-13
fix(bridge-ui): set to null when chainID is not one of the two supported, so prevchain can be checked

cyberhorsey:


dionysuzx (approved):

packages/bridge-ui/src/utils/switchNetwork.ts
```diff
@@ -1,5 +1,43 @@
...
+  const timeout = 5000; // TODO: config?
+
+  const waitForNetworkChange = () => {
+    const srcChainId = get(srcChain)?.id;
+
+    if (srcChainId && srcChainId !== prevChainId) {
+      // We have finally set the chain in the store. We're done here.
+      deferred.resolve();
+    } else if (Date.now() > starting + timeout) {
+      // Wait, what???
```
d1onys1us, line 33: some comments in here i think we should remove if not needed or convert to TODOs

cyberhorsey: This is just copy pasted from Fran's work before, I think we can just leave it since we will be deleting this repo pretty soon

dionysuzx (approved):

jscriptcoder (approved):

PR #14479, opened 2023-08-14
Typo

xnodeer:
In [verify prover logs](https://taiko.xyz/docs/guides/run-a-node/enable-a-prover#verify-prover-logs)  for L2 verify prove logs there is wrong command

should use the command ``docker compose logs -f taiko_client_prover_relayer`` instead but in there is ``docker compose logs -f l3_taiko_client_prover_relayer`` which should be for the L3 node

@d1onys1us @robcrypt25 @MarcusWentz


MarcusWentz (changes requested):
481 files changed with 64 commits. 

Please sync your branch or open a new pull request with a fresh branch.

PR #14489, opened 2023-08-14
feat(protocol): change protocol data layout to enable coverage report

adaki2004:
Forge is only able to run `coverage` in case the contracts by default capable of compiling without any optimization (neither optimizer runs, no compiling `--via-ir `flag).
foundry-issue [link](https://github.com/foundry-rs/foundry/issues/3357#issuecomment-1297192171).
So in order to resolve this i needed to make  optimizations in the storage layout of `TaikoData`. Fortunately there is almost 0 overhead comes from reordering the layout this way and now coverage is running smooth. 

I ran gas-report before and after the changes and compared the results and were neglible.


packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -199,21 +221,11 @@ library TaikoData {
             ethDeposits;
         mapping(address account => uint256 balance) taikoTokenBalances;
         // Slot 7: never or rarely changed
-        uint64 genesisHeight;
-        uint64 genesisTimestamp;
-        uint64 __reserved70;
-        uint64 __reserved71;
+        Slot7 slot7;
```
dantaik, line 224: is this necessary?

dantaik:
> Forge is only able to run `coverage` in case the contracts by default capable of compiling without any optimization (neither optimizer runs, no compiling `--via-ir `flag). foundry-issue [link](https://github.com/foundry-rs/foundry/issues/3357#issuecomment-1297192171). So in order to resolve this i needed to make optimizations in the storage layout of `TaikoData`. Fortunately there is almost 0 overhead comes from reordering the layout this way and now coverage is running smooth.
> 
> I ran gas-report before and after the changes and compared the results and were neglible.

Please place this comment in the TaikoData file, right above the SlotA struct definitation.

dantaik (approved):

Issue #14452, opened 2023-08-11
feat(protocol): allow the proposer to select a prover


Brechtpd:
### Describe the feature request

Currently the proposer and the prover are completely separated. In practice however it's extremely likely block builders already know provers. It's also possible to allow these two to communicate with each other (PBS style).

This allows the proposer to find the prover that is willing to prove the block at the lowest price using an offchain auction, and this can be done on a block per block basis and taking into account the expected gas costs at the time. It also allows the prover to only prove blocks for which it has the capacity to do so.

Conceptually the idea is the same as when a block would be proposed directly with its proof. But because that is currently not possible (proof generation takes too long), instead a bonded promise is attached from the prover.

Onchain things are very simply:
- When the block is proposed, a signature is provided from the prover that it will prove the block. This is done by the prover signing the hash of all the block data (so that the block data itself can remain hidden if necessary when auctioning the block to provers) and specifying the expected fee payment.
- The protocol checks this signature and checks if the prover is bonded.
- If all those checks pass than the proposed block is accepted.
- When the block gets proven, the prover gets his expected fee payment.

When the block isn't proven in time, the bond of the prover gets partially slashed and those funds are used to pay for all the blocks the prover failed to prove.

The prover can also offer the reward for the proof to some other prover. When a block gets proven by another prover and a signature is provided by the expected prover, that prover will get the reward instead. This allows the original prover to avoid getting slashed, and encourages provers to work together to get the proof submitted in time, which benefits the stability of the chain.

Blocks could still be proposed without a specified prover, that will use the open proving path instead.

Benefits:
- Very easy and gas efficient
- Automatically as efficient as possible: the proposer will select the prover willing to prove the block for the lowest price to maximize its own profit. Or not and use the open path.
- Permissionless because any prover can become a prover by just posting the required bond. Lower upfront capital requirements.
- Censorship resistant because the proposer can pick any prover willing to prove the block and can prove the block themself if necessary. Some provers may not want to prove a block and they can just avoid proving those specific blocks without ever being forced to prove any other blocks.

#### Combating centralization

The approach above is great for efficiency, but will also be centralizing, So we now try to ensure prover redundancy for liveness while sacrificing efficiency as little as possible.

The methods below can be combined with one another.

##### Method A. Random (per block) burn per prover

[...]

PR #14510, opened 2023-08-15
feat(protocol): prototype prover abstraction

dantaik:
In this PR, I'm trying to further separate the current prover pool from the core protocol. This detachment facilitates flexibility, allowing the core protocol to easily adapt to alternative or superior prover pool designs without alterations. We're handing the task of finding the best prover pool design to third parties, and our role is to create the *glue* that connects the core protocol to multiple prover pools. This *glue* is embodied by our tokenomics, not by any particular pool.

The concept stems from Brecht's [feature request](https://github.com/taikoxyz/taiko-mono/issues/14452), but this PR concentrates solely on the *glue* aspect and does not incorporate any advanced features. Specifically, it permits a block proposer to select a prover when proposing a block. If the prover's address is `0`, the block is immediately open; otherwise, the selected prover must validate the block within a defined time window.

The functional details are as follows:
- If the assigned prover's address is non-zero and an EOA, the system verifies a signature to confirm the prover's agreement to validate the block at a capped fee-per-gas.
- If the prover is a contract (assumed to be an `IProver` contract), the method `prover.onBlockAssigned` is invoked. This method may include authorization and other logics. If it does not revert, the block can be proposed; otherwise, the proposal is rejected.

In all cases, the protocol enforce the burn of a bond of a minimal value, **_32 x block_deposit_**, in the implementation.

To avert unnecessary gas consumption, there are no notification calls to the prover contract when a block is proven or verified. This design prevents a malicious prover contract from exploiting callbacks like `onBlockProven` or `onBlockVerified` to exhaust gas.

Instead, the prover contract's deployer or owner manages off-chain authentication of block proposers and updates the prover contract's state. All prover contracts must have owners, as they cannot automatically modify their contract state or the behavior of the `onBlockAssigned` function on-chain without core protocol knowledge.

This approach may initially appear undesirable, but it corresponds with the fact that a prover's ability is based on off-chain information.

In this PR, it is also delineated that, generally, the proposer burns a block_deposit to cover the proving fee (block fee) using the block's gasLimit. If the gasUsed is less, a portion is refunded, making the proposer's payment equal to what the prover receives. The exception occurs when a block is not initially open and is verified by a prover other than the assigned one outside the proof window. In this case, we mint `avgFeePerGas * 1.5 * gasUsed` to the actual prover, but the proposer pays his expected amount, i.e., `feePerGas * gasUsed`. The prover's bond is not returned and will cover the additional cost. This almost guarantees the total supply of Taiko token decreases over time.

Tests are not updated and not runnable. I'd like to get your initial feedback first.


packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -170,6 +200,75 @@ library LibProposing {
...
+            // To cover open block reward, we have to use the max of _feePerGas
+            // and state.slotC.avgFeePerGas in the calculation in case
+            // _feePerGas is really small or zero.
+            uint32 bondFeePerGas = _feePerGas > state.slotC.avgFeePerGas
+                ? _feePerGas
+                : state.slotC.avgFeePerGas;
+
+            _bond = LibUtils.calcBlockFee(config, gasLimit, bondFeePerGas)
+                * config.proofBondMultiplier;
+        }
```
Brechtpd, line 269: I would not use `avgFeePerGas` because of two possible issues:
- All payments to the prover by the proposer can be done out of protocol, so `avgFeePerGas` can be 0.
- Provers may be more picky with blocks that are harder to prove (only doing low cost to proof/gas) so the cost per gas for open blocks could be severely miscalculated because those could all be difficult to prove blocks (high cost proof/gas).

I think in most cases, just constant protocol defined constants are the way to go. Easy to think about, less gas to use/update, and nothing can go wrong in the calculation. Another dynamic part of the protocol we don't have to worry about, and can just be set to reasonable constants that don't need to be updated frequently.

packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -129,6 +119,22 @@ library LibUtils {
...
+    )
+        internal
+        pure
+        returns (uint64)
+    {
+        uint64 _gas =
+            gasAmount + LibL2Consts.ANCHOR_GAS_COST + config.blockFeeBaseGas;
+        unchecked {
+            return _gas * feePerGas;
+        }
```
Brechtpd, line 135:
```suggestion
        unchecked {
            return gasAmount * feePerGas;
        }
```
The gas limit is now always the block gas limit, so I think we could just do this now?

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -196,41 +198,36 @@ library LibVerifying {
...
-                    proofReward -= proofReward * proofDelay
-                        * config.rewardMaxDelayPenalty / 10_000 / blk.proofWindow;
-                }
+            if (config.rewardMaxDelayPenalty > 0) {
+                // Give the reward a penalty up to a small percentage.
+                // This will encourage prover to submit proof ASAP.
+                blockFee -= uint64(
+                    uint256(blockFee) * proofDelay
+                        * config.rewardMaxDelayPenalty / 10_000 / blk.proofWindow
+                );
```
Brechtpd, line 230: Similar to above I guess, not sure if we need to incentivize faster provers like this, because this is a tax on provers that want to aggregate as many blocks as possible. If the network guarantees that proofs only to be submitted within the proof window, some blocks being proven faster doesn't really help the network and this extra value being leaked here will end up with users.

Brechtpd (commented):
I think this looks good! Taking into account the gas used makes the calculations a bit more complex than just a flat cost that could be calculated offchain, but I guess not that big of a deal and makes sense for some systems.
- Bond/block vs single bond/prover to be able to prove blocks. It's an interesting one. I'm more in favor of the single bond/prover, but I also see the benefits of the bond/block, but unsure how well that one will work.
  - **bond/block**: should mean lower capital requirements, but only when the prover is proving only a couple of blocks in parallel. And if the bond/block is pretty low, the amount that can be slashed is also low, so probably provers won't really care about losing e.g. 32x normal prover reward which may be in the order of ~$100. Sure it will be something to avoid, but doesn't seem to be something that will reduce their risk taking if they can lower the proof generation costs for all other blocks. A missed proof could have a big impact on the liveness of the chain, though perhaps also limited depending on the proof window and the proving time for a single block.
  - **bond/prover**: a bigger initial capital investment, but for proving a lot of blocks it should be more efficient. The slashing risk for even a single block can now also be much higher because if the bond requirement can be in the order of 1,000x-10,000x normal prover reward, while still ensuring the bond is sufficient to pay for the cost of proving all missed proofs. A small other benefit of this one is lower gas costs because normally the bond only needs to be checked when a block is proposed, so fewer token transfers.
- Generally I'm more in favor of fixed protocol constants vs protocol calculated statistics. 
- Still a bit concerned about centralization risks with nested prover pools but I think it's okay, I guess the same would be true if not explicitly possible, though now can also be done fully trustless.

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: As of now, `avgFeePerGas` is used to calculate the reward for open blocks. If we don't have this value, then we should disallow open blocks and only allow a block with non-zero prover to *become* open after proof window. This may further simplify the protocol code indeed.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -170,6 +200,75 @@ library LibProposing {
...
+        private
+        returns (address _actualProver, uint32 _feePerGas, uint64 _bond)
+    {
+        if (prover == address(0)) {
+            _actualProver = prover;
+            _feePerGas = maxFeePerGas;
+        } else if (!prover.isContract()) {
+            // Verify the prover has authorized this assignment
+            bytes32 hash =
+                keccak256(abi.encodePacked("PROVE_TAIKO_BLOCK", metaHash));
```
davidtaikocha, line 223: Since `metaHash` contains: `block.timestamp` \ `block.number` \ `block.prevrandao` \ `state.slotB.numBlocks`, seems its impossible for a user to sign this payload beforehand?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -110,50 +113,77 @@ library LibProposing {
...
-        blk.proposer = msg.sender;
-        blk.feePerGas = state.slotC.feePerGas;
-        blk.proposedAt = meta.timestamp;
+        unchecked {
+            blk.proofWindow = uint16(
+                (
+                    uint256(state.slotC.avgProofDelay)
+                        * config.proofWindowMultiplier / 100
+                ).min(config.proofMaxWindow).max(config.proofMinWindow)
+            );
```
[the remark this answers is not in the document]
dantaik: I'll remove `avgProofDelay` to make the core protocol simpler. This will disable a small feature: reward early proofs.

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: Will remove `avgFeePerGas` and disable open blocks -- blocks can still become open after the constant proof window

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: will remove as well

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: True...

packages/protocol/contracts/L1/libs/LibUtils.sol
dantaik: this method will be removed

dantaik:
Additional changes:

- proposer pays Ether as proving fee to the assigned prover when the block is proposed, this fee is not refundable. The prover will deposit a bond  in Taiko token which should be large enough. When the assigned prover proves the block, the whole bond is returned when the block is verified; otherwise, half of the bond is transferred to the actual prover. 
- block must have a non-zero prover address, i.e., blocks cannot be open immediately upon proposal.
- commented out the ProverPool code

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -175,101 +158,28 @@ library LibVerifying {
...
-                    newValue: blk.rewardPerGas,
-                    maf: 7200
-                })
-            );
+        if (
+            fc.prover == address(1)
+                || fc.provenAt <= blk.proposedAt + config.proofWindow
+        ) {
+            // Refund all the bond
+            tt.transfer(blk.prover, config.proofBond);
```
Brechtpd, line 179: If we'll keep a bond per block, than `proofBond` may have to be saved in the block because otherwise if `config.proofBond` increases between proposal and here, this may fail because too few tokens that can be transferred (because fewer were deposited at proposal).

Brechtpd (approved):
I really like the simplicity! Paying ETH to the prover directly when the block is proposed is also a nice touch, and makes a lot of sense for its flexibility.

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: Normally the proposer will know it because they will only want to propose blocks that they know exactly how they will behave so they can optimize for the transactions they include in the block. And so because the proposer knows, they will only request a quote from provers after they are done building, it makes sense to also give it to provers so they will also know exactly what block they will have to prove so they can accurately predict their proving costs (or at least more accurately, they don't know how many more blocks they will get to prove and what the Ethereum gas cost will be).

But I totally agree for now it is safer not to include it because proposers will unlikely be very advanced initially and may be very problematic on testnets. In the future I think having the option for the prover to sign off on the exact block data as an option may be useful.

packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -122,10 +111,8 @@ library LibUtils {
             | (uint256(meta.gasLimit) << 176)
             | (uint256(uint160(meta.beneficiary)) << 16);
 
-        inputs[6] = (uint256(uint160(meta.treasury)) << 96);
```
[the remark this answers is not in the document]
dantaik: That's what I have in mind, we can simply hard code it in ZK and the protocol doesn't have to care. What do you think?

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: It's indeed an issue. How about we simply use burn/mint instead to avoid using one extra storage slot per block.

packages/protocol/contracts/L1/IProver.sol
```diff
@@ -0,0 +1,24 @@
...
+/// @notice Defines the function that handle prover assignment.
+
+interface IProver {
+    /// @notice Assigns a prover to a specific block or reverts if none is
+    /// available.
+    /// @param proposer The address of the block proposer.
+    /// @param input The block's BlockMetadataInput data.
+    function onBlockAssigned(
+        address proposer,
+        TaikoData.BlockMetadataInput calldata input
```
dantaik, line 20: @adaki2004 not sure if changing 'calldata' to 'memory' will increase or decrease cost.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -94,182 +83,108 @@ library LibVerifying {
...
-                        * config.rewardMaxDelayPenalty / 10_000 / blk.proofWindow;
-                }
+        unchecked {
+            TaikoToken tt = TaikoToken(resolver.resolve("taiko_token", false));
+            if (
+                fc.prover == address(1)
+                    || fc.provenAt <= blk.proposedAt + config.proofWindow
+            ) {
+                // Refund all the bond
+                tt.mint(blk.prover, config.proofBond);
```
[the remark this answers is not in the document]
dantaik: If the bond is large, then 1% may be too large compared to the actual proving fee...

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -45,47 +48,56 @@ library LibProposing {
...
+            IProver(assignment.prover).onBlockAssigned{ value: msg.value }(
+                input, assignment
+            );
+        } else {
+            bytes32 hash =
+                keccak256(abi.encode(input, msg.value, assignment.expiry));
+            if (assignment.prover != hash.recover(assignment.data)) {
+                revert L1_INVALID_PROVER_SIG();
             }
+            assignment.prover.sendEther(msg.value);
```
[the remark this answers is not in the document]
dantaik: msg.value inside `onBlockAssigned` is the same value as inside this function, so no need to pass it in explicitly.

packages/protocol/contracts/L1/libs/LibUtils.sol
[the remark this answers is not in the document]
Brechtpd: Yeah makes sense, the treasury address is unlikely to change if every, or much more rare than the circuits itself.

packages/protocol/contracts/L1/libs/LibVerifying.sol
Brechtpd: Burn/mint is indeed easier, not sure what the cost difference is between transfer/burn/mint because burn/mint also need to update the total balance.

PR #14529, opened 2023-08-16
chore(bridge-ui-v2): Update tooltips for Processing fee and Recipient address

xiaodino:
Before
Screen Shot 2023-08-16 at 3 13 34 PM

After
Screen Shot 2023-08-16 at 9 45 18 PM

### DIFF
--- a/packages/bridge-ui-v2/src/components/Activities/MobileDetailsDialog.svelte
+++ b/packages/bridge-ui-v2/src/components/Activities/MobileDetailsDialog.svelte
@@ -60,7 +60,7 @@
           <h4 class="text-secondary-content">{$t('activities.header.explorer')}</h4>
           <a
             href={`${chainUrlMap[Number(selectedItem.srcChainId)].explorerUrl}/tx/${selectedItem.hash}`}
-            target="_blank">
+            target="noopener">
             {$t('activities.link.explorer')}
           </a>
         </li>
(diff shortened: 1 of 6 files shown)


cyberhorsey:
Should these still be white even on dark mode?

packages/bridge-ui-v2/src/components/Activities/MobileDetailsDialog.svelte
```diff
@@ -60,7 +60,7 @@
           <h4 class="text-secondary-content">{$t('activities.header.explorer')}</h4>
           <a
             href={`${chainUrlMap[Number(selectedItem.srcChainId)].explorerUrl}/tx/${selectedItem.hash}`}
-            target="_blank">
+            target="noopener">
```
KorbinianK, line 63: noopener is not a valid target. noopener is a rel attribute keyword.

[see Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener)

xiaodino: Reverted the change.

xiaodino:
> Should these still be white even on dark mode?

It's not white. The background color is ```bg-primary-base-background```, which will be updated to another color in the light mode.

KorbinianK (dismissed):

KorbinianK (approved):

dantaik (approved):

PR #14544, opened 2023-08-20
feat(protocol): block reward to incentivize proposers

dantaik:
Mint `proposerRewardPerSecond` tokens per second for each block's proposer based on the block's delay since its parent.  A max reward amount (`proposerRewardMax`) of 32 Taiko token is applied.

I changed my mind not to reward provers to simplify their decision making: if proposers expect a bigger income, they will be willing to pay more proving fees to provers.

### DIFF
--- a/packages/protocol/contracts/L1/ProofVerifier.sol
+++ b/packages/protocol/contracts/L1/ProofVerifier.sol
@@ -8,10 +8,10 @@ pragma solidity ^0.8.20;
 
 import { AddressResolver } from "../common/AddressResolver.sol";
 import { EssentialContract } from "../common/EssentialContract.sol";
-import { Proxied } from "../common/Proxied.sol";
-import { LibZKPVerifier } from "./libs/verifiers/LibZKPVerifier.sol";
 import { IProofVerifier } from "./IProofVerifier.sol";
 import { LibBytesUtils } from "../thirdparty/LibBytesUtils.sol";
+import { LibZKPVerifier } from "./libs/verifiers/LibZKPVerifier.sol";
+import { Proxied } from "../common/Proxied.sol";
 
 /// @title ProofVerifier
 /// @notice See the documentation in {IProofVerifier}.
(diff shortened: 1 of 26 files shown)


Brechtpd:
L2 block rewards are a bit tricky because they differ quite a bit from L1 where proposing a block is free, as in if you are selected it doesn't cost you anything to propose any block. An empty block has the same cost as a full block, a full block provides more fees so it the one that is picked by the L1 block proposer. For an L2 however, an empty L2 block can be much cheaper than a full L2 block because of the L1 costs + the prover costs. So that means that the block reward does not necessarily mean that L2 transactions will be included in a block, it may mean that the smartest thing that is possible is for the proposer to just propose empty blocks because that will provide the highest profit.

There's also the difficulty of the varying L1 gas costs where the subsidy at one point may be very substantial and at other times may be insignificant.  If L1 gas costs are low, proposer may just "mine" these subsidies with empty blocks (and this money will actually end up with the L1 validators, not the proposers/provers).

So in the end this will subsidize transactions for users, which probably makes sense. But we have to be careful not to subsidize empty blocks or useless transactions, because that will just waste money. 

There's also another risk. If Ethereum transactions are made up of mostly L2 transactions, and if every L2 subsidizes its own blocks, than effectively transactions are not subsided because that makes it so the L1 gas cost will simply eat the subsidy of all L2s, resulting in no cheaper transactions for L2 users.

Two changes that I think make sense:
- Don't do a fixed reward per block, instead just emit a reward per L1 block and let it accumulate until the next L2 block is proposed: `config.emmisionPerSecond * (block.timestamp - state.blocks[(b.numBlocks - 1) % config.blockRingBufferSize].proposedAt)`. This makes the amount of tokens minted over time fixed, but the reward per block dynamic so it can change depending on the requirements at some time. The longer it takes for an L2 block to be proposed, the higher the reward for that block. If there are enough transactions with high enough fee to make building a block profitable, than only a tiny or no block reward will be given per block.
- Don't hard-code the emission schedule using hard coded halvings. You can argue for them for chains without any sort of governance or strict monetary rules. L2s have governance, and the L2 token isn't trying to be money. How large the emission should be now also depends on the L1. I would just let governance decide on the emission schedule, they will be in a much better place to decide what makes sense.

dantaik:
> 

Please take another look.

PR #14547, opened 2023-08-21
feat(protocol): remove config.blockMaxTransactions

dantaik:
Brecht, the idea is to use blockMaxGasLimit to control the size of L2 blocks, and remove the number of transaction limitation. Do you think this will be fine?

If we still need such a limit, I think they can also be removed from L1 contract and rely on the circuits to perform the checks.


dantaik:
@adaki2004 when I run `pnpm test`, I got these errors, could you help me fix them?

```
Failing tests:
Encountered 4 failing tests in test/L1/TaikoL1Oracle.t.sol:TaikoL1OracleTest
[FAIL. Reason: Call did not revert as expected] test_L1_OracleProverCanAlwaysOverwriteIfNotSameProof() (gas: 506752)
[FAIL. Reason: Call did not revert as expected] test_L1_OracleProverCannotOverwriteIfSameProof() (gas: 407910)
[FAIL. Reason: Call did not revert as expected] test_L1_if_oracle_prover_can_prove_but_regular_provers_can_not_overwrite() (gas: 469342)
[FAIL. Reason: Call did not revert as expected] test_L1_if_there_is_no_oracle_prover_there_is_no_overwrite_at_all() (gas: 447279)
```

packages/protocol/test/L1/TaikoL1TestBase.sol
```diff
@@ -246,39 +239,6 @@ abstract contract TaikoL1TestBase is TestBase {
         console2.log(str);
     }
 
-    function getInstance(
```
dantaik, line 249: remove the duplicate function to avoid errors

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -202,13 +198,8 @@ library LibProving {
             | (uint256(evidence.parentGasUsed) << 64)
             | (uint256(evidence.gasUsed) << 32);
 
-        // Also hash configs that will be used by circuits
-        inputs[9] = uint256(config.blockMaxGasLimit) << 192
```
dantaik, line 206: @Brechtpd  I feel like we do not need to hash configs in the instance. Let me know what you think.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -123,11 +123,16 @@ library LibProving {
...
         fc.gasUsed = evidence.gasUsed;
 
-        IProofVerifier(resolver.resolve("proof_verifier", false)).verifyProofs({
-            blockId: blockId,
-            blockProofs: evidence.proofs,
-            instance: getInstance(config, resolver, evidence)
-        });
+        bytes32 instance = getInstance(
+            evidence,
+            resolver.resolve("signal_service", false),
```
dantaik, line 128: @Brechtpd We may also skip hashing these addresses as they will never change (can be upgraded). What do you think?

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -28,7 +28,6 @@ contract TaikoL1 is TaikoL1Base {
             blockMaxVerificationsPerTx: 10,
             blockMaxGasLimit: 6_000_000,
             blockFeeBaseGas: 20_000,
-            blockMaxTransactions: 79,
```
Brechtpd, line 31: There's a separate limitation inside the circuits because transactions are handled separately, so there is for example a limited amount of signatures we can check. But yeah we can check this limitation inside the circuits instead as long as that limitation is there.

packages/protocol/contracts/L1/libs/LibProving.sol
Brechtpd: Yeah I'm okay with skipping and hard-coding it in the circuits. Hashing these values here when the proof is being submitted was actually flawed because it made the block non-deterministic if the value changed, so even if we would want to hash these values, it should be done at the proposer side so the proof always remains the same even if the config changes.

PR #14554, opened 2023-08-22
chore(protocol): change the way instance is calculated

dantaik:
@smtmfft @johntaiko Do you think it's easier to calculate the `instance` this way?


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -170,19 +170,6 @@ library LibProving {
...
-        inputs[3] = uint256(evidence.signalRoot);
-        inputs[4] = uint256(evidence.graffiti);
-        inputs[5] = (uint256(uint160(evidence.prover)) << 96)
-            | (uint256(evidence.parentGasUsed) << 64)
-            | (uint256(evidence.gasUsed) << 32);
-
-        assembly {
-            instance := keccak256(inputs, mul(32, 6))
-        }
+        else return keccak256(abi.encode(evidence));
```
johntaiko, line 173: It's easy to write the circuit when the `encode` returned bytes with padding. Like:
```
input: abi.encode(0x5B38Da6a701c568545dCfcB03FcB875f56beddC4(hash), 127(int))
output: 0x0000000000000000000000005b38da6a701c568545dcfcb03fcb875f56beddc4000000000000000000000000000000000000000000000000000000000000007f
```
We only need to combine all fields in evidence manually instead of like a ABI_ENCODE_CIRCUIT. But anyway, the pi-circuit needs to rewrite with new layout.

johntaiko: But, manually combining the evidence has a side effect, because we need always to rewrite the pi-circuit when evidence changed.

dantaik: The new code uses abi.encode, so it does pad each field to a 32 bytes. So this means the new code is easier, right?

dantaik: > But, manually combining the evidence has a side effect, because we need always to rewrite the pi-circuit when evidence changed.

The circuit code will have to be changed regardless.

johntaiko: > The new code uses abi.encode, so it does pad each field to a 32 bytes. So this means the new code is easier, right?

Yeah, same as before, we used `uint256` to cast all fields, this will also be padded to 32 bytes. But, something needs to be confirmed, how is the behavior of encode array, map types(for now, these types are not in the evidence

johntaiko: I mean the fixed length fields are fit to our current pi-circuit

yueawang: encoding change brings circuit change, but fix length so far is fine as all data layout is fixed.
But it can not easily extend to nested, or complex data structure, although a specific circuit is still fine.

johntaiko: But it's not easy to write circuit with dynamic size types, so I hope we will not have any of these types in the future.

smtmfft: > encoding change brings circuit change, but fix length so far is fine as all data layout is fixed. But it can not easily extend to nested, or complex data structure, although a specific circuit is still fine.

> But it's not easy to write circuit with dynamic size types, so I hope we will not have any of these types in the future.

Yes, fixed-len data type and fixed data fields number is the best, otherwise, data fields change (add/remove fixed-len types) is easier than dynamic data type (bytes, string)

smtmfft (approved):

PR #14558, opened 2023-08-22
feat(protocol): allow L2 proposers to bribe L1 validators

dantaik:
This is inspired by Dave's talk at Tea with Taiko.


Brechtpd:
I don't think this is enough because the payment also needs to be related to some (custom) preconditions for inclusion, so a custom contract still seems required to handle that for it to work. I think better not to bring any standard PBS mechanisms in the protocol unless we know exactly what kind of conditions proposers will want to use, and even then better to just leave it external unless there's a good reason not to keep it external? (Without PBS this is basically just a higher transaction tip.)

PR #14310, opened 2023-07-28
refactor(protocol): optimize core protocol and bridge implementation

adaki2004:
Inspired by [Brecht's feature request](https://github.com/taikoxyz/taiko-mono/issues/14452), this PR aims to separate the current prover (pool) contract from the core protocol, allowing various prover contract designs. Taiko will act as the "glue" between the core protocol and different prover contracts.

Core protocol Changes introduced in this PR:

- **Block Proposal**: Blocks must assign a non-zero prover address, and proposers will pay a non-refundable Ether fee to the prover. The prover will burn Taiko tokens as a bond, returned if proof is submitted within 90 minutes; otherwise, the block opens, and half the bond goes to the actual prover.
- **Prover Authorization**: If an EOA, must authorize the proposer with a valid ECDSA signature; if a contract, must have a `onBlockAssigned` function that doesn't revert.
- **Notification Calls**: Removed on-chain notification calls to prover (pool) contract to prevent malicious exploitation.
- **Prover (Pool) Contract and API**: Deleted IProverPool. Taiko and partners will create a Pool API server for proposers to access proving services.
- **Token and Configuration**: Taiko token supply won't increase; decimals are set to 18. TaikoConfig is simplified, loaded from a Library for testing.
- [ ] @davidtaikocha @smtmfft I added two TODOs for you in this PR.
> Taiko Token functions as a bond token for provers.

Bridge changes introduced in this PR:

- Introduced a recallMessage on the source bridge, it will call the message.from to perform custom recall function. (Releasing Ether is no longer handled by the bridge, but the message's from contract.
- Implemented all optimizations captured in this issue: https://github.com/taikoxyz/taiko-mono/issues/14384
- Added NFT vaults

PR #14577, opened 2023-08-24
fix(protocol): LibBridgeRetry fix

adaki2004:
As per Brecht comment: https://github.com/taikoxyz/taiko-mono/pull/14310#discussion_r1302266828


packages/protocol/contracts/bridge/libs/LibBridgeRetry.sol
```diff
@@ -71,32 +71,21 @@ library LibBridgeRetry {
...
             LibBridgeStatus.updateMessageStatus(
                 msgHash, LibBridgeStatus.MessageStatus.FAILED
             );
-
-            // Refund to the specified address, or the user if not specified.
-            address refundTo =
-                message.refundTo == address(0) ? message.user : message.refundTo;
-            refundTo.sendEther(message.value);
-        } else {
-            // Release Ether back to EtherVault if not the last attempt.
```
Brechtpd, line 99: Would keep some of the comments here.

Brechtpd (approved):

PR #14582, opened 2023-08-25
refactor(protocol): remove gasUsed and parentGasUsed

dantaik:


cyberhorsey:
I am out of the loop, why are we removing these?

dantaik:
> I am out of the loop, why are we removing these?

Previously proving fees are calculated based on gas used * fee per gas, now proving fee is decided off-chain, so gas used is no longer used on chain.

cyberhorsey:
> > I am out of the loop, why are we removing these?
> 
> Previously proving fees are calculated based on gas used * fee per gas, now proving fee is decided off-chain, so gas used is no longer used on chain.

Sounds good, in the client is it no longer necessary to us `parentGasUsed` when checking if a proof is valid (ie: we currently do:

```go
       // its valid 
	if parent.GasUsed == parentGasUsed && parent.Hash() == parentHash && blockHash == block.Hash() {
		return true, nil
	}
```

i can remove all parentGasUsed cehcks here, or we should keep them offchain?

cyberhorsey (approved):

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -124,8 +124,6 @@ library TaikoData {
         bytes32 signalRoot;
         bytes32 graffiti;
         address prover;
-        uint32 parentGasUsed;
-        uint32 gasUsed;
```
davidtaikocha, line 128: seems `TaikoL2.anchor` is still using `parentGasUsed` as one of its parameters, remove this from evidence means its not in circuits public input anymore (since right now the public input is `keccak256(abi.encode(evidence))`), not sure if its ok?

Brechtpd: Yep that's the reason we stored the gas_used onchain in the fork choice and had to link it with the parent one, because we somehow had to have the gas_used of the previous block available to be able to correctly update the L2 basefee. Using the gas_used of the current block in the anchor transaction would create a circular dependency when creating the block, so we had to avoid that somehow.

But we will now have access to all 256 previous block headers in the circuits, so we can now get the parent gas used from the block header of the previous block in a verifiable way!

So I agree that we can remove this now, though for a different reason than it seems this PR was created haha

Brechtpd (approved):

Brechtpd:
> > > I am out of the loop, why are we removing these?
> > 
> > 
> > Previously proving fees are calculated based on gas used * fee per gas, now proving fee is decided off-chain, so gas used is no longer used on chain.
> 
> Sounds good, in the client is it no longer necessary to use `parentGasUsed` when checking if a proof is valid (ie: we currently do:
> 
> ```go
>        // its valid 
> 	if parent.GasUsed == parentGasUsed && parent.Hash() == parentHash && blockHash == block.Hash() {
> 		return true, nil
> 	}
> ```
> 
> i can remove all parentGasUsed checks here, or we should keep them still offchain?

You can remove them.

adaki2004 (approved):

PR #14595, opened 2023-08-28
fix(protocol): block reward must be minted

dantaik:
@adaki2004 Dani, there seems to be a test failure, could you please review the changes and also fix the bug?

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -81,14 +81,13 @@ library LibProposing {
             revert L1_TOO_MANY_BLOCKS();
         }
 
+        TaikoToken tt = TaikoToken(resolver.resolve("taiko_token", false));
         if (state.taikoTokenBalances[assignment.prover] >= config.proofBond) {
             unchecked {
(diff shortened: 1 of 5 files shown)


packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -138,7 +137,8 @@ library LibProposing {
                         config.proposerRewardMax
                     );
 
-                    state.taikoTokenBalances[input.beneficiary] += reward;
+                    // Reward must be minted
+                    tt.mint(input.beneficiary, reward);
```
davidtaikocha, line 141: Just wondering should we use `msg.sender` (a L1 address) instead of `input.beneficiary` (a L2 address) here? since `msg.sender` is the account who paid the ether fees here.

dantaik: If the fee is smaller than the reward, then it will be a problem. So maybe it is safer to use input.beneficiary instead of msg.sender.

adaki2004 (approved):

davidtaikocha (approved):

PR #14605, opened 2023-08-30
feat(relayer): Relayer indexer/processor separation and refactor, messaging queue

### LINKED ISSUE #14528: feat(relayer): bridge relayer messaging queue

Relayer should be split up into two subcommands, indexer and processor. Indexer indexes the chain and adds messages to a queue (rabbit, Kafka, etc). Processor reads from the queue.  
  
This will enable high fault tolerance and persistence, especially when the chain is backed up and blocks take a very long time to verify.

A refactor should accompany this to use the same urfave cli package as taiko-client, to make it easier to work across both projects, and similiar subcommand/config pattern.

_No response_

### Additional context

_No response_

cyberhorsey:
Closes #14528 

this PR updates the relayer codebase to be inline with the `taiko-client` architecture, which should make it easier to onboard new people, as well as cross-share work more - once they get familiar with one codebase, they are familiar with all of our Golang applications.

It splits the relayer into two separate apps, the `indexer` and the `processor`, that now communicate asynchronosly through a `Queue` interface, the first implementation of which is a `RabbitMQ` queue (attached picture, processor is on the left reading from the queue, indexer is on the right listenign to the blockchain and adding to the queue).

This has three main advantages:

1) The indexer will be caught up to the tip of the chain much quicker, allowing the bridge-ui to be up-to-date.
2) The persistence of the queue will allow us to more reliably know which messages have been processed, how many behind we are, and which ones failed to be processed.
3) We can spin up as many additional processors as we want, without affecting the indexing, if we see many messages coming in that one processor can not handle.

The main difference is all of the CLI args of the relayer can still be passed in through Environment Variables as well, for easier deployment/configuration.

Latest code is already deployed (matching k8s-config PR is open) to internal-devnet for L1-L2, L2-L1, L2-L3, L3-L2, as well as a RabbitMQ statefulset.

TODO:
- tests for optional flag vars, current only tests requirred (can come in another PR, perhaps a task roger can do to get familiar with this Go codebase as well)
Screenshot from 2023-08-31 12-20-06


dantaik:
@cyberhorsey does this PR address [this issue ](https://github.com/taikoxyz/taiko-mono/issues/14528)?

cyberhorsey:
> @cyberhorsey does this PR address [this issue ](https://github.com/taikoxyz/taiko-mono/issues/14528)?

Yes, but it's only started today and will be maybe two more days.

spachib:
Will it be possible to self host this new relayer setup?

cyberhorsey:
> Will it be possible to self host this new relayer setup?

Yes definitely, you will need to run RabbitMQ where your stuff is deployed. I will add a writeup to the readme along with default ENV values for each layer and mode.

The main "feature loss" here is the relayer can no longer run in L1-L2 processing and L2-L1 processing at the same time (equivilant to the `--layers=both` flag from before), nor can it run in "indexing and processing" mode.

So for instance, for L1-L2, i have this deployed:
- an L1-L2 indexer `go run cmd/main.go indexer` with a `.l1indexer.env`
- an L2-L1 indexer `go run cmd/main.go indexer` with a `.l2indexer.env`
- an L1-L2 processor `go run cmd.main.go processor` with a `.l1processor.env`
- an L2-L1 processor `go run cmd/main.go processor` with a `.l2processor.env`

packages/relayer/indexer/indexer.go
```diff
@@ -0,0 +1,383 @@
...
+	"math/big"
+	"sync"
+	"time"
+
+	"github.com/cyberhorsey/errors"
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
```
davidtaikocha, line 17: Shall we import and reuse this package in this repo: https://github.com/taikoxyz/taiko-client/blob/main/pkg/rpc/ethclient.go ? not sure if this repo will also face those kind of timeout issues?

packages/relayer/indexer/indexer.go
```diff
@@ -0,0 +1,383 @@
...
+		slog.Error("srv shutdown", "error", err)
+	}
+
+	i.wg.Wait()
+}
+
+// nolint: funlen
+func (i *Indexer) Start() error {
+	go func() {
+		if err := i.srv.Start(fmt.Sprintf(":%v", i.httpPort)); err != nil {
```
davidtaikocha, line 221: Seems in `echo`'s doc, they are checking `err != http.ErrServerClosed` instead of `err != nil`, shall we also do that here? https://echo.labstack.com/docs/start-server

packages/relayer/indexer/indexer.go
```diff
@@ -0,0 +1,383 @@
...
+			slog.Error("http srv start", "error", err.Error())
+		}
+	}()
+
+	ctx := context.Background()
```
davidtaikocha, line 226: Do we need a global `context` in `Indexer` struct? (like in taiko-client)

packages/relayer/indexer/indexer.go
```diff
@@ -0,0 +1,383 @@
...
+	}
+
+	i.wg.Add(1)
+
+	go func() {
+		defer func() {
+			i.wg.Done()
+		}()
+
+		if err := i.filter(ctx); err != nil {
```
davidtaikocha, line 239: Shall we retry here? if there are some recoverable errors, like network fluctuations.

packages/relayer/indexer/indexer.go
```diff
@@ -0,0 +1,383 @@
...
+		}
+
+		group, groupCtx := errgroup.WithContext(ctx)
+
+		group.SetLimit(i.numGoroutines)
+
+		for {
+			event := messageSentEvents.Event
+
+			group.Go(func() error {
```
davidtaikocha, line 332: Looks like there is no `group.Wait`?

packages/relayer/processor/processor.go
```diff
@@ -0,0 +1,293 @@
...
+	srcChainId, err := srcEthClient.ChainID(context.Background())
+	if err != nil {
+		return err
+	}
+
+	publicKey := cfg.ProcessorPrivateKey.Public()
+
+	publicKeyECDSA, ok := publicKey.(*ecdsa.PublicKey)
+	if !ok {
+		return err
```
davidtaikocha, line 193: emm i guess here `err == nil`?

packages/relayer/indexer/indexer.go
cyberhorsey: its a few lines below:
```ts
				// wait for the last of the goroutines to finish
				if err := group.Wait(); err != nil {
					return errors.Wrap(err, "group.Wait")
				}
```

packages/relayer/indexer/indexer.go
cyberhorsey: OK, did the rest of the comments, will do this one after the weekend I think!

packages/relayer/processor/process_message.go
```diff
@@ -1,43 +1,86 @@
...
+)
+
+func (p *Processor) eventStatusFromMsgHash(
+	ctx context.Context,
+	gasLimit *big.Int,
+	signal [32]byte,
+) (relayer.EventStatus, error) {
+	var eventStatus relayer.EventStatus
+
+	messageStatus, err := p.destBridge.GetMessageStatus(nil, signal)
```
davidtaikocha, line 36: will suggest adding a timeout for the `ctx` and making `&bind.CallOpts{Context: ctx}` as the first parameter.

dantaik (dismissed):

dantaik:
@cyberhorsey A PR has been merged into main, could you please check if your code needs to be updated?

packages/relayer/processor/process_message.go
cyberhorsey: done, added a new flag for ETHClientTimeout

dantaik (approved):

davidtaikocha (approved):

PR #14606, opened 2023-08-30
feat(protocol): allow Taiko token as L2 fee token

dantaik:
Note that this is only the first step necessary. The bridge also needs to be changed so Ether on L2 will be wrapped as a BridgedERC20 token.

At this point, I think it may not be worthy it to support "Taiko token as the L2 fee token" due to unnecessary implementation complexity. @Brechtpd what do you think?


Brechtpd:
I don't think it's worth it, but I'm also very hesitant of doing it. It's something I think that is both very important and not important at all, at the same time. 

Users in the future will be able to pay in any token using account abstraction, and so then it's only really the proposer that has to handle the L2 fee token. But the proposer needs ETH to pay the L1 gas fees so it will just require additional swaps between ETH and the fee token. So I don't think it adds value to the fee token.

Also adds some non-standard behavior (still a type 1? Apps using msg.value to accept ETH don't work anymore) and extra complexity (bridging between L2s with different fee tokens, also possible msg.value issues, ...).

PR #14611, opened 2023-08-30
fix(protocol): Revert bridged token deployment to Create2Upgradeable

adaki2004:
As it describes:
https://github.com/taikoxyz/taiko-mono/issues/14610

This is a try to solve the issue - I make it draft for now. @cyberhorsey please let me know if it helps (after an upgrade) and if so, i'll make the other ERC token code accordingly + tests tomorrow.

### DIFF
--- a/packages/protocol/contracts/tokenvault/ERC20Vault.sol
+++ b/packages/protocol/contracts/tokenvault/ERC20Vault.sol
@@ -330,23 +330,25 @@ contract ERC20Vault is
         private
         returns (address btoken)
     {
-        ProxiedBridgedERC20 bridgedToken = new ProxiedBridgedERC20();
-
-        btoken = LibVaultUtils.deployProxy(
-            address(bridgedToken),
-            owner(),
-            bytes.concat(
-                bridgedToken.init.selector,
-                abi.encode(
-                    address(_addressManager),
-                    ctoken.addr,
-                    ctoken.chainId,
-                    ctoken.decimals,
-                    ctoken.symbol,
-                    ctoken.name
+        btoken = Create2Upgradeable.deploy(
+            0, // amount of Ether to send
+            keccak256(
+                bytes.concat(
+                    bytes32(ctoken.chainId),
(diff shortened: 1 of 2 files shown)


cyberhorsey:
I have upgraded the L2/L3 vaults and confirms this fixes the issue. Thank you Dani!

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -330,24 +330,26 @@ contract ERC20Vault is
...
-            bytes.concat(
-                bridgedToken.init.selector,
-                abi.encode(
-                    address(_addressManager),
-                    ctoken.addr,
-                    ctoken.chainId,
-                    ctoken.decimals,
-                    ctoken.symbol,
-                    ctoken.name
+        btoken = Create2Upgradeable.deploy(
```
dantaik, line 333: Does this mean all Proxied** contracts are buggy?

cyberhorsey:
> > I have upgraded the L2/L3 vaults and confirms this fixes the issue. Thank you Dani!
> 
> Jeff, i'm not sure if it worked OR Horse was already Birdged somehow and the deploy contract did not even called (Best to try with fresh erc20 contract). Please do (together) some investigation. I opened up another PR #14621 because even if this one works, then this is not an upgradeable BridgedERC20 contract.. so i tried to solve it with that PR i linked but during testing with fresh token fails - tried with Korbi.
> 
> **_UPDATE:_** I was able to validate both of the PRs off-chain (local anvil) with some cheeky hacks (i dont want to commit that ugly raw bytes) and it seems both works so I dont know why it did not work with Korbi's FE!
> 
> Please @cyberhorsey double check and if indeed both works then we should go for the solution: #14621 because it supports upgrade. (if only this one is the working one.. then.. yeah.. sucks)

Just to confirm, it was not HORSE I bridged, but Regular ERC20, which was not working before and worked immediately after the fix and deployed correctly. I will try the new one you recommended today.

PR #14624, opened 2023-08-31
fix(bridge-ui-v2): Fix dialogs being offset

KorbinianK:
### Very interesting bug: 

The CSS rule `backdrop-filter:` (`filter:` as well) on a parent, forces all child elements to be within the bounds of the parent.

This caused dialogs etc to be cut off and centered despite no CSS rule specifying this.

It only happend on dark mode, as we only use the "glass" effect on dark mode which has the filter.

#### Workaround:
apply the style to a pseudo-element..

**Before:**
before

**After:**
after


dantaik (approved):

KorbinianK:
> @KorbinianK @je4nt0nic is the background too bright in dark mode?

For the dialogs you mean? It's according to the design, but yeah I guess if you have large popups like that, it might be too bright. Any suggestions @je4nt0nic ?

xiaodino (approved):

PR #14627, opened 2023-08-31
feat(bridge-ui-v2): Web3Modal theming

KorbinianK:
WIP trying to fix the web3modal using better colors based on light/dark mode


KorbinianK:
closed for now, they are working on v3, let's revisit then

PR #14638, opened 2023-09-02
Update run-a-taiko-node.mdx

anipaul2:
Avoid Using Sudo: Generally, we don't need to use sudo for Docker commands on a personal machine, especially on MacOS. The Docker daemon itself should run with elevated privileges, but individual commands shouldn't require sudo. I removed the sudo and it worked nicely.


MarcusWentz:
Docker might not be setup with admin permission when tested without sudo.

PR #14645, opened 2023-09-04
refactor(protocol): change state.transitions as a ring buffer

dantaik:
`state.transitions` are designed to be a reusable ring buffer, but somehow we missed it in the implementation.


davidtaikocha (dismissed):

adaki2004 (dismissed):

davidtaikocha (dismissed):

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -186,14 +186,21 @@ library TaikoData {
...
         mapping(uint64 blockId_mode_blockRingBufferSize => Block) blocks;
+        // Indexing to transition ids (ring buffer not possible)
         mapping(
             uint64 blockId => mapping(bytes32 parentHash => uint32 transitionId)
             ) transitionIds;
-        mapping(uint64 blockId => mapping(uint32 transitionId => Transition))
-            transitions;
+        // Ring buffer for transitions
+        mapping(
+            uint64 blockId_mode_blockRingBufferSize
```
Brechtpd, line 195:
```suggestion
            uint64 blockId_mod_blockRingBufferSize
```
Or not sure about the meaning of `mode` here.

Brechtpd (dismissed):
I forgot, what was the benefit of having the state transitions separate from the block data?

davidtaikocha (approved):

adaki2004 (approved):

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -86,7 +86,7 @@ library TaikoData {
     /// @dev Struct representing input data for block metadata.
     struct BlockMetadataInput {
         bytes32 txListHash;
-        address beneficiary;
+        address proposer;
```
Brechtpd, line 89: `beneficiary` seems like the better name because it's the fee recipient address, not the proposer?

Brechtpd:
> > I forgot, what was the benefit of having the state transitions separate from the block data?
> 
> The only benefit is that `getBlock` function can return a `Block` struct instead of many small data.
> Screenshot 2023-09-05 at 15 09 45

Ah, then it seems a single mapping could still be possible like before, just using an intermediate struct like this?
```
struct BlockData {
   Block data;
   mapping(uint256 transitionId => Transition) transitions;
}

struct State {
   // Ring buffer for proposed blocks and a some recent verified blocks.
   mapping(uint256 blockId_mode_ringBufferSize => BlockData) blocks;
}
```

Maybe a bit simpler like that, but I guess it doesn't really matter.

PR #14668, opened 2023-09-08
feat(bridge-ui-v2): recallMessage functionality

KorbinianK:
**Work in progress**

the logic got lost along the way, reimplementing it with the new libraries.


packages/bridge-ui-v2/src/libs/proof/Prover.ts
```diff
@@ -1,50 +1,103 @@
...
+    });
+
+    const block = await this._getLatestBlock(destClient, crossChainSyncContract);
+    log('retrieved block', block);
+
+    const clientWithEthProofRequest = destClient as ClientWithEthGetProofRequest;
+
+    if (!block) throw new Error("could not fetch block")
+
+    if (!block || !block.hash) { throw new Error('block  is null') }
```
jscriptcoder, line 61: Maybe we want to delete the line right above, which is already checking if the block is null or undefined?

PR #14675, opened 2023-09-11
test(eventindexer): testing configs

RogerLamTd:


packages/relayer/cmd/flags/common.go
```diff
@@ -126,6 +126,13 @@ var (
 		Value:    10,
 		EnvVars:  []string{"ETH_CLIENT_TIMEOUT"},
 	}
+	CORSOrigins = &cli.StringFlag{
```
cyberhorsey, line 129: im confused why we need this, we have it as an indexer flag already, and it shouldnt be a flag for processor, so it shouldnt be common?

cyberhorsey (dismissed):

packages/relayer/cmd/flags/common.go
RogerLamTd: I saw it listed under commonCategory flags already and processor config did already have CORSOrigins as a flag, I could remove it and make it an indexer only flag?

cyberhorsey (approved):

adaki2004 (approved):

PR #14689, opened 2023-09-13
feat(protocol): Initial draft of hybrid rollup

adaki2004:
Currently draft and needs a LOT of refinement, just opened as a progress. Needs a code walk-through and meeting with Daniel to align more.


packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -35,7 +35,8 @@ contract TaikoL1 is TaikoL1Base {
             proofRegularCooldown: 30 minutes,
             proofOracleCooldown: 15 minutes,
             proofWindow: 90 minutes,
-            proverBond: 10_240e18,
+            proverBondOp: 10_240e18,
```
dantaik, line 38: There is no ZK or OP in the core protocol codebase please. All proofs have a tier, each tier has its own configuration.

dantaik:
The whole idea is about **abstraction**, in the code, we now have different proofs at different tiers, they may be optmistic or ZK, but in the core protocol, we don't use OP or ZK. To config each tier, we implement methods such as `getXXXForTier(...)`

PR #14695, opened 2023-09-15
feat(protocol): Modify LibProposing to accept oracle as assigned prover

adaki2004:
As per request from David:

```
could u open a PR to make our libProposing.sol allow proof assignment to oracle prover? and never charging bond for the oracle prover ?
after a talk with jeff and daniel, we all think its necessary, since we need a prover with unlimited capacity which will be used by our proposer
```

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -70,7 +70,6 @@ library LibProposing {
         // Check prover assignment
         if (
             assignment.prover == address(0)
-                || assignment.prover == LibUtils.ORACLE_PROVER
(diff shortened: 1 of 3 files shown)


davidtaikocha (approved):

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -145,7 +151,9 @@ library LibProposing {
                     );
 
                     // Reward must be minted
-                    tt.mint(input.proposer, reward);
+                    TaikoToken(resolver.resolve("taiko_token", false)).mint(
```
dantaik, line 154: can we use `tt.mint(`?

dantaik (approved):

PR #14700, opened 2023-09-15
feat(bridge-ui-v2): close dialogs with ESC key

KorbinianK:
Some dialogs could not be closed with ESC, some could. Now it should be all of them to be consistent


dantaik (dismissed):

packages/bridge-ui-v2/src/components/Bridge/ProcessingFee/ProcessingFee.svelte
```diff
@@ -44,15 +44,15 @@
       // If so, let's switch to RECOMMENDED method
       selectedFeeMethod = ProcessingFeeMethod.RECOMMENDED;
     }
-
+    removeEscKeyListener();
```
jscriptcoder, line 47: Not sure if we are already closing the modal on component unmounts, but I think it would be also a good idea to remove the listener onDestroy

packages/bridge-ui-v2/src/components/Bridge/Recipient.svelte
```diff
@@ -32,12 +32,13 @@
...
     modalOpen = true;
     addressInput.focus();
+    addEscKeyListener();
   }
 
   function cancelModal() {
     // Revert change of recipient address
     $recipientAddress = prevRecipientAddress;
-
+    removeEscKeyListener();
```
jscriptcoder, line 41: Same here, we could also add this onDestroy, just to make sure that when we are unmounting the whole Bridge component we are also destroying this listener

KorbinianK: I will add the call to onDestroy

packages/bridge-ui-v2/src/components/Bridge/ProcessingFee/ProcessingFee.svelte
KorbinianK: done

dionysuzx (dismissed):

davidtaikocha (dismissed):

dionysuzx (approved):

davidtaikocha (approved):

PR #14704, opened 2023-09-17
chore(bridge-ui-v2): clean amount warning  when left bridge page

xiaodino:
Fix the warning. It should not have a warning when the input text field for amount is empty. 

Screen Shot 2023-09-16 at 10 57 50 PM


dantaik (approved):

KorbinianK:
This does not fix it. The error occurred because the "bridge" action calls clearAmount (and other input clears) after the tx was processed successfully. I fact, it already has been fixed last week and you approved it 😅


https://github.com/taikoxyz/taiko-mono/blob/3ebf0226606b9561ee60fa47ef7d292a3b843678/packages/bridge-ui-v2/src/components/Bridge/Bridge.svelte#L262

KorbinianK (approved):

PR #14715, opened 2023-09-19
feat(protocol): make L2 1559 config upgradable

dantaik:


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -265,7 +279,7 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
         virtual
         returns (EIP1559Config memory)
     {
-        return _eip1559Config;
+        return eip1559Config;
```
davidtaikocha, line 282: See the code below, do we want to change the default min base fee to a higher number?
```solidity
if (_basefee == 0) {
            // To make sure when EIP-1559 is enabled, the basefee is non-zero
            // (Geth never uses 0 values for basefee)
            _basefee = 1;
}
```

davidtaikocha (dismissed):

adaki2004 (dismissed):

davidtaikocha (approved):

adaki2004 (approved):

PR #14721, opened 2023-09-19
feat(protocol): Contestable zkRollup v2 (observations) [DRAFT]

adaki2004:
Here are some observations / bugfixes.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProving.sol
+++ b/packages/protocol/contracts/L1/libs/LibProving.sol
@@ -313,14 +313,12 @@ library LibProving {
                     block.timestamp > tran.timestamp + tier.provingWindow
                         && msg.sender == blk.assignedProver
                 ) revert L1_ASSIGNED_PROVER_NOT_ALLOWED();
-            } else {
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -300,14 +300,19 @@ library LibProving {
             }
 
             if (tid == 1 && tran.prover == blk.assignedProver) {
-                // For the first transition, if the previous prover is still the
-                // assigned prover, we exclusively grant permission to the
-                // assigned approver to re-prove the block, unless the proof
-                // window has elapsed.
+                // For the first transition, (1.) if the previous prover is
+                // still the assigned prover, we exclusively grant permission to
```
dantaik, line 304: a reminder to change line 257

adaki2004: > a reminder to change line 257

To be honest, no i do not think so. We needed at line line 117, but here (line 257), it is like "reProposed" so the time re-starts. I think it is fine as is.

dantaik (approved):

PR #14724, opened 2023-09-19
refactor(protocol): further improve EIP1559 on L2

dantaik:


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -270,16 +215,47 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
     function getEIP1559Config()
         public
-        view
+        pure
         virtual
-        returns (EIP1559Config memory)
+        returns (EIP1559Config memory config)
     {
-        return eip1559Config;
+        config.xscale = 1_488_514_844;
```
dantaik, line 260: @davidtaikocha if we need other values, please change them

davidtaikocha (approved):

adaki2004 (approved):

PR #14739, opened 2023-09-19
fix(docs): fix error in curl command results in Run a Taiko node

chapter087:
The ID specified in the curl command matches the ID displayed in the results. Please correct this as it may be incorrectly identified if the results are incorrect.


d1onys1us:
closed in favor of #14736

PR #14753, opened 2023-09-20
fix(bridge-ui-v2): update InsufficientFunds dialog in mobile view

xiaodino:
Before
Screen Shot 2023-09-19 at 11 02 12 PM

After
Screen Shot 2023-09-19 at 11 48 23 PM

Screen Shot 2023-09-19 at 11 48 45 PM

### DIFF
--- a/packages/bridge-ui-v2/src/components/Transactions/Transaction.svelte
+++ b/packages/bridge-ui-v2/src/components/Transactions/Transaction.svelte
@@ -65,7 +65,7 @@
   {/if}
 
   <div class="sm:w-1/4 md:w-1/5 py-2 flex flex-col justify-center">
-    <Status bridgeTx={item} />
+    <Status bridgeTx={item} isDesktopOrLarger={isDesktopOrLarger} isMobileDetailsDialog={false}/>
   </div>
   <div class="hidden md:flex w-1/5 py-2 flex flex-col justify-center">
     <a
--- a/packages/bridge-ui-v2/src/components/Transactions/Transaction.svelte
+++ b/packages/bridge-ui-v2/src/components/Transactions/Transaction.svelte
@@ -65,7 +65,7 @@
   {/if}
 
   <div class="sm:w-1/4 md:w-1/5 py-2 flex flex-col justify-center">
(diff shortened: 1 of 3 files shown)


packages/bridge-ui-v2/src/components/Transactions/Transaction.svelte
```diff
@@ -65,7 +65,7 @@
   {/if}
 
   <div class="sm:w-1/4 md:w-1/5 py-2 flex flex-col justify-center">
-    <Status bridgeTx={item} />
+    <Status bridgeTx={item} isDesktopOrLarger={isDesktopOrLarger} isMobileDetailsDialog={false}/>
```
KorbinianK, line 68: if we are already passing `isDesktopOrLarger `(should not be passed as prop imho) why do we pass the `isMobileDetailsDialog `as well?

If `isDesktopOrLarger ` is `true` `isMobileDetailsDialog ` will always be `false`. 

I think this needs to done differently

xiaodino: Updated the naming from ```isMobileDetailsDialog``` to ```isFromMobileDetailsDialog```

xiaodino: Trying a different approach.

packages/bridge-ui-v2/src/components/Transactions/Transaction.svelte
```diff
@@ -65,7 +65,7 @@
   {/if}
 
   <div class="sm:w-1/4 md:w-1/5 py-2 flex flex-col justify-center">
-    <Status bridgeTx={item} isDesktopOrLarger={isDesktopOrLarger} isMobileDetailsDialog={false}/>
+    <Status bridgeTx={item} isFromMobileDetailsDialog={false}/>
```
KorbinianK, line 68: A component should be agnostic to where it is used, so if you pass properties they should not be required to be set to false like this. You can do that logic inside the component itself.

Generally, we need to rework the way we claim on mobile, it opens the dialog and then instantly tries to claim. That is not how it should work.
Maybe just close this PR, as this is a hacky workaround instead of a proper fix for the underlying problem. I will work on it.

KorbinianK:
https://github.com/taikoxyz/taiko-mono/pull/14759

PR #14761, opened 2023-09-20
fix(protocol): allow prover assignment with different proving fees

dantaik:
In the off-chain prover assignment, now there will be a list of fees per tier, then when the min tier is chosen for a block, the fee is picked from the list.


packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -84,10 +82,17 @@ library TaikoData {
...
+        uint256 fee;
+    }
+
     struct ProverAssignment {
         address prover;
+        address feeToken;
+        TierFee[] tierFees;
         uint64 expiry;
-        bytes data;
+        bytes signature;
```
adaki2004, line 95: With the rename of data -> signature it seems (based on the renaming only) it does not contain the 'tokenAmount' (for `IProver`) if custom ERC20. 
(This data tho still can be incorporate into the `signature` by `abi.encode(signature, tokenAmount)` so maybe best to use `data` otherwise it can be confusing)

packages/protocol/contracts/libs/LibAddress.sol
```diff
@@ -43,4 +40,24 @@ library LibAddress {
...
+        bytes memory sig
+    )
+        internal
+        view
+        returns (bool valid)
+    {
+        if (
+            AddressUpgradeable.isContract(addr)
+                && IERC1271Upgradeable(addr).isValidSignature(hash, sig)
+                    == EIP1271_MAGICVALUE
```
adaki2004, line 56: Currently this "merges" the IProver and IERC1271 together, right ?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -289,4 +231,87 @@ library LibProposing {
...
+        if (!assignment.prover.isValidSignature(hash, assignment.signature)) {
+            revert L1_ASSIGNMENT_INVALID_SIG();
+        }
+
+        // Find the prover fee using the minimal tier
+        proverFee = _getProverFee(assignment.tierFees, minTier);
+
+        // The proposer irrevocably pays a fee to the assigned prover, either in
+        // Ether or ERC20 tokens.
+        if (assignment.feeToken == address(0)) {
```
adaki2004, line 284: So right now, we also support ERC20 for 'single, EOA provers' not contracts-only ? I like this design.

packages/protocol/contracts/libs/LibAddress.sol
dantaik: Kind of. It's less flexible than the previous one but it works with the tier-based prices better.

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: Yes.

adaki2004 (approved):
LGTM (I will adjust the tests targetting the other branch ‘extra_tier_config’)

PR #14766, opened 2023-09-20
fix(protocol): Hybrid rollup tests and small fixes

adaki2004:
Please find attached the code for broken tests (and new tests as well).

(NOTE: Test `test_L2_getBasefee` still failing.)

2 additions to the code:
- Add back id to metadata (bc. it is needed (to know) when a block is proved for a specific metadata so easier to keep it there instead of off-chain indexing)
- in `LibProving.sol` : ` if (tier.contestBond == 0)` is placed outside of `if (evidence.tier == tran.tier)` because the `tier` variable (line 169) is determined by `evidence.tier.` Which means if` evidence.tier` is `GUARDIAN`, but `tran.tier` cannot be `GUARDIAN` unless it is already `GUARDIAN` :) so it was a chicken and egg. (Now GUARDIAN can prove without a contest)


packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -272,16 +273,16 @@ library LibProposing {
 
     /// @dev Hashing the block metadata.
     function _hashMetadata(TaikoData.BlockMetadata memory meta)
-        private
+        internal
```
dantaik, line 276: For internal methods, lets remove the leading "_"

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -272,16 +273,16 @@ library LibProposing {
...
         returns (bytes32 hash)
     {
         uint256[5] memory inputs;
         inputs[0] = uint256(meta.l1Hash);
         inputs[1] = uint256(meta.mixHash);
         inputs[2] = uint256(meta.txListHash);
-        inputs[3] = (uint256(meta.timestamp) << 192)
-            | (uint256(meta.l1Height) << 128) | (uint256(meta.gasLimit) << 96);
+        inputs[3] = (uint256(meta.id)) | (uint256(meta.timestamp) << 64)
+            | (uint256(meta.l1Height) << 128) | (uint256(meta.gasLimit) << 192);
```
dantaik, line 285: `(uint256(meta.gasLimit) << 192` to `(uint256(meta.gasLimit) << 160`? as gasLimit is a uint32

packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -124,4 +125,24 @@ library LibUtils {
             lastVerifiedBlockId: b.lastVerifiedBlockId
         });
     }
+
+    function getInstance(
```
dantaik, line 129: Now the "instance" is only an internal concept to PSE-ZKEVM, It should be removed from libs/*.sol.

dantaik: I suggest to move it to PseZkVerifer.sol

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -144,8 +144,6 @@ library LibVerifying {
                 blk = state.blocks[slot];
                 if (blk.blockId != blockId) revert L1_BLOCK_MISMATCH();
 
-                tid = LibUtils.getTransitionId(state, blk, slot, blockHash);
```
dantaik, line 147: Why delete this? Before a block can be verified, you need to find the right trnasition, you need to look up the tid based the parent hash (`blockHash` in the code).

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -202,7 +202,41 @@ library LibProving {
...
+            emit TransitionProved({
+                blockId: blk.blockId,
+                parentHash: evidence.parentHash,
+                blockHash: evidence.blockHash,
+                signalRoot: evidence.signalRoot,
+                prover: msg.sender,
+                proofBond: 0,
+                tier: evidence.tier
+            });
+        } else if (evidence.tier == tran.tier) {
```
dantaik, line 239: line 240 and 253 needs to be updated to remove "top tier" related content.

packages/protocol/contracts/L1/libs/LibVerifying.sol
dantaik: Should we split LibVerifying into one more file called LibGenesis.sol and bring the init function there.

dantaik (changes requested):

packages/protocol/contracts/L1/libs/LibProposing.sol
adaki2004: Yes, because 128 is position where the `l1Height` starts and since `l1Height` is 64, so `gasLimit` can only start at 192.

packages/protocol/contracts/L1/libs/LibUtils.sol
adaki2004: Ok, i needed for testing. Anyways i put this function in the `TestBase` too then.

packages/protocol/contracts/L1/libs/LibVerifying.sol
adaki2004: Yep thanks for catching! 
Sorry was a test issue and i removed the wrong line ! (Fixed, commited soon)
I do not think we should split this file, but as you like!

dantaik (approved):

packages/protocol/test/L1/TaikoL1TestBase.sol
```diff
@@ -188,6 +203,26 @@ abstract contract TaikoL1TestBase is TestBase {
         );
     }
 
+    function getInstance(
```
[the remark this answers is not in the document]
adaki2004: Yes i was wondering. If this is ok by you to make it public, sure let's do it!

PR #14767, opened 2023-09-20
feat(relayer): fix cost calculation for isProfitable

RogerLamTd:


packages/relayer/processor/process_message.go
```diff
@@ -243,11 +245,11 @@ func (p *Processor) sendProcessMessageCall(
 		auth.GasLimit = 3000000
 	} else {
 		// otherwise we can estimate gas
-		gas, cost, err = p.estimateGas(ctx, event.Message, proof)
+		gas, _, err = p.estimateGas(ctx, event.Message, proof)
```
cyberhorsey, line 248: instead of `_`, lets just remove `cost` from being returned by this function now.

packages/relayer/processor/process_message.go
```diff
@@ -243,11 +245,11 @@ func (p *Processor) sendProcessMessageCall(
...
 		// and if gas estimation failed, we just try to hardcore a value no matter what type of event,
 		// or whether the contract is deployed.
 		if err != nil || gas == 0 {
-			cost, err = p.hardcodeGasLimit(ctx, auth, event, eventType, canonicalToken)
+			_, err = p.hardcodeGasLimit(ctx, auth, event, eventType, canonicalToken)
```
cyberhorsey, line 252: same here, this function can just return `err` now.

packages/relayer/processor/process_message.go
```diff
@@ -461,3 +457,47 @@ func (p *Processor) saveMessageStatusChangedEvent(
...
+				return errors.Wrap(err, "p.destBridge.SuggestGasPrice")
+			}
+			auth.GasPrice = gasPrice
+
+			return nil
+		}
+	} else {
+		auth.GasTipCap = gasTipCap
+
+		return nil
```
cyberhorsey, line 478: dont think this return nil is needed, since it just returns nil anyuways below this statement.

packages/relayer/processor/process_message.go
```diff
@@ -461,3 +457,47 @@ func (p *Processor) saveMessageStatusChangedEvent(
...
+	}
+
+	return nil
+}
+
+func (p *Processor) getCost(ctx context.Context, auth *bind.TransactOpts) (*big.Int, error) {
+	if auth.GasPrice == nil {
```
cyberhorsey, line 485: its usually better to do != nil checks, so i would change this to
```go
if auth.GasTipCap != nil
```

cyberhorsey (dismissed):

packages/relayer/processor/process_message.go
```diff
@@ -243,11 +245,11 @@ func (p *Processor) sendProcessMessageCall(
 		auth.GasLimit = 3000000
 	} else {
 		// otherwise we can estimate gas
-		gas, cost, err = p.estimateGas(ctx, event.Message, proof)
+		gas, err = p.estimateGas(ctx, event.Message, proof)
 		// and if gas estimation failed, we just try to hardcore a value no matter what type of event,
 		// or whether the contract is deployed.
 		if err != nil || gas == 0 {
```
davidtaikocha, line 251: shall we print this `err` in logs here? or its safe to ignore?

packages/relayer/processor/process_message.go
```diff
@@ -461,3 +452,45 @@ func (p *Processor) saveMessageStatusChangedEvent(
...
+}
+
+func (p *Processor) getCost(ctx context.Context, auth *bind.TransactOpts) (*big.Int, error) {
+	if auth.GasTipCap != nil {
+		bn, err := p.destEthClient.BlockNumber(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		blk, _ := p.destEthClient.BlockByNumber(context.Background(), new(big.Int).SetUint64(bn))
```
davidtaikocha, line 484: This error will never happen so it can be ignored?

packages/relayer/processor/process_message.go
```diff
@@ -256,20 +258,14 @@ func (p *Processor) sendProcessMessageCall(
 		}
 	}
 
-	gasTipCap, err := p.destEthClient.SuggestGasTipCap(ctx)
+	err = p.setGasTipOrPrice(ctx, auth)
```
davidtaikocha, line 261:
```go
if err = p.setGasTipOrPrice(ctx, auth); err != nil {
  // ...
}
```
will save u one line of code :) but yeah its not necessary.

packages/relayer/processor/process_message.go
RogerLamTd: if its not nil then hardcode so safe to ignore?

packages/relayer/processor/process_message.go
RogerLamTd: adding the check, nice spot

davidtaikocha: and i see u fetch the latest head height, and query the block by that number, seems they are the same?

```go
blk, err := p.destEthClient.BlockByNumber(ctx, nil)
```

davidtaikocha:
```suggestion
		blk, _ := p.destEthClient.BlockByNumber(ctx, new(big.Int).SetUint64(bn))
```

can't understand why use `context.Background()` here

packages/relayer/processor/process_message.go
```diff
@@ -461,3 +451,48 @@ func (p *Processor) saveMessageStatusChangedEvent(
...
+		}
+
+		blk, err := p.destEthClient.BlockByNumber(context.Background(), new(big.Int).SetUint64(bn))
+		if err != nil {
+			return nil, err
+		}
+
+		cfg := params.NetworkIDToChainConfigOrDefault(p.destChainId)
+
+		baseFee := eip1559.CalcBaseFee(cfg, blk.Header())
```
davidtaikocha, line 490: whats the difference between this and `blk.BaseFee()`?

RogerLamTd: eip1559.CalcBaseFee calculates basefee based on gasUsed of the previous block

davidtaikocha: emmm for L2 => L1 relayer i think this will work, but for L1 => L2 relayer, u should try `TaikoL2.getBasefee`?

packages/relayer/processor/process_message.go
RogerLamTd: adding slog.Info for debugging purposes

packages/relayer/processor/process_message.go
```diff
@@ -461,3 +452,55 @@ func (p *Processor) saveMessageStatusChangedEvent(
...
+			gasPrice, err := p.destEthClient.SuggestGasPrice(context.Background())
+			if err != nil {
+				return errors.Wrap(err, "p.destBridge.SuggestGasPrice")
+			}
+			auth.GasPrice = gasPrice
+		}
+	} else {
+		auth.GasTipCap = gasTipCap
+
+		return nil
```
cyberhorsey, line 471: this was marked as resolved but i dont think it was fixed, i think this `return nil` is extraneous, 

i think thi sfunction can be:

```go
func (p *Processor) setGasTipOrPrice(ctx context.Context, auth *bind.TransactOpts) error {
	gasTipCap, err := p.destEthClient.SuggestGasTipCap(ctx)
	if err != nil {
		if IsMaxPriorityFeePerGasNotFoundError(err) {
			auth.GasTipCap = FallbackGasTipCap
		} else {
			gasPrice, err := p.destEthClient.SuggestGasPrice(context.Background())
			if err != nil {
				return errors.Wrap(err, "p.destBridge.SuggestGasPrice")
			}
			auth.GasPrice = gasPrice
		}
		return nil
	} 

	auth.GasTipCap = gasTipCap

	return nil
```

eliminating the else clause completely

packages/relayer/processor/process_message.go
```diff
@@ -461,3 +452,55 @@ func (p *Processor) saveMessageStatusChangedEvent(
...
+}
+
+func (p *Processor) getCost(ctx context.Context, auth *bind.TransactOpts) (*big.Int, error) {
+	if auth.GasTipCap != nil {
+		blk, err := p.destEthClient.BlockByNumber(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		cfg := params.NetworkIDToChainConfigOrDefault(p.destChainId)
```
cyberhorsey, line 484: we only need to use this in the `else` clause below, its not used in the `if p.taikoL2 != nil` clause, so we can move it there, above `baseFee = eip1559.CalcBaseFee`.

packages/relayer/processor/process_message.go
```diff
@@ -461,3 +452,55 @@ func (p *Processor) saveMessageStatusChangedEvent(
...
+
+		var baseFee *big.Int
+
+		if p.taikoL2 != nil {
+			gasUsed := uint32(blk.GasUsed())
+			timeSince := uint64(time.Since(time.Unix(int64(blk.Time()), 0)))
+			baseFee, err = p.taikoL2.GetBasefee(&bind.CallOpts{}, timeSince, gasUsed)
```
cyberhorsey, line 491: we should pass the ctx in `bind.CallOpts` like this:
```go
&bind.CallOpts{
Context: ctx,
}
```

cyberhorsey (dismissed):

cyberhorsey (approved):

packages/relayer/processor/process_message.go
```diff
@@ -461,3 +452,52 @@ func (p *Processor) saveMessageStatusChangedEvent(
...
+func (p *Processor) getCost(ctx context.Context, auth *bind.TransactOpts) (*big.Int, error) {
+	if auth.GasTipCap != nil {
+		blk, err := p.destEthClient.BlockByNumber(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		var baseFee *big.Int
+
+		if p.taikoL2 != nil {
```
davidtaikocha, line 484: So when starting a L2 => L1 relayer, `p.taikoL2` will be `nil`?

packages/relayer/processor/process_message.go
```diff
@@ -461,3 +452,52 @@ func (p *Processor) saveMessageStatusChangedEvent(
...
+			gasUsed := uint32(blk.GasUsed())
+			timeSince := uint64(time.Since(time.Unix(int64(blk.Time()), 0)))
```
davidtaikocha, line 486: `blk.Time()` since `0`?

RogerLamTd: This achieves time.Since() gives time.Now().Sub(y), this achieves time.Now() - blk.Time(). The 0 is the second param to time.Unix(), should work as intended!

davidtaikocha (approved):

PR #14769, opened 2023-09-21
fix(bridge-ui-v2): Add z-index for close button on mobile

xiaodino:
The ChainSelect dialogs and some dialogs are not being able to be closed on mobile. 

After following https://ymoondhra.medium.com/how-to-run-localhost-on-your-iphone-4110a54d1896 to set up localhost on iPhone, use Safari iPhone web inspector to debug. It should be fixed with this PR.

### DIFF
--- a/packages/bridge-ui-v2/src/components/Bridge/ProcessingFee/ProcessingFee.svelte
+++ b/packages/bridge-ui-v2/src/components/Bridge/ProcessingFee/ProcessingFee.svelte
@@ -157,7 +157,7 @@
 
   <dialog id={dialogId} class="modal" class:modal-open={modalOpen}>
     <div class="modal-box relative px-6 py-[35px] md:rounded-[20px] bg-neutral-background">
-      <button class="absolute right-6 top-[35px]" on:click={cancelModal}>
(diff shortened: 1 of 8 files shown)


KorbinianK:
When you start vite with the `--host` flag, you can access the local instance in your network :)

packages/bridge-ui-v2/src/components/Bridge/ProcessingFee/ProcessingFee.svelte
```diff
@@ -157,7 +157,7 @@
 
   <dialog id={dialogId} class="modal" class:modal-open={modalOpen}>
     <div class="modal-box relative px-6 py-[35px] md:rounded-[20px] bg-neutral-background">
-      <button class="absolute right-6 top-[35px]" on:click={cancelModal}>
```
KorbinianK, line 160: Looking at all the close buttons, they are "almost" identical, when they really should be "completely" identical. Let's deploy this fix, but I think we should have a separate Dialog Component as a future improvement

KorbinianK (approved):

yueawang (approved):

dantaik (approved):

PR #14783, opened 2023-09-22
fix(protocol): Fix tests

adaki2004:
Fix failing tests.


dantaik (approved):

dantaik:
one test still fails

adaki2004:
> one test still fails

Yes, that is the `test_L2_getBasefee` - i will have a look, but i am less familiar with that one so it might take some time.

adaki2004:
> > one test still fails
> 
> Yes, that is the `test_L2_getBasefee` - i will have a look, but i am less familiar with that one so it might take some time.

This function returns 0 is the issue:
The supplied function parameters in the test above are:
 **config.xscale: 1488514844
  config.xscale: 1488514844
  _gasExcess: 0
  _basefee: 0**
  
```
  function calculatePrice(
          uint128 xscale,
          uint128 yscale,
          uint64 xExcess,
          uint64 xPurchase
      )
          internal
          pure
          returns (uint256)
      {
          assert(xscale != 0 && yscale != 0);
          uint64 _xPurchase = xPurchase == 0 ? 1 : xPurchase;
          uint256 _before = _calcY(xExcess, xscale);
          uint256 _after = _calcY(xExcess + _xPurchase, xscale);
          return (_after - _before) / _xPurchase / yscale;
      }
```

@dantaik maybe do you know by heart what is wrong with the inputs ? Should not these need to be recalculated somewhere (in the `test_L2_getBasefee` ?)

PR #14810, opened 2023-09-25
chore(protocol): Contestable rollup additions (design docs, documentation, etc.)

adaki2004:
As per title.


dantaik:
Can I suggest to also submit the source file for the diagram?

adaki2004:
@d1onys1us let me know if i shall just throw out the files under `contract-documentation` as it is auto-generated (i edited by hand when i realized .. )

@cyberhorsey just tagging you cause some events changed too so it might effect indexer/relayer.

packages/website/pages/docs/concepts/proving.mdx
```diff
@@ -3,11 +3,11 @@ import { Callout } from "nextra-theme-docs";
...
+The purpose of proving blocks is to give certainty to bridges about the execution that happened in the rollup. To rely on some state that happened inside of the rollup, a bridge will want a proof that everything was done correctly. On Taiko you can run a node as a prover and prove blocks, permissionlessly. This means that you can examine the proposed blocks on the TaikoL1 contract, and generate proofs for them. Currently, any prover can create proofs for proposed blocks. This means that the number of "state transitions" has no upper bound, because we don't know what is the correct state transition yet. Only first prover with a valid proof of the correct state transition will receive the reward of `ETH` (and possibly any `ERC20` if prover pool implementation favors it).
```
dantaik, line 6:
```suggestion
The purpose of proving blocks is to give certainty to bridges about the execution that happened in the rollup. To rely on some state that happened inside of the rollup, a bridge will want a proof that everything was done correctly. On Taiko you can run a node as a prover and prove blocks, permissionlessly. This means that you can examine the proposed blocks on the TaikoL1 contract, and generate proofs for them. Currently, any prover can create proofs for proposed blocks. This means that the number of "state transitions" has no upper bound, because we don't know what is the correct state transition yet. Only first prover with a valid proof of the correct state transition will receive Ether or ERC20 token as the proving fee.
```

dantaik (approved):

dantaik (approved):

dantaik:
@d1onys1us @MarcusWentz could you please take a look at the new/modified docs?

dantaik (approved):

PR #14811, opened 2023-09-25
feat(bridge-ui-v2): NFT bridge stepper

KorbinianK:
Stepper component that will be used for the NFT bridge and in the next update for ERC20

image

image

image


dantaik:
If the last step has finished, should we simply hide the step widget because the widget below will display the final status, no?

KorbinianK:
> If the last step has finished, should we simply hide the step widget because the widget below will display the final status, no?

I wouldn't hide it, as it will shift and change the entire layout. It also resembles a timeline and flow the user went through. And a stepper doesn't make too much sense if it is realistically only for one additional step, as the first one is already "selected".

What do you think @je4nt0nic?

dantaik (approved):

jscriptcoder (approved):

jscriptcoder:
Nice job on the stepper @KorbinianK. This definitely looks better than what we currently have in the bridge.

xiaodino (approved):

PR #14821, opened 2023-09-26
feat(protocol): emit events in LibTaikoToken

dantaik:


adaki2004:
In protocol/package.json i thin the `test` command shall look like:
`"test": "forge test -vvv --gas-report --fuzz-seed $(date +%s) --match-path test/**/*.t.sol",`

(Otherwise it seems it tries to run simulation tests which already removed)

adaki2004:
> In protocol/package.json i think the `test` command shall look like: `"test": "forge test -vvv --gas-report --fuzz-seed $(date +%s) --match-path test/**/*.t.sol",`
> 
> (Otherwise it seems it tries to run simulation tests which already removed)

Pushed to base branch.

davidtaikocha (approved):

adaki2004 (approved):

PR #14823, opened 2023-09-26
fix(protocol): Foundry has an issue with the seed

adaki2004:
A trial branch to investigate the issue: https://github.com/foundry-rs/foundry/issues/5913


adaki2004:
@davidtaikocha is it possible to update cargo in ci config ?
(I conclude something like this happens to us as well: https://stackoverflow.com/questions/72896871/cargo-test-keeps-failing-on-github-actions-but-passes-on-local-machine)

PR #14830, opened 2023-09-26
feat(protocol): Guardian can return unprovable bonds

adaki2004:


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -219,6 +219,15 @@ library LibProving {
                 revert L1_INVALID_EVIDENCE();
             }
 
+            if (
+                evidence.proof.length != 0
+                    && evidence.proof[0] == keccak256("UNPROVABLE")
```
dantaik, line 224: How about change `UNPROVABLE` to `RETURN_LIVENESS_BOND`

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -219,6 +219,15 @@ library LibProving {
                 revert L1_INVALID_EVIDENCE();
             }
 
+            if (
+                evidence.proof.length != 0
```
dantaik, line 223:
```suggestion
                evidence.proof.length == 1
```

adaki2004: I think != 0 is easier because then in the GuardianVerifier.sol it can stay as is - if for whatever reason the guardian is not returning the liveness bond. What do you think ?

dantaik: In GuardianVerifier.sol we can simply remove "if (evidence.proof.length != 0) revert INVALID_PROOF();"

packages/protocol/contracts/L1/verifiers/GuardianVerifier.sol
```diff
@@ -38,7 +38,10 @@ contract GuardianVerifier is EssentialContract, IVerifier {
         external
         view
     {
-        if (evidence.proof.length != 0) revert INVALID_PROOF();
+        if (
```
dantaik, line 41: How about we remove this first If-statement

PR #14835, opened 2023-09-27
feat(protocol): update `DeployOnL1` script based on Contestable zkRollup changes

davidtaikocha:


adaki2004 (approved):

packages/protocol/contracts/L1/tiers/OptimisticRollupConfigProvider.sol
```diff
@@ -80,3 +81,10 @@ contract OptimisticRollupConfigProvider is ITierProvider {
         else return LibTiers.TIER_OPTIMISTIC;
     }
 }
+
+/// @title ProxiedOptimisticRollupConfigProvider
+/// @notice Proxied version of the parent contract.
+contract ProxiedOptimisticRollupConfigProvider is
```
dantaik, line 87: David, I don't think we will ever need to upgrade this contract. We can simply set "tier_provider" address to a different address.

dantaik: All these config providers are also meant to be *stateless*.

davidtaikocha: Changed: https://github.com/taikoxyz/taiko-mono/commit/2c66a941eec88c0a729d0c384fb24a1000edbe5a

dantaik (approved):

PR #14838, opened 2023-09-27
feat(protocol): update A5 1559 params

davidtaikocha:
At this point `gasIssuedPerSecond` is `2_666_666` (sell `8m * 4` gas per 12 seconds), change it to `666_666` (sell `8m` gas per 12 seconds)


adaki2004 (approved):
Lgtm

adaki2004:
To overcome the issue of not having an owner (hence pre-deployed contract so EssentialContract._init() never called) we need a mechanism to introduce a new owner via proxy upgrades.
With this solution, the call of `updateEIP1559Config` works but tests obviously fails (which calls the `TaikoL2._init()`)

dantaik (approved):

dantaik (approved):

PR #14848, opened 2023-09-28
chore(bridge-ui-v2): Add isDeployedCrossChain.test.ts

xiaodino:


packages/bridge-ui-v2/src/libs/token/isDeployedCrossChain.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      srcChainId,
+      destChainId,
+    });
+
+    // Then
+    expect(result).toEqual(true);
+  });
+
+  it('should return false if not configured', async () => {
+    mockToken.addresses[destChainId] = zeroAddress;
```
KorbinianK, line 52: The reason this test works is because `getCrossChainAddress() `reads the configured destinationAddress (`zeroAddress` in this case) and treats it as valid. This is actually an error and should be fixed.

[this line](https://github.com/taikoxyz/taiko-mono/blob/95337bce36c5e9e4b07955501baadec8ddf93a2c/packages/bridge-ui-v2/src/libs/token/getCrossChainAddress.ts#L35) should also check if it is `!zeroAddress`.

You are only testing preconfigured tokens that have some form of address at the destination. What we actually need it for, is when we don't have that address. We check if we can find one via `getCrossChainAddress(),` if we do, it is deployed. So the test should also handle tokens without any destination address set.

dantaik (approved):

dantaik (dismissed):

packages/bridge-ui-v2/src/libs/token/isDeployedCrossChain.test.ts
dantaik: @xiaodino

packages/bridge-ui-v2/src/libs/token/getCrossChainAddress.ts
```diff
@@ -95,6 +95,13 @@ export async function getCrossChainAddress({
 
     // return await destTokenVaultContract.read.canonicalToBridged([BigInt(destChainId), bridgedAddress]);
   } else {
+    if (!(srcChainId in routingContractsMap) || !(destChainId in routingContractsMap[srcChainId])) {
```
jscriptcoder, line 98: Nitpick, how about?
```ts
if (!routingContractsMap?.[srcChainId]?.[destChainId]) { ... }
```
does the same job and looks cleaner. Just a suggestion

packages/bridge-ui-v2/src/libs/token/isDeployedCrossChain.test.ts
xiaowheat: @KorbinianK  Line 31 and 32 have checked  if it is ```!zeroAddress```.

packages/bridge-ui-v2/src/libs/token/getCrossChainAddress.ts
xiaowheat: Updated. Thanks

dantaik:
After a second thought, I think I perfer close this PR without merging, and only bring this feature back when we are ready to use it.

PR #14849, opened 2023-09-28
refactor(protocol): rework on L2 1559 (AMM and Standard impl) - DO NOT MERGE

dantaik:


davidtaikocha:
> @davidtaikocha could you help me fix a test in GenerateGenesis.g.sol? My docker is not working correctly. (you can search TODO in that file).

will do

Brechtpd:
The vanilla EIP-1559 formula depends on fixed block times which we don't enforce for our L2. The other reason for not using the normal formula was because we'd need to support dynamically sized blocks in an efficient way, which I guess with continuations/chunked prover we hopefully can do. But I would still say the current formula is much nicer and completely independent of block times or block sizes.

How was the current implementation buggy? I did think the implementation was more complex than necessary with the scaling which I didn't think was necessary, but the approximated formula and formula worked correctly with the more straight forward implementation that was used for the proving fee at the time, so seems like the main code should have worked.

dantaik:
> The vanilla EIP-1559 formula depends on fixed block times which we don't enforce for our L2. The other reason for not using the normal formula was because we'd need to support dynamically sized blocks in an efficient way, which I guess with continuations/chunked prover we hopefully can do. But I would still say the current formula is much nicer and completely independent of block times or block sizes.
> 
> How was the current implementation buggy? I did think the implementation was more complex than necessary with the scaling which I didn't think was necessary, but the approximated formula and formula worked correctly with the more straight forward implementation that was used for the proving fee at the time, so seems like the main code should have worked.

The current approach has bugs, but I didn't manage to figure out exactly where the bug is. The `exp` op is also hard to debug. 

I agree on the time issues you mentioned, I'm actually working on a Uniswap AMM style pool to offer gas for purchase per block and refill the gas based on block time. I think that version will be better than this vanilla version and is much easier to debug without the `exp` ops.

Brechtpd:
I am still strongly in favor of using exp because that will actually be the most vanilla EIP-1559 formula available in Ethereum, because the exp based formula will be used for the base fee calculation for the data blobs in EIP-4844. So that means that this mechanism is already part of Ethereum, just in a different part: https://eips.ethereum.org/EIPS/eip-4844#blob-gasprice-update-rule. It doesn't seem worth it to me to make yet another EIP-1559 style mechanism, especially because the exp based one is a better way to do it than the normal EIP-1559 even without taking into account our reasons why we can't use the vanilla one.

The formula used in EIP-4844 is slightly different and `exp` is approximated in a different way but I didn't think it was worth it to change it it basically the same, but if the code is being changed anyway than it makes sense to me to just stick to the EIP-4844 code as close as possible.

davidtaikocha:
workflow errors fixed in the latest commit.

PR #14850, opened 2023-09-28
chore(bridge-ui-v2): pre-select destination chain

KorbinianK:
If we only have two chains, the other chain should be selected as destination by default

### DIFF
--- a/packages/bridge-ui-v2/src/libs/chain/chains.ts
+++ b/packages/bridge-ui-v2/src/libs/chain/chains.ts
@@ -20,6 +20,14 @@ function mapChainConfigToChain(chainId: string, chainConfig: ChainConfigMap[numb
   };
 }
 
+export const chainIdToChain = (chainId: number): Chain => {
+  const chain = chains.find((chain) => chain.id === chainId);
+  if (!chain) {
+    throw new Error(`Chain with id ${chainId} not found`);
+  }
+  return chain;
+};
+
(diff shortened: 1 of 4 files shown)


packages/bridge-ui-v2/src/libs/chain/chains.ts
```diff
@@ -20,6 +20,14 @@ function mapChainConfigToChain(chainId: string, chainConfig: ChainConfigMap[numb
   };
 }
 
+export const chainIdToChain = (chainId: number): Chain => {
+  const chain = chains.find((chain) => chain.id === chainId);
+  if (!chain) {
+    throw new Error(`Chain with id ${chainId} not found`);
```
jscriptcoder, line 26: This exception doesn't seem to be handled anywhere. Maybe we want to catch it and present the user with a error toast?

dantaik (approved):

xiaodino (approved):

packages/bridge-ui-v2/src/libs/chain/chains.ts
KorbinianK: This error is at least currently impossible to reach as the only place this util is used is inside the chainselector wrapper. 
There we specifically pick a chain from our configured chains and not any from the user, so adding more code such as a toast that will never appear is a bit redundant I think.

If any UI component will use it for something else in the future, the error needs to be caught there though.

PR #14851, opened 2023-09-28
feat(protocol): support SGX proofs

adaki2004:
First PoC - basically verifier does not do attestation verification on-chain - just a signature check.

A set of initial ( `addToRegistry()` ) trusted enclaves identified uniquely by their (initial) ECDSA public key can sign data. 

In our case exactly this:

```
return keccak256(
            abi.encode(
                evidence.metaHash,
                evidence.parentHash,
                evidence.blockHash,
                evidence.signalRoot,
                evidence.graffiti,
                assignedProver,
                newPubKey
            )
        );
```

Once public key is used by an enclave another has to be used next time.

Let me know if this is what we discussed or not.


packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,101 @@
...
+        address currentInstancePubKey,
+        address newPubKey
+    )
+        public
         pure
+        returns (bytes32 signedHash)
     {
-        revert UNIMPLEMENTED();
+        return keccak256(
+            abi.encode(
```
Brechtpd, line 95: Adding the prover here still makes sense I nkow think

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,101 @@
...
+
+        bytes32 signedInstance = getSignedHash(evidence,proofData.currentPubKey, proofData.newPubKey);
+
+        if(!proofData.currentPubKey.isValidSignature(signedInstance, proofData.signature)) {
+            revert SGX_INVALID_PROOF_SIGNATURE();
+        }
+
+        // Invalidate current key, because it cannot be used again (side-channel attacks).
+        sgxRegistry[proofData.currentPubKey] = false;
+        sgxRegistry[proofData.newPubKey] = true;
```
Brechtpd, line 82: Some optimizations possibilities probably, but not important now.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,101 @@
...
 
-    error UNIMPLEMENTED();
+    struct ProofData {
+        address currentPubKey;
+        address newPubKey;
+        bytes signature;
+    }
+    
+    /// @dev One SGX instance is uniquely identified (on-chain) by it's ECDSA public key. Once that public key is used (by proof verification) it has to be overwritten by a new one (representing the same instance). This is due to side-channel protection.
+    mapping(address sgxInstance => bool truested) sgxRegistry;
```
Brechtpd, line 28:
```suggestion
    mapping(address sgxInstance => bool trusted) sgxRegistry;
```

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,101 @@
...
 
     /// @notice Initializes the contract with the provided address manager.
     /// @param _addressManager The address of the address manager contract.
     function init(address _addressManager) external initializer {
         EssentialContract._init(_addressManager);
     }
 
+    /// @notice Adds trusted SGX instances to the registry.
+    /// @param trustedInstances The address array of trusted SGX instances.
+    function addToRegistry(address[] memory trustedInstances) external onlyOwner {
```
Brechtpd, line 43: Adding new SGX addresses by already registered addresses still TODO, or do you think this should be implemented in the owner contract? I would just add it here directly so it's always available.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,101 @@
...
+        bool isContesting,
+        TaikoData.BlockEvidence calldata evidence
     )
         external
+    {
+        // Do not run proof verification to contest an existing proof
+        if (isContesting) return;
+
+        // At least (worst case) the 2 x 32 bytes shall be encoded (current and new pubKey)
+        if(evidence.proof.length < 64) {
```
Brechtpd, line 64: May as well add the signature length and do a hard equality check?

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
adaki2004: Yeah, i thought we agree in the latter (`onlyOwner` kind of style) but can easily add it.
But question is:
When an already registered SGX address is trying to add a new one: Is it adding a new (like new instance) address ? or it is changing the address of itself ?
Since i think SGX cannot fire away transactions, it should happen via signatures.. If via signatures -> then it would need to set his instance a new one as well (bc. of the attack vector you told me).

Brechtpd: Ah I thought we ended up agreeing to do both haha

But yeah for the register, it would be adding a new address (with a new ID) and then similar to the proof verification, overwrite the address used by the signature with a new one.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,156 @@
...
+            sgxRegistry[uniqueVerifiers] = trustedInstances[i];
+
+            emit InstanceAdded(uniqueVerifiers, trustedInstances[i]);
+
+            uniqueVerifiers++;
+        }
+
+        // Invalidate current key, because it cannot be used again (side-channel
+        // attacks).
+        sgxRegistry[instanceId] = newPubKey;
```
Brechtpd, line 98: This code and 
```
if (sgxRegistry[instanceId] != signer) {
    revert SGX_NOT_VALID_SIGNER_OR_ID_MISMATCH();
}
```

I think easy to extract in it's own function so it can be reused by both `addToRegistryBySgxInstance` and `verifyProof`.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,156 @@
...
+        external
+        onlyOwner
+    {
+        for (uint256 i; i < trustedInstances.length; i++) {
+            sgxRegistry[uniqueVerifiers] = trustedInstances[i];
+
+            emit InstanceAdded(uniqueVerifiers, trustedInstances[i]);
+
+            uniqueVerifiers++;
+        }
```
Brechtpd, line 62: Can also be extracted to some internal function so it can be reused by `addToRegistryByOwner` and `addToRegistryBySgxInstance`.

Brechtpd (commented):
Looks good! Just some refactoring suggestions.

dantaik:
I need to learn more about SGX, but it seems no public keys are associated with those from Intel.

Brechtpd:
> I need to learn more about SGX, but it seems no public keys are associated with those from Intel.

The public/private keys are generated inside the SGX program and so we have full flexibility of what what we can use to verify proofs, after verifying the attestation that the program is running safely.

adaki2004:
Just linking Patryk's issue [here](https://github.com/gramineproject/gramine/discussions/1579), so that we can see the design proposals from Intel SGX experts.

dantaik (changes requested):

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,165 @@
...
+
+    /// @dev For gas savings, we shall assign each SGX instance with an id
+    /// so that when we need to set a new pub key, just write storage once.
+    uint256 public uniqueVerifiers;
+
+    /// @dev One SGX instance is uniquely identified (on-chain) by it's ECDSA
+    /// public key. Once that public key is used (by proof verification) it has
+    /// to be overwritten by a new one (representing the same instance). This is
+    /// due to side-channel protection.
+    mapping(uint256 instanceId => address sgxInstance) public sgxRegistry;
```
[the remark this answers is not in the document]
adaki2004: The reason is gas savings. Because then we can just associate an instance with an Id, and then we do not need to write storage 2 times (we need to wipe away the addresses pretty 'frequently') , please see this:

https://github.com/taikoxyz/taiko-mono/pull/14851#discussion_r1340348014

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,165 @@
...
     )
         external
+    {
+        // Do not run proof verification to contest an existing proof
+        if (isContesting) return;
+
+        // Size is: 224 bytes
+        // Struct encoding data+
+        // 2x32 bytes (id + address on a word) + 65 bytes (r,s,v)
+        if (evidence.proof.length != 224) {
```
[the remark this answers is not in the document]
adaki2004: 2 problems i see here: 
- if not decodeable this will revert silently (hard to debug)
- on [branch](https://github.com/taikoxyz/taiko-mono/pull/14851#discussion_r1340348014) where tier combined SGX+ZK we need to have size data, because we concatenate 2 byestreams to 1 (in evidence.proof) so easier to have such in my opinion.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,165 @@
...
 contract SGXVerifier is EssentialContract, IVerifier {
-    uint256[50] private __gap;
+    event InstanceAdded(
+        uint256 indexed instanceId, address indexed instanceInitPubKey
+    );
 
-    error UNIMPLEMENTED();
+    struct ProofData {
+        uint256 id;
+        address newPubKey;
```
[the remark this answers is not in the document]
adaki2004: [Here](https://mbed-tls.readthedocs.io/en/latest/security-advisories/mbedtls-security-advisory-2020-04/) is a description.
To avoid leakage, we need to change it relatively frequently.
kép

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,165 @@
...
     function verifyProof(
         // blockId is unused now, but can be used later when supporting
         // different types of proofs.
         uint64,
-        address,
-        bool,
-        TaikoData.BlockEvidence calldata
+        address prover,
+        bool isContesting,
+        TaikoData.BlockEvidence calldata evidence
```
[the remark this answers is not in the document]
adaki2004: Hmm, i do not think so, because everyone can call it and result in zero effect. (Taiko protocol has tied state change mechanisms to it, so i feel it harmless.)

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -7,38 +7,165 @@
...
+        addTrustedInstances(trustedInstances);
+    }
+
+    /// @notice Adds trusted SGX instances to the registry by another SGX
+    /// instance.
+    /// @param instanceId The id of the SGX instance who is adding new members.
+    /// @param newPubKey The new address of the instance.
+    /// @param trustedInstances The address array of trusted SGX instances.
+    /// @param signature The signature proving authenticity.
+    function addToRegistryBySgxInstance(
```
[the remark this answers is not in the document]
adaki2004: 1. First we add trusted instances (attestation off-chain checked) with: `addToRegistryByOwner()`
2. Then when a machine is trusted, it can add other trusted machines (with signature) so no need of permission, because only trusted machine signature guards it.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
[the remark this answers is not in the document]
adaki2004: > I'm not familiar with SGX, SGX can automatically generate a new pubkey after each run? 
it should, yes!

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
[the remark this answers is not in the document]
dantaik: what about this SSTORE `sgxRegistry[id]`?

packages/protocol/test/L1/SgxVerifier.t.sol
```diff
@@ -0,0 +1,81 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import { Test } from "forge-std/Test.sol";
+import { console2 } from "forge-std/console2.sol";
+import { TaikoL1 } from "../../contracts/L1/TaikoL1.sol";
+import { SGXVerifier } from "../../contracts/L1/verifiers/SGXVerifier.sol";
+import { TaikoL1TestBase } from "./TaikoL1TestBase.sol";
+
+contract SgxVerifier is TaikoL1TestBase {
```
dantaik, line 10:
```suggestion
contract TestSgxVerifier is TaikoL1TestBase {
```

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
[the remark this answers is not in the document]
adaki2004: No (or not mecessarily, no, do not think so.) Anyone could submit the trusted machine’s signtaure it is “harmless”, the base principle is that a trusted instance is only doing “good” - hence trusted.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -6,39 +6,177 @@
...
+    uint256 public uniqueVerifiers;
 
-    error UNIMPLEMENTED();
+    /// @dev One SGX instance is uniquely identified (on-chain) by it's ECDSA
+    /// public key (or rather ethereum address). Once that address is used (by
+    /// proof verification) it has to be overwritten by a new one (representing
+    /// the same instance). This is due to side-channel protection. Also this
+    /// public key shall expire after some time. (For now it is a long enough 6
+    /// months setting.)
+    mapping(uint256 instanceId => InstanceData sgxInstance) public sgxRegistry;
```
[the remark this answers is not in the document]
adaki2004: > Again, a mapping with uint as key and a totoal number uniqueVerifiers is the same as an array, so why not use
> `Instance[] public instances`

Actually, i do think your SgxVerifier2 is eliminating the need of this. (by keeping track of non-expired addresses, so i’d adapt that approach).

adaki2004:
@Brechtpd merged all SGX related code together in this branch. (Verifier + Tier config).
May i ask you to have another look on the SgxVerifier contract as proposals from Daniel and the “original” version got merged together ?

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -6,39 +6,154 @@
...
-        revert UNIMPLEMENTED();
+        if (oldInstance != address(0)) {
+            if (!isInstanceValid(oldInstance)) revert SGX_INVALID_AUTH();
+            instances[oldInstance] = 1;
+        }
+
+        if (newInstance == address(0)) revert SGX_INVALID_INSTANCE();
+        if (instances[newInstance] != 0) revert SGX_INSTANCE_REGISTERED();
+
+        instances[newInstance] = block.timestamp;
```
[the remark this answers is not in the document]
adaki2004: @dantaik This is what i was trying to say with this:
https://github.com/taikoxyz/taiko-mono/pull/14851#discussion_r1347310094

Please decide you 2 and then let me know.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -6,39 +6,154 @@
...
+import { LibBytesUtils } from "../../thirdparty/LibBytesUtils.sol";
 
 import { TaikoData } from "../TaikoData.sol";
 
 import { IVerifier } from "./IVerifier.sol";
 
-/// @title GuardianVerifier
-// TODO(dani): implement this verifier.
+/// @title SGXVerifier
 contract SGXVerifier is EssentialContract, IVerifier {
```
dantaik, line 20: Can we use `SgxVerifier`?

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -6,39 +6,154 @@
...
+    }
+
+    /// @notice Adds trusted SGX instances to the registry by another SGX
+    /// instance.
+    /// @param newAddress The new address of the instance.
+    /// @param trustedInstances The address array of trusted SGX instances.
+    /// @param signature The signature proving authenticity.
+    function registerBySgxInstance(
+        address newAddress,
+        address[] memory trustedInstances,
```
dantaik, line 68: what's the difference between 'newAddress' and 'trustedInstances', why `trustedInstances` cannot include `newAddress`?

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
```diff
@@ -6,39 +6,154 @@
...
+
+        address newInstance =
+            address(bytes20(LibBytesUtils.slice(evidence.proof, 0, 20)));
+
+        bytes memory signature = LibBytesUtils.slice(evidence.proof, 20);
+
+        address oldInstance =
+            getSignedHash(evidence, prover, newInstance).recover(signature);
+
+        _replaceInstance(oldInstance, newInstance);
```
dantaik, line 112: I believe https://github.com/taikoxyz/taiko-mono/pull/14911/commits/1dd22238278415ec94e3e1d0aa8876806502177a is clearer

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
adaki2004: It can, so that the first (or last) can be the instance's new address, sure, can do it.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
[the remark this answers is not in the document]
dantaik: How can you save gas if you want to mark an address to be invalid on chain so it cannot be registered again?

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
adaki2004: Can do that, but from gas perspective i still favors the 'id' based solution as Brecht pointed out too. https://github.com/taikoxyz/taiko-mono/pull/14851#discussion_r1350934563

When there is a new instance (and it will be relatively frequent) there will be a 20k gas cost to write them, otherwise if the instance is associated with and ID
then the ID's place in the mapping would be modified:
```
struct InstanceData {
        address activeAddress;
        uint256 setAt;
    }

mapping(uint256 instanceId => address sgxInstance) public sgxRegistry;
```

Very hard to work into the same direction if i constantly get 2 different ones. Please make your call on the design.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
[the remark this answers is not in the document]
adaki2004: Actually I do not see a need for checking if an address was already registered or not.
Example:

Bob is an instance, and is in charge of changing his own address. 
A: Either he can keep track of his addresses and not use what was used before
B: Re-use old addresses (if he has lot of addresses, it will be almost impossible for leakage).
C: Use an incremental variable/word if he generates address from mnemonic/seed phrases so that he does not even have to keep track of the old ones.

adaki2004:
Note: I'll leave this PR as is, until you (@Brechtpd , @dantaik ) make a final call about the design. (ID based vs. non-ID based).

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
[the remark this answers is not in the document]
Brechtpd: The new private key will be generated from randomness inside the SGX program, so the new public key will also be random. It's the SGX program that does this, so we don't have to verify that the old public key cannot be used anymore, because the correct execution of the program already achieves that.

packages/protocol/contracts/L1/verifiers/SGXVerifier.sol
Brechtpd: Logically having them separate makes more sense to me. `trustedInstances` is the list of addresses that an SGX instance has verified the remote attestations of, and `newAddress` is his own new public key for the next SGX proof that SGX instance will generate. Combining them is a bit confusing I feel.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        // Invalidate current key, because it cannot be used again (side-channel
+        // attacks).
+        sgxRegistry[instanceId] =
+            InstanceData(newInstance, uint64(block.timestamp));
+
+        emit InstanceChanged(
+            instanceId, oldInstance, newInstance, uint64(block.timestamp)
+        );
```
Brechtpd, line 147: Maybe move these things to an internal function so `registerInstancesBySgx` uses the exact same code. Because currently `InstanceChanged` is thrown here, but not in `registerInstancesBySgx` when exactly the same thing is done. So both using the same internal function makes sure they are consistent.

Brechtpd (approved):
Looks good, only one small suggestion

packages/protocol/test/TestBase.sol
```diff
@@ -47,6 +48,11 @@ abstract contract TestBase is Test {
     address internal Xavier = getRandomAddress();
     address internal Yasmine = getRandomAddress();
     address internal Zachary = getRandomAddress();
+    address internal SGX_X_0 = vm.addr(0x4);
```
dantaik, line 51: @adaki2004 please try to avoid adding test-specific code into TaikoL1TestBase, this base class is supposed to be clean and decoupled from actual tests

packages/protocol/test/L1/TaikoL1TestBase.sol
```diff
@@ -68,15 +70,38 @@ abstract contract TaikoL1TestBase is TestBase {
...
+        sgxZkVerifier.init(address(addressManager));
+
         gv = new GuardianVerifier();
         gv.init(address(addressManager));
 
-        cp = new OptimisticRollupConfigProvider();
+        gp = new GuardianProver();
+        gp.init(address(addressManager));
+        address[5] memory initMultiSig;
+        initMultiSig[0] = David;
```
dantaik, line 89: Test test specific code should go to their own class, the base class is suppose to be clean and lean.

dantaik (approved):

packages/protocol/test/L1/TaikoL1TestBase.sol
dantaik: I added two TODOs in the test files, you can take care of them later and go ahead to merge this PR.

dantaik:
@adaki2004 in the SGX verifier, should we somehow associate the proof with the image id, by image, I mean the specific software we run inside the SGX, otherwise, the SGX may simply run a very different binary, then all will be wrong.

adaki2004:
> @adaki2004 in the SGX verifier, should we somehow associate the proof with the image id, by image, I mean the specific software we run inside the SGX, otherwise, the SGX may simply run a very different binary, then all will be wrong.

Not the proof but maybe the actual contract itself can be a good idea! (Like a contract version OR tier config setting for example of we want to do something like: till blockheight X we want to support version 1.0. Affer that 1.01 so basically we support it right away with these multi-tiered configs).
The reason is to the contract level association that the remote attestation verification takes care of this. When a new program/image is created, it requires a distinct remote attestation for validation. The most straightforward is to maintain separate SGX contracts for each version.

dantaik:
> example if we want to do something like: till blockheight X we want to support version 1.0. Affer that 1.01 so basically we support it right away with these multi-tiered configs OR even tge verifyProof() can determine the blockcheight X (X+1) issue bc it has the blockId, various solutions, so proof level image association is not necessary).
> The reason is to the contract level association that the remote attestation verification takes care of this. When a new program/image is created, it requires a distinct remote attestation for validation. The most straightforward is to maintain separate SGX contracts for each version.

But how can a Taiko block chain viewers knows which image/version we are using for each block?

Brechtpd:
In cases of bug fixes then everything is transparent except for provers who will need to update their software which will contain the necessary information I'd assume. In case of hard forks the block ranges will be part of the node. I don't think there's a difference between the different prover types (zk, SGX) of how versions will be handled. zk has the verification keys, SGX currently has the "bootstrap nodes", or when we can verify the attestations onchain that will define the version for the SGX verifier.

dantaik:
> In cases of bug fixes then everything is transparent except for provers who will need to update their software which will contain the necessary information I'd assume. In case of hard forks the block ranges will be part of the node. I don't think there's a difference between the different prover types (zk, SGX) of how versions will be handled. zk has the verification keys, SGX currently has the "bootstrap nodes", or when we can verify the attestations onchain that will define the version for the SGX verifier.

If I understand correctly, the ZK keys which can be calculated from the circuit codebase, but the SGX "bootstrap nodes" setup cannot be verified and has to be trusted. If there is also an image ID that can be verified by using the guest program such as zeth, then the solution is more trustworthy. But I guess it is hard to achieve?

Brechtpd:
They are pretty similar in that to verify the verification keys in the zk verifier contract, you would indeed have to verify the code and regenerate the same verification keys and check if they match the ones in the smart contract. Registering those verification keys is itself a trusted process, because the smart contract cannot verify if those keys are correct. But it is verifiable.

With SGX you would have to check the source code, and then generate the `MRENCLAVE` value which is the hash of the program being run, and then verify the that the remote attestation published for the bootstrap node matches the expected `MRENCLAVE`. The trusted process here is that those bootstrap nodes are being registered to the smart contract without the smart contract verifying that they are indeed running the correct program. But as long as the remote attestations of those bootstrap nodes are made available offchain, anybody can verify them so it's exactly the same as with the zk verifier.

The only benefit of being able to verify the remote attestations onchain is that there is no dependency on those bootstrap nodes, so it's potentially more decentralized because those bootstrap nodes may refuse verifying other SGX instances or they may go offline. But even with onchain remote attestation verification, setting the expected `MRENCLAVE` for the smart contract verifying the remote attestation verification would be a trusted step, so nothing really changes to the security or the trustlessness of the process. Both approaches are as verifiable and as trustless as the other, and are the same here as for the zk verifier.

dantaik:
> They are pretty similar in that to verify the verification keys in the zk verifier contract, you would indeed have to verify the code and regenerate the same verification keys and check if they match the ones in the smart contract. Registering those verification keys is itself a trusted process, because the smart contract cannot verify if those keys are correct. But it is verifiable.
> 
> With SGX you would have to check the source code, and then generate the `MRENCLAVE` value which is the hash of the program being run, and then verify the that the remote attestation published for the bootstrap node matches the expected `MRENCLAVE`. The trusted process here is that those bootstrap nodes are being registered to the smart contract without the smart contract verifying that they are indeed running the correct program. But as long as the remote attestations of those bootstrap nodes are made available offchain, anybody can verify them so it's exactly the same as with the zk verifier.
> 
> The only benefit of being able to verify the remote attestations onchain is that there is no dependency on those bootstrap nodes, so it's potentially more decentralized because those bootstrap nodes may refuse verifying other SGX instances or they may go offline. But even with onchain remote attestation verification, setting the expected `MRENCLAVE` for the smart contract verifying the remote attestation verification would be a trusted step, so nothing really changes to the security or the trustlessness of the process. Both approaches are as verifiable and as trustless as the other, and are the same here as for the zk verifier.

I think a trusted set up of MRENCLAVE is acceptable as long as it is verifiable by third parties. Is it possible to achieve this without onchain remote attestation verification?

Brechtpd:
Yes, because onchain remote attestation verification does not make any difference.

1. `MRENCLAVE`: hash of code/data that is run in SGX
2. Remote attestation: Verifies that the SGX instance is running the `MRENCLAVE` program correctly in SGX
3. We bind the SGX instance to an initial public key

No need for any trusted setup or onchain remote attestation verification, it is as trustless as registering the verification keys for the zk circuits (ignoring the trust in intel of course).

PR #14860, opened 2023-09-29
test(protocol): decouple 1559 logic from TaikoL2, implement AMM style 1559

dantaik:
In this PR, I've implemented a 1559 basefee per gas adjustment algorithm, leveraging Uniswap's AMM formulas. I ran simulations over 400 blocks: 
- 95% of blocks have block delays centered around 3 seconds
- 5% have delays centered around 100 seconds
- Gas used per block follows a normal distribution around 4.3 million (sourced from A5 testnet data).

Comparison Chart

### Observations:
- The AMM-based approach struggles with blocks that have longer delays (5%). 
- Longer delays lead to more gas being added to the AMM pool, causing a significant drop in the basefee per gas.
- The issue is exacerbated when there are no subsequent blocks to consume the extra gas, causing the basefee per gas to not recover -- this may be the reason that the current A5 testnet's basefee per gas is always 1 (the min value).

Given that 100-second block delays are relatively common in L2, I'm currently leaning towards the standard 1559 approach and against incorporating time into the calculation. 

## How to Run Tests:
```console
cd packages/protocol
python3 test/L2/blocktime_gasused_gen.py
forge test --mt test_1559 -vvv
```
CSV data will output to the console for chart plotting in Excel.

@Brechtpd I'd love to get your feedback.


Brechtpd:
I don't see how the vanilla EIP-1559 can ever work for us though because it's super easy to manipulate. Just submit empty blocks to push the base fee down as much as you want before you submit your actual block, rendering the EIP-1559 system useless. In the vanilla EIP-1559 time is incorporated in the formula, just not implicitly because Ethereum has fixed block times, and there is always a block. No block in Ethereum would be an empty block, which would also push the base fee down (as you would want to happen). But the amount of empty blocks is limited in time and proposers don't want to miss their chance to make money with an actual block. For our current L2, this is not the case because blocks can be proposed at any time and so you can manipulate the basefee freely.

The basefee staying down if there are no blocks is also the expected (and required) behavior. If blocks don't use more than the target gas limit/s, than the basefee should keep going down. EIP-1559 is a way to extract congestion fees. If there is no congestion, the base fee should be 0 (or 1), otherwise it would be a tax on users.

dantaik:
> I don't see how the vanilla EIP-1559 can ever work for us though because it's super easy to manipulate. Just submit empty blocks to push the base fee down as much as you want before you submit your actual block, rendering the EIP-1559 system useless. In the vanilla EIP-1559 time is incorporated in the formula, just not implicitly because Ethereum has fixed block times, and there is always a block. No block in Ethereum would be an empty block, which would also push the base fee down (as you would want to happen). But the amount of empty blocks is limited in time and proposers don't want to miss their chance to make money with an actual block. For our current L2, this is not the case because blocks can be proposed at any time and so you can manipulate the basefee freely.
> 
> The basefee staying down if there are no blocks is also the expected (and required) behavior. If blocks don't use more than the target gas limit/s, than the basefee should keep going down. EIP-1559 is a way to extract congestion fees. If there is no congestion, the base fee should be 0 (or 1), otherwise it would be a tax on users.

Somehow I'm not convinced that the standard 1559 cannot apply to variable block times. Before PoS, the proof of work chain doesn't guarantee the block time to be very constant, and I think 1559 also works even for PoW. Anyway, I decoupled 1559 code from TaikoL2 so we can easily switch to new implementations easily.

Brechtpd:
> Somehow I'm not convinced that the standard 1559 cannot apply to variable block times. Before PoS, the proof of work chain doesn't guarantee the block time to be very constant, and I think 1559 also works even for PoW. Anyway, I decoupled 1559 code from TaikoL2 so we can easily switch to new implementations easily.

The average block time was still fixed at 13s, and the thing EIP-1559 does is target an average gas/s so that was fine. And of course also the very important fact that it's only possible to propose a single block. If people could propose multiple blocks in a single propose slot it would also not work, because otherwise why would a proposer make a block larger than the target to make the basefee increase?

PR #14866, opened 2023-09-30
feat(protocol): sync L2 base fee per gas from L1

dantaik:
The prevailing belief has been that L2's block space is independent of L1, and therefore, the EIP-1559 base fee per gas (`basefee`) on L2 should be calculated without reference to L1's `basefee`. 

Upon further consideration, it seems impractical to maintain low L2 base fees when L1 is highly congested. Offering cheap block space on L2 during such times could lead to inefficient resource allocation. This PR aims to address this by syncing the L1 and L2 base fees.

This PR aims to align the L2 base fee with the L1 base fee, challenging the previous assumption that L2's block space should be independent of L1's. Specifically, the L2 base fee is set to be 1/100th of the L1 base fee.


Brechtpd:
I don't really understand how this could work:
- The main goal of the L2 basefee is so that we can target a safe gas/s that the rollup can sustain. If the L2 basefee is not linked to the activity on L2, it fails to do this. The L1 cost of a tx and the L2 work required by that tx are also not related.
- If the L1 basefee is high, I would think in practice that the L2 basefee will probably go down because the price an L2 transaction wants to pay to get included on the L2 gets eaten up more by the L1 fees required by the L1 proposer. So if the L2 transaction is willing to pay 10 gwei, but the proposer already needs 8 gwei to submit the transaction with profit on-chain, only 2 gwei could be eaten up by the basefee. For transactions that want to pay only 8 gwei, the proposer will have to wait until the basefee goes down to 0, or the L1 cost goes down. This makes sense because that means the the transactions are only willing to pay the raw L1 cost to get included, which means that the transaction does not want to pay a priority fee, and so there is no congestion (or at least nobody willing to pay for priority inclusion) if these transactions get included. If the L1 cost then goes down to 6 gwei, then there could be congestion again to get all these 8 gwei transactions included. It is true that the target gas/s can be undershot because compared to an L1 the baselevel profitability of a transaction is not simply 0 but changes with the L1. Sot here could be transactions, but depending on the L1 cost they could or could not be used (when not taking subsidies into account). This is similar to how there would simply be no transaction on L1.
- There is no target gas target/s without time, and there is no congestion without time. How the block submission works is also critical for what can work here. The vanilla EIP-1559 formula would work if only a single L2 block with variable block size can be submitted per L1 block (or some other mechanism that is equivalent to this).

PR #14867, opened 2023-09-30
refactor(protocol): make sure L1.init called in genesis, not the first block

dantaik:
David, in the previous code, the block.number can be either 0 or 1. With our current way of genesis the L2, the value should always be 0, is it?

@davidtaikocha @adaki2004 could one of you run the genesis test, my docker is broken locally.


davidtaikocha:
I think its intended for unit tests (`block.number` can be either `0` or `1`)? since for genesis generation, we wont need / run a `init` function, we just put the values in the slots: https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/utils/generate_genesis/taikoL2.ts#L343

PR #14885, opened 2023-10-02
refactor(protocol): reimplement curve(amm)-based 1559 for L2a

dantaik:
The previous implementation is buggy. I think the new one is correct and simpler. The core code is ready for review, but a few TODOs  remain:

- [x] some helper functions are missing implementation
- [x] add more comments and improve related documentation
- [x] add/fix tests


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -24,36 +24,31 @@ import { TaikoL2Signer } from "./TaikoL2Signer.sol";
 contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
     using LibMath for uint256;
 
+    uint64 public constant GAS_TARGET_PER_L1_BLOCK = 150 * 1e6;
```
adaki2004, line 27: hmm, should not it be 15 mill ?

packages/protocol/contracts/L2/1559/Lib1559Math.sol
```diff
@@ -0,0 +1,71 @@
...
+    )
+        private
+        pure
+        returns (uint256 baseFeePerGas)
+    {
+        int256 diff = _ethQty(
+            gasExcess + gasInBlock, gasTarget, adjustmentQuotient
+        ) - _ethQty(gasExcess, gasTarget, adjustmentQuotient);
+        baseFeePerGas =
+            uint256(diff) / gasInBlock / LibFixedPointMath.SCALING_FACTOR_1E18;
```
Brechtpd, line 52: I think doing the derivative at the current `gasExcess` is a bit more accurate like how Vitalik does it, because we're working on the state after the previous block (and so we are not calculating the slope for the current block, in which case calculating the diff like is done here would be more accureate). But I guess the biggest thing is that it supports 0 gas blocks and the accuracy is not dependent on how big `gasInBlock` is (UNTESTED so I may do the scaling wrong):
```
baseFeePerGas = _ethQty(gasExcess, gasTarget, adjustmentQuotient) / (gasTarget * adjustmentQuotient * LibFixedPointMath.SCALING_FACTOR_1E18)
```

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -158,24 +141,41 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
+        (baseFeePerGas, gasExcess) = Lib1559Math.calcBaseFee({
+            numL1Blocks: latestSyncedL1Height == 0
+                ? 0
+                : l1Height - latestSyncedL1Height,
+            gasExcess: gasExcess,
+            gasInBlock: gasInBlock,
+            gasTarget: GAS_TARGET_PER_L1_BLOCK,
+            adjustmentQuotient: ADJUSTMENT_QUOTIENT
+        });
+        if (gasExcess == 0) gasExcess = 1;
```
Brechtpd, line 162: This one is just to prevent the storage slot to be reset to 0?

Brechtpd (commented):
Nice! Using the `l1height` instead of the L2 timestamp also makes sense to optimize having to store the timestamp otherwise (with a small dependency on the L1 block times now, but yeah that's not changing any time soon :)

packages/protocol/contracts/L2/1559/Lib1559Math.sol
dantaik: Will test it out

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: I was assuming we scale Ethereum by 10 times :) We can adjust this number later. I'll also add a comment.

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: Yes, here I use 0 to indicate the dynamic basefee calculate is disabled.

packages/protocol/contracts/L2/1559/Lib1559Math.sol
dantaik: It turns out the derivative based base fee calc is more accurate:

```solidity
def basefee1(gas_used):
    diff = eth_qty(gas_excess_issued + gas_used) - eth_qty(gas_excess_issued)
    return int(diff / gas_used /SCALE)

def basefee2(): //  turns out to be better
    return eth_qty(gas_excess_issued) / SCALE / TAIKO_TARGET / ADJUSTMENT_QUOTIENT

```

dantaik:
@Brechtpd @adaki2004 this PR is ready for another round of review. Tests are not ready.

packages/protocol/test/L2/Lib1559Math.t.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import { console2 } from "forge-std/console2.sol";
+import { TestBase } from "../TestBase.sol";
+import { TaikoL2 } from "../../contracts/L2/TaikoL2.sol";
+
+contract TestLib1559Math is TestBase {
+// function test_1559ParamCalculation() external {
```
dantaik, line 9: I'll be working on this test file later

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -24,158 +24,102 @@ import { TaikoL2Signer } from "./TaikoL2Signer.sol";
...
-    uint64 public parentTimestamp; // slot 4
+    uint128 public gasExcess; // slot 4
     uint64 public latestSyncedL1Height;
-    uint64 public gasExcess;
 
     uint256[146] private __gap;
 
     // Captures all block variables mentioned in
     // https://docs.soliditylang.org/en/v0.8.20/units-and-global-variables.html
-    event Anchored(
```
dantaik, line 54: @davidtaikocha David, I removed may data from this event as those data are available in the block header. Let me know if it's OK.

packages/protocol/contracts/L2/1559/Lib1559Math.sol
```diff
@@ -0,0 +1,71 @@
...
+        if (gasTarget == 0 || adjustmentQuotient <= 1 || gasInBlock == 0) {
+            revert EIP1559_INVALID_PARAMS();
+        }
+
+        uint128 issuance = numL1Blocks * gasTarget;
+        _gasExcess = issuance >= gasExcess ? 0 : gasExcess - issuance;
+
+        _baseFeePerGas =
+            _calcBaseFee(_gasExcess, gasInBlock, gasTarget, adjustmentQuotient);
+        _gasExcess += gasInBlock;
```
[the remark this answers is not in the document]
dantaik: should have been fixed.

Brechtpd (approved):

adaki2004 (approved):

PR #14887, opened 2023-10-03
feat(protocol): notify the verifier when a proof lost a contestation

dantaik:
Inspired by Vitalik's [multi-prover video](https://www.google.com/search?q=vitalik+multi+prover+youtube&oq=vitalik+multi+prover+youtube&gs_lcrp=EgZjaHJvbWUyBggAEEUYOdIBCDYzMzRqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8#fpstate=ive&vld=cid:598c69d0,vid:6hfVzCWT6YI,st:0), verifiers can now receive notifications from TaikoL1 when their proofs are contested and proven to be wrong. Verifiers can implement custom logic to handle these notifications, such as disabling themselves.

Screenshot 2023-10-03 at 13 26 54

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProving.sol
+++ b/packages/protocol/contracts/L1/libs/LibProving.sol
@@ -294,7 +294,7 @@ library LibProving {
                 tier: evidence.tier
             });
         } else {
-            // The new tier is higher than the previous tier, we  are in the
+            // The new tier is higher than the previous tier, we are in the
             // proving mode. This works even if this transition's contester is
             // address zero.
 
@@ -377,6 +377,24 @@ library LibProving {
                         false
                     );
                 } else {
+                    // Notify the verifier that a proof it verified has been
+                    // contested and invalidated. This allows the verifier to
+                    // implement logic for self-disabling.
+                    if (tran.tier != 0) {
+                        ITierProvider.Tier memory prevTier = ITierProvider(
+                            resolver.resolve("tier_provider", false)
+                        ).getTier(tran.tier);
+
+                        address verifier =
+                            resolver.resolve(prevTier.verifierName, true);
+
+                        if (verifier != address(0)) {
+                            IVerifier(verifier).handleLostContestation(
+                                blockId, tran.prover, tran.blockHash
+                            );
+                        }
+                    }
+
                     // In the event that the contester is the winner, half of
                     // the validity bond is designated as the reward, to be
                     // divided equally between the new prover and the contester.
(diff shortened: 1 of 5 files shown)


dantaik:
> Well, logically looks good to me but on the other hand this can be done off-chain. Each and every client who seeks for feedback can listen / index `TransitionProved` events and if one comes after (they submitted a valid proof), with same `parentHash` but different `blockHash` and `signalRoot` gives the same result.

The goal is to make it possible to disable verifiers automatically. You can also argument that this can also be done manually after a decision is made off chain.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -377,6 +377,24 @@ library LibProving {
                         false
                     );
                 } else {
+                    // Notify the verifier that a proof it verified has been
```
Brechtpd, line 380: I don't understand all the different phases yet of this design, but shouldn't this be done in the `evidence.tier == tran.tier` case? Because if just another proving system has a different result, how can the verifier make a self assessment using that information?

Brechtpd (commented):
We actually implemented this before, I think pretty much directly after Vitalik proposed it: https://github.com/taikoxyz/taiko-mono/blob/69e936264d1f8bbd32c7884ddbf5925dfe2c8aaf/packages/protocol/contracts/L1/v1/V1Proving.sol#L280-L285

But it got removed in one of the simplification refactorings I guess.

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: The idea is if a proof A on tier 3 is contested by another proof B on the same tier, and a proof C on tier 5 has proven that B is correct and A is not, then we notify the verifier for tier 3,...

dantaik: The example I gave above may not be clear. Let me try this:

1. Alice submits a block header X, along with a tier 2 proof.
2. Bob challenges Alice's claim by presenting a different header Y. No proof is required for Y, and even if provided, the protocol ignores it. This means the challenger can be any user, not necessarily a prover.
3. Charley validates that Y is the correct header hash using a tier 3 proof. Only at this point do we suspect that the tier 2 proof might be flawed. Alice's tier 2 proof was initially verified but is now overridden by Charley's tier 3 proof. The tier 2 verifier is notified and may opt for self-disabling.

Brechtpd: I see, so there is some subjectiveness that comes with this approach, compared to the two conflicting proofs for the same proof type which always means it's buggy.

dantaik: Yes

dantaik:
Closing this PR as the feature introduce is not used immediately. We can come back for it later.

Issue #14881, opened 2023-10-02
Make proposer rewards distribution more fair


adaki2004:
### Describe the feature request

Currently the first proposer within an L1 block gets the TKO compensation which does lead (at least on testnet definitely) to 'gas war' situations which fuels by speculation that TKO might worth something on mainnet. 
On the other hand the design is what it makes it doable, so maybe best to change it for example like:

1. Accumulate the "rewards" by every seconds (kind of same as now, but not reset to 0 at all.)
2. Give 1/3 of the accumulated rewards to each every proposers

### Describe alternatives you've considered

Description of the alternatives you've considered here.

### Additional context

Additional context here.

dantaik:
Given the discussion with Brecht yesterday, lets postpone this improvement.

adaki2004:
> Given the discussion with Brecht yesterday, lets postpone this improvement.

I understand it may be out of interest for now, but was already half cooked, so i just leave it as is. We can close that PR then.
https://github.com/taikoxyz/taiko-mono/pull/14888

dantaik:
Fixed

PR #14888, opened 2023-10-03
feat(protocol):  implemented partial and more fair block proposal rewards

adaki2004:
I know it may be not included (if we cut out the whole feature - see Brecht's comment in discord) but since the changes were just couple of lines, i tried it.

A more fair distribution for proposers. Will create some simulation.

Basic idea is:
There is a "pot". This does not get resetted, but the emission is set and the pot grows with it. 
With each and every `proposeBlock()` the 1/3 of the pot is paid out to the proposer. -> So everyone is getting 'something' and the halving % can be changed.
1/3 means the diff between 1st and 2nd proposer reward is closer (than it would be 1/2).


adaki2004:
As Daniel said [here](https://github.com/taikoxyz/taiko-mono/issues/14881#issuecomment-1744480435) we might not use it or out of interest, just leaving it as is - and we can close this PR without deleting the branch.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -101,26 +101,29 @@ library LibProposing {
...
-                    LibTaikoToken.creditToken(
-                        state, resolver, msg.sender, reward, true
-                    );
-                }
+        // First proposer per L1 block updates the L1 block height and the accumulated reward.
+        if(state.slotC.latestRewardAccL1Height < block.number) {
+            // The very first proposal within the protocol starts the 'clock' and from that point on we calculate the accumulation.
+            if(state.slotC.latestRewardAccL1Height != 0) {
+                // Calculate the addition to the accumulated rewards by getting how many blocks elapsed since the last 'sync' then multiply it by 12 (L1 block time) and the reward issued per second.
+                state.slotC.accumulatedReward += uint128((block.number - state.slotC.latestRewardAccL1Height) * 12 * config.proposerRewardPerSecond);
```
dantaik, line 109: Lets use  `proposerReviewPerL1Block` to replace 12 * config.proposerRewardPerSecond`

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -101,26 +101,29 @@ library LibProposing {
...
-
-                    // Reward must be minted
-                    LibTaikoToken.creditToken(
-                        state, resolver, msg.sender, reward, true
-                    );
-                }
+        // First proposer per L1 block updates the L1 block height and the accumulated reward.
+        if(state.slotC.latestRewardAccL1Height < block.number) {
+            // The very first proposal within the protocol starts the 'clock' and from that point on we calculate the accumulation.
+            if(state.slotC.latestRewardAccL1Height != 0) {
```
dantaik, line 107: How about use 'lastProposedHeight' to represent the L1 block number in which the last block proposal occurred. 

Later we can do:
```
if (proposerRewardPerL1Block > 0) {
    if (block.number != state.slotC.lastProposedHeight) {
       state.slotC.accProposerReward += config.proposerRewardPerL1Block * (block.numeber - state.slotC.lastProposedHeight);
      if (state.slotC.accProposerReward > proposerRewardMax) state.slotC.accProposerReward = proposerRewardMax;
       state.slotC.lastProposedHeight = block.number;
    }

    reward = state.slotC.accProposerReward/3;
    state.slotC.accProposerReward -= reward;
}
```

adaki2004:
Simulation results:
Settings:
Emission rate: 0.1 TKO / sec
Average proposeBlock() per L1 ethereum block: 4.95

Over 200 - only be able to test 200 due to gas constraints - Ethereum blocks (which is roughly 3/4 hour): 

239.3 TKO paid out.

That is 3.1 million TKO per year (if the above numbers are the average).

Every proposer gets something not the first one.
Paid out per block diagram is as below, so depending on how many proposers we had per L1 block, it is between 0 - 3 TKO per every 12 seconds.

kép

With 1.65 L2 blocks (per L1 block) the rewards are relatively the same (rounding diff): 237.89 TKO in 200 blocks.

The distribution is:
kép

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: Assuming each Taiko block time is 3 seconds, then in each Etheruem block, there will be 4 L2 blocks. Their reward will be:

1/3, 2/9, 4/27, 8/81, or in percentage: 41.54%, 27.69%, 18.46%, and 12.31%,
with 3 blocks, 47.37%, 31.58%, and 21.05%;
with 2 blocks: 60% and 40% respectively.

dantaik (changes requested):

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: The ratios looks good

packages/protocol/contracts/L1/libs/LibProposing.sol
adaki2004: `latestRewardAccL1Height` is the latest L1 height the reward was pooled into the `accumulatedReward`. (It always happens in the first `proposeBlock()` per given L1 block. You want me to change the name ? I can do that, it indeed might be confusing.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -99,27 +99,46 @@ library LibProposing {
...
+                        (extraRewardMinted + state.slotC.accumulatedReward).min(
+                            config.proposerRewardMax
+                        )
                     );
+                }
+
+                if (state.slotC.accumulatedReward != 0) {
+                    // The current proposer receives a fixed percentage of the
+                    // reward pool.
+                    reward = state.slotC.accumulatedReward
```
dantaik, line 131: `state.slotC.accumulatedReward * config.proposerRewardPoolPctg / 100 ` may overflow but `state.slotC.accumulatedReward /100 * config.proposerRewardPoolPctg ` won't.

adaki2004: In such case it may, but i think we have bigger problems if the
`state.slotC.accumulatedReward = type(uint128).max - 1`, bc it would basically mean halted chain since long years.

adaki2004: If we are fine by losing a bit due to rounding error, i'll modify it back, no issue :)

Brechtpd:
In what ways do we think this will be more fair? Trying to understand the thought process, I might miss something.

adaki2004:
> In what ways do we think this will be more fair? Trying to understand the thought process, I might miss something.

So before, the whole reward went into the first `proposeBlock()` transactor per given L1 and reward then resetted.

Now this is a pool, accumulated every second (every L1 block, so every 12 seconds) and X portion is going to the proposer and then halves.. So basically now everyone gets something.

dantaik:
> In what ways do we think this will be more fair? Trying to understand the thought process, I might miss something.

Previously, if 3 blocks are proposed inside the same L1 block, the first proposer get 100% reward, the second and the third get 0; if we set the new parameter to 50%, then the first proposer get 50%, the second one get 25%, the third one get 12.5% (and another 12.5% remains)

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: I prefer `state.slotC.accumulatedReward /100 * config.proposerRewardPoolPctg` :)

adaki2004: Done.

Brechtpd:
In practice though, we are pretty much limited by one proposer per Ethereum block, because any proposer that is not the first in the L1 block is building on top of state that is unknown to him, so the profitability of the block will be unknown so very risky for a proposer to do that (almost a certainty that that proposer will lose money when doing this, subsidy or not, because the 1st proposer is in a much much better position).

I do agree that giving all the reward for only the first block in the L1 block could be inefficient, because in theory the subsidy can be spread over multiple L2 blocks to make all of them profitable, instead of just giving all of it to the first one and then the subsidy being wasted by just giving it to the ETH validator when it's too much, and the other blocks are simply not profitable so won't be submitted. So having some partial distribution mechanism may be worthwhile, which would be triggered if suddenly multiple L2 blocks could be made that are profitable, which could certainly be the case especially when the L1 fees go down.

dantaik:
We may still config the on-chain reward to 0 and conduct off-chain reward periodically.

dantaik (approved):

PR #14898, opened 2023-10-04
improve Dani's code

dantaik:


packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -99,42 +99,47 @@ library LibProposing {
...
+                            config.proposerRewardMax
+                        )
+                    );
+                }
+
+                if (state.slotC.accumulatedReward != 0) {
+                    // The current proposer receives a fixed percentage of the
+                    // reward pool.
+                    reward = state.slotC.accumulatedReward / 100
+                        * config.proposerRewardPoolPctg;
```
adaki2004, line 132: Do not think it will happen often, but safer to do multiplication first then division because `state.slotC.accumulatedReward < 100` will result in 0 token.

adaki2004 (approved):

PR #14911, opened 2023-10-06
refactor(protocol): implemented an alternative SGX verifier

dantaik:


packages/protocol/contracts/L1/verifiers/SgxVerifier2.sol
```diff
@@ -0,0 +1,139 @@
...
+import { EssentialContract } from "../../common/EssentialContract.sol";
+import { Proxied } from "../../common/Proxied.sol";
+import { LibBytesUtils } from "../../thirdparty/LibBytesUtils.sol";
+
+import { TaikoData } from "../TaikoData.sol";
+
+import { IVerifier } from "./IVerifier.sol";
+
+/// @title SgxVerifier2
+contract SgxVerifier2 is EssentialContract, IVerifier {
```
adaki2004, line 20: This contract lacking the following functionality:
- registerInstancesByAnotherSGXInstance (Please see `addToRegistryBySgxInstance` in `SgxVerifier.sol`)


I guess the one below is somewhat the one, but it lacks the functionality to add, other trusted party's address.     
```
// TODO(dani): who will call this function?
    function replaceInstance(
        address newInstance,
        bytes memory signature
    )
        external
    {
        address oldInstance = keccak256(
            abi.encode("REGISTER_SGX_INSTANCE", newInstance)
        ).recover(signature);

        _replaceInstance(oldInstance, newInstance);
    }
```

packages/protocol/contracts/L1/verifiers/SgxVerifier2.sol
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    function _replaceInstance(
+        address oldInstance,
+        address newInstance
+    )
+        private
+    {
+        if (oldInstance != address(0)) {
+            if (!isInstanceValid(oldInstance)) revert SGX_INVALID_AUTH();
```
adaki2004, line 125: The SGX instance shall have a chance to update/replace his own 'old' instance even in case it is expired.

packages/protocol/contracts/L1/verifiers/SgxVerifier2.sol
adaki2004: Beside that LGTM.

packages/protocol/contracts/L1/verifiers/SgxVerifier2.sol
```diff
@@ -0,0 +1,139 @@
...
+    /// @param _addressManager The address of the address manager contract.
+    function init(address _addressManager) external initializer {
+        EssentialContract._init(_addressManager);
+    }
+
+    function registerInstance(address instance) external onlyOwner {
+        _replaceInstance(address(0), instance);
+    }
+
+    // TODO(dani): who will call this function?
```
adaki2004, line 48: https://github.com/taikoxyz/taiko-mono/pull/14911/files#r1349496658

packages/protocol/contracts/L1/verifiers/SgxVerifier2.sol
dantaik: Dani, why the SGX instance wants to add other third party addresses? Should we just use the owner's registerInstance() function?

adaki2004: Hey Daniel,
So implementing Brecht’s proposal , see as:

“Still an interesting alternative approach without any onchain attestion verification: just use a set of initial set of "bootstrap" SGX nodes when we launch the rollup that have published their attestation somewhere and so everyone can check it offchain. These bootstrap SGX nodes can then verify attestations from other SGX machines that want to generate proofs (and these themselves are also allowed to verify attestations from others). So this way onchain we simply just have to check a simple signature for SGX machines that want to register themselves so their SGX proofs can be accepted.”

Also see the reference here in the first branch:

https://github.com/taikoxyz/taiko-mono/pull/14851#discussion_r1340351054

dantaik: I still don't know the whole SGX solution well, so lets skip this question and leave it to you, Patryk, and Brecht. 

The new implementation is simply a more concise implementation of SGXVerifier.sol

packages/protocol/contracts/L1/verifiers/SgxVerifier2.sol
```diff
@@ -0,0 +1,140 @@
...
+        // Do not run proof verification to contest an existing proof
+        if (isContesting) return;
+
+        // Size is: 87 bytes
+        // 2 bytes + 20 bytes + 65 bytes = 87
+        if (evidence.proof.length <= 20) {
+            revert SGX_INVALID_PROOF_SIZE();
+        }
+
+        address newInstance =
```
dantaik, line 83: @adaki2004 One question I have is whether we should make "newInstance" here also a list: `newInstances`, given that the `replaceInstance()` function takes an array.

dantaik:
@adaki2004 could you merge this PR then work on the tests?

PR #14922, opened 2023-10-09
feat(protocol): lower top tier proof cost

dantaik:
The goal is to make the muti-sig based guardian prover's cost to be as small as possible. We now config the number of blocks to verify after each proveBlock tranaction inside the tier config, instead of using a global config for all tiers.


davidtaikocha (approved):

adaki2004:
> The goal is to make the muti-sig based guardian prover's cost to be as small as possible. We now config the number of blocks to verify after each proveBlock tranaction inside the tier config, instead of using a global config for all tiers.

Logic wise LGTM, but i have one other concern.

This introduces extra variable + some logic which ( at first glance seems not significant but) results in +10% on average and +30% in median `proveBlock()` transactions which seems a bit of a high price to make multi-sig guardian calls cheaper - and hopefully those guardians shall be rare - while the regular users pay more.

base branch gas tests:
kép


this one:
kép

adaki2004 (approved):

Issue #14910, opened 2023-10-06
Investigate multisig options for guardian


adaki2004:
### Describe the feature request

Investigate the way how guardian shall operate behind a multisig wallet/contract.
Use gnosis ?
Use custom contract ?

### Describe alternatives you've considered

Description of the alternatives you've considered here.

### Additional context

Additional context here.

adaki2004:
High level overview:
**Pre-requisite 1:** We have 5 wallets (Alice, Bob, Cecile, Dave, Elliot) possibly the signing (private) key is sitting in a secure, cloud environment. (On different instances, with different providers, GCP / AWS / Tencent, etc.) -> They will be the multisig signer.

**Pre-requisite 2:** We have a 3/5  (just an example) multisig contract (`address: 0x1`), so at least 3 address  needs to sign a transaction with a multisig, which then fires away the `proveBlock()`.

**Pre-requisite 3:** `address: 0x1` is set as `guardian`.

**Pre-requisite 4:** There is a backend monitoring/alerting system observing the chain.

Workflow:
1. There is a need for the `GUARDIAN` to submit it's proof. (Example: either because GuardianTier-1 tier was contested or alerting system even figured out without contestation that is indeed incorrect or there is an unproveable block, etc.). There is monitoring/backend system for it.
2. Monitoring system (obtaining the signing keys for Alice+Bob+Cecile) will trigger calls to sign the multisig transaction.
3. Multisig, when it has the minimum signers, will fire away the `proveBlock()` transaction.



Options we have:
**A:** Using SAFE for firing away transactions (most of their codebase is open-source, but i think not the backend).
**B:** Custom solution

Current issues:
Currenlty Safe is not on Sepolia (at least in the UI, it is not supported).

Brechtpd:
I'll add these because they are also options, but they are very fancy options that are much more complex and only make sense if we want to distribute this among a lot of signers, which would get too expensive using standard signatures:
- https://geometry.xyz/notebook/Optimized-BLS-multisignatures-on-EVM
- https://ethereum.org/en/staking/dvt/

But again, much more complicated than standard signatures, so it would really have to be worth it to use in such a security critical part (so at this point: probably not).

Alternatively something like this could be something done as a required proof for each block, while something else that is much slower can be reserved for when something actually needs to be overridden.

dantaik:
[...]

PR #14923, opened 2023-10-09
feat(protocol): create contract to manage multi-sig prover for guardian proofs

dantaik:
See discussion [here](https://github.com/taikoxyz/taiko-mono/issues/14910). I'm assuming the % of blocks requires guardian proofs are very small. 

@adaki2004 you can help me to finish the PR and test it for gas-used, then compare with gnosis safe's solution. @Brechtpd I'm not concerned about the custom solution as is fairly easy.

### DIFF
--- a/packages/protocol/contracts/L1/GuardianProver.sol
+++ b/packages/protocol/contracts/L1/GuardianProver.sol
@@ -1,0 +1,82 @@
+// SPDX-License-Identifier: MIT
+//  _____     _ _         _         _
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/L1/GuardianProver.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: MIT
+//  _____     _ _         _         _
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+
+pragma solidity ^0.8.20;
```
dantaik, line 7: @adaki2004 please move this file to /L1/provers/

adaki2004: Ok, will do!

PR #14929, opened 2023-10-09
feat(protocol): reward proposer using gas used on L2

dantaik:
This PR has the following changes:
- L1 taiko token block reward (or proposer reward) has been moved from L1 to L2
- In each L2 block's anchor tx, block reward are minted to the parent block's propser, not the current block's proposer
- The amount of block reward now also depends on the block's gas used. L2 tracks a moving average of recent blocks' gas used, `avgGasUsed`, then for any block whose gas used is `gasUsed`, we calculate a ratio `gasUsed/avgGasUsed` then cap it to `2.0`, then we mint `maxBlockReward * min(2.0, gasUsed/avgGasUsed)`, where `maxBlockReward` is calculated the same way as previously. 

This PR has the following benefit:

- The gas used for proposing a block is reduced
- Proposing empty blocks will end up receiving very small block reward.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -34,9 +34,6 @@ library TaikoData {
         uint32 blockFeeBaseGas;
         // The maximum allowed bytes for the proposed transaction list calldata.
         uint24 blockMaxTxListBytes;
-        // Amount of token to reward to the first block propsoed in each L1
-        // block.
-        uint128 proposerRewardPerL1Block;
         uint128 proposerRewardMax;
         uint8 proposerRewardPoolPctg;
         // ---------------------------------------------------------------------
@@ -102,7 +99,6 @@ library TaikoData {
         uint64 l1Height;
         uint32 gasLimit;
         address coinbase; // L2 coinbase
-        uint96 rewardBase;
         TaikoData.EthDeposit[] depositsProcessed;
     }
 
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -30,12 +30,8 @@ library TaikoData {
         uint64 maxBlocksToVerifyPerProposal;
         // The maximum gas limit allowed for a block.
         uint32 blockMaxGasLimit;
-        // The base gas for processing a block.
-        uint32 blockFeeBaseGas;
         // The maximum allowed bytes for the proposed transaction list calldata.
         uint24 blockMaxTxListBytes;
-        uint128 proposerRewardMax;
-        uint8 proposerRewardPoolPctg;
         // ---------------------------------------------------------------------
         // Group 3: Proof related configs
         // ---------------------------------------------------------------------
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -102,6 +102,7 @@ library TaikoData {
         uint64 l1Height;
         uint32 gasLimit;
         address coinbase; // L2 coinbase
+        uint96 rewardBase;
         TaikoData.EthDeposit[] depositsProcessed;
     }
 
--- a/packages/protocol/contracts/L2/TaikoL2.sol
+++ b/packages/protocol/contracts/L2/TaikoL2.sol
@@ -26,6 +26,14 @@ import { TaikoL2Signer } from "./TaikoL2Signer.sol";
 contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
     using LibMath for uint256;
 
+    struct Config {
+        uint64 blockGasTarget;
+        uint256 basefeeAdjustmentQuotient;
+        uint256 proposerRewardPerL1Block;
+        uint256 proposerRewardMax;
+        uint256 proposerRewardPoolPctg;
+    }
+
     struct VerifiedBlock {
         bytes32 blockHash;
         bytes32 signalRoot;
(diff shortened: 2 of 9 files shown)


Brechtpd:
Not sure I understand the purpose of the average gas used, could you explain more?

Otherwise reposting from Discord because I think no matter what is done, this is a hard limitation:

> For an L1, putting transactions in a block is basically free. So that means that as long as a transaction pays a `fee > 0` (or can make money in another way), it can be included and be profitable. For an L2 however, this is not the case anymore because including a transaction in a block has a cost: The tx data cost, and for a zk rollup the cost to prove the execution of that transaction. So that creates a lower bound for which a proposer would ever include a tx inside an L2 block, because if the tx cannot at least pay for that, not including it will be more profitable. That means that a block reward subsidy is only helpful in getting transactions included between this `tx_data_cost + tx_proving_cost` and `tx_total_cost`, where the tx total cost also includes all the general overhead costs (transaction costs without the data part, proof verification) mostly shared between all the transactions. 

> So for example if there are only transactions that pay a fee below this lower bound in the mempool, an empty block will be the most profitable block to create. In which case the subsidy isn't really helpful to get people's transactions included, or keep all the actual proposers/provers live because creating and proving empty blocks is not representative to the actual work that needs to be done with full blocks.

> In the cases where the subsidy isn't necessary, the complete subsidy is sent to the L1 validator because only the proposer that pays the most to the L1 validator gets his block included.

I think taking `gas_used` into account doesn't really change anything to this, it only makes creating an "empty" block a little more expensive. I don't think it helps to get more real transactions included because it doesn't impact the profitability of a transaction, which it the only important part in the decision to include the tx in a block or not. How expensive a (fake) empty block is is of no concern for this.

As a result, I don't think it's really possible to reward proposers. Either it's a subsidy for users so they can pay lower transaction costs, or it's free money for L1 validators.

dantaik:
Brecht, a counter argument is that if there are profitable transactions in L2 mempool, then the best strategy for a proposer may be getting those transaction included to maximize the income (L2 fees + block rewards). Even these transactions are hard to zk-prove, the L2 fees will most likely justify the extra proving cost. After all, proving cost is much smaller than L2 fees, otherwise, the zk rollup idea will become invalid.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -25,26 +26,38 @@ import { TaikoL2Signer } from "./TaikoL2Signer.sol";
...
+        uint128 blockRewardPoolMax;
+        uint8 blockRewardPoolPctg;
+    }
+
     struct VerifiedBlock {
         bytes32 blockHash;
         bytes32 signalRoot;
     }
 
+    uint32 public constant ANCHOR_GAS_DEDUCT = 40_000;
```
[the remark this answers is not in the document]
davidtaikocha: We have a script to run it in a foundry node, but it's not accurate neither, the best way to estimate it is to run it in internal devnet, will also do it later.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -139,51 +160,125 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
+        accumulatedReward -= maxBlockReward;
+
+        if (
+            parentGasUsed <= ANCHOR_GAS_DEDUCT
+                || avgGasUsed <= ANCHOR_GAS_DEDUCT || parentProposer == address(0)
+        ) {
+            return 0;
+        }
+
+        address tt = resolve("taiko", true);
```
davidtaikocha, line 267:
```suggestion
        address tt = resolve("taiko_token", true);
```
The name `"taiko"` on L2 is used (in L2 signal service) which will be resolved to this `ProxiedTaikoL2` contract, so we may need to use another name here: 

```solidity
bytes32 syncedSignalRoot = ICrossChainSync(resolve("taiko", false))
            .getCrossChainSignalRoot(signalProof.height);
```

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -139,51 +160,125 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
+        if (tt == address(0)) return 0;
+
+        // The ratio is in [0-20000]
```
davidtaikocha, line 270:
```suggestion
        // The ratio is in [0-200]
```

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -34,9 +34,6 @@ library TaikoData {
         uint32 blockFeeBaseGas;
```
davidtaikocha, line 34: Not related to this PR but seems this config is no longer used?

Brechtpd:
> Brecht, a counter argument is that if there are profitable transactions in L2 mempool, then the best strategy for a proposer may be getting those transaction included to maximize the income (L2 fees + block rewards). Even these transactions are hard to zk-prove, the L2 fees will most likely justify the extra proving cost. After all, proving cost is much smaller than L2 fees, otherwise, the zk rollup idea will become invalid.

Not sure I understand, because profitable transactions of course will be included, but only if they are above the lower bound `tx_data_cost + tx_proving_cost` (when subsidized sufficiently) or `tx_total_cost` (when not subsidized): 
> That means that a block reward subsidy is only helpful in getting transactions included between this tx_data_cost + tx_proving_cost and tx_total_cost, where the tx total cost also includes all the general overhead costs (transaction costs without the data part, proof verification) mostly shared between all the transactions.

> Even these transactions are hard to zk-prove, the L2 fees will most likely justify the extra proving cost. After all, proving cost is much smaller than L2 fees, otherwise, the zk rollup idea will become invalid.

Any transaction that pays less than `tx_data_cost + tx_proving_cost`  will never be included, no matter how big the subsidy.

The argument isn't really if zk rollups are cheaper or more scalable (scalable yes, cheaper not really obvious!), it's how the subsidy effects things where proposers are only driven by max profitability. Could the subsidy be used to do include transactions? Of course! But will it?

packages/protocol/contracts/L1/TaikoData.sol
dantaik: removed

dantaik:
> > Brecht, a counter argument is that if there are profitable transactions in L2 mempool, then the best strategy for a proposer may be getting those transaction included to maximize the income (L2 fees + block rewards). Even these transactions are hard to zk-prove, the L2 fees will most likely justify the extra proving cost. After all, proving cost is much smaller than L2 fees, otherwise, the zk rollup idea will become invalid.
> 
> Not sure I understand, because profitable transactions of course will be included, but only if they are above the lower bound `tx_data_cost + tx_proving_cost` (when subsidized sufficiently) or `tx_total_cost` (when not subsidized):
> 
> > That means that a block reward subsidy is only helpful in getting transactions included between this tx_data_cost + tx_proving_cost and tx_total_cost, where the tx total cost also includes all the general overhead costs (transaction costs without the data part, proof verification) mostly shared between all the transactions.
> 
> > Even these transactions are hard to zk-prove, the L2 fees will most likely justify the extra proving cost. After all, proving cost is much smaller than L2 fees, otherwise, the zk rollup idea will become invalid.
> 
> Any transaction that pays less than `tx_data_cost + tx_proving_cost` will never be included, no matter how big the subsidy.
> 
> The argument isn't really if zk rollups are cheaper or more scalable (scalable yes, cheaper not really obvious!), it's how the subsidy effects things where proposers are only driven by max profitability. Could the subsidy be used to do include transactions? Of course! But will it?

I don't anticipate that proposers will include unprofitable transactions; efforts to enforce this would likely be futile.

packages/protocol/contracts/L2/TaikoL2.sol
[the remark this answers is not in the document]
dantaik: Added a TODO for you.

davidtaikocha (approved):
Changes LGTM besides one conflicting file

dantaik:
Some real transactions in the L2 mempool are profitable, and the rest are not. By "profitable", I mean a transaction's fee is greater than the proposer's overall cost and can positively contribute to a proposer's net income. No proposer will or shall not include non-profitable real transactions (unless there is a good MEV opportunity from that transaction).

The block reward incentivizes proposers to include those profitable transactions even when the initial fees collected from all such transactions are still too small to cover the cost. With block rewards, these profitable transactions will be processed, and users may transact more to gradually make the new blocks to generate more yield to cover the cost. Again, those real transactions that don't increase a proposer's revenue will never be included. 

I hope this addresses your concerns.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -139,51 +160,125 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
+        config.blockRewardPoolPctg = 40; // 40%
+    }
+
+    // In situations where the network lacks sufficient transactions for the
+    // proposer to profit, they are still obligated to pay the prover the
+    // proving fee, which can be a substantial cost compared to the total L2
+    // transaction fees collected. As a solution, Taiko mints additional Taiko
+    // tokens per second as block rewards. It's important to note that if
+    // multiple blocks are proposed within the same L1 block, only the first one
+    // will receive the block reward.
```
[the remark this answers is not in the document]
dantaik: fixed

PR #14932, opened 2023-10-10
feat(protocol): implement a multisig GuardianProver

adaki2004:
See this initial PR here: https://github.com/taikoxyz/taiko-mono/pull/14923
And the issue/discussion here: https://github.com/taikoxyz/taiko-mono/issues/14910

Moved it to a new PR, because previous was based on branch `contestable-zkrollup`, but we need the sgx eventually, so easier to use this branch as base now.


packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -0,0 +1,117 @@
...
+        ) revert INVALID_PROOF();
+        bytes32 hash = keccak256(abi.encode(blockId, evidence, unprovable));
+
+        uint256 approvalBits = blocks[hash];
+        require(approvalBits != DONE);
+
+        approvalBits |= uint8(1 << getGuardianId(msg.sender));
+
+        if (isApproved(approvalBits)) {
+            if (unprovable) {
```
dantaik, line 74: we don't need `unprovable` as a parameter for this function because if the block is unprovable, then evidence.proof should be constructed to have the right value.

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -0,0 +1,117 @@
...
+        external
+    {
+        if (
+            evidence.proof.length != 0
+                && evidence.tier != LibTiers.TIER_GUARDIAN
+        ) revert INVALID_PROOF();
+        bytes32 hash = keccak256(abi.encode(blockId, evidence, unprovable));
+
+        uint256 approvalBits = blocks[hash];
+        require(approvalBits != DONE);
```
dantaik, line 69: revert SOMETHING

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -0,0 +1,117 @@
...
+            bytes memory data = abi.encodeWithSignature(
+                "proveBlock(uint64,bytes)", blockId, abi.encode(evidence)
+            );
+
+            (bool success, bytes memory result) =
+                resolve("taiko", false).call(data);
+            if (!success) {
+                revert PROVE_FAILED(result);
+            }
+            blocks[hash] = DONE;
```
dantaik, line 88: maybe we can remove `blocks[hash] = DONE;` and use `delete blocks[hash]`.

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    mapping(uint256 id => address guardian) public guardians;
+    mapping(bytes32 => uint256 approvalBits) public blocks;
+
+    uint256[48] private __gap;
+
+    error INVALID_GUARDIAN_SET();
+    error INVALID_GUARDIAN();
+    error INVALID_PROOF();
+    error PROVE_FAILED(bytes);
```
dantaik, line 29: PROVING_FAILED

packages/protocol/contracts/L1/provers/GuardianProver.sol
adaki2004: I think does not really matter where we construct this.

If we construct it off-chain, then we need to modify the 
```
        if (
            evidence.proof.length != 0
                && evidence.tier != LibTiers.TIER_GUARDIAN
        ) revert INVALID_PROOF();
```
Otherwise we just need to set `unprovable = true` off-chain, and do not have to bother by encoding the constant string to bytes.

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    /// @notice Adds guardians
+    /// @param multisigParticipants The address array of the guardians.
+    function addGuardians(address[] memory multisigParticipants)
+        external
+        onlyOwner
+    {
+        // Always send 5 addresses, even if you just want to modify 1 address,
+        // so that (Bob, 0, 0, 0, 0) will set Bob only with index 1.
+        if (multisigParticipants.length != MAX_GUARDIANS) {
```
dantaik, line 45: MAX_GUARDIANS => NUM_GUARDIANS

dantaik (changes requested):

packages/protocol/contracts/L1/provers/GuardianProver.sol
adaki2004: I think `delete` is not 100% good, because it does not signal if a transition is DONE or not. (Or done but deleted). I'd keep it as is.

adaki2004:
> Maybe we should one admin-only function: `replaceAddress(old, new)`

We can use the `changeGuardians()` for that.

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -0,0 +1,121 @@
...
+
+/// @title GuardianProver
+contract GuardianProver is EssentialContract {
+    event GuardiansAdded(address[]);
+
+    uint256 public constant NUM_GUARDIANS = 5;
+    uint256 public constant REQUIRED_GUARDIANS = 3;
+    uint256 private constant DONE = type(uint256).max;
+
+    mapping(address signer => uint256 id) public guardians;
```
dantaik, line 23: should we just maintain a list: `address[5] public signers`

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -0,0 +1,121 @@
...
+        external
+        onlyOwner
+    {
+        // Always send 5 addresses, even if you just want to modify 1 address,
+        // so that (Bob, 0, 0, 0, 0) will set Bob only with index 1.
+        if (multisigParticipants.length != NUM_GUARDIANS) {
+            revert INVALID_GUARDIAN_SET();
+        }
+
+        for (uint256 i = 1; i <= NUM_GUARDIANS; i++) {
```
dantaik, line 52: this doesn't remove the previous/replaced signer, right?

packages/protocol/contracts/L1/provers/GuardianProver.sol
dantaik: I'm in favor of removing the "evidence.proof.length != 0" check.

packages/protocol/contracts/L1/provers/GuardianProver.sol
dantaik: What about we emit an event instead of having `blocks[hash] = DONE;`

dantaik (changes requested):

dantaik (approved):

dantaik (approved):

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -0,0 +1,122 @@
...
+            if (guardian == address(0)) revert INVALID_GUARDIAN();
+
+            // In case there is a pending 'approval' and we call setGuardians()
+            // with
+            // an existing guardian but with different array position (id), then
+            // accidentally 2 guardian signatures could lead to firing away a
+            // proveBlock() transaction.
+            uint256 id = guardianIds[guardian];
+
+            if (id != 0) {
```
dantaik, line 62: @adaki2004 please review this part again.

PR #14935, opened 2023-10-11
Try merge Dani's idea with mine

dantaik:
Dani, please take a look.

### DIFF
--- a/packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
+++ b/packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
@@ -22,18 +22,28 @@ contract SgxVerifier is EssentialContract, IVerifier {
 
     uint256 public constant INSTANCE_EXPIRY = 180 days;
 
-    mapping(address instance => uint256 registeredAt) public instances;
+    event InstancesChanged();
+
+    struct Instance {
+        address addr;
+        uint64 effectiveSince;
+    }
+
+    Instance[] public instances;
+    // mapping(address instance => uint256 registeredAt) public instances;
 
     uint256[49] private __gap;
 
-    event InstanceRegistered(
-        address indexed replaced, address indexed instance, uint256 registeredAt
-    );
-
     error SGX_INVALID_AUTH();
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -101,59 +114,77 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
 
-        if (newInstance == address(0)) revert SGX_INVALID_INSTANCE();
-        if (instances[newInstance] != 0) revert SGX_INSTANCE_REGISTERED();
+        if (id != instances.length) {
+            // Move the last element to the emply slot
+            Instance memory last = instances[instances.length - 1];
+            instances[id - 1] = last;
+        }
+        instances.pop();
+    }
```
adaki2004, line 184: I have several problems with this.
1. How do we know which instance belongs to which ID ? (Especially if we change them with removal) Always indexing the chain ? Also SGX verifiers who wants to add with signature (`addInstanceBySgx`), needs to index for the correct id ? 
2. We want to remove Alice, she is with the 8th id, and what if there are no purged instances here?
Example:

instance.length = 10
Alice id = 8

`Instance memory last = instances[instances.length - 1];`
last is Bob
So after this:
`instances[7] = Bob;`
`instances[8] = Alice;`
`instances[9] = Dave;`

In this case why do we remove Dave ?

I feel this solution is unnecessarily complex, and would push to use this (older version) instead, as was also good with Brecht:
https://github.com/taikoxyz/taiko-mono/blob/4867aedcb0f6ae29ff8383b47da16e676bd3f253/packages/protocol/contracts/L1/verifiers/SGXVerifier.sol

Brechtpd:
Would be very helpful to have some explanation why this version would be preferred over the original one because it's hard to figure out from code alone. This way it's possible to efficiently discuss the different versions (that are all very similar because this contract really doesn't do very much).

adaki2004:
Closing this as agreed to go forward with the ID based mechanism.

PR #14936, opened 2023-10-11
feat(protocol): alpha-7: use EIP4844 blob for DA

dantaik:
@Brechtpd Please let me know if the way I use 4844 is correct in this PR.


packages/protocol/contracts/common/Lib4844.sol
```diff
@@ -0,0 +1,61 @@
...
+        bytes32 versionHash,
+        uint256 x,
+        uint256 y,
+        bytes1[48] memory commitment,
+        bytes1[48] memory proof
+    )
+        internal
+        view
+    {
+        if (x >= FIELD_ELEMENTS_PERBLOB) revert POINT_X_TOO_LARGE();
```
Brechtpd, line 38:
```suggestion
        if (x >= BLS_MODULUS) revert POINT_X_TOO_LARGE();
```

Though it seems like the precompile will also do this check.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -42,12 +53,29 @@ contract PseZkVerifier is EssentialContract, IVerifier {
         // Do not run proof verification to contest an existing proof
         if (isContesting) return;
 
-        bytes32 instance = getInstance(prover, evidence);
+        ProofData memory data = abi.decode(evidence.proof, (ProofData));
+        bytes32 instance = getInstance(prover, data.txListHash, evidence);
+
+        // Verify blob
+
+        // Question: What if x is larger than the blob data size
```
Brechtpd, line 61: x is just a random point in the field, not a point index

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -42,12 +53,29 @@ contract PseZkVerifier is EssentialContract, IVerifier {
...
+
+        // Question: how to calculate pointCommitment and pointProof
+        // offchain?
```
Brechtpd, line 64: It's a standard KZG commitment I think so should be plenty of libraries available to do this from the data. The proof does also depend on the randomness, so having it calculated directly by the prover next to the actual proof seems to make sense to me.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -42,12 +53,29 @@ contract PseZkVerifier is EssentialContract, IVerifier {
...
+        Lib4844.point_evaluation_precompile({
+            versionHash: blobVersionHash,
+            x: uint256(instance) % Lib4844.FIELD_ELEMENTS_PERBLOB,
```
Brechtpd, line 67:
```suggestion
            `x: keccak(blobVersionHash, 0)` % Lib4844.BLS_MODULUS,
```

The `0` is not correct but would need some extra data exposed by the proof (or already in the proof, but would have to be extracted, not sure what makes the most sense).

Brechtpd (commented):
General approach looks good to me.

packages/protocol/contracts/common/Lib4844.sol
dantaik: fixed

packages/protocol/contracts/4844/BlobHashReader.yulp
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+//  _____     _ _         _         _
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+
+// An implemenatation of IBlobHashReader
+object "BlobHashReader" {
```
dantaik, line 8: @davidtaikocha @Brechtpd take a look a this file and the way it's used. I'm not entire sure whether this works as I still dont' know how to fire a blob tx.

dantaik:
@Brechtpd one more question for you: is it possible to calculate the version hash of a blob before its tx is confirmed on chain using the blob's binary data. I assume it yes, but need to double check with you.

Brechtpd:
> @Brechtpd one more question for you: is it possible to calculate the version hash of a blob before its tx is confirmed on chain using the blob's binary data. I assume it yes, but need to double check with you.

I also think so yes, because as far as I can also see it really is just a simple KGZ commitment of the data itself, and that is then hashed and a version added to it. Looking at the geth code: https://github.com/protolambda/go-ethereum/blob/ff35a44c775cf34f45ff3e4202a63d8c6c836c37/crypto/kzg/kzg.go#L24C23-L24C39. The only external data is the KZG CRS value which is a fixed value.

It also looks like the versioned hash even has to be pre-computed and made part of the transaction itself so that it only needs to be verified and not computed by block builders: https://hackmd.io/@protolambda/blobs_l2_tx_usage#Blob-TX-contents

At the bottom of that page there's a also a nice explanation of how things work together with a zk proof that's pretty cool, would have been useful to have found this page earlier!

PR #14938, opened 2023-10-11
fix(protocol): fix a bug reported by Quillaudit

dantaik:
@adaki2004 Dani, this is a clear bug, but how come our A5 test bridge still works? or does it? Maybe you and @KorbinianK can take a look at some ERC20 bridge txs to check who received the bridged tokens.

### DIFF
--- a/packages/protocol/contracts/tokenvault/ERC20Vault.sol
+++ b/packages/protocol/contracts/tokenvault/ERC20Vault.sol
@@ -144,7 +144,7 @@ contract ERC20Vault is
         IBridge.Message memory message;
 
         (message.data, _amount) = _encodeDestinationCall({
-            user: message.user,
+            user: msg.sender,
             token: opt.token,
             amount: opt.amount,
             to: opt.to


adaki2004:
> @adaki2004 Dani, this is a clear bug, but how come our A5 test bridge still works? or does it? Maybe you and @KorbinianK can take a look at some ERC20 bridge txs to check who received the bridged tokens.

It indeed a bug but not modifying the working behavior by nature.

1. So when we call on the source chain sendToken it will trigger:
```
    function _encodeDestinationCall(
        address user,
        address token,
        address to,
        uint256 amount
    )
```

it basically puts together the `receiveToken()` parameters to be called with on the destination chain.

2. The receiveToken() params as follows:
```
    function receiveToken(
        CanonicalERC20 calldata ctoken,
        address from,
        address to,
        uint256 amount
    )
```

The user address in this case will be 0, but it does not matter (technically) because we use the from only in the even emitted, so not modifying the working behavior.

Good catch, it shall be indexable and so we need to fill this properly!

adaki2004 (approved):

dantaik:
Screenshot 2023-10-11 at 22 00 09

But we use the `to` address to send tokens/Ether, right? So it matters and user should not have received any tokens after briding.

adaki2004:
> Screenshot 2023-10-11 at 22 00 09
> 
> But we use the `to` address to send tokens/Ether, right? So it matters and user should not have received any tokens after briding.

Yes, but the `to` address is correct. This is the `from` address in this `receiveToken()` call, which only used in the event emitted.

davidtaikocha (approved):

PR #14944, opened 2023-10-12
feat(protocol): support multi-hop message bridging

dantaik:
This is a missing feature that was long over-looked.

TODO: add a test


packages/protocol/contracts/bridge/libs/LibBridgeProcess.sol
```diff
@@ -73,21 +77,40 @@ library LibBridgeProcess {
...
+
+            for (uint256 i; i < proofs.length - 1; ++i) {
+                IBridge.IntermediateProof memory iproof =
+                    abi.decode(proofs[i], (IBridge.IntermediateProof));
+                // perform inclusion check
+                verified = LibSecureMerkleTrie.verifyInclusionProof(
+                    bytes.concat(LibSignalService.getSignalSlot(app, signal)),
+                    hex"01",
+                    iproof.mkproof,
+                    iproof.signalRoot
```
adaki2004, line 97: I think supplying `iproof.signalRoot` with the arguemnts does not guarantee that the proof is not 'made-up' (tamper proof).

In current version, how we call the verifyInclusionProof is like that:
kép

So we know the signalRoot can be trusted because it is synced with the anchor, no ? 🤔

adaki2004: Maybe it is overkill and there are easier solutions, my .02:

To have a trusted `signalRoot`, so with an example `L2B` have a trusted , synched `L2A` and `L2C` signal root, i drafted a quick design.

So the L1 <-> L2 would not been touched !


BUT in case the direction is not like that (rather L2 <-> L2), we could have a registry contract which only out special relayer could update. See diagram:

kép

Now, when someone wants to bridge from `L2A` to `L2B` 2 proof needed:

1. First proof: the registry synced signal root confirm inclusion (of message)
2. Second proof: The L1 relayed (with anchor), common SignalService contract contains L2A SignalRoot as a storage slot with value: 1

In the UI / Client + also in the contract:
- If it is a L1 <-> L2 operation proof.length = 1
- Otherwise validate the first proof based on this `'Taiko Deployed Registry'` and the second shall be that signalRoot is within the L1 common relayed one.

packages/protocol/contracts/bridge/libs/LibBridgeProcess.sol
```diff
@@ -41,18 +44,19 @@ library LibBridgeProcess {
...
-    /// @param proof The proof of the message hash from the source chain.
+    /// @param proofs The list of merkle inclusion proofs.
     /// @param checkProof A boolean flag indicating whether to verify the signal
     /// receipt proof.
     function processMessage(
         LibBridgeData.State storage state,
         AddressResolver resolver,
         IBridge.Message calldata message,
-        bytes calldata proof,
+        bytes[] calldata proofs,
```
adaki2004, line 54: Same mechanism shall be applied to `recallMessage` ? It would be a bit more complex as that one requires full merkle proof from the Birdge 🤔

adaki2004: But it will be a bit more difficult, because we do not use the `SignalService` for `recallMessage` but full merkle proofs from the destination `Bridge`. But how can we do this '2 step' proving without it ? 🤔 
Maybe relaying over the `BrideRoot` (or the stateRoot of the L2) in LibVerifying.sol just like we relay the `signalRoot`:
```
                    ISignalService(resolver.resolve("signal_service", false))
                        .sendSignal(signalRoot);
```

Then we need 2 proofs: 1 signalProof, and 1 full merkle ? 🤔

packages/protocol/contracts/bridge/libs/LibBridgeProcess.sol
```diff
@@ -73,21 +77,40 @@ library LibBridgeProcess {
...
+            bytes32 signal = msgHash;
+
+            bool verified;
+
+            for (uint256 i; i < proofs.length - 1; ++i) {
+                IBridge.IntermediateProof memory iproof =
+                    abi.decode(proofs[i], (IBridge.IntermediateProof));
+                // perform inclusion check
+                verified = LibSecureMerkleTrie.verifyInclusionProof(
+                    bytes.concat(LibSignalService.getSignalSlot(app, signal)),
```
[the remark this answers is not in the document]
dantaik: Brecht, the `app` var will be updated in the loop, so the key will also be different.

packages/protocol/contracts/bridge/libs/LibBridgeProcess.sol
dantaik: You are right, recallMessage shall also be updated accordingly. Let me do it later today.

dantaik:
This PR is ready to review again, now all `proof` become`proofs` to support multiple-hop bridging, including recall message and processMessage.

There is an implementation change in how to verify a message has failed on the destination change: Now when a message's status is FAIED, we'll send a derived signal on the destination chain. On the source chain, we use the same function `isSignalReceived` to verify the signal is sent.

packages/protocol/contracts/bridge/libs/LibBridgeProcess.sol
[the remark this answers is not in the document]
Brechtpd: Ah yeah I shouldn't have said the same `app`, just meant that a signal slot specific key is generated for each loop iteration, while in some cases you may just want to use the direct value itself if the tree works like that. But if this only supports multiple levels of these signal specific trees, that makes sense.

adaki2004 (approved):
Functionally LGTM - deployment fails (maybe bc. of new libraries unlinked? - unf. the error message is not so descriptive).

dantaik:
closed due to https://github.com/taikoxyz/taiko-mono/pull/14954

PR #14951, opened 2023-10-13
refactor(protocol): refactor bridge refine code using TaikoL1's pattern

dantaik:
No logical changes involved, purely a refactor to make the overall contract code looks more like the way we organized TaikoL1's code.


adaki2004 (approved):
LGTM, but there i an issue (deployment).

PR #14954, opened 2023-10-13
feat(protocol): support multi-hop bridging & add SGX prover

dantaik:
The following changes have been made:

- Enabled multi-hop bridging (eg. L2a->L2b, and L1 -> L3), this is enabled by SignalService's new multi-hop signal verification.
- Signal's slot now takes `block.chainid` as an extra input.
- Recall message do not use status-storage to check if a message has failed on destination chain, now for failed message, a new/derived signal is sent on the destination chain. 
- Merged Bridge code into a simple file for simplicity
- Removed two unused contracts
- Fixed DeployOnL1 script
- Added SGX provers by Dani


packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -77,57 +78,69 @@ contract SignalService is ISignalService, EssentialContract {
...
+            signalRoot
         );
     }
 
     /// @notice Get the storage slot of the signal.
+    /// @param chainId The address's chainId.
     /// @param app The address that initiated the signal.
     /// @param signal The signal to get the storage slot of.
-    /// @return signalSlot The unique storage slot of the signal which is
+    /// @return  The unique storage slot of the signal which is
```
dantaik, line 132:
```suggestion
    /// @return The unique storage slot of the signal which is
```

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -77,57 +78,69 @@ contract SignalService is ISignalService, EssentialContract {
...
         view
-        validChainId(srcChainId)
-        validApp(app)
-        validSignal(signal)
         returns (bool)
     {
-        SignalProof memory signalProof = abi.decode(proof, (SignalProof));
-        bytes32 syncedSignalRoot = ICrossChainSync(resolve("taiko", false))
-            .getCrossChainSignalRoot(signalProof.height);
+        if (app == address(0)) return false;
```
dantaik, line 91: @adaki2004 Dani, this is the function that you need to carefully review. The multi-hop proof is supported in SignalService, not the Bridge.

Acutally it also make sense to remove `skipProofCheck` from the Bridge contract and add the function to SignalService.sol, but when I tried to do it, quite a few tests failed. Maybe you can give it a try.

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -77,57 +79,78 @@ contract SignalService is ISignalService, EssentialContract {
...
+        Proof memory p = abi.decode(proof, (Proof));
+        if (p.storageProof.length == 0) return false;
+
+        for (uint256 i; i < p.hops.length; ++i) {
+            if (p.hops[i].signalRoot == 0) return false;
+            if (p.hops[i].storageProof.length == 0) return false;
+        }
+
+        // Check a chain of inclusion proofs. If this chain is chainA, and the
+        // message is sent on chainC, and we have chiainB in the middle, we
```
dantaik, line 106:
```suggestion
        // message is sent on chainC, and we have chainB in the middle, we
```

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -77,57 +79,78 @@ contract SignalService is ISignalService, EssentialContract {
     }
 
     /// @inheritdoc ISignalService
-    function isSignalReceived(
+    function proveSignalReceived(
```
dantaik, line 82: @cyberhorsey @KorbinianK please also review this change as it will affect how proofs are generated and encoded.

cyberhorsey: thanks for the heads up. yes I will have to make some relayer abi changes. I'll make a PR targeting this branch in the morning.

packages/protocol/contracts/signal/SignalService.sol
adaki2004: > @adaki2004 Dani, this is the function that you need to carefully review. The multi-hop proof is supported in SignalService, not the Bridge.
> 
> Acutally it also make sense to remove `skipProofCheck` from the Bridge contract and add the function to SignalService.sol, but when I tried to do it, quite a few tests failed. Maybe you can give it a try.

Done.
You had also 2 questions (ERC721 and ERC1155 tests related)  and i'm reflecting here:
```
// TODO(dani): we have to overwrite Alice address, otherise
// test_onMessageRecalled_721 will fail. Do you know why?
```
```
// TODO(dani): why chaning Amilia's address will fail the test?
```

Because we need to create the `getPreDeterminedDataBytes()` pre-testing. (We need to gain this data from the source chain’s `sendToken()`’s `message.data` but in order to do that i need to pre-calculate it (off-the-official-tests), locally.
 This will bake in the `msg.sender `context, which is Alice/Amelia (in ERC721 it was former, otherwise latter). So this is why it fails, because it affects the predetermined data bytes. 
 Now I modified and Amelia is the caller, so a bit unified the 2 tests, but the reason was that - to not change them.

dantaik:
@adaki2004  shall we merge this PR to its base branch?

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -107,22 +107,25 @@ contract SignalService is EssentialContract, ISignalService {
...
         if (signalRoot == 0) return false;
 
         for (uint256 i; i < p.hops.length; ++i) {
             Hop memory hop = p.hops[i];
             bytes32 slot = getSignalSlot(
                 hop.chainId,
-                resolve(hop.chainId, "taiko", false),
-                hop.signalRoot
+                resolve(chainId, "taiko", false),
+                hop.signalRoot ^ bytes32(hop.chainId)
```
adaki2004: I think XOR is not necessary here, as `signalSlot` already contains the `chainId`.

adaki2004:
> @adaki2004 shall we merge this PR to its base branch?

Good to go from my side.

adaki2004 (approved):

PR #14969, opened 2023-10-16
docs(website): add relayer and eventindexer swagger docs, add back swap guide

dionysuzx:
@2manslkh i added back the swap guide as it seems users were looking for it, so i think we can keep it here for now. but i didn't add it back into the navbar, for the same reason we removed the guide (not really a "core" thing).

that being said, i think we'll eventually need a way to extend our guides beyond "core guides" and into more fun tutorials. for example, how to user taiko's eventindexer or relayer API to do something, or how to modify taiko's client to do something.

### DIFF
--- a/packages/website/package.json
+++ b/packages/website/package.json
@@ -19,13 +19,13 @@
   },
   "devDependencies": {
     "@heroicons/react": "^2.0.18",
-    "@types/node": "^20.8.3",
-    "@types/react": "^18.2.25",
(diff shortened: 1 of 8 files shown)


davidtaikocha (approved):

dantaik (approved):

2manslkh:
Maybe 🛠️Tinkerer Guides for building on top of existing products and 🔎Explorer Guides for ecosystem exploration?

PR #14971, opened 2023-10-16
fix(protocol): Revert long imports (hardhat seems issues parsing remappings)

adaki2004:
It seem that the pnpm package update with `latest` introduced some incompatibilities.

It is now not able to parse the remappings properly.
 `Error HH415: Two different source names` (A similar hardhat issue: https://github.com/NomicFoundation/hardhat/issues/4399)
 
 1. Renamed the mappings and use the old ones -> This issue is solved
 2. Now defi-wonderland (or the error is misleading) says no exported `lib/utils `during genesis tests.
 3. Reverted the defi-wonderland package to the version we had previously, but issue still persists.
 
 Could you have a look at on it too, please ?

PR #14972, opened 2023-10-16
refactor(protocol): fix issue introduced by short-imports

dantaik:


dantaik:
> @dantaik Hmm, I do not really get the purpose of this PR 🤔 . What i see here is, that you did (almost) the exact same thing as me, on the other branch you closed. https://github.com/taikoxyz/taiko-mono/pull/14971/files
> 
> Plus, you need to change ethers version to `"ethers": "^5.7.2",` otherwise it fails - as you see in the CI/CD pipeline.

Almost the same, but not the same. Fewer solidity files are changed.

BTW, so the problem is changed by ethers.js version or both the long imports AND the ethers.js version?

dantaik:
Let me bring back yours and merge it first.

PR #14989, opened 2023-10-19
fix(protocol): Fix LibAddress

adaki2004:
2 issues we had - and never figured it out before. Let me explain.

**Nr.1.:**
`AddressUpgradeable.sendValue` is not proper (not properly implemented, or documented up to debate) -> So i reverted back the old.

```
    function sendValue(address payable recipient, uint256 amount) internal {
        require(address(this).balance >= amount, "Address: insufficient balance");
        (bool success, ) = recipient.call{value: amount}("");
        require(success, "Address: unable to send value, recipient may have reverted");
    }
```

This function - if recipient is `0x0` (as our `EtherVault` is 0 and non-existent on L1) - would return the funds to `msg.sender()` and `success` would be `true`. -> Resulting the funds never actually sent to the Bridge.

**Nr. 2.:**
```
try IERC165Upgradeable(addr).supportsInterface(interfaceId) returns (
                bool _result
            ) {
                result = _result;
            } catch {}
```
If called on EOA (if `addr` is EOA) it will revert and that revert would not be catched by the try-catch ! It is misused here. (If we call it on a contract and it reverts, that is fine (because our trusted `ERCXXXTokenVault` would never revert), but not fine reverting on EOAs. )

Please see convo here: https://forum.soliditylang.org/t/call-for-feedback-the-future-of-try-catch-in-solidity/1497

### DIFF
--- a/packages/protocol/contracts/libs/LibAddress.sol
+++ b/packages/protocol/contracts/libs/LibAddress.sol
@@ -30,7 +30,7 @@ library LibAddress {
     /// @param amount The amount of Ether to send in wei.
     function sendEther(address to, uint256 amount) internal {
         // Check for zero-value or zero-address transactions
-        if (to == address(0)) return;
+        if (to == address(0)) revert ETH_TRANSFER_FAILED();
 
         // Attempt to send Ether to the recipient address
         (bool success,) = payable(to).call{ value: amount }("");
@@ -65,14 +65,11 @@ library LibAddress {
         view
         returns (bool valid)
     {
-        if (
-            AddressUpgradeable.isContract(addr)
-                && IERC1271Upgradeable(addr).isValidSignature(hash, sig)
-                    == EIP1271_MAGICVALUE
-        ) {
-            valid = true;
-        } else if (ECDSAUpgradeable.recover(hash, sig) == addr) {
-            valid = true;
+        if (AddressUpgradeable.isContract(addr)) {
+            return IERC1271Upgradeable(addr).isValidSignature(hash, sig)
+                == EIP1271_MAGICVALUE;
+        } else {
+            return ECDSAUpgradeable.recover(hash, sig) == addr;
         }
     }
 }
--- a/packages/protocol/contracts/libs/LibAddress.sol
+++ b/packages/protocol/contracts/libs/LibAddress.sol
@@ -36,7 +36,7 @@ library LibAddress {
         (bool success,) = payable(to).call{ value: amount }("");
 
         // Ensure the transfer was successful
-        if (!success) {
+if (!success)  revert ETH_TRANSFER_FAILED();
             revert ETH_TRANSFER_FAILED();
         }
     }
--- a/packages/protocol/contracts/libs/LibAddress.sol
+++ b/packages/protocol/contracts/libs/LibAddress.sol
@@ -20,10 +20,21 @@ import { IERC1271Upgradeable } from
 library LibAddress {
     bytes4 private constant EIP1271_MAGICVALUE = 0x1626ba7e;
 
-    /// @dev Wrap this into a new function so the parameter `to` is `address`
-    /// instead of `address payable`.
+    /// @dev Sends Ether to the specified address. It is recommended to avoid
+    /// using `.transfer()` due to potential reentrancy issues.
+    /// Reference:
+    /// https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now
+    /// @param to The recipient address.
+    /// @param amount The amount of Ether to send in wei.
     function sendEther(address to, uint256 amount) internal {
-        AddressUpgradeable.sendValue(payable(to), amount);
+        // Check for zero-value or zero-address transactions
+        if (amount == 0 || to == address(0)) return;
+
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/libs/LibAddress.sol
```diff
@@ -34,11 +45,12 @@ library LibAddress {
         view
         returns (bool result)
     {
-        try IERC165Upgradeable(addr).supportsInterface(interfaceId) returns (
-            bool _result
-        ) {
-            result = _result;
-        } catch { }
+        if (AddressUpgradeable.isContract(addr)) {
```
dantaik, line 48: if (!AddressUpgradeable.isContract(addr))  return false;

adaki2004: > if (!AddressUpgradeable.isContract(addr)) return false;

It returns false if not contract.

packages/protocol/contracts/libs/LibAddress.sol
```diff
@@ -20,10 +20,21 @@ import { IERC1271Upgradeable } from
...
+    /// https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now
+    /// @param to The recipient address.
+    /// @param amount The amount of Ether to send in wei.
     function sendEther(address to, uint256 amount) internal {
-        AddressUpgradeable.sendValue(payable(to), amount);
+        // Check for zero-value or zero-address transactions
+        if (amount == 0 || to == address(0)) return;
+
+        // Attempt to send Ether to the recipient address
+        (bool success,) = payable(to).call{ value: amount }("");
```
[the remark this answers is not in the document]
dantaik: Even if the data is "", we can still call the receive() function.

packages/protocol/contracts/libs/LibAddress.sol
dantaik: Yeah, just a trick to reduce the level of {}

dantaik: No worries.

packages/protocol/contracts/libs/LibAddress.sol
```diff
@@ -20,10 +20,25 @@ import { IERC1271Upgradeable } from
...
+    /// @dev Sends Ether to the specified address. It is recommended to avoid
+    /// using `.transfer()` due to potential reentrancy issues.
+    /// Reference:
+    /// https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now
+    /// @param to The recipient address.
+    /// @param amount The amount of Ether to send in wei.
     function sendEther(address to, uint256 amount) internal {
-        AddressUpgradeable.sendValue(payable(to), amount);
+        // Check for zero-value or zero-address transactions
+        if (to == address(0)) return;
```
dantaik, line 33: shouldn't we fail here instead of return?

packages/protocol/contracts/libs/LibAddress.sol
```diff
@@ -20,10 +20,25 @@ import { IERC1271Upgradeable } from
...
+
+        // Attempt to send Ether to the recipient address
+        (bool success,) = payable(to).call{ value: amount }("");
+
+        // Ensure the transfer was successful
+        if (!success) {
```
dantaik, line 39: ### 
```suggestion
if (!success)  revert ETH_TRANSFER_FAILED();
```

packages/protocol/contracts/libs/LibAddress.sol
adaki2004: No, because it may be an NFT transfer - no ether involved or so, so no fail.

dantaik: if to ==0 and we do not revert, then the Ether will remain the current contract, which doesn't seem correct.

packages/protocol/contracts/libs/LibAddress.sol
```diff
@@ -20,10 +20,23 @@ import { IERC1271Upgradeable } from
...
+    /// @dev Sends Ether to the specified address. It is recommended to avoid
+    /// using `.transfer()` due to potential reentrancy issues.
+    /// Reference:
+    /// https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now
+    /// @param to The recipient address.
+    /// @param amount The amount of Ether to send in wei.
     function sendEther(address to, uint256 amount) internal {
-        AddressUpgradeable.sendValue(payable(to), amount);
+        // Check for zero-value or zero-address transactions
+        if (to == address(0)) return;
```
dantaik, line 33: Shall we remove AddressUpgradeable import?

adaki2004: We can,but it has the `isContract()` check, which we need - or we can port over.

dantaik: I see, we still need it elsewhere.

dantaik (approved):

Brechtpd:
Looks good to me. Maybe one general comment is that there are no gas limits on any of these external calls, so depending on where they are used it may introduce some problems depending on where they are used, so probably good to add this as a warning to those functions.

PR #14992, opened 2023-10-20
feat(protocol): remove block reward

dantaik:
After thinking it over for a few days, I am gradually leaning towards not giving block rewards. If we want to ensure liveness during the bootstrap phase, we should propose a block ourselves every few seconds, without concerning if such behaviour will end up with a profit or a loss.

### DIFF
--- a/packages/protocol/contracts/L2/TaikoL2.sol
+++ b/packages/protocol/contracts/L2/TaikoL2.sol
@@ -11,7 +11,6 @@ import { ICrossChainSync } from "../common/ICrossChainSync.sol";
 import { ISignalService } from "../signal/ISignalService.sol";
 import { Proxied } from "../common/Proxied.sol";
 import { LibMath } from "../libs/LibMath.sol";
-import { TaikoToken } from "../L1/TaikoToken.sol";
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -28,37 +27,26 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
     mapping(uint256 blockId => bytes32 blockHash) public l2Hashes;
     mapping(uint256 l1height => ICrossChainSync.Snippet) public snippets;
 
     // A hash to check the integrity of public inputs.
     bytes32 public publicInputHash; // slot 3
-
     uint128 public gasExcess; // slot 4
-    uint128 public accumulatedReward;
-
     address public parentProposer; // slot 5
```
davidtaikocha, line 40: Can be removed too?

dantaik: Removed

Brechtpd (approved):

PR #14994, opened 2023-10-20
refactor(protocol): improve TaikoToken and allow pausing Bridge/Vaults

dantaik:
The following changes are made:

- AddressManager can set an name to address(0), but in reality, the in-production AddressManager's `getAddress` function should be a pure function.
- TaikoToken no longer has the `burn` or `mint` function
- No special handling for Taiko token in ERC20Vault.
- Removed unnecessary OZ imports from TaikoToken.sol
- Allow admin to pause Bridge and Vaults - @je4nt0nic @KorbinianK please note that our UI may have to reflect the pause status (true or false). when the Bridge/Vaults are paused, no user transactions are allowed.
- Disable pausing TaikoToken transfers
- @davidtaikocha EtherVault can no longer be address(0) now.


packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -234,8 +230,7 @@ contract ERC20Vault is
         if (token == address(0)) revert VAULT_INVALID_TOKEN();
 
         if (amount > 0) {
-            if (isBridgedToken[token] || token == resolve("taiko_token", true))
-            {
+            if (isBridgedToken[token]) {
```
adaki2004, line 233: TaikoToken (now) has no public burn function.
(I think special handling is somewhat necessary for that.)

packages/protocol/contracts/L1/TaikoToken.sol
```diff
@@ -80,42 +63,6 @@ contract TaikoToken is
...
-        public
-        onlyFromNamed2("erc20_vault", "taiko")
-    {
-        _mint(to, amount);
-    }
-
-    /// @notice Burns tokens from the specified address.
-    /// @param from The address to burn tokens from.
-    /// @param amount The amount of tokens to burn.
-    function burn(
```
adaki2004, line 109: We should not remove it (IMO) but just let it burnable - if anyone wants to burn.. be it. (Otherwise it effects the ERC20Vault as commented below.)

dantaik: Who would call it then? I think it's OK not to have it. If we really need it, for example, when we have a DAO to burn some tokens, we can upgrade the token contract. But before that, this feature won't be used intentionally.

adaki2004: > Who would call it then?

`ERC20Vault.sol `line 225 ? 🤔

dantaik:
@davidtaikocha could you please review Genesis related code change?

packages/protocol/contracts/L1/TaikoToken.sol
adaki2004: @dantaik

packages/protocol/contracts/tokenvault/ERC20Vault.sol
adaki2004: @dantaik I think TKO can be a bridged token so effectively this (ERC20Vault.sol line 234) will try to burn it -> But has no burn function, no ? 🤔

adaki2004: Ah so, we are not deploying TaikoToken on L2 ourselves ? 🤔 but let it be “bridged” via the BridgedERC20 ?
If so it is fine, then no comment!

Brechtpd (approved):

davidtaikocha:
> @davidtaikocha could you please review Genesis related code change?

One small issue is we should also init TaikoL2's `_reentry ` && `_paused ` with `1` in genesis generation script too (instead of leaving them unset as `0`), opened a PR here: https://github.com/taikoxyz/taiko-mono/pull/15002

davidtaikocha (approved):

PR #14995, opened 2023-10-20
feat(protocol): allow pause the Bridge and Vaults

dantaik:
Changes:

- Allow admin to pause Bridge and Vaults
- Disable pausing TaikoToken transfers

@je4nt0nic @KorbinianK please note that our UI may have to reflect the pause status (true or false). when the Bridge is paused, no user transactions are allowed.

Pending: https://github.com/taikoxyz/taiko-mono/pull/14996


adaki2004:
@dantaik I think we shall also halt all the inflow and outflow so pause the `depositEtherToL2()` at the same time, because it adds up to the `Bridge` / `EtherVault` amount (on respective chains) otherwise would make an issue analysis much harder.

dantaik:
> depositEtherToL2

But depositEtherToL2 is not done through the Bridge.

adaki2004:
> > depositEtherToL2
> 
> But depositEtherToL2 is not done through the Bridge.

No via the Bridge but the bridge is taking the custody of the funds. And in case of there is an issue , plus the possibility of other inflaw (via `depositEtherToL2()`) it would make the issue analysis much harder.
kép

dantaik:
> > > depositEtherToL2
> > 
> > 
> > But depositEtherToL2 is not done through the Bridge.
> 
> No via the Bridge but the bridge is taking the custody of the funds. And in case of there is an issue, the possibility of other inflaw (via `depositEtherToL2()`) it would make the issue analysis much harder. kép

I see your point now. Good catch, what if we simply require each L1 must have an EtherValut? Please see this PR: https://github.com/taikoxyz/taiko-mono/pull/14996, lets continue the discussion there.

adaki2004:
> > > > depositEtherToL2
> > > 
> > > 
> > > But depositEtherToL2 is not done through the Bridge.
> > 
> > 
> > No via the Bridge but the bridge is taking the custody of the funds. And in case of there is an issue, the possibility of other inflaw (via `depositEtherToL2()`) it would make the issue analysis much harder. kép
> 
> I see your point now. Good catch, what if we simply require each L1 must have an EtherValut? Please see this PR: #14996, lets continue the discussion there.

Personally i think the easiest solution is something like:
`IBridge(bridgeAddress).isPaused() ` in the `depositEtherToL2` ?
What do you think ? We would not break any tests, and everything would stay as is.

PR #14996, opened 2023-10-20
what if we require etherValut to be non-zero?

dantaik:
@adaki2004 do you remember why we allow 0-address for EtherVault on L1?

DO NOT MERGE -- tests are failing


adaki2004:
> @adaki2004 do you remember why we allow 0-address for EtherVault on L1?
> 
> DO NOT MERGE -- tests are failing

Yes, because it would have 'no effect' - it would just "return".
kép

packages/protocol/contracts/L1/libs/LibDepositing.sol
```diff
@@ -37,11 +38,7 @@ library LibDepositing {
             revert L1_INVALID_ETH_DEPOSIT();
         }
 
-        address to = resolver.resolve("ether_vault", true);
-        if (to == address(0)) {
-            to = resolver.resolve("bridge", false);
-        }
-        to.sendEther(msg.value);
+        resolver.resolve("ether_vault", false).sendEther(msg.value);
```
adaki2004, line 41: It still does not prevent depositing to the vault 🤔

dantaik: True, but I guess it's OK, isn't it?

adaki2004: I think we are exactly here (haven't solved it): https://github.com/taikoxyz/taiko-mono/pull/14995#issuecomment-1772414615

The only difference is that we do not store the funds within the Bridge, but the EtherVault.. Still the funds would be bridged over, what we are trying to avoid (in case paused).

dantaik: We are trying to avoid Bridge bug, if the bridge has a bug, we pause it, the Taiko Rollup still works with the bridge bing paused.

Then we can also pause the EtherVault. Once the Vault is paused, the releaseEther functions will be disabled, but sending Ether to the vault is still allowed.

adaki2004: I see the point but my main counter argument is: to investigate a possible Bridge/EtherVault issue (effectively) we need to shut down everything which is bridge related (inflow and outflow) to see the balances. In my mind, Bridge and EtherVault is highly connected to each other.

Imagine we need to investigate bridge/ethervault leakage/minting out of thin air, and somehow we do not allow to use the Bridge but people could directly send ether to the TaikoL1. It would complicate things.

adaki2004: If you are comfortable this way, then it is fine by me. Changes look good (tho tests are failing).

dantaik:
@adaki2004  please help me fix the tests.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -63,8 +64,6 @@ contract Bridge is EssentialContract, IBridge {
         _;
     }
 
-    receive() external payable { }
```
dantaik, line 66: if ether_vault is non-zero, the the Bridge shall not be allowed to receive token directly.

adaki2004: This causes the tests to fail. Why ? The EtherVault needs to send the funds from EtherVault to the Bridge. This code is needed.

adaki2004: Will add back in my test fix branch.

PR #14998, opened 2023-10-20
fix(protocol): Fix failing tests

adaki2004:
Per title.


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -64,6 +64,8 @@ contract Bridge is EssentialContract, IBridge {
         _;
     }
 
+    receive() external payable { }
```
dantaik, line 67: I wonder why we have to have this?

dantaik (approved):

Issue #14991, opened 2023-10-19
Update to walletconnect v3


KorbinianK:
### Describe the feature request

Look into WCv3, maybe it fixes some of the weird styling problems we currently have.

[Documentation](https://docs.walletconnect.com/web3modal/about)

### Describe alternatives you've considered

Description of the alternatives you've considered here.

### Additional context

Additional context here.

KorbinianK:
We should also go back to the single button, but I assume the chain icon is also not displayed in the button itself then? Please check.

We went with this approach of two buttons because the ETH balance did not update automatically but it seems you can update it by calling `fetchBalance`. [See here] (https://github.com/WalletConnect/web3modal/issues/1402). 
Maybe v3 does that automatically, if not, we can still call it ourselves in the relevant actions (separate PR). But let's go back to the old, single button as soon as we are done with L2-L2 and NFT.

KorbinianK:
Reverted until further investigation: https://github.com/taikoxyz/taiko-mono/pull/15069

PR #14999, opened 2023-10-20
feat(bridge-ui-v2): upgrade walletconnect from v2 to v3

xiaodino:
Before
Screen Shot 2023-10-22 at 5 15 50 PM
Screen Shot 2023-10-22 at 5 15 44 PM

After
Screen Shot 2023-10-22 at 5 12 01 PM

Screen Shot 2023-10-22 at 5 12 07 PM

Screen Shot 2023-10-22 at 5 12 42 PM

https://github.com/taikoxyz/taiko-mono/issues/14991


packages/pos-dashboard/package.json
KorbinianK, line 1: These files are not bridge related and should not be touched by this PR

packages/bridge-ui/src/components/Transactions/Transaction.svelte
KorbinianK, line 1: This is the old bridge ui

KorbinianK (changes requested):
You changed the old bridge code as well as PoS packages. This was a bridge UI v2 only task. Please make sure to only touch relevant code.

xiaodino:
> You changed the old bridge code as well as PoS packages. This was a bridge UI v2 only task. Please make sure to only touch relevant code.

Good point. Reverted those changes.

KorbinianK:
Screen Shot 2023-10-22 at 5 12 42 PM

Is there a way to display the chain icon here as well?

packages/bridge-ui-v2/src/libs/connect/web3modal.ts
```diff
@@ -1,57 +1,51 @@
...
+    '--w3m-color-fg-1': 'var(--primary-content)',
+    '--w3m-color-bg-1': 'var(--primary-background)',
+    '--w3m-color-bg-2': 'var(--neutral-background)',
+    '--w3m-color-overlay': 'none',
+    '--w3m-accent-color': 'var(--neutral-background)',
+    '--w3m-accent-fill-color': 'var(--dark-background)',
   },
-  ethereumClient,
-);
+  themeMode: (localStorage.getItem('theme') as 'dark' | 'light') ?? 'dark',
```
KorbinianK, line 50: The theme switching does not work correctly. It stays on the initial one until you refresh the page.

image

KorbinianK: Ok, so it is possible as per the demo:
image

It would also be great to combine the two "buttons" we have now manually and use the combined one from walletconnect again. Assuming they fixed the hook to update the balance.

xiaodino:
> image
> Is there a way to display the chain icon here as well? I would assume so, as the component has the name `networkPlaceholder`
> 
> **Update:** Ok, so it is possible as per the demo: image
> 
> It would also be great to combine the two "buttons" we have now manually and use the combined one from walletconnect again. Assuming they fixed the hook to update the balance.

I spent some time on displaying the chain icon. The source code https://github.com/WalletConnect/web3modal/blob/V3/packages/wagmi/src/client.ts#L32 doesn't expose a way to set the chain icon for networkPlaceholder nor for imageSrc. 

I don't find a hack way nor a private API to set it. It's a minor UI change. I will keep it as it is and will revisit this when I have more time in another PR. Would you agree?

packages/bridge-ui-v2/src/libs/connect/web3modal.ts
xiaodino: Good catch. Updated https://github.com/taikoxyz/taiko-mono/pull/14999/files#diff-555d989884c6cc2a88ae4fde18e1e462fc54d9c57c6666ee244fcee1cc21d837L13 to resolve it.

davidtaikocha (dismissed):

davidtaikocha (dismissed):

KorbinianK:
> > image
> > Is there a way to display the chain icon here as well? I would assume so, as the component has the name `networkPlaceholder`
> > **Update:** Ok, so it is possible as per the demo: image
> > It would also be great to combine the two "buttons" we have now manually and use the combined one from walletconnect again. Assuming they fixed the hook to update the balance.
> 
> I spent some time on displaying the chain icon. The source code https://github.com/WalletConnect/web3modal/blob/V3/packages/wagmi/src/client.ts#L32 doesn't expose a way to set the chain icon for networkPlaceholder nor for imageSrc.
> 
> I don't find a hack way nor a private API to set it. It's a minor UI change. I will keep it as it is and will revisit this when I have more time in another PR. Would you agree?

[Seems like a known issue](https://github.com/WalletConnect/web3modal/issues/1405) might have saved you some time. I will create a follow up issue

KorbinianK:
image
image

The text is barely visible on both themes, please update.

KorbinianK:
> I spent some time on displaying the chain icon. The source code https://github.com/WalletConnect/web3modal/blob/V3/packages/wagmi/src/client.ts#L32 doesn't expose a way to set the chain icon for networkPlaceholder nor for imageSrc.
> 
> I don't find a hack way nor a private API to set it. It's a minor UI change. I will keep it as it is and will revisit this when I have more time in another PR. Would you agree?

Ok, there is an open issue: https://github.com/WalletConnect/web3modal/issues/1405 might have saved you some time. I will create a follow up task for us

xiaodino:
> image image
> The text is barely visible on both themes, please update.


Good catch. I didn't notice that when setting the theming for Web3Modal. Updated
Screen Shot 2023-10-25 at 9 51 54 AM
Screen Shot 2023-10-25 at 9 56 17 AM

KorbinianK (approved):

davidtaikocha (approved):

PR #15000, opened 2023-10-20
feat(relayer): Relayer multihop

cyberhorsey:
This Pr adds functionality for multihop relayer bridge processing.

Main functionality:
- Waits for headers to sync between the source chain, all intermediary hop chains, and finally, the last hop chain to the destination chain, so all the proofs will be valid when calling `bridge.proveMessageReceived`.
- Generate the main proof
- If hops exist, generate the hop proofs, using the previous signalroot of the previous hop proof as the storage slot.
- Encode the new Signalproof with hops and submit.

TODO in future PR:
- Need to index `CrossChainSynced` events for source, all hops, and destination, so I can easily query when a block hash was synced, to make sure we wait for that header to be synced instead of "latest". on L1 to L3 it wont slow down anything too much, just a few blocks, becasue they sync in anchor tx. for L2 to L2 it will slow things down right now.

Example of L1-L3 hop on interval devnet processed: https://l3blockscoutapi.internal.taiko.xyz/tx/0x66fe2932d700b400df578614e7646c5a5e33cd13126ccf5166326a367e034ce5


packages/relayer/proof/encoded_signal_proof.go
```diff
@@ -47,16 +67,90 @@ func (p *Prover) EncodedSignalProof(
...
+		if err != nil {
+			return nil, 0, errors.Wrap(err, "hopSignalService.GetSignalSlot")
+		}
+
+		encodedHopStorageProof, nextSignalRoot, err := p.encodedStorageProof(
+			ctx,
+			hop.Caller,
+			hop.SignalServiceAddress,
+			common.Bytes2Hex(hopStorageSlotKey[:]),
+			int64(blockNum),
```
cyberhorsey, line 115: upon reviewing my own code, this only works for one hop, we need to add `blockNum` as a `hopParam`.

cyberhorsey: resolved now

dantaik (approved):

dantaik:
@cyberhorsey I notice there is an example TX, which is great, I was wondering if it works... :D

cyberhorsey:
> @cyberhorsey I notice there is an example TX, which is great, I was wondering if it works... :D

yep, I have ran it L1 to L3, but not L2 to L2 yet, or anything with more than a single hop (L1 to L4, etc). theoretically it should work since it's configured to support an arbitrary number of hops that should be encoded the same way the contract expects.

packages/relayer/processor/process_message.go
```diff
@@ -88,39 +89,118 @@ func (p *Processor) processMessage(
...
+			common.Bytes2Hex(key[:]),
+			msgBody.Event.Raw.BlockHash,
+			blockNum,
+		)
+	} else {
+		// get latest synced header since not every header is synced from L1 => L2,
+		// and later blocks still have the storage trie proof from previous blocks.
+		latestSyncedSnippet, err = p.destHeaderSyncer.GetSyncedSnippet(&bind.CallOpts{}, 0)
+		if err != nil {
+			return errors.Wrap(err, "taiko.GetSyncedHeader")
```
RogerLamTd, line 183:
```suggestion
			return errors.Wrap(err, "taiko.GetSyncedSnippet")
```

packages/relayer/proof/encoded_signal_proof.go
```diff
@@ -47,16 +68,90 @@ func (p *Prover) EncodedSignalProof(
...
+		signalServiceAddress,
+		key,
+		blockHeader.Height.Int64(),
+	)
+
+	if err != nil {
+		return nil, 0, errors.Wrap(err, "p.encodedStorageProof")
+	}
+
+	slog.Info("successfully generated main storage proof proof")
```
davidtaikocha, line 97:
```suggestion
	slog.Info("successfully generated main storage proof")
```

RogerLamTd (approved):

davidtaikocha (approved):

PR #15001, opened 2023-10-21
fix(bridge-ui-v2) Fix build

xiaodino:


packages/bridge-ui-v2/src/app.config.ts
```diff
@@ -49,7 +49,7 @@ export const transactionConfig = {
 };
 
 export const toastConfig = {
-  duration: 5000,
+  duration: 50000,
```
KorbinianK, line 52: As I wrote in my last PR, my next PR (already in draft) will fix the toasts causing the build to fail. Has nothing to do with the duration, and why should they be 50sec? Due to moving several toasts outside of bridge component, they weren't updated to the new version. The new PR can be merged Monday.

KorbinianK (changes requested):

KorbinianK (changes requested):
This doesn't fix the build problem.
Notifications should not be visible for 50 seconds!

As I wrote in my last PR, the fix is already prepared (draft PR). Can be merged Monday.

PR #15005, opened 2023-10-22
feat(pos-dashboard): upgrade wagmi from 0.12.16 to 1.4.4

xiaodino:


cyberhorsey:
I think pos dashboard can be deleted since it's only for Eldfell and we no longer use any of those staking mechanisms, unless we want to repurpose it to allow users to view events from the later staking methods and remove all the forms/etc.

PR #15010, opened 2023-10-23
feat(website): propose to change our slogan and short description

dantaik:
welcome constructive feedbacks :)

Lets discuss this in Istanbul.

### DIFF
--- a/packages/website/components/Home/Hero.tsx
+++ b/packages/website/components/Home/Hero.tsx
@@ -43,10 +43,10 @@ export function Hero() {
           <div className="relative z-10 w-3/4 pt-16 pb-20 text-left lg:py-48">
             <div className="pl-[max(env(safe-area-inset-left),1.5rem)]">
               <h1 className="font-grotesk text-4xl md:text-5xl tracking-tight text-neutral-900 dark:text-neutral-100">
-                A <span className="text-[#e81899]">based</span> rollup
+                <span className="text-[#e81899]">Frictionless</span> and <span className="text-[#e81899]">Secure</span> Ethereum Scaling
               </h1>
               <p className="font-groteskmedium mt-3 text-lg text-neutral-600 sm:text-xl md:mt-5 dark:text-neutral-100">
-                Decentralized, Ethereum-equivalent (Type 1) ZK-EVM.
+                Taiko is a type-1 zkEVM, a Based rollup with multi-proofs.
               </p>
               <div className="mt-10 flex flex-col sm:flex-row md:justify-left">
                 <div className="inline-flex mb-4 sm:mb-0 sm:mr-4">
--- a/packages/website/components/Home/Hero.tsx
+++ b/packages/website/components/Home/Hero.tsx
@@ -46,7 +46,7 @@ export function Hero() {
                 <span className="text-[#e81899]">Frictionless</span> and <span className="text-[#e81899]">Secure</span> Ethereum Scaling
               </h1>
               <p className="font-groteskmedium mt-3 text-lg text-neutral-600 sm:text-xl md:mt-5 dark:text-neutral-100">
-                Taiko is the first type 1 zkEVM rollup with multi-proofs.
+                Taiko is the first based rollup and type 1 zkEVM with multi-proofs.
               </p>
               <div className="mt-10 flex flex-col sm:flex-row md:justify-left">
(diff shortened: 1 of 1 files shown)


dantaik:
@d1onys1us I don't see the website is deployed on Vercel automatically to offer a preview

packages/website/components/Home/Hero.tsx
```diff
@@ -43,10 +43,10 @@ export function Hero() {
           <div className="relative z-10 w-3/4 pt-16 pb-20 text-left lg:py-48">
             <div className="pl-[max(env(safe-area-inset-left),1.5rem)]">
               <h1 className="font-grotesk text-4xl md:text-5xl tracking-tight text-neutral-900 dark:text-neutral-100">
-                A <span className="text-[#e81899]">based</span> rollup
+                <span className="text-[#e81899]">Frictionless</span> and <span className="text-[#e81899]">Secure</span> Ethereum Scaling
               </h1>
               <p className="font-groteskmedium mt-3 text-lg text-neutral-600 sm:text-xl md:mt-5 dark:text-neutral-100">
-                Decentralized, Ethereum-equivalent (Type 1) ZK-EVM.
+                Taiko is a type-1 zkEVM, a Based rollup with multi-proofs.
```
KorbinianK, line 49: Nott a fan of the double "a" in the description makes it somehow sound like we're just one of many. 

Maybe something more like:

> "Taiko is the first type1-zkEVM rollup with multi-proofs"

> "Taiko brings multi-proofs to the zk-EVM landscape"

But I suppose then we need the keyword "based" in the slogan again.

Brechtpd:
I would avoid using first, especially if it's also a prediction because we don't know if it will be true. It also seems a bit strange for us to focus on being first for something while being one of the last big zk-evms projects to launch (vs "Taiko is not aiming to be your first ZK-EVM. Taiko is aiming to be your last ZK-EVM" style messaging before).

KorbinianK:
> I would avoid using first, especially if it's also a prediction because we don't know if it will be true. It also seems a bit strange for us to focus on being first for something while being one of the last big zk-evms projects to launch (vs "Taiko is not aiming to be your first ZK-EVM. Taiko is aiming to be your last ZK-EVM" style messaging before).

I see your point, however to the best of our knowledge right now it is true, right? Making this statement true.

I liked "first" as it highlights that we have something new and unique (right now) that differentiates us from the other zk-EVMs. I also considered "only" but that is certainly going to change. 

Maybe we just claim it entirely? 
`"Taiko is the type1-zkEVM rollup with multi-proofs"`

When we use it verbally, we can then emphasise the "the".

dantaik:
> Taiko is the type1-zkEVM rollup with multi-proofs

Interesting, I like this idea. I sounds like: there is no other rollups like us, "Taiko is **the** (only) type1-zkEVM rollup with multi-proofs"*

mirzdame:
> > Taiko is the type1-zkEVM rollup with multi-proofs
> 
> Interesting, I like this idea. I sounds like: there is no other rollups like us, "Taiko is **the** (only) type1-zkEVM rollup with multi-proofs"*
> 
> Now we have:
> 
> 1. Taiko is **the** type1-zkEVM rollup with multi-proofs
> 2. Taiko is the **first** type1-zkEVM rollup with multi-proofs
> 3. Taiko is **the** based type1-zkEVM rollup with multi-proofs (Daniel's favorite)
> 4. Taiko is the **first** based type1-zkEVM rollup with multi-proofs
> 5. Taiko is **the** based rollup with multi-proofs and type1-zkEVM
> 6. Taiko is the **first** based rollup with multi-proofs and  type1-zkEVM
> 
> We also need to decide between:
> 
> 1. type-1 zkEVM
> 2. type 1 zkEVM
> 3. type1 zkEVM
> 4. type1-zkEVM

I would perhaps add 2 more to the list (& my favourite).

7. Taiko -- a permissionless & based Type 1 zkEVM 
8. Taiko -- a permissionless & based Ethereum-equivalent zkEVM

While I think our multi-proof attribute is very important, I propose keeping the slogan to the 3 main attributes that mean the most to the widest possible audience (both technical and non-technical). 1. permissionless, 2. based, 3. Type 1 (or we could use Ethereum-equivalent if we want to be less technical). Instead I would make sure that we cover multi-proof design in our sub-header and other key sections of our pitch/website/decks. 

Re. question on how to write T1, I vote for #2 (Type 1 zkEVM)

PR #15011, opened 2023-10-23
feat(protocol): allow using both blob & calldata for DA

dantaik:
This PR enables users to use EIP-4844 blob to carry the transaction list when a block is proposed. The calldata based approach still works.

When `txList` is empty, we'll try to read the blob version hash.


packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -77,20 +118,24 @@ contract PseZkVerifier is EssentialContract, IVerifier {
...
+            abi.encodePacked(
                 evidence.metaHash,
                 evidence.parentHash,
                 evidence.blockHash,
                 evidence.signalRoot,
                 evidence.graffiti,
-                prover
+                prover,
+                blobHash,
+                txListHash
```
Brechtpd, line 138: Will also need to contain the x and y points so we have them in the circuits.

packages/protocol/contracts/L1/verifiers/IVerifier.sol
```diff
@@ -15,6 +15,8 @@ interface IVerifier {
         uint64 blockId,
         address prover,
         bool isContesting,
+        bool usingBlob,
+        bytes32 blobHash,
```
Brechtpd, line 19: Maybe cleaner to make it part of the `BlockEvidence`?

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -21,6 +23,19 @@ contract PseZkVerifier is EssentialContract, IVerifier {
 
     error L1_INVALID_PROOF();
 
+    struct PointProof {
+        bytes32 txListHash;
```
Brechtpd, line 27: Why not call this one blobHash as well?

dantaik: My understanding is that this value is the keccak hash of the RLP-encoded txList, provided by the actual zk prover. When 4844 blob is used for DA, the protocol no longer care about the keccak hash of the RLP-encoded txList, but this value is used by the protocol to calculate a random x (the instance value in our impl).

Let me know if you understand it differently.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
dantaik: The x value is the instance itself, the y value cannot be calculated without x, which is the instance. so y should not be used to calculate the instance.

dantaik: See:

```solidity
Lib4844.evaluatePoint({
        blobHash: blobHash,
        x: uint256(instance) % Lib4844.BLS_MODULUS,
        y: pf.pointValue,
        commitment: pf.pointCommitment,
        proof: pf.pointProof
});
```

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -170,42 +180,43 @@ library LibProposing {
...
-        inputs[3] = uint256(meta.extraData);
-        inputs[4] = (uint256(meta.id)) | (uint256(meta.timestamp) << 64)
-            | (uint256(meta.l1Height) << 128) | (uint256(meta.gasLimit) << 192);
-        inputs[5] = uint256(uint160(meta.coinbase));
-        inputs[6] = uint256(keccak256(abi.encode(meta.depositsProcessed)));
-
-        assembly {
-            hash := keccak256(inputs, 224 /*mul(7, 32)*/ )
-        }
+        return keccak256(
```
dantaik, line 202: @smtmfft please review this change. If you still prefer the old approach let me know.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
Brechtpd: Ah yeah, I didn't see it was used like that! 

I think the keccak hash approach makes sense because it's easy to do and we need to support hashing the data anyway for the calldata method. There are a couple of other options but they are more complicated: https://github.com/taikoxyz/zkevm-circuits/pull/106#issuecomment-1752993436

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
Brechtpd: `x` is the instance after the modulus so some extra work in the circuit needed to calculate it, but shouldn't complicate things too much (hopefully).

We do really need `y` as part of the instance though so because the circuit needs to have access to it, that's why we calculate it in the smart contract so that we know the correct value.

I would propose only hashing the required data for the random point `x`: `keccak(blobHash, txListHash)`.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -126,14 +129,19 @@ library TaikoData {
     /// 10 slots reserved for upgradability, 3 slots used.
     struct Block {
         bytes32 metaHash; // slot 1
-        address assignedProver; // slot 2
+        // blobHash represents the EIP-4844 versioned hash of the blob that
+        // carries the transaction list, or the keccack hash of the RPP-encoded
```
adaki2004, line 133:
```suggestion
        // carries the transaction list, or the keccack hash of the RLP-encoded
```

packages/protocol/contracts/L1/verifiers/IVerifier.sol
adaki2004: Keeping it in the `BlockEvidence`so that API stay same (+ data related to evidence), i think too might be a good idea.

dantaik: Both  `usingBlob` and `blobHash` are sloaded from storage when a block is to be proven. If we move blobHash into evidence, we still need to check `evicence.blobHash == blobHash`, I don't think it's a good idea to request the prover to provide duplicate datea in evidence.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
dantaik: `the keccak hash approach makes sense` => no change? Sorry please be more specific :)

packages/protocol/contracts/L1/verifiers/IVerifier.sol
adaki2004: What if put `blobHash` into the `metaHash` , saving some gas ? So that we do not need to store the `blk.blobHash` but it is done indirectly via `metaHash`  (`function hashMetadata(TaikoData.BlockMetadata memory meta`)) ?

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
dantaik: Modified as suggested.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
Brechtpd: Yep no changes!

packages/protocol/contracts/L1/verifiers/IVerifier.sol
Brechtpd: > I don't think it's a good idea to request the prover to provide duplicate datea in evidence.

The prover doesn't necessarily has to provide the same `TaikoData.BlockEvidence` struct as struct used in `IVerifier`, but yeah that does mean creating a new struct and copying some values over to the `IVerifier` one.

dantaik:
> Going over the code with Dani, we noticed `txListStart` and `txListEnd` got removed. This means the same blob cannot contain multiple L2 blocks so potentially not efficient. I think it also means we don't know the length of the txList in the blob so we cannot check that it is valid.

Brecht, yes, currently there is no way to check blob length in the solidity code, so if we have the start/end index, they will have to be checked in circuits. But I guess I removed the start/end mainly because I don't think the same blobs are going to be reused a lot, maybe I'm wrong, but we can add this small feature back when we realize it's going to be useful.

Brechtpd:
> > Going over the code with Dani, we noticed `txListStart` and `txListEnd` got removed. This means the same blob cannot contain multiple L2 blocks so potentially not efficient. I think it also means we don't know the length of the txList in the blob so we cannot check that it is valid.
> 
> Brecht, yes, currently there is no way to check blob length in the solidity code, so if we have the start/end index, they will have to be checked in circuits. But I guess I removed the start/end mainly because I don't think the same blobs are going to be reused a lot, maybe I'm wrong, but we can add this small feature back when we realize it's going to be useful.

There's still the validity check concern if we don't know the length. We may need to add the total length of the txList in the blob data itself at the start and decode it from there, I'm not sure if our current encoding of the tx list allows us to set hard rules for validity. I forgot if the txList is also RLP encoded, if so that could be sufficient. @smtmfft

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -29,68 +44,106 @@ contract PseZkVerifier is EssentialContract, IVerifier {
...
+            instance = calcInstance({
+                prover: input.prover,
+                evidence: evidence,
+                pointValue: pf.pointValue,
+                metaHash: input.metaHash
+            });
+
+            Lib4844.evaluatePoint({
+                blobHash: input.blobHash,
+                x: calc4844PointEvalX(input.blobHash, pf.txListHash),
```
dantaik, line 71: @Brechtpd please take a look at this file. 
x is calculatd based on blobHash and txListHash, bacause of this, txListHash is currently not hashed into instance. Let me know if you think this is OK?

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -29,68 +44,106 @@ contract PseZkVerifier is EssentialContract, IVerifier {
...
 
+    function calc4844PointEvalX(
+        bytes32 blobHash,
+        bytes32 txListHash
+    )
+        public
+        pure
+        returns (uint256)
+    {
+        return uint256(blobHash ^ txListHash) % Lib4844.BLS_MODULUS;
```
dantaik, line 123: @Brechtpd Do you think this will work as well, as an cheaper alternative to keccak?

smtmfft:
> There's still the validity check concern if we don't know the length. We may need to add the total length of the txList in the blob data itself at the start and decode it from there, I'm not sure if our current encoding of the tx list allows us to set hard rules for validity. I forgot if the txList is also RLP encoded, if so that could be sufficient. @smtmfft

For circuit part, we need to know the length, as RLP(raw_bytes) need length to uniquely mapping to the txlist hash, but that value can be sent by contract? no need to in blob?

dantaik:
> > > Going over the code with Dani, we noticed `txListStart` and `txListEnd` got removed. This means the same blob cannot contain multiple L2 blocks so potentially not efficient. I think it also means we don't know the length of the txList in the blob so we cannot check that it is valid.
> > 
> > 
> > Brecht, yes, currently there is no way to check blob length in the solidity code, so if we have the start/end index, they will have to be checked in circuits. But I guess I removed the start/end mainly because I don't think the same blobs are going to be reused a lot, maybe I'm wrong, but we can add this small feature back when we realize it's going to be useful.
> 
> There's still the validity check concern if we don't know the length. We may need to add the total length of the txList in the blob data itself at the start and decode it from there, I'm not sure if our current encoding of the tx list allows us to set hard rules for validity. I forgot if the txList is also RLP encoded, if so that could be sufficient. @smtmfft

According to @smtmfft, RLP encoding has data size as part of it.

Screenshot 2023-10-26 at 15 57 23

smtmfft:
> > > There's still the validity check concern if we don't know the length. We may need to add the total length of the txList in the blob data itself at the start and decode it from there, I'm not sure if our current encoding of the tx list allows us to set hard rules for validity. I forgot if the txList is also RLP encoded, if so that could be sufficient. @smtmfft
> > 
> > 
> > For circuit part, we need to know the length, as RLP(raw_bytes) need length to uniquely mapping to the txlist hash, but that value can be sent by contract? no need to in blob?
> 
> Smart contract doesn't know the size of the blob, or we can always use the 4844 blob max size?



> > > There's still the validity check concern if we don't know the length. We may need to add the total length of the txList in the blob data itself at the start and decode it from there, I'm not sure if our current encoding of the tx list allows us to set hard rules for validity. I forgot if the txList is also RLP encoded, if so that could be sufficient. @smtmfft
> > 
> > 
> > For circuit part, we need to know the length, as RLP(raw_bytes) need length to uniquely mapping to the txlist hash, but that value can be sent by contract? no need to in blob?
> 
> Smart contract doesn't know the size of the blob, or we can always use the 4844 blob max size?

AFAIC, the only chance of malicious behavior in circuit is rlc(txlist_raw_bytes) == rlc(txlist_raw_bytes + padding_0), which leads to hash conflict. Put the length at the blob beginning could avoid this problem if contract does not know the size.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
Brechtpd: It needs to be part of the instance: https://github.com/taikoxyz/taiko-mono/pull/15031#discussion_r1372355040

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
Brechtpd: Would keep the keccak method because it's easier to do in the circuit and it only needs to be done once. Doing the xor also requires quite a lot of lookups, and we don't have a dedicated circuit to do that specifically like we do with keccak.

Brechtpd:
> According to @smtmfft, RLP encoding has data size as part of it.

I think it's okay for each tx independently because those are RLP encoded, but the list of the transactions itself is not (or is it @smtmfft, I remember at some point it was but I think it's not anymore right?).

Blobs are also big: 128KB! There's also only a very limited amount of blobs per L1 I think: current target is 3, so sharing the blob between L2 blocks seems required in the long term unless that number gets increased or L2s just all start submitting very large blocks. I think it would be reasonable to add back the start/end list indices and we can just force the start to be `0` for now. I think that's a reasonable approach unless there's a better way to do it that's simple and flexible.

dantaik:
> > There's still the validity check concern if we don't know the length. We may need to add the total length of the txList in the blob data itself at the start and decode it from there, I'm not sure if our current encoding of the tx list allows us to set hard rules for validity. I forgot if the txList is also RLP encoded, if so that could be sufficient. @smtmfft
> 
> For circuit part, we need to know the length, as RLP(raw_bytes) need length to uniquely mapping to the txlist hash, but that value can be sent by contract? no need to in blob?

Even if the data can be provided by the proposer in contract, I don't think it can be trusted.

smtmfft:
> I think it's okay for each tx independently because those are RLP encoded, but the list of the transactions itself is not (or is it @smtmfft, I remember at some point it was but I think it's not anymore right?).

txList has the length as it is still follows rlp encoding rule. But 1 reason we need extra length is because we need to cover invalid case where the decoder has no idea how long it is, and the txlist hash in circuit is not that reliable due to properties of RLC(0 padding agnostic)
I just imaged a very special case: say raw_txlist is 200 bytes which is un-decodable, but if padding with some 0s, the new_txlist whose length is say 210, it's then decodable. In circuit, because of RLC, raw_txlist & new_raw_txlist could be mapped to the same hash if the length is not specified by instance. Then, this invalid rlp can be decoded and be sent to following modules like evm. Although I don't see any immediate harm as the decoded tx is certainly not executable as padding 0 is always the signature, this is still a defect in the system.

BTW: is new Hi/Lo arch already gets rid of RLC??

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
dantaik: fixed

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
dantaik: fixed

dantaik:
> > I think it's okay for each tx independently because those are RLP encoded, but the list of the transactions itself is not (or is it @smtmfft, I remember at some point it was but I think it's not anymore right?).
> 
> txList has the length as it is still follows rlp encoding rule. But 1 reason we need extra length is because we need to cover invalid case where the decoder has no idea how long it is, and the txlist hash in circuit is not that reliable due to properties of RLC(0 padding agnostic) I just imaged a very special case: say raw_txlist is 200 bytes which is un-decodable, but if padding with some 0s, the new_txlist whose length is say 210, it's then decodable. In circuit, because of RLC, raw_txlist & new_raw_txlist could be mapped to the same hash if the length is not specified by instance. Then, this invalid rlp can be decoded and be sent to following modules like evm. Although I don't see any immediate harm as the decoded tx is certainly not executable as padding 0 is always the signature, this is still a defect in the system.
> 
> BTW: is new Hi/Lo arch already gets rid of RLC??

Not a direct comment, but do you think with 4844, the blob is always the fixed size, padding with zeros to fill the blob's default size?

smtmfft:
> > > I think it's okay for each tx independently because those are RLP encoded, but the list of the transactions itself is not (or is it @smtmfft, I remember at some point it was but I think it's not anymore right?).
> > 
> > 
> > txList has the length as it is still follows rlp encoding rule. But 1 reason we need extra length is because we need to cover invalid case where the decoder has no idea how long it is, and the txlist hash in circuit is not that reliable due to properties of RLC(0 padding agnostic) I just imaged a very special case: say raw_txlist is 200 bytes which is un-decodable, but if padding with some 0s, the new_txlist whose length is say 210, it's then decodable. In circuit, because of RLC, raw_txlist & new_raw_txlist could be mapped to the same hash if the length is not specified by instance. Then, this invalid rlp can be decoded and be sent to following modules like evm. Although I don't see any immediate harm as the decoded tx is certainly not executable as padding 0 is always the signature, this is still a defect in the system.
> > BTW: is new Hi/Lo arch already gets rid of RLC??
> 
> Not a direct comment, but do you think with 4844, the blob is always the fixed size, padding with zeros to fill the blob's default size?

It can, but we still need the right length, because in RLP spec, it ends at the last meaningful byte, so if length is with padding, the rlp decoder in theory should return an error.

Brechtpd:
Important thing that I don't want people to miss: only 3 blobs can be used in a single L1 block on average! And you always pay for the full blob. These parameters will likely be used for a long time, so we have to take them into account.

> BTW: is new Hi/Lo arch already gets rid of RLC??

RLC is still used for the keccak data that needs to be hashed.

dantaik:
@Brechtpd given that there is noway to get the blob length in solidity, I suggest that we take the bob length and encoding related questions/topics offline, maybe you wanna set up a call with @smtmfft ?

Brechtpd:
> @Brechtpd given that there is noway to get the blob length in solidity, I suggest that we take the bob length and encoding related questions/topics offline, maybe you wanna set up a call with @smtmfft ?

We can work around the length not being passed in by having it in the data stream itself, I don't think there's too much too discuss there, what we do will depend on what we get from the smart contract! Previously we had the start and end indices from the smart contract so we just could use those.

My biggest question is are we okay with just being able to do a single block per blob. To be able to support multiple blocks per blob we need at least an extra offset per block as an input from the smart contract. This could have a big impact on fees, especially because Taiko blocks may be at disadvantage because all L2s will be competing to get one of these 3 blobs per L1 block on average. So even if we don't want to support it right now, having the offset there and forcing it to 0 for now seems like a good way to make things more future proof so we don't have to keep changing the interface between the smart contract and proofs for this in the future.

dantaik:
> > @Brechtpd given that there is noway to get the blob length in solidity, I suggest that we take the bob length and encoding related questions/topics offline, maybe you wanna set up a call with @smtmfft ?
> 
> We can work around the length not being passed in by having it in the data stream itself, I don't think there's too much too discuss there, what we do will depend on what we get from the smart contract! Previously we had the start and end indices from the smart contract so we just could use those.
> 
> My biggest question is are we okay with just being able to do a single block per blob. To be able to support multiple blocks per blob we need at least an extra offset per block as an input from the smart contract. This could have a big impact on fees, especially because Taiko blocks may be at disadvantage because all L2s will be competing to get one of these 3 blobs per L1 block on average. So even if we don't want to support it right now, having the offset there and forcing it to 0 for now seems like a good way to make things more future proof so we don't have to keep changing the interface between the smart contract and proofs for this in the future.

I see what you mean. Did you mean 3 blobs per L1 bock, I read something like " a long term maximum of 16 blobs per block".

Let me to file an issue to add blob sharing support to the next version, for this A6 version, I just want to auditors to be able to review the basic blob support. It doesn't have to be the final protocol version.

Brechtpd:
These are the latest settings I know of: 
```
TARGET_BLOB_GAS_PER_BLOCK | 393216
GAS_PER_BLOB              | 2**17

MAX_BLOB_GAS_PER_BLOCK    | 786432
LIMIT_BLOBS_PER_TX        | 2**12
```

So `393216/2**17 = 3` blobs as the target. Similar to the block gas limit, a block can contain up to 2x that, but this increases the blob basefee, so the target being 3 is the one that can actually be sustained.

I don't know if these parameters could still change before mainnet, or when they would get raised at a later time. I think likely we can only expect a raise when DAS is actually used for blob data, so probably quite a while off in that case.

smtmfft:
> These are the latest settings I know of:
> 
> ```
> TARGET_BLOB_GAS_PER_BLOCK | 393216
> GAS_PER_BLOB              | 2**17
> 
> MAX_BLOB_GAS_PER_BLOCK    | 786432
> LIMIT_BLOBS_PER_TX        | 2**12
> ```
> 
> So `393216/2**17 = 3` blobs as the target. Similar to the block gas limit, a block can contain up to 2x that, but this increases the blob basefee, so the target being 3 is the one that can actually be sustained.

The blob is exclusively used by one tx sender, right? so I think the target 3 is not enough(workable) if all L2s are competing for them, we then will see another war type blob-war :)

dantaik:
> > These are the latest settings I know of:
> > ```
> > TARGET_BLOB_GAS_PER_BLOCK | 393216
> > GAS_PER_BLOB              | 2**17
> > 
> > MAX_BLOB_GAS_PER_BLOCK    | 786432
> > LIMIT_BLOBS_PER_TX        | 2**12
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So `393216/2**17 = 3` blobs as the target. Similar to the block gas limit, a block can contain up to 2x that, but this increases the blob basefee, so the target being 3 is the one that can actually be sustained.
> 
> The blob is exclusively used by one tx sender, right? so I think the target 3 is not enough(workable) if all L2s are competing for them, we then will see another war type blob-war :)

I think the blob can be used by as many transactions as possible before its deleted. We need to cache the blobVersionHash somewhere in our protocol...

Brechtpd:
> I think the blob can be used by as many transactions as possible before its deleted. We need to cache the blobVersionHash somewhere in our protocol...

It is possible, but then you decouple the data transaction from the propose transaction which will likely be pretty problematic because that data will be outdated/stolen quickly, so for a based rollup this is quite tricky if not all data/propose transactions are in the same L1 block.

Still, caching the block hash or not, offsets to the txLists in the blob would be needed.

dantaik:
> > I think the blob can be used by as many transactions as possible before its deleted. We need to cache the blobVersionHash somewhere in our protocol...
> 
> It is possible, but then you decouple the data transaction from the propose transaction which will likely be pretty problematic because that data will be outdated/stolen quickly, so for a based rollup this is quite tricky if not all data/propose transactions are in the same L1 block.
> 
> Still, caching the block hash or not, offsets to the txLists in the blob would be needed.

Will get this done in another PR.

dantaik:
@Brechtpd what pending comments do you think I need to address?

Brechtpd (approved):

Brechtpd:
Nothing sorry, just forgot to approve after all comments were addressed.

PR #15025, opened 2023-10-24
feat(protocol): update A6 configurations

davidtaikocha:
Still need some benchmarks for circuits later on, for internal review at first.


packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
```diff
@@ -82,8 +87,14 @@ contract TaikoA6TierProvider is ITierProvider {
...
     function getMinTier(uint256 rand) public pure override returns (uint16) {
-        if (rand % 1000 == 0) return LibTiers.TIER_SGX_AND_PSE_ZKEVM;
-        else if (rand % 100 == 0) return LibTiers.TIER_SGX;
+        // If we assume the block time is 3 seconds, and the proof generation
+        // time is ~90 mins
+        // and half of the blocks are unprovable: 90 * 60 / 3 / 2  = 900.
+        // TODO(david): tune this value based on the A6 circuits benchmark.
+        if (rand % 900 == 0) return LibTiers.TIER_SGX_AND_PSE_ZKEVM;
+        else if (rand % 100 == 0) return LibTiers.TIER_SGX; // 1% of the blocks
+            // will be slected to require a SGX proof.
```
Brechtpd, line 96: 1% seems low. Why not 100%? For testing purposes?

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -262,14 +262,28 @@ contract TaikoL1 is
...
-            blockRingBufferSize: 403_210,
+            chainId: 167_008,
+            // Assume the block time is 3s, the protocol will allow ~1 month of
+            // new blocks
+            // without any verification.
+            blockMaxProposals: 864_000,
+            blockRingBufferSize: 864_010,
+            // Can be overridden by the tier config.
             maxBlocksToVerifyPerProposal: 10,
+            // Limited by the PSE zkEVM circuits.
```
Brechtpd, line 279: Because people complain about it and there is a path for the PSE circuits to be able to support arbitrary block sizes, maybe we can just set this to 15M gas?

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -262,14 +262,28 @@ contract TaikoL1 is
...
             blockMaxGasLimit: 8_000_000,
+            // Each go-ethereum transaction has a size limit of 128KB,
+            // and right now txList is still saved in calldata, so we set it
+            // to 120KB.
             blockMaxTxListBytes: 120_000,
-            livenessBond: 10_240e18,
+            livenessBond: 10_240e18, // 10_240 Taiko token
```
Brechtpd, line 285: Seems like a strangely specific number, why isn't it just a round number? :)

packages/protocol/contracts/L1/TaikoL1.sol
davidtaikocha: Ah good news! Changed to 15M.

packages/protocol/contracts/L1/TaikoL1.sol
davidtaikocha: Changed to a round number: `10_000`

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
davidtaikocha: If so, then I think the optimistic tier in this A6 provider can be removed? emmm do we want to do that?

dantaik: Brecht, I think we should not make it 100% to also test the lowest tier. A a prover, we can always submit SGX proofs even if the block requires a lower tier proof.

packages/protocol/contracts/L1/TaikoToken.sol
```diff
@@ -47,8 +47,8 @@ contract TaikoToken is
         ERC20SnapshotUpgradeable.__ERC20Snapshot_init_unchained();
         ERC20VotesUpgradeable.__ERC20Votes_init_unchained();
 
-        // Mint 1 billion tokens
-        _mint(_recipient, 1_000_000_000 ether);
+        // Mint max Uint128 for testnet.
+        _mint(_recipient, type(uint128).max);
```
dantaik, line 51: why max? I think we should stick to 1 billion, it should just work, no?

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
dantaik: 10%?

packages/protocol/contracts/L1/TaikoToken.sol
davidtaikocha: yep it should work, but just thought 1 billion supply with current `10_000 liveness bond` + `20 * 10_000 optimistic proof validity bond`, even if we keep all tokens, we are only able to provide ~5000 blocks optimistic proof under 28 hrs (cooldown window), and if we assume 3 seconds as block time, we may have 1200 blocks per hour, so lets reduce the bond?

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -262,14 +262,28 @@ contract TaikoL1 is
...
+            blockRingBufferSize: 864_010,
+            // Can be overridden by the tier config.
             maxBlocksToVerifyPerProposal: 10,
-            blockMaxGasLimit: 8_000_000,
+            // Limited by the PSE zkEVM circuits.
+            blockMaxGasLimit: 15_000_000,
+            // Each go-ethereum transaction has a size limit of 128KB,
+            // and right now txList is still saved in calldata, so we set it
+            // to 120KB.
             blockMaxTxListBytes: 120_000,
```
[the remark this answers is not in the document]
davidtaikocha: image
Yep when a transaction comes into mempool, there will be a check that its size should be smaller than `txMaxSize`, but I also think that it's just a limitation set by geth, and not by Ethereum protocol.

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
davidtaikocha: Changed to 10%

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -262,14 +262,27 @@ contract TaikoL1 is
         virtual
         returns (TaikoData.Config memory)
     {
+        // All hard-coded configurations:
+        // - treasury: 0xdf09A0afD09a63fb04ab3573922437e1e637dE8b
+        // - blockMaxTxs: 80 (limited by the PSE zkEVM circuits)
```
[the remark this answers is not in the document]
davidtaikocha: No it's just hard-coded in both circuits and node at this point.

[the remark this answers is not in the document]
davidtaikocha: But if there is an optimization in circuits for increasing `block.gasLimit` to 15M, are we safe to increase this number too? @Brechtpd

packages/protocol/contracts/L1/TaikoToken.sol
davidtaikocha: yep 250 liveness bond makes sense to me, changed: https://github.com/taikoxyz/taiko-mono/pull/15025/commits/2af5f28c3c5aae10be2aa10cb81a5ce5cd9d7396

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -262,14 +262,27 @@ contract TaikoL1 is
...
             maxBlocksToVerifyPerProposal: 10,
-            blockMaxGasLimit: 8_000_000,
+            // Limited by the PSE zkEVM circuits.
+            blockMaxGasLimit: 15_000_000,
+            // Each go-ethereum transaction has a size limit of 128KB,
+            // and right now txList is still saved in calldata, so we set it
+            // to 120KB.
             blockMaxTxListBytes: 120_000,
-            livenessBond: 10_240e18,
+            livenessBond: 10_000e18, // 10_000 Taiko token
```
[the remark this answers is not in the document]
davidtaikocha: Changed to 250 based on the latest comment.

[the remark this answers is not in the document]
davidtaikocha: Changed to 250 based on the latest comment: https://github.com/taikoxyz/taiko-mono/pull/15025#discussion_r1371007603

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -262,14 +262,27 @@ contract TaikoL1 is
         virtual
         returns (TaikoData.Config memory)
     {
+        // All hard-coded configurations:
+        // - treasury: 0xdf09A0afD09a63fb04ab3573922437e1e637dE8b
+        // - blockMaxTxs: 80 (limited by the PSE zkEVM circuits)
+        // - anchorGasLimit: 250_000 (based on internal devnet, its ~220_000
```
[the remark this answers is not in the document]
davidtaikocha: Right, it's part of the L2 consensus, all anchor transactions should use this value as gas limit.

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
```diff
@@ -10,7 +10,11 @@ import { ITierProvider, LibTiers } from "./ITierProvider.sol";
 
 /// @title TaikoA6TierProvider
 contract TaikoA6TierProvider is ITierProvider {
-    uint96 private constant UNIT = 10_000e18; // 10000 Taiko token
+    uint96 private constant UNIT = 10_000e18; // 10_000 Taiko token (equal to
```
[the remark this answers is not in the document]
davidtaikocha: Changed.

packages/protocol/contracts/L1/TaikoL1.sol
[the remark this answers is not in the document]
Brechtpd: Yeah should be okay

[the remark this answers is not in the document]
davidtaikocha: Sounds good updated to 150: https://github.com/taikoxyz/taiko-mono/pull/15025/commits/381a2b7ec4a0f95ca80b5204518216497db2e8e1 cc @smtmfft

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -173,7 +175,11 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
 
     /// @notice Returns EIP1559 related configurations
     function getConfig() public pure virtual returns (Config memory config) {
-        config.gasTargetPerL1Block = 15 * 1e6 * 10; // 10x Ethereum gas target
+        // 4x Ethereum gas target, if we assume most of the time, L2 block time
+        // is 3s, and each block is full (gasUsed is 15_000_000), then its
+        // ~60_000_000, if the  network is congester than that, the base fee
+        // will increase.
+        config.gasTargetPerL1Block = 15 * 1e6 * 4;
```
dantaik, line 182: Initially the network won't be used much, so the L1 capacity may higher than our expected block size. So maybe on top of this calculation, divide it by a factor of 8? Then gradually we change the factor to 1?

dantaik: Otherwise, we'll always see 0 basefee

davidtaikocha: For testnet, what we observed is that the network's usage is just simply highly related to community campaigns, but yeah for mainnet, that factor makes sense to me.

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
```diff
@@ -82,8 +80,14 @@ contract TaikoA6TierProvider is ITierProvider {
     }
 
     function getMinTier(uint256 rand) public pure override returns (uint16) {
-        if (rand % 1000 == 0) return LibTiers.TIER_SGX_AND_PSE_ZKEVM;
-        else if (rand % 100 == 0) return LibTiers.TIER_SGX;
+        // If we assume the block time is 3 seconds, and the proof generation
+        // time is ~90 mins and half of the blocks are unprovable:
+        // 90 * 60 / 3 / 2 = 900.
+        // TODO(david): tune this value based on the A6 circuits benchmark.
+        if (rand % 900 == 0) return LibTiers.TIER_SGX_AND_PSE_ZKEVM;
```
dantaik, line 87: This doesn't have to be really accurate.I prefer 1000 or 500, easier for manual calculation.

dantaik: I think @Brechtpd would prefer a higher number. so lets settle for 500 first then increate it if our SGX servers are not busy?

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
```diff
@@ -9,10 +9,8 @@ pragma solidity ^0.8.20;
 import { ITierProvider, LibTiers } from "./ITierProvider.sol";
 
 /// @title TaikoA6TierProvider
+/// @dev Assuming liveness bound is 250TKO.
```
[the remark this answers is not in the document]
davidtaikocha: Added

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
davidtaikocha: Changed to `500`

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
```diff
@@ -82,8 +84,11 @@ contract TaikoA6TierProvider is ITierProvider {
     }
 
     function getMinTier(uint256 rand) public pure override returns (uint16) {
-        if (rand % 1000 == 0) return LibTiers.TIER_SGX_AND_PSE_ZKEVM;
-        else if (rand % 100 == 0) return LibTiers.TIER_SGX;
+        if (rand % 500 == 0) return LibTiers.TIER_SGX_AND_PSE_ZKEVM; // 0.2% of the blocks
```
smtmfft, line 87: One remaining concern here is this randomly selected block is unprovable due to unsupported tx, high rws, or some unknown issues(hidden bugs). 
For example, if a popular partner who still use legacy tx launches on taiko, like the one we met in A5, for a period of time all blocks are filled with at least on of their txs, which makes all blocks in that period unprovable, no matter what the rand is.
Should we have provable blocks (could be pure anchor block) at fixed intervals playing as ZK checkpoints without purely rely on rand?

dantaik: How can you have provable blocks at fixed intervals? If a block is not provable, we need to prove it using our guardian prover.

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: So should we divided this number by a factor? Our testnet volume won't be larger than Ethereum mainnet, in terms of gas.

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
smtmfft: Yes, here we rely heavily on oracle, I just thought besides the rand picking, maybe we can intentionally submit a few zk blocks to kind of refresh the zk ratio requirement.

packages/protocol/contracts/L2/TaikoL2.sol
adaki2004: Did some simulation. 

About "dynamically changing" values in the protocol:
Keeping track of the average gas / L1 variable and then adjust the behavior (all 3 values: gasExcess, target, quotient) to “show” a good look (like 1-10gwei baseFee) can technically be feasible but very challenging, computation heavy and is fee market manipulation, so I would not go this way.

 Also very important to note that the 3 values (gasExcess, target, quotient) are heavily interconnected ! If we change the target (+quotient) the gasExcess needs to change as well ! Otherwise the 1559 math will not work.
 
 When **60mil** and adjustment quotient is **8**, gas excess (calculated) is **19542030010** .
 
 THis is the simulation with the parameters (you can see in the labels). 
 
**Above target:**
block (if gas
**Below target:**
block (if gas
**Around target:**
1698392033_target_traffic


  When we run the same simulation with a very low (wrong gas excess ) value: 1 it looks like this:  ( See y-axis with 1e-9) 
  
block (if gas
below target)
1698392540_target_traffic


  When we use a huge one: 19542030010 (original good) * 10  (See y-axis with 1e41)
  
1698392740_above_average_traffic
below target)
Basefee chart
 
 So this is my proposal:
**1.** Do not set the init gasExcess to wrong value :) (What would be the initial correct value, see point  
**2.** What is the most important (and currently lacking in TaikoL2) to have a setter to modify these 3 variables all together otherwise things would look more ugly.
**3.** Monitor A5 - what was the traffic in the first 5 days, I mean average gasUsed / L1 block, and use this as our base target, and calculate the gasExcess accordingly. It is a good starting point, and when things will differ highly, just use the function in point 2.

dantaik: @adaki2004 could you please add the setter function? We can still use a not-so-accurate initial gasExcess then use the function to turn the parameter later on. I don't think picking the initial value right for gasExcess shall be important.

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
dantaik: We can certainly do that, and prove it proactively even with ZKPs even if ZKPs are not required.

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: see https://github.com/taikoxyz/taiko-mono/pull/15052

dantaik (approved):

dantaik (approved):

Brechtpd (approved):

PR #15027, opened 2023-10-24
feat(protocol): Enable designated address to propose 1st block post-genesis.

dantaik:
This feature allows a specific address to exclusively propose the first block in Taiko following mainnet launch.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -45,6 +45,7 @@ library LibProposing {
     error L1_UNAUTHORIZED();
 
     /// @dev Proposes a Taiko L2 block.
+
     function proposeBlock(
         TaikoData.State storage state,
         TaikoData.Config memory config,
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -61,10 +62,8 @@ library LibProposing {
         // However, if the "proposer" address is set to a non-zero value, we
         // ensure that only that specific address has the authority to propose
         // blocks.
-        address proposer = resolver.resolve("proposer", true);
-        if (proposer != address(0) && msg.sender != proposer) {
-            revert L1_UNAUTHORIZED();
-        }
+        TaikoData.SlotB memory b = state.slotB;
+        if (!_isProposerPermitted(b, resolver)) revert L1_UNAUTHORIZED();
```
adaki2004, line 66: I'd add `if (slotB.numBlocks == 1)` here first (not in the `_isProposerPermitted`) because this will be called at every `proposeBlock()`, so in case we do not check it here, we will have 2 operations:

JUMP: 8 gas
EQ (equal): 3 gas

So "long term", cheaper to check the 1st block here. Otherwise LGTM.

dantaik: good feedback

davidtaikocha (approved):

PR #15031, opened 2023-10-24
feat(protocol): Blob hash to meta

adaki2004:


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -196,8 +196,6 @@ library LibProving {
                     prover: msg.sender,
                     isContesting: evidence.tier == tran.tier
                         && tier.contestBond != 0,
-                    usingBlob: blk.usingBlob,
-                    blobHash: blk.blobHash,
```
Brechtpd, line 200: `evidence.blobHash` is passed in here, but it doesn't seem to be verified anywhere that it is indeed the same `blobHash` that was hashed inside `blk.metaHash`. To do that, the prover will have to provide all the meta data again, and then that data needs to be hashed again and checked that the hash matches the one in `blk.metaHash`.

dantaik: Good finding. Dani, maybe we should keep the code as-is if  passing in metadata as one additional parameter then hash it will probably cost more gas then storing these two variables (after ring buffer loops)?

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -135,13 +138,12 @@ library TaikoData {
         bytes32 blobHash; // slot 2
```
dantaik, line 138: Dani, if we remove `usingBlob`, we can also remove `blobHash` from Block, move both of them to Metadata.

dantaik: Then when a block is to be proven, the prover will also have to pass in the metadata again then check the hash against blk.metaHash.

adaki2004:
@dantaik @Brechtpd 

Thanks for the feedback. I added the `BlockMetadata` to the proving and checking if it makes up the same `metahash` in `LibProving`.

Made some gas-consumption benchmark tests, and this is expected. ( `config.blockMaxProposals = 10 config.blockRingBufferSize = 12` in `TaikoL1.t.sol`).

This one's TaikoL1 tests:
kép

This one's Tiers tests:
kép

Base (4844) TaikoL1 tests:
kép

Base (4844) Tiers tests:
kép


Result TLDR (as expected):
- proposeBlock: -> around 18.000 gas cheaper (as expected due to no storage writes)
- proveBlock: -> around 1.700 gas more expensive (extra calldata)
- verifyBlock: -> (no significant change)

Please let me know if this solution is still lacking something or not. If not, then make a call which one to use.

I like this one better because we did not break the interface (for verifyProof() with 2 extra variables) and also significantly cheaper.

dantaik:
Dani, I think 1.700 more gas is fine, it's better than changing a slot from 1 to 2, which is 2,900 more gas. So this new implementation is cheaper.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -80,6 +82,24 @@ library LibProposing {
             revert L1_TOO_MANY_BLOCKS();
         }
 
+        bytes32 blobOrTxListHash;
```
dantaik, line 85: I prefer to use `blobHash` and you can add the same comment for it:  `//or txListHash (if Blob not yet supported)`

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -93,12 +113,14 @@ library LibProposing {
...
                 // number as the L2 mixHash.
                 difficulty: bytes32(block.prevrandao * b.numBlocks),
+                blobHash: blobOrTxListHash,
                 extraData: extraData,
                 id: b.numBlocks,
                 timestamp: uint64(block.timestamp),
                 l1Height: uint64(block.number - 1),
                 gasLimit: config.blockMaxGasLimit,
                 coinbase: msg.sender,
+                usingBlob: blobUsed,
```
dantaik, line 123: use the same name? "blobused"?

packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -121,4 +121,34 @@ library LibUtils {
...
+
+        uint256[7] memory inputs;
+        inputs[0] = uint256(meta.l1Hash);
+        inputs[1] = uint256(meta.difficulty);
+        inputs[2] = uint256(meta.blobHash);
+        inputs[3] = uint256(meta.extraData);
+        inputs[4] = (uint256(meta.id)) | (uint256(meta.timestamp) << 64)
+            | (uint256(meta.l1Height) << 128) | (uint256(meta.gasLimit) << 192);
+        inputs[5] =
+            uint256(uint160(meta.coinbase)) | (uint256(uUsingBlob) << 160);
```
dantaik, line 147: you can simply use `(meta.usingBlob ? 1: 0)`

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -187,6 +191,9 @@ library LibProving {
         {
             address verifier = resolver.resolve(tier.verifierName, true);
 
+            // Auto-filled from meta value, so that no interface changes (no
+            // need to modify verifyProof)
+            evidence.usingBlob = meta.usingBlob;
```
dantaik, line 196: so the value of evidence.usingBlob is simply overwritten? Somehow I don't like the placeholder idea. Why not just pass in usingBlob directly to the verifier?

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -104,9 +106,11 @@ library TaikoData {
         bytes32 metaHash;
         bytes32 parentHash;
         bytes32 blockHash;
+        bytes32 blobHash;
```
dantaik, line 109: I think we can simply pass both "Metadata" and "Evicence" to the verifier, and remove duplicate fields, such as `blobHash` and `usingBlob` from Evidence, we can also delete `metaHash` from Evidence.

packages/protocol/contracts/L1/libs/LibProving.sol
adaki2004: Wanted to keep the verifyProof() as clean as possible:

```
    function verifyProof(
        uint64 blockId,
        address prover,
        bool isContesting,
        TaikoData.BlockEvidence memory evidence
    )
```

 Since others do not use the `blobHash` or `blobUsed` (like SGX). But if you prefer, i can add it back.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -110,13 +110,15 @@ contract TaikoL1 is
         external
         nonReentrant
     {
+        (
+            TaikoData.BlockEvidence memory evidence,
+            TaikoData.BlockMetadata memory metaData
```
dantaik, line 115: can we simply call it `meta` for consistence?

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -110,13 +110,15 @@ contract TaikoL1 is
...
+            input, (TaikoData.BlockEvidence, TaikoData.BlockMetadata)
+        );
         TaikoData.Config memory config = getConfig();
         uint8 maxBlocksToVerify = LibProving.proveBlock(
-            state,
-            config,
-            AddressResolver(this),
-            blockId,
-            abi.decode(input, (TaikoData.BlockEvidence))
+            state, config, AddressResolver(this), blockId, evidence, metaData
```
dantaik, line 121: metadata already has an `id` field, so I think we can stop passing `blockId` to LibProving, but we probably want to check `meta.id == blockId`.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -185,19 +189,21 @@ library LibProving {
         // It's obvious that proof verification is entirely decoupled from
         // Taiko's core protocol.
         {
-            address verifier = resolver.resolve(tier.verifierName, true);
+            IVerifier.VerifierInput memory verifierInput;
```
dantaik, line 192: Can we do 

```
= Input({
a: a,
b: b
})
```
so if you miss any field, the compiler will complain.

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -75,7 +76,8 @@ contract GuardianProver is EssentialContract {
     /// @dev Called by guardians to approve a guardian proof
     function approveEvidence(
         uint64 blockId,
```
dantaik, line 78: blockId can be removed

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -44,49 +44,40 @@ contract PseZkVerifier is EssentialContract, IVerifier {
...
+                prover: input.prover,
+                evidence: evidence,
+                pointValue: pf.pointValue
+            });
+
             Lib4844.evaluatePoint({
-                blobHash: blobHash,
-                x: uint256(x) % Lib4844.BLS_MODULUS,
+                blobHash: input.blobHash,
+                x: uint256(instance) % Lib4844.BLS_MODULUS,
```
dantaik, line 70: please use  `bytes32 x = keccak256(abi.encodePacked(blobHash, pf.txListHash));` as the x value.

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -11,6 +11,7 @@ import { Proxied } from "../../common/Proxied.sol";
 
 import { LibTiers } from "../tiers/ITierProvider.sol";
 import { TaikoData } from "../TaikoData.sol";
+import { TaikoL1 } from "../TaikoL1.sol";
```
dantaik, line 14: remove this.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -137,8 +127,6 @@ contract PseZkVerifier is EssentialContract, IVerifier {
                 evidence.signalRoot,
                 evidence.graffiti,
                 prover,
-                blobHash,
```
dantaik, line 140: Dani, I think we still need `blobHash` and `txListHash` to be hashed into instnce.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
adaki2004: x is a uint256

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
adaki2004: I think not necessary, because `metaHash` is there (and metaHash contains them (in blobHash variable hashed into it)).

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -100,23 +100,17 @@ contract TaikoL1 is
...
-    /// select the right implementation version.
     /// @param input An abi-encoded {TaikoData.BlockEvidence} object.
-    function proveBlock(
-        uint64 blockId,
-        bytes calldata input
-    )
-        external
-        nonReentrant
-    {
+    function proveBlock(bytes calldata input) external nonReentrant {
```
dantaik, line 104: I'm afraid for this function (not the one in LibProving) , we still need the blockId to route to different impls later if we hard fork.

I'll add it back.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -32,7 +33,8 @@ library LibProposing {
         uint256 proverFee,
         uint16 minTier,
         TaikoData.BlockMetadata meta,
-        bool usingBlob
+        bytes32 blobHash,
```
dantaik, line 36: meta already has 'blobHash` and 'blobUsed`. You can remove them from the event.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -129,19 +131,14 @@ library TaikoData {
...
-        bytes32 blobHash; // slot 2
-        address assignedProver; // slot 3
+        address assignedProver; // slot 2
         uint96 livenessBond;
-        uint64 blockId; // slot 4
+        uint64 blockId; // slot 3
         uint64 proposedAt;
         uint32 nextTransitionId;
         uint32 verifiedTransitionId;
         uint16 minTier;
```
dantaik, line 140: minTier can also be moved MetaData as it is not used during verification. If you do so, you can also remove `minTier` from Event BlockProposed together with `blobHash` and `blobUsed`

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -67,17 +67,20 @@ library LibProving {
...
-        uint64 slot = blockId % config.blockRingBufferSize;
+        uint64 slot = meta.id % config.blockRingBufferSize;
         TaikoData.Block storage blk = state.blocks[slot];
 
         // Check the integrity of the block data. It's worth noting that in
         // theory, this check may be skipped, but it's included for added
         // caution.
-        if (blk.blockId != blockId || blk.metaHash != evidence.metaHash) {
+        if (
+            blk.blockId != meta.id || blk.metaHash != evidence.metaHash
```
dantaik, line 81: Evidence and Metadata are both passed to proveBlock, therefore, you should remove `metahash` from `Evidence`.

packages/protocol/contracts/L1/verifiers/IVerifier.sol
```diff
@@ -11,13 +11,18 @@ import { TaikoData } from "../TaikoData.sol";
 /// @title IVerifier Interface
 /// @notice Defines the function that handles proof verification.
 interface IVerifier {
+    struct VerifierInput {
+        uint64 blockId;
+        address prover;
+        bool isContesting;
+        bool blobUsed;
+        bytes32 blobHash;
```
dantaik, line 19: here in Input, you need 'metaHash'

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -44,49 +44,42 @@ contract PseZkVerifier is EssentialContract, IVerifier {
...
 
         bytes32 instance;
-
-        if (usingBlob) {
+        if (input.blobUsed) {
             PointProof memory pf = abi.decode(proof.pointProof, (PointProof));
-            bytes32 x = keccak256(abi.encodePacked(blobHash, pf.txListHash));
+
+            instance = calcInstance({
+                prover: input.prover,
```
dantaik, line 63: yes, instance must have `metaHash` (no longer part of evidence),  `txListHash` to be hashed in.

dantaik: Suggest to use evidence as the first param

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
dantaik: You still need to hash-in txListHash besides metahash.

adaki2004: LiProposing.sol line 84: 
`bytes32 blobHash; //or txListHash (if Blob not yet supported)`
So `txListHash` is actually hashed into the metaHash - under the name `blobHash`.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
adaki2004: > yes, instance must have `metaHash` (no longer part of evidence), `txListHash` to be hashed in.

Please see my thoughts on why not txListHash has to be hashed in.
https://github.com/taikoxyz/taiko-mono/pull/15031#discussion_r1372232411

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
adaki2004: @Brechtpd am i wrong maybe here ? 🤔

Brechtpd: `pf.txListHash` still needs to be put into the instance because it's used to generate the random point, and so in the circuit we have to verify if it is indeed the correct value by hashing the data in the circuit.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
dantaik: I think it is still needed, if not, how can the circuits check that the right value is used for calculating the x: 

```
 Lib4844.evaluatePoint({
                blobHash: input.blobHash,
                x: uint256(
                    keccak256(abi.encodePacked(input.blobHash, pf.txListHash))
                    ) % Lib4844.BLS_MODULUS,
                y: pf.pointValue,
                commitment: pf.pointCommitment,
                proof: pf.pointProof
            });
```

PR #15041, opened 2023-10-26
refactor(protocol): additional changes

dantaik:


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -252,6 +258,7 @@ library LibProving {
                     tran.validityBond / 4 + tran.contestBond
                 );
                 tran.contester = address(0);
+                tran.validityBond = 0;
```
dantaik, line 261: Dani, do you think this is necessary?

adaki2004: In my opinion, it is not necessary but arguable.
Given the nature of the proofs (ZK / SGX) there shall be only 1 verifiable path / fork. 
In case the ZK proof (or SGX) allows more proofs to be verified, creating an invalid (alternative) transition, guardian overwrites it - but I just guess here, that the original prover who submitted a wrong proof was unintentional, so i'd return the validityBond, but it is your call.

dantaik: I don't think this has nothing to do with ZK or SGX. This is the guardian prover provers a block, then give back the liveness bond.

adaki2004: Yes, i'm aware. I think it has to do with ZK or SGX. Because when Guardian 'comes into action' ? When the previous tiers fail (for whatever reason).

PR #15046, opened 2023-10-26
refactor(protocol): split Evidence to Transition + TierProof

dantaik:
As you see, this is logically clearer.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -72,6 +72,11 @@ library TaikoData {
         uint256 fee;
     }
 
+    struct TierProof {
+        uint16 tier;
+        bytes proof;
+    }
+
     struct ProverAssignment {
         address prover;
         address feeToken;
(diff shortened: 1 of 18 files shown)


adaki2004 (approved):

davidtaikocha (approved):

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -80,10 +85,12 @@ library TaikoData {
         bytes signature;
     }
 
+    struct BlockParams {
```
dantaik, line 88: Going forward we can put more parameter inside this struct without changing the proposeBlock's ABI

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -92,29 +99,27 @@ library TaikoData {
...
+        bytes32 depositsHash; // slot 5
+        address coinbase; // L2 coinbase, // slot 6
         uint64 id;
         uint32 gasLimit;
-        uint64 timestamp; // slot 6
+        uint64 timestamp; // slot 7
         uint64 l1Height;
         uint16 minTier;
         bool blobUsed;
-        TaikoData.EthDeposit[] depositsProcessed; // slot 7
```
dantaik, line 102: Removal of deposits make the metadata smaller for proveBlock() to use, and it will also simplified the way we hash the metadata.

davidtaikocha (approved):

PR #15049, opened 2023-10-27
fix(protocol): fix the `message.to == etherVault` bug

dantaik:
Reported by Victor from QuillAudits: if message.to == etherVault bug, then message can call releaseEther to drain all Ether from the audit.

I also added a modifier `onlyWhenNamed` so make sure if we change the "ether_vault" mapping to a different address **mistakenly** (we will probably register more address with different names, and it is possible that we accidentally change "ether_vault"), the the original EtherVault contract will no longer trust (or give access to) the Bridge contract. Otherwise, if the bridge still have access to the old EtherVault contract, message.to can be set to the old EtherVault contract to withdraw Ether.

- [ ] @adaki2004 can you verify the bug is real or not?
- [ ] @adaki2004 could you remove authorize() from EtherVault and use `onlyFromNamed("bridge")` instead?

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -148,9 +148,7 @@ contract Bridge is EssentialContract, IBridge {
         isMessageRecalled[msgHash] = true;
 
         // Release necessary Ether from EtherVault.
-        EtherVault(resolve("ether_vault", false)).releaseEther(
-            address(this), message.value
-        );
+        EtherVault(resolve("ether_vault", false)).releaseEther(message.value);
 
         // Execute the recall logic based on the contract's support for the
         // IRecallableSender interface
@@ -197,9 +195,7 @@ contract Bridge is EssentialContract, IBridge {
         if (!received) revert B_NOT_RECEIVED();
 
         address payable ethVault = resolve("ether_vault", false);
-        EtherVault(ethVault).releaseEther(
-            address(this), message.value + message.fee
-        );
+        EtherVault(ethVault).releaseEther(message.value + message.fee);
 
         Status status;
         uint256 refundAmount;
@@ -272,7 +268,7 @@ contract Bridge is EssentialContract, IBridge {
         }
 
         address payable ethVault = resolve("ether_vault", false);
-        EtherVault(ethVault).releaseEther(address(this), message.value);
+        EtherVault(ethVault).releaseEther(message.value);
 
         // Attempt to invoke the messageCall.
         if (_invokeMessageCall(message, msgHash, gasleft())) {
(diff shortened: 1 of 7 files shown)


dantaik:
@adaki2004 could you remove authorize() from Bridge and use `onlyFromNamed("bridge")` instead?

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -100,11 +100,14 @@ contract Bridge is EssentialContract, IBridge {
         if (message.to == address(0)) revert B_INVALID_TO();
         if (message.to == destBridge) revert B_INVALID_TO();
 
+        address etherValut = resolve("ether_vault", false);
```
adaki2004, line 103:
```suggestion
        address etherVault = resolve("ether_vault", false);
```

adaki2004:
> Reported by Victor from QuillAudits: if message.to == etherVault bug, then message can call releaseEther to drain all Ether from the audit.
>     * [ ]  @adaki2004 can you verify the bug is real or not?

I analyzed the code and i see no bug here - maybe a "mistake" like: why would anyone want to send ether/erc20 to our EtherVault ? Is there a way to get a more detailed description / workflow of the problem or a contact to Victor ?

How i see:
EtherVault has a receive() fallback, so the `_invokeMessageCall(message, msgHash, gasLimit)` could succeed.

If it succeeds, how anyone would get over this check below in the `recallMessage` ? 
```
bool received = _proveSignalReceived(
            _signalForFailedMessage(msgHash), message.destChainId, proof
        );
```

I do not see here a bug (maybe we can restrict the sending to EtherVault, the before mentioned reasons, to be stupid-proof) but i so far i could not figure out how it could drain all ether. 🤔

dantaik:
@adaki2004 a design decision to be made here: if EtherVault is not shared by more than one Bridge, maybe we simply take custody of Ether in the Bridge, I think that's even simpler. 

If we do need to share the same Ether Vault among multiple Bridges, then we still need the authorize() function. Now the question is: **do we really want to share the EtherVault among multiple Bridges/Rollups, what can such a shared EtherVault offer**

adaki2004:
> @adaki2004 a design decision to be made here: if EtherVault is not shared by more than one Bridge, maybe we simply take custody of Ether in the Bridge, I think that's even simpler.
> 
> If we do need to share the same Ether Vault among multiple Bridges, then we still need the authorize() function. Now the question is: **do we really want to share the EtherVault among multiple Bridges/Rollups, what can such a shared EtherVault offer**

I let it answer @Brechtpd

But just to highlight this design convo is a bit unrelated to the initial finding: a 'draining issue if `EtherVault` is the `message.to` reported by Quill'. I still could not figure it out how is it possible, could you please ask Victor to comment in this PR ?

PR #15051, opened 2023-10-27
fix(protocol): fix 2 bridge bugs reported by Victor

dantaik:
Reported by Victor from QuillAudits:

1.  if message.to == etherVault bug, then message can call releaseEther to drain all Ether from the audit.

2. currently attacker can steal all tokens in the ERC20Vault through the same bug I mentioned, calling the onMessageRecalled which also has a constraint that shouldn't be called unless failed, currently through this bug that is not handled fully attacker is able to bypass this regardless and steal all tokens

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -99,14 +99,15 @@ contract Bridge is EssentialContract, IBridge {
 
         if (message.to == address(0)) revert B_INVALID_TO();
         if (message.to == destBridge) revert B_INVALID_TO();
+        if (message.to == resolve(message.destChainId, "ether_vault", true)) {
+            revert B_INVALID_TO();
+        }
 
-        address etherVault = resolve("ether_vault", false);
-        if (message.to == etherVault) revert B_INVALID_TO();
         // Ensure the sent value matches the expected amount.
         uint256 expectedAmount = message.value + message.fee;
         if (expectedAmount != msg.value) revert B_INVALID_VALUE();
 
-        etherVault.sendEther(expectedAmount);
+        resolve("ether_vault", false).sendEther(expectedAmount);
 
         _message = message;
         // Configure message details and send signal to indicate message
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -99,8 +99,8 @@ contract Bridge is EssentialContract, IBridge {
 
         if (
             message.to == address(0) || message.to == destBridge
-                || message.to == resolve(message.destChainId, "taiko", true)
                 || message.to == resolve(message.destChainId, "ether_vault", true)
+                || message.to == resolve(message.destChainId, "taiko", true)
         ) {
             revert B_INVALID_TO();
         }
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -199,7 +199,7 @@ contract Bridge is EssentialContract, IBridge {
         // Process message differently based on the target address
         if (
             message.to == address(0) || message.to == address(this)
-                || message.to == ethVault || message.to == resolve("taiko", true)
+                || message.to == ethVault || message.to == resolve("taiko", false)
         ) {
             // Handle special addresses that don't require actual invocation but
             // mark message as DONE
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -97,9 +97,11 @@ contract Bridge is EssentialContract, IBridge {
             revert B_INVALID_CHAINID();
         }
 
-        if (message.to == address(0)) revert B_INVALID_TO();
-        if (message.to == destBridge) revert B_INVALID_TO();
-        if (message.to == resolve(message.destChainId, "ether_vault", true)) {
+        if (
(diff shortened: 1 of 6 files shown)


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -202,7 +207,10 @@ contract Bridge is EssentialContract, IBridge {
         uint256 refundAmount;
 
         // Process message differently based on the target address
-        if (message.to == address(this) || message.to == address(0)) {
+        if (
+            message.to == address(0) || message.to == address(this)
+                || message.to == ethVault || message.to == resolve("taiko", true)
```
Brechtpd, line 212: Is this only added here for messages that were already sent before adding the new checks? Just want to make sure I don't miss anything.

dantaik: Without this, we'd have to rely on all source AddressManagers to be configured correctly; with this, the Bridge is considered more secure.

Brechtpd (approved):

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -97,8 +97,13 @@ contract Bridge is EssentialContract, IBridge {
             revert B_INVALID_CHAINID();
         }
 
-        if (message.to == address(0)) revert B_INVALID_TO();
-        if (message.to == destBridge) revert B_INVALID_TO();
+        if (
+            message.to == address(0) || message.to == destBridge
```
dantaik, line 101: @Brechtpd Here, maybe I can simply remove this if-statement.

Brechtpd: Ah yeah, that makes sense!

Brechtpd (approved):

davidtaikocha (approved):
LGTM, after merging this into `4844`, lets open another cherry-pick PR targeting to `contestable-zkrollup`?

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -154,9 +149,23 @@ contract Bridge is EssentialContract, IBridge {
         bool support =
             message.from.supportsInterface(type(IRecallableSender).interfaceId);
         if (support) {
+            _ctx = Context({
+                msgHash: msgHash,
+                from: address(this),
+                srcChainId: message.srcChainId
+            });
```
Brechtpd, line 156: I guess for `onMessageRecalled` the context could simply be passed into the function because of the fixed function call, but makes sense to use the same mechanism as the normal call.

Brechtpd (approved):

packages/protocol/contracts/bridge/Bridge.sol
dantaik: Not sure I understand you correctly. I think vaults must call back to the Bridge through `context()` to make sure the onMessageCalled is not only called by bridge, but only called within the `recallMessage` function.

Brechtpd: Isn't that special case handled by having the `assert(message.from != address(this));` in the normal `_invokeMessageCall`? So the context from the bridge can only contain the bridge address itself for `recallMessage`?

Not suggesting any changes, but still interesting to know to make sure I fully understand things. :)

Brechtpd: Oh never mind, I see know why that can't work!

dantaik: In `sendMessage`, we have `_message.from = msg.sender;` this means message.from should never be "address(this)" -- so the assert will never be triggered.

 But a user can still specify `message.to` to a vault, and the `message.data= abi.encode(Vault.onMessageRecalled.selector, ...)`.  Since the vault trusts the Bridge -- its `recallMessage` function has a `onlyFromNamed("bridge")` modifier, the `onMessageRecalled` can be executed if the user process the message on the destination.

So any bridge-enabled app, all the current token vaults included, must verify that certain functions are not only callable by the bridge, but only by the bridge's processMessage or recallMessage function, not both.

dantaik: The security is so important, do you think we should set up a call to discuss about it?

dantaik:
> LGTM, after merging this into `4844`, lets open another cherry-pick PR targeting to `contestable-zkrollup`?

David, there are actually some outstanding issues with multi-hop bridging, and its not really easy to resolve soon.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: Also, I realized some design assmptions are not valid for enabling the sharing of certain components, such as EtherVault, token vaults, signal service, the bridge, among multiple L2s. To truly make L2->L2 transfer of asset possible, there may be quite some changes... Still working on it.

Brechtpd: > The security is so important, do you think we should set up a call to discuss about it?

Looking at these issues, I realized I'm really not familiar with how the bridge works, so for me I think it would be best to first spend some more time learning the latest code.

If there's more changes coming, maybe we can do a call after everybody had time to look into those? Or we need a bridge session in Turkey. :)

dantaik: Ok, there will be more changes, mostly to:

1) simplify the vault codebase
2) enable L2-L2 bridging (still not working)

dantaik: Let me merge this one first.

PR #15052, opened 2023-10-27
fix(protocol): Handle 1559 vars together

adaki2004:
As per convo here: https://github.com/taikoxyz/taiko-mono/pull/15025#discussion_r1372647928 - we need to handle (and re-set) gasExcess together with the target (and quotient).

### DIFF
--- a/packages/protocol/contracts/L1/TaikoL1.sol
+++ b/packages/protocol/contracts/L1/TaikoL1.sol
@@ -258,7 +258,7 @@ contract TaikoL1 is
     /// @return Config struct containing configuration parameters.
     function getConfig()
         public
-        pure
+        view
         virtual
         returns (TaikoData.Config memory)
     {
@@ -293,7 +293,7 @@ contract TaikoL1 is
         });
     }
 
-    function isConfigValid() public pure returns (bool) {
+    function isConfigValid() public view returns (bool) {
         return LibVerifying.isConfigValid(getConfig());
     }
 }
--- a/packages/protocol/contracts/L2/TaikoL2.sol
+++ b/packages/protocol/contracts/L2/TaikoL2.sol
@@ -25,8 +25,9 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
     using LibMath for uint256;
 
     struct Config {
+        uint128 gasExcess;
         uint64 gasTargetPerL1Block;
-        uint256 basefeeAdjustmentQuotient;
+        uint8 basefeeAdjustmentQuotient;
     }
 
     // Mapping from L2 block numbers to their block hashes.
@@ -36,10 +37,8 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
 
     // A hash to check the integrity of public inputs.
     bytes32 public publicInputHash; // slot 3
-    // Will init it with value 1 in genesis generation script to simply enable
-    // the L2 1559.
-    uint128 public gasExcess; // slot 4
-    uint64 public latestSyncedL1Height; // slot 5
+    uint64 public latestSyncedL1Height; // slot 4
+    Config public baseFeeConfig; // slot 5
 
     uint256[145] private __gap;
 
@@ -56,7 +55,9 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
     /// @param _addressManager Address of the {AddressManager} contract.
     function init(
         address _addressManager,
-        uint128 _gasExcess
+        uint128 _gasExcess,
+        uint64 _gasTargetPerL1Block,
+        uint8 _basefeeAdjustmentQuotient
     )
         external
         initializer
@@ -73,7 +74,10 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
             l2Hashes[parentHeight] = blockhash(parentHeight);
         }
 
-        gasExcess = _gasExcess;
+        baseFeeConfig.gasExcess = _gasExcess;
+        baseFeeConfig.gasTargetPerL1Block = _gasTargetPerL1Block;
+        baseFeeConfig.basefeeAdjustmentQuotient = _basefeeAdjustmentQuotient;
+
         (publicInputHash,) = _calcPublicInputHash(block.number);
     }
 
@@ -115,7 +119,8 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
 
         // Verify the base fee per gas is correct
         uint256 basefee;
-        (basefee, gasExcess) = _calc1559BaseFee(config, l1Height, parentGasUsed);
+        (basefee, config.gasExcess) =
+            _calc1559BaseFee(config, l1Height, parentGasUsed);
         if (!skipFeeCheck() && block.basefee != basefee) {
             revert L2_BASEFEE_MISMATCH();
         }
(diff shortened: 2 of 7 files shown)


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -174,13 +185,13 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
-    function getConfig() public pure virtual returns (Config memory config) {
-        // 4x Ethereum gas target, if we assume most of the time, L2 block time
-        // is 3s, and each block is full (gasUsed is 15_000_000), then its
-        // ~60_000_000, if the  network is congester than that, the base fee
-        // will increase.
-        config.gasTargetPerL1Block = 15 * 1e6 * 4;
-        config.basefeeAdjustmentQuotient = 8;
+    function getConfig() public view virtual returns (Config memory config) {
+        // // 4x Ethereum gas target, if we assume most of the time, L2 block
+        // time
```
dantaik, line 190: reformat this comment please

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -56,7 +55,9 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
     /// @param _addressManager Address of the {AddressManager} contract.
     function init(
         address _addressManager,
-        uint128 _gasExcess
+        uint128 _gasExcess,
```
dantaik, line 58: use Config here?

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -25,8 +25,9 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
     using LibMath for uint256;
 
     struct Config {
+        uint128 gasExcess;
```
dantaik, line 28: if we do not use the Config object, then we can align the storage to use one less slot -- sharing the same slot with latestSyncedL1Height by changing uint128 to uint96?

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -133,7 +138,13 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
         publicInputHash = publicInputHashNew;
         latestSyncedL1Height = l1Height;
 
-        emit Anchored(blockhash(parentId), gasExcess);
+        emit Anchored(blockhash(parentId), config.gasExcess);
+    }
+
+    /// @notice Sets EIP1559 related configurations
+    /// @param config The new config settings.
+    function setConfig(Config memory config) public virtual onlyOwner {
```
dantaik, line 146: we should check the config is valid, for example, value > 0 or something.

packages/protocol/contracts/L2/TaikoL2.sol
adaki2004: We use the config object as storage var.

adaki2004: We need it because gasExcess is a constantly changing variable. not a static config var.

dantaik: Yes, i mean, get ride of the Config object, and layout the storage as:

```solidity
uint96 gasExcess; // slot 2
uint64 gasTargetPerL1Block;
uint8 basefeeAdjustmentQuotient
uint64 public latestSyncedL1Height;
```

this will only take one slot

adaki2004: I'd not reduce the gasExcess size.
kép

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -115,7 +128,8 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
 
         // Verify the base fee per gas is correct
         uint256 basefee;
-        (basefee, gasExcess) = _calc1559BaseFee(config, l1Height, parentGasUsed);
+        (basefee, config.gasExcess) =
```
dantaik, line 131: this won't write to storage, you have to do`baseFeeConfig.gasExcess`

dantaik: I think you can get rid of  `Config memory config = getConfig();`

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -133,7 +147,23 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
+        emit Anchored(blockhash(parentId), config.gasExcess);
+    }
+
+    /// @notice Sets EIP1559 related configurations
+    /// @param _baseFeeConfig The new config settings.
+    function setConfig(Config memory _baseFeeConfig)
+        public
+        virtual
+        onlyOwner
+        validConfig(
```
dantaik, line 159: validConfig(config)?

packages/protocol/contracts/L2/TaikoL2.sol
adaki2004: I can get rid of that struct, sure.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -36,30 +37,41 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
 
     // A hash to check the integrity of public inputs.
     bytes32 public publicInputHash; // slot 3
-    // Will init it with value 1 in genesis generation script to simply enable
-    // the L2 1559.
-    uint128 public gasExcess; // slot 4
-    uint64 public latestSyncedL1Height; // slot 5
+    uint64 public latestSyncedL1Height; // slot 4
```
dantaik, line 40: @adaki2004 somehow I prefer this setter to be added to a subcontract of TaikoL2.sol and keep TaikoL2.sol as is. Do you think it's a better idea?

adaki2004: You mean this one:
`setConfig()` ?

TBH i dont know if it makes sense to create a contract just for 3 variables.. (?).

dantaik: I mean add a TaikoL2EIP1559Configurable.sol so inherit  TaikoL2 then add/override whatever functions necessary.

adaki2004: What are you doing here ? I totally do not get it why we need to work like this.
https://github.com/taikoxyz/taiko-mono/pull/15053/files

dantaik: I'm not addressing the same thing, you are adding a new feature, I'm just optimizing the storage size for the old contract without a new feature.

adaki2004: > TaikoL2EIP1559Configurable.sol

Done.

adaki2004: > I'm not addressing the same thing, you are adding a new feature, I'm just optimizing the storage size for the old contract without a new feature.

Took over your size reductions here.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -258,7 +258,7 @@ contract TaikoL1 is
     /// @return Config struct containing configuration parameters.
     function getConfig()
         public
-        pure
+        view
```
dantaik, line 261: this is necessary for sub-contracts to override by reading from storage

dantaik (approved):

PR #15053, opened 2023-10-27
refactor(protocol): change gas eip1559 types

dantaik:
The previous datatypes are too large


adaki2004:
We are about to address them here:
https://github.com/taikoxyz/taiko-mono/pull/15052
And it lacks the possibility to set the 3 variables (excess, target, quotient) **all together** - which is crucial.

adaki2004:
Delete this please, i'll handle it in A6 configs!

adaki2004:
Addressed all size reductions here: https://github.com/taikoxyz/taiko-mono/pull/15052

PR #15057, opened 2023-10-27
feat(protocol): Use authorize instead of onlyFrom

adaki2004:
Use authorize with vaults to be able to share common vaults across different L2 instances.

### DIFF
--- a/packages/protocol/contracts/bridge/EtherVault.sol
+++ b/packages/protocol/contracts/bridge/EtherVault.sol
@@ -8,30 +8,21 @@ pragma solidity ^0.8.20;
 
 import { EssentialContract } from "../common/EssentialContract.sol";
 import { LibAddress } from "../libs/LibAddress.sol";
+import { AuthorizationBase } from "../common/AuthorizationBase.sol";
 import { Proxied } from "../common/Proxied.sol";
 
 /// @title EtherVault
 /// @notice This contract is initialized with 2^128 Ether and allows authorized
 /// addresses to release Ether.
 /// @dev Only the contract owner can authorize or deauthorize addresses.
-contract EtherVault is EssentialContract {
+contract EtherVault is AuthorizationBase, EssentialContract {
     using LibAddress for address;
 
-    mapping(address addr => bool authorized) public isAuthorized;
-    uint256[49] private __gap;
+    uint256[50] private __gap;
 
-    event Authorized(address indexed addr, bool authorized);
     event EtherReleased(address indexed to, uint256 amount);
 
-    error VAULT_PERMISSION_DENIED();
     error VAULT_INVALID_RECIPIENT();
-    error VAULT_INVALID_PARAMS();
-
-    modifier onlyAuthorized() {
-        // Ensure the caller is authorized to perform the action
-        if (!isAuthorized[msg.sender]) revert VAULT_PERMISSION_DENIED();
-        _;
-    }
 
     receive() external payable { }
 
@@ -78,11 +69,7 @@ contract EtherVault is EssentialContract {
     /// @param addr Address to set the authorized status of.
     /// @param authorized Authorized status to set.
     function authorize(address addr, bool authorized) public onlyOwner {
-        if (addr == address(0)) revert VAULT_INVALID_PARAMS();
-        if (isAuthorized[addr] == authorized) revert VAULT_INVALID_PARAMS();
-
-        isAuthorized[addr] = authorized;
-        emit Authorized(addr, authorized);
+        _authorize(addr, authorized);
     }
 }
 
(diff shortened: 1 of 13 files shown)


packages/protocol/contracts/common/AuthorizationBase.sol
```diff
@@ -0,0 +1,37 @@
...
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+
+pragma solidity ^0.8.20;
+
+import { EssentialContract } from "../common/EssentialContract.sol";
+
+/// @title AuthorizationBase
+/// @notice Every contract which needs to have authorization (EtherVault, ERCXXVault)
+abstract contract AuthorizationBase {
```
dantaik, line 13: is EssentialContract

packages/protocol/contracts/common/AuthorizationBase.sol
```diff
@@ -0,0 +1,37 @@
...
+    error VAULT_PERMISSION_DENIED();
+    error VAULT_INVALID_PARAMS();
+
+    modifier onlyAuthorized() {
+        // Ensure the caller is authorized to perform the action
+        if (!isAuthorized[msg.sender]) revert VAULT_PERMISSION_DENIED();
+        _;
+    }
+
+    function _authorize(address addr, bool authorized) internal {
```
dantaik, line 30: _authorize => authorize external

packages/protocol/contracts/common/AuthorizationBase.sol
adaki2004: Tried the idea, but it would create unambigous parent dependencies (EtherVault vs. ERCXXXVault)

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -372,6 +374,14 @@ contract ERC20Vault is
             ctokenDecimal: ctoken.decimals
         });
     }
+
+    /// @notice Sets the authorized status of an address, only the owner can
+    /// call this function.
```
dantaik, line 379: Do we have to write this in every sub-contract?

packages/protocol/contracts/common/AuthorizationBase.sol
adaki2004: Ot making baseNFTVault ingeriting from that one can do the job. ok, will try later.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
adaki2004: Can rearrange these if i make BaseNFTVault inheriting from the AuthorizationBase. Will do later.

packages/protocol/contracts/common/AuthorizationBase.sol
dantaik: > Or making baseNFTVault ingeriting from that one can do the job. ok, will try later.

If possile, this is the preferred approach.

packages/protocol/contracts/bridge/EtherVault.sol
```diff
@@ -8,30 +8,21 @@ pragma solidity ^0.8.20;
...
 import { LibAddress } from "../libs/LibAddress.sol";
+import { AuthorizationBase } from "../common/AuthorizationBase.sol";
 import { Proxied } from "../common/Proxied.sol";
 
 /// @title EtherVault
 /// @notice This contract is initialized with 2^128 Ether and allows authorized
 /// addresses to release Ether.
 /// @dev Only the contract owner can authorize or deauthorize addresses.
-contract EtherVault is EssentialContract {
+contract EtherVault is AuthorizationBase, EssentialContract {
```
dantaik, line 18: AuthorizationBase => AuthorizableContract? This will match better with how we name EssentialContract

dantaik:
@adaki2004 there is one my issue bridge fixed, could you base your PR on top of this branch: bridgable_app

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -121,12 +121,6 @@ contract ERC20Vault is
         _;
     }
 
-    /// @notice Initializes the contract with the address manager.
-    /// @param addressManager Address manager contract address.
-    function init(address addressManager) external initializer {
-        EssentialContract._init(addressManager);
```
davidtaikocha, line 127: Shall we still keep this `init` method and call `AuthorizableContract.init(addressManager)` here?

dantaik:
*migrated into https://github.com/taikoxyz/taiko-mono/pull/15061**

PR #15061, opened 2023-10-27
refactor(protocol): refactor bridge/vault/signal for multi-chain deployment

dantaik:
The following major changes are included in this PR:

- refactored token vault contracts, now they have a common base to handle interaction with the Bridge contract.
- Bridge contract and token vaults no longer check the `to` address, it will handled on the destination bridge/vault without failure.
- Bridge allows `message.to` to be a TaikoL1 contract address.
- Merged EtherVault into Bridge to simplify deployment
- Added a AuthorizableContract contract.
- SignalService is now an AuthorizableContract contract.
- New contract owner must claim the ownership.
- Allow owner to pause proving/verifying blocks.
- Ensure proposer is EOA when use calldata for DA (otherwise it's hard for client to download the calldata)
- Ensure chain ID is a uint64 so it can be made part of the address of each pre-deployment contract.
- SignalService's proof need to have an crossChainSync address.

After this PR: multihop bridging shall be truly enabled. The relayer needs to be modified accordingly, @cyberhorsey :

All reviewers, please read:

Read this: https://github.com/taikoxyz/taiko-mono/blob/bridgable_app/packages/protocol/docs/multihop_bridging_deployment.md

If you have questions, please DM me in discord to schedule a zoom call ASAP!


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -202,7 +210,11 @@ contract Bridge is EssentialContract, IBridge {
         uint256 refundAmount;
 
         // Process message differently based on the target address
-        if (message.to == address(this) || message.to == address(0)) {
+        if (
+            message.to == address(0) || message.to == address(this)
+                || message.to == ethVault
+        ) {
+            // || message.to == resolve("taiko", false)
```
dantaik, line 217: I removed this line: let me know if this may be a problem?

The reason to remove this one is to make Bridge sharable by multiple L2s, so there will be multiple "taiko" addresses, associated with different chainIds.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -88,18 +86,14 @@ contract Bridge is EssentialContract, IBridge {
...
-            isDestChainEnabled(message.destChainId);
+        (bool destChainEnabled,) = isDestChainEnabled(message.destChainId);
 
         // Verify destination chain and to address.
         if (!destChainEnabled) revert B_INVALID_CHAINID();
         if (message.destChainId == block.chainid) {
             revert B_INVALID_CHAINID();
         }
 
-        if (message.to == address(0)) revert B_INVALID_TO();
```
dantaik, line 100: remove checking for the `to` address, they are considered normal cases and are handled properly by the Bridge on the destination chain.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -213,7 +212,7 @@ contract Bridge is EssentialContract, IBridge {
         // Process message differently based on the target address
         if (
             message.to == address(0) || message.to == address(this)
-                || message.to == ethVault || message.to == resolve("taiko", false)
```
dantaik, line 216: Bridge allows `message.to` to be a TaikoL1 contract address

davidtaikocha:
@dantaik workflow errors fixed in the latest commits

packages/protocol/contracts/common/EssentialContract.sol
```diff
@@ -63,7 +66,7 @@ abstract contract EssentialContract is OwnableUpgradeable, AddressResolver {
     /// @notice Initializes the contract with an address manager.
     /// @param _addressManager The address of the address manager.
     function _init(address _addressManager) internal virtual override {
-        OwnableUpgradeable.__Ownable_init_unchained();
+        Ownable2StepUpgradeable.__Ownable2Step_init();
```
adaki2004, line 69: See : https://github.com/taikoxyz/taiko-mono/pull/15061/files#r1376345113

packages/protocol/contracts/common/AddressManager.sol
```diff
@@ -31,52 +31,52 @@ interface IAddressManager {
...
+        uint256 indexed chainId,
         bytes32 indexed name,
         address newAddress,
         address oldAddress
     );
 
     /// @notice Initializes the owner for the upgradable contract.
     function init() external initializer {
-        OwnableUpgradeable.__Ownable_init();
+        Ownable2StepUpgradeable.__Ownable2Step_init();
```
adaki2004, line 48: See: https://github.com/taikoxyz/taiko-mono/pull/15061/files#r1376345113

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -51,15 +55,19 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
     function init(
-        address _addressManager,
+        address _signalService,
         uint64 _gasExcess
     )
         external
         initializer
     {
-        EssentialContract._init(_addressManager);
+        Ownable2StepUpgradeable.__Ownable2Step_init();
```
adaki2004, line 67: See: https://github.com/taikoxyz/taiko-mono/pull/15061/files#r1376345113

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -110,15 +109,26 @@ contract SignalService is EssentialContract, ISignalService {
...
+        bytes32 signalRoot = ICrossChainSync(p.crossChainSync).getSyncedSnippet(
+            p.height
+        ).signalRoot;
+
         if (signalRoot == 0) return false;
 
         for (uint256 i; i < p.hops.length; ++i) {
             Hop memory hop = p.hops[i];
+
+            bytes32 label = authorizedAddresses[hop.signalRootRelay];
```
cyberhorsey, line 126: what if we are doing L1 to L4, and `TaikoL2` as a predeployed proxy at the same address across multiple chains (L2, L3)? we are unable to authorize the same address for more than one chainID now?

cyberhorsey:
PR with updates for relayer: https://github.com/taikoxyz/taiko-mono/pull/15088
until this PR is deployed on internal devnet i'll be unable to test, but will do so then!

packages/protocol/contracts/signal/SignalService.sol
dantaik: This is indeed a very good feedback. @davidtaikocha can we make sure the pre-deployed contracts all have unique addresses per chain, for example, if we deploy TaikoL2 on a chain with id 12345, can we use 12345 to prefix the address of the deployment of TaikoL2?

davidtaikocha: I will update the deployment script to implement this feature.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -94,9 +95,15 @@ library LibProposing {
 
             if (blobHash == 0) revert L1_NO_BLOB_FOUND();
         } else {
+            // The proposer must be an Externally Owned Account (EOA) for
+            // calldata usage. This ensures that the transaction is not an
+            // internal one, making calldata retrieval more straightforward.
+            if (msg.sender != tx.origin) revert L1_PROPOSER_NOT_EOA();
```
adaki2004, line 101: I guess we want to eliminate this:
`EOA` ->` Contract A` -> `Contract B` (TaikoL1)

Wondering what happens if the `Contract A `would use `delegatecall()` to preserve the EOA context with rubbish (non findeable, or not easily decodeable) TX list.

```

taikoL1.delegatecall(
            abi.encodeWithSignature("proposeBlock(bytes, bytes)", params, made_up_txn_list)
        );
```

adaki2004: `TransparentProxyUpgradeable` is also using `delegateCall()`, and since `TaikoL1` is upgradeable, the proxy also uses the delegateCall() pattern. 
So others could easily do that too, no ?

adaki2004: @dantaik @davidtaikocha 
Maybe something like this in TaikoL1.sol ? 🤔  (But if this is a proxy, then it may complicate things, hmm)
https://ethereum.stackexchange.com/questions/128924/is-there-a-way-to-check-if-a-call-is-a-call-or-a-delegatecall

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -144,7 +162,7 @@ contract SignalService is EssentialContract, ISignalService {
     /// @return The unique storage slot of the signal which is
     /// created by encoding the sender address with the signal (message).
     function getSignalSlot(
-        uint256 chainId,
+        uint64 chainId,
```
adaki2004, line 165: @cyberhorsey be careful with this one (if you need to implement similar on relayer side.. if you rely on this call it's ok), because it modifies the hash ! (uint256 vs uint64)

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: > I guess we want to eliminate this: `EOA` ->` Contract A` -> `Contract B` (TaikoL1)
> 
> Wondering what happens if the `Contract A` would use `delegatecall()` to preserve the EOA context with rubbish (non findeable, or not easily decodeable) TX list.
> 
> ```
> 
> taikoL1.delegatecall(
>             abi.encodeWithSignature("proposeBlock(bytes, bytes)", params, made_up_txn_list)
>         );
> ```

using delegateCall in contractA will make the sender to be the contractA address and the origin to be the user, so the if-statement will return true to revert.

dantaik: > https://ethereum.stackexchange.com/questions/128924/is-there-a-way-to-check-if-a-call-is-a-call-or-a-delegatecall

this approach prevents our contract to be called as code by third party contracts. I'm not sure we need to do it...

dantaik: We should use this feature to prevent our code to be used if our code has a restrictive license, so other people cannot deploy a copy of our code but want to call our deployed contract using delegate call.

packages/protocol/contracts/signal/SignalService.sol
dantaik: Should we simply change the code here to use uint256?

packages/protocol/contracts/L1/libs/LibProposing.sol
adaki2004: > using delegateCall in contractA will make the sender to be the contractA address 

delegateCall() will preserve the context of msg.sender, msg.data, mag.value, so in this case EOA would be msg.sender 🤔  ?

dantaik: If you are interacting with contractA which delegatecall TaikoL1, then your transaction is not a Taiko transaction (although the code is the same), our client won't listen to contractA's events...

adaki2004: > If you are interacting with contractA which delegatecall TaikoL1, then your transaction is not a Taiko transaction (although the code is the same), our client won't listen to contractA's events...

Yeah, right ! Thankfully it seems false alarm then. ;)

Brechtpd: Ah more discussion about this here! I want to raise another concern (https://github.com/taikoxyz/taiko-mono/pull/15090#discussion_r1377821887):
> This may break integrating Taiko with for example PBS because that will likely require some external contract, unless we do something like allowing an extra callback to be specified in the propose function that can do some extra logic.

packages/protocol/contracts/signal/SignalService.sol
cyberhorsey: ty, I have already covered it in the relayer PR!

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: I responded there and here is a PR for checking the EOA differently https://github.com/taikoxyz/taiko-mono/pull/15095

adaki2004:
@dantaik please consider reverting this (EOA) commit: https://github.com/taikoxyz/taiko-mono/pull/15061/commits/c63159d4bd2c2e8470b91ce6e614cc00802e4854 (introduced with the PR https://github.com/taikoxyz/taiko-mono/pull/15095#event-10829502314)

`extcodesize` is 0 during constructor call so it could make our client life harder.
Link: https://consensys.github.io/smart-contract-best-practices/development-recommendations/solidity-specific/extcodesize-checks/

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -94,9 +95,15 @@ library LibProposing {
 
             if (blobHash == 0) revert L1_NO_BLOB_FOUND();
         } else {
+            // The proposer must be an Externally Owned Account (EOA) for
+            // calldata usage. This ensures that the transaction is not an
+            // internal one, making calldata retrieval more straightforward.
+            if (!msg.sender.isEOA()) revert L1_PROPOSER_NOT_EOA();
```
adaki2004, line 101: This one: https://github.com/taikoxyz/taiko-mono/pull/15061#issuecomment-1788574549

packages/protocol/contracts/L1/libs/LibTaikoToken.sol
```diff
@@ -43,7 +43,7 @@ library LibTaikoToken {
         AddressResolver resolver,
         uint256 amount
     )
-        internal
+        external
```
Brechtpd, line 46: Ideally there's a delay for withdrawing to protect the proposer from a prover that withdraws tokens before the propose, making the propose tx fail.

It is possible to work around this with additional conditions that would stop the propose tx to be valid in PBS, but feels like this should simply be a standard feature so proposers don't have to care about that. A delay of a couple of L1 blocks seems more than enough.

packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -105,20 +70,77 @@ library LibUtils {
...
+            blockId == 0 ? state.slotB.lastVerifiedBlockId : blockId;
+        uint64 slot = _blockId % config.blockRingBufferSize;
+
+        TaikoData.Block storage blk = state.blocks[slot];
+
+        if (blk.blockId != _blockId) revert L1_BLOCK_MISMATCH();
+        if (blk.verifiedTransitionId == 0) revert L1_TRANSITION_NOT_FOUND();
+
+        TaikoData.TransitionState storage transition =
+            state.transitions[slot][blk.verifiedTransitionId];
```
Brechtpd, line 100: All this code seems duplicated from `getVerifyingTransition`, seems like that function should simply be reused here.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -51,15 +55,19 @@ contract TaikoL2 is EssentialContract, TaikoL2Signer, ICrossChainSync {
...
+
+        if (_signalService == address(0)) revert L2_INVALID_PARAM();
+        signalService = _signalService;
```
Brechtpd, line 70: Below there's logic for both `if (block.number > 1)` and `if (block.number > 0) {`. Shouldn't we just force the function to be called in a single block number? That would make things more clear.

packages/protocol/contracts/bridge/IBridge.sol
```diff
@@ -8,18 +8,17 @@ pragma solidity ^0.8.20;
 
 /// @title IBridge
 /// @notice The bridge used in conjunction with the {ISignalService}.
-/// @dev Ether is held by Bridges on L1 and by the EtherVault on L2,
-/// not by token vaults.
+/// @dev Ether is held by Bridges on L1 and L2s.
```
Brechtpd, line 11: There's a couple of `(auto filled)` and `For ERC20 transfers` comments that I think are outdated.

packages/protocol/contracts/bridge/IBridge.sol
```diff
@@ -8,18 +8,17 @@ pragma solidity ^0.8.20;
...
         // Message ID.
         uint128 id;
         // Message sender address (auto filled).
         address from;
         // Source chain ID (auto filled).
-        uint256 srcChainId;
+        uint64 srcChainId;
         // Destination chain ID where the `to` address lives (auto filled).
-        uint256 destChainId;
+        uint64 destChainId;
```
Brechtpd, line 21: `user` seems a bit like a strange name for the use of this variable if I understand correctly. Maybe `processor` or something would be closer to the actual use?

packages/protocol/contracts/common/ICrossChainSync.sol
```diff
@@ -14,16 +14,24 @@ pragma solidity ^0.8.20;
...
+    /// @param syncedInBlock The ID of this chain's block where the sync
+    /// happened.
     /// @param blockHash The hash of the synced block.
     /// @param signalRoot The root hash representing cross-chain signals.
     event CrossChainSynced(
-        uint64 indexed srcHeight, bytes32 blockHash, bytes32 signalRoot
+        uint64 indexed remoteBlockId,
+        uint64 syncedInBlock,
+        bytes32 blockHash,
+        bytes32 signalRoot
```
Brechtpd, line 34: Nitpick: a bit of a strange order of these variables, `syncedInBlock` of the current chain is mixed between variable of the source chain. Also the there's a `remote` in front of `blockId`, but not in front of `blockHash` and `signalRoot` which are also `remote` if I understand correctly. So something like would be more clear I think:
```
uint64 syncedInBlock,
uint64 indexed blockId,
bytes32 blockHash,
bytes32 signalRoot
```

packages/protocol/contracts/tokenvault/BaseVault.sol
```diff
@@ -0,0 +1,68 @@
...
+    function name() public pure virtual returns (bytes32);
+
+    function checkProcessMessageContext()
+        internal
+        view
+        returns (IBridge.Context memory ctx)
+    {
+        if (msg.sender != resolve("bridge", false)) {
+            revert VAULT_PERMISSION_DENIED();
+        }
```
Brechtpd, line 49: Code could be shared with the same code in ` checkRecallMessageContext()` with a nice `onlyFromBridge` modifier I think.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -295,7 +260,7 @@ contract ERC20Vault is
             // If it's a canonical token
             ERC20Upgradeable t = ERC20Upgradeable(token);
             ctoken = CanonicalERC20({
-                chainId: block.chainid,
+                chainId: uint64(block.chainid),
```
Brechtpd, line 263: - ` _encodeDestinationCall` doesn't seem like a great name for this function because it does a lot more then just encode the call data!
- Is the `isBridgedToken[token]` check necessary if we can just do `ctoken = bridgedToCanonical[token];` and check if it's 0 or not (totally not sure about this, would need to check more).
- Below the balance change it checked when doing the `transferFrom`. That's of course great but could use some comments that describes why it's done like that! (`transferFrom` can subtract a fee from the amount etc...)

Brechtpd (commented):
Haven't checked everything in detail yet, but here are some comments from a first run through the code. Nothing big.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -101,6 +108,7 @@ contract TaikoL1 is
     )
         external
         nonReentrant
+        whenNotPaused
```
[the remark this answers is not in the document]
dantaik: The BlockMetadata, Transition, TireProof struct/definition may change, if we have blockId, then we can decode the `input` differently based on the value of blockId.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -128,27 +136,35 @@ contract TaikoL1 is
     /// @param maxBlocksToVerify Max number of blocks to verify.
     function verifyBlocks(uint64 maxBlocksToVerify) external nonReentrant {
         if (maxBlocksToVerify == 0) revert L1_INVALID_PARAM();
+        if (state.slotB.provingPaused) revert L1_PROVING_PAUSED();
```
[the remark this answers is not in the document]
dantaik: fixed

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -110,30 +98,103 @@ library LibProposing {
...
             });
         }
+
+        // Following the Merge, the L1 mixHash incorporates the
+        // prevrandao value from the beacon chain. Given the possibility
+        // of multiple Taiko blocks being proposed within a single
+        // Ethereum block, we must introduce a salt to this random
+        // number as the L2 mixHash.
+        unchecked {
+            meta.difficulty = meta.blobHash
```
davidtaikocha, line 125: Seems `meta.blobHash` is still uninitialized at this point, so this "salt" is not working?

dantaik: You are right, fixed. Please take another look.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -208,6 +285,7 @@ library LibProposing {
         return keccak256(
             abi.encode(
                 "PROVER_ASSIGNMENT",
+                taikoAddress,
```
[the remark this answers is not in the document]
Brechtpd: Should we hash `metaHash

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -110,30 +98,104 @@ library LibProposing {
...
+                uint256(params.txListByteOffset) + params.txListByteSize
+                    > MAX_BYTES_PER_BLOB
+            ) {
+                revert L1_TXLIST_OFFSET();
+            }
+
+            if (
+                params.txListByteSize == 0
+                    || params.txListByteSize > config.blockMaxTxListBytes
+            ) revert L1_TXLIST_SIZE();
```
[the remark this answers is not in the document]
dantaik: It is now _impossible_ to submit a block with 0 bytes calldata as DA, but it is possible to submit a blob with no data at all (if Ethereum allows it).

Did you mean that we should allow submit 0-byte DA for a block? I think maybe allowing 0-data should also be fine.

packages/protocol/contracts/L1/libs/LibProposing.sol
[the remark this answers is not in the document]
dantaik: In Assignment, there is a parameter called `maxBlockId`, this is the max L2 block Id that this current block being proposed can take. Do you think we should also add `maxProposedIn` as the max L1 block ID that this block must be proposed in? Adding it should be easy.

We can also add an `assignment.metaHash` there, but if this value is 0, we allow the block to have any metahash, if the value is not zero, then we check the two meta hashes match. What do you think?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -235,7 +313,7 @@ library LibProposing {
 
         // Hash the assignment with the blobHash, this hash will be signed by
         // the prover, therefore, we add a string as a prefix.
-        bytes32 hash = hashAssignment(assignment, blobHash);
+        bytes32 hash = hashAssignment(assignment, address(this), blobHash);
```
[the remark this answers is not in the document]
dantaik: Good point. This may also apply to other sendEther case.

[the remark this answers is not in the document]
dantaik: After a second thought: if the prover can change the gas used when receive Ether, then the proposer should be carefully and should probably not choose to use the prover. In this PR, https://github.com/taikoxyz/taiko-mono/pull/15200, I applied a max gas, but the value is set to 200K to prevent the worst case.

packages/protocol/contracts/L1/libs/LibProposing.sol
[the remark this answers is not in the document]
dantaik: Please take a look at this PR: https://github.com/taikoxyz/taiko-mono/pull/15201

packages/protocol/contracts/L1/libs/LibUtils.sol
dantaik: getVerifyingTransition and getVariables in LibUtils are no longer used, so I removed them.

packages/protocol/contracts/bridge/IBridge.sol
dantaik: remove them.

packages/protocol/contracts/bridge/IBridge.sol
dantaik: The 'user' is actually the msg.sender on the source chain to send a message, so maybe 'owner' is a better name? I made the change, let me know if you like it.

packages/protocol/contracts/common/ICrossChainSync.sol
dantaik: Please take a look at this PR: https://github.com/taikoxyz/taiko-mono/pull/15202

packages/protocol/contracts/bridge/IBridge.sol
dantaik: message.from is the processor on the source chain; message.to is the process on the dest chain.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
dantaik: 1. renamed `_encodeDestinationCall` to `_handleMessage`

dantaik: 3. Done. comment added.

dantaik: 2. I need sometime to check if isBridgedToken can be removed.

packages/protocol/contracts/L1/libs/LibTaikoToken.sol
dantaik: - [ ] This has not been done yet, how exactly do you like this to be implemented. I feel like this can be done later.

packages/protocol/contracts/bridge/IBridge.sol
```diff
@@ -8,31 +8,30 @@ pragma solidity ^0.8.20;
...
-        uint256 destChainId;
-        // User address of the bridged asset.
-        address user;
-        // Destination address.
+        // Source chain ID.
+        uint64 srcChainId;
+        // Destination chain ID where the `to` address lives.
+        uint64 destChainId;
+        // The owner of the message.
+        address owner;
```
adaki2004, line 23: owner is OK, but need to rename the GenerateGenesis.g.sol `line 190` and `215` accordingly.

packages/protocol/contracts/L2/TaikoL2.sol
dantaik: block number can only be 0 (real L2 genesi) or 1 (in tests). 
I optimized the code to make it easier to understand.

packages/protocol/contracts/L1/libs/LibProposing.sol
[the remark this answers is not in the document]
Brechtpd: Hmm I think I also just meant that allowing it it should be fine, bot no big deal if not.

packages/protocol/contracts/L1/libs/LibTaikoToken.sol
Brechtpd: I think something super simple should be okay:
- Prover requests a certain amount of tokens to be withdrawable, this is stored in a mapping and this simply overwrites any previously requested amount. This also sets the timestamp when the amount will be withdrawable
- The withdraw function now simply checks that the amount to withdraw was requested earlier and it's now possible to withdraw that amount.

Agreed that it can be done later, just necessary to make the proposer <-> prover completely trustless.

packages/protocol/contracts/bridge/IBridge.sol
Brechtpd: Oh yeah owner is a good name!

> The 'user' is actually the msg.sender on the source chain to send a message, so maybe 'owner' is a better name? I made the change, let me know if you like it.

It's not necessarily `msg.sender` when using the bridge directly right?

> message.from is the processor on the source chain; message.to is the process on the dest chain.

I see `message.from` is kind of the simulated `msg.sender` I think, with `message.user` having some special processing powers. But agreed owner is a better name!

packages/protocol/contracts/L1/libs/LibProposing.sol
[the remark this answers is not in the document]
dantaik: Oh, then maybe using an array of one byte should be fine.

packages/protocol/contracts/L1/libs/LibTaikoToken.sol
dantaik: https://github.com/taikoxyz/taiko-mono/issues/15207

Filed a feature request

PR #15065, opened 2023-10-28
feat(bridge-ui-v2): add amount input to 1155 tokens

KorbinianK:
Amount input for 1155 tokens:

- handle and validate the input
- optimize approval check
- introduce new wrapper util to get contracts by type / token type


packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
```diff
@@ -154,26 +154,42 @@
...
-    const { value } = event.target as HTMLInputElement;
-    if (
-      $selectedToken.type === TokenType.ERC721 ||
-      $selectedToken.type === TokenType.ERC1155 ||
-      !$selectedToken.decimals
-    ) {
-      $enteredAmount = BigInt(value);
+    if ($selectedToken.type === TokenType.ERC1155) {
+      // For ERC1155, no decimals are allowed
+      if (value.includes('.') || value.includes(',')) {
```
NguyenTuanCanh, line 168: `if (/[.,]/.test(value)) {`
I think this is better.

packages/bridge-ui-v2/src/components/Bridge/NFTBridgeSteps/ImportStep.svelte
```diff
@@ -203,7 +204,15 @@
       canProceed = false;
     }
   } else if (importMethod === 'scan' && $selectedNFTs && $selectedNFTs.length > 0 && $destinationChain && scanned) {
-    canProceed = true;
+    if ($selectedNFTs && $selectedNFTs[0].type === TokenType.ERC1155) {
```
NguyenTuanCanh, line 207:
```
 function validateManualImport() {
    if (detectedTokenType === TokenType.ERC721 || detectedTokenType === TokenType.ERC1155) {
      return (
        addressInputState === AddressInputState.VALID &&
        idInputState === IDInputState.VALID &&
        (detectedTokenType === TokenType.ERC1155 ? $enteredAmount > BigInt(0) : true) &&
        isOwnerOfAllToken
      );
    }
    return false;
  }

  function validateScanImport() {
    if ($selectedNFTs && $selectedNFTs.length > 0 && $destinationChain && scanned) {
      if ($selectedNFTs[0].type === TokenType.ERC1155) {
        return $enteredAmount > BigInt(0);
      } else if ($selectedNFTs[0].type === TokenType.ERC721) {
        return true;
      }
    }
    return false;
  }
  $: canProceed =
  importMethod === 'manual' && validateManualImport()
    ? true
    : importMethod === 'scan' && validateScanImport()
    ? true
    : false;
```
    
    I think you should do something similar to this to make the code look clearer. This one looks too bad

packages/bridge-ui-v2/src/components/Bridge/Actions.svelte
```diff
@@ -49,6 +49,7 @@
 
   //TODO: this should probably be checked somewhere else?
```
NguyenTuanCanh, line 50:
```
  function getSpenderAddress(spenderAddresses: any, tokenType: string) {
    const addressMap: Record<string, string> = {
      [TokenType.ERC1155]: 'erc1155VaultAddress',
      [TokenType.ERC721]: 'erc721VaultAddress',
    };

    return spenderAddresses?.[addressMap[tokenType]] || '';
  }

  //TODO: this should probably be checked somewhere else?
  export async function checkTokensApproved() {
    if ($selectedToken?.type === TokenType.ERC721 || $selectedToken?.type === TokenType.ERC1155) {
      if ($account?.address && $network?.id && $destNetwork?.id) {
        const currentChainId = $network?.id;
        const destinationChainId = $destNetwork?.id;
        const token = $selectedToken as NFT;

        const result = await checkOwnershipOfNFT(token as NFT, $account?.address, currentChainId);

        if (!result.every((item) => item.isOwner === true)) {
          return false;
        }

        const wallet = await getConnectedWallet();
        const spenderAddresses = routingContractsMap[currentChainId][destinationChainId];

        const args: RequireApprovalArgs = {
          tokenAddress: token.addresses[currentChainId],
          owner: wallet.account.address,
          spenderAddress: getSpenderAddress(spenderAddresses, token.type),
          tokenId: BigInt(token.tokenId),
          chainId: currentChainId,
        };

        if (token.type === TokenType.ERC1155) {
          try {
            const bridge = bridges[token.type] as ERC1155Bridge;

            // Let's check if the vault is approved for all ERC1155
            const result = await bridge.isApprovedForAll(args);
            allTokensApproved = result;
          } catch (error) {
            console.error('isApprovedForAll error');
          }
        } else if (token.type === TokenType.ERC721) {
          const bridge = bridges[token.type] as ERC721Bridge;

          // Let's check if the vault is approved for all ERC721
          try {
            const requiresApproval = await bridge.requiresApproval(args);
            allTokensApproved = !requiresApproval;
          } catch (error) {
            console.error('isApprovedForAll error');
          }
        }
      }
    }
  }
```
I think this will be better. Avoid repeating code

packages/bridge-ui-v2/src/components/Bridge/Amount.svelte
KorbinianK: Essentially comes down to readability, but as regex is a bit more performant and it is happening on typing input, I agree! Thanks

packages/bridge-ui-v2/src/components/Bridge/NFTBridgeSteps/ImportStep.svelte
KorbinianK: Absolutely! 👍  There are still a couple of places where functions grew too much and should be deconstructed into smaller individual parts. Good catch, will improve this

xiaodino (dismissed):

dantaik (dismissed):

dantaik:
@KorbinianK It seems there is no change in this PR somehow.

KorbinianK:
> @KorbinianK It seems there is no change in this PR somehow.

@dantaik oh... because the other PR got merged before this one. I'll close it then...

PR #15090, opened 2023-10-31
fix(protocol): ensure proposer is EOA when use calldata for DA

dantaik:


davidtaikocha (approved):

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -94,9 +95,15 @@ library LibProposing {
 
             if (blobHash == 0) revert L1_NO_BLOB_FOUND();
         } else {
+            // The proposer must be an Externally Owned Account (EOA) for
+            // calldata usage. This ensures that the transaction is not an
+            // internal one, making calldata retrieval more straightforward.
```
Brechtpd, line 100: This may break integrating Taiko with for example PBS because that will likely require some external contract, unless we do something like allowing an extra callback to be specified in the propose function that can do some extra logic.

I also seem to remember this check wasn't 100% safe to detect all cases (when not used in delegatecall), but I can't immediately find anything on it and and I also can't think of any ways to cheat this check.

dantaik: I think we can come back to fix this if this really an issue for PBS. It's too early to predict how the integration will look like exactly, maybe it's just a contract-based signature check. So fixing it too early will require the client to do quite some work...

For EOA checking, i switched to use this approach: https://github.com/taikoxyz/taiko-mono/pull/15095

PR #15092, opened 2023-10-31
fix(protocol): Change sgx instanceId to 4bytes

adaki2004:
As per title.
According to Brecht, to minimize a possibility to spam the sgx bootstrap nodes successfully, we shall increase the range from 65K max possible ids.

### DIFF
--- a/packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
+++ b/packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
@@ -110,14 +110,14 @@ contract SgxVerifier is EssentialContract, IVerifier {
         // Do not run proof verification to contest an existing proof
         if (isContesting) return;
 
-        // Size is: 87 bytes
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -110,14 +110,14 @@ contract SgxVerifier is EssentialContract, IVerifier {
         // Do not run proof verification to contest an existing proof
         if (isContesting) return;
 
-        // Size is: 87 bytes
-        // 2 bytes + 20 bytes + 65 bytes (signature) = 87
-        if (evidence.proof.length != 87) revert SGX_INVALID_PROOF();
+        // Size is: 89 bytes
+        // 4 bytes + 20 bytes + 65 bytes (signature) = 89
```
dantaik, line 114: 8 bytes instead?

PR #15102, opened 2023-11-01
feat(protocol): allow pausing proving

dantaik:


Brechtpd (approved):
Should we also pause `verifyBlocks`? Just thinking of ways where we can prevent things from going wrong as quickly as possible with the least amount of transactions. If we see invalid proofs being submitted it will take a while to fix them all, preventing blocks from being finalized when being paused gives us all the time we need to fix all things without having to worry about things taking some time.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -83,6 +83,8 @@ library LibVerifying {
         // Retrieve the latest verified block and the associated transition used
         // for its verification.
         TaikoData.SlotB memory b = state.slotB;
+        if (b.provingPaused) return;
```
dantaik, line 86: @Brechtpd if proving is paused, verification is effectively paused as well, but we didn't revert here, simply return (doing nothing)

dantaik:
> Should we also pause `verifyBlocks`? Just thinking of ways where we can prevent things from going wrong as quickly as possible with the least amount of transactions. If we see invalid proofs being submitted it will take a while to fix them all, preventing blocks from being finalized when being paused gives us all the time we need to fix all things without having to worry about things taking some time.

Done

Brechtpd (approved):

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -129,11 +132,18 @@ contract TaikoL1 is
...
     function verifyBlocks(uint64 maxBlocksToVerify) external nonReentrant {
         if (maxBlocksToVerify == 0) revert L1_INVALID_PARAM();
+        if (state.slotB.provingPaused) revert L1_PROVING_PAUSED();
+
         LibVerifying.verifyBlocks(
             state, getConfig(), AddressResolver(this), maxBlocksToVerify
         );
     }
 
+    /// @notice Pause block proving.
```
adaki2004, line 142:
```suggestion
    /// @notice Pause block proving.
    /// @param pause True if paused.
```

PR #15104, opened 2023-11-01
fix(protocol): deploy TransparentUpgradeableProxy for bridged token differently

dantaik:
@adaki2004 @davidtaikocha 
we should deploy a ProxiedBridgedERC1155 before hand, registered it as "proxied_bridged_erc1155" in AddressManager, then only deploy a TransparentUpgradeableProxy when necessary. The current deployment strategy is very expensive -- each new bridged token will trigger a deployment of the implementation, this is not necessary.

@adaki2004 please finish this PR for the other two vaults, then fix tests.
@davidtaikocha please fix the deployment scripts.


adaki2004:
> @adaki2004 please finish this PR for the other two vaults, then fix tests.

Done.

packages/protocol/test/tokenvault/ERC1155Vault.t.sol
```diff
@@ -204,6 +208,20 @@ contract ERC1155VaultTest is TestBase {
 
         vm.deal(address(bridge), 100 ether);
 
+        address erc1155_common_logic = LibBridgedTokenDeployer
```
dantaik, line 211: rename to proxiedBridgedERC1155?

packages/protocol/utils/generate_genesis/taikoL2.ts
```diff
@@ -208,19 +187,19 @@ async function generateContractConfigs(
...
                     contractArtifacts.ProxiedSingletonTaikoL2,
-                    addressMap,
+                    addressMap
                 );
                 break;
             case "ProxiedSingletonBridge":
                 bytecode = linkContractLibs(
                     contractArtifacts.ProxiedSingletonBridge,
-                    addressMap,
+                    addressMap
```
dantaik, line 196: somehow my git-triggered autoformat  will use a different style.

packages/protocol/test/TestBase.sol
```diff
@@ -51,6 +54,22 @@ abstract contract TestBase is Test {
     address internal SGX_Z = getRandomAddress();
 }
 
+library LibBridgedTokenDeployer {
+    function deployLogicContract(
```
dantaik, line 58: can we deploy the ERC token implementation contract the same way as any other contract without using Create2?

adaki2004: A simple 'new' makes it for sure , just re-used the previous approach if we ever wanted to have deterministic deployments. But i can modify it.

dantaik: I don't think we need deterministic deployments unless I missed something. Only the proxy needs to be deterministic per token.

dantaik: Actually maybe we do have a problem here, maybe not a big one:

currently the btoken address is not deterministic.

dantaik: I prepare a draft PR for you @adaki2004

adaki2004: Yes, i agree. We only do care about 1 implementation (which will be shared) and different proxies. I think it has nothing to do with `CREATE2` - only in case we want to have some addresses deterministic.

PR #15105, opened 2023-11-02
feat(protocol): allow DA slicing and blob sharing

dantaik:


packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -62,7 +25,7 @@ library LibUtils {
         uint64 blockId,
         bytes32 parentHash
     )
-        internal
+        external
```
dantaik, line 28: some changes to make sure the TaikoL1's code is smaller than the limit.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -117,30 +93,100 @@ library LibProposing {
...
+                }
+            }
+
+            if (params.txListByteSize == 0) revert L1_TXLIST_OFFSET_SIZE();
+            if (
+                uint256(params.txListByteOffset) + params.txListByteSize
+                    > config.blockMaxTxListBytes
+            ) {
+                revert L1_TXLIST_TOO_LARGE();
+            }
```
Brechtpd, line 184: I think this check is only useful if `config.blockMaxTxListBytes` is set to the blob size, because if it's not the blob size we still have to check in the circuit if the range is valid anyway. But the same constant is also used with the calldata tx list so not sure if that will be the case or not. Currently they would have very similar limits, so setting it to the blob size would make sense.

Brechtpd (approved):
Just one question, otherwise looks good!

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: Change the code to also check against MAX_BYTES_PER_BLOB

PR #15109, opened 2023-11-02
refactor(bridge-ui-v2): Create a function to check a condition.

xxxigm:
There are some overly long combined conditions, I think we should put them into an array for processing. This helps make the code cleaner.

### DIFF
--- a/packages/bridge-ui-v2/src/components/Bridge/NFTBridgeSteps/ConfirmationStep.svelte
+++ b/packages/bridge-ui-v2/src/components/Bridge/NFTBridgeSteps/ConfirmationStep.svelte
@@ -143,8 +143,21 @@
     }
   }
 
+  const _notAllowBridge = ()  => {
+    const conditions = [
+      !$bridgeService,
+      !$selectedToken,
+      !$network,
+      !$destNetwork?.id,
+      !$account?.address,
+    ]
+
+    return conditions.some(Boolean)
+  }
+
   async function bridge() {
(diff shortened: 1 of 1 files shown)


packages/bridge-ui-v2/src/components/Bridge/NFTBridgeSteps/ConfirmationStep.svelte
```diff
@@ -143,8 +143,21 @@
...
+    const conditions = [
+      !$bridgeService,
+      !$selectedToken,
+      !$network,
+      !$destNetwork?.id,
+      !$account?.address,
+    ]
+
+    return conditions.some(Boolean)
+  }
```
KorbinianK, line 156:
```suggestion
  const canBridge = ()  => {
    const conditions = [
      $bridgeService,
      $selectedToken,
      $network,
      $destNetwork?.id,
      $account?.address,
    ]
    return conditions.some(Boolean)
  }
```

as per your own standard from a previous PR, it would be better to use the positive check.

NguyenTuanCanh: @KorbinianK I think it must be 'every', why 'some'?

KorbinianK: Yes absolutely, for the inverted check it should be `every`.

2manslkh:
@KorbinianK are we gonna merge this PR?

KorbinianK:
Stale. Let's close it. I add the suggested changes to another PR.

PR #15113, opened 2023-11-02
make btoken address deterministic

dantaik:
If we do this, we need to keep the changes made by David which uses LibDeploy.

### DIFF
--- a/packages/protocol/contracts/libs/LibDeploy.sol
+++ b/packages/protocol/contracts/libs/LibDeploy.sol
@@ -12,7 +12,8 @@ import { TransparentUpgradeableProxy } from
     "lib/openzeppelin-contracts/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 
 library LibDeploy {
-    /// @dev Deploys a contract (via proxy)
+    /// @dev Deploys a TransparentUpgradeableProxy proxy with deterministic
+    /// address before a given implementation.
     function deployDetermisticUpgradableProxy(
         address owner,
         address logic,
@@ -22,6 +23,7 @@ library LibDeploy {
         external
         returns (address)
     {
+        // TODO: check owner, logic
         return Create2Upgradeable.deploy(
             0,
             salt,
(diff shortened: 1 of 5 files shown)


packages/protocol/contracts/libs/LibDeploy.sol
```diff
@@ -0,0 +1,34 @@
...
+        bytes memory data
+    )
+        external
+        returns (address)
+    {
+        return Create2Upgradeable.deploy(
+            0,
+            salt,
+            abi.encodePacked(
+                type(TransparentUpgradeableProxy).creationCode,
```
dantaik, line 29: not tested

packages/protocol/script/DeployOnL1.s.sol
```diff
@@ -326,7 +329,8 @@ contract DeployOnL1 is Script {
             "erc721_vault",
             address(erc721Vault),
             bytes.concat(
-                erc721Vault.init.selector, abi.encode(addressManagerForSingletons)
+                erc721Vault.init.selector,
```
dantaik, line 332: my commit triggerred autoformat

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
```diff
@@ -352,13 +350,13 @@ contract ERC1155Vault is BaseNFTVault, ERC1155ReceiverUpgradeable {
...
         );
-        btoken = address(
-            new TransparentUpgradeableProxy(
-                resolve("proxied_bridged_erc1155", false),
-                owner(),
-                data
-            )
-        );
+
+        btoken = LibDeploy.deployDetermisticUpgradableProxy({
```
dantaik, line 354: @adaki2004 please do not merge this PR. 

There maybe an issue with this approach.

Is it possible that someone else deploy a btoken before we do, so the address is already used, so we cannot deploy the contract in vaults?

dantaik: I think it is indeed an issue. We should not merge this PR :)

PR #15122, opened 2023-11-03
feat(protocol): add airdrop contracts (not final)

### LINKED ISSUE #15026: Airdrop contract (+mock backend) creation with merkle proofs

As per title. (Currently handle with low prio).

### Additional context

Additional context here.

adaki2004:
As per issue: https://github.com/taikoxyz/taiko-mono/issues/15026
(The merkle tree build up is on purpose written in typescript - not solidity in foundry tests - hence we can reuse at production.)


packages/protocol/contracts/airdrop/ClaimAirdrop.sol
```diff
@@ -0,0 +1,89 @@
+// SPDX-License-Identifier: MIT
+//  _____     _ _         _         _
```
dantaik, line 2: suggest to move this to contracts/team/airdrop

dantaik:
May I suggest something more generic and can support both NFT and ERC20 tokens.



The claim contract can be a abstract contract with a virtual funciton:

```solidity
function claimWithData(bytes calldata) virtual external;
```


In this contract we implement:

```solidity
funciton claim(bytes calldata data, bytes mkproof) external {
    bytes32 v = kecceck256(data);
    if (claimed[v]) revert CLAIMED_ALREADY();
 
    LibBinaryMerkelTree.validateProof(getSlot(v), uint(1), mkproof, this.rootHash);
    claimed[v] = true;
    claimWithData(data);
}
```

A ERC20 claim contract can implement claimWithData as:

```solidity
function claimWithData(bytes calldata data) external override {
   (address token, address recipient, uint amount) = abi.decode(data, (address, address, uint));
   token.transferFrom(someVault, recipient, amount);
}
```

packages/protocol/contracts/team/airdrop/ClaimAirdrop.sol
```diff
@@ -0,0 +1,89 @@
...
+
+            emit UserClaimed(msg.sender, allowance);
+            return;
+        }
+        revert CLAIMED_ALREADY();
+    }
+
+    /// @notice Set new Merkle Root
+    /// @param _merkleRoot Root of merkle tree
+    function setMerkleRoot(bytes32 _merkleRoot) external onlyOwner {
```
dantaik, line 75: Setting the merkle root in init()?

adaki2004: > Setting the merkle root in init()?

Can do that as well, but still having to have this feature (setter) so that in case the merkle tree changes (bc. we add a new leaf, or so) we can easily set it (again).

adaki2004:
> May I suggest something more generic and can support both NFT and ERC20 tokens.
> 
> The claim contract can be a abstract contract with a virtual funciton:
> 
> ```solidity
> function claimWithData(bytes calldata) virtual external;
> ```
> 
> In this contract we implement:
> 
> ```solidity
> funciton claim(bytes calldata data, bytes mkproof) external {
>     bytes32 v = kecceck256(data);
>     if (claimed[v]) revert CLAIMED_ALREADY();
>  
>     LibBinaryMerkelTree.validateProof(getSlot(v), uint(1), mkproof, this.rootHash);
>     claimed[v] = true;
>     claimWithData(data);
> }
> ```
> 
> A ERC20 claim contract can implement claimWithData as:
> 
> ```solidity
> function claimWithData(bytes calldata data) external override {
>    (address token, address recipient, uint amount) = abi.decode(data, (address, address, uint));
>    token.transferFrom(someVault, recipient, amount);
> }
> ```
> 

Good idea, will do so !

> I further suggest:
> 
>     * This contract should not hold any token, another contract (vault) should hold the token and set allowance for the claim contract.

Hmm i'm not so sure in this, i mean i had the same idea initially and was thinking what would be the actual benefit ? I see extra complexity (+1 contract + logic + tests + deployments) while the end result is the same, and not necessarily safer or more flexible.

>     * This contract should also have a claim deadline, after that, nothing can be claimed. No need to withdraw token as there is no token in this contract.

Deadline ok, but please let me know why you think we need extra contract as i see it as an overhead.

dantaik:
No extra contract needed. We have our vault to keep tokens the DAO has or the company has, we just need that contract to set up allowance to this airdrop claiming contract.

adaki2004:
LibBinaryMerkelTree.validateProof(getSlot(v), uint(1), mkproof, this.rootHash);
    claimed[v] = true;
    claimWithData(data);

@dantaik 
Generalizing it (to an abstract parent class) is a good idea tho tHis is not how we validate merkle proofs (for off-chain allowlist). We do not need to sotre thing in a contract storage. We can use the openzeppelin merkel proof validation code - and please see the off-chain typescript code how i did it.

packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
```diff
@@ -0,0 +1,68 @@
...
+        if (merkleRoot == 0x0) {
+            revert CLAIM_NOT_STARTED();
+        }
+
+        ERC20InputParams memory input =
+            abi.decode(inputData, (ERC20InputParams));
+
+        // Create the leafs - a leaf consist of the 2 data encoded (address and
+        // allowance). See claimList.ts and buildMerkleTree.ts for more info !
+        bytes32 leaf = keccak256(abi.encodePacked(msg.sender, input.allowance));
```
Brechtpd, line 46: Would allow people to claim for different addresses as well instead of requiring `msg.sender` to do it itself.

adaki2004: I see the point, but from a practical standpoint, how you “chose” from the (example) 10K database ?
You (Bob) goes to the website, and via the UI you cherry-pick Alice address to claim for her ? 🤔 would not it make sense that everyone claims himself/herself ?

Brechtpd: Yeah you check for eligibility for all your addresses and then claim them using an address that is most convenient. For example I have an address I just use for doing small transactions that is easier to use than other addresses that may require more work (hardware wallet/smart wallet). It can also be safer in those cases because you don't have to do any transaction from those wallets that contain more wallets so have to be more careful when doing transactions for those.

packages/protocol/contracts/team/airdrop/BaseAirdrop.sol
```diff
@@ -0,0 +1,73 @@
...
+
+import { Proxied } from "../../common/Proxied.sol";
+
+/// @title BaseAirdrop
+/// Contract for managing Taiko token airdrop for eligible users
+abstract contract BaseAirdrop is OwnableUpgradeable {
+    mapping(bytes32 => bool) public leafClaimed;
+    address public tokenVaultAddress; // Vault
+    address public tokenAddress; // Token
+    bytes32 public merkleRoot = 0x0;
```
dantaik, line 22:
```suggestion
    bytes32 public merkleRoot;
```

packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
```diff
@@ -0,0 +1,68 @@
...
+        verifyProof(input.merkleProof, leaf);
+
+        if (!leafClaimed[leaf]) {
```
dantaik, line 49: some logics are common and shall be put into the base class, only 

```
 bool success = IERC20(tokenAddress).transferFrom(
                tokenVaultAddress, msg.sender, input.allowance
            );
            
```

is specific.

dantaik (approved):

packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
dantaik: Please also create an ERC721 airdrop contract to learn what are shard logics.

packages/protocol/contracts/team/airdrop/BaseAirdrop.sol
```diff
@@ -0,0 +1,73 @@
...
+
+    error CLAIM_NOT_STARTED();
```
dantaik, line 24: errors not used in this contract should be deleted.

packages/protocol/test/team/airdrop/ERC20Airdrop.t.sol
```diff
@@ -0,0 +1,138 @@
...
+        vm.prank(Alice, Alice);
+        airdrop.claim(abi.encode(input));
+    }
+
+    function test_claim_but_with_invalid_allowance() public {
+        airdrop.setMerkleRoot(merkleRoot);
+
+        // These proofs are coming from 'pnpm run buildMerkle'
+        bytes32[] memory merkleProof = new bytes32[](3);
+        merkleProof[0] =
```
dantaik, line 67: our code is so simple and we should have very simple tests for our own code, we are not supposed to test the proof verification, it must have been tested by OZ.

packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
adaki2004: Sure, but it would mean everyone could claim (and check eligibility) for everyone.. if which is OK with you, then it is OK with me :) 
Another Q for the UI then. So you go to the platform, and you connect your wallet ?  Or just there is a sinple input text form, where you can initiate a query per specific address ?
And if address is eligible, then you can request for proofs.. and can claim for anybody ?

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -0,0 +1,101 @@
...
+
+/// @title MerkleClaimable
+/// Contract for managing Taiko token airdrop for eligible users
+abstract contract MerkleClaimable is OwnableUpgradeable {
+    mapping(bytes32 => bool) public isClaimed;
+    bytes32 public merkleRoot;
+    uint128 public claimStart;
+    uint128 public claimEnd;
+
+    event Claimed(bytes32 hash);
```
adaki2004, line 22: reserved slots

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -0,0 +1,101 @@
...
+import { MerkleProofUpgradeable } from
+    "lib/openzeppelin-contracts-upgradeable/contracts/utils/cryptography/MerkleProofUpgradeable.sol";
+
+/// @title MerkleClaimable
+/// Contract for managing Taiko token airdrop for eligible users
+abstract contract MerkleClaimable is OwnableUpgradeable {
+    mapping(bytes32 => bool) public isClaimed;
+    bytes32 public merkleRoot;
+    uint128 public claimStart;
+    uint128 public claimEnd;
```
adaki2004, line 20: change to uint64 (all time)

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -0,0 +1,101 @@
...
+        uint128 _claimEnd,
+        bytes32 _merkleRoot
+    )
+        internal
+    {
+        OwnableUpgradeable.__Ownable_init();
+
+        if (_merkleRoot == 0x0) {
+            revert INVALID_MERKLE_ROOT();
+        }
```
adaki2004, line 83: move to internal

packages/protocol/utils/airdrop/buildMerkleTree.ts
```diff
@@ -0,0 +1,83 @@
+import { MerkleTree } from "merkletreejs/dist/MerkleTree";
+import { claimList } from "./airdrop_db/claimList";
```
adaki2004, line 2: custom json file input from cmd

dantaik: This is the only pending issue @adaki2004

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -39,6 +44,7 @@ abstract contract MerkleClaimable is OwnableUpgradeable {
         bytes32[] calldata proof
     )
         external
+        nonReentrant
```
adaki2004, line 49: I think this is not necessary, we solved it 'in-house' already.
A contract is vulnerable if it `calls`, `sends` or `transfers` to an untrusted address (malicious contract with fallback function).
A design to prevent that is check-effects interaction. (https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html)
We already have that given (by line 52 + line 58), because the only place an untrusted address can be invoked, is within `_claimWIthData()` but we already marked the hash as `used`.

dantaik (approved):

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
dantaik: If you are sending an ERC1155 token to an target address and the target address is a contract and implement a ERC1155 callback, then the callback can call this contract to re-enter.

adaki2004: > If you are sending an ERC1155 token to an target address and the target address is a contract and implement a ERC1155 callback, then the callback can call this contract to re-enter.

Yes. But it would revert because the hash is already “used”. (Line 60)

PR #15128, opened 2023-11-04
improve the airdrop contract design

dantaik:
@adaki2004 I thought this is easy for you after you working on the bridge and vaults...

### DIFF
--- a/packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
+++ b/packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
@@ -15,6 +15,10 @@ import { Proxied } from "../../common/Proxied.sol";
 
 /// @title MerkleClaimable
 /// Contract for managing Taiko token airdrop for eligible users
+// TODO(dani): add claimStart and claimEnd timestamp so claim() can only be
+// called between these two timestamps.
+// Better to add non-reentrance guard to claim().
+
 abstract contract MerkleClaimable is OwnableUpgradeable {
     mapping(bytes32 => bool) public isClaimed;
     bytes32 public merkleRoot;
--- a/packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
+++ b/packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
@@ -29,7 +29,7 @@ abstract contract MerkleClaimable is OwnableUpgradeable {
     error INVALID_PROOF();
 
     function claim(bytes calldata data, bytes32[] calldata proof) external {
-        bytes32 hash = keccak256(abi.encode("CLAIM)TAIKO_AIRDROP", data));
+        bytes32 hash = keccak256(abi.encode("CLAIM_TAIKO_AIRDROP", data));
         if (isClaimed[hash]) revert CLAIMED_ALREADY();
 
         if (!MerkleProofUpgradeable.verify(proof, merkleRoot, hash)) {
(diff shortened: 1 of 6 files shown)


packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -0,0 +1,51 @@
...
+    mapping(bytes32 => bool) public isClaimed;
+    bytes32 public merkleRoot;
+
+    event Claimed(bytes32 hash);
+
+    error CLAIMED_ALREADY();
+    error INVALID_PROOF();
+
+    function claim(bytes calldata data, bytes32[] calldata proof) external {
+        bytes32 hash = keccak256(abi.encode("CLAIM)TAIKO_AIRDROP", data));
```
dantaik, line 32:
```suggestion
        bytes32 hash = keccak256(abi.encode("CLAIM_TAIKO_AIRDROP", data));
```

adaki2004:
Sure! Will take over and continue (most prob on Monday). Thanks for chiming in.

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -0,0 +1,51 @@
...
+import { MerkleProofUpgradeable } from
+    "lib/openzeppelin-contracts-upgradeable/contracts/utils/cryptography/MerkleProofUpgradeable.sol";
+
+import { Proxied } from "../../common/Proxied.sol";
+
+/// @title MerkleClaimable
+/// Contract for managing Taiko token airdrop for eligible users
+// TODO(dani): add claimStart and claimEnd timestamp so claim() can only be
+// called between these two timestamps.
+// Better to add non-reentrance guard to claim().
```
adaki2004, line 20: `Line 33` already prevents reentrancy, because the only time a contract can be invoked is from the call _claimWithData() but by then (in case of a reentrancy) the hash is already registered as claimed.

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -0,0 +1,51 @@
...
+    mapping(bytes32 => bool) public isClaimed;
+    bytes32 public merkleRoot;
+
+    event Claimed(bytes32 hash);
+
+    error CLAIMED_ALREADY();
+    error INVALID_PROOF();
+
+    function claim(bytes calldata data, bytes32[] calldata proof) external {
+        bytes32 hash = keccak256(abi.encode("CLAIM_TAIKO_AIRDROP", data));
```
adaki2004, line 32: been playing around with it for a while now (testing the off-chain, on-chain scripts) now, and figured it is not going to work out properly, i'd suggest something like this:

```

        bytes32 hash = keccak256(abi.encode("CLAIM_TAIKO_AIRDROP", user, amount));
```

The reason is, calldata has some 'marker' appended to the data itself.

So if i do this:

        `console2.logBytes(abi.encode("CLAIM_TAIKO_AIRDROP", data));`

The result is this:
`0x000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000013434c41494d5f5441494b4f5f41495244524f500000000000000000000000000000000000000000000000000000000000000000000000000000000000000000400000000000000000000000007e5f4552091a69125d5dfcb7b8c2659029395bdf0000000000000000000000000000000000000000000000000000000000000064` 

But if i decode the value (as proposed above):
`0x00000000000000000000000000000000000000000000000000000000000000600000000000000000000000007e5f4552091a69125d5dfcb7b8c2659029395bdf00000000000000000000000000000000000000000000000000000000000000640000000000000000000000000000000000000000000000000000000000000013434c41494d5f5441494b4f5f41495244524f5000000000000000000000000000`

So it is more easy to create the raw data off-chain if we dont need to consider solidity encoding structure.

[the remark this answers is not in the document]
adaki2004: PR for that: https://github.com/taikoxyz/taiko-mono/pull/15162

PR #15161, opened 2023-11-09
feat(protocol): add maxBlockId to ProverAssignment

dantaik:
Allow checking if the assignment has expired using block ID.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -306,7 +306,7 @@ library LibProposing {
     {
         // Check assignment not expired
         if (
-            block.timestamp >= assignment.expiry
+            block.timestamp > assignment.expiry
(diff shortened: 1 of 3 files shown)


davidtaikocha (approved):

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -296,13 +298,17 @@ library LibProposing {
...
+        uint64 blockId,
         TaikoData.ProverAssignment memory assignment
     )
         private
         returns (uint256 proverFee)
     {
         // Check assignment not expired
-        if (block.timestamp >= assignment.expiry) {
+        if (
+            block.timestamp >= assignment.expiry
```
adaki2004, line 309: remove equal

dantaik, line 309:
```suggestion
            block.timestamp > assignment.expiry
```

PR #15173, opened 2023-11-10
feat(protocol): allow pay L1 builder a conditional tip

dantaik:
I'm not sure we should add this feature to the protocol. Just to show now the metahash can be used this way.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -98,7 +98,11 @@ library LibProposing {
         // If we choose to persist all data fields in the metadata, it will
         // require additional storage slots.
         unchecked {
+            TaikoData.Block storage parentBlk =
+                state.blocks[(b.numBlocks - 1) % config.blockRingBufferSize];
+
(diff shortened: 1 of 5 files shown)


davidtaikocha (approved):
The implementation idea LGTM

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -325,22 +336,28 @@ library LibProposing {
...
+        uint256 tip;
         if (assignment.feeToken == address(0)) {
-            // Paying Ether
             if (msg.value < proverFee) revert L1_ASSIGNMENT_INSUFFICIENT_FEE();
-            assignment.prover.sendEther(proverFee);
             unchecked {
-                // Return the extra Ether to the proposer
-                uint256 refund = msg.value - proverFee;
-                if (refund != 0) msg.sender.sendEther(refund);
+                tip = proverFee - msg.value;
```
adaki2004, line 343: change order

dantaik: fixed

Brechtpd (approved):

PR #15200, opened 2023-11-13
feat(protocol): restrict max gas paying prover

dantaik:


davidtaikocha (approved):

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -28,6 +28,11 @@ library LibProposing {
     // field element has 32 bytes.
     uint256 public constant MAX_BYTES_PER_BLOB = 4096 * 32;
 
+    // Max gas paying the prover. This should be large enough to prevent the
+    // worst cases, usually block proposer shall be aware the risks and only
+    // choose provers that cannot consume too much gas when receiving Ether.
```
Brechtpd, line 33: Easy for provers to give the proposer an address that is an EOA and then deploy a smart contract to that address that consumes a lot of gas!

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -28,6 +28,11 @@ library LibProposing {
...
+    uint256 public constant MAX_GAS_PAYING_PROVER = 200_000;
```
Brechtpd, line 34: Seems a bit high but fair enough

Brechtpd (approved):

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: Yes, but provers are most likely entities that win their businesses over time by behaving well, some metrics are important for proposer when to consider which provers to use.

Brechtpd: Shouldn't they just pick the prover that offers the lowest fee? Anything that is more trust based seems like a downside for the system to work as well and efficient as possible. Any trust based system would also be more centralizing.

PR #15208, opened 2023-11-14
feat(protocol): move prover assignment verification to hook

dantaik:
Brecht, David, Dani, I'm trying to prototype this hook idea so more logics can be extracted away from the protocol. Take a look at this, so later we can brainstorm a bit. No rush though. 

- Daniel


adaki2004:
I have no strong opinion on this, imo it has pros and cons.
I like that simplifies and makes code (`LibProposing`) more readable, but also on the other hand it adds some overhead. (2600 extra gas cost for extrnal function call + a deployment of a new standalone contract). Beside that LGTM.

dantaik:
This PR is ready for you to review (the idea), tests are failing and will be fixed only when the idea is approved.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -0,0 +1,172 @@
...
+                || assignment.maxBlockId != 0 && meta.id > assignment.maxBlockId
+                || assignment.maxProposedIn != 0
+                    && block.number > assignment.maxProposedIn
+        ) {
+            revert HOOK_ASSIGNMENT_EXPIRED();
+        }
+
+        // Hash the assignment with the blobHash, this hash will be signed by
+        // the prover, therefore, we add a string as a prefix.
+        bytes32 hash = hashAssignment(assignment, address(this), meta.blobHash);
```
adaki2004, line 82: According to the function declaration (`hashAssignment()`) i think it shall be `msg.sender`
```suggestion
        bytes32 hash = hashAssignment(assignment, msg.sender, meta.blobHash);
```

packages/protocol/test/L1/TaikoL1TestBase.sol
```diff
@@ -310,16 +316,17 @@ abstract contract TaikoL1TestBase is TestBase {
...
     )
         internal
         view
         returns (bytes memory signature)
     {
-        bytes32 digest =
-            LibProposing.hashAssignment(assignment, taikoAddr, blobHash);
+        bytes32 digest = 0;
+        // TODO
+        // AssignmentHook.hashAssignment(assignment, taikoAddr, blobHash);
```
adaki2004, line 329: I cannot comment `line 383` for whatever reason (most probably because it has not changed) but it is not valid anymore:
`tko.approve(address(L1), amountTko);`

The deployed `assignmentHook` has to be approved as the `spender`.

dantaik: Tests need to changed later.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
dantaik: I think both will work.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -229,35 +224,46 @@ library LibProposing {
...
         });
+
+        {
+            TaikoToken tko = TaikoToken(resolver.resolve("taiko_token", false));
+            uint256 tkoBalance = tko.balanceOf(address(this));
+
+            // Run all hooks.
+            // Note that address(this).balance has been updated with msg.value,
+            // prior to any code in this function has been executed.
+            uint256 ethBalance = address(this).balance - msg.value;
```
Brechtpd, line 249: I think no ETH is stored in this contract, so this could simply be `address(this).balance`? Same thing below I think where instead of having to keep track of `msgValue`, it could always just be  `address(this).balance`.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -229,35 +224,46 @@ library LibProposing {
...
+
+            // Run hooks
+            for (uint256 i; i < params.hookCalls.length; ++i) {
```
Brechtpd, line 252: Is it necessary/useful to support multiple hooks? Seems like only a single hook will probably be used, and even then this single hook itself can call into multiple other smart contracts if necessary.

Brechtpd (approved):
Looks good, just some small comments.

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: I think one hook is good enough for default scenario, but I want to support multiple hooks so proposers, with their imagination, can do extra things.

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: True.  I added some comments and removed `ethBalance`

PR #15212, opened 2023-11-14
docs(website): changed chainlist link to list taiko chains and testnets

0xhwk:
Changed chainlist link so users can see taiko chains directly and 'Include testnets' checked as documents suggest.


2manslkh (approved):
makes sense

dionysuzx (approved):

PR #15214, opened 2023-11-15
RGB Code Information Added for Taiko Branding

alibabacapital:
A refined version of the sentence with the added RGB information.


KorbinianK:
I don't think this adds anything of value. The link to figma already has all the information.

PR #15234, opened 2023-11-20
feat(protocol): Add airdrop contracts

adaki2004:
Same as here: https://github.com/taikoxyz/taiko-mono/pull/15122 but that one had lots of conflicts so rebased to the new base.

`ERC20Airdrop2`'s missing `getBalance` also implemented (with tests) for unlocking.


packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
```diff
@@ -0,0 +1,93 @@
...
+    /// @return balance The balance the user successfully claimed
+    /// @return withdrawableAmount The amount available to withdraw
+    function getBalance(address user)
+        public
+        view
+        returns (uint256 balance, uint256 withdrawableAmount)
+    {
+        balance = claimedAmount[user];
+        if (balance == 0) return (0, 0);
+        uint256 timeBasedAllowance = balance * uint64(block.timestamp - claimEnd) / withdrawalWindow;
```
dantaik, line 84: block.timestamp maybe super large if the user wait long...

dantaik: we need a test for this contract

adaki2004: > block.timestamp maybe super large if the user wait long...

Withdraw only allows to withdraw within the window. (Tho people could still query the readonly function of getBalance() so i’ll cap it.

adaki2004: > we need a test for this contract

Added one testing the withdrawal part, but will add +1-2 extra more.

dantaik:
Two questions:

1. If the USDC contract code must be deployed as-is on the byte-level, why don't we simply copy the byte-code into this project?
2. If the design is generic, we can/should test the design with a mock native token contract, not with a real one, which will avoid copying so many test contract files into this repo.

adaki2004:
> Two questions:
> 
>     1. If the USDC contract code must be deployed as-is on the byte-level, why don't we simply copy the byte-code into this project?
> 
>     2. If the design is generic, we can/should test the design with a mock native token contract, not with a real one, which will avoid copying so many test contract files into this repo.

I think you meant to ask these questions in this PR https://github.com/taikoxyz/taiko-mono/pull/15250, no ?

But if we are here, my answers:
1. Yes, i agree this is they way, i just wanted a clean USDC contract which represents the actual state. We can copy the USDC bytecode and deploy it - on mainnet/testnet.
2. I would say yes and no at the same time.
Generic in terms of any kind of native assets could be plugged into the system.
Non-generic when it comes to each individual contracts - like USDC. It requires a 'translator' which maps the abbreviations (which in the USDC case is the burn function). So i wanted to E2E test a valid USDC contract, so that we are sure we know what we are doing, how we doing it, and it works on testnets and mainnets. I think it is definitely worth having +some extra files for safety- in the test folder, then figuring out live on testnet what is wrong.

dantaik:
@adaki2004 May I ask you to rebased this PR with branch `proxy`?

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -0,0 +1,90 @@
...
+    "lib/openzeppelin-contracts-upgradeable/contracts/access/OwnableUpgradeable.sol";
+import { ReentrancyGuardUpgradeable } from
+    "lib/openzeppelin-contracts-upgradeable/contracts/security/ReentrancyGuardUpgradeable.sol";
+
+import { MerkleProofUpgradeable } from
+    "lib/openzeppelin-contracts-upgradeable/contracts/utils/cryptography/MerkleProofUpgradeable.sol";
+
+/// @title MerkleClaimable
+/// Contract for managing Taiko token airdrop for eligible users
+abstract contract MerkleClaimable is ReentrancyGuardUpgradeable, OwnableUpgradeable {
```
dantaik, line 19: Lets use EssentialContract as the base

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -0,0 +1,90 @@
...
+        uint64 _claimEnd,
+        bytes32 _merkleRoot
+    )
+        external
+        onlyOwner
+    {
+        _setConfig(_claimStart, _claimEnd, _merkleRoot);
+    }
+
+    function _init() internal {
```
dantaik, line 77: To follow open zepplin's naming conv, please rename it.

dantaik, line 77:
```suggestion
    function _MerkleClaimable_init() internal {
```

packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
```diff
@@ -0,0 +1,101 @@
...
+        view
+        returns (uint256 balance, uint256 withdrawableAmount)
+    {
+        balance = claimedAmount[user];
+        // If balance is 0, or if the timestamp is above the (withdrawWindow + claimEnd) then there
+        // is no more balance and withdrawable amount
+        if (balance == 0 || uint64(block.timestamp) > (claimEnd + withdrawalWindow)) return (0, 0);
+        // Balance might be positive before end of claiming (claimEnd - if claimed already) but
+        // withdrawable is 0.
+        if (uint64(block.timestamp) < (claimEnd)) return (balance, 0);
```
dantaik, line 89:
```suggestion
        if (block.timestamp < claimEnd) return (balance, 0);
```

packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
```diff
@@ -0,0 +1,101 @@
...
+    /// @return withdrawableAmount The amount available to withdraw
+    function getBalance(address user)
+        public
+        view
+        returns (uint256 balance, uint256 withdrawableAmount)
+    {
+        balance = claimedAmount[user];
+        // If balance is 0, or if the timestamp is above the (withdrawWindow + claimEnd) then there
+        // is no more balance and withdrawable amount
+        if (balance == 0 || uint64(block.timestamp) > (claimEnd + withdrawalWindow)) return (0, 0);
```
dantaik, line 86: If the time is way beyond the withdrawal window, then the user can withdraw all its unwithdrawn amount, not 0.

packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
```diff
@@ -0,0 +1,101 @@
...
+        // Balance might be positive before end of claiming (claimEnd - if claimed already) but
+        // withdrawable is 0.
+        if (uint64(block.timestamp) < (claimEnd)) return (balance, 0);
+
+        uint256 timeBasedAllowance =
+            balance * (uint64(block.timestamp) - claimEnd) / withdrawalWindow;
```
dantaik, line 92: still buggy here.

dantaik (changes requested):

packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
adaki2004: Yep, thanks, corrected.

packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
```diff
@@ -0,0 +1,104 @@
...
+        returns (uint256 balance, uint256 withdrawableAmount)
+    {
+        balance = claimedAmount[user];
+        // If balance is 0 then there is no balance and withdrawable amount
+        if (balance == 0) return (0, 0);
+        // Balance might be positive before end of claiming (claimEnd - if claimed already) but
+        // withdrawable is 0.
+        if (block.timestamp < claimEnd) return (balance, 0);
+
+        // Hard cap timestamp - so range cannot go over - to get more allocation over time.
```
dantaik, line 91: block.timestamp.min(claimEnd + withdrawalWindow)

dantaik, line 91: block.timestamp.min(claimEnd + withdrawalWindow)

dantaik (approved):

PR #15247, opened 2023-11-22
fix(protocol): use ERC1967Proxy

dantaik:
The following changes are made in this PR:

1. Use `RC1967Proxy` instead of `TransparentUpgradeableProxy`. `TransparentUpgradeableProxy` has the issue that there is an proxy admin and the contract also has an owner, and the admin and owner cannot be the same address - but we only have one DAO/governor address.
2. Split the original `EssentialContract` into an `OwnerUUPSUpgradable.sol` and a new `EssentialContract` so `AddressManager` is now an `OwnerUUPSUpgradable` contract.
3. Removed all ProxiedXXXContract as all owner contracts must be used behind a proxy. This is enforced by calling `_disableInitializers` in constructor.
4. Simplified test imports to make it more readable.
5. Add airdrop contracts
This PR doesn't have any logic changes in our main contracts besides using RC1967Proxy.

TODO:
- [x] @davidtaikocha: please fix the genesis generation scripts.
- [x] @davidtaikocha: verify packages/protocol/script/DeployOnL1.s.sol and fix issues there
- [x] @dantaik : Deploy Governor contract - A timelock and a governor contract has been deployed and the timelock is now the owner of all Ownerable contracts. @davidtaikocha please test the deployment.


adaki2004:
> `TransparentUpgradeableProxy` has an `admin` address and the logic contract behind it has an `owner` address. These two addresses cannot be the same due to the implementation of `TransparentUpgradeableProxy`. It will be problematic if we assign DAO governor as both the admin and the owner. Please see the `TestEssentialContract.t.sol` tests for details.
> 
> `ERC1967Proxy` doesn't have the protocol level admin, upgradability permission is granted to the logic contract's owner address.
> 
> We shall use `ERC1967Proxy` to replace `TransparentUpgradeableProxy` in all cases. @davidtaikocha

From my perspective having 2 "entities" is fine and indeed desired, why ?
One general change in the contracts: new implementation deployment (via DAO / Proxy Admin)
Other generic change in the contracts : config settings change (like base fee adjustment).

- DAO may need to upgrade the implementation contract via the proxy - but to make it 100% legit, it shall be a `TimeLock` contract. Meaning it will take 7 days (for example) to make those changes effective. So leaving some time for people to exit if they do not like those new design changes. (It is even listen on L2 beat)
However, sometimes we need to act very fast and we cannot wait 7 days, and this is where the admin comes in to scope.

So if one and only "entity" can do these updates, then there is a fear that heavy protocol changes can go through in a second - no time for people to react if they like the change or not.

So having a quick (config change, aka owner) and a slow (impl. change, aka admin) entity makes sense to me.

Brechtpd:
Reading through the docs of `TransparentUpgradeableProxy`, its goal is to fix this potential exploit: https://medium.com/nomic-foundation-blog/malicious-backdoors-in-ethereum-proxies-62629adf3357. Not the most practical one, but a fair enough thing to try to fix.

The solution they propose also seams reasonable: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/330c39b66218de5fa4f0f62d4824ae3c710bb40f/contracts/proxy/transparent/TransparentUpgradeableProxy.sol#L35-L39. But maybe a bit troublesome in practice to have an extra intermediate contract here, I'm not sure?

> From my perspective having 2 "entities" is fine and indeed desired, why ?

In theory you could do both with the same address using the function selector as an extra decision factor, but I think a fair point.

dantaik:
> > `TransparentUpgradeableProxy` has an `admin` address and the logic contract behind it has an `owner` address. These two addresses cannot be the same due to the implementation of `TransparentUpgradeableProxy`. It will be problematic if we assign DAO governor as both the admin and the owner. Please see the `TestEssentialContract.t.sol` tests for details.
> > `ERC1967Proxy` doesn't have the protocol level admin, upgradability permission is granted to the logic contract's owner address.
> > We shall use `ERC1967Proxy` to replace `TransparentUpgradeableProxy` in all cases. @davidtaikocha
> 
> From my perspective having 2 "entities" is fine and indeed desired, why ? One general change in the contracts: new implementation deployment (via DAO / Proxy Admin) Other generic change in the contracts : config settings change (like base fee adjustment).
> 
> * DAO may need to upgrade the implementation contract via the proxy - but to make it 100% legit, it shall be a `TimeLock` contract. Meaning it will take 7 days (for example) to make those changes effective. So leaving some time for people to exit if they do not like those new design changes. (It is even listen on L2 beat)
>   However, sometimes we need to act very fast and we cannot wait 7 days, and this is where the owner comes in to scope.
> 
> So if one and only "entity" can do these updates, then there is a fear that heavy protocol changes can go through in a second - no time for people to react if they like the change or not.
> 
> So having a quick (config change, aka owner) and a slow (impl. change, aka admin) entity makes sense to me.

There are two actors who can change the config/impl of contracts: the **DAO** and the **security council**.  Regardless who performs the change, the changes may either be **proxy upgrade** or **onlyOwner functions**.

I believe it is incorrect to say that the owner can only perform onlyOwner transactions and the DAO can only do contract upgrades.

Brechtpd (approved):
Looks good to me. I wouldn't say ERC1967Proxy is generally better than TransparentUpgradeableProxy because they both have their issues, so depends on what is deemed less bothersome. I'm fine with both.
 
 > I believe it is incorrect to say that the owner can only perform onlyOwner transactions and the DAO can only do contract upgrades.
 
Not sure I understand what you mean with this. Isn't that the two roles we have on the protocol level? In practice they may overlap, but if the two roles are not seen as separate than the protocol should also just see them as one no?

dantaik:
> sn't that the two roles we have on the protocol level

The protocol assume there is only one role: the owner, which can perform `onlyOwner` ops and upgrade the proxy's impl.

cyberhorsey:
@dantaik proxy_scripts has been merged into this PR, i think it can be merged into base branch now if @davidtaikocha approves the rest of this PR.

davidtaikocha:
> @dantaik proxy_scripts has been merged into this PR, i think it can be merged into base branch now if @davidtaikocha approves the rest of this PR.

I will do the integration test in internal devnet today, and after that, will approve this PR.

davidtaikocha (approved):

PR #15235, opened 2023-11-20
feat(eventindexer): Update eventindexer post hooks, add AssigmentHook to deploy script, update chart apis

cyberhorsey:
Hooks has changed the functionality of th protocol quite a bit ,and this eventindexer needs updates.

This PR:
1) Adds `AssignmentHook` to `DeployOnL1.s.sol`, otherwise it can not be tested with the deploy with this code, nor can it be tested in TaikoClient. @dantaik @adaki2004 please review this part.
2) Updates bindings
2) Moves `averageProofReward` calculations to the `AssignmentHook.BlockAssigned` event, away from `LibProposing.BlockProposed` event.
3) Calculates proof rewards as an array, unique by `feeTokenAddress`, as they can now be paid in any token.
4) Updates the time-series generator to generate multiple charts for proof rewards, one for each feeTokenAddress used.
5) Removes `proposer reward` calculations.
6) Updates chart API, so the frontend stats page can now display a different chart for each proof reward, separated by feeTokenAddress.

PR #15249, opened 2023-11-22
refactor(protocol): improve contract deployment script

dantaik:
The following changes are made in this PR:

1. Use `RC1967Proxy` instead of `TransparentUpgradeableProxy`. `TransparentUpgradeableProxy` has the issue that there is an proxy admin and the contract also has an owner, and the admin and owner cannot be the same address.
2. Split the original `EssentialContract` into an `OwnerUUPSUpgradable.sol` and a new `EssentialContract` so `AddressManager` is now an `OwnerUUPSUpgradable` contract.
3. Removed all ProxiedXXXContract as all owner contracts must be used behind a proxy. This is enforced by calling `_disableInitializers` in constructor.
4. Simplified test imports to make it more readable.

This PR doesn't have any logic changes in our main contracts besides using RC1967Proxy.

TODO:
- [ ] David: please fix the genesis generation scripts.
- [ ] David: verify packages/protocol/script/DeployOnL1.s.sol and fix issues there
- [ ] Daniel: Deploy Governor contract


cyberhorsey:
@dantaik note that in this pr: https://github.com/taikoxyz/taiko-mono/pull/15235
I have added `AssignmentHook` to this deploy script as well.

PR #15250, opened 2023-11-23
feat(protocol): Add native ERC20 token support 'plugin' to ERC20Vault via Registry+Translator

adaki2004:
So, this seemingly a big PR but it is NOT !

Let me summarize the things and it will be more understandable.

### Why cannot we just bring the USDC contracts (from [taiko/USDC](https://github.com/taikoxyz/USDC/pull/1)) to this repo's contract folder ? And why I use those contracts only in `test/helper/usdc` ?

Because those are not identical (modified contracts only for E2E tests). The reason is the compiler settings (solc version and optimization). 
Here (in the tests) i can change the pragma version (to 8.20) but 'in real life' we have to stick to the old USDC contract compiler versions and optimizations.

You **do not have to bother** with the contracts in `test/helper/usdc`, because they are the official USDC contracts - slightly modified to work with .8.20, see mods:
There are some things deprecated with higher version of the compiler, which is still relevant in circle’s USDC code (lack of receive does not trigger a warning, constructor visibilty,`.now`, `.add` -> BUT in the `test/helper/usdc` folder, it is fine, i modified these files to be working with .8.20 but the Circle’s Bridged USDC Standard is clear on that, we shall **NOT** modify the code and have the exact same byte code.
https://github.com/circlefin/stablecoin-evm/blob/master/doc/bridged_USDC_standard.md

So this is why, during testing, I used these modified USDC contracts - to have an E2E tests with our ERC20Vault.

## Ok so now the changes:

- introduced an `ERC20Hook` contract, deployable - which will handle the 'mappings' native token vs. deployed token on L2.
- In particular, we need to keep track of some data:
```
    struct DeployedToCanoincalData {
        address canonicalAddress;
        uint8 burnFunctionSignature;
    }
    mapping(address => address) public canonicalToPredeployed;
    mapping(address l2Address => DeployedToCanoincalData) public predeployedToCanonical;
```
1. `canonicalToPredeployed` is needed in `_getOrDeployBridgedToken()`, so when we we check if a custom contract is deployed already for a specific native coin (in our case/example it is USDC).
2. `predeployedToCanonical` is needed in `_handleMessage`, so when we "birdge back" to the canonical chain the already bridged USDC. Because we "minted" during the bridging, we now need to burn. But the burn function differs from our "BridgedERC20" standard - we need to keep that in mind. (Only minterRole can burn and only in case if minterRole address owning those tokens).

This was the easiest design coming to my mind, with hooks, because maybe in the future we can support other native tokens (USDT), the only thing we need to do is pre-deploy a USDT the same way, check its' burn function and add it to the interface if different.

Referring to this doc too:
https://github.com/taikoxyz/USDC/blob/0d7d65af85ce59b5ae8452b15d125159f21555e4/README.md

Please let me know if you have any questions or suggestions !


Brechtpd:
For the different solc versions, could this help: https://ethereum.stackexchange.com/a/153746?

adaki2004:
> For the different solc versions, could this help: https://ethereum.stackexchange.com/a/153746?

Tried :( Not working. And one other issue is the optimizer settings - which differs for USDC too.

kép

Brechtpd:
Two ideas:
- Instead of modifying the behavior of ERC20Vault, maybe better to create a separate USDCTranslator contract that handles the differences. So ERC20Vault -> USDCTranslator (which implements IMintableERC20) -> USDC (custom smart contract). This seems cleaner and more flexible because then these translator contracts can do whatever is needed to make the custom contracts conform to the ERC20Vault behavior.
- Rename ERC20Hook to ERC20Regsitry so it's more clear what its function is

adaki2004:
> 
>     * Rename ERC20Hook to ERC20Regsitry so it's more clear what its function is

Done, thanks for the feedback.

>     * Instead of modifying the behavior of ERC20Vault, maybe better to create a separate USDCTranslator contract that handles the differences. So ERC20Vault -> USDCTranslator (which implements IMintableERC20) -> USDC (custom smart contract). This seems cleaner and more flexible because then these translator contracts can do whatever is needed to make the custom contracts conform to the ERC20Vault behavior.

An observations to this - as i started implementing. We can definitely do that (i think) but it comes with the fact that this 'USDCTranslator' needs to be:
- MinterRole given (which is OK, we just grant it minter role, it is OK)
- Needs to be an 'approved()' address -> This contract will be the one (and not the ERC20Vault!!) who need to be approved for bridging back from L2 to L1.  So the UI also needs extra logic to check to whom 'approve()'... Because this wrapper/translator will be the one responsible for transferFrom().

I'm implementing in the meantime, will try to do it within today, adjust code+tests, and you can decide if the flexibility worth the effort of breaking the 'way' we bridge (hence some UI logic changes).

adaki2004:
>  Instead of modifying the behavior of ERC20Vault, maybe better to create a separate USDCTranslator contract that handles the differences. So ERC20Vault -> USDCTranslator (which implements IMintableERC20) -> USDC (custom smart contract). This seems cleaner and more flexible because then these translator contracts can do whatever is needed to make the custom contracts conform to the ERC20Vault behavior.

Funcional wise it is done. Now every native ethereum asset could be plugged in to the system - if we want to support it (stEth, USDT, USDC, etc.) - just need to predeploy their current 'up-to-date' smart contract from ethereum (with their respective proxies) onto L2 with genesis (or right after the genesis) and then update the `ERC20NativeRegistry`.

Please have a look at the design diagram and then check the code ! I tried to put explanations everywhere possible.

Here is a high level design diagram about the usage.

@dantaik , @Brechtpd i advise to start from the diagram and see points in order, 1,2,3,4.. etc. to get to know the flow.

kép

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -87,6 +88,21 @@ contract ERC20Vault is BaseVault {
     error VAULT_MESSAGE_NOT_FAILED();
     error VAULT_MESSAGE_RELEASED_ALREADY();
 
+    /// @dev If a native token issuer (e.g.: Circle/Lido) revokes minter role from our contracts
+    /// (once they get the ownership), we won't be able to mint / burn the tokens. But we still want
+    /// to mint these tokens like "USDC ⭀31337" style, but then we need to reset the old mapping.
```
Brechtpd, line 93: Last part  about `"USDC ⭀31337" style` could use some more information because I don't really understand what it means from reading this.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -227,8 +243,18 @@ contract ERC20Vault is BaseVault {
 
         // If it's a bridged token
         if (bridgedToCanonical[token].addr != address(0)) {
+            // erc20NativeRegistry shall be address(0) on L1 and CAN be address(0) on L2 too, if
+            // there is no native support
```
Brechtpd, line 247: Would be cool if these changes would not be needed, but some things are not totally clear yet to me so I don't know.

Brechtpd (commented):
> Needs to be an 'approved()' address -> This contract will be the one (and not the ERC20Vault!!) who need to be approved for bridging back from L2 to L1. So the UI also needs extra logic to check to whom 'approve()'... Because this wrapper/translator will be the one responsible for transferFrom().

Could you work around this by calling back into ERC20Vault from the translator contract so that ERC20Vault still does the transfers requiring approval? Seems like it could be possible in a safe way because msg.sender would the the translator contract and ERC20Vault would know the token contract that is linked to that translator? Not saying it's a good idea, but seems like there are could be potential solutions that do not require this to be a special case.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
dantaik: the name and symbol has been changed in another PR. This ⭀ is removed.

adaki2004: > Last part about `"USDC ⭀31337" style` could use some more information because I don't really understand what it means from reading this.

It was just an example. (And we used these namings for `BridgedERC20` assets)
So the situation (I tried to explain) as follows:
1. Taiko predeploys USDC on L2.
2. People bridging USDC into that contract (via the translator obv.)
3. The ownership of USDC on L2 will be transferred eventually to Circle.
4. Circle for whatever reason decides to cut the translator/tokenvault's roles (minter role).

In this case we need to reset the mapping too in the registry, but also has to reset this `canonicalToBridged[chainId][canonicalAddrToReset]`, otherwise we will not be able to bridge USDC anymore (like` USDC ⭀31337` as our `BridgedERC20` token, hence the mapping points to the contract for whcih we has no minter rights anymore.)

packages/protocol/contracts/tokenvault/ERC20Vault.sol
adaki2004: Please see this: https://github.com/taikoxyz/taiko-mono/pull/15250#discussion_r1405134515

adaki2004:
> > Needs to be an 'approved()' address -> This contract will be the one (and not the ERC20Vault!!) who need to be approved for bridging back from L2 to L1. So the UI also needs extra logic to check to whom 'approve()'... Because this wrapper/translator will be the one responsible for transferFrom().
> 
> Could you work around this by calling back into ERC20Vault from the translator contract so that ERC20Vault still does the transfers requiring approval? Seems like it could be possible in a safe way because msg.sender would the the translator contract and ERC20Vault would know the token contract that is linked to that translator? Not saying it's a good idea, but seems like there are could be potential solutions that do not require this to be a special case.

Well, i tried with delegateCall() + removing the nativeContractAddress from the translators but so far no success. Will need to revisit on Monday.

dantaik:
Somehow I found my comments have not been submitted successfully (or I submitted it somewhere else):

- If Circle requires their contract to be deployed bytecode as-is, we should simply copy their bytecode to this contract without creating our own https://github.com/taikoxyz/USDC repo.
- We should avoid integration testing with USDC and other future third-party tokens, and focus on unit-testing our own code. I suggest to completely remove USDC related code from our mono repo.
- I suggest to rename Translator to Adaptor as this is what it is from a design-pattern perspective.

adaki2004:
> Somehow I found my comments have not been submitted successfully (or I submitted it somewhere else):

Yes, you asked them in the Airdrop branch, but it is ok. :)

>     * If Circle requires their contract to be deployed bytecode as-is, we should simply copy their bytecode to this contract without creating our own https://github.com/taikoxyz/USDC repo.

Agreed.

>     * We should avoid integration testing with USDC and other future third-party tokens, and focus on unit-testing our own code. I suggest to completely remove USDC related code from our mono repo.

Well, the addition to they layer above (aka with Translators/Adaptors) means that some tokens have custom burn/mint functions. Like we see with Circle USDC's burn. But i'll try to shrinken the contracts and make it less, but only once i reached the 'final' version. (Only missing piece is: make the adatpor calls delegate so that the ERC20Vault is the "approved" contract not the translator - for native tokens).

>     * I suggest to rename Translator to Adaptor as this is what it is from a design-pattern perspective.

Ok.

adaki2004:
@Brechtpd @dantaik 
Functional wise it is ready ! Please have a look if something is not OK or clear.

Missing things:

- [ ] Bring USDC bytecode (from other repo) here and incorporate into the L2 deployment process
- [ ] Mock actual USDC code under `/test/helper/usdc` into a smaller codebase
- [ ] Use other proxy implementation for registry and vault

adaki2004:
As opened a new on top of branch `proxy`, closing this.

PR #15257, opened 2023-11-25
(DEBUG BRANCH) Approve shall be applied on ER20Vault not translator

adaki2004:
@Brechtpd 
This is where i'm at.

I use the command for testing. For whatever reason the 'minting' is not working.  (You can clearly see my debug prints :))
`forge test --match-test test_20Vault_bridge_native_usdc_back_to_l1 -vv`

Any idea what am i doing wrong with the delegateCall.. ?

### DIFF
--- a/packages/protocol/contracts/tokenvault/erc20/translators/UsdcTranslator.sol
+++ b/packages/protocol/contracts/tokenvault/erc20/translators/UsdcTranslator.sol
@@ -10,10 +10,14 @@ import "lib/openzeppelin-contracts-upgradeable/contracts/token/ERC20/ERC20Upgrad
 
 import "./BaseTranslator.sol";
 
-interface IUsdcBurn {
+interface IUsdc {
     /// @notice Burns `amount` tokens from address(this)
     /// @param amount The amount of tokens to burn.
     function burn(uint256 amount) external;
+    /// @notice Mint `amount` tokens to `to`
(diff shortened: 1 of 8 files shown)


packages/protocol/contracts/tokenvault/erc20/translators/UsdcTranslator.sol
```diff
@@ -27,32 +31,46 @@ contract UsdcTranslator is BaseTranslator {
...
     /// @param amount The amount of tokens to mint.
-    function mint(address account, uint256 amount) public onlyFromNamed("erc20_vault") {
-        IERC20TokenVaultRelayer(nativeTokenContract).mint(account, amount);
+    function mint(
+        address token,
+        address account,
+        uint256 amount
+    )
+        public
+        onlyFromNamed("erc20_vault")
```
Brechtpd, line 40: I think you can just remove this check, the important things is that msg.sender will be the ERC20Vault for the USDC token contract, so only delegate calls done by ERC20Vault will have that property.

adaki2004:
https://ethereum.stackexchange.com/questions/125217/delegatecall-failing -> Tried, not helped.

adaki2004:
Storage layout:
kép

adaki2004:
Storage layout now - and still, `delegatecall()` does not resolve the mint function properly.
kép

PR #15261, opened 2023-11-26
feat(protocol): Make Governor and TimelockControler upgradable

dantaik:


packages/protocol/script/DeployOnL1.s.sol
```diff
@@ -165,17 +164,26 @@ contract DeployOnL1 is Script {
             owner: timelock
         });
 
-        address governor = address(new TaikoGovernor(IVotes(taikoToken), _timelock));
-        console2.log("governor: ", governor);
+        address governor = LibDeployHelper.deployProxy({
```
Brechtpd, line 167: Any reason for the governor to be upgradeable?

dantaik: No particular reason except for avoiding launching a new DAO before adopting new governor modules/features

Brechtpd: Just thinking it should be pretty easy to switch the governor out right? But yeah doesn't matter much!

Brechtpd (approved):

PR #15263, opened 2023-11-27
docs(website): update list of socials 

sillybillyy:
Since there is an Taiko Lab's Official Account on LinkedIn, Suggested to update Taiko's List of Socials to confirm legitimacy and build follower in preparation once in need of any recruitment in the future.

also I'm thinking of notion as well (ref:https://taikoxyz.notion.site/Taiko-Jobs-828fd7232d2c4150a11e10c8baa910a2)
but I think that the url might be change so up to discussion but the main url (https://taikoxyz.notion.site/Taiko-Jobs) isn't practically work at all and need of registration instead.


dantaik:
People are still using LinkedIn?!!!

macvts:
better than nothing xD
Utilizing those existing never wasted 😇

PR #15275, opened 2023-11-27
feat(protocol): Native token support

adaki2004:
Same as: https://github.com/taikoxyz/taiko-mono/pull/15250 but rebased to proxy branch.

Changes are the native token support plugin changes. (+the bug from ERC20Vault's onMessageRecall also incorporated)

High level design overview:
285492728-868649bb-63cc-408d-ac0b-288cfc56de99

Known missing things:

- [ ] Bring USDC bytecode (from other repo) here and incorporate into the L2 deployment process (not strictly related to this feature, but in general to USDC support)
- [x] Mock actual USDC code under `/test/helper/usdc` into a smaller codebase


Brechtpd (commented):
It feels like it may be worth it to also let the standard tokens go through an adapter that simply implements the BaseAdapter in the simplest way. This will reduce the number of different code paths so it should be easier and more straightforward to test.

packages/protocol/contracts/tokenvault/erc20/adapters/UsdcAdapter.sol
```diff
@@ -0,0 +1,79 @@
...
+        ERC20Upgradeable(token).transferFrom({
+            from: account,
+            to: address(this), //ERC20Vault - via delegatecall()
+            amount: amount
+        });
+        IUsdc(token).burn(amount);
+    }
+
+    /// @notice We need to keep this because tokenVault will use the SafeERC20Upgradeable's
+    /// safeTransfer, which eventuall calls into this.
```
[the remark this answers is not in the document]
adaki2004: No - these can be deleted ! (And deleting them with the next commit) because the bridged tokens (be it native or not) are not subject to `transfer` or `transferFrom` (except USDC but that is a special case and we handle via the `burn()` function as that one mandates these adapter contracts).

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -86,6 +87,23 @@ contract ERC20Vault is BaseVault {
     error VAULT_INVALID_SRC_CHAIN_ID();
     error VAULT_MESSAGE_NOT_FAILED();
     error VAULT_MESSAGE_RELEASED_ALREADY();
+    error VAULT_ERROR_NATIVE_BURN();
+    error VAULT_ERROR_NATIVE_MINT();
+
+    /// @dev If a native token issuer (e.g.: Circle/Lido) revokes minter role from our contracts
+    /// (once they get the ownership), we won't be able to mint / burn the tokens. But we still want
+    /// to mint these tokens like "USDC ⭀31337" style, but then we need to reset the old mapping.
```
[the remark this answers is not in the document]
adaki2004: Okay, so i'll try to explain the necessity of it via an example!
1. We predeploy the USDC contracts to L2. We call the registry's `changeCustomToken()` to create a mapping between USDC (on L1) and USDC (on. L2).
2. Eventually, when bridge is "safe", Circle needs to have the ownership over the L2 USDC contract(s).
3. For whatever reason (hack in bridge, no trust in Taiko, rival bridges, etc.) Circle will revoke the minter role from our ERC20Vault.

After this, Bob wants to bridge USDC from L1 to L2 via our bridge contract, but it will revert since we do not have minter role anymore. But he also not able to bridge the token into a newly created (BridgedERC20 token named for example `"USDC ⭀31337"` -> This is just a nameing convention we used with BridegdERC20 tokens, this is why i said this in the above example) because such token cannot be created hence the `canonicalToBridged[chainId][canoncalAddr]` will contain the old address.. So we need a mechanism to wipe this data away -> to support further the bridging of the USDC (even tho not 'natively' but via `BridgedERC20`)

Let me know if it makes sense, or not.

dantaik:
@adaki2004 I thought you are going to remove all USDC related code from our tests so we can have pure unit tests for our own code. Aren't you?

adaki2004:
> @adaki2004 I thought you are going to remove all USDC related code from our tests so we can have pure unit tests for our own code. Aren't you?

I'm not saying removing (because we need to test the integration of USDC especially the `burn()` which is a special USDC "feature"- hence we doing this branch) but reducing the files.

Currently we have 20-25 files -> What if I flatten them to 2 files ? (1 for proxy, 1 for implementation).

adaki2004:
> @adaki2004 I thought you are going to remove all USDC related code from our tests so we can have pure unit tests for our own code. Aren't you?

@dantaik Please see the latest commit - and file-level changes.

I think it is fine having circle contracts as 2 (flattened) files because of several reasons:
- We testing for the actual contracts (integration) - and since we need adapters to adapt custom native tokens, we need to test the custom behaviour (which Circle case is the `burn()`).
- All these files (tho in the tests i named them USDC) relevant to all stable coins circle is issuing.
From Circle:
`"...a bridged USDC token contract to Circle follows a set procedure that's also extensible to bridged EURC."`

adaki2004:
@davidtaikocha Could you please check first from deployment point of view ? :) Especially the USDC Native one.
I'm pretty sure there are things i missed, but i tried to query the contracts storage.
Weird is the proxy does not have any storage (?) then how it links to the implementation.. ? 

Here is what i got after the commands:
kép

packages/protocol/contracts/tokenvault/ERC20Vault.sol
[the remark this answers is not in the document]
Brechtpd: Instead of saying "But we still want to mint these tokens like "USDC ⭀31337" style", wouldn't it more clear to just say that when that happens it still needs to be possible to bridge tokens using the standard `BridgedERC20`?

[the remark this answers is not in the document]
adaki2004: Yes that would be the ideal i was thinking about it too! :) 
But it requires all the time to “query” if we can bridge native way or not, which is expensive i think. (Bc of that canonicalToBridge mapping for that canonical “occupied” with an address which is not mintable).
Do you have any ideas with which we could do it a cheap and efficient way ? 🤔

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -227,8 +263,26 @@ contract ERC20Vault is BaseVault {
...
+            }
+            // Depending on native token (or not), it handles the burn accordingly.
+            _executeBurnOrMint(
+                adapter != address(0) ? adapter : token,
+                token,
+                msg.sender,
+                amount,
+                adapter != address(0),
+                false
+            );
```
Brechtpd, line 283: Very similar code as in ` onMessageRecalled`, I think best if it is shared in a single other function.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -300,4 +375,49 @@ contract ERC20Vault is BaseVault {
...
+        } else {
+            (bool success,) = adapter.delegatecall(
+                abi.encodeWithSignature(
+                    "burn(address,address,uint256)", tokenContract, toOrFrom, amount
+                )
+            );
+
+            if (!success) {
+                revert VAULT_BURN_ERROR();
+            }
```
Brechtpd, line 420: More code reuse when just storing `"mint(address,address,uint256)"`/`"burn(address,address,uint256)"` in a string based on `isMint`. Then all other code could be shared.

packages/protocol/contracts/tokenvault/erc20/adapters/BaseAdapter.sol
```diff
@@ -0,0 +1,41 @@
...
+
+    /// @notice Burns `amount` tokens from the `from` address.
+    /// @param token The erc20 token contract.
+    /// @param from The account from which the tokens will be burned.
+    /// @param amount The amount of tokens to burn.
+    function burn(address token, address from, uint256 amount) external;
+}
+
+/// @title BaseAdapter
+/// @notice It serves as a wrapper between the deployed USDC and the ERC20Vault - an extra layer for
```
Brechtpd, line 28:
```suggestion
/// @notice It serves as a wrapper between the token and the ERC20Vault - an extra layer for
```

packages/protocol/contracts/tokenvault/erc20/adapters/BaseAdapter.sol
```diff
@@ -0,0 +1,41 @@
...
+/// flexibility. It is not an ERC20 contract, but we need to implement the interfaces the ERC20Vault
+/// calls and relay over to other (native) contracts. We do not even need this contract to be
+/// initialized, because we need the implementation of it, the EssentialContract is only because of
+/// storage compatibility with ERC20Vault.
```
Brechtpd, line 32: Does it need to be storage compatible? Seems safer if the contract simply doesn't touch storage at all.

packages/protocol/contracts/tokenvault/erc20/adapters/BaseAdapter.sol
```diff
@@ -0,0 +1,41 @@
...
+abstract contract BaseAdapter is EssentialContract, IERC20TokenVaultRelayer {
+    uint256[50] private __gap;
```
Brechtpd, line 34: Does it need to be a `EssentialContract`? Seems safer to be not to have `uint256[50] private __gap;` in here of do any storage writes.

packages/protocol/contracts/tokenvault/erc20/adapters/BaseAdapter.sol
```diff
@@ -0,0 +1,41 @@
...
+
+    /// @notice Initializes the contract.
+    /// @param _addressManager The address manager
+    function init(address _addressManager) external initializer {
+        EssentialContract._Essential_init(_addressManager);
+    }
```
Brechtpd, line 40: Can be removed?

packages/protocol/contracts/tokenvault/erc20/adapters/UsdcAdapter.sol
```diff
@@ -0,0 +1,53 @@
...
+
+/// @title UsdcAdapter
+/// @notice It serves as a wrapper between the deployed USDC and the ERC20Vault - an extra layer for
+/// flexibility. It is not an ERC20 contract, but we need to implement the interfaces the ERC20Vault
+/// calls and relay over to the USDC contract. The reason this contract needs to be 'custom' is the
+/// different mechanisms how each native tokens handling some functions. In this case the
+/// 'troublemaker' is the burn function which behaves differently (in USDC) than usual.
+contract UsdcAdapter is BaseAdapter {
+    /// @notice Mints tokens to an account.
+    /// @dev Only an ERC20Vault can call this function.
```
Brechtpd, line 31: Can be called by any contract, so best to remove this I think.

packages/protocol/contracts/tokenvault/erc20/adapters/UsdcAdapter.sol
```diff
@@ -0,0 +1,53 @@
...
+    /// @param account The account to mint tokens to.
+    /// @param amount The amount of tokens to mint.
+    function mint(address token, address account, uint256 amount) public {
+        IUsdc(token).mint(account, amount);
+    }
+
+    /// @notice Burns tokens from an account.
+    /// @dev Only an ERC20Vault can call this function.
```
Brechtpd, line 39: Can be called by any contract, so best to remove this I think.

packages/protocol/contracts/tokenvault/erc20/adapters/BridgedERC20Adapter.sol
```diff
@@ -0,0 +1,33 @@
...
+
+import "./BaseAdapter.sol";
+import "../../IMintableERC20.sol";
+
+/// @title BridgedERC20Adapter
+/// @notice It serves as an extension for the ERC20Vault to handle BridgedERC20 tokens in a same
+/// fashion as it handles native tokens, so the interface can be unified.
+contract BridgedERC20Adapter is BaseAdapter {
+    /// @notice Mints tokens to an account.
+    /// @dev Only an ERC20Vault can call this function.
```
Brechtpd, line 19: Can be called by any contract, so best to remove this I think.

packages/protocol/contracts/tokenvault/erc20/adapters/BridgedERC20Adapter.sol
```diff
@@ -0,0 +1,33 @@
...
+    /// @param account The account to mint tokens to.
+    /// @param amount The amount of tokens to mint.
+    function mint(address token, address account, uint256 amount) public {
+        IMintableERC20(token).mint(account, amount);
+    }
+
+    /// @notice Burns tokens from an account.
+    /// @dev Only an ERC20Vault can call this function.
```
Brechtpd, line 27: Can be called by any contract, so best to remove this I think.

packages/protocol/contracts/tokenvault/erc20/registry/ERC20NativeRegistry.sol
```diff
@@ -0,0 +1,104 @@
...
+    event Erc20CustomDeleted(address l1Address, address l2Counterpart);
+
+    error ERC20_HOOK_INVALID_ADDRESS();
+    error ERC20_HOOK_INVALID_BURN_SIGNATURE_ID();
+
+    function init(address _addressManager) external initializer {
+        EssentialContract._Essential_init(_addressManager);
+    }
+
+    /// @notice Adds a custom, pre-deployed contract mapping to it's L1 (parent chain) counterpart
```
Brechtpd, line 46:
```suggestion
    /// @notice Adds a custom, pre-deployed contract mapping to its L1 (parent chain) counterpart
```

packages/protocol/contracts/tokenvault/erc20/registry/ERC20NativeRegistry.sol
```diff
@@ -0,0 +1,104 @@
...
+    /// AddressManager
+    /// @param deleteCustom Indicates deleting both of the mappings
+    /// @param chainId ChainId the canonical originates from
+    function changeCustomToken(
+        address l1Address,
+        address deployedCounterpart,
+        bytes32 adapterName,
+        uint64 chainId,
+        bool deleteCustom
+    )
```
Brechtpd, line 60: Feels a bit strange to have a single function with a `deleteCustom` variable while it doesn't really share any code. Why not create two functions, one for adding and one for removing?

packages/protocol/contracts/tokenvault/erc20/adapters/BaseAdapter.sol
adaki2004: No, it does not. :) At one point i planned it differently (deployable behind a proxy + callbacks) and is kind of a leftover, removing now.

dantaik:
@adaki2004 given that we can upgrade bridged tokens to local tokens, I'll close this one for now. Lets test the native token support in A6 without pre-deploying the USDC token.

PR #15288, opened 2023-11-29
feat(protocol): allow changing Bridged ERC20 token on dest chain

dantaik:
Now for any canonical token, we allow the bridged token address (`btokenOld`) to be updated to another address (`btokenNew`). Once this is done, `btokenOld` will be in a outbound migration mode, and `btokenNew` will be in a inbound mode (normal operations will still work).

This is to enable our Bridge/Vaults to handle USDC Native Token Standard.


packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -70,22 +79,68 @@ contract BridgedERC20 is
...
+
     /// @notice Mints tokens to an account.
     /// @dev Only an ERC20Vault can call this function.
     /// @param account The account to mint tokens to.
     /// @param amount The amount of tokens to mint.
-    function mint(address account, uint256 amount) public onlyFromNamed("erc20_vault") {
+    function mint(address account, uint256 amount) public nonReentrant whenNotPaused {
+        if (migratingFrom == address(0)) {
+            if (msg.sender != resolve("erc20_vault", true)) revert RESOLVER_DENIED();
+            emit Transfer(address(0), account, amount);
```
adaki2004, line 120: This even is emitted by the mint() function.

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -70,22 +79,66 @@ contract BridgedERC20 is
...
+        migratingFrom = address(0);
+        emit Migration(migratingFrom, migratingTo);
+    }
+
     /// @notice Mints tokens to an account.
-    /// @dev Only an ERC20Vault can call this function.
     /// @param account The account to mint tokens to.
     /// @param amount The amount of tokens to mint.
-    function mint(address account, uint256 amount) public onlyFromNamed("erc20_vault") {
+    function mint(address account, uint256 amount) public nonReentrant whenNotPaused {
```
adaki2004, line 117: So with this function what we basically say is:
- We can mint (into this token) if there is no incoming migration AND caller is the ERC20Vault
OR
- We can mint if there is an incoming migration (into this contract) AND the old contract is calling it.

But this is the "new" contract, no ? Shall not we allow incoming mints from the ERC20Vault, too ?

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -70,22 +79,66 @@ contract BridgedERC20 is
...
+        if (migratingFrom == address(0)) {
+            if (msg.sender != resolve("erc20_vault", true)) revert RESOLVER_DENIED();
+            emit Transfer(address(0), account, amount);
```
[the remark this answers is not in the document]
adaki2004: > Sorry I don't get it, can you elaborate a bit more?

Just meant, the `emit Transfer() `event is a duplicate if we emit it here too, because the `_mint()` function emits it already.

packages/protocol/contracts/tokenvault/BridgedERC20.sol
dantaik: updated the logics, please review again.

packages/protocol/contracts/tokenvault/BridgedERC20.sol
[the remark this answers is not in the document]
dantaik: Fixed

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -70,19 +79,64 @@ contract BridgedERC20 is
         srcDecimals = _decimals;
     }
 
+    function startOutboundMigration(address to)
```
adaki2004, line 82: We do not have `stopOutboundMigration()` -> But i guess that is fine-> most prob. not necessary, if we close the `stopInboundMigration` in the new contract.

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -70,19 +79,64 @@ contract BridgedERC20 is
...
+        external
+        nonReentrant
+        whenNotPaused
+        onlyFromNamed("erc20_vault")
+    {
+        if (migratingTo != address(0)) revert BRIDGED_TOKEN_PERMISSION_DENIED();
```
adaki2004, line 88:
```suggestion
        if (migratingTo != address(0)) revert BRIDGED_TOKEN_MIGRATION_ONGOING();
```

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -70,19 +79,64 @@ contract BridgedERC20 is
...
+        emit Migration(migratingFrom, migratingTo);
+    }
+
+    function startInboundMigration(address from)
+        external
+        nonReentrant
+        whenNotPaused
+        onlyFromNamed("erc20_vault")
+    {
+        if (migratingFrom != address(0)) revert BRIDGED_TOKEN_PERMISSION_DENIED();
```
adaki2004, line 101:
```suggestion
        if (migratingTo != address(0)) revert BRIDGED_TOKEN_MIGRATION_ONGOING();
```

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -70,19 +79,64 @@ contract BridgedERC20 is
...
+        if (from == address(0)) revert BRIDGED_TOKEN_INVALID_PARAMS();
+
+        migratingFrom = from;
+        emit Migration(migratingFrom, migratingTo);
+    }
+
+    function stopInboundMigration() external nonReentrant whenNotPaused onlyOwner {
+        if (migratingFrom == address(0)) revert BRIDGED_TOKEN_PERMISSION_DENIED();
```
adaki2004, line 109:
```suggestion
       if (migratingTo != address(0)) revert BRIDGED_TOKEN_MIGRATION_ONGOING();
```

packages/protocol/contracts/tokenvault/adaptors/USDCAdaptor.sol
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        usdc.mint(account, amount);
+    }
+
+    /// @dev Warning:
+    /// 1) Users must set up allowances for this adaptor so USDC can be transferred from users
+    /// addresses to the adaptor.
+    /// 2) This adaptor must be granted the correct role by the USDC contract in order to burn
+    /// native USDC tokens.
```
adaki2004, line 45: I think it is very convenient (for the Bridge team) to ask for `approve()` on the `ERC20Vault` - in general. If we introduce deployable, standalone contracts (like this), it might make their life easier to always query which contract to be 'approved' - especially if we will see lots of native support.

So what i'd do (and i know it only bc. Brecht advised me on the other branch) is exactly the same as you doing/implementing here, but this contract could be an "implementation only" contract, without being `EssentialContract`. -> A "stateless" delegate contract.

And the ERC20Vault simply would delegatecall into this contract - so basically this contract just an "extension" of that ERC20Vault - and acting on behalf of that.
See here:
https://github.com/taikoxyz/taiko-mono/blob/081c1b9ac23ac287d0ba06fa9e9474780c26a5b8/packages/protocol/contracts/tokenvault/ERC20Vault.sol#L405

adaki2004 (approved):
LGTM

davidtaikocha (approved):

PR #15333, opened 2023-12-05
chore(relayer): add Makefile for relayer

fenghaojiang:
chore: add Makefile for relayer


cyberhorsey:
I don't think this project requires a Makefile as building the relayer with go build is quite standard and the makefile doesn't add any reduction in complexity.

PR #15335, opened 2023-12-05
fix(bridge-ui-v2): configuredCustomTokens in file/document is inconsistency

conanxyz:
I attempted to deploy the bridge frontend based on the documentation and noticed a discrepancy between the configuredCustomToken and the file name.
```
cp config/sample/configuredCustomTokens.example config/configuredCustomTokens.json
```
The README document here uses 'configuredCustomTokens.example,' while in the actual 'config' directory, it is 'configuredCustomToken.example.' I believe it should be changed to 'configuredCustomTokens.example,' and the references in the code should be updated also


KorbinianK (approved):
Good spot!

davidtaikocha (approved):

PR #15340, opened 2023-12-06
refactor(protocol): Variable Declaration Updates

codeesura:
Changed `const` declarations to `let` for objects that are reassigned, ensuring better compliance with JavaScript best practices and improving the flexibility of our code.


davidtaikocha:
According to JavaScript best practice, for `Object`s, I think we should `const` at those places.

PR #15348, opened 2023-12-07
feat(bridge-ui-v2): Update encoded signal proof in BridgeProver.ts

xiaodino:
Tested with L2a and L2b.
I haven't tested multiple hops, which is L2a <-> L2b.

Screen Shot 2023-12-07 at 11 20 30 PM


davidtaikocha (approved):

packages/bridge-ui-v2/scripts/vite-plugins/generateChainConfig.ts
```diff
@@ -86,6 +86,8 @@ async function storeTypes(sourceFile: SourceFile) {
     members: [
       { name: 'L1', value: 'L1' },
       { name: 'L2', value: 'L2' },
+      { name: 'L2a', value: 'L2a' },
```
KorbinianK, line 89: If we do this we "hardcode" our layers again.

packages/bridge-ui-v2/src/libs/bridge/types.ts
```diff
@@ -9,6 +9,7 @@ export enum MessageStatus {
   RETRIABLE,
   DONE,
   FAILED,
+  NewOnlyOwner,
```
KorbinianK, line 12: Where does this status come from? MessageStatus in the contract is unchanged and this should mirror it.

packages/bridge-ui-v2/src/libs/chain/types.ts
```diff
@@ -3,6 +3,8 @@ export type ChainID = bigint;
 export enum LayerType {
   L1 = 'L1',
   L2 = 'L2',
+  L2a = 'L2a',
+  L2b = 'L2b',
```
KorbinianK, line 7: Same as above

packages/bridge-ui-v2/scripts/vite-plugins/generateChainConfig.ts
```diff
@@ -151,6 +153,8 @@ async function buildChainConfig(sourceFile: SourceFile, configuredChainsConfigFi
 enum LayerType {
   L1 = 'L1',
   L2 = 'L2',
+  L2a = 'L2a',
```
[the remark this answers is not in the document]
xiaodino: We can use the name such as L2a Internal and L2b Internal

packages/bridge-ui-v2/src/libs/bridge/types.ts
xiaodino: Synced with Jeff. NewOnlyOwner is used internally in relayer. Removed.

KorbinianK (approved):

PR #15358, opened 2023-12-08
feat(protocol): allow L1 timelock to be the L2 owner

dantaik:
Before this PR, the ownership of the TaikoL2 contract is really not given to the DAO as DAO lives on L1. 

This PR introduces a new contract CrossChainOwned.sol that checks the owner address sent signals from L1 to L2 to authorize onlyOwner transactions.

- [ ] @davidtaikocha The L2 genesis script may need some changes as the init() function of TaikoL2 changed.
- [x] @dantaik will add a test for it.

### DIFF
--- a/packages/protocol/contracts/L2/CrossChainOwned.sol
+++ b/packages/protocol/contracts/L2/CrossChainOwned.sol
@@ -11,20 +11,22 @@ import "lib/openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
 import "../common/EssentialContract.sol";
 import "../signal/ISignalService.sol";
 
+/// @title CrossChainOwned
+/// @notice This contract's owner lives on another chain who uses signal for transaction approval.
 abstract contract CrossChainOwned is EssentialContract {
-    mapping(bytes32 approvalHashe => bool executed) public executed; // slot 1
-    uint64 public ownerChainId; // slot 2
+    uint64 public ownerChainId; // slot 1
     uint64 public nextXchainTxId;
-    uint256[48] private __gap;
+    uint256[49] private __gap;
 
     event TransactionExecuted(uint64 indexed txId, bytes32 indexed approvalHash);
 
     error INVALID_PARAMS();
     error TX_NOT_APPROVED();
     error TX_REVERTED();
+    error NOT_CALLABLE();
 
-    function executeTransaction(bytes calldata txdata, bytes calldata proof) external {
-        bytes32 approvalHash = _isTxApproved(txdata, proof);
+    function executeApprovedTransaction(bytes calldata txdata, bytes calldata proof) external {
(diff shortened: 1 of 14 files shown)


packages/protocol/contracts/L2/CrossChainOwned.sol
```diff
@@ -0,0 +1,117 @@
...
+        external
+    {
+        if (executor != address(0) && executor != msg.sender) {
+            revert INVALID_EXECUTOR();
+        }
+
+        bytes32 approvalHash = _isTransactionApproved(txdata, proof, executor);
+        if (approvalHash == 0) revert TX_NOT_APPROVED();
+
+        (bool success,) = address(this).call(txdata);
```
davidtaikocha, line 43: If we let `owner()` be L1 `timelock`, shall we modify those `onlyOwner` methods in TaikoL2 (e.g. `function withdraw(address token, address to) external onlyOwner`), to make it only executable by the L2 executor here instead of `owner()`?

packages/protocol/contracts/L2/CrossChainOwned.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    event TransactionExecuted(uint64 indexed txId, bytes32 indexed approvalHash);
+
+    error INVALID_PARAMS();
+    error INVALID_EXECUTOR();
+    error TX_NOT_APPROVED();
+    error TX_REVERTED();
+    error NOT_CALLABLE();
+
+    function executeApprovedTransaction(
```
Brechtpd, line 29: Why does this function check its own proof instead of doing things similarly to  what `checkProcessMessageContext` where it's called directly by the bridge contract and just the original sender on the source chain needs to be checked?

packages/protocol/contracts/L2/CrossChainOwned.sol
dantaik: Yes, but we don't override `onlyOwner` modifier, we overide `_checkOwner`. It's done in the code.

packages/protocol/contracts/L2/CrossChainOwned.sol
dantaik: This new contract doesn't use the Bridge, instead, it interacts with the SignalService directly on the source chain directly.

Brechtpd: Any reason for that? Seems like this type of thing will be a common thing that people have to do for these cross layer transactions. It feels a bit strange that even we wouldn't use the standard bridge to do this.

dantaik: Using the SignalService is intuitive to me, but we can also use the Bridge. By using the bridge, maybe we can avoid the introduction of the new CrossChainOwned contract. I'll give it a try and then lets compare the two solutions.

PR #15359, opened 2023-12-08
test(protocol): SGX signature cross-check (DO NOT MERGE)

adaki2004:
Hey Patryk,

As discussed here is a quick test with explainer how to test that the SGX instance is creating the exact same signature as expected.

The scope is not to test the whole `verifyProof()` function, that data is properly sized and packed (That the size is: `89 bytes 4 bytes(id) + 20 bytes(new instance address) + 65 bytes (signature) = 89`) but only the signature validity - if assuming data to be signed is valid and correct.

You can (and shall) modify some elements like:
- Your SGX instance's private key and ethereum address `SgxVerifier.t.sol` line `78` and `81`
- If you have the chance to sign dummy data within your SGX instance, then you do not have to modify those dummy values in `SgxVerifier.t.sol` line `56-64`, just inject those.

You can run the tests by (entering into `packages/protocol` folder) this command:
`forge test --match-test 'test_SgxSignatureOnly' -vv`

Please shoot me if you have further questions.

### DIFF
--- a/packages/protocol/test/L1/SgxVerifier.t.sol
+++ b/packages/protocol/test/L1/SgxVerifier.t.sol
@@ -44,7 +44,9 @@ contract TestSgxVerifier is TaikoL1TestBase {
         // the sgx instance is the one expected.
         // What is the scope of this test ?
         // Basically checking these 2 lines in SgxVerifier.sol:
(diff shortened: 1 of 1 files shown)


adaki2004:
@pbeza closing as seems working!

PR #15368, opened 2023-12-11
refactor(protocol): use Bridge to send cross-chain owned contract transactions

dantaik:
This new implementation, compared with the one using signal service directly, does seem to be more clean and straight forward.


packages/protocol/contracts/L2/TaikoL2EIP1559Configurable.sol
```diff
@@ -21,14 +21,7 @@ contract TaikoL2EIP1559Configurable is TaikoL2 {
...
     /// @param newGasExcess The new gas excess
-    function setConfigAndExcess(
-        Config memory config,
-        uint64 newGasExcess
-    )
-        external
-        virtual
-        onlyOwner
-    {
+    function setConfigAndExcess(Config memory config, uint64 newGasExcess) external {
```
adaki2004, line 24: Do not we need `onlyOwner` here still ?

Brechtpd (approved):

adaki2004:
I cannot comment on un-changed lines so i put it here.
Tests are failing because the `setConfigAndExcess` is onlyOwner now.

Fixes are (`TaikoL2.t.sol`):
kép

and

kép

dantaik:
> setConfigAndExcess

Fixed

dantaik:
@davidtaikocha could you fix issue in TaikoL2.ts, now the TaikoL2 contract needs the address manager's address, not the signal service's address in init().

davidtaikocha (approved):

PR #15395, opened 2023-12-18
feat(bridge-ui-v2): L2 <-> L2 bridging with multi-hop

dantaik:
recreate Ruby's PR


packages/bridge-ui-v2/config/sample/configuredBridges.example
```diff
@@ -8,7 +8,7 @@
         "erc20VaultAddress": "",
         "erc721VaultAddress": "",
         "erc1155VaultAddress": "",
-        "crossChainSyncAddress": "",
+        "taikoAddress": "",
```
KorbinianK, line 11: Not sure this rename was necessary, the relayer also refers to those as CrossChainSync and the contracts implement the `ICrossChainSync` interface.

packages/bridge-ui-v2/config/sample/configuredBridges.example
```diff
@@ -44,7 +44,7 @@
...
         "signalServiceAddress": ""
       }
```
KorbinianK, line 49: Add hops example

packages/bridge-ui-v2/config/schemas/configuredBridges.schema.json
```diff
@@ -31,20 +31,39 @@
...
             "required": [
               "bridgeAddress",
               "erc20VaultAddress",
               "erc721VaultAddress",
               "erc1155VaultAddress",
-              "crossChainSyncAddress",
-              "signalServiceAddress"
+              "taikoAddress",
+              "signalServiceAddress",
+              "hops"
```
KorbinianK, line 66: Hops should not be required. L1-L2 does not need hops afaik.

packages/bridge-ui-v2/scripts/vite-plugins/generateCustomTokenConfig.ts
```diff
@@ -43,7 +43,7 @@ export function generateCustomTokenConfig() {
         const isValid = validateJsonAgainstSchema(configuredTokenConfigFile, configuredChainsSchema);
 
         if (!isValid) {
-          throw new Error('encoded configuredBridges.json is not valid.');
+          throw new Error('encoded configuredRelayer.json is not valid.');
```
KorbinianK, line 46:
```suggestion
          throw new Error('encoded configuredCustomTokens.json is not valid.');
```

packages/bridge-ui-v2/src/libs/bridge/types.ts
```diff
@@ -171,6 +171,7 @@ export type RequireApprovalArgs = {
 export type ClaimArgs = {
   msgHash: Hash;
   message: Message;
+  receipt?: TransactionReceipt;
```
KorbinianK, line 174: If you add this here, you should omit it from the derived types such as releaseArgs.

packages/bridge-ui-v2/src/libs/bridge/types.ts
```diff
@@ -198,7 +199,14 @@ export type AddressConfig = {
   etherVaultAddress?: Address;
   erc721VaultAddress: Address;
   erc1155VaultAddress: Address;
-  crossChainSyncAddress: Address;
+  taikoAddress: Address;
+  signalServiceAddress: Address;
+  hops: Array<HopAddressConfig>;
```
KorbinianK, line 204: optional

packages/bridge-ui-v2/src/libs/relayer/RelayerAPIService.ts
```diff
@@ -6,7 +6,7 @@ import { Buffer } from 'buffer';
 import { bridgeABI } from '$abi';
 import { routingContractsMap } from '$bridgeConfig';
 import { apiService } from '$config';
-import type { BridgeTransaction, MessageStatus } from '$libs/bridge';
+import { type BridgeTransaction, MessageStatus } from '$libs/bridge';
```
KorbinianK, line 9: why? Both are used as type.

KorbinianK (changes requested):
There are also a lot of formatting changes that the linter should fix.

Run `pnpm svelte:check && pnpm lint:fix`

packages/bridge-ui-v2/src/libs/bridge/ETHBridge.ts
```diff
@@ -109,12 +109,19 @@ export class ETHBridge extends Bridge {
...
+    const { msgHash, message, receipt } = args;
     const srcChainId = Number(message.srcChainId);
     const destChainId = Number(message.destChainId);
 
     if (messageStatus === MessageStatus.NEW) {
-      const proof = await this._prover.encodedSignalProof(msgHash, srcChainId, destChainId);
+      const hops = routingContractsMap[srcChainId][destChainId].hops;
+
+      let proof;
+      if (hops.length > 0) {
```
KorbinianK, line 120: If hops are optional this will fail, just a heads up :)

packages/bridge-ui-v2/src/libs/proof/BridgeProver.ts
```diff
@@ -1,27 +1,213 @@
...
+    // Get the block number from last hop chain to dest chain
+    blockNumber = await this.getBlockNumber(hopChainId, destChainId, destCrossChainSyncAddress);
+
+    // Generate main storage proof with receipt.blockNumber
+    const { proof, rlpEncodedStorageProof } = await this.encodedStorageProof({
+      msgHash,
+      clientChainId: srcChainId,
+      contractAddress: srcBridgeAddress,
+      proofForAccountAddress: srcSignalServiceAddress,
+      blockNumber: receipt.blockNumber,
```
KorbinianK, line 152: I see this the only place we actually need the receipt. From what I understood about the proofs, we don't necessarily need the exact block, just any after. Technically 'latest' works, but it could not be synced. So if we fetch the latest synced block, so the one you fetch above, doesn't this work as well?

xiaodino: Good question. Let me verify it.

packages/bridge-ui-v2/config/sample/configuredBridges.example
xiaodino: Added

packages/bridge-ui-v2/src/libs/proof/BridgeProver.ts
xiaodino: We don't need the receipt here. Removed.

packages/bridge-ui-v2/src/libs/bridge/ETHBridge.ts
xiaodino: Update.

packages/bridge-ui-v2/src/libs/relayer/RelayerAPIService.ts
xiaodino: Reverted the change.

packages/bridge-ui-v2/src/libs/bridge/types.ts
xiaodino: Updated.

packages/bridge-ui-v2/src/libs/bridge/types.ts
xiaodino: Updated. We don't need receipt.

packages/bridge-ui-v2/scripts/vite-plugins/generateCustomTokenConfig.ts
xiaodino: Updated.

packages/bridge-ui-v2/config/schemas/configuredBridges.schema.json
xiaodino: Updated.

packages/bridge-ui-v2/config/sample/configuredBridges.example
xiaodino: Reverted the change.

KorbinianK (approved):

xiaodino (approved):

PR #15416, opened 2023-12-26
fix(protocol): use safeTransferFrom to transfer ERC721 to unspecified address `to`

paweenpit:
Use `safeTransferFrom` to ensure target address `to` implements `onERC721received`

### DIFF
--- a/packages/protocol/contracts/tokenvault/ERC721Vault.sol
+++ b/packages/protocol/contracts/tokenvault/ERC721Vault.sol
@@ -106,7 +106,7 @@ contract ERC721Vault is BaseNFTVault, IERC721Receiver, IERC165Upgradeable {
             if (ctoken.chainId == block.chainid) {
                 token = ctoken.addr;
                 for (uint256 i; i < tokenIds.length; ++i) {
-                    ERC721Upgradeable(token).transferFrom({
+                    ERC721Upgradeable(token).safeTransferFrom({
                         from: address(this),
                         to: to,
                         tokenId: tokenIds[i]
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/tokenvault/ERC721Vault.sol
```diff
@@ -106,7 +106,7 @@ contract ERC721Vault is BaseNFTVault, IERC721Receiver, IERC165Upgradeable {
             if (ctoken.chainId == block.chainid) {
                 token = ctoken.addr;
                 for (uint256 i; i < tokenIds.length; ++i) {
-                    ERC721Upgradeable(token).transferFrom({
+                    ERC721Upgradeable(token).safeTransferFrom({
```
dantaik, line 109: could you also change "BridgedERC721.sol"  from
`return ERC721Upgradeable.transferFrom(from, to, tokenId);` to 
`return ERC721Upgradeable.safeTransferFrom(from, to, tokenId);`

dantaik: There is another `.transferFrom` call in this file, please also replace it with 'safeTransferFrom'

paweenpit: The other `.transferFrom` in this file sends the `ERC721` to `address(this)` (this contract) which already implements `onERC721received`. So there is no need to use `.safeTransferFrom`. Using `.transferFrom` will optimize a bit more gas.

paweenpit: for "BridgedERC721.sol", I'd recommend overriding `.safeTransferFrom` to call `ERC721Upgradeable.safeTransferFrom(from, to, tokenId);` and use this function elsewhere rather than implementing `.safeTransferFrom` in the `.transferFrom`.

dantaik:
@paweenpit thank you for this PR. Inspired by your change, I made this PR with more changes and bug fixes.https://github.com/taikoxyz/taiko-mono/pull/15422

PR #15422, opened 2023-12-27
fix(protocol): fix bridge token transfer check

dantaik:
Inspired by https://github.com/taikoxyz/taiko-mono/pull/15416/files

Fixed before-transfer check by overriding the right function.


packages/protocol/contracts/tokenvault/ERC721Vault.sol
```diff
@@ -234,7 +234,7 @@ contract ERC721Vault is BaseNFTVault, IERC721ReceiverUpgradeable {
                 });
 
                 for (uint256 i; i < op.tokenIds.length; ++i) {
-                    t.transferFrom(user, address(this), op.tokenIds[i]);
```
paweenpit, line 237: I believe there is no need to use `.safeTransferFrom` here as ERC721 is transferred to this contract which has already implemented `onERC721received`

paweenpit:
added another [PR#15423](https://github.com/taikoxyz/taiko-mono/pull/15423) to this branch.

davidtaikocha (approved):

paweenpit (approved):

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -160,4 +118,11 @@ contract BridgedERC20 is BridgedERC20Base, IERC20MetadataUpgradeable, ERC20Upgra
     function _burnToken(address from, uint256 amount) internal override {
         _burn(from, amount);
     }
+
+    function _transfer(address from, address to, uint256 amount) internal virtual override {
```
adaki2004, line 122: Why not override `_beforeTokenTransfer()`, same as you did with ERC1155 ?

adaki2004 (commented):
Beside my comments, LGTM, one small reminder:
- We removed the `whenNotPaused` modifier from the bridged tokens, which seems OK to me, since they can be transferred (among users) but bridge contracts would anyways paused, so it's ok i guess.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -159,8 +159,7 @@ contract TaikoL2 is CrossChainOwned, TaikoL2Signer, ICrossChainSync {
         if (token == address(0)) {
             to.sendEther(address(this).balance);
         } else {
-            IERC20 t = IERC20(token);
-            t.transfer(to, t.balanceOf(address(this)));
+            IERC20(token).transfer(to, IERC20(token).balanceOf(address(this)));
```
adaki2004, line 162: Should not we use SafeERC20 ?

adaki2004, line 162: Should not we use SafeERC20 ? As like in AssignmentHook ?
https://github.com/taikoxyz/taiko-mono/pull/15423/files

adaki2004, line 162: Should not we use SafeERC20 ? As like in AssignmentHook ?
https://github.com/taikoxyz/taiko-mono/pull/15423/files

dantaik: Fixed

dantaik:
> Beside my comments, LGTM, one small reminder:
> 
> * We removed the `whenNotPaused` modifier from the bridged tokens, which seems OK to me, since they can be transferred (among users) but bridge contracts would anyways paused, so it's ok i guess.

We checked the following in `_beforeTokenTransfer`:

```
if (paused()) revert INVALID_PAUSE_STATUS();
```

dantaik:
@adaki2004 Please review again.

adaki2004 (approved):

PR #15426, opened 2023-12-27
fix(bridge-ui-v2): ETHBridge Handling for Different Destination Addresses

joaolago1113:
### Overview
This pull request introduces a critical fix in the ETHBridge logic, particularly addressing an issue where the transaction value was incorrectly set to zero when the destination address differed from the sender's address.

### Problem
Previously, in the ETHBridge `_prepareTransaction` function, the transaction value was erroneously set to zero if the `to` address (destination) was not the same as the `owner` (sender). This behavior was due to the way the `value` variable was being assigned, resulting in transactions with a value of zero when sent to a different address.

### Solution
To resolve this, the logic for determining the transaction value has been modified. The updated logic now accurately sets the `value` based on whether the transaction is self-directed or to another address. This ensures that the correct amount is assigned to the transaction regardless of the destination address.

### Changes
- Modified the `_prepareTransaction` function in `ETHBridge.ts`.
- Introduced two temporary variables `valueForSelf` and `valueForOther` to hold the values for self-directed and other-directed transactions, respectively.
- Implemented conditional logic to correctly assign the transaction value based on the destination address.

### Impact
This fix is crucial for the proper functioning of the bridge when transactions are intended for addresses other than the sender's. It ensures the correct value is set for the transaction, thus preventing failures or incorrect transaction behavior.

### Testing
The changes have been tested for both scenarios - sending to the same address and to a different address. The transactions now correctly reflect the intended value.

### Additional Notes
This fix does not fully implement the feature mentioned in the existing `TODO` comment about supporting simultaneous transactions to self and another address. That feature enhancement remains to be implemented in the future.

### DIFF
--- a/packages/bridge-ui-v2/src/libs/bridge/ETHBridge.ts
+++ b/packages/bridge-ui-v2/src/libs/bridge/ETHBridge.ts
@@ -27,7 +27,8 @@ export class ETHBridge extends Bridge {
 
     // TODO: contract actually supports bridging to ourselves as well as
     //       to another address at the same time
-    const [senderAmount, recipientAmount] = to.toLowerCase() === owner.toLowerCase() ? [amount, BigInt(0)] : [BigInt(0), amount];
+    const [senderAmount, recipientAmount] =
+      to.toLowerCase() === owner.toLowerCase() ? [amount, BigInt(0)] : [BigInt(0), amount];
     let value;
     if (senderAmount === BigInt(0)) {
       value = recipientAmount;
--- a/packages/bridge-ui-v2/src/libs/bridge/ETHBridge.ts
+++ b/packages/bridge-ui-v2/src/libs/bridge/ETHBridge.ts
@@ -27,7 +27,13 @@ export class ETHBridge extends Bridge {
 
     // TODO: contract actually supports bridging to ourselves as well as
     //       to another address at the same time
-    const [value] = to.toLowerCase() === owner.toLowerCase() ? [amount, BigInt(0)] : [BigInt(0), amount];
+    const [senderAmount, recipientAmount] = to.toLowerCase() === owner.toLowerCase() ? [amount, BigInt(0)] : [BigInt(0), amount];
+    let value;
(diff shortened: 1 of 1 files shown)


KorbinianK (dismissed):
Great spot!

davidtaikocha (dismissed):

KorbinianK:
@joaolago1113 Please fix the linting errors

KorbinianK (approved):

davidtaikocha (approved):

adaki2004:
Okay @joaolago1113, i checked that comment and context in the code and i have more clarity now.

So you referred to.

```
    // TODO: contract actually supports bridging to ourselves as well as
    //       to another address at the same time
```

This comment is from @jscriptcoder , from 6 months ago.

2 possible things it could mean.

**1.** 
Bridged contracts actually supports bridging to 
A.: ourselves (so who initiated the transaction)
AND (as well as)
B: another (custom) address (see screenshot's red circle below)

BUT not at the same time (execution time), just the wording is not ideal. Same time means more: 'now = finally = either. or something liek that.

kép

OR (and i think this is more probable)
**2.** 

There is a possibility to bridge `ETH` with a `ERC20/ERC721/ERC1155` bridging, together, literally at the same time (with same TXN), (but NOT to a different address, and so also the wording is not ideal here ) but this feature is not supported in the UI (currently).

@joaolago1113 so no need to modify the contracts, as functional wise works as we expect, no need to have a new feature which can handle 2 recipients and amounts (at the same time) on the destination chain. (it would involve additional complexity, both on the UI and in the contract)

@KorbinianK, what do you think what Fran could have referred to here ? It is a 6 months old commit, from him and i think he meant the 2nd one.

Commit link: https://github.com/taikoxyz/taiko-mono/commit/4b639d7a5315c20a6766fb2b59d0ce5d3b973453 

P.S.:  (or if you are here Fran, if you know please let us know :) )
https://github.com/taikoxyz/taiko-mono/commit/4b639d7a5315c20a6766fb2b59d0ce5d3b973453

PR #15435, opened 2023-12-29
refactor(protocol): USDCAdaptor gas optimization

Kittikornl:
change to immutable

### DIFF
--- a/packages/protocol/contracts/tokenvault/adaptors/USDCAdaptor.sol
+++ b/packages/protocol/contracts/tokenvault/adaptors/USDCAdaptor.sol
@@ -24,20 +24,24 @@ interface IUSDC {
 
 /// @title USDCAdaptor
 contract USDCAdaptor is BridgedERC20Base {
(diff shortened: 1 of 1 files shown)


adaki2004:
We do not want this to be immutable - but changeable - with upgrades.

Kittikornl:
Immutable can be changed with upgrades

adaki2004:
> Immutable can be changed with upgrades

Well, yes and no at the same time.
Yes, because we could modify the logic to work, but no because the USDC (in your code) is not stored on storage slot 0.
 
Tho we would be able to change to a different address, the `IUSDC public immuable USDC;` as you proposed is not stored in slot 0, because the compiler does not allocate storage slot for immutable, but rather the `__gap[0]` would be on slot 0. And it would require to have` __gap` size on 50 (not 49) and use the gap later (if we want to change), so I think it is good as is, but thanks. Please keep contributing, and try to make it more meaningful. Tiny optimizations are helpful too, but we need to keep other considerations in mind.

kép

PR #15437, opened 2023-12-30
chore(bridge-ui-v2): Improve tooltip ux

mintcnn:
- update dotenv config to fix error 'env not found' when running `pnpm dev` on Node 18
- improve the UX of the tooltip
    - open the tooltip when hover or click
    - close the tooltip when the mouse leaves or click outside with some delay

### DIFF
--- a/packages/bridge-ui-v2/src/components/Tooltip/Tooltip.svelte
+++ b/packages/bridge-ui-v2/src/components/Tooltip/Tooltip.svelte
@@ -40,7 +40,7 @@
   });
 </script>
 
-<div class={classes} role="presentation" on:mouseleave={() => closeTooltip(1000)}>
+<div class={classes} role="presentation" on:mouseleave={() => closeTooltip(200)}>
   <button
     aria-haspopup="dialog"
     aria-controls={tooltipId}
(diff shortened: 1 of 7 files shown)


KorbinianK:
There is nothing wrong with the env file loading. It's expected in the package and works just fine.

mintcnn:
> There is nothing wrong with the env file loading. It's expected in the package and works just fine.
Reproduced! but did not find the error again. I will revert the change for dotenv!

packages/bridge-ui-v2/src/components/Tooltip/Tooltip.svelte
```diff
@@ -31,22 +33,21 @@
...
   });
 
   onDestroy(() => {
     closeTooltip();
-    document.removeEventListener('click', closeTooltip);
   });
 </script>
 
-<div class={classes}>
+<div class={classes} role="presentation" on:mouseleave={() => closeTooltip(1000)}>
```
KorbinianK, line 43:
```suggestion
<div class={classes} role="presentation" on:mouseleave={() => closeTooltip(200)}>
```

KorbinianK (dismissed):

KorbinianK:
@mintcnn can you rebase your branch to `alpha-6` please? Then we merge it. I changed the timeout to 200ms, but I like the behaviour.

mintcnn:
> Contributor
@KorbinianK 
I already created the new PR since rebasing to alpha-6 has a lot of changes and conflict files, please help review again
https://github.com/taikoxyz/taiko-mono/pull/15449

PR #15456, opened 2024-01-05
chore(relayer): remove unnecessary conversions

estensen:
Use linter to detect unnecessary type conversions
https://github.com/mdempsky/unconvert


davidtaikocha (approved):

cyberhorsey:
should we be merging PRs to main right now instead of alpha-6? I don't think so, it will create conflicts when merging alpha-6 back to main.

davidtaikocha:
Closed due to: https://github.com/taikoxyz/taiko-mono/pull/15456#issuecomment-1879814527

PR #15473, opened 2024-01-10
fix(protocol): fix immutable-variable bug in UUPSUpgradeable for L2 predeployment

dantaik:
Please carefully review this PR. 

The original issue: the OZ's UUPSUpgradeable implementation of has the following code:

```solidity
abstract contract UUPSUpgradeable is IERC1822Proxiable, ERC1967Upgrade {
    address private immutable __self = address(this);
...
}
```

When we pre-deploy a sub-contract of UUPSUpgradeable to L2 genesis, immutable variable `__self` will not be initialized and there is no storage slot for this variable (see https://medium.com/@ajaotosinserah/a-comprehensive-guide-to-implementing-constant-and-immutable-variables-in-solidity-4026ebadc6aa). 

We need to remove immutable variables from the code of all contracts we pre-deploy.

### DIFF
--- a/packages/protocol/contracts/common/UUPSUpgradeableAlt.sol
+++ b/packages/protocol/contracts/common/UUPSUpgradeableAlt.sol
@@ -23,7 +23,7 @@ import "lib/openzeppelin-contracts/contracts/proxy/ERC1967/ERC1967Upgrade.sol";
  * lib/openzeppelin-contracts/contracts/proxy/utils/UUPSUpgradeable.sol
  * without the immutable variable so we can pre-deploy these contracts on L2.
  */
-abstract contract UUPSUpgradeable is IERC1822Proxiable, ERC1967Upgrade {
+abstract contract UUPSUpgradeableAlt is IERC1822Proxiable, ERC1967Upgrade {
     uint256[50] private __gap;
 
     error DISALLOWED_WHEN_NOT_DELEGAATED();
--- a/packages/protocol/genesis/GenerateGenesis.g.sol
+++ b/packages/protocol/genesis/GenerateGenesis.g.sol
@@ -88,7 +88,7 @@ contract TestGenerateGenesis is Test, AddressResolver {
 
         vm.startPrank(addressManager.owner());
 
-        addressManager.upgradeTo(address(newAddressManager));
+        addressManagerProxy.upgradeTo(address(newAddressManager));
 
         vm.stopPrank();
     }
@@ -109,7 +109,7 @@ contract TestGenerateGenesis is Test, AddressResolver {
 
         vm.startPrank(addressManager.owner());
 
-        addressManager.upgradeTo(address(newAddressManager));
+        addressManagerProxy.upgradeTo(address(newAddressManager));
 
         vm.stopPrank();
     }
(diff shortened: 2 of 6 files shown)


packages/protocol/contracts/common/UUPSUpgradeableAlt.sol
```diff
@@ -0,0 +1,114 @@
...
+    uint256[50] private __gap;
+
+    error DISALLOWED_WHEN_NOT_DELEGAATED();
+    error DISALLOWED_WHEN_IS_DELEGATED();
+    /**
+     * @dev Check that the execution is being performed through a delegatecall call and that the
+     * execution context is a proxy contract with an implementation (as defined in ERC1967).
+     */
+
+    modifier onlyProxy() {
```
dantaik, line 36: For these two modifiers, please compare the current implementations with the original ones in `lib/openzeppelin-contracts/contracts/proxy/utils/UUPSUpgradeable.sol`

dantaik: @Brechtpd Please also help me review the code.

Brechtpd:
@dantaik There is a way to actually keep the immutable variables by manually patching the bytecode before predeploying to emulate what happens in the constructor, @adaki2004 and @davidtaikocha found a way to do, but unsure how easy it is to do in practice.

But will also check this approach as well.

packages/protocol/genesis/GenerateGenesis.g.sol
```diff
@@ -88,7 +88,7 @@ contract TestGenerateGenesis is Test, AddressResolver {
 
         vm.startPrank(addressManager.owner());
 
-        addressManager.upgradeTo(address(newAddressManager));
+        addressManagerProxy.upgradeTo(address(newAddressManager));
```
dantaik, line 91: It seems weird that there is no `new ERC1967Proxy` in this file.

packages/protocol/contracts/common/UUPSUpgradeableAlt.sol
Brechtpd: I think this check it less strong than the original one because, without `_self`,  it's impossible here to check that the correct implementation contract is being called without having access to the "real" `address(this)` of the implementation contract. So we can only check that something is delegated, but not that it is delegated to the correct implementation contract, which is what the original `onlyProxy` does check.

dong77: I also prefer to substitute the address in the compiled bytecode to "initialize" immutable variables correctly. Lets wait for @davidtaikocha on that.

cyberhorsey: is there a strong reason for the switch to this proxy versus the pre existing TtansparentUpgradableProxy we had before which I believe doesn't have this issue (since we were able to upgrade contacts before?)

afaik, main feature of this proxy is you can eventually disable upgradability.

is it better to patch this, or use an out of the box proxy solution that's worked for us in the past?

adaki2004:
Maybe let's wait if we can fix predeployment with inserting the address into the correct place OR change to Transparent ? This will break the 'standard' - not only code(storage) wise but functional wise.

PR #15486, opened 2024-01-12
feat(protocol): fix issues in AssignmentHook

adaki2004:
From team SigP.:

- fix bug in ERC20 proverFee payment (previous from TaikoL1, now from L2 block's coinbase address)
- add L2's chain id and the hook address to the assignment hash calculation
- check duplicate hooks
- check TKO balance change strict using `==` rather than `>=`.


packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -261,8 +261,10 @@ library LibProposing {
             }
 
             // Check that after hooks, the Taiko Token balance of this contract
-            // have increased by at least config.livenessBond
-            if (tko.balanceOf(address(this)) < tkoBalance + config.livenessBond) {
+            // have increased by the same amount as config.livenessBond (to prevent)
+            // multiple draining payments by a malicious proposer nesting the same
+            // hook.
+            if (tko.balanceOf(address(this)) != tkoBalance + config.livenessBond) {
```
davidtaikocha, line 267: emm I guess another reason why we use `<` here is to allow hook providers to charge extra `TKO`?

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -32,7 +32,7 @@ contract AssignmentHook is EssentialContract, IHook {
         uint64 expiry;
         uint64 maxBlockId;
         uint64 maxProposedIn;
-        bytes32 metaHash;
```
davidtaikocha, line 35: emmm since we also allow other implementation of hooks, there is a change others may want to use it?

packages/protocol/contracts/L1/libs/LibProposing.sol
adaki2004: It is dangerous to handle like that. Hooks -imo - are not like 3rd party "providers" but more just an external libraries, extending the functionality.

Scneario:
1. Proposer gets a prover's signature to prove a block.
2. A malicious proposer can embedded that data into multiple hooks (per same` proposeBlock()`) which will drain the TKO balance - as liveness bond.
So basically this is to guarantee that the proposer does not overpay by embedding ("replaying") his/her signed assigment per same hook array.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
Brechtpd: I would opt to keep `metaHash` and put it in the data signed by the prover. Then it's still an option to use it or not, and provers can actually know beforehand what they will have to prove.

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: I would say to just limit the protocol to a single hook. If a prover/proposer wants to do multiple function calls that's fine, but they'll have to create a new hook smart contract to do that. Seems like a niche use case that we don't see people using directly, and if they want to do something like that they still can, but we keep the complexity outside of the main protocol.

With data blobs now (and even more efficient data copies from memory coming when still using calldata) it'll be more flexible and less complex to move custom logic on top of the propose function instead of inside it as well, so we'll be able to move more stuff out in future versions.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
Brechtpd: A combination of `maxBlockId` and `maxProposedIn` could also achieve most of that, but indeed not multiple blocks inside a single data blob.

packages/protocol/contracts/L1/libs/LibProposing.sol
adaki2004: Ok, verdict:
- I leave the API as is, but restrict to only call 1 hook. (revert if more)
- Also leave the equal here, as if the hook provider charges more, it will keep the TKO, and not forwards it to TaikoL1.

dantaik, line 267: Why we need to be strict on this?

dantaik:
> * remove `metaHash` in assignment since it is unused

It is not used in this hook, but maybe used in another.

packages/protocol/contracts/L1/libs/LibProposing.sol
adaki2004: > Why we need to be strict on this?

First: Who likes to overpay ? :) 
Second: Currently the design now allows replaying the same hook.data multiple times, and then prover pays multiple times the liveness bond within the same proposeBlock(). This is preventing it.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -32,7 +32,7 @@ contract AssignmentHook is EssentialContract, IHook {
...
+        uint64 chainId;
```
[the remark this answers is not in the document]
adaki2004: Please see latest commit, i already added it back.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -29,6 +29,7 @@ contract AssignmentHook is EssentialContract, IHook {
 
     struct ProverAssignment {
         address feeToken;
+        uint64 chainId;
```
dantaik, line 32: I don't think we need to add chainId here -- simply put chainid() to the `hashAssignment` function's`abi.encode`?

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -141,13 +144,16 @@ contract AssignmentHook is EssentialContract, IHook {
...
     )
         public
-        pure
+        view
         returns (bytes32)
     {
         return keccak256(
             abi.encode(
                 "PROVER_ASSIGNMENT",
+                assignment.chainId,
```
dantaik, line 153: chainid() here

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: > It is dangerous to handle like that. Hooks -imo - are not like 3rd party "providers" but more just an external libraries, extending the functionality.
> 
> Scneario:
> 
> 1. Proposer gets a prover's signature to prove a block.
> 2. A malicious proposer can embedded that data into multiple hooks (per same` proposeBlock()`) which will drain the TKO balance - as liveness bond.
>    So basically this is to guarantee that the proposer does not overpay by embedding ("replaying") his/her signed assigment per same hook array.

our own hook is simply one implementation, we don't even check if the hook used is our impl or another one from a third party.

dantaik: The provider will check the hook addresses and only sign the Assignment if all checks pass.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -243,6 +244,10 @@ library LibProposing {
             IERC20 tko = IERC20(resolver.resolve("taiko_token", false));
             uint256 tkoBalance = tko.balanceOf(address(this));
 
+            // Allow max one hook - for now - but keep the API for
+            // possibility to have more in the future.
+            if (params.hookCalls.length > 1) revert L1_TOO_MANY_HOOKS();
```
dantaik, line 249: Unless there is a bug, I still prefer allowing multiple hooks.

dantaik (changes requested):
My feedback summary:
- keep multiple hooks, so hooks can stack on top of each other instead of wrap each other for max reuseability
- remove blockid as a parameter, use `blockid()` directly inside the hash funciton
- keep the "!=" check -- we can change it back to ">=" later if we have a scenario

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -120,7 +128,9 @@ contract AssignmentHook is EssentialContract, IHook {
                 refund = msg.value - input.tip;
             }
             // Paying ERC20 tokens
-            IERC20(assignment.feeToken).safeTransferFrom(msg.sender, blk.assignedProver, proverFee);
+            IERC20(assignment.feeToken).safeTransferFrom(
```
dantaik, line 131: This fixes the bug found by SigmaPrime:
> Also on the AssignmentHook when paying in fee tokens rather than native tokens it transfers from msg.sender which is actually the TaikoL1 contract. See here. This seems like it should be calling meta.coinbase or otherwise restricting it to the caller of blockPropose().

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
dantaik, line 32: it doesn't work if we simply add chainid here without verifying it's the same as the L1 contract's configured chainId.

adaki2004:
> My feedback summary:
> 
>     * keep multiple hooks, so hooks can stack on top of each other instead of wrap each other for max reuseability
> 
>     * remove blockid as a parameter, use `blockid()` directly inside the hash funciton
> 
>     * keep the "!=" check -- we can change it back to ">=" later if we have a scenario

Done - thanks for the pr for chaind.

dantaik (approved):

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -120,7 +127,9 @@ contract AssignmentHook is EssentialContract, IHook {
...
+                meta.coinbase, blk.assignedProver, proverFee
```
dantaik, line 131: fix bug in ERC20 proverFee payment (previous from TaikoL1, now from L2 block's coinbase address)

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -129,25 +138,29 @@ contract AssignmentHook is EssentialContract, IHook {
...
         public
-        pure
+        view
         returns (bytes32)
     {
         return keccak256(
             abi.encode(
                 "PROVER_ASSIGNMENT",
-                taikoAddress,
+                ITaikoL1(taikoL1Address).getConfig().chainId,
```
dantaik, line 160: add L2's chain id to the assignment hash calculation

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik, line 267: check TKO balance change strict using == rather than >=.

dantaik: @adaki2004  with [this change](https://github.com/taikoxyz/taiko-mono/pull/15492/files), maybe we don't have to check "==" here?

davidtaikocha (approved):
And client updates are ready too: https://github.com/taikoxyz/taiko-client/pull/502

packages/protocol/contracts/L1/libs/LibProposing.sol
kirk-baird: I still recommend having a strict equality, for two reasons
a) I don't see a benefit for transferring more than the liveness bond, it would increase the cost to the prover and the extra funds would be stuck in the TaikoL1 contract.
b) If multiple hooks are available which can transfer TKO to the TaikoL1 contract the proposer could attempt to use both e.g. If there is a third party hook, a proposer could attempt to use both hooks.

adaki2004: > I still recommend having a strict equality, for two reasons

I think this is a strict equalitsy, no ? (if NOT -> revert)
`if (tko.balanceOf(address(this)) != tkoBalance + config.livenessBond) revert L1_LIVENESS_BOND_NOT_RECEIVED`

kirk-baird: Yep this PR currently has a strict equality, which I currently see as a nicer solution. I was responding to @dantaik comment about potentially moving back to inequality.
 
 > @adaki2004 with [this change](https://github.com/taikoxyz/taiko-mono/pull/15492/files), maybe we don't have to check "==" here?

PR #15492, opened 2024-01-13
fix(protocol): disallow duplicate hooks

dantaik:


packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -246,14 +247,21 @@ library LibProposing {
             // Run all hooks.
             // Note that address(this).balance has been updated with msg.value,
             // prior to any code in this function has been executed.
+            address prevHook;
             for (uint256 i; i < params.hookCalls.length; ++i) {
+                if (uint160(prevHook) >= uint160(params.hookCalls[i].hook)) {
+                    revert L1_INVALID_HOOK();
```
adaki2004, line 253: Is it possible that in the future Hooks will depend on each other and call sequence matters ? If so, it might not be possible to rank them based on integers, no ? 🤔

dong77: It the order matters, then one has to develop and deploy a new hook that internally calls other hooks, or one hook that just have logics from more other hooks.

davidtaikocha (approved):

packages/protocol/contracts/L1/libs/LibProposing.sol
adaki2004: Ok, be it. :) i guess we can 'worry' about this later, i dont think we will have multiple hooks (soon) anyways.

adaki2004 (approved):

PR #15493, opened 2024-01-14
feat(protocol): update a protocol comment

davidtaikocha:
Right now we still use `80` as `blockMaxTxs ` in consensus.


Brechtpd (approved):
Realistically speaking about how we will likely use the PSE circuits, I think we shouldn't have any specific limitation for this anymore. But of course also doesn't really hurt to keep it in for now.

smtmfft (approved):

PR #15506, opened 2024-01-16
feat(protocol): Strike price to token grants

adaki2004:
TKO tokens now shall be bought when withdrawn for a strike price (per TKO in stable token) - it still can be 0.

### DIFF
--- a/packages/protocol/contracts/team/TimelockTokenPool.sol
+++ b/packages/protocol/contracts/team/TimelockTokenPool.sol
@@ -59,15 +59,12 @@ contract TimelockTokenPool is EssentialContract {
         uint32 unlockPeriod;
         // Strike price per TKO (1e18) in stables (e.g. USDC) in wei.
         uint64 strikePrice;
-    }
-
-    struct Recipient {
+        // Withdrawn already
         uint128 amountWithdrawn;
-        Grant[] grants;
     }
 
     uint256 public constant MAX_GRANTS_PER_ADDRESS = 8;
-    uint256 public constant ONE_TKO_UNIT = 1 * 10 ** 18;
+    uint128 public constant ONE_TKO_UNIT = 1 * 10 ** 18;
 
     address public taikoToken;
     address public strikePriceToken;
@@ -75,7 +72,7 @@ contract TimelockTokenPool is EssentialContract {
     uint128 public totalAmountGranted;
     uint128 public totalAmountVoided;
     uint128 public totalAmountWithdrawn;
-    mapping(address recipient => Recipient) public recipients;
+    mapping(address recipient => Grant[]) public recipients;
     uint128[43] private __gap;
 
     event Granted(address indexed recipient, Grant grant);
@@ -116,14 +113,14 @@ contract TimelockTokenPool is EssentialContract {
     /// same recipient.
     function grant(address recipient, Grant memory g) external onlyOwner {
         if (recipient == address(0)) revert INVALID_PARAM();
-        if (recipients[recipient].grants.length >= MAX_GRANTS_PER_ADDRESS) {
+        if (recipients[recipient].length >= MAX_GRANTS_PER_ADDRESS) {
             revert TOO_MANY();
         }
 
         _validateGrant(g);
 
         totalAmountGranted += g.amount;
-        recipients[recipient].grants.push(g);
+        recipients[recipient].push(g);
         emit Granted(recipient, g);
     }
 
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/team/TimelockTokenPool.sol
```diff
@@ -19,6 +19,13 @@ import "lib/openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
 import "lib/openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
 import "../common/EssentialContract.sol";
 
+/// @title IDecimals
+/// @dev Serves for calculating the common denominator (TKO vs. strike price in stables)
+interface IDecimals {
```
dantaik, line 24: You don't have to define this interface even if you want to use the decimals() function, you can use IERC20 from OZ. But I don't think you should use IDecimals in the first place as we can assume TKO's decimals is always 18 on Both Taiko L1 and L2.

packages/protocol/contracts/team/TimelockTokenPool.sol
```diff
@@ -168,18 +198,28 @@ contract TimelockTokenPool is EssentialContract {
 
     function _withdraw(address recipient, address to) private {
         Recipient storage r = recipients[recipient];
+
+        uint256 price;
```
dantaik, line 202: here price should be renamed to "cost"?

packages/protocol/contracts/team/TimelockTokenPool.sol
```diff
@@ -168,18 +198,28 @@ contract TimelockTokenPool is EssentialContract {
...
         uint128 amount;
 
         uint256 rGrantsLength = r.grants.length;
         for (uint128 i; i < rGrantsLength; ++i) {
-            amount += _getAmountUnlocked(r.grants[i]);
+            uint128 perGrantAmount;
+
+            perGrantAmount = _getAmountUnlocked(r.grants[i]);
```
dantaik, line 209: it will be great to merge these 3 lines into the _getAmountUnlocked and rename it as _getAmountAndCostUnlocked

dantaik: Here is a logical error: `r.amountWithdrawn` is the total amount withdrawn for all grants assigned to this address, previously this is ok, but now each grant has a different strike price, you have to calculated the cost per grant with its own `amountWithdrawn`. The Recipient must be modified to remove `amountWithdrawn` and put `amountWithdrawn` inside each grant`. How tests didn't capture this?

packages/protocol/test/team/TimelockTokenPool.t.sol
```diff
@@ -9,32 +9,44 @@ contract MyERC20 is ERC20 {
     }
 }
 
+contract USDC is ERC20 {
+    constructor(address recipient) ERC20("USDC", "USDC") {
+        _mint(recipient, 1_000_000_000e6);
+    }
+
+    function decimals() public view virtual override returns (uint8) {
```
dantaik, line 17: this is not used at all.

dantaik (changes requested):

packages/protocol/contracts/team/TimelockTokenPool.sol
adaki2004: `decimals()` is not part of `IER20.sol` but OK, i can remove and have it fixed unit.

packages/protocol/contracts/team/TimelockTokenPool.sol
adaki2004: Because it is a loop per grant. And it multiply the per grant amount with per grant price, so there is no logical error here.

adaki2004: Amount withdrawn is TKO.
I calculated here the cost per grant (in the loop).

amountWithdrawn and amountWithdrawable is per overall tokens not per grant! (You wrote that part of the code).

Who really cares about amountWithdrawn per grant ? If Y portion is withdrawable on grant A, and X portion is withdrawable on grant B, than -> amount withdrawabel is : Y + X

packages/protocol/test/team/TimelockTokenPool.t.sol
adaki2004: It is function overloading, so that USDC (with 6 decimals) works as expected. If you change this part, test will fail.

packages/protocol/contracts/team/TimelockTokenPool.sol
adaki2004: Done.
Tests were not catching because i set the strikePrices to 0 every existing tests, and added 2 new, but they did not intermediary steps, just X years after 100% fully unlocks, so this kind of case did not come up. Now i modified all files to have a non-zero strike price.

dantaik:
I prefer another implementation in https://github.com/taikoxyz/taiko-mono/pull/15522 with minimal changes.

PR #15511, opened 2024-01-16
fix(eventindexer): Revise the names of the db env vars

lxcong:
Revise the names of the MySQL env vars read from the CLI to match the documentation and the .default.env file.


cyberhorsey:
PRs should not be opened to main right now, they should be opened to alpha-6 branch.

PR #15514, opened 2024-01-17
feat(protocol): Sgx improvements

adaki2004:
Changes:
- Restrict call to TaikoL1 and ProofVerifierAggregator. 
- eip712 style hashing

### DIFF
--- a/packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
+++ b/packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
@@ -17,6 +17,7 @@ pragma solidity 0.8.20;
 import "lib/openzeppelin-contracts/contracts/utils/cryptography/ECDSA.sol";
 import "../../common/EssentialContract.sol";
 import "../../thirdparty/LibBytesUtils.sol";
+import "../ITaikoL1.sol";
 import "../TaikoData.sol";
 import "./IVerifier.sol";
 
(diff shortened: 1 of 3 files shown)


Brechtpd (approved):

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -59,6 +60,17 @@ contract SgxVerifier is EssentialContract, IVerifier {
     error SGX_INVALID_INSTANCES();
     error SGX_INVALID_PROOF();
 
+    /// @dev Modifier that ensures the caller is either taikoL1 or SgxAndZkVerifier.
+    modifier onlyTaikoOrCombinedVerifier() {
+        if (
+            msg.sender != resolve("taiko", true)
+                && msg.sender != resolve("tier_sgx_and_pse_zkevm", true)
```
dyxushuai, line 67:
```suggestion
                || msg.sender != resolve("tier_sgx_and_pse_zkevm", true)
```

Maybe you mean match one of these two addresses?

dyxushuai, line 67:
```suggestion
                || msg.sender != resolve("tier_sgx_and_pse_zkevm", true)
```
Do you mean match one of these two addresses?

adaki2004: Yes, i mean. So good as is.

IF NOT taiko AND NOT combined verifier either -> revert.

johntaiko (approved):

PR #15006, opened 2023-10-22
feat(repo): add a contract to handle token granting and unlocking

dantaik:
Addressing https://github.com/taikoxyz/taiko-mono/issues/14986

PR #15522, opened 2024-01-18
feat(protocol): enable strike price to token grants

dantaik:
Minimal change to add strike price support to TimelockTokenPool. All tests are migrated from Dani's PR https://github.com/taikoxyz/taiko-mono/pull/15506

### DIFF
--- a/packages/protocol/test/team/TimelockTokenPool.t.sol
+++ b/packages/protocol/test/team/TimelockTokenPool.t.sol
@@ -131,9 +131,7 @@ contract TestTimelockTokenPool is TaikoTest {
         (amountOwned, amountUnlocked, amountWithdrawn, amountToWithdraw, costToWithdraw) =
             pool.getMyGrantSummary(Alice);
 
-        // TODO(dani): who did you figured out this number? It's really hard to maintain
-        uint256 amount1 = 5_000_000_317_097_919_837_645;
-
+        uint256 amount1 = uint128(10_000e18) * uint64(block.timestamp - unlockStart) / unlockPeriod;
         uint256 expectedCost = amount1 / ONE_TKO_UNIT * strikePrice1;
 
         assertEq(amountOwned, 10_000e18);
@@ -217,7 +215,7 @@ contract TestTimelockTokenPool is TaikoTest {
         (amountOwned, amountUnlocked, amountWithdrawn, amountToWithdraw, costToWithdraw) =
             pool.getMyGrantSummary(Alice);
 
-        uint256 amount1 = 5_000_000_317_097_919_837_645;
+        uint256 amount1 = uint128(10_000e18) * uint64(block.timestamp - grantStart) / grantPeriod;
         uint256 expectedCost = amount1 / ONE_TKO_UNIT * strikePrice1;
 
         assertEq(amountOwned, amount1);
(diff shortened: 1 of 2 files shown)


packages/protocol/test/team/TimelockTokenPool.t.sol
```diff
@@ -136,62 +178,80 @@ contract TestTimelockTokenPool is TaikoTest {
...
+        (amountOwned, amountUnlocked, amountWithdrawn, amountToWithdraw, costToWithdraw) =
             pool.getMyGrantSummary(Alice);
+
+        expectedCost = (10_000e18 - amount1) / ONE_TKO_UNIT * strikePrice1;
         assertEq(amountOwned, 10_000e18);
         assertEq(amountUnlocked, 10_000e18);
         assertEq(amountWithdrawn, amount1);
-        assertEq(amountWithdrawable, 10_000e18 - amount1);
+        assertEq(amountToWithdraw, 10_000e18 - amount1);
+        // TODO(dani): the following assert fails
```
dantaik, line 242: @adaki2004 Please take a look at this TODO - uncommenting the next line will fail the test. I'm not sure it's a test issue or implementation issue.

packages/protocol/test/team/TimelockTokenPool.t.sol
```diff
@@ -72,62 +94,82 @@ contract TestTimelockTokenPool is TaikoTest {
...
+        assertEq(amountToWithdraw, 0);
+        assertEq(costToWithdraw, 0);
 
         vm.warp(unlockCliff + 1);
 
-        (amountOwned, amountUnlocked, amountWithdrawn, amountWithdrawable) =
+        (amountOwned, amountUnlocked, amountWithdrawn, amountToWithdraw, costToWithdraw) =
             pool.getMyGrantSummary(Alice);
+
+        // TODO(dani): who did you figured out this number? It's really hard to maintain
```
adaki2004, line 134: It was you who figured this out, not me !
https://github.com/taikoxyz/taiko-mono/pull/15006/commits/73f27713052f43b38aea72d874ae5e0cbe87d0b4

And I agree, it is really ahrd to maintain.

packages/protocol/test/team/TimelockTokenPool.t.sol
adaki2004: Checked quikcly - no issue on any end, it is a rounding error. Will check tomo where exactly that loss.
```
  expected cost: 49990000
  costToWithdraw: 50000000
```

adaki2004 (approved):

Brechtpd (approved):

adaki2004:
@dantaik One question to the concept with an example - if i understand it well. Let's say: 

1. Alice gets a grant 1, with a strike price 0.1 USD/token, e.g.: 100 tokens.
2. Alice gets a grant 2, with a strike price 1.5 USD/token, eg.: 50 token again.
3. Later, when Alic tokens are unlocked, she is about to withdraw tokens -> but the token price e.g.: is 0.4.

In this design there is no way for Alice to refuse (or not `withdraw`) grant 2, and pay for just grant 1 - because all happens in the same transaction so basically all or nothing. And in this case Alice would pay 10 + 75 = 80 for 200 tokens, which worth 80 on the market. Isn't it a possible scenario ?

PR #15528, opened 2024-01-19
fix(protocol): fix guardian prover bug

adaki2004:
The blelow code section.If transition is correct but contested AND guardian came with the (same) correct proof, this check would not allow it to execute properly.

```
            // When contestBond is zero for the current tier, it signifies
            // it's the top tier. In this case, it can overwrite existing
            // transitions without contestation.
            if (tran.blockHash == ts.blockHash && tran.signalRoot == ts.signalRoot) {
                revert L1_ALREADY_PROVED();
            }
```

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProving.sol
+++ b/packages/protocol/contracts/L1/libs/LibProving.sol
@@ -226,7 +226,7 @@ library LibProving {
             assert(tier.validityBond == 0);
 
             // It means prover is right (not the contester)
-            bool proverIsRight = tran.blockHash == ts.blockHash && tran.signalRoot == ts.signalRoot;
+            bool sameTransition = tran.blockHash == ts.blockHash && tran.signalRoot == ts.signalRoot;
             // We should outright prohibit the use of zero values for both
             // blockHash and signalRoot since, when we initialize a new
             // transition, we set both blockHash and signalRoot to 0.
@@ -249,7 +249,7 @@ library LibProving {
             ts.prover = msg.sender;
 
             if (ts.contester != address(0)) {
-                if (!proverIsRight) {
+                if (!sameTransition) {
                     // At this point we know that the contester was right
                     tko.transfer(ts.contester, ts.validityBond >> 2 + ts.contestBond);
                 }
--- a/packages/protocol/contracts/L1/libs/LibProving.sol
+++ b/packages/protocol/contracts/L1/libs/LibProving.sol
@@ -224,6 +224,9 @@ library LibProving {
 
         if (tier.contestBond == 0) {
             assert(tier.validityBond == 0);
+
+            // It means prover is right (not the contester)
+            bool proverIsRight = tran.blockHash == ts.blockHash && tran.signalRoot == ts.signalRoot;
             // We should outright prohibit the use of zero values for both
             // blockHash and signalRoot since, when we initialize a new
             // transition, we set both blockHash and signalRoot to 0.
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -224,12 +224,6 @@ library LibProving {
 
         if (tier.contestBond == 0) {
             assert(tier.validityBond == 0);
-            // When contestBond is zero for the current tier, it signifies
-            // it's the top tier. In this case, it can overwrite existing
-            // transitions without contestation.
```
cyberhorsey, line 229: because of this code:
```ts
 ts.blockHash = tran.blockHash;
            ts.signalRoot = tran.signalRoot;
            ts.prover = msg.sender;

            if (ts.contester != address(0)) {
                // At this point we know that the contester was right
                tko.transfer(ts.contester, ts.validityBond >> 2 + ts.contestBond);
                ts.contester = address(0);
                ts.validityBond = 0;
            }
            ```
            
            even though the contester was wrong, he will be rewarded, and the guardian prover will now be the prover of the block, and get the original provers reward? so anyone can contest any block, to steal proof money.

**Review by cyberhorsey — CHANGES_REQUESTED** [review]

File: packages/protocol/contracts/L1/libs/LibProving.sol (hunk 1 above)
Comment (cyberhorsey) [reply packages/protocol/contracts/L1/libs/LibProving.sol:229]: i think we need this:
```solidity
if (ts.contester != address(0) && tran.blockHash != ts.blockHash && tran.signalRoot != ts.signalRoot) {
                // At this point we know that the contester was right
                tko.transfer(ts.contester, ts.validityBond >> 2 + ts.contestBond);
                ts.contester = address(0);
                ts.validityBond = 0;
            } else {
               // we know the contester was wrong, penalizse them or something
            }
 ```

Comment (cyberhorsey) [reply packages/protocol/contracts/L1/libs/LibProving.sol:229]: also this block further down:
```ts
emit TransitionProved({
                blockId: blk.blockId,
                tran: tran,
                prover: msg.sender,
                validityBond: 0,
                tier: proof.tier
            });
            ```
            
i think we dont want to emit this event, in this case, we should have a new event like `ContestationDismissed` or something, in the case of the contester being wrong and the guardian prover proving the original proof is correct.

Comment (cyberhorsey) [reply packages/protocol/contracts/L1/libs/LibProving.sol:229]: And perhaps we should charge a user to stake something to contest, so they can be penalized when they contest incorrectly, and trying to attack/halt chain.

Comment (adaki2004) [reply packages/protocol/contracts/L1/libs/LibProving.sol:229]: > i think we need this:
> 
> ```solidity
> if (ts.contester != address(0) && tran.blockHash != ts.blockHash && tran.signalRoot != ts.signalRoot) {

> ```

Yes, almost.. :) We need to move that check one side inner, but good observation i changed it !

Comment (adaki2004) [reply packages/protocol/contracts/L1/libs/LibProving.sol:229]: > i think we dont want to emit this event, in this case, we should have a new event like `ContestationDismissed` or something, in the case of the contester being wrong and the guardian prover proving the original proof is correct.

Well, as everything is retrievable with the transaction(s) - like it is seen it is a tier guardian, i think it is fine as is, but can add that if it helps!

File: packages/protocol/contracts/L1/libs/LibProving.sol [hunk 2]
```diff
@@ -252,8 +249,10 @@ library LibProving {
             ts.prover = msg.sender;
 
             if (ts.contester != address(0)) {
-                // At this point we know that the contester was right
-                tko.transfer(ts.contester, ts.validityBond >> 2 + ts.contestBond);
+                if (!proverIsRight) {
```
dantaik, line 252:
```suggestion
                if (!sameTransition) {
```

dantaik (approved):

davidtaikocha (approved):

PR #15529, opened 2024-01-19
fix(protocol): refactor proving logics

dantaik:
Previous, if a transition is contested, the oracle prover cannot prove the transition is indeed valid.

See the new test: test_L1_GuardianProverProveTheOrigianlTransitionIsCorrect


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -222,217 +224,138 @@ library LibProving {
...
-            // Regardless of whether the previous prover or the contester
-            // emerges as the winner, we consistently erase the contest history
-            // to make this transition appear entirely new.
-            ts.prover = msg.sender;
-            ts.validityBond = tier.validityBond;
-            ts.contester = address(0);
-            ts.contestBond = 1; // to save gas
-            ts.timestamp = uint64(block.timestamp);
-            ts.tier = proof.tier;
+    /// @dev Chesk the msg.sender (the new prover) against the block's assigned prover.
```
dantaik, line 338:
```suggestion
    /// @dev Check the msg.sender (the new prover) against the block's assigned prover.
```

davidtaikocha (approved):

PR #15530, opened 2024-01-19
refactor(protocol): alternative proving logics

dantaik:
Trying to optimize the code to make it subjectively more readable.


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -78,7 +78,9 @@ library LibProving {
         returns (uint8 maxBlocksToVerify)
     {
         // Make sure parentHash is not zero
-        if (tran.parentHash == 0) revert L1_INVALID_TRANSITION();
+        if (tran.parentHash == 0 || tran.blockHash == 0 || tran.signalRoot == 0) {
```
dantaik, line 81: The only difference with the old implementation is that in all cases, blockHash and signalRoot cannot be zero.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -222,208 +228,136 @@ library LibProving {
...
+                    reward = ts.validityBond >> 2;
+                    tko.transfer(ts.contester, ts.contestBond + reward);
                 }
-                ts.contester = address(0);
-                ts.validityBond = 0;
+            } else {
+                if (sameTransition) revert L1_ALREADY_PROVED();
+                // Contest the existing transition and prove it to be invalid
+                reward = ts.validityBond >> 1;
+                ts.contestations += 1;
```
dantaik, line 263: this is something missed from the previous implantation and not tested.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -222,208 +226,140 @@ library LibProving {
...
-                    // the previous prover -- 1/4 each
-                    reward = ts.contestBond >> 2;
+                emit TransitionProved({
+                    blockId: blk.blockId,
+                    tran: tran,
+                    prover: msg.sender,
+                    validityBond: 0,
+                    tier: proof.tier
+                });
+            } else {
```
adaki2004, line 311: According to this new logic, there shall be an extra check (but please confirm) somewhere here. 

Because this [line](https://github.com/taikoxyz/taiko-mono/blob/29e82230d09080b263c6aabb937550c9fa7a63ca/packages/protocol/contracts/L1/libs/LibProving.sol#L243) will allow people to contest with lower tiers !

Example, please follow along:

So tiers are: 1.Optimistic, 2. Sgx, 3. ZK+Sgx, 4. Guardian

1. Bob proves a Zk+SGX block (because the default tier is 3).
2. Alice wants to contest the block and she submits a proof with SGX , and she will succeed to contest.

packages/protocol/contracts/L1/libs/LibProvingAlt.sol
```diff
@@ -0,0 +1,365 @@
...
+
+            if (returnLivenessBond) {
+                tko.transfer(blk.assignedProver, blk.livenessBond);
+                blk.livenessBond = 0;
+            }
+        }
+
+        bool sameTransition = tran.blockHash == ts.blockHash && tran.signalRoot == ts.signalRoot;
+
+        if (proof.tier > ts.tier) {
```
dantaik, line 243: @adaki2004 I created this file to live side-by-side with the previous implementation. 

You said there may be a bug (lower tier can contest a higher tier proof), according to the new test test_L1_ContestingWithLowerTierProofRevertspn, it's not possible.

packages/protocol/contracts/L1/libs/LibProvingAlt.sol
```diff
@@ -0,0 +1,365 @@
...
+        //
+        // It's obvious that proof verification is entirely decoupled from
+        // Taiko's core protocol.
+        {
+            address verifier = resolver.resolve(tier.verifierName, true);
+            // The verifier can be address-zero, signifying that there are no
+            // proof checks for the tier. In practice, this only applies to
+            // optimistic proofs.
+            if (verifier != address(0)) {
+                bool isContesting = proof.tier == ts.tier && tier.contestBond != 0;
```
adaki2004, line 212: Tiny thing, but GuardianVerifier does not check `isContesting` inside - as it just checks who is the prover (as indeed the "guardian_prover"). And in case it will in the future (i guess not) it is better to remove this check.

```suggestion
                bool isContesting = proof.tier == ts.tier;
```

adaki2004:
LGTM. Tried to make it a little bit more readable (tho this is subjective as you also pointed out), please have a look at on this [PR](https://github.com/taikoxyz/taiko-mono/pull/15540/files) and make a call if it worth to incorporate.

packages/protocol/contracts/L1/libs/LibProvingAlt.sol
dantaik: This change should be revert. Done.

adaki2004 (approved):
LGTM - imo you can use the alt as the 'normal' `LibProving.sol`

davidtaikocha (approved):

PR #15537, opened 2024-01-20
fix(protocol): disallow zero point value in KZG to avoid zero polynomial

dantaik:
@Brechtpd I'm not sure this is the right way to do it, take a look.

### DIFF
--- a/packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
+++ b/packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
@@ -63,12 +63,12 @@ contract PseZkVerifier is EssentialContract, IVerifier {
 
         bytes32 instance;
         if (ctx.blobUsed) {
+            PointProof memory pf = abi.decode(zkProof.pointProof, (PointProof));
+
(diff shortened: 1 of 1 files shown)


Brechtpd:
As far as I understand, disallowing it wouldn't allow commitment to blobs with all zeros, so that will be problematic because then we couldn't prove a block like that. So if we would want to make that impossible, I think we would already have to prevent the block to be submitted at proposal time and reject blocks being proposed with the KZG commitment set to the all-zero KZG commitment.

But I'm not sure if that even is a special case that would need to be handled inside the circuits. Also it's unlikely at this point that we would even use the PSE circuit for this at all so I don't think anything to worry about. For the zkVM circuits I think we should probably just use the normal way to verify the blob data instead of using this trick and measure the performance and only do the trick when it's really necessary, because ideally we just use the standard code to calculate the KZG commitment there.

Brechtpd:
We certainly cannot disallow it at the prover side because like the comment also says it would prevent the block from being verified. I don't think we have to change anything for reasons given above.

PR #15544, opened 2024-01-22
feat(protocol): Add possibility to filter grant withdrawals

adaki2004:
The background of this PR is the issue described [here](https://github.com/taikoxyz/taiko-mono/pull/15522#issuecomment-1900069175).

So this PR introduces feature that X (`Alice`) can activate (and withdraw) grants after one another (or the first 2 (cheapest) but not yet the 3rd, etc.) - how it makes sense.

Usually, this PR enforces (see `GRANT_PRICE_SHALL_INCREASE`) that :
- `strike price grant1` <= `strike price grant2`<= `strike price grant3` <= `strike price grant4` etc

Then Alice can decide she is willing to withdraw only the first 2, for the 3rd and 4th she: either does not have the money OR the market price is higher so does not worth to withdraw it.

This enforcement (`GRANT_PRICE_SHALL_INCREASE`) can be removed (if there is a good argument) and still the cap on the grants withdraw number is a nice feature. I just assumed usually the first grant is more 'generous' the the upcoming ones - from price perspective.

### DIFF
--- a/packages/protocol/contracts/team/TimelockTokenPool.sol
+++ b/packages/protocol/contracts/team/TimelockTokenPool.sol
@@ -17,6 +17,7 @@ pragma solidity 0.8.20;
 import "lib/openzeppelin-contracts/contracts/utils/cryptography/ECDSA.sol";
 import "lib/openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
 import "lib/openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
+import "../libs/LibMath.sol";
 import "../common/EssentialContract.sol";
 
 /// @title TimelockTokenPool
@@ -35,6 +36,7 @@ import "../common/EssentialContract.sol";
 /// - team members, advisors, etc.
 /// - grant program grantees
 contract TimelockTokenPool is EssentialContract {
+    using LibMath for uint256;
     using SafeERC20 for IERC20;
 
     struct Grant {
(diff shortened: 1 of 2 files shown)


dantaik:
@adaki2004 Your comment [here](https://github.com/taikoxyz/taiko-mono/pull/15522#issuecomment-1900069175) is totally valid, however, I'd prefer solving this issue with another approach: encourage our teammates to use different addresses to receive different grants.

maybe we can simplify the implementation by enforcing each address can only receive one single grant? Being simpler is better.

adaki2004:
> @adaki2004 Your comment [here](https://github.com/taikoxyz/taiko-mono/pull/15522#issuecomment-1900069175) is totally valid, however, I'd prefer solving this issue with another approach: encourage our teammates to use different addresses to receive different grants.
> 
> maybe we can simplify the implementation by enforcing each address can only receive one single grant? Being simpler is better.

Okay, i take it !
Either way (1 grant per address) or people could decide on it once they know the side-effects, i'm fine.

dantaik:
> > @adaki2004 Your comment [here](https://github.com/taikoxyz/taiko-mono/pull/15522#issuecomment-1900069175) is totally valid, however, I'd prefer solving this issue with another approach: encourage our teammates to use different addresses to receive different grants.
> > maybe we can simplify the implementation by enforcing each address can only receive one single grant? Being simpler is better.
> 
> Okay, i take it ! Either way (1 grant per address) or people could decide on it once they know the side-effects, i'm fine.

Lets do 1 grant per address - code will be simplified.

PR #15546, opened 2024-01-22
feat(relayer): Allow relayer to run in "target single transaction hash" mode

cyberhorsey:
This lets us help debug a potential issue with a bridge message, or manually process for easier testing.

### DIFF
--- a/packages/relayer/processor/process_single.go
+++ b/packages/relayer/processor/process_single.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"strings"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/taikoxyz/taiko-mono/packages/relayer/bindings/bridge"
 	"github.com/taikoxyz/taiko-mono/packages/relayer/pkg/queue"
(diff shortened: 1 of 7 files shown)


RogerLamTd (approved):

packages/relayer/processor/process_single.go
```diff
@@ -0,0 +1,60 @@
...
+
+	receipt, err := p.srcEthClient.TransactionReceipt(ctx, *p.targetTxHash)
+	if err != nil {
+		return err
+	}
+
+	for _, log := range receipt.Logs {
+		topic := log.Topics[0]
+
+		if topic == bridgeAbi.Events["MessageSent"].ID {
```
mask-pp, line 27: LGTM, but a 
nitpick it will be better to use ParseMessageSent func([reference link](https://github.com/taikoxyz/taiko-client/blob/bbe9ed764e00029f54baec9e7c51664ee36c489b/bindings/gen_guardian_prover.go#L907)).

mask-pp (approved):

packages/relayer/processor/process_single.go
cyberhorsey: I agree, I can update this PR tomorrow morning to use that instead.

cyberhorsey: done!

PR #15555, opened 2024-01-24
feat(protocol): enforce an invocation delay for bridged messages

dantaik:
Brecht previously proposed enhancing the Bridge's security by implementing a mandatory delay before message invocation. In this draft implementation, by default, the invocation delay is set to 0, allowing the Bridge to operate as usual. However, when configured with a non-zero delay, each message requires a two-step process (true for both `processMessage` and `recallMessage`). The first step verifies the message with a Merkel proof and timestamps it, while the second step is responsible for the actual message invocation. A challange is that the previous message.fee is designed for a one-step process, now with a two-step process, in theory we need two fees, one for each step, but this is not implemented in this PR as it will have data incompatibility issue.

We should deploy a bridge *watchdog* software which always monitors all `MessageReceived` events, ensuring that messages scheduled for invocation exist on their source chain. If a discrepancy is detected, the watchdog will use `pauseMessages` to halt the invocation of these messages and initiate an investigation.

Tests specific to the non-zero invocation delay feature will be added once we decide to adopt this enhancement.


Brechtpd:
Looks like indeed it does what I think it should do!

Optimism article for reference: https://blog.oplabs.co/two-step-withdrawals/

One thing to stress here is that the delay would only be enabled for L2->L1 bridging because everywhere else it is not required for safety (we have time to monitor and fix those transactions and revert them if needed).

One UX improvement that I think it is important: Currently it's only possible to commit to a message after the block has already finalized. However, this isn't needed. We can allow users to commit to the message with any given blockhash/signalroot. Then when the message is processed we check if the blockhash/signalroot is a valid one for a finalized block. The big benefit of this approach is that this allows the delay to work in parallel with the block finalization delay. In the current implementation, the delay is additive. So for example if the block finalization delay is 1 day and the bridge delay is 12 hours, as long as the user commits to his bridge transaction within the first 12 hours, no extra delay will be noticed by the user because for the user it still looks like a delay of 1 day. In the current implementation the withdrawal delay would be at minimum 36 hours.

Not sure how useful `pauseMessages` will be compared to just pausing the bridge completely because once somebody has exploited the bridge and we also mark directly which one, it will either quickly be exploited by others and so we won't be able to keep up with pausing individual messages, or there is no bug at all but still our monitoring software is broken so also not a great situation to continue with the bridging.

cyberhorsey:
just commenting here to say I'm watching the PR and the watchdog service would be an additional subcommand on the relayer, and when/if this PR is finalized, I will handle that part.

dantaik:
>  We can allow users to commit to the message with any given blockhash/signalroot. 

The current approach does have the downside that all delays are accumulative which impacts UX negatively. 

However, if we allow users to "register" messages without checking the signal is actually received using a verified signalRoot, then a spammer can register  tons of fake messages to spam the bridge watchdog. The watchdog cannot afford banning all these messages, but if one of them is not banned but get executed, then the whole design won't work.

dantaik:
I'm ok with pausing the entire bridge by the watchdog -- a hot wallet so it cannot be the owner address.
I think the pauseMessages function offers some flexibility. You can also see it as a way of deleting a message.

adaki2004:
> The first step verifies the message with a Merkel proof and timestamps it, while the second step is responsible for the actual message invocation. A challange is that the previous message.fee is designed for a one-step process, now with a two-step process, in theory we need two fees, one for each step, but this is not implemented in this PR as it will have data incompatibility issue.

Actually it can be done without modifying the interface. Fee is a uint256.
So lower/upper 16bytes (128bits) are used by the first/second one. On the UI it just needs to use 2 clear variables, and upshift one to the upper uint128 section - and handle it in the code properly.
 
> when/if this PR is finalized, I will handle that part

Just a note, 2nd invocation does not require a proof so it can be empty significantly reducing the gas cost.

Brechtpd:
> However, if we allow users to "register" messages without checking the signal is actually received using a verified signalRoot, then a spammer can register tons of fake messages to spam the bridge watchdog. The watchdog cannot afford banning all these messages, but if one of them is not banned but get executed, then the whole design won't work.

Spamming is very expensive though, while the watchdog just has to keep a map (blockhash -> list of all message acks for this blockhash) and then only verify the messages for blockhashes that are known by the watchdog in its own L2 node. Messages without known blockhash can be pruned after a day or so because it's impossible to predict future blockhashes that much in advance so they cannot be valid (curretnly it's even basically impossible to predict after more than 1 L1 block time).

dantaik:
@Brechtpd @dantaik @cyberhorsey @KorbinianK  this PR is ready for you to review now. We need to make a decision quickly whether we should adopt this approach. 👍 if you think we should.

packages/protocol/genesis/GenerateGenesis.g.sol
```diff
@@ -141,6 +141,7 @@ contract TestGenerateGenesis is Test, AddressResolver {
     }
 
     function testSingletonBridge() public {
+        // TODO: the next line reverts with RESOLVER_ZERO_ADDR(1, "bridge")
```
dantaik, line 144: @davidtaikocha can you look at this issue?

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -31,21 +31,38 @@ contract Bridge is EssentialContract, IBridge {
...
+        // The time a message is marked as received on the destination chain
+        uint64 receivedAt;
+        // The address that can execute the message after the invocation delay without an extra
+        // delay.
+        // For a failed message, preferredExecutor's value doesn't matter as only the owner can
+        // invoke the message.
+        address preferredExecutor;
+    }
+
+    uint256 public constant INVOCATION_EXTRA_DELAY = 10 minutes;
```
dantaik, line 49: delete it

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -193,58 +251,78 @@ contract Bridge is EssentialContract, IBridge {
...
+                receivedAt: uint64(block.timestamp),
+                preferredExecutor: message.gasLimit == 0 ? message.owner : msg.sender
+            });
+        }
 
-            if (_invokeMessageCall(message, msgHash, gasLimit)) {
-                status = Status.DONE;
-            } else {
-                status = Status.RETRIABLE;
+        (uint256 invocationDelay, uint256 invocationExtraelay) = getInvocationDelays();
```
adaki2004, line 270: Small typo:
```suggestion
        (uint256 invocationDelay, uint256 invocationExtraDelay) = getInvocationDelays();
```

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -138,42 +180,58 @@ contract Bridge is EssentialContract, IBridge {
...
 
-        ISignalService signalService = ISignalService(resolve("signal_service", false));
-
-        if (!signalService.isSignalSent(address(this), msgHash)) revert B_MESSAGE_NOT_SENT();
-
-        bool received = _proveSignalReceived(
-            signalService, _signalForFailedMessage(msgHash), message.destChainId, proof
-        );
-        if (!received) revert B_NOT_FAILED();
+        bool isMessageNew = messageReception[msgHash].receivedAt == 0;
```
Brechtpd, line 185: `isMessageNew` a bit of a confusing name because it can both mean a message with the status NEW (checked just before this), or a message that already has its message hash committed. Maybe `isMessageReceived` or `isMessageCommitted` (but this would require a rename of `messageReception` to `messageCommitments` to make that work well I think).

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -279,15 +357,11 @@ contract Bridge is EssentialContract, IBridge {
...
-            );
+            _updateMessageStatus(msgHash, Status.DONE);
         } else if (isLastAttempt) {
-            // Update the message status to "FAILED"
-            _updateMessageStatus(
-                ISignalService(resolve("signal_service", false)), msgHash, Status.FAILED
-            );
+            _updateMessageStatus(msgHash, Status.FAILED);
+        } else {
+            // this transaction will revert
```
Brechtpd, line 364: Will it? Or does this comment just mean the message will remain retriable?

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -367,11 +441,42 @@ contract Bridge is EssentialContract, IBridge {
...
+    /// @return invocationExtraelay The extra delay in second (to be added to invocationDelay) if
+    /// the transactor is not the preferredExecutor who acknowledged the reception of this message.
+    function getInvocationDelays()
+        public
+        view
+        virtual
+        returns (uint256 invocationDelay, uint256 invocationExtraelay)
+    {
+        invocationDelay = 0;
+        invocationExtraelay = 0;
```
Brechtpd, line 458: This code will use `block.chainid` here to select the correct values? If so, maybe the L1 values can still be defined as public constants in this smart contract, this function then just selects them when on L1, otherwise this function just returns zeros.

Brechtpd (commented):
I think we should definitely use this.

I would of course prefer the parallelizable method that supplies the blockhash immediately when committing to the Merkle proof because of UX improvements (faster withdrawals and less manual steps to go through, now it requires three distinct steps for users to do for withdrawals: withdrawal message tx, wait hours, commit message tx, wait hours, process withdrawal tx). But if this the solution we go for, then all good of course.

packages/protocol/contracts/bridge/Bridge.sol
Brechtpd: I kind of like `messageProofs`/`isMessageProven` more than `messageCommitments` that I first thought would be a good name. If nobody else comes up with a better name/has any objections in the next couple of hours I'll do the rename in this PR!

packages/protocol/contracts/bridge/Bridge.sol
Brechtpd: Ah I see. I guess the benefit of the `block.chainid` is no different contracts needed, but indeed not much difference.

adaki2004:
> One UX improvement that I think it is important: Currently it's only possible to commit to a message after the block has already finalized. However, this isn't needed. We can allow users to commit to the message with any given blockhash/signalroot. Then when the message is processed we check if the blockhash/signalroot is a valid one for a finalized block. The big benefit of this approach is that this allows the delay to work in parallel with the block finalization delay. In the current implementation, the delay is additive. So for example if the block finalization delay is 1 day and the bridge delay is 12 hours, as long as the user commits to his bridge transaction within the first 12 hours, no extra delay will be noticed by the user because for the user it still looks like a delay of 1 day. In the current implementation the withdrawal delay would be at minimum 36 hours.

I think we are indeed lacking this feature, right @dantaik ? 🤔

Brechtpd:
Right! So I think Reception will additionally store the supplied `signalRoot` and `blockID`, set in the first step. The `signalRoot` is used as is to verify the given merkle proof.

Then the second step we just verify that at `blockID` the stored `signalRoot` matches the one return by the cross chain sync snippet.

dantaik:
> Right! So I think Reception will additionally store the supplied `signalRoot` and `blockID`, set in the first step. The `signalRoot` is used as is to verify the given merkle proof.
> 
> Then the second step we just verify that at `blockID` the stored `signalRoot` matches the one return by the cross chain sync snippet.

Brecht's suggestion does have the benefit of having a shorter delay between message sending and execution with a potential drawback of a higher cost (due to a larger receipt). However, I assume there will be a lot more changes in SignalService or new code to be added elsewhere. That's why I am reluctant to adopt *given our current launch schedule*.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -544,11 +544,15 @@ contract Bridge is EssentialContract, IBridge {
         view
         returns (bool)
     {
-        return signalService.proveSignalReceived({
+        try signalService.proveSignalReceived({
```
adaki2004, line 547: Common misconception with solidity which is coming from the syntax of `try ... catch` is that it will catches every reverts, but it will not, by default catches custom error inside the expression.

See a good explanation here:
https://forum.soliditylang.org/t/call-for-feedback-the-future-of-try-catch-in-solidity/1497

If we want to somehow propoerly try-catch custom errors, here is a pseudo:
```
// A low level call to the contract
 (bool success, bytes memory returnData) =
            address(signalService).call( 
                abi.encodePacked(
                    signalService.proveSignalReceived.selector,
                    abi.encode(srcChainId, app,signal,proof)
                )
            );
        if (success) { 
            // proveSignalReceived completed successfully (did not revert)
            return true;
        } else { 
            // proveSignalReceived reverted.
            return false;
        }
```

Brechtpd: I would say in our case try/catch would be the one to use though, unless I misunderstand something. We're only interested in catching errors in the called function, and the additional checks on the smart contract existing before the call should revert because it means that the smart contract is not setup correctly, so still reverting on those would be the expected behavior.

Of course in the end they are almost the same, and should be fine to go with both.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -138,44 +179,64 @@ contract Bridge is EssentialContract, IBridge {
...
-        if (isMessageRecalled[msgHash]) revert B_RECALLED_ALREADY();
 
-        address signalService = resolve("signal_service", false);
+        if (messageStatus[msgHash] != Status.NEW) revert B_STATUS_MISMATCH();
 
-        if (!ISignalService(signalService).isSignalSent(address(this), msgHash)) {
-            revert B_MESSAGE_NOT_SENT();
-        }
+        uint64 receivedAt = proofReceipt[msgHash].receivedAt;
+        bool isMessageNotProven = receivedAt == 0;
```
Brechtpd, line 186: [Double negations are not not bad](https://cleankotlin.nl/blog/double-negations)

Just giving the reason why I changed it to the positive case when I did the renaming, I think for most people easier to think about the positive case instead of a name being the negative case.

adaki2004: yeah i was searching for the `if( ! xxx)` too :), tho no strong opinion on that.

packages/protocol/contracts/bridge/Bridge.sol
adaki2004: > Of course in the end they are almost the same, and should be fine to go with both.

You might (most prob.) right here, but i still prefer a solution we already 'know' from `_invokeMessage`. The reason i dig up that article is, because i know at one point in time we struggled with `try..catch` (maybe was misused (?)) but we even removed it.

Anyways as it is ambigous - hence diff. behavior in other languages - i still prefer this one.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: > [Double negations are not not bad](https://cleankotlin.nl/blog/double-negations)
> 
> Just giving the reason why I changed it to the positive case when I did the renaming, I think for most people easier to think about the positive case instead of a name being the negative case.

Changed back since you prefer the original naming.

Brechtpd (approved):

adaki2004:
@KorbinianK @cyberhorsey if you have any questions let me know. 
TLDR here so we can refer to later:
This will require some (ui/relayer) changes like: 
- delay enabled or not ?  (L2 most probably not, L1 i guess will have it)
- if delay enabled calling "claim" (aka `processMessage`) twice and in the right time window, etc.

adaki2004 (approved):

PR #15559, opened 2024-01-24
feat(protocol): enable remote attestation in SGX prover

adaki2004:
On-chain attestation verification is now supported in the protocol

- Currently the verifyProof() works the same way as it should
- registerInstance() is a permissionless function to allow registering instances with a one-time attestation verification method


dantaik:
> On-chain attestation verification is now supported in the protocol
> 
> * Currently the verifyProof() works the same way as it should
> * registerInstance() is a permissionless function to allow registering instances with a one-time attestation verification method

If we launch SGX proofs with RA, I want the new proofs as a higher tier proofs than the current SGX proofs without RA. The new code is more complicated than I expected. Testing them side-by-side seems to be a good idea.

adaki2004:
> > On-chain attestation verification is now supported in the protocol
> > 
> > * Currently the verifyProof() works the same way as it should
> > * registerInstance() is a permissionless function to allow registering instances with a one-time attestation verification method
> 
> If we launch SGX proofs with RA, I want the new proofs as a higher tier proofs than the current SGX proofs without RA. The new code is more complicated than I expected. Testing them side-by-side seems to be a good idea.

Nothing changed at all (in terms of `verifyProof()`) so i see no reason having 2 SGX proofs - because they are exactly the same.
The only change is the on-chain attestation - which is done once per instance.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -19,6 +19,8 @@ import "../../common/EssentialContract.sol";
 import "../../thirdparty/LibBytesUtils.sol";
 import "../ITaikoL1.sol";
 import "./IVerifier.sol";
+import { IAttestation } from "../../thirdparty/onchainRA/interfaces/IAttestation.sol";
```
dantaik, line 22: as you can see in other files, we basically switched away from using `import {...} from xxx` to `import xxx` unless you want to rename the imported contracts

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -77,37 +81,32 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
-                newInstance,
-                extraInstances
-            )
-        );
-        address oldInstance = ECDSA.recover(signedHash, signature);
-        if (!_isInstanceValid(id, oldInstance)) revert SGX_INVALID_INSTANCE();
+        address automataDcapAttestation = (resolve("automata_dcap_attestation", true));
 
-        _replaceInstance(id, oldInstance, newInstance);
+        // Still possible to be backward compatible and have the onlyOwner tpye of method registering instances.
```
dantaik, line 95:
```suggestion
        // Still possible to be backward compatible and only the owner can register instances.
```

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -77,37 +81,32 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
+                revert SGX_INVALID_ATTESTATION();
+            }
+
+            address[] memory _address = new address[](1);
+            _address[0] = address(bytes20(LibBytesUtils.slice(attestation.localEnclaveReport.reportData, 0, 20)));
+
+            return _addInstances(_address)[0];
+        }
+        // A special return variable, signaling that this method is not supported at the moment, because a contract with name "automata_dcap_attestation" not deployed.
+        return type(uint256).max;
```
dantaik, line 109: the code can be improved using:

```
if (automataDcapAttestation == address(0)) {
   return type(uint256).max;
}
// logics for automataDcapAttestation not zero.
```

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -77,37 +81,32 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
-        if (!_isInstanceValid(id, oldInstance)) revert SGX_INVALID_INSTANCE();
+        address automataDcapAttestation = (resolve("automata_dcap_attestation", true));
 
-        _replaceInstance(id, oldInstance, newInstance);
+        // Still possible to be backward compatible and have the onlyOwner tpye of method registering instances.
+        if (automataDcapAttestation != address(0)) {
+            (bool verified, ) = IAttestation(automataDcapAttestation).verifyParsedQuote(attestation);
 
-        ids = _addInstances(extraInstances);
+            if(!verified) {
```
dantaik, line 99: shorten this as one line.

packages/protocol/contracts/thirdparty/onchainRA/lib/QuoteV3Auth/V3Struct.sol
```diff
@@ -0,0 +1,79 @@
...
+    struct ParsedECDSAQuoteV3AuthData {
+        bytes ecdsa256BitSignature; // 64 bytes
+        bytes ecdsaAttestationKey; // 64 bytes
+        EnclaveReport pckSignedQeReport; // 384 bytes
+        bytes qeReportSignature; // 64 bytes
+        ParsedQEAuthData qeAuthData;
+        ParsedCertificationData certification;
+    }
+
+    struct ParsedV3QuoteStruct {
```
dantaik, line 74: ParsedV3QuoteStruct => ParsedV3Quote

packages/protocol/contracts/thirdparty/onchainRA/utils/SigVerifyLib.sol
```diff
@@ -0,0 +1,119 @@
...
+import "./RsaVerify.sol";
+import "./BytesUtils.sol";
+
+// Library for verifying signatures
+// Supports verifying signatures with the following algorithms:
+// - RS256
+// - ES256
+// - RS1
+
+contract SigVerifyLib is ISigVerifyLib {
```
dantaik, line 14: this is not a lib, I'd suggest to remove "Lib" from both `SigVerifyLib` and `ISigVerifyLib`.

packages/protocol/contracts/thirdparty/onchainRA/interfaces/ISigVerifyLib.sol
```diff
@@ -0,0 +1,62 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+interface ISigVerifyLib {
```
dantaik, line 4: A interface for a lib?

packages/protocol/contracts/thirdparty/onchainRA/interfaces/ISigVerifyLib.sol
```diff
@@ -0,0 +1,62 @@
...
+        CertSigAlgorithm sigAlg;
+    }
+
+    enum Algorithm {
+        RS256,
+        ES256,
+        RS1
+    }
+
+    function verifyAttStmtSignature(bytes memory tbs, bytes memory signature, PublicKey memory publicKey, Algorithm alg)
```
dantaik, line 36: can we have a full name? verifyAttestationStatementSignature? If Stmt is stamement, where is the Statement struct defined, or what's the different between an "Attestation" and an "AttestationStatement"?

dantaik: Need a better name for `tbs`

packages/protocol/contracts/thirdparty/onchainRA/AutomataDcapV3Attestation.sol
```diff
@@ -0,0 +1,651 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
dantaik, line 2: Are we using the code from the head commit of https://github.com/automata-network/automata-dcap-v3-attestation/tree/main/contracts/interfaces?

Can we try to also install it as a sub module rather than copying its code over?

dantaik (changes requested):

dantaik:
I'm trying to optimize the submodule, will mark this PR ready to be reviewed later.

packages/protocol/test/L1/TaikoL1TestBase.sol
```diff
@@ -24,6 +24,17 @@ abstract contract TaikoL1TestBase is TaikoTest {
     GuardianProver public gp;
     TaikoA6TierProvider public cp;
     Bridge public bridge;
+    
+    // For SGX remote attestation
+    AutomataDcapV3Attestation attestation;
```
dantaik, line 29: we should not add subclass variables to the base, the base will get larger and larger...

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -134,7 +134,7 @@ library LibProposing {
 
             if (params.blobHash != 0) {
                 // We try to reuse an old blob
-                if (!isBlobReusable(state, config, params.blobHash)) {
+                if (isBlobReusable(state, config, params.blobHash)) {
```
dantaik, line 137: revert this change

.gitmodules
```diff
@@ -14,3 +14,9 @@
 	path = packages/protocol/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
 	branch = chore/v1.5.1
+[submodule "packages/protocol/lib/solady"]
```
dantaik, line 17: we still need to specify branch, tag, or commit

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -193,7 +193,9 @@ library LibProposing {
         // of multiple Taiko blocks being proposed within a single
         // Ethereum block, we must introduce a salt to this random
         // number as the L2 mixHash.
-        meta.difficulty = keccak256(abi.encodePacked(block.prevrandao, b.numBlocks, block.number));
+        unchecked {
+            meta.difficulty = meta.blobHash ^ bytes32(block.prevrandao * b.numBlocks * block.number);
```
dantaik, line 197: revert this change

packages/protocol/contracts/L1/provers/Guardians.sol
```diff
@@ -94,9 +94,8 @@ abstract contract Guardians is EssentialContract {
             _approvals[version][hash] |= 1 << (id - 1);
         }
 
-        uint256 _approval = _approvals[version][hash];
-        approved = isApproved(_approval);
-        emit Approved(operationId, _approval, approved);
+        approved = isApproved(_approvals[version][hash]);
```
dantaik, line 97: revert

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -199,7 +199,7 @@ contract TaikoL2 is CrossChainOwned, TaikoL2Signer, ICrossChainSync {
     /// block id is greater than or equal to the current block number.
     function getBlockHash(uint64 blockId) public view returns (bytes32) {
         if (blockId >= block.number) return 0;
-        if (blockId + 256 >= block.number) return blockhash(blockId);
+        if (blockId >= block.number - 256) return blockhash(blockId);
```
dantaik, line 202: also revert

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -40,20 +40,17 @@ contract Bridge is EssentialContract, IBridge {
     mapping(bytes32 msgHash => bool recalled) public isMessageRecalled;
     mapping(bytes32 msgHash => Status) public messageStatus; // slot 3
     Context private _ctx; // // slot 4,5,6pnpm
-    mapping(address => bool) public addressBanned;
```
dantaik, line 43: undo this change

packages/protocol/contracts/team/TimelockTokenPool.sol
```diff
@@ -186,7 +186,7 @@ contract TimelockTokenPool is EssentialContract {
         totalCostPaid += costToWithdraw;
 
         IERC20(taikoToken).transferFrom(sharedVault, to, amountToWithdraw);
-        IERC20(costToken).safeTransferFrom(recipient, sharedVault, costToWithdraw);
+        IERC20(costToken).transferFrom(recipient, sharedVault, costToWithdraw);
```
dantaik, line 189: undo changes in this file

packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
```diff
@@ -43,6 +43,6 @@ contract ERC20Airdrop is MerkleClaimable {
 
     function _claimWithData(bytes calldata data) internal override {
         (address user, uint256 amount) = abi.decode(data, (address, uint256));
-        IERC20(token).transferFrom(vault, user, amount);
+        IERC20Upgradeable(token).transferFrom(vault, user, amount);
```
dantaik, line 46: undo changes

packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
```diff
@@ -14,7 +14,7 @@
 
 pragma solidity 0.8.20;
 
-import "lib/openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+import "lib/openzeppelin-contracts-upgradeable/contracts/token/ERC20/ERC20Upgradeable.sol";
```
dantaik, line 17: undo changes

packages/protocol/contracts/thirdparty/automata-attestation/lib/EnclaveIdStruct.sol
```diff
@@ -0,0 +1,28 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
dantaik, line 2: for all new files, please use the exact pragma as we used in all existing solidity files.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -256,27 +256,22 @@ contract ERC20Vault is BaseVault {
     {
         checkRecallMessageContext();
 
-        (CanonicalERC20 memory ctoken,,, uint256 amount) =
+        (, address token,, uint256 amount) =
```
dantaik, line 259: undo changes

packages/protocol/contracts/tokenvault/ERC721Vault.sol
```diff
@@ -234,7 +234,7 @@ contract ERC721Vault is BaseNFTVault, IERC721ReceiverUpgradeable {
                 });
 
                 for (uint256 i; i < op.tokenIds.length; ++i) {
-                    t.safeTransferFrom(user, address(this), op.tokenIds[i]);
+                    t.transferFrom(user, address(this), op.tokenIds[i]);
```
dantaik, line 237: undo changes

packages/protocol/contracts/tokenvault/adaptors/USDCAdaptor.sol
```diff
@@ -19,7 +19,7 @@ import "../BridgedERC20Base.sol";
 interface IUSDC {
     function burn(uint256 amount) external;
     function mint(address to, uint256 amount) external;
-    function transferFrom(address from, address to, uint256 value) external returns (bool);
+    function transferFrom(address from, address to, uint256 amount) external;
```
dantaik, line 22: undo changes

packages/protocol/script/SetDcapParams.s.sol
```diff
@@ -0,0 +1,91 @@
...
+import "forge-std/Script.sol";
+import "forge-std/console2.sol";
+
+import "../contracts/thirdparty/automata-attestation/AutomataDcapV3Attestation.sol";
+import "../contracts/L1/verifiers/SgxVerifier.sol";
+import "../test/automata-attestation/utils/DcapTestUtils.t.sol";
+import "../test/automata-attestation/utils/V3JsonUtils.t.sol";
+
+contract SetDcapParams is Script, DcapTestUtils, V3JsonUtils {
+    //@todo: Wand Yue -> after deployed with DeployOnL1, you have the parameters, fill in the proper
```
dantaik, line 26: todo should in format: "// TODO(Yue Wang): do something"

packages/protocol/script/config_dcap_sgx_verifier.sh
```diff
@@ -0,0 +1,31 @@
+PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 \
```
dantaik, line 1: what is this private key???

packages/protocol/test/L1/TaikoL1LibProvingWithTiers.t.sol
```diff
@@ -172,9 +172,9 @@ contract TaikoL1LibProvingWithTiers is TaikoL1TestBase {
             // This proof cannot be verified obviously because of
             // signalRoot instead of blockHash
             uint16 minTier = meta.minTier;
+
```
dantaik, line 175: revert changes

packages/protocol/test/L1/TaikoL1TestBase.sol
```diff
@@ -68,6 +68,8 @@ abstract contract TaikoL1TestBase is TaikoTest {
             })
         );
 
+        vm.warp(1_695_435_682);
```
dantaik, line 71: please confirm this change is necessary.

packages/protocol/test/TaikoTest.sol
```diff
@@ -42,7 +42,17 @@ import "../contracts/test/erc20/FreeMintERC20.sol";
 import "./DeployCapability.sol";
 import "./HelperContracts.sol";
 
-abstract contract TaikoTest is Test, DeployCapability {
+// For SGX remote attestation
```
dantaik, line 45: Do not add SGX specific settings in this file, you can create a TaikoSGXTest.sol which uses TaikoTest as its parent.

dantaik (changes requested):

packages/protocol/test/L1/TaikoL1TestBase.sol
adaki2004: It is necessary, this is a timestamp from which the attestation is valid.But I'm placing it to the SGX relevant tests.

packages/protocol/script/config_dcap_sgx_verifier.sh
adaki2004: I guess smtmfft used private key.

.gitmodules
```diff
@@ -14,3 +14,10 @@
 	path = packages/protocol/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
 	branch = chore/v1.5.1
```
dantaik, line 16: tag: v1.5.1

packages/protocol/test/TaikoTest.sol
```diff
@@ -72,7 +72,8 @@ abstract contract TaikoTest is Test, DeployCapability {
     address internal Zachary = randAddress();
     address internal SGX_X_0 = vm.addr(0x4);
     address internal SGX_X_1 = vm.addr(0x5);
-    address internal SGX_Y = randAddress();
+    address internal SGX_Y =
+        vm.addr(0x9b1bb8cb3bdb539d0d1f03951d27f167f2d5443e7ef0d7ce745cd4ec619d3dd7);
```
dantaik, line 76: move 3 SGX variable to sub-contract

dantaik (changes requested):

.gitmodules
adaki2004: Can you try to incorporate them (and test) ? I suck so much today morning and yesterday with submodules 😄

.gitmodules
dantaik: solady still missing a tag/commit/branch.

packages/protocol/script/config_dcap_sgx_verifier.sh
dantaik, line 1: @smtmfft can you take a look and remove the private key?

packages/protocol/foundry.toml
```diff
@@ -22,6 +22,7 @@ fs_permissions = [
   { access = "read-write", path = "./deployments" },
   { access = "read", path = "./test" },
   { access = "read", path = "./genesis" },
+  { access = "read", path = "./test/onchainRA/assets/0923"},
```
dantaik, line 25: I think "onchainRA" should be repalced

dantaik (approved):

packages/protocol/contracts/thirdparty/automata-attestation/interfaces/IAttestation.sol
```diff
@@ -0,0 +1,11 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { V3Struct } from "../lib/QuoteV3Auth/V3Struct.sol";
+
+interface IAttestation {
+    function verifyAttestation(bytes calldata data) external returns (bool);
```
smtmfft, line 7: I think we can remove this one from the interface as it costs more gas and actually never be called now. 
Even the code can be also removed, however, let's keep the code for merely reference now, and clean it up after we  have enough tests.

dantaik (approved):

smtmfft (approved):

packages/protocol/contracts/thirdparty/automata-attestation/interfaces/IAttestation.sol
adaki2004: Maybe best to keep original interface ? Does not harm.

fyi: also added a thirdparty reference readme.md in contracts/thirdparty, please check if that is correct or i should remove you repo's link.

smtmfft: No harm, I will create another PR to move sth (will be) deprecated to test sol. Feels like a advantage of foundry as you can alway keep the last version in test because it is still sol.

packages/protocol/contracts/thirdparty/automata-attestation/utils/BytesUtils.sol
```diff
@@ -0,0 +1,400 @@
+// SPDX-License-Identifier: BSD 2-Clause License
+pragma solidity 0.8.24;
+
+// Inspired by ensdomains/dnssec-oracle - BSD-2-Clause license
+// https://github.com/ensdomains/dnssec-oracle/blob/master/contracts/BytesUtils.sol
+
+library BytesUtils {
```
Brechtpd, line 7: Seems some inefficient code here, could be worthwhile to improve some of these and see if it actually makes a difference or not at some point (not important now).

packages/protocol/script/SetDcapParams.s.sol
```diff
@@ -0,0 +1,89 @@
...
+    bytes32 public mrEnclave = vm.envBytes32("MR_ENCLAVE");
+    bytes32 public mrSigner = vm.envBytes32("MR_SIGNER");
+
+    function run() external {
+        require(ownerPrivateKey != 0, "PRIVATE_KEY not set");
+        require(dcapAttestationAddress != address(0), "ATTESTATION_ADDRESS not set");
+
+        vm.startBroadcast(ownerPrivateKey);
+
+        // all in one
```
Brechtpd, line 41: What does this mean? :)

packages/protocol/script/config_dcap_sgx_verifier.sh
```diff
@@ -0,0 +1,20 @@
...
+QEID_PATH=/test/automata-attestation/assets/0923/identity.json \
+V3_QUOTE_PATH=/test/automata-attestation/assets/0923/v3quote.json \
+MR_ENCLAVE=0xdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef \
+MR_SIGNER=0xdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef \
+forge script script/SetDcapParams.s.sol:SetDcapParams \
+    --fork-url http://localhost:8545 \
+    --broadcast \
+    --ffi \
+    -vvvv \
+    --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 \ #foundry test key
```
Brechtpd, line 19: This private key always needs to match `PRIVATE_KEY`? If so, can we reuse `PRIVATE_KEY` here?

packages/protocol/script/SetDcapParams.s.sol
```diff
@@ -0,0 +1,89 @@
...
+            parseEnclaveIdentityJson(enclaveIdJson);
+        AutomataDcapV3Attestation(dcapAttestationAddress).configureQeIdentityJson(parsedEnclaveId);
+        console.log("qeIdParsedSuccess: %s", qeIdParsedSuccess);
+    }
+
+    function configureTcbInfoJson() internal {
+        string memory tcbInfoJson = vm.readFile(string.concat(vm.projectRoot(), tcbInfoPath));
+        (bool tcbParsedSuccess, TCBInfoStruct.TCBInfo memory parsedTcbInfo) =
+            parseTcbInfoJson(tcbInfoJson);
+        // string memory fmspc = "00606a000000";
```
Brechtpd, line 71: Can be removed?

packages/protocol/test/L1/SgxVerifier.t.sol
```diff
@@ -3,12 +3,68 @@ pragma solidity 0.8.24;
...
+            parseTcbInfoJson(tcbInfoJson);
+        require(tcbParsedSuccess, "tcb parsed failed");
+        attestation.configureTcbInfoJson(fmspc, parsedTcbInfo);
+
+        (bool qeIdParsedSuccess, EnclaveIdStruct.EnclaveId memory parsedEnclaveId) =
+            parseEnclaveIdentityJson(enclaveIdJson);
+        require(qeIdParsedSuccess, "qeid parsed failed");
+        attestation.configureQeIdentityJson(parsedEnclaveId);
+
+        registerAddress("automata_dcap_attestation", address(attestation));
```
Brechtpd, line 65: Could some of this setup code be shared with the `SetDcapParams` script? Seems like that would be ideal because then the setup script is also tested, currently they are different code paths so have to keep them both up to date (so easy to forget).

packages/protocol/test/L1/SgxVerifier.t.sol
```diff
@@ -3,12 +3,68 @@ pragma solidity 0.8.24;
...
     function deployTaikoL1() internal override returns (TaikoL1) {
         return
             TaikoL1(payable(deployProxy({ name: "taiko", impl: address(new TaikoL1()), data: "" })));
     }
 
+    function setUp() public override {
+        // Call the TaikoL1TestBase setUp()
+        super.setUp();
+
+        vm.warp(1_695_435_682);
```
Brechtpd, line 42: Maybe add a short comment of the reason why this is done, similar to what is done in `AutomataDcapV3AttestationTest`.

packages/protocol/test/automata-attestation/AutomataDcapV3AttestationTest.t.sol
```diff
@@ -0,0 +1,110 @@
...
+        string memory fmspc = "00606a000000";
+        (bool tcbParsedSuccess, TCBInfoStruct.TCBInfo memory parsedTcbInfo) =
+            parseTcbInfoJson(tcbInfoJson);
+        require(tcbParsedSuccess, "tcb parsed failed");
+        attestation.configureTcbInfoJson(fmspc, parsedTcbInfo);
+
+        (bool qeIdParsedSuccess, EnclaveIdStruct.EnclaveId memory parsedEnclaveId) =
+            parseEnclaveIdentityJson(enclaveIdJson);
+        require(qeIdParsedSuccess, "qeid parsed failed");
+        attestation.configureQeIdentityJson(parsedEnclaveId);
```
Brechtpd, line 74: Same here as before, I think would be nice to share this set up code between the different places.

packages/protocol/test/automata-attestation/utils/DcapTestUtils.t.sol
```diff
@@ -0,0 +1,265 @@
...
+        pure
+        returns (bool success, TCBInfoStruct.TCBInfo memory tcbInfo)
+    {
+        JSONParserLib.Item memory root = JSONParserLib.parse(tcbInfoJsonStr);
+        JSONParserLib.Item[] memory children = root.children();
+        JSONParserLib.Item[] memory tcbInfoObj;
+
+        uint256 tcbInfoIndex = INDEX_ERROR;
+
+        for (uint256 x = 0; x < root.size(); x++) {
```
Brechtpd, line 30: `x` as the iterator variable? Bold move.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -77,37 +81,29 @@ contract SgxVerifier is EssentialContract, IVerifier {
         ids = _addInstances(_instances);
     }
 
-    /// @notice Adds SGX instances to the registry by another SGX instance.
```
Brechtpd, line 80: So we won't support SGX instances verifying other SGX instances anymore? It's much cheaper, but of course does require additional code to do it in SGX so maybe not worth it anymore?

Depending on our trust in the onchain code, verifying the RA in a smart contract + SGX may also be worth it from a security perspective, but of course again only if we think it's worth it to support the SGX verification path as well.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -77,37 +81,29 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
-            )
-        );
-        address oldInstance = ECDSA.recover(signedHash, signature);
-        if (!_isInstanceValid(id, oldInstance)) revert SGX_INVALID_INSTANCE();
+        address automataDcapAttestation = (resolve("automata_dcap_attestation", true));
 
-        _replaceInstance(id, oldInstance, newInstance);
+        if (automataDcapAttestation == address(0)) {
+            return type(uint256).max;
+        }
```
Brechtpd, line 95: Why doesn't this simply revert?

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -77,37 +81,29 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
+
+        // Still possible to be backward compatible and register instances by the owner.
```
Brechtpd, line 97: Comment seems irrelevant at this place?

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -77,37 +81,29 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
+        (bool verified,) = IAttestation(automataDcapAttestation).verifyParsedQuote(attestation);
+
+        if (!verified) revert SGX_INVALID_ATTESTATION();
+
+        address[] memory _address = new address[](1);
+        _address[0] =
+            address(bytes20(Bytes.slice(attestation.localEnclaveReport.reportData, 0, 20)));
```
Brechtpd, line 104:
```suggestion
        _address[0] = address(bytes20(attestation.localEnclaveReport.reportData));
```
New tricks!

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
adaki2004: Reason is: we might not support (for whatever reason) the on-chain attestastion, so it can be address(0) - and it will be backward compatible with previous versions.

packages/protocol/script/config_dcap_sgx_verifier.sh
adaki2004: @smtmfft

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
Brechtpd: This code still needs updating to make it backward compatible or no? Could be that I'm missing something.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
adaki2004: I think (or thought) this is what we agreed:
kép

adaki2004: so no 'addInstances' by trusted instances.. they can/should add themselves.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
adaki2004: We allow not having set a "dcap_attesation" in the address manager. If we add it, we enable the feature. If we dont add it, feature is not enabled.
Backward compatible with tests (too) :)

Brechtpd: So instead of reverting it returns an invalid ID instead? Reverting seems nicer so it's more clear it's disabled I think, but fair enough.

packages/protocol/script/SetDcapParams.s.sol
smtmfft: it calls 5 functions in run() for register platform, set MrEnclave, and set sgx instance id. for test purpose it's ok, for real use case, maybe we should call individual function instead.

packages/protocol/script/config_dcap_sgx_verifier.sh
smtmfft: Not that familiar with foundry, so it follows another script, will double check.

packages/protocol/script/SetDcapParams.s.sol
smtmfft: Yes

packages/protocol/contracts/thirdparty/automata-attestation/utils/BytesUtils.sol
```diff
@@ -0,0 +1,400 @@
...
+            mask = type(uint256).max; // Set to maximum value of uint256
+        } else {
+            mask = 256 ** (32 - len) - 1;
+        }
+
+        assembly {
+            let srcpart := and(mload(src), not(mask))
+            let destpart := and(mload(dest), mask)
+            mstore(dest, or(destpart, srcpart))
+        }
```
Brechtpd, line 293:
```suggestion
        assembly {
            mcopy(dest, src, len)
        }
```

packages/protocol/contracts/thirdparty/automata-attestation/utils/BytesUtils.sol
```diff
@@ -0,0 +1,400 @@
...
+    function compareBytes(bytes memory a, bytes memory b) internal pure returns (bool) {
+        if (a.length != b.length) {
+            return false;
+        }
+        for (uint256 i = 0; i < a.length; i++) {
+            if (a[i] != b[i]) {
+                return false;
+            }
+        }
+        return true;
```
Brechtpd, line 398:
```suggestion
        return keccak256(a) == keccak256(b);
```

packages/protocol/test/L1/SgxVerifier.t.sol
smtmfft: it can be, it because Dani & I write different scripts independently.

Brechtpd (commented):
Just some easy optimizations in BytesUtils, saves ~150k gas.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -77,37 +81,29 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
-    function addInstances(
-        uint256 id,
-        address newInstance,
-        address[] calldata extraInstances,
-        bytes calldata signature
-    )
+    /// @notice Adds an SGX instance after the attestation is verified
+    /// @param attestation The parsed attestation quote.
+    /// @return id The respective instanceId
+    function registerInstance(V3Struct.ParsedV3QuoteStruct calldata attestation)
```
Brechtpd, line 87: Security feature idea: use the already stored `addedAt` so that SGX instances that are registered with onchain RA verification are only usable after some short time delay. This we can monitor for issues in this smart contract and reject instances that actually don't have a valid RA.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
adaki2004: As discussed - leave it as is (for now) and when SGX is ready to support registration service, we can add it back.

packages/protocol/test/automata-attestation/common/AttestationTestBase.t.sol
```diff
@@ -0,0 +1,75 @@
...
+        string memory fmspc = "00606a000000";
+        (bool tcbParsedSuccess, TCBInfoStruct.TCBInfo memory parsedTcbInfo) =
+            parseTcbInfoJson(tcbInfoJson);
+        require(tcbParsedSuccess, "tcb parsed failed");
+        attestation.configureTcbInfoJson(fmspc, parsedTcbInfo);
+
+        (bool qeIdParsedSuccess, EnclaveIdStruct.EnclaveId memory parsedEnclaveId) =
+            parseEnclaveIdentityJson(enclaveIdJson);
+        require(qeIdParsedSuccess, "qeid parsed failed");
+        attestation.configureQeIdentityJson(parsedEnclaveId);
```
Brechtpd, line 71: Nice. Can this code also be shared with the code in `SetDcapParams`? I know it's much less that can be shared, but still having the same setup code for the actual smart contract calls between the real deployment and tests seems very useful.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -32,10 +34,14 @@ contract SgxVerifier is EssentialContract, IVerifier {
     /// bootstrapping the network with trustworthy instances.
     struct Instance {
         address addr;
-        uint64 addedAt; // We can calculate if expired
+        uint64 addedAt; // We can calculate if expired or not yet valid
+        bool checkValidityCoolTime; // Owner added sgx instances not needed to have a cooldown for
+            // validity
```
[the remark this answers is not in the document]
smtmfft: In raiko, should we make this expire-register rotation happen automatically? 
Pro is all auto, raiko can read the INSTANCE_EXPIRY setting and rotate before that.
Con is prover needs to config their private key into raiko to call register, and we need to handle the error cases which make raiko more complex.
Maybe keep raiko simpler is better.

BTW: now the register is done manually, through that forge script config_dcap_xx.s.sol.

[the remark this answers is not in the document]
adaki2004: > (I was also thinking first to have a `validUntil` instead of doing `validFrom + INSTANCE_EXPIRY`, but doing the `validFrom + INSTANCE_EXPIRY` seems more flexible so we can update `INSTANCE_EXPIRY` at any time by updating the smart contract and then all instances automatically have the updated lifetime which seems more convenient to me)

Good one! Added.

packages/protocol/test/automata-attestation/common/AttestationTestBase.t.sol
adaki2004: Added a common base. 
Please @smtmfft can you have a look too ? :)

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -32,10 +34,12 @@ contract SgxVerifier is EssentialContract, IVerifier {
     /// bootstrapping the network with trustworthy instances.
     struct Instance {
         address addr;
-        uint64 addedAt; // We can calculate if expired
+        uint64 validFrom;
```
dantaik, line 37: validSince?

packages/protocol/script/config_dcap_sgx_verifier.sh
smtmfft: Generally this 2 keys can be different to make each function (could) have its own sender rather than forge test trigger. Here for testing they are the same. So, no problem here.

smtmfft (approved):

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -161,13 +176,25 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
+        }
+
         for (uint256 i; i < _instances.length; ++i) {
             if (_instances[i] == address(0)) revert SGX_INVALID_INSTANCE();
 
-            instances[nextInstanceId] = Instance(_instances[i], uint64(block.timestamp));
+            instances[nextInstanceId] = Instance(_instances[i], validSince);
             ids[i] = nextInstanceId;
 
             emit InstanceAdded(nextInstanceId, _instances[i], address(0), block.timestamp);
```
Brechtpd, line 200: Should the event also use `validSince` instead of `block.timestamp`. Seems more useful.

packages/protocol/contracts/L1/verifiers/SgxVerifier.sol
```diff
@@ -177,13 +204,16 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
         instances[id] = Instance(newInstance, uint64(block.timestamp));
         emit InstanceAdded(id, newInstance, oldInstance, block.timestamp);
     }
 
     function _isInstanceValid(uint256 id, address instance) private view returns (bool) {
         if (instance == address(0)) return false;
         if (instance != instances[id].addr) return false;
-        return instances[id].addedAt + INSTANCE_EXPIRY > block.timestamp;
+        if (instances[id].validSince > block.timestamp) return false;
+        return instances[id].validSince + INSTANCE_EXPIRY > block.timestamp;
```
Brechtpd, line 217:
```suggestion
        return instances[id].validSince <= block.timestamp && block.timestamp <= instances[id].validSince + INSTANCE_EXPIRY;
```

This is really just a suggestion, and I know it's exactly the same, but my brain has trouble with these checks if they are not in the order that make sense to me. And the current code is especially tricky because the first part of the check needs to result to `true` when it's invalid, while the second part needs to result in `true` when it is valid!

Brechtpd (approved):

PR #15563, opened 2024-01-25
feat(protocol): allow someone to ack message reception then invoke the message before others to earn the fee

dantaik:
Allow someone to ack the reception of a message then be the preferred transactor to invoke the transaction before any other addresses (including the message's owner). This solves the fee issue:

> A challenge is that the previous message.fee is designed for a one-step process, now with a two-step process, in theory we need two fees, one for each step, but this is not implemented in this PR as it will have data incompatibility issue.

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -176,9 +176,7 @@ contract Bridge is EssentialContract, IBridge {
             messageReceive[failureSignal].timestamp = uint64(block.timestamp);
         }
 
-        Receive memory _receive = messageReceive[failureSignal];
-
-        if (block.timestamp >= getInvocationDelay() + _receive.timestamp) {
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -95,8 +95,9 @@ contract Bridge is EssentialContract, IBridge {
         external
         onlyFromNamed("bridge_watchdog")
     {
+        uint64 _timestamp = toPause ? type(uint64).max : uint64(block.timestamp);
```
Brechtpd, line 98: I guess have to be a bit careful here because the timestamp is used to do math on, and so if that math is done on uint64s it may overflow. I think currently it's fine because the value is upcasted to uint256s when it's used in math, and also the transactions reverting with overflow is okay for these messages.

Brechtpd (approved):

PR #15565, opened 2024-01-25
fix(protocol): fix bug in LibBytesUtils.toBytes32 (TKO-07)

dantaik:
Screenshot 2024-01-25 at 19 15 28


dantaik:
> Should be something like this no?
> 
> ```
>  function toBytes32(bytes memory _bytes) internal pure returns (bytes32) {
>         require(_bytes.length >= 32, "array is too small"););
>         return abi.decode(_bytes, (bytes32)); // will truncate if input length > 32 bytes
>  }
> ```
> 
> because we just have to prevent reading in uninitialized memory. This will make sure the array is large enough to always have data set of the array itself.
> 
> Alternatively, as recommended, we could also do this:
> 
> ```
>  function toBytes32(bytes memory _bytes) internal pure returns (bytes32) {
>         if (_bytes.length == 0) {
>              return 0;
>         } else if (_bytes.length < 32) {
>             bytes32 ret;
>             assembly {
>                 ret := mload(add(_bytes, 32))
>             }
>             return ret;
>         }
> 
>         return abi.decode(_bytes, (bytes32)); // will truncate if input length > 32 bytes
> }
> ```
> 
> This should also work because memory is aligned on 32 bytes (well, I believe so, would have to double check), so we can be sure that the accessed memory is set correctly for all array lengths, except for input length 0 so it's a special case.

I'm not sure about this part:
```
 else if (_bytes.length < 32) {
>             bytes32 ret;
>             assembly {
>                 ret := mload(add(_bytes, 32))
>             }
>             return ret;
>         }
```

What if we simply return 0 for the size is smaller than 32.

dantaik:
this function is only used by LibMerkleTrie, which doesn't have test coverage. It's hard to tell the impact of such a change.

dantaik:
Screenshot 2024-01-25 at 20 01 21


If we look at one of the use cases of toBytes32, it seems the input can be indeed smaller than 32 bytes.

Brechtpd (approved):

Brechtpd:
Oh wait, I misread the text! `bytes` are NOT always padded to 32 bytes.

image

So in the short case, we still have potential issues because the padding bytes are not guaranteed to be reset, though they are probably also not directly manipulable. Still, not sure what the best approach is now.

adaki2004:
> Should be something like this no?
> 
> ```
>  function toBytes32(bytes memory _bytes) internal pure returns (bytes32) {
>         require(_bytes.length >= 32, "array is too small"););
>         return abi.decode(_bytes, (bytes32)); // will truncate if input length > 32 bytes
>  }
I think this one is NOT ok, because what Daniel also copied:

`// Nodes smaller than 31 bytes aren't hashed.`

> Alternatively, as recommended, we could also do this:
> 
> ```
>  function toBytes32(bytes memory _bytes) internal pure returns (bytes32) {
>         if (_bytes.length == 0) {
>              return 0;
>         } else if (_bytes.length < 32) {
>             bytes32 ret;
>             assembly {
>                 ret := mload(add(_bytes, 32))
>             }
>             return ret;
>         }
> 
>         return abi.decode(_bytes, (bytes32)); // will truncate if input length > 32 bytes
> }
> ```

I dont think this one is good either, because if bytes is 1..31 length long, mload still can overflow.

I'd simply do this:
```
function toBytes32(bytes memory _bytes) internal pure returns (bytes32) {
        if (_bytes.length < 32) {
            return bytes32(_bytes); // It will not over-read
        }
        return abi.decode(_bytes, (bytes32)); // will truncate if input length >32 bytes
            // 32 bytes
    }
```

Brechtpd:
Testing this out, I'm actually not seeing the "short" bytes behavior given above by chatGPT (well, not the greatest source on this as maybe expected), no matter the length, it seems aligned to 32 bytes always, with the first 32 bytes being the length.

But `bytes32(_bytes)` does indeed also work correctly, a bit surprising because such a non-obvious thing to cast. It also zeros out any byte values after the specified length, so this seems like a very safe and good solution, even though it is non-trivial why it works. :)

kirk-baird:
> Testing this out, I'm actually not seeing the "short" bytes behavior given above by chatGPT (well, not the greatest source on this as maybe expected), no matter the length, it seems aligned to 32 bytes always, with the first 32 bytes being the length.

Yep I don't believe gpt is correct here. Length will always be 32 byte word for `bytes`. 

The direct casting does sound like the right solution to me. It will truncate longer than 32 `bytes`, it also handles shorter than 32 bytes and specifically the length zero case which isn't handled by the current code. Though I haven't written test cases so would be good to check all of these scenarios.

packages/protocol/contracts/thirdparty/LibBytesUtils.sol
```diff
@@ -112,16 +112,17 @@ library LibBytesUtils {
...
-        if (_bytes.length < 32) {
+        if (_bytes.length == 0) {
+            return 0;
+        } else if (_bytes.length < 32) {
             bytes32 ret;
             assembly {
                 ret := mload(add(_bytes, 32))
             }
-            return ret;
+            return ret >> (8 * (32 - _bytes.length));
```
adaki2004, line 122: This is problematic:

`ret := mload(add(_bytes, 32)`

because it can overread the _bytes if less than 32 bytes.

Proposed solution:
```
function toBytes32(bytes memory _bytes) internal pure returns (bytes32) {
        if (_bytes.length < 32) {
            return bytes32(_bytes); // It will not over-read
        }
        return abi.decode(_bytes, (bytes32)); // will truncate if input length >32 bytes
            // 32 bytes
    }
```

Kirk also confirmed this one.

packages/protocol/test/thirdparty/LibBytesUtils.t.sol
```diff
@@ -0,0 +1,25 @@
...
+
+        uint256 v = uint256(0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20);
+        assertEq(LibBytesUtils.toBytes32(byteArray), bytes32(v));
+
+        byteArray = new bytes(31);
+        for (uint256 i = 0; i < 31; i++) {
+            byteArray[i] = bytes1(uint8(i + 1));
+        }
+
+        v = uint256(0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20) >> 8;
```
dantaik, line 22: I'm not sure if the expected value shall be 
`v = uint256(0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20) >> 8;` or `v = uint256(0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20) >> 8 << 8;`

packages/protocol/test/thirdparty/LibBytesUtils.t.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.20;
+
+import "../TaikoTest.sol";
+import "../../contracts/thirdparty/LibBytesUtils.sol";
+
+contract TestLibBytesUtils is TaikoTest {
+    function testToBytes32() public {
+        bytes memory byteArray = new bytes(32);
```
Brechtpd, line 9: Ideally loops over lengths from [0..32], or at least zero length.

dantaik:
@Brechtpd @adaki2004 please take a look at the tests.

packages/protocol/test/thirdparty/LibBytesUtils.t.sol
```diff
@@ -6,20 +6,61 @@ import "../../contracts/thirdparty/LibBytesUtils.sol";
...
+        //  0x0102030405060708090a0b0c0d0e0f1011121314151617000000000000000000
+        //  0x0102030405060708090a0b0c0d0e0f1011121314151617180000000000000000
+        //  0x0102030405060708090a0b0c0d0e0f1011121314151617181900000000000000
+        //  0x0102030405060708090a0b0c0d0e0f101112131415161718191a000000000000
+        //  0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b0000000000
+        //  0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c00000000
+        //  0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d000000
+        //  0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e0000
+        //  0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f00
+        //  0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20
```
adaki2004, line 53: This test - imo - is incorrect.

For example, you have this variable:
`0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20`

And you want to say, that bytes1 (0x01) is the data you want to interpret (let's say this is the 'nodeId').
`0x01  -> data you wanna convert to with this to bytes32()
02030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20 -> consecutive memory area contains this 'rubbish' data
`

Padding means, (solidity is little endian!) it should look like this:
`0x0000000000000000000000000000000000000000000000000000000000000001` and not the one which is in the test.

Simple (`bytes32(input)`) will do that job as is written here:
https://docs.soliditylang.org/en/v0.8.24/types.html#conversions-between-elementary-types

kép

packages/protocol/contracts/thirdparty/LibBytesUtils.sol
```diff
@@ -112,16 +112,19 @@ library LibBytesUtils {
     }
 
     function toBytes32(bytes memory _bytes) internal pure returns (bytes32) {
-        if (_bytes.length < 32) {
+        if (_bytes.length == 0) {
+            return 0;
+        } else if (_bytes.length < 32) {
```
adaki2004, line 117:
```
if (_bytes.length < 32) {
            // https://docs.soliditylang.org/en/v0.8.24/types.html#conversions-between-elementary-types
            return bytes32(_bytes);
        }
```

packages/protocol/test/thirdparty/LibBytesUtils.t.sol
Brechtpd: I think the test is right! bytesN types are padded on the right and types like uint are padded on the left. This makes some sense if you see the bytesN as just fixed size byte arrays, and so zero bytes are added to the end of the array. The original implementation also makes that clear, because the bytes were simply mloaded and so padding zeros (if not overwritten) are on the right in the `bytes` data.

But yes, `bytes32(input)` does have this exact behavior and so also passes the test! So that is also my preferred solution.

dantaik:
@Brechtpd implementation optimized. Take another look.

adaki2004 (approved):

Brechtpd (approved):

packages/protocol/contracts/thirdparty/LibBytesUtils.sol
```diff
@@ -112,16 +112,7 @@ library LibBytesUtils {
...
-            bytes32 ret;
-            assembly {
-                ret := mload(add(_bytes, 32))
-            }
-            return ret;
-        }
-
-        return abi.decode(_bytes, (bytes32)); // will truncate if input length >
-            // 32 bytes
+        return _bytes.length < 32 ? bytes32(_bytes) : abi.decode(_bytes, (bytes32));
```
Brechtpd, line 115:
```suggestion
        return bytes32(_bytes);
```

It could actually just be this, it works for all array lengths

Brechtpd (changes requested):

davidtaikocha (approved):

Brechtpd (approved):

PR #15567, opened 2024-01-25
fix(protocol): use safeTransferFrom (TKO-09)

dantaik:
Screenshot 2024-01-25 at 19 27 12

### DIFF
--- a/packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
+++ b/packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
@@ -15,14 +15,11 @@
 pragma solidity 0.8.20;
 
 import "lib/openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
-import "lib/openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
 import "./MerkleClaimable.sol";
 
 /// @title ERC20Airdrop
 /// Contract for managing Taiko token airdrop for eligible users
 contract ERC20Airdrop is MerkleClaimable {
-    using SafeERC20 for IERC20;
-
     address public token;
     address public vault;
     uint256[48] private __gap;
@@ -46,6 +43,6 @@ contract ERC20Airdrop is MerkleClaimable {
 
     function _claimWithData(bytes calldata data) internal override {
         (address user, uint256 amount) = abi.decode(data, (address, uint256));
-        IERC20(token).safeTransferFrom(vault, user, amount);
+        IERC20(token).transferFrom(vault, user, amount);
     }
 }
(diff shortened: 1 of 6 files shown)


dantaik:
Screenshot 2024-01-25 at 20 17 42

@adaki2004 not sure for USDC adaptor, whether we should use safeTransferFrom as well. Do you remember where we get this IUSDC interface?

packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
```diff
@@ -43,6 +46,6 @@ contract ERC20Airdrop is MerkleClaimable {
 
     function _claimWithData(bytes calldata data) internal override {
         (address user, uint256 amount) = abi.decode(data, (address, uint256));
-        IERC20Upgradeable(token).transferFrom(vault, user, amount);
```
Brechtpd, line 46: No need to use it here because it's only for TKO (and TKO follows the ERC20 standard correctly)?

packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
```diff
@@ -71,7 +73,7 @@ contract ERC20Airdrop2 is MerkleClaimable {
     function withdraw(address user) external ongoingWithdrawals {
         (, uint256 amount) = getBalance(user);
         withdrawnAmount[user] += amount;
-        IERC20Upgradeable(token).transferFrom(vault, user, amount);
```
Brechtpd, line 74: Same here, not necessary?

adaki2004:
> Screenshot 2024-01-25 at 20 17 42
> 
> @adaki2004 not sure for USDC adaptor, whether we should use safeTransferFrom as well. Do you remember where we get this IUSDC interface?

No need action here either, because i think this is an invalid comment from SigmaPrime (related to USDC). They follow the standard of IERC20 and returns a bool.
I think they mixed it up with USDT, which indeed does not return true. Will ask them in the chat.

adaki2004:
> > Screenshot 2024-01-25 at 20 17 42
> > @adaki2004 not sure for USDC adaptor, whether we should use safeTransferFrom as well. Do you remember where we get this IUSDC interface?
> 
> No need action here either, because i think this is an invalid comment from SigmaPrime (related to USDC). They follow the standard of IERC20 and returns a bool. kép
> 
> I think they mixed it up with USDT, which indeed does not return true. Will ask them in the chat.

@dantaik Kirk confirmed (see telegram), he mixed it with USDT, all good, can stay as is, as USDC follows the standard.

adaki2004 (approved):

Brechtpd (approved):

PR #15568, opened 2024-01-25
fix(protocol): new way to calculate meta.difficulty (TKO-11)

dantaik:
Screenshot 2024-01-25 at 19 38 49

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -111,7 +111,7 @@ library LibProposing {
         unchecked {
             meta = TaikoData.BlockMetadata({
                 l1Hash: blockhash(block.number - 1),
-                difficulty: keccak256(bytes.concat(bytes32(block.prevrandao))),
+                difficulty: 0, // to be initialized below
                 blobHash: 0, // to be initialized below
                 extraData: params.extraData,
                 depositsHash: keccak256(abi.encode(depositsProcessed)),
@@ -187,6 +187,13 @@ library LibProposing {
             meta.txListByteOffset = 0;
             meta.txListByteSize = uint24(txList.length);
         }
+
+        // Following the Merge, the L1 mixHash incorporates the
+        // prevrandao value from the beacon chain. Given the possibility
+        // of multiple Taiko blocks being proposed within a single
+        // Ethereum block, we must introduce a salt to this random
+        // number as the L2 mixHash.
+        meta.difficulty = keccak256(abi.encodePacked(block.prevrandao, b.numBlocks, block.number));
 
         // Use the difficulty as a random number
         meta.minTier = ITierProvider(resolver.resolve("tier_provider", false)).getMinTier(
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -193,9 +193,7 @@ library LibProposing {
         // of multiple Taiko blocks being proposed within a single
         // Ethereum block, we must introduce a salt to this random
         // number as the L2 mixHash.
-        unchecked {
-            meta.difficulty = meta.blobHash ^ bytes32(block.prevrandao * b.numBlocks * block.number);
-        }
+        meta.difficulty = keccak256(abi.encodePacked(block.prevrandao, b.numBlocks, block.number));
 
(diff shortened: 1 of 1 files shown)


adaki2004 (approved):

Brechtpd:
What about just using `block.prevrandao`? Any benefits of artificially modifying the randomness per L2 if multiple ones are submitted in same L1 block or the randomness not matching the L1 value directly? 

If we want to make it unique per L2 than hashing in `address(this)` would ensure this. But I don't immediately see a reason to make it more difficult (no pun intended) than just passing through `block.prevrandao` directly but I could be missing something.

dantaik:
> prevrandao

I like the idea.

dantaik:
> > prevrandao
> 
> I like the idea.

how about we simply hash it `keccak256(bytes.concat(bytes32(block.prevrandao)))`

dantaik:
Wait, I think we want to make sure each L2 block has a different difficulty, otherwise L2 blocks proposed in the same L1 block always have the same min-tier...

Brechtpd:
I think realistically speaking, proposers will just avoid submitting blocks when it's not the cheapest tier. Because if a certain proof is cheap enough to not really make a difference on their profitability, that proof should just always be in the base tier because that's better security. And if the proof is very expensive, well it simply will not be profitable to submit a block with the fees users are used to paying for most of the other blocks. If a zk proof costs $20, well that's a pretty large spike in expected fees users have to pay that is not predictable. This compared to the pretty good predictability of L1 fees and L2 fees (at least for a short time) because of EIP-1559.

So I think randomness in the proofs will result in a worse user experience because either users have to overpay all the time (bad ux because higher fees), or the proposers will simply wait one (or more) L1 blocks so the randomness is more favorable and then be able to submit their block when it is actually profitable (bad ux because now transactions take longer to finalize).

adaki2004:
> I think realistically speaking, proposers will just avoid submitting blocks when it's not the cheapest tier. Because if a certain proof is cheap enough to not really make a difference on their profitability, that proof should just always be in the base tier because that's better security. And if the proof is very expensive, well it simply will not be profitable to submit a block with the fees users are used to paying for most of the other blocks. If a zk proof costs $20, well that's a pretty large spike in expected fees users have to pay that is not predictable. This compared to the pretty good predictability of L1 fees and L2 fees (at least for a short time) because of EIP-1559.
> 
> So I think randomness in the proofs will result in a worse user experience because either users have to overpay all the time (bad ux because higher fees), or the proposers will simply wait one (or more) L1 blocks so the randomness is more favorable and then be able to submit their block when it is actually profitable (bad ux because now transactions take longer to finalize).

That would imply 100% of the blocks, shall have 100% of the proofs (sgx, zk, guardian) immediately there together - otherwise makes no sense anyways, no ? Because simply noone would want to propose something which costs more than "needed" ? 🤔

Brechtpd:
I think it enforces at least some fixed configuration without randomness. For example SGX + guardian as the base proofs (because both SGX and guardian are very cheap to verify, so it doesn't make much sense not to always have them, it won't impact user fees in any significant way). Then it could still have SGX + guardian + ZK as a way to override the SGX + guardian tier, because ZK at the base level may be expensive (remains to be seen though), but that would only be done when needed if the blockhash is actually wrong, not randomly, so users never have to pay the ZK cost normally (but we do lower security and increase the time to finality by not having it part of the base proofs).

dantaik:
In the future we can try some other ideas, for example, determine the minTier of block by the next proposeBlock transaction, or use VRF. But for now, with minimal modification, we have the following option:

1. ` meta.difficulty = block.prevrandao` - most simply but difficulty is shared by multiple L2 blocks proposed in the same L1 block.
2. `meta.difficulty = keccak256(abi.encodePacked(block.prevrandao, b.numBlocks, block.number));` - the proposed change where each L2 block has its own pseudo number, proposers cannot influence the value.
3. `meta.difficulty = meta.blobHash ^ bytes32(block.prevrandao * b.numBlocks * block.number);` - the current A6 impl, an issue with it is that block proposers can carefully construct blob to proactively influence the difficulty value.

Among the above 3 options. I still think #2 is preferred over the other 2.

Brechtpd (approved):

Brechtpd:
For the difficulty value itself I think it is indeed fine, especially because contracts shouldn't depend on it much as well. Random minimum tiers I think is problematic but I guess that's another discussion.

PR #15573, opened 2024-01-25
feat(protocol): cap reward at new proof's validity bond (TKO-13, 23)

dantaik:
Screenshot 2024-01-25 at 21 10 35
Screenshot 2024-01-25 at 21 58 13

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProvingAlt.sol
+++ b/packages/protocol/contracts/L1/libs/LibProvingAlt.sol
@@ -352,8 +352,11 @@ library LibProvingAlt {
         }
 
         unchecked {
-            if (reward > tier.validityBond) {
-                tko.transfer(msg.sender, reward - tier.validityBond);
+            if (reward >= tier.validityBond) {
+                // If the reward is bigger than the validity bond to be deposited,
+                // we simply cap the reward at tier.validityBond, otherwise, a prover
+                // will be incentivized to submit a invalid higher-tier (verifiable)
+                // proof to contest a valid lower tier proof to win extra reward.
             } else {
                 tko.transferFrom(msg.sender, address(this), tier.validityBond - reward);
             }
--- a/packages/protocol/contracts/L1/libs/LibProvingAlt.sol
+++ b/packages/protocol/contracts/L1/libs/LibProvingAlt.sol
@@ -357,6 +357,9 @@ library LibProvingAlt {
                 // we simply cap the reward at tier.validityBond, otherwise, a prover
                 // will be incentivized to submit a invalid higher-tier (verifiable)
                 // proof to contest a valid lower tier proof to win extra reward.
+
+                // Note that this behavior is different from LibProving's implementation
+                // where the cap is not applied.
             } else {
                 tko.transferFrom(msg.sender, address(this), tier.validityBond - reward);
             }
--- a/packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
+++ b/packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
@@ -37,8 +37,8 @@ contract TaikoA6TierProvider is EssentialContract, ITierProvider {
         if (tierId == LibTiers.TIER_OPTIMISTIC) {
             return ITierProvider.Tier({
                 verifierName: "tier_optimistic",
-                validityBond: 1000 ether, // TKO
-                contestBond: 1000 ether, // TKO
+                validityBond: 250 ether, // TKO
+                contestBond: 250 ether, // TKO
                 cooldownWindow: 24 hours,
                 provingWindow: 2 hours,
                 maxBlocksToVerify: 10
@@ -59,8 +59,8 @@ contract TaikoA6TierProvider is EssentialContract, ITierProvider {
         if (tierId == LibTiers.TIER_SGX_AND_PSE_ZKEVM) {
             return ITierProvider.Tier({
                 verifierName: "tier_sgx_and_pse_zkevm",
-                validityBond: 250 ether, // TKO
-                contestBond: 250 ether, // TKO
+                validityBond: 1000 ether, // TKO
+                contestBond: 1000 ether, // TKO
                 cooldownWindow: 24 hours,
                 provingWindow: 6 hours,
                 maxBlocksToVerify: 6


dantaik:
I'm not sure this is the correct handling. On one hand, optimistic proofs should have a higher validity bond, but if we adopt this "feature", then higher tier proofs should have larger validity bonds.

@Brechtpd @adaki2004 let me know what you think.

Brechtpd:
The system really is complex to think about and figure out all the ways it could go wrong or be exploited. Testing it will also be very challenging.

I'm still of the opinion that this code is too complex for something that will never be used because it's only activated when somebody thinks they can exploit the system (otherwise they know they will lose money, so they won't put anything invalid onchain). Ironically, the more complex we make this system, the more likely it will be used because the chance there will be bugs in this code increases, and so attacking the system is something that people will want to take advantage of.

So I would call for something that is as easy as possible, even if for some reason it is less efficient or has less features, because all of that is just theoretical. I don't know what that is currently, but less is more. Especially in this case because again, if this constesting code doesn't contain bugs/exploits, it will never be used. And that's a very good thing.

adaki2004:
If ZK will be cheap, i just have them together with SGX and Guardian - all 3 together.

The likelihood of an exploiter to ever be capable of exploiting 2 different system (sgx + zk), resulting in a consensus of a false transition is almost non-existent. 2, different systems needs to have a consensus in a false (but same) new blockhash, highly unlikely.

And in case the 3 (ZK, SGX, Guardian) disagree, it can be a 2/3 multisig in a sense. In this way also the guardian cannot be a centralization factor anymore - but a supportive leg !

If ZK is not cheap to have them all 100% of the blocks ? Good question, i leave it for you to make the call :).

Brechtpd:
SGX + guardian is I think already very high security, and the good news is that's is extremely cheap as well! ZK indeed we'll have to see, but that shouldn't take long to get cheap enough as well I feel (as an optimist).

dantaik:
> The system really is complex to think about and figure out all the ways it could go wrong or be exploited. Testing it will also be very challenging.
> 
> I'm still of the opinion that this code is too complex for something that will never be used because it's only activated when somebody thinks they can exploit the system (otherwise they know they will lose money, so they won't put anything invalid onchain). Ironically, the more complex we make this system, the more likely it will be used because the chance there will be bugs in this code increases, and so attacking the system is something that people will want to take advantage of.
> 
> So I would call for something that is as easy as possible, even if for some reason it is less efficient or has less features, because all of that is just theoretical. I don't know what that is currently, but less is more. Especially in this case because again, if this constesting code doesn't contain bugs/exploits, it will never be used. And that's a very good thing.

Identifying bugs in SGX and ZK code base is more challenging than finding bugs in BCR! The very reason BCR exists is that proofs may be verifiable but wrong. With BCR, we have the luxury to correct wrong state-transition without overloading ourself with over-proving each and every block. If guardian provers are engaged in each and every blocks, we'll never be able to live without them. I don't like guardian provers, when ZK is ready (well tested) ,we should drop guardians as soon as possible. Then the question from dapp developers still hold: ZK is now perfect (bug-free) but they may still be expensive, how can we further lower the cost of layer two transactions? BCR offers a solution for this question, without guardian provers in the picture.

Merging different types of proofs (SGX, ZK, and guardians) into one doesn't seem to be an elegant solution to me.

Brechtpd:
> Identifying bugs in SGX and ZK code base is more challenging than finding bugs in BCR! The very reason BCR exists is that proofs may be verifiable but wrong. 

I don't think contestation will help find or identify bugs in the proving systems, see below why. 

But even if it would, it wouldn't even really identify bugs though. We would just know it could be exploited, but we have no idea what the bug/exploit actually is. So we can see something is wrong in a proof system, which is valuable information so we could disable the proof system or something like that, but we can't really use it directly to make the exploited proof better. We still have to set up tests and review/audit the code of the proving systems like normal.

>  With BCR, we have the luxury to correct wrong state-transition without overloading ourself with over-proving each and every block.

Comes at a pretty high price though (more security assumptions and much higher bridging latency). 

>  If guardian provers are engaged in each and every blocks, we'll never be able to live without them. I don't like guardian provers, when ZK is ready (well tested) ,we should drop guardians as soon as possible.

If the argument is that guardian provers (or some POA selected validators in the future perhaps) are so powerful that prover bugs will never be exploited, than I think that's a very nice feature and shows how powerful they are! 

Following this reasoning though we can also say that prover bugs will never be exploited as long as there is a multi-sig or security counsel active that has the power to revert invalid state transitions. The only logical thing for hackers to do is to only exploit bugs they know when they are sure they will be able to steal money. And this will never be the case as long as there are any security precautions active.

So I also don't see what the difference here is between the guardian provers proving each block or not. As long as they are there, why would an attacker put an invalid blockhash onchain if they know it will be reverted? They would only lose money by doing so AND make it public that they can exploit the proving system (extremely very valuable information). So the only rational thing for hackers to do is to only exploit the system when they can both exploit the proofs themselves AND be sure they can DDOS/exploit the guardian prover keys/....

Of course, in theory I agree. In practice, things seem much more complicated and having an extra set of validators as one possible proving system seems very valuable. 

>  Then the question from dapp developers still hold: ZK is now perfect (bug-free) but they may still be expensive, how can we further lower the cost of layer two transactions? BCR offers a solution for this question, without guardian provers in the picture.

Agreed that the mechanism itself can be valuable! Just that there are different ways to implement it, and I think it's worthwhile to think how we can keep the core mechanism while simplifying things as much as possible. It's like the proving fee, it took a couple of iterations and eventually we found something very simple that works.

> Merging different types of proofs (SGX, ZK, and guardians) into one doesn't seem to be an elegant solution to me.

Personally I believe it's the most powerful and elegant solution because, except for cost, it is has no downsides and it's the highest level of security and has the fastest finality. I think it's the only way to truly make a truly scalable chain, onchain finalization delays should be instantaneous, everything else is not scalable in the long term. Assuming that indeed a prover with zero bugs is a utopia, multiple proof together is the only way to get there.

> I would argue that BCR is valuable as long as finding bugs in specific proofing system is harder than finding bugs in BCR itself.

Yeah but knowing there is a bug != knowing how to fix it. And as explained above, if the contestation actually works, the proving systems could still be extremely buggy and we would have no idea because nobody would publicize they can exploit a prover and not get their price (== stolen L2 funds) (or if we would know, but then the contestation would be buggy as well, and all the L2 funds would be lost).

> I thought SGX is simple and straightforward, but looking at the new SGX w/ RA, I started to doubt it.

Nothing is easy, but there are different kind of bugs. Bugs where you forget something obvious or made a dumb mistake, those are good bugs. Then there are non-obvious bugs that are there because of complexity and difficult scenario's.

adaki2004:
> With BCR, we have the luxury to correct wrong state-transition without overloading ourself with over-proving each and every block. If guardian provers are engaged in each and every blocks, we'll never be able to live without them.

It comes at a cost of 'contestation + cooldown time' which is a prolonged period compared to instant finalization. We (or maybe just me) was making claims that fraud proof systems take very long time to exit/withdraw/bridge and now we are almost at the same UX - while not gaining anything in return.

The 'plus' here (not cost because BCR meant to be cost effective) is the extra security - that actually we have a proving system (sgx or guardian, or both) in case contested.

> Merging different types of proofs (SGX, ZK, and guardians) into one doesn't seem to be an elegant solution to me.

To me, it seems an obvious solution. I know your take on the guardians, and also on ZK. I share the same vision, if ZK is expensive, we shall not have it by default for every blocks, otherwise it also a horrible UX (cost).

But with SGX + guardian (together) we will not give up anything from the security, it comes with a tiny extra cost (compared to fraud or sgx only) while we gain instant finality.

I totally get the reasoning, why you want to have guardian - but not always. -> It looks not so nice. But from security wise you better have it - even if just "sometimes" (or never) submitting proof. 

Question is: what do we want to sacrifize ? 
- Better user experience (instant finalization) -> proofs together
- Small improvement in cost (bc. guardian is not always there) -> proofs separate, contestable

Actually, having those proofs together would make us look even more decentralized, hence the guardian (as a single entity per say) cannot rule the chain alone, they need to come to a conclusion together with the SGX.

When can we kill guardian ? When ZK is affordable and bugfree for a long-time, and we also tested it in prod.

Why we cannot expect hackers to identify and exploit ZK bugs in our system ?
In both of the case (contestable, or just 1 tier proof aggregated), the hacker would be simply an idiot (or a white hat) doing so. Why ? Because spending hours in front of the PC, while trying to exploit a bug which will be then overridden by the guardian anyways (proofs together or not)..

So while i like the multi-tier possibility of the design, and would keep it in (for app-chains, L3s, other L2s, etc.), but just the possibility of it for now, and configure mainnet a 1-tier solution.

Maybe gathering all the pros, cons, side effects and have a vote on it, could also help decide.

Personally, i'd prefer paying 0.01x more, to have my transaction and bridge finalized instantly.

d1onys1us:
I think that the most elegant solution is all proofs at once with no contestation mechanism, and we were criticized for this on Twitter as well just recently. That is because it has the upsides already mentioned:
- Highest security
- Instant finalization
- More deleted code and simplified logic

I also particularly agree that contestation is a mechanism that people will not use except for exploitation, resulting in dead code and a wider attack surface. I just can't imagine someone running a contester regularly as if they think it's something they can profit from, therefore the only reason again being to exploit things.

The downside of this is clear as well: slightly more expensive in the short term, but I feel it makes sense to make this deliberate tradeoff. 

Also I don't know the scope of such refactoring, but I really agree that contestation is an attack surface and will be unused for security purposes, but instead for attack purposes. I feel that if users want super cheap transactions at the cost of less security they should consider using a validium pre-danksharding instead.

About the guardian provers: I agree that they are not great to rely on completely, and there can be some improvements there.  I'm not sure what is best to ensure their liveness, but as we saw yesterday we had a lot of people go down. I still think they are useful if we can improve their liveness incentives (and do things like mutex-locking between each other for upgrades), and also distribute them a bit more.

PR #15574, opened 2024-01-25
feat(protocol): check 4844 staticcall return values (TKO-22)

dantaik:
@davidtaikocha You many need to verify this PR is indeed a valid enhancement.  Comment the additional checks out if necessary when you try use 4844.

Screenshot 2024-01-25 at 21 33 06


davidtaikocha:
> @davidtaikocha You many need to verify this PR is indeed a valid enhancement. Comment the additional checks out if necessary when you try use 4844.
> 
> Screenshot 2024-01-25 at 21 33 06

Will do the verification soon

Brechtpd (approved):

davidtaikocha (approved):

PR #15577, opened 2024-01-25
feat(protocol): allow bridge to ban addresses

dantaik:


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -207,7 +216,7 @@ contract Bridge is EssentialContract, IBridge {
         // Process message differently based on the target address
         if (
             message.to == address(0) || message.to == address(this)
-                || message.to == address(signalService)
+                || message.to == address(signalService) || addressBanned[message.to]
```
Brechtpd, line 219: Cleaner to just add the other address in `addressBanned` as well I think.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -75,6 +78,12 @@ contract Bridge is EssentialContract, IBridge {
         _ctx.msgHash == bytes32(PLACEHOLDER);
     }
 
+    function banAddress(address addr, bool toBan) external onlyOwner nonReentrant whenNotPaused {
```
Brechtpd, line 81: `whenNotPaused` needed? Seems reasonable to also be able to update these when the bridge is paused if for some reason this is related to some exploit.

dantaik: It can be removed.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: I still think we should leave other checks here as-is to avoid operational mistakes, for example, forgetting to ban the signalService. 

Banning other addresses are not so critical but useful.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: Removed

adaki2004 (approved):

packages/protocol/contracts/bridge/Bridge.sol
Brechtpd: Ah yes good point that it could easily be forgotten if the signal service address is updated or is not yet known when `init` is called.

Brechtpd (approved):

PR #15582, opened 2024-01-26
fix(protocol): Correct decoding (TKO-03)

adaki2004:
image


Brechtpd (approved):
General question: do we need recall? It seems like a complex thing, also complex to test, and a source of multiple bugs presumably also because it's hard to test and not being used.

So do we think it is just a nice to have? Do we think it's something critical that will actually be used? I'm just trying to find out if this feature makes sense from a risk/reward point of view because the bridge itself is so critical for security, the reward really needs to be worth the risk.

dantaik (approved):

PR #15587, opened 2024-01-27
chore(protocol): change the A6 config so higher tier has larger bonds

dantaik:


davidtaikocha (approved):

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
```diff
@@ -37,8 +37,8 @@ contract TaikoA6TierProvider is EssentialContract, ITierProvider {
         if (tierId == LibTiers.TIER_OPTIMISTIC) {
             return ITierProvider.Tier({
                 verifierName: "tier_optimistic",
-                validityBond: 1000 ether, // TKO
-                contestBond: 1000 ether, // TKO
+                validityBond: 250 ether, // TKO
+                contestBond: 500 ether, // TKO
```
dantaik, line 41: contestBond is higher now because if the original proof is invalid, then we don't want the proof to be contested; if the original proof is indeed valid, the contester wins 125 TKO, this means a 25% return for the 500 contestBond.

packages/protocol/contracts/L1/tiers/TaikoA6TierProvider.sol
```diff
@@ -59,8 +59,8 @@ contract TaikoA6TierProvider is EssentialContract, ITierProvider {
         if (tierId == LibTiers.TIER_SGX_AND_PSE_ZKEVM) {
             return ITierProvider.Tier({
                 verifierName: "tier_sgx_and_pse_zkevm",
-                validityBond: 250 ether, // TKO
-                contestBond: 250 ether, // TKO
+                validityBond: 1000 ether, // TKO
+                contestBond: 2000 ether, // TKO
```
dantaik, line 63: The higher the tier is ,the higher the bonds are supposed to be.

davidtaikocha (approved):

adaki2004 (approved):

PR #15600, opened 2024-01-29
fix(protocol): address miscellaneous feedbacks from Sigma Prime (TKO26)

adaki2004:
kép

Please see this modification + some of my concerns/comments in the notion file at raw TKO_26!

https://github.com/taikoxyz/taiko-mono/pull/15600

Some notes to subpoints:
Nr. 6 Unnecessary functions in SignalService and AddressManager. → It would create a diff design scheme we followed, and i understand it is deployment overhead, but would involve to change OwnerUUPS and others inheriting from it OR NOT inheriting from that contract but a similar, without having the pasue(), unPause().

Nr. 13: Cannot find this part:
” [85-87] of LibVerifying.sol contain inequalities where we have for an unsigned inte-
ger x which will never return true. “

Nr. 15.:  Total Supply Of TKO Token Difficult To Determine
@Daniel Wang 
As burning is not an actual burn (in our contract and complex code, i dont exactly know what is the best was to decouple)

Nr. 19: Related Asset(s): L1/hooks/AssignmentHook.sol

It can if tip is, high, but tip is set by the proposer, why would anyone want to give such (unrelaistic amount), they would just shoot in their legs, because it even has to send it with msg.value.

Nr. 21:
Provers May Contest Their Own Proof

I do not see anyone benefitting from this. Contest bond shall be big enough.

Also if some may submitted (accidentally ?) a wrong proof, he (the original prover) can correct it’s mistake by contesting and re-proving with a higher tier, so in such case it does not lose all his bonds.


packages/protocol/contracts/tokenvault/ERC1155Vault.sol
```diff
@@ -144,7 +144,9 @@ contract ERC1155Vault is BaseNFTVault, ERC1155ReceiverUpgradeable {
             }
         }
 
-        _to.sendEther(msg.value);
+        if (msg.value > 0) {
```
dantaik, line 147: I think we should still allow sending 0 ETH to message.owner, in case this addresss "receive" function is not empty.

Same for other changes.

dantaik (changes requested):

packages/protocol/contracts/L1/gov/TaikoTimelockController.sol
```diff
@@ -19,8 +19,6 @@ import
 import "../../common/OwnerUUPSUpgradable.sol";
 
 contract TaikoTimelockController is OwnerUUPSUpgradable, TimelockControllerUpgradeable {
-    uint256[50] private __gap;
```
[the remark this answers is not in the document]
adaki2004: Because `__gap` is usually needed to reserve storage space (and mainly in parent contracts), but here there is no storage variables in this contract (yet) so we can simply add new storage variables any time we upgrade.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -176,7 +175,7 @@ library LibProving {
 
         // The new proof must meet or exceed the minimum tier required by the
         // block or the previous proof; it cannot be on a lower tier.
-        if (proof.tier == 0 || proof.tier < meta.minTier || proof.tier < ts.tier) {
+        if (proof.tier == 0 || proof.tier < ts.tier) {
```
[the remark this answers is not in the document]
adaki2004: Yes.
kép

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -90,11 +90,16 @@ library LibVerifying {
                 || config.blockMaxTxListBytes > 128 * 1024 // calldata up to 128K
                 || config.livenessBond == 0 || config.ethDepositRingBufferSize <= 1
                 || config.ethDepositMinCountPerBlock == 0
+            // Audit recommendation, and gas tested. Processing 32 deposits (as initially set in
+                // TaikoL1.sol) costs 72.502 gas. So a safe bet to set max cap around 100K (cca 45)
+                // as average propose block is somewhere 170-200K depending on the tier). It is
+                // anyway set by Taiko Labs.
+                || config.ethDepositMaxCountPerBlock > 45
```
Brechtpd, line 97: Seems like on Ethereum the limit is 16: 
- https://eips.ethereum.org/EIPS/eip-4895#why-no-gas-costs-for-the-withdrawal-type
- https://benjaminion.xyz/eth2-annotated-spec/phase0/beacon-chain/#max-operations-per-block

Because these deposits are actually pretty expensive as they require L1 work I don't think we have to follow this same value, but 16 would be the very conservative value.

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
Brechtpd: I guess it may be used as some kind of trigger that a bridge transaction was just done (receive function could figure that out by checking `msg.sender`, so maybe useful in some bridge specific contracts?) but otherwise there is no extra context what happened.

Either way fine for me, but if we do add the check I would do the `if (msg.value > 0) {` inside `sendEther`, it looks messy to see it everywhere.

Brechtpd (commented):
- I would move LibUint512Math and TaikoL2Signer to the test folder, they are only used in the tests. Only `GOLDEN_TOUCH_ADDRESS` is actually used and so can just be copied to `TaikoL2.sol`.
- We may want to update the Merkle trie/RLP decoding to the latest versions used in optimism: https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock/src/libraries. I haven't diffed them, but seems likely there are also some optimizations in there. In any case, we should definitely check if the new versions contain fixes or improvements.

packages/protocol/contracts/L1/gov/TaikoTimelockController.sol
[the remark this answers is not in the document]
dantaik: If you remove `__gap` here, the sub-contract's storage layout will be totally shifted.

A contract having no storage doesn't mean it won't have storage variable in the future. We need to keep it.

packages/protocol/contracts/L1/tiers/ITierProvider.sol
```diff
@@ -23,7 +23,7 @@ interface ITierProvider {
         uint96 contestBond;
         uint24 cooldownWindow;
         uint16 provingWindow;
-        uint8 maxBlocksToVerify;
+        uint8 maxBlocksToVerifyWithTier;
```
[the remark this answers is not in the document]
dantaik: I still think `maxBlocksToVerify` is good enough, as 'tier.maxBlocksToVerify' already tells it's 'per proof'

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
dantaik: Sending a zero Ether to an address is a valid use case, I don't think our bridge should disable it. I think we should just make the invocation without checking if msg.value is zero. Calling a default receive function is very cheap with 0 ether.

dantaik (changes requested):

dantaik:
> * I would move LibUint512Math and TaikoL2Signer to the test folder, they are only used in the tests. Only `GOLDEN_TOUCH_ADDRESS` is actually used and so can just be copied to `TaikoL2.sol`.

See this PR: https://github.com/taikoxyz/taiko-mono/pull/15604

> * We may want to update the Merkle trie/RLP decoding to the latest versions used in optimism: https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock/src/libraries. I haven't diffed them, but seems likely there are also some optimizations in there. In any case, we should definitely check if the new versions contain fixes or improvements.

I'll work on this one later.

packages/protocol/contracts/L1/libs/LibVerifying.sol
adaki2004: I make it max 32 then.

adaki2004:
> > * I would move LibUint512Math and TaikoL2Signer to the test folder, they are only used in the tests. Only `GOLDEN_TOUCH_ADDRESS` is actually used and so can just be copied to `TaikoL2.sol`.
> 
> See this PR: #15604


Addressed this here too. And all the changes you were recommending.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -206,6 +205,13 @@ library LibProving {
             // The verifier can be address-zero, signifying that there are no
             // proof checks for the tier. In practice, this only applies to
             // optimistic proofs.
+            if (
+                verifier == address(0)
+                    && keccak256(abi.encodePacked(tier.verifierName))
+                        != keccak256(abi.encodePacked("tier_optimistic"))
```
dantaik, line 211: make keccak256(abi.encodePacked("tier_optimistic")) a constant var

packages/protocol/contracts/L1/libs/LibProvingAlt.sol
```diff
@@ -109,7 +109,7 @@ library LibProvingAlt {
 
         // The new proof must meet or exceed the minimum tier required by the
         // block or the previous proof; it cannot be on a lower tier.
-        if (proof.tier == 0 || proof.tier < meta.minTier || proof.tier < ts.tier) {
+        if (proof.tier == 0 || proof.tier < ts.tier) {
```
dantaik, line 112: add the optimistic name check here.

packages/protocol/contracts/L1/verifiers/PseZkVerifier.sol
```diff
@@ -80,7 +81,7 @@ contract PseZkVerifier is EssentialContract, IVerifier {
                 pointProof: pf.pointProof
             });
         } else {
-            assert(zkProof.pointProof.length == 0);
+            if (zkProof.pointProof.length != 0) revert L1_BLOB_NOT_USED();
```
dantaik, line 84: search for "assert"

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
```diff
@@ -193,7 +193,9 @@ contract ERC1155Vault is BaseNFTVault, ERC1155ReceiverUpgradeable {
             }
         }
         // Send back Ether
-        message.owner.sendEther(message.value);
+        if (message.value > 0) {
```
dantaik, line 196: revert

dantaik (changes requested):

packages/protocol/contracts/L1/libs/LibProving.sol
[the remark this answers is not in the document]
dantaik: In this PR, lets remove LibProving and rename LibProvingAlt to LibProving

packages/protocol/contracts/L1/libs/LibProvingAlt.sol
dantaik: Dani, I reviewed the code more carefully. I think we should simply remove:

```
  if (ts.tier < meta.minTier) {
                revert L1_UNEXPECTED_TRANSITION_TIER();
            }
```

and keep `if (proof.tier == 0 || proof.tier < meta.minTier || proof.tier < ts.tier)`.


The reason we have questions now, is because previously 
```
  if (ts.tier < meta.minTier) {
                revert L1_UNEXPECTED_TRANSITION_TIER();
            }
```
is 

```
  assert (ts.tier >= meta.minTier);
```
this is to really make sure we know assert must be true. That's why we use asset, but once asserts are converted to if-revert, it will cause a confusion that the case may actually happen.


So maybe we should **keep other asserts there without converting them to if-revert.**

adaki2004: Okay, so just keep as is, and remove what you suggests here. But what about assert ? (Shall we put this back:
`assert (ts.tier >= meta.minTier);` ? I guess no)

dantaik (approved):

packages/protocol/contracts/L1/gov/TaikoTimelockController.sol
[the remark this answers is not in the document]
dantaik: Dani, you need to revert this change.

packages/protocol/contracts/L1/gov/TaikoGovernor.sol
```diff
@@ -33,8 +33,6 @@ contract TaikoGovernor is
     GovernorVotesQuorumFractionUpgradeable,
     GovernorTimelockControlUpgradeable
 {
-    uint256[50] private __gap;
```
dantaik: please revert this change

packages/protocol/contracts/L1/gov/TaikoGovernor.sol
```diff
@@ -91,7 +91,7 @@ contract TaikoGovernor is
         return 50_400; // 1 week
     }
 
-    // The number of votes required in order for a voter to become a proposer
+    // The number of votes required in order for a voter to become a vote proposer
```
dantaik, line 94: revert this change

dantaik, line 94:
```suggestion
    // The number of votes required in order for a voter to become a proposer
```

dantaik (approved):

packages/protocol/contracts/L1/libs/LibProvingAlt.sol
```diff
@@ -22,15 +22,16 @@ import "../verifiers/IVerifier.sol";
 import "../TaikoData.sol";
 import "./LibUtils.sol";
 
-/// @title LibProvingAlt
-/// @notice An alternative library for handling block contestation and proving in the Taiko
+/// @title LibProving
```
dantaik, line 25: Keeping the file name as LIbProvingAlt is a good idea. Later we can have a simple PR by using `git mv LibProvingAlt.sol LibProving.sol`

packages/protocol/contracts/L1/libs/LibProvingAlt.sol
```diff
@@ -108,7 +110,7 @@ library LibProvingAlt {
         // become available. In cases where a transition with the specified
         // parentHash does not exist, the transition ID (tid) will be set to 0.
         (uint32 tid, TaikoData.TransitionState storage ts) =
-            _createTransition(state, blk, meta, tran, slot);
```
dantaik, line 111: seems this change was reverted by you.

dantaik (approved):

Brechtpd (approved):

PR #15608, opened 2024-01-30
feat(protocol): adopt optimism new trie codebase

dantaik:
From Brecht:

> We may want to update the Merkle trie/RLP decoding to the latest versions used in optimism: https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock/src/libraries. I haven't diffed them, but seems likely there are also some optimizations in there. In any case, we should definitely check if the new versions contain fixes or improvements.


packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -182,8 +182,11 @@ contract AssignmentHook is EssentialContract, IHook {
         pure
         returns (uint256)
     {
-        for (uint256 i; i < tierFees.length; ++i) {
+        for (uint256 i; i < tierFees.length;) {
             if (tierFees[i].tier == tierId) return tierFees[i].fee;
+            unchecked {
+                ++i;
+            }
```
adaki2004, line 189: No need to have this optimizaton, bc. fro 0.8.22:

kép

Unchecked loop increemnt optimizer:
kép
https://docs.soliditylang.org/en/latest/internals/optimizer.html#unchecked-loop-optimizer

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -260,6 +260,9 @@ library LibProposing {
                 );
 
                 prevHook = params.hookCalls[i].hook;
+                unchecked {
+                    ++i;
+                }
```
adaki2004, line 265: Same as here:
https://github.com/taikoxyz/taiko-mono/pull/15608/files#r1470670145

packages/protocol/contracts/L1/provers/Guardians.sol
```diff
@@ -57,8 +57,11 @@ abstract contract Guardians is EssentialContract {
 
         // Delete current guardians data
         uint256 guardiansLength = guardians.length;
-        for (uint256 i; i < guardiansLength; ++i) {
+        for (uint256 i; i < guardiansLength;) {
             delete guardianIds[guardians[i]];
+            unchecked {
+                ++i;
+            }
```
adaki2004, line 64: Same as here:
https://github.com/taikoxyz/taiko-mono/pull/15608/files#r1470670145

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -168,4 +171,18 @@ contract SignalService is AuthorizableContract, ISignalService {
...
+    /// TrieNodes.
+    function _transcode(bytes memory proof) internal pure returns (bytes[] memory proofs) {
+        RLPReader.RLPItem[] memory nodes = RLPReader.readList(proof);
+        proofs = new bytes[](nodes.length);
+
+        for (uint256 i; i < nodes.length;) {
+            proofs[i] = RLPReader.readBytes(nodes[i]);
+            unchecked {
+                ++i;
+            }
```
adaki2004, line 185: Same as here: https://github.com/taikoxyz/taiko-mono/pull/15608/files#r1470670145

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
dantaik: Good to know.

Brechtpd (approved):

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -170,6 +173,17 @@ contract SignalService is AuthorizableContract, ISignalService {
...
+    /// @notice Translate a RLP-encoded list of RLP-encoded TrieNodes into a list of LP-encoded
+    /// TrieNodes.
+    function _transcode(bytes memory proof) internal pure returns (bytes[] memory proofs) {
+        RLPReader.RLPItem[] memory nodes = RLPReader.readList(proof);
+        proofs = new bytes[](nodes.length);
+
+        for (uint256 i; i < nodes.length; ++i) {
+            proofs[i] = RLPReader.readBytes(nodes[i]);
+        }
+    }
```
adaki2004, line 185: Obviously prefer not having this on-chain (and i know Daniel already opened a PR ).

We can tho manage to upgrade in a ways that: X messageId is old version implementation verification - after X this new ?
So basically having them living both side by side for some time.. Tehn upgrade the contracts again and exclude the old.

This way we can be compatible by not causing too much damage.

dantaik: We can do it in two PRs, after merging this one,  changing this PR to support the tokenID based handling  --https://github.com/taikoxyz/taiko-mono/pull/15609. The old code can still be deleted.

dantaik: Wait, the `messageId` is only visble on the Bridge level, not on the SignalService level, so we have to figure out another way ...

davidtaikocha (approved):

packages/protocol/contracts/signal/SignalService.sol
dantaik: I'll simply ask Korbi to use eth_getstorageat to get the SignalService's implementation address, then if address == old_impl, then use the relayer's returned proof as-is, if the addrss == new_impl, the bridge dapp should RLP decode then proof, then ABI-encode the decoded list, then use the generated data.

adaki2004 (approved):

PR #15609, opened 2024-01-30
refactor(protocol): breaking change by removing signal service overhead

dantaik:
@cyberhorsey If we merge this PR, it means the merkle proof is no longer RLP-encoded (from a `bytes[]`), but will be ABI-encoded.

@davidtaikocha once the internal devnet with 4844 is live, please upgrade our signal servce to the new impleentation in this PR. We have never tried upgrading SignalService before.

https://www.notion.so/Upgrade-SignalService-Bridge-App-Relayer-6dab176b86f04c168bd2c8299b782ba7


packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -39,13 +38,13 @@ contract SignalService is AuthorizableContract, ISignalService {
     struct Hop {
         address signalRootRelay;
         bytes32 signalRoot;
-        bytes storageProof;
+        bytes[] storageProof;
```
dantaik, line 41: these are breaking changes as relayer proof data needs to be re-formatted.

adaki2004: Actually i like it better than translating the encoded proof to a format like [adopt_optimism_new_trie_code](https://github.com/taikoxyz/taiko-mono/tree/adopt_optimism_new_trie_code) bc. it will create some on-chain overhead.

dantaik: I like this approach too, but if we merge this PR then the bridge relayer will have to be changed and we have a lot more work to do. Let the bridge relayer fix the existing bugs first.

dantaik:
@davidtaikocha once the internal devnet with 4844 is live, please upgrade our signal servce to the new impleentation in this PR. We have never tried upgrading SignalService before.

cyberhorsey:
@dantaik you're missing a `_transcode(p.storageProof)` on line 146 in SignalService.sol, which should also remove the `_transcode()`.

adaki2004:
Build fails:
kép

adaki2004 (approved):

Brechtpd (approved):
The `_transcode` function is now unused and can be deleted.

PR #15613, opened 2024-01-30
feat(protocol): get rid of new compiler warnings

dantaik:
undone the implementation() function


dantaik:
I'm trying to figure out :
Screenshot 2024-01-30 at 20 29 56

Brechtpd:
Should be possible already using `eth_getstorageat` as explained in the comments: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/61117c4db8497ba489d5e1e127565a011ed6907a/contracts/proxy/ERC1967/ERC1967Proxy.sol#L33

image

I assume the reason not to expose it directly onchain is to lower the cost of deploying the proxy contract because public functions require extra opcodes. Even though in this case the deployment cost isn't really that important, seems like the method to get the implementation contract is already easy.

dantaik:
> Should be possible already using `eth_getstorageat` as explained in the comments: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/61117c4db8497ba489d5e1e127565a011ed6907a/contracts/proxy/ERC1967/ERC1967Proxy.sol#L33
> 
> image
> 
> I assume the reason not to expose it directly onchain is to lower the cost of deploying the proxy contract because public functions require extra opcodes. Even though in this case the deployment cost isn't really that important, seems like the method to get the implementation contract is already easy.

With this approach, the dapp/client side then has no way of reading the impl to customize some logics then.

dantaik:
> I'm trying to figure out : Screenshot 2024-01-30 at 20 29 56

I have to make this change to get rid of the compiler warnings: https://github.com/taikoxyz/taiko-mono/pull/15613/commits/f36f5abb8330cd591ec575eddf5b52e0d0ec146e

Any idea?

packages/protocol/contracts/libs/LibDeploy.sol
```diff
@@ -17,8 +17,19 @@ pragma solidity 0.8.24;
 import "lib/openzeppelin-contracts/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 import "lib/openzeppelin-contracts-upgradeable/contracts/access/OwnableUpgradeable.sol";
 
+/// @title TaikoERC1967Proxy
+/// @dev A proxy that allows for reading its implementation address.
+contract TaikoERC1967Proxy is ERC1967Proxy {
+    constructor(address _logic, bytes memory _data) payable ERC1967Proxy(_logic, _data) { }
+
+    function implementation() public view returns (address) {
```
dantaik, line 25: If we explosed this address, then we can merge [this breaking change](https://github.com/taikoxyz/taiko-mono/pull/15609) without modifying the relayer's mk-proof data and only need to modify the Bridge dapp to check the bridge implementation address, if the impl is the old one, we use the bridge's return data as-is, otherwise, RLP-decode the date first into a bytes[] then use the data.

Brechtpd: I still don't really follow, shouldn't the proof data be versioned then? Because the smart contract gets some data from offchain in some format, and so may have to convert that data or not to work with the new Merkle trie lib. The current implementation used onchain doesn't say anything about the proof format supplied from offchain. Unless we of course simply force the bridge to use the updated format and so then it will have to do the conversion offchain by reading the implementation contract with `eth_getstorageat` and so supplying the expected format onchain?

dantaik:
> > With this approach, the dapp/client side then has no way of reading the impl to customize some logics then.
> 
> Dapp indeed couldn't, but client can right? And in the end it's the client that has to prepare the data so should be good?
> 
> In cases it's needed in the smart contract, the more standard way I think would be to expose a version in the implementation contract, but yeah using the implementation contract address directly also works.

I tried to add a version, but more codes are needed to verify version numbers are different among implementation. So using the impl address as the version works better.

dantaik:
> but client can right?

Client can, yes.

packages/protocol/contracts/libs/LibDeploy.sol
dantaik: There are multiple approaches possible, one of them is the dapp queries the underlying SignalService implemenation address and if it is address(abc), then it queries the relayer with an extra data "?ss_addr=abc". The relayer then convert the data to match the address abc.

Brechtpd: But this is completely offchain no? No need to expose the implementation contract in a way that works onchain?

dantaik: Having this new function doesn't hurt, right? And it may be very useful in some cases. So why not have it?

Brechtpd: Right, but `eth_getstorageat` can get you the same value without any changes?

I don't have anything against exposing the implementation contract onchain btw! I just don't think it's required to do what needs to be done.

Brechtpd: > Having this new function doesn't hurt, right? And it may be very useful in some cases. So why not have it?

Yeah agree, the deployment cost for the proxy is not important for us.

dantaik: If we have this `implementation()` function now on A6, then we can modify the Bridge dapp as follows:

- query implementation(), it returns either address_a (old) or address_b (new with the new OP code) for SignalService on, for example L1. The dapp doesn't know exactly when the upgrade happends, it doesn't have to know.
- the dapp get the proof from the relayer.
- in case of address_a, do nothing, just use whatever the relayer returns
- in case of address_b, inside the dapp, convert the relayer's proof_data into the new format, then use it.

packages/protocol/contracts/common/TaikoProxy.sol
```diff
@@ -0,0 +1,28 @@
...
+
+import "lib/openzeppelin-contracts/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+import "lib/openzeppelin-contracts-upgradeable/contracts/access/OwnableUpgradeable.sol";
+
+/// @title TaikoProxy
+/// @dev An ERC1967Proxy that allows for reading its implementation address.
+contract TaikoProxy is ERC1967Proxy {
+    constructor(address _logic, bytes memory _data) payable ERC1967Proxy(_logic, _data) { }
+
+    function implementation() public view returns (address) {
```
Brechtpd, line 25: One thing to be careful for now though is that a call to `implementation()` will always end up in the proxy contract, so any function called `implementation()` in the implementation contract cannot be reached.

Perhaps giving it more of unique name like `getProxyImplementation()` would be better because a conflict is already much more unlikely.

packages/protocol/contracts/libs/LibDeploy.sol
dantaik: Ok, maybe it's not right :(

Brechtpd: Yeah I got the idea, alternatively instead of calling `implementation()` the bridge dapp could also just call `eth_getstorageat`, but that is indeed more exotic so having the view function there may be simpler so seems good.

packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
```diff
@@ -58,15 +58,15 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
     }
 
     function pause() public virtual whenNotPaused {
-        _authorizePause(msg.sender);
         _paused = _TRUE;
         emit Paused(msg.sender);
+        _authorizePause(msg.sender);
```
dantaik, line 63: not sure why this will get rid of the new compiler error.

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -35,14 +35,14 @@ contract TaikoL2 is CrossChainOwned, ICrossChainSync {
...
-    // Golden touch address
-    address public constant GOLDEN_TOUCH_ADDRESS = 0x0000777735367b36bC9B61C50022d9D0700dB4Ec;
-
     struct Config {
         uint32 gasTargetPerL1Block;
         uint8 basefeeAdjustmentQuotient;
     }
 
+    // Golden touch address
+    address public constant GOLDEN_TOUCH_ADDRESS = 0x0000777735367b36bC9B61C50022d9D0700dB4Ec;
```
dantaik, line 44: fix lint warning

packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
Brechtpd: It's because in `SignalService.sol` and `AddressManager.sol`, `_authorizePause` always simply reverts no matter what, so any code that follows it is indeed unreachable in those cases.

Brechtpd (approved):

packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
dantaik: ah, so compiling the sub-contracts will end up giving warnings for the parent-contract. Interesting, but if we think about SignalService and AddressManager, some code unreachable seems to be the right behavior.

davidtaikocha (approved):

PR #15620, opened 2024-01-31
feat(protocol): Support hex quote

smtmfft:
SGX outputs hex, update test script to support hex directly rather than previous through JSON file which needs extra util to convert hex to json.


adaki2004:
Generally looks good.
E.g. maybe we should remove code like this: 
`// import "forge-std/console.sol";` from the deployed codes.

Also, one thing, is the failing unit tests:
kép
(Tried to update branch but it does not let me do it, says head branch modified)
kép

dantaik (approved):

dantaik:
Screenshot 2024-01-31 at 21 03 08

smtmfft:
Not sure why my compile works, I will double check.

smtmfft:
got the reason, before that sol relied on LibByteUtils which exists in thirdparty/, after the PR submission, upstream changed to another Bytes util, so the old one is removed after branch merged, I will switch to new reference and resolve new conflicts soon.

adaki2004 (approved):

PR #15639, opened 2024-02-02
refactor(protocol): move gov/ & verifiers/ up

dantaik:
Moving `gov/` and `verifiers/ `from `contracts/L1/ `to `contracts/` to make the solidity folder more flat.

One reason for doing so is that I believe `contracts/thirdparty/automata-attestation/` introduced in https://github.com/taikoxyz/taiko-mono/pull/15559/files# shall be moved to `contracts/verifier/automata-attestation/` because we didn't simply use thirdparty code as-is but made quite some modifications. Leaving the new code in `thirdparty` may not make sense.


dantaik:
Lets draft another PR after https://github.com/taikoxyz/taiko-mono/pull/15559 is merged to avoid conflicts

adaki2004:
> Lets draft another PR after #15559 is merged to avoid conflicts

I put this on low prio, and can also cause conflicts / dependency issues on SigmaPrime end, when they will try to incorporate their tests into our alpha-6 branch. I'd definitely wait until then, not to cause them headaches.

PR #15641, opened 2024-02-02
fix(protocol): imporve bridge `_proveSignalReceived` and fix genesis test

dantaik:


adaki2004:
Seems genesis tests still failing.

adaki2004 (approved):

dantaik:
> Seems genesis tests still failing.

fixed

davidtaikocha (approved):

Brechtpd:
If we want to catch reverts, why don't we just add a try/catch in `_proveSignalReceived` instead of this approach of manually constructing the calldata any manually doing the call on the address?

PR #15646, opened 2024-02-02
test(protocol): improve 1559 tests

dantaik:
I changed the logging in this 1559 simulation, only to find out that the `currentBaseFee` never change for `test_simulation_lower_traffic`, `test_simulation_higher_traffic`, and `test_simulation_target_traffic`., You can run this command to verify:

```
forge test --evm-version cancun   -vvv --mt test_simulation_lower_traffic
```

Will investigate into it during the weekend. @dantaik @davidtaikocha


adaki2004:
> Will investigate into it during the weekend. @dantaik @davidtaikocha

Just FY - Not 100% sure if anything is buggy.. We were every time below the gas target of 15M per block - so the behavior at this point is expected. (!)

I'd suggest to set the gas target in the new private testnet - relatively - very low and set up spamming with parametrizable load. This way we can simulate high / low load.

My ideal scenario would be:
1. Set a limit to 200.000.
Why 200K ? The easiest way is for example a script which constantly trying to bridge ether. (`sendMessage`).
Send ether on average costs 95-100K. (See [here](https://explorer.katla.taiko.xyz/tx/0x1a1dc48e95c110fc6e79ccc892546e7e6b2e02647f226ae813b0847b6832d324))
2. Simulate the fact that around the beginning of a network there might be some block with 0% target for a short time. (This is to test the 'ramp up' when transactions start coming)
3. Send 1 - 3 `sendMessage` in each Taiko L2 block and see the sinus wave, and the base fee changing. (Constantly 3 sendMessage would increase to "infinity", long term below will reduce it to 1 (as we see now) and long term around average will stabilize it).

PR #15657, opened 2024-02-05
chore(protocol): move `automata-attestation` and `verifiers` up in file hierarchy

dantaik:
- Moved contracts/thirdparty/automata-attestation to contracts/automata-attestation as we made quite some modification to the original code. Keeping the files in thirdparty/ may confuses oursleves.
- Moved contracts/L1/verifiers to contracts/verifiers so all proof related code is more independent of the core protocol code.
- zero logical change and used 'git mv' to keep all the revision history


adaki2004:
Will notify SigmaP - as it might refer to their workflow of integrating their tests.

adaki2004 (approved):

Brechtpd (approved):

PR #15661, opened 2024-02-05
feat(protocol): New status for 2 step briding

adaki2004:
For the `bridge UI` + `relayer` it needs to be know when the first process message is done and if is is done - and can proceed to the second transaction. (+ for the `Bridge solvency checker` it is a valuable information too)

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -32,7 +32,8 @@ contract Bridge is EssentialContract, IBridge {
         RETRIABLE,
         DONE,
         FAILED,
-        RECALLED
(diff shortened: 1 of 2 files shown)


Brechtpd:
Wouldn't it also be okay for anybody that wants to check if the message was already proven to check `proofReceipt[msgHash].receivedAt > 0`? Or is the new `PROOF_SUBMITTED` status easier to check for?

adaki2004:
> Wouldn't it also be okay for anybody that wants to check if the message was already proven to check `proofReceipt[msgHash].receivedAt > 0`? Or is the new `PROOF_SUBMITTED` status easier to check for?

True, missed that. Going to delete this PR.

PR #15665, opened 2024-02-06
feat(protocol): check if addresses ever reregistered in SGXProver

adaki2004:
One address shall be registered (during attestation) only once, otherwise it could bypass this contract's expiry check by always registering with the same attestation and getting multiple valid `instanceId`s.

While during proving, it is technically possible to register the old addresses, it is less of a problem, because the `instanceId` would be the same for those addresses and can be easily deleted without the possibility of re-registering again - while if we dont rate limit at registration level, it could shoot up legit (and same) sgx provers with different `instanceId`s.

### DIFF
--- a/packages/protocol/contracts/verifiers/SgxVerifier.sol
+++ b/packages/protocol/contracts/verifiers/SgxVerifier.sol
@@ -52,14 +52,21 @@ contract SgxVerifier is EssentialContract, IVerifier {
     /// public key shall expire after some time. (For now it is a long enough 6
     /// months setting.)
     mapping(uint256 instanceId => Instance) public instances; // slot 2
+    /// @dev One address shall be registered (during attestation) only once, otherwise it could
+    /// bypass this contract's expiry check by always registering with the same attestation and
+    /// getting multiple valid instanceIds. While during proving, it is technically possile to
+    /// register the old addresses, it is less of a problem, because the instanceId would be the
+    /// same for those addresses.
+    mapping(address instanceAddress => bool alreadyAttested) public attestationRegistered; // slot 3
 
-    uint256[48] private __gap;
+    uint256[47] private __gap;
 
     event InstanceAdded(
         uint256 indexed id, address indexed instance, address replaced, uint256 validSince
     );
     event InstanceDeleted(uint256 indexed id, address indexed instance);
 
+    error SGX_ALREADY_ATTESTED();
     error SGX_DELETE_NOT_AUTHORIZED();
     error SGX_INVALID_ATTESTATION();
     error SGX_INVALID_INSTANCE();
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/verifiers/SgxVerifier.sol
```diff
@@ -52,14 +52,21 @@ contract SgxVerifier is EssentialContract, IVerifier {
     /// public key shall expire after some time. (For now it is a long enough 6
     /// months setting.)
     mapping(uint256 instanceId => Instance) public instances; // slot 2
+    /// @dev One address shall be registered (during attestation) only once, otherwise it could
+    /// bypass this contract's expiry check by always registering with the same attestation and
+    /// getting multiple valid instanceIds. While during proving, it is technically possible to
+    /// register the old addresses, it is less of a problem, because the instanceId would be the
+    /// same for those addresses and if deleted - the attestation cannot be reused anyways.
+    mapping(address instanceAddress => bool alreadyAttested) public attestationRegistered; // slot 3
```
dong77, line 60: suggest to name addressRegistered

dantaik (approved):

packages/protocol/contracts/verifiers/SgxVerifier.sol
```diff
@@ -52,14 +52,21 @@ contract SgxVerifier is EssentialContract, IVerifier {
     /// public key shall expire after some time. (For now it is a long enough 6
     /// months setting.)
     mapping(uint256 instanceId => Instance) public instances; // slot 2
+    /// @dev One address shall be registered (during attestation) only once, otherwise it could
+    /// bypass this contract's expiry check by always registering with the same attestation and
+    /// getting multiple valid instanceIds. While during proving, it is technically possible to
+    /// register the old addresses, it is less of a problem, because the instanceId would be the
+    /// same for those addresses and if deleted - the attestation cannot be reused anyways.
```
Brechtpd, line 59: > While during proving, it is technically possible to register the old addresses, it is less of a problem, because the instanceId would be the same for those addresses and if deleted - the attestation cannot be reused anyways.

Could you explain a bit more, not sure I understand.

Brechtpd (approved):

packages/protocol/contracts/verifiers/SgxVerifier.sol
adaki2004: > > While during proving, it is technically possible to register the old addresses, it is less of a problem, because the instanceId would be the same for those addresses and if deleted - the attestation cannot be reused anyways.
> 
> Could you explain a bit more, not sure I understand.

Yes, so we have 2 approaches to make addresses valid for time `X`.

Nr.1: At registry (`addInstances()` or `registerInstance()`)
Nr.2: When a proof is submitted and we wipe away the old address, and use a new one.

What i meant here is, that the issue (of adding multiple times the address) arises at `nr.1`, at registration, because in case someone, a malicious but possibly 'attestable' instance registers itself multiple ("infinite") times it gets instance ids N, N+1, N+2, N+M... ids.. So basically making itself multiple valid instances within the contract.

The other case, where we exchanging `old` with `new`, it is not a severe problem to reuse an old address, because in any case it turns out to be kind of shady, it is "just" one, and we easily ban (delete) here with 1 simple function call, and cannot vierfy attestation again. Maybe my explanation is not so 'english', let me know how is best to adjust in the comments.

PR #15669, opened 2024-02-06
feat(relayer): two-step bridge + watchdog + full merkle proof

cyberhorsey:
1) Update the relayer `indexer` subcommand to be able to index either `MessageSent` or `MessageReceived`, add the event name to the queue name to have separate queues for each flow. We will now deploy two relayer indexers for each source-dest chain, to feed either queue (to be either Processed, or checked if they need to be Suspended).

2) Update the relayer `processor` to be able to handle the the new `two-step` bridge, as well as full merkle proofs: if a `proofReceipt` exists onchain, we are in the second step, if it doesnt, we are in the first step. The processor needs to be able to wait after sending the "MessageReceive" call if we are in the first step, and wait to finalize it. Ifi t is in the "MessageExecuted" step, the relayer can return early and acknowledge the queue message.

If a relayer exits early after processing a call in the first step, the message needs to be unacknowledged and go back onto the queue, to repeat the process, except detect it now has a ProofReceipt and can be processed without a proof.

3) Add a new `watchdog` subcommand, which reads from the new `MessageReceived` queue, and calls `isMessageSent` on the source bridge. If it exists, acknowledge the queue message succesfully. If it doesn't exist, we need to transact `suspend message` on chain for this transaction, and add that to the database for querying via a UI.

4) Add a new migration, repository, interface, implementation, and route to the relayer `API` subcommand for:
`GET /suspendedTransactions` - list all suspended transactions in a paginated response.

5) increment a Prometheus counter when we suspend, so we can use this for an alert to alert David and I.

6) Update proof changes with latest signal service/bridge code.

TODO:
- Through the UI, we probably want a way to be able to suspend an individual transaction. However, we need to be careful about how we do this - most likely the safest way is simple through the frontend using a watchdog private key. If we want to have it in the backend, then we need a way to authenticate known persons who can suspend transactions.

## Summary by CodeRabbit

- **New Features**
	- Added watchdog software for monitoring and processing messages between Ethereum chains.
	- Introduced suspended transactions handling and repository.
	- New event handling and subscription logic based on event names.
	- Enhanced message processing with proof retrieval methods.
	- Added functionality for encoding storage proofs.
	- New HTTP route for retrieving suspended transactions.
- **Enhancements**
	- Streamlined ABI definitions across ERC20, ERC721, and ERC1155 vaults.
	- Improved event definitions and added new fields in various structures for better data representation.
	- Updated server configurations to include suspended transactions repository.
	- Refined gas-related operations with new utility functions.
- **Bug Fixes**
	- Fixed message processing logic and error handling in various components.
- **Documentation**
	- Added SQL migration script documentation for creating `suspended_transactions` table.
- **Refactor**
	- Renamed and refactored methods and variables for clarity and consistency across the system.
- **Tests**
	- Introduced and updated test cases for new configurations, encoding methods, and gas operations.


cyberhorsey:
@dantaik @Brechtpd is calling "bridge.isMessageSent" on the source chain a good enough "watchdog check" when we get a MessageReceived event on the destination chain?

Or should I actually look up in the database that we found an existing `MessageSent` event from the source chain, as well?

Brechtpd:
> @dantaik @Brechtpd is calling "bridge.isMessageSent" on the source chain a good enough "watchdog check" when we get a MessageReceived event on the destination chain?
> 
> Or should I actually look up in the database that we found an existing `MessageSent` event from the source chain, as well?

I think they are pretty much identical really looking at the code because done directly together, and the `isMessageSent` code is very easy so extremely low chance there could be some bug related with just that code. So depending on `isMessageSent` sounds good to me.

cyberhorsey:
> > @dantaik @Brechtpd is calling "bridge.isMessageSent" on the source chain a good enough "watchdog check" when we get a MessageReceived event on the destination chain?
> > Or should I actually look up in the database that we found an existing `MessageSent` event from the source chain, as well?
> 
> I think they are pretty much identical really looking at the code because done directly together, and the `isMessageSent` code is very easy so extremely low chance there could be some bug related with just that code. So depending on `isMessageSent` sounds good to me.

OK sounds good, thats what I'm doing already, so this should be sufficient. When we deploy an internal devnet we can see how it operates.

davidtaikocha (approved):

dantaik (approved):

dantaik (approved):

packages/bridge-ui-v2/.env.bak
KorbinianK, line 1: Lots of files for bridge-ui-v2 that **should not be commited**. I guess because dave moved it to `bridge-ui` now.

KorbinianK (dismissed):
Lots of files for bridge-ui-v2 that should **not be committed**. I guess because dave renamed the package to `bridge-ui`.

cyberhorsey:
> Lots of files for bridge-ui-v2 that should **not be committed**. I guess because dave renamed the package to `bridge-ui`.

Gotcha, bad merge from main then. I will remove the whole folder!

xiaodino (approved):

packages/protocol/test/automata-attestation/assets/0923/v3quote.json
```diff
@@ -1,57 +1,57 @@
 {
-    "header": {
-        "version": "0x0300",
-        "attestationKeyType": "0x0200",
-        "teeType": "0x00000000",
-        "qeSvn": "0x0a00",
-        "pceSvn": "0x0f00",
-        "qeVendorId": "0x939a7233f79c4ca9940a0db3957f0607",
-        "userData": "0x12ce6af1e4a81e0ecdac427b99bb029500000000"
+  "header": {
```
KorbinianK, line 2: This file should not change

dantaik: +1

packages/protocol/test/signal/SignalService.t.sol
```diff
@@ -2,6 +2,7 @@
 pragma solidity 0.8.24;
 
 import "../TaikoTest.sol";
+import "forge-std/src/console2.sol";
```
dantaik, line 5: @cyberhorsey I added your test in another PR: [add_jeff_test](https://github.com/taikoxyz/taiko-mono/pull/16090)

davidtaikocha (approved):

dantaik:
@cyberhorsey shall we merge this PR?

mask-pp (approved):

cyberhorsey:
FYI This is just waiting for some changes I've asked Ruby to do with regards to ERC20/721/1155s (now that we have a generic invocation method, the previous method of checking which type of token we have no longer works).

KorbinianK (approved):

PR #15671, opened 2024-02-07
feat(protocol)!: sync state root rather than signal service's storage root

dantaik:
`verifyMerkleProof` in SignalService remains to be re-implemented.


packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -37,8 +37,8 @@ contract SignalService is AuthorizableContract, ISignalService {
     // storageProof represents ABI-encoded tuple of (key, value, and proof)
     // returned from the eth_getProof() API.
     struct Hop {
-        address signalRootRelay;
-        bytes32 signalRoot;
+        address sianglService;
```
adaki2004, line 40:
```suggestion
        address signalService;
```

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -139,18 +145,18 @@ contract TaikoL2 is CrossChainOwned, ICrossChainSync {
...
+        emit CrossChainSynced(uint64(block.number), l1Height, l1BlockHash, parentBlockHeader.stateRoot);
 
         // Update state variables
         l2Hashes[parentId] = blockhash(parentId);
         snippets[l1Height] = ICrossChainSync.Snippet({
             remoteBlockId: l1Height,
             syncedInBlock: uint64(block.number),
             blockHash: l1BlockHash,
-            signalRoot: l1SignalRoot
+            stateRoot: parentBlockHeader.stateRoot
```
adaki2004, line 159: This is the L2's parentBlock's `stateRoot` not the L1 state root. 🤔 Are you sure this is what we want ?

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -124,6 +125,11 @@ contract TaikoL2 is CrossChainOwned, ICrossChainSync {
             parentId = block.number - 1;
         }
 
+        // Verify parentBlockHader match the block hash
+        if (parentBlockHeader.hashBlockHeader() != blockhash(parentId)) {
```
adaki2004, line 129: We are on L2 and it seems with this you are comparing L2, but should not we need L1 ? 🤔

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -139,18 +141,18 @@ contract TaikoL2 is CrossChainOwned, ICrossChainSync {
             revert L2_BASEFEE_MISMATCH();
         }
 
-        // Store the L1's signal root as a signal to the local signal service to
+        // Store the L1's block hash as a signal to the local signal service to
         // allow for multi-hop bridging.
-        ISignalService(resolve("signal_service", false)).sendSignal(l1SignalRoot);
-        emit CrossChainSynced(uint64(block.number), l1Height, l1BlockHash, l1SignalRoot);
+        ISignalService(resolve("signal_service", false)).sendSignal(l1BlockHash);
```
Brechtpd, line 146:
```suggestion
        ISignalService(resolve("signal_service", false)).sendSignal(l1StateRoot);
```
I think.

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -117,36 +117,33 @@ contract SignalService is AuthorizableContract, ISignalService {
...
                 chainId, // use label as chainId
-                hop.signalRootRelay,
-                hop.signalRoot // as a signal
+                hop.signalService,
+                hop.stateRoot // as a signal
             );
 
             bool verified = SecureMerkleTrie.verifyInclusionProof(
-                bytes.concat(slot), hex"01", hop.storageProof, signalRoot
+                bytes.concat(slot), hex"01", hop.storageProof, stateRoot
```
Brechtpd, line 138: Here we'll also have to prove the signal root account state root (aka the signal root). But we can cache it so only the first user that wants to use this state root has to do it. 

```
mapping(address signalService => mapping(bytes32 stateRoot => bytes32 signalRoot)) public cachedSignalRoots;

if (cachedSignalRoots[hop.signalService][stateRoot] == bytes32(0)) {
  bool accountVerified = SecureMerkleTrie.verifyInclusionProof(
      bytes.concat(hop.signalService), signalRoot, _transcode(hop.accountProof), stateRoot
  );
  if (!accountVerified) return false;

  // Cache for reuse for future users so they don't have to verify the account proof anymore
  cachedSignalRoots[hop.signalService][stateRoot] = signalRoot;
}

bool verified = SecureMerkleTrie.verifyInclusionProof(
    bytes.concat(slot), hex"01", _transcode(hop.storageProof), cachedSignalRoots[hop.signalService][stateRoot]
);
if (!verified) return false;
```

Brechtpd (commented):
I think this looks good. 

For me the only chose there still is it to either use the state root as the main thing that bridging works on (currently done in this PR) or just use the block hash.

Pros of working on the state root vs using the block header:
- Less logic work to do on chain, no need to do any block header logic onchain. Gas cost to verify the block header onchain should be quite small because block headers are small, but still quite a lot of code necessary to do it.
- Simple drop in replacement for the signal root, all interfaces basically remain the same so just a different piece of data that is being synced, but now we just also first have to prove the signal root using an account Merkle proof against this state root.

Cons of working on the state root vs using the block header:
- We still have to have special logic to sync both the block header and the state root between chains. We could just always work on the block header, and this could actually be more efficient overall because we don't need to do bridging for each block, while storing the state root for each block does cost quite a bit of gas. This simplifies some of the core code in the protocol a bit, but I guess not too much.

I think at the current stage we are in, I would say the state root approach is the better one to minimize changes.

d1onys1us:
fyi @dantaik if you can hold off on merging this until i fix the tags that would be great. i didn't push the v1.0.0 protocol major release yet which establishes the base API. will take a look tomorrow.

dantaik:
> fyi @dantaik if you can hold off on merging this until i fix the tags that would be great. i didn't push the v1.0.0 protocol major release yet which establishes the base API. will take a look tomorrow.

Please mark this as ready when you are done.

packages/protocol/contracts/verifiers/PseZkVerifier.sol
```diff
@@ -144,7 +144,6 @@ contract PseZkVerifier is EssentialContract, IVerifier {
             abi.encode(
                 tran.parentHash,
                 tran.blockHash,
-                tran.signalRoot,
```
Brechtpd, line 147: We need to verify the value is correct in the zk proof, so `tran.stateRoot` should still be here.

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -93,59 +100,64 @@ contract SignalService is AuthorizableContract, ISignalService {
...
 
-        bytes32 signalRoot = ICrossChainSync(p.crossChainSync).getSyncedSnippet(p.height).signalRoot;
-
-        if (signalRoot == 0) return false;
+        bytes32 stateRoot = ICrossChainSync(p.crossChainSync).getSyncedSnippet(p.height).stateRoot;
+        if (stateRoot == 0) revert SS_CROSS_CHAIN_SYNC_ZERO_STATE_ROOT();
 
+        // If a signal is sent from chainA -> chainB -> chainC (this chain), we verify the proofs in
+        // the following order:
+        // 1. using chainC's latest stateRoot to verify that chainB's TaikoL1/TaikoL2 contract has
```
Brechtpd, line 127:
```suggestion
        // 1. using chainC's latest parent's stateRoot to verify that chainB's TaikoL1/TaikoL2 contract has
```
Maybe a bit clearer that way

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -93,59 +100,64 @@ contract SignalService is AuthorizableContract, ISignalService {
...
+        // sent a given hop stateRoot on chainB using its own signal service.
+        // 2. using the verified hop stateRoot to verify that the source app on chainA has sent a
+        // signal using its own signal service.
+        // We always verify the proofs in the reversed order.
```
Brechtpd, line 131:
```suggestion
        // We always verify the proofs in the reversed order (top to bottom).
```
This is how I think about it, but still a bit confusing anyway. :)

packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -139,18 +141,19 @@ contract TaikoL2 is CrossChainOwned, ICrossChainSync {
             revert L2_BASEFEE_MISMATCH();
         }
 
-        // Store the L1's signal root as a signal to the local signal service to
+        // Store the L1's block hash as a signal to the local signal service to
```
Brechtpd, line 144:
```suggestion
        // Store the L1's state root as a signal to the local signal service to
```

packages/protocol/contracts/verifiers/PseZkVerifier.sol
dantaik: Good catch

Brechtpd (approved):

dionysuzx (approved):

d1onys1us:
@dantaik this is okay to merge. lets not worry about breaking changes for now, we will not release the v1.0.0 yet. i will tag bcr audited tag in another way (`0.x.y+bcr.audited`)

PR #15678, opened 2024-02-07
test(protocol): add Sigma Prime tests

kirk-baird:
# What has been done

Added the following testing artifacts from SIgma Prime review
- `verifiers/`
- `thirdparty/`
- `common/`
- `L1/gov/`

Tests that don't port easily and haven't been included.
- `Bridge.t.sol`
- `tokenvault/`
- `SignalService.t.sol`
- `TaikoL1.t.sol`
- `TaikoL2.t.sol`


dantaik (approved):

packages/protocol/test/thirdparty/LibFixedPointMath.t.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: UNLICENSED
```
dantaik, line 1: @kirk-baird Do you mind adding this license to all files that you add and maybe also add 
```
/// @author Your Name <your@email.address>
```

kirk-baird: Yep done, I believe I've got all of the files now.

I'm not a lawyer but I believe since we were contracted by you, you've got ownership of the tests IP so should be able to use your license if you'd like?

adaki2004 (approved):

packages/protocol/test/thirdparty/LibFixedPointMath.t.sol
dantaik: Using our license works for us. Thank you.

dantaik: I'll update the license myself later:)

PR #15679, opened 2024-02-08
fix(protocol): fix recall not working with bridged tokens

Brechtpd:
The token passed into `receiveToken` is always the canonical token, but the recall code checks if `bridgedToCanonical[ctoken.addr].addr` exists to know if it's a bridged or canonical token, which seems incorrect because it's never a bridged token that is passed in. So the current code does not support recalling on bridged token contracts correctly because all tokens are always seen as canonical.

Previous code:
```
if (bridgedToCanonical[ctoken.addr].addr != address(0)) {
    IBridgedERC20(ctoken.addr).mint(message.owner, amount);
} else {
    ERC20(ctoken.addr).safeTransfer(message.owner, amount);
}
```

The first check would always be false because `ctoken.addr` is always a canonical token and so the bridged path would never be taken. So if indeed the bridged token needs to be used on the chain, it would not work because it would also go through the canonical token path.

I believe the code to transfer the tokens in `onMessageRecalled` should be the same code that is used in `receiveToken` with the same token address deduction to know if the bridged variant needs to be used or the canonical one.

There also seems to be no test that tests the ERC20 recall, so I'm trying to add one but it's not working yet because I'm no familiar with the test code. But working on it!


adaki2004 (approved):
LGTM.

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
```diff
@@ -170,36 +149,20 @@ contract ERC1155Vault is BaseNFTVault, ERC1155ReceiverUpgradeable {
...
+        // Transfer the ETH and tokens back to the owner
+        _transferTokens(ctoken, message.owner, tokenIds, amounts);
         message.owner.sendEther(message.value);
 
         // Emit TokenReleased event
         emit TokenReleased({
             msgHash: msgHash,
             from: message.owner,
-            token: nft.addr,
+            token: ctoken.addr,
```
adaki2004, line 165: shouldn't it be the addr? (returned by `_transferTokens(ctoken, message.owner, tokenIds, amounts);` )

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -261,14 +254,8 @@ contract ERC20Vault is BaseVault {
...
-        if (amount > 0) {
-            if (bridgedToCanonical[ctoken.addr].addr != address(0)) {
-                IBridgedERC20(ctoken.addr).mint(message.owner, amount);
-            } else {
-                ERC20(ctoken.addr).safeTransfer(message.owner, amount);
-            }
-        }
-
+        // Transfer the ETH and tokens back to the owner
+        _transferTokens(ctoken, message.owner, amount);
```
adaki2004, line 258: same, maybe catch the token address return value and emit in the `TokenReleased` ?

packages/protocol/contracts/tokenvault/ERC721Vault.sol
```diff
@@ -152,34 +135,19 @@ contract ERC721Vault is BaseNFTVault, IERC721ReceiverUpgradeable {
...
-        // send back Ether
+        // Transfer the ETH and tokens back to the owner
+        _transferTokens(ctoken, message.owner, tokenIds);
         message.owner.sendEther(message.value);
 
         emit TokenReleased({
             msgHash: msgHash,
             from: message.owner,
-            token: nft.addr,
+            token: ctoken.addr,
```
adaki2004, line 150: Same as above ? Emit address returned by `_transferTokens(ctoken, message.owner, tokenIds);`) ?

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
Brechtpd: Yeah I stuck with the previous behavior where the canonical token is the one that is always logged. But indeed using the actual token address on the chain probably better.

packages/protocol/contracts/tokenvault/ERC721Vault.sol
dantaik: Maybe we emit both the ctoken address and the actual token address?

adaki2004: Dont think it is necessary as it might not even exists (bridged). recall is always called on the source chain, so from TokenReleased (msgHash is from the same source chain) event we could identify the transactions.

dantaik: If bridged token doesn't know exist on the local chain, address(0) can be used.

Brechtpd: `TokenSent`, `TokenReceived`, and `TokenReleased` now all log the actual token address. I don't know what's convenient for offchain applications so can't really judge what data should be logged, but having those three events log the token addresses in a consistent way seems to make sense to me. So if the canonical address is something that would be helpful, I would add it to all three (or if not, to none).

dantaik: @Brechtpd I do think canonical address shall also be logged which will be convenient for the off-chain apps who doesn't keep track of all mapping between canonical token and the actual token, given that the extra cost is really small.

dantaik (approved):

PR #15680, opened 2024-02-08
feat(protocol): fix signal service multi-hop proof verification bugs

dantaik:
There are two bugs:
1. We should not allow users to specify the `crossChainSync` address, it should always be the "taiko" address on the curernt chain.
2. We should check that each hop relay -- the middleman contract that sends stateRoot as a signal to its local signal service. Previously this address is not verified correctly using the AuthorizableContract contract. Now we removed AuthorizableContract and introduce a HopRelayRegistry to verify hop relays explicitly. Note that if the signal service disabled multi-hop bridging, the HopRelayRegistry is not needed.


packages/protocol/utils/generate_genesis/taikoL2.ts
```diff
@@ -464,12 +464,6 @@ async function generateContractConfigs(
                 _paused: 1, // _FALSE
                 // Ownable2Upgradeable
                 _owner: ownerSecurityCouncil,
-                authorizedAddresses: {
```
dantaik, line 467: @davidtaikocha please check if this is the right modification. AuthorizableContract.sol is removed, and SignalService now needs an AddressMamager to init.

dionysuzx (approved):

d1onys1us:
approving as merge master. i removed the `!` as we wont worry about breaking change for now.

Brechtpd (approved):
Approved because I think this correctly links the difference chains, but still with low certainty.

dantaik:
> Approved because I think this correctly links the difference chains, but still with low certainty.

Lets disable multi-hop for now in the code.

PR #15711, opened 2024-02-09
feat(protocol)!: reserve 2 slots for TaikoData.Transition [wait for base PR to be merged first]

dantaik:


packages/protocol/test/L1/TaikoL1LibProvingWithTiers.t.sol
```diff
@@ -786,12 +786,10 @@ contract TaikoL1LibProvingWithTiers is TaikoL1TestBase {
...
-                TaikoData.Transition memory tran = TaikoData.Transition({
-                    parentHash: parentHash,
-                    blockHash: blockHash,
-                    stateRoot: stateRoot,
-                    graffiti: 0x0
-                });
+                TaikoData.Transition memory tran;
+                tran.parentHash = parentHash;
+                tran.blockHash = blockHash;
+                tran.stateRoot = stateRoot;
```
Brechtpd, line 792: I would still use the normal struct initialization, otherwise if the struct does still change it will be tricky to track all the places it needs to be updated. (Generally it's even more gas efficient like that because I think now solidity zeros the memory first and then needs to still overwrite it anyway, but the difference is so small it doesn't really matter).

dantaik: I think in test the current approach is fine. And even if we use the "normal struct initialisation", we cannot prevent individual fields from being reassigned new values. We still have to know what shall be assigned and what shall not.

Brechtpd (approved):

packages/protocol/test/L1/TaikoL1LibProvingWithTiers.t.sol
dantaik: Closed this one to use https://github.com/taikoxyz/taiko-mono/pull/15716

PR #15712, opened 2024-02-09
feat(protocol): Cache source chain signal service storage proof

adaki2004:
As discussed moved to another PR and also only caching the outter (source chain's signal service root).


packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -35,18 +35,28 @@ import "./ISignalService.sol";
 contract SignalService is AuthorizableContract, ISignalService {
     using SafeCast for uint256;
 
+    /// @dev Meaning, for this chainId A, signal service contract B, in given snapshot world state
+    /// root 'C', then here is a storage proof for that signal service.
+    mapping(
+        uint256 signalServiceChainId
+            => mapping(
+                address signalServiceAddress
+                    => mapping(bytes32 worldStateRoot => bytes signalStorageRoot)
```
Brechtpd, line 44: Do store the `signalStorageRoot` directly as a `bytes32` because that only takes up a single slot. Storing it as `bytes` requires two storage slots (the length of the array (32 bytes) + the actual storage root (32 bytes)). You can just compare against `bytes32(0)` for the non-existing case.

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -136,18 +145,31 @@ contract SignalService is AuthorizableContract, ISignalService {
             Hop memory hop = p.hops[i];
             if (hop.stateRoot == stateRoot) revert SS_INVALID_HOP_PROOF();
 
-            bytes32 label = authorizedAddresses[hop.relayerContract];
+            bytes32 label = authorizedAddresses[hop.relay];
             if (label == 0) revert SS_HOP_RELAYER_UNAUTHORIZED();
 
             uint64 hopChainId = uint256(label).toUint64();
 
+            // Do not cache hop, just the source chain's signal service storage root.
```
Brechtpd, line 153: Would cache the hops as well: https://github.com/taikoxyz/taiko-mono/pull/15683#discussion_r1484217054

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -156,16 +178,28 @@ contract SignalService is AuthorizableContract, ISignalService {
...
+
+        if (cachedRoot.length == 0) {
+            (verified, toBeCachedStorageRoot) = LibTrieProof.verifyWithAccountProof(
+                stateRoot, signalService, slot, hex"01", merkleProof, hex""
+            );
+        } else {
+            (verified, toBeCachedStorageRoot) = LibTrieProof.verifyWithAccountProof(
+                stateRoot, signalService, slot, hex"01", merkleProof, cachedRoot
+            );
+        }
```
Brechtpd, line 202: Just pass in the cached root as is, let the function do what is necessary based on its value, you already handle it anyway in `LibTrieProof`.

Brechtpd (approved):

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -35,19 +35,28 @@ import "./ISignalService.sol";
 contract SignalService is AuthorizableContract, ISignalService {
     using SafeCast for uint256;
 
+    /// @dev Meaning, for this chainId A, signal service contract B, in given snapshot world state
+    /// root 'C', then here is a storage proof for that signal service.
```
dantaik, line 39: Dani, I have a different idea of caching the info.

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -35,19 +35,28 @@ import "./ISignalService.sol";
...
+    mapping(
+        uint256 signalServiceChainId
+            => mapping(
+                address signalServiceAddress
+                    => mapping(bytes32 worldStateRoot => bytes32 signalStorageRoot)
+            )
+    ) public verifiedStorageRoots;
```
dantaik, line 46: How is this cache used then?

adaki2004: What do you mean ?
It saves at LibTrieProof the heavy on-chain calculation of the storage root.

packages/protocol/contracts/signal/SignalService.sol
adaki2004: Feel free to proceed, however you like, maybe best on a separate PR to see.

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -137,18 +145,30 @@ contract SignalService is AuthorizableContract, ISignalService {
...
+                hop.relay,
+                hop.stateRoot,
+                hop.merkleProof,
+                verifiedStorageRoots[hopChainId][hop.relay][hop.stateRoot]
             );
             stateRoot = hop.stateRoot;
         }
 
-        verifyMerkleProof(stateRoot, srcChainId, app, signal, p.merkleProof);
+        verifiedStorageRoots[srcChainId][app][stateRoot] = verifyMerkleProof(
```
dantaik, line 164: In my opinion, the underlying LibTrieProof library shall know nothing about the whole caching idea.

dantaik: I'd improve the ICrossChainSync interface and its underlying implementation to support caching of state roots from multiple chains.

adaki2004:
Closing as abandoned.

PR #15715, opened 2024-02-09
feat(protocol): Add votes and snapshot plugins to BridgedERC20

adaki2004:
So we were discussing "delegation" during the airdrop. I'm fine with it but if i recall, Daniel and Keng vouched for a one transaction: claim and delegation bundled together (so from inside the airdrop claiming contract).

To achieve this we can use the `delegateBySig()` function.

Daniel's advice was make `BridgeERC20` implement `ERC20Votes` and `ERC20Snapshot` since this way all bridge contracts could have this feature and we dont need to make extra changes to the protoocl to work with Taiko tokens. (I could not put the modules into `BridgedERC20Base` because of inheritance linearization)


packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -65,13 +73,25 @@ contract BridgedERC20 is BridgedERC20Base, IERC20MetadataUpgradeable, ERC20Upgra
...
     }
 
+    /// @notice Creates a new token snapshot.
+    function snapshot() public onlyOwner {
+        _snapshot();
+    }
+
+    /// @notice Delegate votes from the original sender (tx.origin) to `delegatee`.
+    function delegateByTxOrigin(address delegatee) public virtual {
+        _delegate(tx.origin, delegatee);
```
dantaik, line 92: I don't think it's a good idea to use tx.origin here. Maybe there is another way.

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -119,16 +139,49 @@ contract BridgedERC20 is BridgedERC20Base, IERC20MetadataUpgradeable, ERC20Upgra
...
     function _beforeTokenTransfer(
         address, /*from*/
         address to,
         uint256 /*amount*/
     )
         internal
-        virtual
-        override
+        view
+        override(ERC20Upgradeable, ERC20SnapshotUpgradeable)
```
dantaik, line 151: do we also need to call `super._before...`

dantaik:
Dani, I also suggest suggest this into two PRs, one to add snapshot and delegate function to the bridgedERC20 token; the other one to solve the airdrop-and-delegate issue -- if we can do it in on tx, it's great, otherwise, we can still use more than one transaction, *but BridgedERC20 should not use tx.origin just because we have the airdrop-and-delegate objective*.

dantaik:
We can use this function in `lib/openzeppelin-contracts-upgradeable/contracts/token/ERC20/extensions/ERC20VotesUpgradeable.sol`:

```solidity
  /**
     * @dev Delegates votes from signer to `delegatee`
     */
    function delegateBySig(
        address delegatee,
        uint256 nonce,
        uint256 expiry,
        uint8 v,
        bytes32 r,
        bytes32 s
    ) public virtual override {
  ...

```

Before the airdrop transaction, we collect a delegateSignature from the user, making sure its valid; then the user proceed with the airdrop claim function. Two signatures, one transaction.

Later, we asynchronously fire many `delegateBySig` transactions behind the scene. The misson will be accomlished this way.

adaki2004:
> Dani, I also suggest suggest this into two PRs, one to add snapshot and delegate function to the bridgedERC20 token; the other one to solve the airdrop-and-delegate issue -- if we can do it in on tx, it's great, otherwise, we can still use more than one transaction, _but BridgedERC20 should not use tx.origin just because we have the airdrop-and-delegate objective_.

I rather do it in one PR here - as this is the reason why we doing it. The changes are not so big, i dont think we shall fractionalize that much.

adaki2004:
> ```solidity
> delegateBySig
> ```

Great suggestion. Made it working via this way, and it will be in 1 transaction as you proposed - together with the claim - as you and Keng proposed for.

This way the user pays and transact, no need to store it.

Anyways, if we were about to do with 2 transactions, then i would do the :
1. claim airdrop
2. delegate (by the user, non signature based, since it is easier on the ui).

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
```diff
@@ -59,7 +76,15 @@ abstract contract MerkleClaimable is EssentialContract {
         }
 
         isClaimed[hash] = true;
-        _claimWithData(data);
```
Brechtpd, line 62: Seems a bit like a strange setup where the claiming has an extra internal function to extend the functionality. More logical to me would be to have an external `claimAndDelegate()` function (or something like that) that would call `claim` and also calls `delegateBySig()`.

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -17,14 +17,22 @@ pragma solidity 0.8.24;
...
 
 /// @title BridgedERC20
 /// @notice An upgradeable ERC20 contract that represents tokens bridged from
 /// another chain.
-contract BridgedERC20 is BridgedERC20Base, IERC20MetadataUpgradeable, ERC20Upgradeable {
+contract BridgedERC20 is
+    BridgedERC20Base,
+    IERC20MetadataUpgradeable,
+    ERC20SnapshotUpgradeable,
+    ERC20VotesUpgradeable
```
Brechtpd, line 34: The downside of having to do one size fits all, now all bridged tokens need to support the onchain voting functionality because we want it for TKO (will be nicer in the next protocol version!). :)

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -65,13 +73,21 @@ contract BridgedERC20 is BridgedERC20Base, IERC20MetadataUpgradeable, ERC20Upgra
...
+        __ERC20Permit_init(_name);
 
         // Set contract properties
         srcToken = _srcToken;
         srcChainId = _srcChainId;
         srcDecimals = _decimals;
     }
 
+    /// @notice Creates a new token snapshot.
+    function snapshot() public onlyOwner {
```
Brechtpd, line 87: By default the owner of this smart contract is also the owner of the vault. So will we have a system to give token projects ownership of the bridged contract or snapshots will simply only be used for TKO?

Brechtpd (commented):
I think delegating with a signature is a good approach. Claiming for other addresses should still be allowed I feel so that users don't have to interact with each address that has an airdrop separately, but this would be off the default path so something they could just use etherscan for or something. The UI requiring the same address claiming the tokens to also sign and delegate seems like a good default flow.

> Anyways, if we were about to do with 2 transactions, then i would do the :

Seems easy (and I think even necessary for general case) to support both cases.

packages/protocol/contracts/team/airdrop/MerkleClaimable.sol
dantaik: I think from the op's perspective, we want to make sure if some one is in the airdrop list, the address must claim by itself, no way can claim for it, so we filter out dead addresses. And we also want to make sure whenever you claim, you must delegate to someone.

dantaik: However, the extra function shall go to ERC20Aridrop contracts, not in this parent contract.

packages/protocol/contracts/tokenvault/BridgedERC20.sol
dantaik: We cannot transfer the ownership to some else, so we need another role for snapshot function.

packages/protocol/test/team/airdrop/SigUtil.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.24;
+
+// Creating the hashTypedDataV4 hash type signing
```
dantaik, line 4: The OZ library doesn't offer something?

packages/protocol/test/team/airdrop/SigUtil.sol
```diff
@@ -0,0 +1,55 @@
...
+        public
+        view
+        returns (bytes32)
+    {
+        return
+            keccak256(
+                abi.encodePacked(
+                    "\x19\x01",
+                    DOMAIN_SEPARATOR,
+                    getStructHash(_permit)
```
dantaik, line 51: shall we simply do the following without defining a struct?

```
abi.encodePacked(
                    "\x19\x01",
                    DOMAIN_SEPARATOR,
                    keccak256(
                abi.encode(
                    _DELEGATION_TYPEHASH,
                    _delegate.delegatee,
                    _delegate.nonce,
                    _delegate.expiry
                )
            );
     ```

Comment (adaki2004) [reply packages/protocol/test/team/airdrop/SigUtil.sol:51]: It is exactly the same. It is just a helper sig util. The true "thing" is to re-create on tha UI..! We need to test it back and forth.

File: packages/protocol/test/team/airdrop/SigUtil.sol (hunk 6 above)
Comment (dantaik) [inline packages/protocol/test/team/airdrop/SigUtil.sol:4]: OZ has library code to support EIP 712,

File: packages/protocol/contracts/tokenvault/BridgedERC20.sol (hunk 4 above)
Comment (adaki2004) [reply packages/protocol/contracts/tokenvault/BridgedERC20.sol:34]: Yes, my proposal was a special check in Bridge -> if TKO then mint into a modified contract.
Now every bridge contract can basically basically used in a DAO, which i'd say does not hurt, right ? Maybe it is a feature.. ? 
Deployment size does not matter bc. we anyway pre-deploy the implementation. If snapsot and voting is not used frequently also the transfer is not 'heavy' (same cost) so i dont see it a bug but rather maybe a cool feature(?).
What do you see the downsizes or side effects of it ?
Taiko can even be a chain which boost your token and make "DAO-able" :D

File: packages/protocol/test/team/airdrop/SigUtil.sol (hunk 6 above)
Comment (dantaik) [reply packages/protocol/test/team/airdrop/SigUtil.sol:4]: See:
```
import "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";

contract MessageSigner is EIP712 {
    string private constant SIGNING_DOMAIN = "MessageSignerDomain";
    string private constant SIGNATURE_VERSION = "1";

    struct Message {
        address from;
        string message;
    }

    bytes32 private constant MESSAGE_TYPEHASH = keccak256(
        "Message(address from,string message)"
    );

    function getMessageHash(Message memory message) public view returns (bytes32) {
        return _hashTypedDataV4(keccak256(abi.encode(
            MESSAGE_TYPEHASH,
            message.from,
            keccak256(bytes(message.message)) // Correctly hash the string message
        )));
    }
}
```

Comment (dantaik) [reply packages/protocol/test/team/airdrop/SigUtil.sol:4]: I also prefer to move the code to ERC20Airrop.sol or its parent contract.

File: packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol [hunk 8]
```diff
@@ -41,8 +41,19 @@ contract ERC20Airdrop is MerkleClaimable {
...
-    function _claimWithData(bytes calldata data) internal override {
+    function _claimWithData(
+        bytes calldata data,
+        address delegatee,
+        uint256 nonce,
+        uint256 expiry,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) internal override {
```
dantaik, line 52: If both ERC20Airdrop and ERC20Airdrop2 allow claiming, we shall create a parent contract and move shared code there.

packages/protocol/contracts/team/airdrop/ERC721Airdrop.sol
```diff
@@ -40,7 +40,15 @@ contract ERC721Airdrop is MerkleClaimable {
         vault = _vault;
     }
 
-    function _claimWithData(bytes calldata data) internal override {
+    function _claimWithData(
```
dantaik, line 43: ERC721Airdrop shall not be changed at all -- delegation shall not be something MerkleClaimable shall care.

dantaik (changes requested):

dantaik:
@adaki2004 I still think we should split snapshot from airdrop and have 2 PRs.

packages/protocol/contracts/team/airdrop/ERC721Airdrop.sol
adaki2004: Then how do you achieve the desired one transaction per airdrop ? I already told you and Keng i prefer the 2 time transactions, because of errors too for example (UI does not consturcts sig things properly, etc.)

adaki2004:
> @adaki2004 I still think we should split snapshot from airdrop and have 2 PRs.

Please proceed as you wish, i'll be OOF most of today and tomorrow- or wait till monday. Also would be good if the requirements would be more clear not changing from day to day.
First it was asked to bundle into one transaction, now it is possible (with built in sigantures), and you say now: `ERC721Airdrop shall not be changed at all`. If requirements are not perfectly known (and changing) it creates wasted efforts like this and double work.

dantaik:
@adaki2004 please review: https://github.com/taikoxyz/taiko-mono/pull/15727, if this PR is merged, I think you can revert all files except BridgedERC20.sol

packages/protocol/contracts/tokenvault/BridgedERC20.sol
dantaik: Please see https://github.com/taikoxyz/taiko-mono/pull/15726

dantaik (changes requested):

adaki2004:
> Before the airdrop transaction, we collect a delegateSignature from the user, making sure its valid; then the user proceed with the airdrop claim function. Two signatures, one transaction.
> 
> Later, we asynchronously fire many `delegateBySig` transactions behind the scene. The misson will be accomlished this way.
> 
> We can also ask the user to transact his `delegateBySig` transaction after receiving the airdrop.

What if we collect and the signature will not be valid afterwards ?
Either expires, tho we can ask for a long time OR they delegate before we fire away and the signature nonce is not valid anymore ?

I guess it is not a huge issue as it means the user delegated on his own.

adaki2004:
Closing as we implemented it in 2 separate PRs. (Was easier to open up a new one with BridgedERC20 than revert some changes here.)

PR #15725, opened 2024-02-10
chore(starter-dapp): clean up .default.env/README.md

ymittal:
We already have a `.default.env` file under `/starter-dapp`, so we can clean up `README.md` to reflect that.


d1onys1us:
these are just meant to be an example values, i don't think we want to populate the .default.env with these alpha-2 values. will close this PR for now because we haven't been actively maintaining this repo and might remove it / move it out.

PR #15727, opened 2024-02-10
feat(protocol): allow one-tx claim and delegation for bridged ERC20 tokens

dantaik:


Brechtpd (approved):
I think this works, I would choose composition over inheritance here because I find it easier to follow, but not any real difference between the two: https://github.com/taikoxyz/taiko-mono/pull/15731.

adaki2004 (approved):

PR #15728, opened 2024-02-10
feat(protocol)!: introduce CrossChainCache to replace CrossChainSync

dantaik:
This is an early demo of the idea of a cross-chain Cache (for block hash, state root, and signal service's storage root) that can be used by TaikoL1, TaikoL2, and Signal service.

I would like to get your feedback before proceeding with the implementation and bug fixes, @Brechtpd @adaki2004


Brechtpd:
I can't really give any comments because I don't know what the expected goals are.

The method in Dani's PR where the rollup contract stores the core blockchain related information it's responsible for (and also may have some extra super powers that allows overwriting this data in case of a security bug), and there being some other non-core contract that is responsible for storing data related to optimizing bridge transactions (the signal contract in this case) seems to make a lot of sense to me. You could store this bridging related data in some other contract, but to me the signal contract is that contract because all bridging transactions will go through there in practice anyway. If other dapps want to read cached data from there they can do so, if they also want to be able to write data in there then quite a lot of things done currently in the signal contract would also have to be moved there to make it trustless. Conceptually that could make sense, but in practice the bridging will be centralized to the signal contract (which is actually not a good thing, so booster rollups!).

dantaik:
We have a cross-chain data sync store, it's not explicit, it's called ICrossChainSync and its now part of both TaikoL1 and TaikoL2. I don't feel it's right to cache data inside the signal service as another cross-chain sync data store/cache. This job should be consolidated into a delighted contract and let signal service store the signals only.

I also don't think it's a good idea that the underlying LibTrieProof.sol shall know anything about the idea of cache at all. It shall only provide the primitive function to be used to verify the account proofs, the storage proofs, or the full proofs. The signal service should decide which functions to use based on the cross-chain data cache.

adaki2004:
I'm not a fan of design, storage and api changes 5 days before we kind of agreed to freeze the software. I have multiple problems with this.
- protocol shall be the first one to freeze because all other components (bridge ui, relayer, client) are adjusted based on it.
- we cannot make the upgrade because this way the A6 and A7 will be incompatible so we would need another testnet.

We could say "internal devnet is fine" but compared to A6 we already have many changes, personally i dont feel comfortable if we do not let the public "stress test" these new features/changes.

We are introducing high risks - with breaking design changes, where we cannot upgrade A6 - for maybe only small amount of benefits (if any) - but definitely new risks.

Brechtpd:
> We have a cross-chain data sync store, it's not explicit, it's called ICrossChainSync and its now part of both TaikoL1 and TaikoL2. I don't feel it's right to cache data inside the signal service as another cross-chain sync data store/cache. This job should be consolidated into a delighted contract and let signal service store the signals only.

The cached data in the signal contract is signal contract related data though. It only caches data related to proofs against signal contract accounts. This is different compared to the general block chain state like the blockhash and the chain state root. So having the signal service cache its own data seems reasonable enough to me.

> I also don't think it's a good idea that the underlying LibTrieProof.sol shall know anything about the idea of cache at all. It shall only provide the primitive function to be used to verify the account proofs, the storage proofs, or the full proofs. The signal service should decide which functions to use based on the cross-chain data cache.

I agree ideally it shouldn't be part of `LibTrieProof`. On the other hand, the caching logic is super simple and if it's not done there it's done elsewhere so doesn't seem like the biggest difference if we're only interested in a single code path (and I think we are, I don't immediately see this logic being used elsewhere).

One other concern I also have with putting more time in changing things how they are done now is that it won't be compatible with future protocol versions because bridging will be done very differently in designs with synchronous composabilty like the booster design (on top of already different style of doing things in booster rollups).

dantaik:
I have a better idea, I will close this one and open a new PR.

PR #15731, opened 2024-02-10
feat(protocol): claim and delegation (composition style)

Brechtpd:
Just as a demo of composition instead of inheritance

### DIFF
--- a/packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
+++ b/packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
@@ -42,20 +42,27 @@ contract ERC20Airdrop is MerkleClaimable {
         vault = _vault;
     }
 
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/team/airdrop/ERC20Airdrop.sol
```diff
@@ -42,20 +42,27 @@ contract ERC20Airdrop is MerkleClaimable {
         vault = _vault;
     }
 
-    function _claimWithData(bytes calldata data, bytes memory extraData) internal override {
-        (address user, uint256 amount) = abi.decode(data, (address, uint256));
+    function claimAndDelegate(
```
dantaik, line 45: I'm ok with this approach, but we need to override the claim function because we would like to maximize the delegation transaction.

dantaik (approved):

PR #15733, opened 2024-02-10
Docker-compose.yml and Bridge.sol updated security - Separate Environment Variables 

cryptoryda:
### # 💡**Separate Environment Variables**

- Instead of defining environment variables directly in the docker-compose.yml file, consider using a separate .env file to store sensitive information like passwords. This enhances security and makes it easier to manage different configurations for different environments.

## 👉Guide:

**Separate Environment Variables:** 

- Create a .env file in the same directory as your `docker-compose.yml` file with the following content:

```
- [1] MYSQL_DATABASE=relayer 
- [2] MYSQL_ROOT_PASSWORD=passw00d
```

**Bridge.sol**

- Function Visibility: 
Some functions like `suspendMessages`,  `banAddress`, and `processMessage` have external visibility, meaning they can be called from outside the contract. These functions perform critical operations such as modifying the state of the contract (banning addresses, suspending messages). This broadens the attack surface and could lead to unauthorized actions if not properly secured.

### PR - value:

- Main intend is to secure the `MYSQL_DATABASE` and `MYSQL_ROOT_PASSWORD` Key values.

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -97,9 +97,12 @@ contract Bridge is EssentialContract, IBridge {
         bytes32[] calldata msgHashes,
         bool toSuspend
     )
-        external
+    //@audit - function can be called from outside of thee contract if it is external
+       internal
         onlyFromOwnerOrNamed("bridge_watchdog")
     {
+          //@audit - 4 Improved visibility check and removal of magic numbers.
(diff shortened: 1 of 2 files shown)


KorbinianK:
@cryptoryda Please create two separate PRs, one for each package. We will not merge PRs that change more than one at the same time. Also make sure all the linting passes (semvar titles etc).

PR #15739, opened 2024-02-11
fix(protocol): need to fix a bug in LibTrieProof (or its test)

dantaik:
`forge test --evm-version cancun -vvv --mt test_verifyFullMerkleProof` fails

### DIFF
--- a/packages/protocol/contracts/libs/LibTrieProof.sol
+++ b/packages/protocol/contracts/libs/LibTrieProof.sol
@@ -26,6 +26,7 @@ library LibTrieProof {
      * @param stateRoot The merkle root of state tree.
      * @param addr The address of contract.
      * @param slot The slot in the contract.
+     * @param value The value to be verified.
      * @param mkproof The proof obtained by encoding storage proof.
      * @return verified The verification result.
      */
--- a/packages/protocol/contracts/libs/LibTrieProof.sol
+++ b/packages/protocol/contracts/libs/LibTrieProof.sol
@@ -28,7 +28,6 @@ library LibTrieProof {
      * @param slot The slot in the contract.
      * @param value The value to be verified.
      * @param mkproof The proof obtained by encoding storage proof.
-     * @return verified The verification result.
      */
     function verifyFullMerkleProof(
         bytes32 stateRoot,
@@ -39,7 +38,6 @@ library LibTrieProof {
     )
         internal
         pure
-        returns (bool verified)
     {
         (bytes[] memory accountProof, bytes[] memory storageProof) =
             abi.decode(mkproof, (bytes[], bytes[]));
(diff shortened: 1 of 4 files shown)


adaki2004 (approved):

adaki2004:
Issue was padding:
`bytes.concat(value)`

In previous code we simply passed the `hex"01"` to `SecureMerkleTrie` but now it went through an intermediary call (from `SignalService` -> `LibTrieProof` -> `SecureMerkleTrie` and the conversion +` bytes.concat` padded it like this:
`0x0100000000000000000000000000000000000000000000000000000000000000`
but we need this:
`0x01`

packages/protocol/contracts/libs/LibTrieProof.sol
```diff
@@ -18,22 +18,21 @@ library LibTrieProof {
...
      * @param slot The slot in the contract.
-     * @param value The value to be verified.
      * @param mkproof The proof obtained by encoding storage proof.
      * @return verified The verification result.
      */
     function verifyFullMerkleProof(
         bytes32 stateRoot,
         address addr,
         bytes32 slot,
-        bytes32 value,
```
Brechtpd, line 36: I guess could still take `bytes memory value` without extra complications to allow the lib to be used to check other values.

Brechtpd (approved):

packages/protocol/contracts/libs/LibTrieProof.sol
adaki2004: Sure thing i thought the same too, tho i tried to play around with correct conversions.

So in SignalService we do: hex"01", it translates to bytes32 and by default already it is wrong (padding and trailing zeros)
0x0100000000000000000000000000000000000000000000000000000000000000

How can we overcome this easiest ?

Brechtpd: But if you change `bytes32 value` to `bytes memory value` in `verifyFullMerkleProof` there won't be a conversion to bytes32 first right?

adaki2004: should be, yes, totally missed that. 🤣  will do real quick.

packages/protocol/contracts/libs/LibTrieProof.sol
```diff
@@ -18,22 +18,22 @@ library LibTrieProof {
...
     error LTP_INVALID_ACCOUNT_PROOF();
+    error LTP_INVALID_INCLUSION_PROOF();
 
     /**
      * Verifies that the value of a slot in the storage of an account is value.
      *
      * @param stateRoot The merkle root of state tree.
      * @param addr The address of contract.
      * @param slot The slot in the contract.
-     * @param value The value to be verified.
```
Brechtpd, line 28: comment can be added back

PR #15740, opened 2024-02-12
feat(protocol)!: improve protocol based on Brecht's internal review

Brechtpd:
These are some simplifications/optimization/small fixes/questions I have after doing a first pass review of the code, though certainly still parts that require more looking into. Nothing super important though here.

It's certainly not my goal to get all these small modifications in the current code base, but maybe some of them could and some of them could be done in a later version if deemed not important enough to still change the code at this point. I do also think that unnecessary changes should be avoided at this point.

I will also add comments to the different parts in this PR with some more context where needed.

Overall impression: it's hard to review because of the complexity of the different systems. There's a lot of code and the different optimizations that work on top of each other that just make things difficult to verify. The amount of different tricks you need to keep in your head just for all the gas optimizations to store the list of blocks and transitions is already big, and if you add the contestation part on top of it, it makes it hard. While the rollup basically just stores a list of blocks and a mapping of transitions, the current implementation makes this a challenge to easily understand. There's also a lot of features that result in a lot of conditional branches which does not make things easier.

There seems to be quite some tests missing I feel. Which I can understand very well because the amount of features and if/else cases that need to be tested is big. Take for example just the `AssignmentHook` mechanism and the amount of different things that would need to be tested to truly test all the features: payment with ETH, payment with ERC20 token, limited by time/block id/parent hash/meta hash, paying an ETH tip to coinbase, multiple hooks which also impact some of there features in non-trivial ways,... And this is only for the successful paths, you would still have to test for all the error paths as well. It's a lot of work.

I know a lot of my "keep things simple" comments have been ignored before, so I don't have any illusions or really goals to change things at this point. But I don't come with only complaints today because I actually have a simplified Taiko implementation that I started working on for the booster rollups: https://github.com/Brechtpd/taiko-mono/pull/1. It also contains a new design for how to do multi-provers and making sure buggy provers are automatically disabled. The mechanism used there would already greatly decrease the complexity compared to the current contestation system (and fixes issues I have raised in the past with that).


packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -137,7 +137,6 @@ library TaikoData {
         uint64 timestamp; // slot 6 (90 bits)
         uint16 tier;
         uint8 contestations;
-        bytes32[4] __reserved;
```
Brechtpd, line 140: `TransitionState` and `Block` are only stored in mappings and are never used as an input, so don't really necessarily need these reserved gaps. Unless if it's to maintain compatibility with onchain smart contracts using this data, then they should be kept because then still useful I think.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -43,6 +43,11 @@ contract TaikoL1 is
     TaikoData.State public state;
     uint256[100] private __gap;
 
+    modifier whenProvingNotPaused() {
```
Brechtpd, line 46: Just some simplification for the verification logic when proving is paused or nothing needs to be verified, nothing important.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -108,26 +109,13 @@ contract AssignmentHook is EssentialContract, IHook {
 
         // The proposer irrevocably pays a fee to the assigned prover, either in
         // Ether or ERC20 tokens.
-        uint256 refund;
```
Brechtpd, line 111: Here also just a simplification of the logic.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -114,6 +117,12 @@ library LibProposing {
                 blobHash: 0, // to be initialized below
                 extraData: params.extraData,
                 depositsHash: keccak256(abi.encode(depositsProcessed)),
+                // TODO(Brecht): A bit limiting this is required to be the same address as
+                // msg.sender
+                // And fees will be collected on L2 so if a smart contract is used as the proposer
+                // it requires
+                // the same contract to be deployed on L2 at the same address owned by the same
+                // entity.
```
Brechtpd, line 125: Having the address as an option I think would still be useful, otherwise have to take into account L1/L2 differences. Also could limit the composability of the smart contract because have to be very careful now which contract is `msg.sender` when submitting blocks.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -155,14 +164,11 @@ library LibProposing {
                 }
             }
 
+            // Check that the txList data range is within the max size of a blob
             if (uint256(params.txListByteOffset) + params.txListByteSize > MAX_BYTES_PER_BLOB) {
                 revert L1_TXLIST_OFFSET();
             }
 
-            if (params.txListByteSize == 0 || params.txListByteSize > config.blockMaxTxListBytes) {
```
Brechtpd, line 162: Here also just some simplification

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -198,38 +205,25 @@ library LibProposing {
...
-        blk.verifiedTransitionId = 0;
-
-        // Verify assignment authorization; if prover's address is an IProver
-        // contract, transfer Ether and call "validateAssignment" for
-        // verification.
-        // Prover can charge ERC20/NFT as fees; msg.value can be zero. Taiko
-        // doesn't mandate Ether as the only proofing fee.
-        blk.assignedProver = params.assignedProver;
+        // Create the block that will be stored onchain
+        TaikoData.Block memory blk = TaikoData.Block({
```
Brechtpd, line 209: First create the block in memory. Benefits:
- Makes sure all variables are initialized. Which really is all of them each time, would be strange if some variable would not need updating.
- Reduces gas because the block is used in the assignment hook, and so that currently requires reading from storage again. Currently that is also very expensive because the block has `bytes32[7] __reserved`, which means reading 7 slots that are not worm yet (7*800 = 5600 gas, and all other slots that are still warm also still take 100 gas).

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -145,26 +146,32 @@ library LibProving {
         // Taiko's core protocol.
         {
             address verifier = resolver.resolve(tier.verifierName, true);
-            // The verifier can be address-zero, signifying that there are no
```
Brechtpd, line 148: Just moved this check below to share, nothing else

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -145,26 +146,32 @@ library LibProving {
...
                     metaHash: blk.metaHash,
                     blobHash: meta.blobHash,
+                    // TODO(Brecht): Quite limiting this is required to be the same address as
+                    // msg.sender,
+                    // less flexibility on the prover's side for proof generation/proof submission
+                    // using
+                    // multiple accounts.
+                    // Added msgSender to allow the prover to be any address in the future.
                     prover: msg.sender,
+                    msgSender: msg.sender,
```
Brechtpd, line 163: Added different field to allow the `prover` address to be more flexible, while not breaking verifiers that require a specific `msg.sender` to work (like the current guardian proof verifier). This is now explicit instead of it being implicit before that the prover was also required to be `msg.sender`.

packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -136,17 +140,17 @@ library LibVerifying {
         // blockHash on L2.
         bytes32 blockHash = state.transitions[slot][tid].blockHash;
         bytes32 stateRoot;
-        uint64 processed;
+        uint64 numBlocksVerified;
```
Brechtpd, line 143: Just a rename, nothing else

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -45,10 +45,7 @@ contract GuardianProver is Guardians {
 
         if (approved) {
             deleteApproval(hash);
-            bytes memory data =
-                abi.encodeCall(ITaikoL1.proveBlock, (meta.id, abi.encode(meta, tran, proof)));
-            (bool success,) = resolve("taiko", false).call(data);
-            if (!success) revert PROVING_FAILED();
+            ITaikoL1(resolve("taiko", false)).proveBlock(meta.id, abi.encode(meta, tran, proof));
```
Brechtpd, line 48: Doing the normal way is a bit safer, but shouldn't really matter.

packages/protocol/contracts/L1/provers/Guardians.sol
```diff
@@ -48,34 +49,41 @@ abstract contract Guardians is EssentialContract {
...
     {
+        // We need at least MIN_NUM_GUARDIANS and at most 255 guardians (so the approval bits fit in
+        // a uint256)
         if (_newGuardians.length < MIN_NUM_GUARDIANS || _newGuardians.length > type(uint8).max) {
             revert INVALID_GUARDIAN_SET();
         }
-        if (_minGuardians < _newGuardians.length >> 1 || _minGuardians > _newGuardians.length) {
+        // Minimum number of guardians to approve is at least equal or greater than half the
+        // guardians (rounded up) and less or equal than the total number of guardians
+        if (_minGuardians < (_newGuardians.length + 1) >> 1 || _minGuardians > _newGuardians.length)
```
Brechtpd, line 59: Just made it so the minimum number of guardians is rounded up. With 5 guardians, otherwise only 2 would be required with this check, which seems low.

packages/protocol/contracts/L1/provers/Guardians.sol
```diff
@@ -48,34 +49,41 @@ abstract contract Guardians is EssentialContract {
...
+        {
             revert INVALID_MIN_GUARDIANS();
         }
 
-        // Delete current guardians data
```
Brechtpd, line 58: Also just some simplifications here below. This code seems written like it tries to optimize gas as much as possible, but this code will almost never be called, so simplicity should be preferred.

packages/protocol/contracts/L2/TaikoL2EIP1559Configurable.sol
```diff
@@ -19,34 +19,34 @@ import "./TaikoL2.sol";
 /// @title TaikoL2EIP1559Configurable
 /// @notice Taiko L2 with a setter to change EIP-1559 configurations and states.
 contract TaikoL2EIP1559Configurable is TaikoL2 {
-    Config private _config;
+    Config public customConfig;
```
Brechtpd, line 22: Just a rename here and the code below, nothing more

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -46,7 +46,10 @@ contract Bridge is EssentialContract, IBridge {
         address preferredExecutor;
     }
 
-    uint256 internal constant PLACEHOLDER = type(uint256).max;
+    // The slot in transient storage of the call context
+    // This is the keccak256 hash of "bridge.ctx_slot"
+    bytes32 private constant _CTX_SLOT =
+        0xe4ece82196de19aabe639620d7f716c433d1348f96ce727c9989a982dbadc2b9;
```
Brechtpd, line 52: Now uses transient storage to store the call context. Depending on the transaction, this may or may not save gas because transient storage does not have the refund issue like with normal storage. So the good thing about it is that it always reduces the max amount of gas needed in a transaction, and it doesn't depend on how much can be refunded.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -89,7 +92,6 @@ contract Bridge is EssentialContract, IBridge {
     /// @param _addressManager The address of the {AddressManager} contract.
     function init(address _addressManager) external initializer {
         __Essential_init(_addressManager);
-        _ctx.msgHash == bytes32(PLACEHOLDER);
```
Brechtpd, line 92: This code should of course have been `_ctx.msgHash = bytes32(PLACEHOLDER);`, but it doesn't really matter except a bit of extra gas cost for the first tx that uses it. Both `bytes32(PLACEHOLDER)` and `bytes32(0)` cases were handled correctly in the code.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -256,6 +254,10 @@ contract Bridge is EssentialContract, IBridge {
         whenNotPaused
         sameChain(message.destChainId)
     {
+        // TODO(Brecht): `message.owner`, but this is the `msg.sender` on the source chain.
+        // If the address is not owned by the same entity on the destination chain
+        // (e.g. can be the case for smart wallets/general contracts) this can give unexpected
+        // results (especially with refunding).
```
Brechtpd, line 260: Have to be careful with using address from the source chain on the destination chain! It really shouldn't be done.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -358,9 +358,10 @@ contract Bridge is EssentialContract, IBridge {
         whenNotPaused
         sameChain(message.destChainId)
     {
-        // If the gasLimit is set to 0 or isLastAttempt is true, the caller must
-        // be the message.owner.
-        if (message.gasLimit == 0 || isLastAttempt) {
+        // If isLastAttempt is true, the caller must be the message.owner.
+        // TODO(Brecht): why not allow anyone to call when message.gasLimit == 0?
+        // There is no fee to be earned, and there won't be any gas limit anyway.
+        if (isLastAttempt) {
```
Brechtpd, line 364: Explained in the comment, any reason to check for the `message.gasLimit == 0` case before?

packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
```diff
@@ -25,7 +25,12 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
     uint8 private constant _FALSE = 1;
     uint8 private constant _TRUE = 2;
 
-    uint8 private _reentry; // slot 1
+    // The slot in transient storage of the reentry lock
+    // This is the keccak256 hash of "ownerUUPS.reentry_slot"
+    bytes32 private constant _REENTRY_SLOT =
+        0xa5054f728453d3dbe953bdc43e4d0cb97e662ea32d7958190f3dc2da31d9721a;
```
Brechtpd, line 31: Also use transient storage for the reentry lock. Same explanation as for the bridge context how this helps.

packages/protocol/contracts/tokenvault/BaseVault.sol
```diff
@@ -52,8 +52,8 @@ abstract contract BaseVault is EssentialContract, IRecallableSender, IERC165Upgr
         returns (IBridge.Context memory ctx)
     {
         ctx = IBridge(msg.sender).context();
-        address sender = resolve(ctx.srcChainId, name(), false);
-        if (ctx.from != sender) revert VAULT_PERMISSION_DENIED();
+        address selfOnSourceChain = resolve(ctx.srcChainId, name(), false);
+        if (ctx.from != selfOnSourceChain) revert VAULT_PERMISSION_DENIED();
```
Brechtpd, line 56: Just a rename

packages/protocol/contracts/tokenvault/BridgedERC1155.sol
```diff
@@ -56,11 +55,14 @@ contract BridgedERC1155 is
         external
         initializer
     {
-        if (_srcToken == address(0) || _srcChainId == 0 || _srcChainId == block.chainid) {
-            revert BTOKEN_INVALID_PARAMS();
-        }
+        // Check if provided parameters are valid.
+        // The symbol and the name can be empty for ERC1155 tokens so we use some placeholder data
+        // for them instead.
+        LibBridgedToken.validateInputs(_srcToken, _srcChainId, "foo", "foo");
```
Brechtpd, line 61: Just moved this shared code to a helper function.

packages/protocol/contracts/tokenvault/BridgedERC20Base.sol
```diff
@@ -33,25 +33,26 @@ abstract contract BridgedERC20Base is EssentialContract, IBridgedERC20 {
     error BB_MINT_DISALLOWED();
 
     function changeMigrationStatus(
-        address addr,
```
Brechtpd, line 36: Just renames below for clarity, and moved out the shared conditional to `_isMigratingOut()`.

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
```diff
@@ -122,11 +116,13 @@ contract ERC1155Vault is BaseNFTVault, ERC1155ReceiverUpgradeable {
         // Check context validity
         IBridge.Context memory ctx = checkProcessMessageContext();
 
-        address _to = to == address(0) || to == address(this) ? from : to;
+        // Don't allow sending to disallowed addresses.
+        // Don't send the tokens back to `from` because `from` is on the source chain.
+        if (to == address(0) || to == address(this)) revert VAULT_INVALID_TO();
```
Brechtpd, line 121: Same issue here with using `from` from the source chain on the destination chain. It's safer to let this fail and let the user recall on the source chain.

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
```diff
@@ -156,8 +152,6 @@ contract ERC1155Vault is BaseNFTVault, ERC1155ReceiverUpgradeable {
         (CanonicalNFT memory ctoken,,, uint256[] memory tokenIds, uint256[] memory amounts) =
             abi.decode(message.data[4:], (CanonicalNFT, address, address, uint256[], uint256[]));
 
-        if (ctoken.addr == address(0)) revert VAULT_INVALID_TOKEN();
-
```
Brechtpd, line 160: Removed this check because it's not necessary? There is no similar check in `receiveToken` which depends on the same data. So either the check is needed in both or in none of them at least.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -130,10 +131,13 @@ contract ERC20Vault is BaseVault {
 
         btokenOld = canonicalToBridged[ctoken.chainId][ctoken.addr];
 
+        // TODO(Brecht): if the ctoken is on the current chain, should we check the
+        // symbol/name/decimals against the actual token?
```
Brechtpd, line 135: Strangely enough `_handleMessage` gets that data directly from the token contract, but it's not checked here. Any reason?

packages/protocol/contracts/tokenvault/ERC721Vault.sol
```diff
@@ -134,16 +130,9 @@ contract ERC721Vault is BaseNFTVault, IERC721ReceiverUpgradeable {
     {
         checkRecallMessageContext();
 
-        if (message.owner == address(0)) revert VAULT_INVALID_USER();
-        if (message.srcChainId != block.chainid) {
-            revert VAULT_INVALID_SRC_CHAIN_ID();
-        }
-
```
Brechtpd, line 141: These types of checks are also only done here in recall for ERC721 tokens, but not in any of the other vaults. So if they are necessary, which I don't think is the case, but if they are they need to be done in other vaults as well.

packages/protocol/contracts/tokenvault/LibBridgedToken.sol
```diff
@@ -32,4 +51,18 @@ library LibBridgedToken {
...
+        return string(
+            abi.encodePacked(
+                "ethereum:",
+                Strings.toHexString(uint160(srcToken), 20),
+                "@",
+                Strings.toString(srcChainId),
+                "/tokenURI?uint256="
+            )
+        );
+    }
```
Brechtpd, line 67: Taken from Optimism so that the URI for bridged tokens still points to where the data can be found. Seems reasonable?

packages/protocol/contracts/verifiers/PseZkVerifier.sol
```diff
@@ -61,17 +61,13 @@ contract PseZkVerifier is EssentialContract, IVerifier {
 
         ZkEvmProof memory zkProof = abi.decode(proof.data, (ZkEvmProof));
 
-        bytes32 instance;
```
Brechtpd, line 64: Just a small simplification, nothing more

packages/protocol/contracts/verifiers/SgxVerifier.sol
```diff
@@ -70,7 +70,6 @@ contract SgxVerifier is EssentialContract, IVerifier {
     error SGX_DELETE_NOT_AUTHORIZED();
     error SGX_INVALID_ATTESTATION();
     error SGX_INVALID_INSTANCE();
-    error SGX_INVALID_INSTANCES();
```
Brechtpd, line 73: In this file also just some very small simplifications and removed (to me) some unnecessary checks.

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: I'm ok with being able to customize the coinbase. Maybe a seperate PR for it?

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -314,6 +321,8 @@ library LibProving {
                 // In scenarios where this transition is not the first one, we
                 // straightforwardly reset the transition prover to address
                 // zero.
+                // TODO(Brecht): Is it sure that in all cases all the neccessary data is stored
```
dantaik, line 324: I ran through different branches in the proving code and it seems all fileds are initialized.

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: It seems we need to open a few PR to handle 'msg.sender' better.

packages/protocol/contracts/tokenvault/ERC1155Vault.sol
dantaik: I think it's OK to remove this check as if ctoken.addr is zero, the transfer function will revert below.

packages/protocol/contracts/tokenvault/ERC721Vault.sol
dantaik: I think we should add `if (message.srcChainId != block.chainid)` check to the other two vaults as well.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
dantaik: I think changeBridgedToken is called only on the target chain but ctoken is on the remote/source chain.

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -45,10 +45,7 @@ contract GuardianProver is Guardians {
 
         if (approved) {
             deleteApproval(hash);
-            bytes memory data =
-                abi.encodeCall(ITaikoL1.proveBlock, (meta.id, abi.encode(meta, tran, proof)));
-            (bool success,) = resolve("taiko", false).call(data);
-            if (!success) revert PROVING_FAILED();
```
dantaik, line 51: The original implementation catch different proving errors and convert them into PROVING_FAILED. I guess it's fine to not do so.

dantaik:
I got this warning so I have to spend some time to learn the transient storage.

```
Warning (2394): Transient storage as defined by EIP-1153 can break the composability of smart contracts: Since transient storage is cleared only at the end of the transaction and not at the end of the outermost call frame to the contract within a transaction, your contract may unintentionally misbehave when invoked multiple times in a complex transaction. To avoid this, be sure to clear all transient storage at the end of any call to your contract. The use of transient storage for reentrancy guards that are cleared at the end of the call is safe.
```

I think a valuable feedback is about `msg.sender` in 1) mesage ownership, and 2) block.coinbase, and 3) transition.prover. We probably need to work on these improvements.

adaki2004:
> I got this warning so I have to spend some time to learn the transient storage.

Great article [here](https://soliditylang.org/blog/2024/01/26/transient-storage/);

Quote:
"`An expected canonical use case for transient storage is cheaper reentrancy locks, which can be readily implemented with the opcodes as showcased below. However, given the caveats mentioned in the specification of EIP-1153, utmost care has to be taken for more advanced use cases of transient storage to preserve the composability of your smart contract. To raise awareness of this issue, for the time being, the compiler will emit a warning on use of tstore in assembly.`"

packages/protocol/contracts/bridge/Bridge.sol
adaki2004: My understanding is kinda changed, as Brecht highlighted.
If `message.gasLimit == 0 `-> we allow only the owner to call this. But why ? No fees to be earned by anyone so it is an extra check but might be unnecessary. What is the reason noone can call this retry on behalf of the user, as he/she does not gain any fees ?
And user can still decide if this is the `lastAttempt`, he/she can just call it with true and message status will be permanently set.

adaki2004:
Just as a benchmark, gas tested and improvements for propose/prove/verify: 
(non-tiered - tiered tests)
Propose: 9 - 10%
Prove: 1.8 - 3.5 %
Verify: 5 - 8%

dantaik:
> Just as a benchmark, gas tested and improvements for propose/prove/verify:  (non-tiered - tiered tests) Propose: 9 - 10% Prove: 1.8 - 3.5 % Verify: 5 - 8%

not sure I understand these numbers...

packages/protocol/contracts/bridge/Bridge.sol
dantaik: > we allow only the owner to call this. But why 
Because in some scenarios the users want to make the call when the timing is right, it has nothing to do with fees, it's about the timing.

packages/protocol/contracts/tokenvault/ERC721Vault.sol
adaki2004: I think none of these checks (as pointed out and deleted by Brecht is not necessary).
`message.owner == address(0)` -> Already checked in sendMessage so it does not get till the recall. If a message is maliciously constructed, proving already throw.
`message.srcChainId != block.chainid` -> Also alredy checked in sendMessage, and in any case a message is maliciously constructed, it cannot make through a proof check an also a valid context check in `checkRecallMessageContext` because it could not even go through the `sendMessage`.

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: See https://github.com/taikoxyz/taiko-mono/pull/15743

packages/protocol/contracts/bridge/Bridge.sol
adaki2004: > Because in some scenarios the user wants to make the call when the timing is right, it has nothing to do with fees, it's about the timing.

What kind of timing, or why timing ? You mean timing 'when' cheap - no congestion ? If so, still not hurt if someone else claims for us when txn would be expensive for us and we wait till the proper time. In the meantime if anyone claims for us (free) it is OK - in my opinion.

Or you mean some other kind of timing reason ?

adaki2004:
> > Just as a benchmark, gas tested and improvements for propose/prove/verify: >  (non-tiered - tiered tests) Propose: 9 - 10% Prove: 1.8 - 3.5 % Verify: 5 - 8%
> 
> not sure I understand these numbers...

Old code gas report:
Pasted Graphic 1


Brecht's code gas report:
Pasted Graphic

packages/protocol/contracts/L1/libs/LibProving.sol
Brechtpd: Currently yes! But I feel they should be separate, that's also one of the reasons we bake in the prover address in the proof so it cannot be stolen by someone else, so if somebody would like to submit someone's else's proof they should be able to do so, the prover address baked in the proof would always be seen as the actual prover to the protocol.

packages/protocol/contracts/bridge/Bridge.sol
Brechtpd: Timing is indeed an interesting case I hadn't thought of. Though this would only be for retrying messages that already failed a first time, so seems like if it was timing sensitive, a retry as fast as possible afterwards would be okay.

But fair enough, can't hurt to keep it for this potential edge case!

packages/protocol/contracts/tokenvault/LibBridgedToken.sol
Brechtpd: I just copied it directly from optimism, I haven't read the spec in any detail: https://github.com/ethereum-optimism/optimism/blob/6120d22fc8382f92c990d2c81231fa8e981ed7dd/packages/contracts-bedrock/src/universal/OptimismMintableERC721.sol#L68

packages/protocol/contracts/L1/provers/GuardianProver.sol
Brechtpd: Oh right, didn't think of that nuance.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
Brechtpd: Oh I see, I thought it also used some stored data in that case, but seems not!

packages/protocol/contracts/tokenvault/ERC721Vault.sol
dantaik: Fine with me.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: Seems it's indeed an issue.

dantaik (approved):

adaki2004 (approved):

packages/protocol/contracts/bridge/Bridge.sol
dantaik: Lets fix it with new PRs.

PR #15741, opened 2024-02-12
chore(protocol): fix genesis tests and add minor changes

dantaik:
```
      - name: protocol - Deploy L1 Contracts
        working-directory: ./packages/protocol
        run: |
          anvil &
          while ! nc -z localhost 8545; do
            sleep 1
          done
          pnpm deploy:foundry
```
still fails. 

Reverting nonReentrant implementation to the previous non-transient approach can make the test pass. Therefore, I guess the new nonReentrant implementation does have some side effect we need to figure out.

@Brechtpd may I suggest to keep the origiinal nonReentrant implementation in your PR then create a seperate PR for it?

## Update
I removed some uncessary nonReentrant guards from some function, now the above script succeeds. It does reflect that the behavior of the new nonReentrant is different and we need to care. Please review carefully if these nonReentrant usage can be removed.

### DIFF
--- a/packages/protocol/contracts/L1/provers/Guardians.sol
+++ b/packages/protocol/contracts/L1/provers/Guardians.sol
@@ -41,7 +41,14 @@ abstract contract Guardians is EssentialContract {
     /// @notice Set the set of guardians
     /// @param _newGuardians The new set of guardians
     /// @param _minGuardians The minimum required to sign
-    function setGuardians(address[] memory _newGuardians, uint8 _minGuardians) external onlyOwner {
+    function setGuardians(
+        address[] memory _newGuardians,
+        uint8 _minGuardians
+    )
+        external
+        onlyOwner
+        nonReentrant
+    {
         // We need at least MIN_NUM_GUARDIANS and at most 255 guardians (so the approval bits fit in
         // a uint256)
         if (_newGuardians.length < MIN_NUM_GUARDIANS || _newGuardians.length > type(uint8).max) {
--- a/packages/protocol/contracts/L1/provers/Guardians.sol
+++ b/packages/protocol/contracts/L1/provers/Guardians.sol
@@ -41,14 +41,7 @@ abstract contract Guardians is EssentialContract {
     /// @notice Set the set of guardians
     /// @param _newGuardians The new set of guardians
     /// @param _minGuardians The minimum required to sign
-    function setGuardians(
-        address[] memory _newGuardians,
-        uint8 _minGuardians
-    )
-        external
-        onlyOwner
-        nonReentrant
-    {
+    function setGuardians(address[] memory _newGuardians, uint8 _minGuardians) external onlyOwner {
         // We need at least MIN_NUM_GUARDIANS and at most 255 guardians (so the approval bits fit in
         // a uint256)
         if (_newGuardians.length < MIN_NUM_GUARDIANS || _newGuardians.length > type(uint8).max) {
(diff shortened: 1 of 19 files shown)


packages/protocol/contracts/L1/provers/Guardians.sol
```diff
@@ -41,14 +41,7 @@ abstract contract Guardians is EssentialContract {
     /// @notice Set the set of guardians
     /// @param _newGuardians The new set of guardians
     /// @param _minGuardians The minimum required to sign
-    function setGuardians(
-        address[] memory _newGuardians,
-        uint8 _minGuardians
-    )
-        external
-        onlyOwner
-        nonReentrant
```
dantaik, line 50: this fixes the `pnpm deploy:foundry` failure, not sure why.

adaki2004: The reason is this:
```
 Error: 
Transaction Failure: 0x63379e0e6ddddf647171f9fcde954b7a10c04896edc09938851840a7bbeaad7d
eth_chainId
EIP-3855 is not supported in one or more of the RPCs used.
```

So anvil i guess does not have dencun upgrade yet!

Brechtpd:
> Reverting nonReentrant implementation to the previous non-transient approach can make the test pass. Therefore, I guess the new nonReentrant implementation does have some side effect we need to figure out.

Oh that would be bad! If it is indeed different we have to check because it really should behave 100% identical (except for gas costs of course).

> I removed some uncessary nonReentrant guards from some function, now the above script succeeds. It does reflect that the behavior of the new nonReentrant is different and we need to care. Please review carefully if these nonReentrant usage can be removed.

I think it's safer to keep all functions `nonReentrant` just in case. Even if a specific function does not need it itself, there can still be strange behavior when a `nonReentrant` function that can call an external contract can call into a function of itself that does state changes and doesn't have the guard. So ideally any function that does state changes should have the non-reentrant guard to make sure self calls never can have unexpected consequences.

adaki2004:
> > Reverting nonReentrant implementation to the previous non-transient approach can make the test pass. Therefore, I guess the new nonReentrant implementation does have some side effect we need to figure out.
> 
> Oh that would be bad! If it is indeed different we have to check because it really should behave 100% identical (except for gas costs of course).

Issue is coming from this, i commented already:
https://github.com/taikoxyz/taiko-mono/pull/15745#discussion_r1485824839

adaki2004:
> > Reverting nonReentrant implementation to the previous non-transient approach can make the test pass. Therefore, I guess the new nonReentrant implementation does have some side effect we need to figure out.
> 
> Oh that would be bad! If it is indeed different we have to check because it really should behave 100% identical (except for gas costs of course).

Maybe bc. foundry does not suportts it yet ? It seems anvil does not.

https://github.com/taikoxyz/taiko-mono/pull/15741#discussion_r1485842477

packages/protocol/contracts/L1/provers/Guardians.sol
dantaik: oh, that's why!

dantaik:
New all tests pass (removed a line from genesis test)

Brechtpd:
Looks like we may need to explicitly enable cancun in anvil like we currently have to do in forge to be able to use the latest opcodes.

image

But I have never run the deployment script so not sure how to best test this out.

dantaik:
> Looks like we may need to explicitly enable cancun in anvil like we currently have to do in forge to be able to use the latest opcodes.
> 
> image
> 
> But I have never run the deployment script so not sure how to best test this out.

change the work flow to use ` anvil --hardfork cancun`

PR #15745, opened 2024-02-12
chore(protocol): with transient storage, the default should be 0, not 1

dantaik:
With this change, the transient slot is actually cleared earlier in the call frames rather than at the end of the tx. But realisticly, it makes no difference.

### DIFF
--- a/packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
+++ b/packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
@@ -44,7 +44,7 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
         if (_loadReentryLock() == _TRUE) revert REENTRANT_CALL();
         _storeReentryLock(_TRUE);
         _;
-        _storeReentryLock(_FALSE);
+        _storeReentryLock(0);
     }
 
     modifier whenPaused() {


packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
```diff
@@ -44,7 +44,7 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
         if (_loadReentryLock() == _TRUE) revert REENTRANT_CALL();
         _storeReentryLock(_TRUE);
         _;
-        _storeReentryLock(_FALSE);
```
dantaik, line 47: I wonder if this `_storeReentryLock(_FALSE)` call is even necessary. A transient storage is cleared automatically at the end of the transition, no?

adaki2004: Transient stroage cleared when transaction ends (for all the contracts). But contract A might call B's function (which is `nonReentrant`) multiple times. So if we don't clear, the second call would fail.

kép

packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
```diff
@@ -44,7 +44,7 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
...
+        _storeReentryLock(0);
```
dantaik, line 47: using 0 is the same as `clearing the storage`.

Brechtpd: Agreed that this is less confusing, only a single false value now instead of both `0` and `_FALSE`.

Brechtpd (approved):

adaki2004:
Cannot we make:
`uint8 private constant _FALSE = 1;`
to
`uint8 private constant _FALSE = 0;` ? 🤔

Brechtpd:
That makes pausing slightly more expensive, but yeah that doesn't matter, so simply changing the false value does seem like a good approach as well.

dantaik:
A second thought is that we should not merge this PR....

dantaik:
> Cannot we make: `uint8 private constant _FALSE = 1;` to `uint8 private constant _FALSE = 0;` ? 🤔

PR #15748, opened 2024-02-12
fix(protocol): add require statements instead of if-else

cryptoryda:
### Add require statements instead of if-else
**require statment provide more security then if-else conditions**

Example:
`  require(!_isInitialized, "Contract is already initialized");`


dantaik:
We actually prefer the if-revert approach.

PR #15757, opened 2024-02-12
ci(protocol): forge fmt --check in protocol workflow

dionysuzx:
## how things work
- lefthook should handle this for devs, but if devs disable lefthook with `LEFTHOOK=0` (which is fine!), then this lint job will error out


dantaik (approved):

adaki2004 (approved):

dantaik:
@d1onys1us Not sure why this PR cannot be merged.

PR #15758, opened 2024-02-13
feat(protocol): Add prover param

Brechtpd:
4 tests are still failing.

Also with the contesting and stuff, it's not as clear cut when to use `msg.sender` and the supplied prover... So this might be a more confusing feature than I first thought because there will need to be clear rules when which is used. So not so convinced anymore we should add this.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -73,6 +73,7 @@ library TaikoData {
     }
 
     struct TierProof {
+        address prover;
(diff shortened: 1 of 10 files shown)


Brechtpd:
Yeah agreed we can leave it for now.

>  I think the issue you raised in the other PR about 'message.owner == msg.sender' in vaults is more critical. Do you want to fix it?

Yeah will look into fixing that.

Brechtpd:
Closing for now because not that important and pretty tricky.

PR #15759, opened 2024-02-13
chore(protocol): manage deps with node_modules instead of submodules

dionysuzx:
what do you think of this change @adaki2004 @dantaik ? i am inspired from:
- https://github.com/PaulRBerg/foundry-template?tab=readme-ov-file#installing-dependencies

i always have issues with git submodules and dont really like them.

## how to update

after merging run these commands locally (make sure your index is clean):
```
git submodule deinit -f --all 
git clean -fd
```

## Summary by CodeRabbit

- **Refactor**
	- Updated import paths for OpenZeppelin contracts and other external libraries across various contracts to use the `@openzeppelin/contracts-upgradeable` namespace or correct project structure paths.
	- Adjusted import paths in scripts and tests to reflect reorganization of project structure or dependencies.


adaki2004:
This looks good to me. So  `pnpm install` , and works out of the blue, right ?

adaki2004:
> > This looks good to me. So `pnpm install` , and works out of the blue, right ?
> 
> yes, that is all. and the dependencies are frozen by tag, just as before.

say no more. ✅

adaki2004 (approved):

dantaik:
I like the idea, but it seems there are test errors.

packages/protocol/remappings.txt
```diff
@@ -0,0 +1,6 @@
+@openzeppelin/contracts-upgradeable/=node_modules/@openzeppelin/contracts-upgradeable/
```
dantaik, line 1: how about "ozu" and "oz"?

dantaik:
I like this idea, but please make sure all tests pass - there is at least a failure

adaki2004:
> I like this idea, but please make sure all tests pass - there is at least a failure

@d1onys1us i think the failure is coming from an import (and wrong remappings). That one is removed on branch https://github.com/taikoxyz/taiko-mono/pull/15781, and once incorporated into main, the base of this PR can be updated.

packages/protocol/remappings.txt
d1onys1us: i would prefer `@openzepplin/(contracts|contracts-upgradeable)/*` because:
- readability > brevity for smart contract code
- this is the name of the npm package by openzeppelin

dantaik:
@davidtaikocha I think you may need to review this PR. If I remembered correctly, you also need the codebase to work with Hardfork.

dantaik (approved):

.gitmodules
```diff
@@ -1,27 +0,0 @@
...
-	path = packages/protocol/lib/openzeppelin-contracts-upgradeable
-	url = https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable
-	tag = v4.8.2
-
-[submodule "packages/protocol/lib/openzeppelin-contracts"]
-	path = packages/protocol/lib/openzeppelin-contracts
-	url = https://github.com/OpenZeppelin/openzeppelin-contracts
-	tag = v4.8.2
-
-[submodule "packages/protocol/lib/forge-std"]
```
dantaik, line 14: you should also remove `.git/modules/packages` but not sure this is actually part of the git history.

d1onys1us: i updated the PR description. we will need to execute a command:
```
git submodule deinit -f --all 
git clean -fd
```
after pulling the changes from main.

davidtaikocha:
> @davidtaikocha I think you may need to review this PR. If I remembered correctly, you also need the codebase to work with Hardfork.

I think the codebase is still in `node_modules` right? so if the `pnpm test:genesis` test can be passed, i'm ok with that.

PR #15761, opened 2024-02-13
feat(protocol)!: re-implement multi-hop bridging with optional caching

dantaik:
This is the best mutli-hop bridging optimization I can achieve. 

@KorbinianK @cyberhorsey @xiaodino please note that the `getSyncedSnippet` now does not take any paramter and only returns the latest synced data. Please verify if this is going to work for Bridge relayer/dapp.

@Brechtpd @adaki2004 Please review SignalService. It manages all cross-chain multi-hop bridging functions without relying another contract.

See https://github.com/taikoxyz/taiko-mono/blob/improve_signal_service_and_multihop/packages/protocol/docs/multihop_bridging_deployment.md

## Summary by CodeRabbit

- **New Features**
	- Enhanced cross-chain data relay functionality with additional parameters for improved specificity and security.

- **Refactor**
	- Streamlined signal verification process for increased efficiency.
	- Simplified genesis contract setup for better maintainability.

- **Chores**
	- Removed unused variables and functions to clean up the codebase.


packages/protocol/contracts/libs/LibTrieProof.sol
```diff
@@ -20,37 +20,42 @@ library LibTrieProof {
...
+    ///
+    /// @param rootHash The merkle root of state tree or the account tree. If accountProof's length
+    /// is zero, it is used as the account's storage root, otherwise it will be used as the state
+    /// root.
+    /// @param addr The address of contract.
+    /// @param slot The slot in the contract.
+    /// @param value The value to be verified.
+    /// @param accountProof The account proof
+    /// @param storageProof The storage proof
+    /// @return storageRoot The account's storageRoot
```
dantaik, line 33:
```suggestion
    /// @return storageRoot The account's storage root
```

dantaik:
> I actually use `getSyncedSnippet` by passing in a blockNum for multi-hop bridging:
> 
> ```go
> if p.hops != nil {
> 		var hopEthClient ethClient = p.srcEthClient
> 
> 		for _, hop := range p.hops {
> 			hop.blockNum = blockNum
> 
> 			_, err := p.waitHeaderSynced(ctx, hop.headerSyncer, hopEthClient, blockNum)
> 
> 			if err != nil {
> 				return errors.Wrap(err, "p.waitHeaderSynced")
> 			}
> 
> 			snippet, err := hop.headerSyncer.GetSyncedSnippet(&bind.CallOpts{
> 				Context: ctx,
> 			},
> 				hop.blockNum,
> 			)
> 
> 			slog.Info("hop synced snippet",
> 				"syncedInBlock", snippet.SyncedInBlock,
> 				"blockNum", hop.blockNum,
> 				"blockHash", common.Bytes2Hex(snippet.BlockHash[:]),
> 			)
> 
> 			if err != nil {
> 				return errors.Wrap(err, "hop.headerSyncer.GetSyncedSnippet")
> 			}
> 
> 			blockNum = snippet.SyncedInBlock
> 
> 			hopEthClient = hop.ethClient
> 		}
> 
> 		blockNum, err = p.waitHeaderSynced(ctx, p.destHeaderSyncer, hopEthClient, blockNum)
> 		if err != nil {
> 			return errors.Wrap(err, "p.waitHeaderSynced")
> 		}
> ```
> 
> In this way, you can see that I use it as a way to find out which block num on the hop chain the previous blockHash was synced in, and then wait for that header to be synced on the next hop chain. In this way, I dont need to wait for the "latest" header to be synced - I can just check if the block number the hop the blockHash was synced in.
> 
> I could check events, and just poll the database for the events instead, but we may want to try and "freeze" as much relayer code as possible.

@cyberhorsey could you take another look at the PR. I reverted some changes.

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -71,113 +58,89 @@ contract SignalService is EssentialContract, ISignalService {
...
-            Hop memory hop = p.hops[i];
-            if (hop.stateRoot == stateRoot) revert SS_INVALID_HOP_PROOF();
-
-            if (!hrr.isRelayRegistered(_srcChainId, hop.chainId, hop.relay)) {
-                revert SS_INVALID_RELAY();
+            if (hop.cacheChainData) {
+                if (isLastHop) _relayChainData(_chainId, _SIGNAL_ROOT, signalRoot);
+                else if (isFullProof) _relayChainData(_chainId, _STATE_ROOT, hop.rootHash);
+                else _relayChainData(_chainId, _SIGNAL_ROOT, hop.rootHash);
             }
```
Brechtpd, line 122:
```suggestion
            if (hop.cacheChainData && isFullProof) {
                _relayChainData(_chainId, _SIGNAL_ROOT, signalRoot);
            }
```

Mostly following up on my comment in the previous PR https://github.com/taikoxyz/taiko-mono/pull/15736#discussion_r1486860388 because I still believe we only have to cache signal roots.

Let's say you have to prove something on another L2 and the only thing that is available on the current chain is the L1 state root. To prove something like that we need the L2 account + storage proof (to prove the signal), and then the L1 account + storage proof (to prove the L2 state root that was used, which is made available by the taiko contract). 
- Current method: we would cache is the L2 state root and then the L1 signal root. If we next have to prove something on L1, we immediately have the L1 signal root so that's efficient. However, if we want to prove something on the L2 again, then the only thing we have is the L2 state root. So we again have to provide an account + storage proof to prove another signal, and now because of the special case on `isLastHop` we do cache the signal root so future signals on that L2 will now only have to provide a storage proof.
- Updated method: we would cache the L2 signal root and the L1 signal root. Now if we have to prove something on L2 again, we immediately have the necessary signal root available. So no need for another full proof to be able to cache the L2 signal root, and we also save on caching the the L2 state root which will never be used (the path to the L2 signal root is always shorter than the path to the L2 state root).

Generally speaking, I think the intermediate hops will always be full proofs, and for each of these we should always cache the signal root and the signal root only, so for future signals on those chains we can immediately depend on a trusted signal root. The last proof can be a partial proof or a full proof depending if one of the signal roots was cached previously. If the last proof is a partial proof, that means we verify against a signal root already and so need to cache anything anymore because it will already be written to the local signal service contract (otherwise a partial proof would not work).

Or perhaps I'm misunderstanding how it works, also a possibility.

Brechtpd (commented):
This implementation is nice! Except for the caching comment which is just optimization related, I believe it's also correct, but even though the code is simple it will still take me a bit more time and more staring at it to be sure of that.

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -20,49 +20,36 @@ import "../common/ICrossChainSync.sol";
...
 
-    struct Proof {
-        uint64 height;
-        bytes merkleProof;
-        // Ensure that hops are ordered such that those closer to the signal's source chain come
-        // before others.
-        Hop[] hops;
-    }
+    bytes32 private constant _STATE_ROOT = bytes32("state_root");
+    bytes32 private constant _SIGNAL_ROOT = bytes32("signal_root");
```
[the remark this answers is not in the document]
dantaik: Done

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -71,113 +58,89 @@ contract SignalService is EssentialContract, ISignalService {
...
-            _srcSignal = hop.stateRoot;
+            bytes32 kind = isFullProof ? _STATE_ROOT : _SIGNAL_ROOT;
+            _signal = signalForChainData(_chainId, kind, hop.rootHash);
+            _chainId = hop.chainId;
+            _app = _signalService;
         }
 
-        verifyMerkleProof(stateRoot, _srcChainId, _srcApp, _srcSignal, p.merkleProof);
+        if (!isSignalSent(_app, _signal)) revert SS_LOCAL_CHAIN_DATA_NOT_FOUND();
         return true;
```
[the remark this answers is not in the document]
dantaik: Done

adaki2004:
Ok, after going through the code i have a question, with an example:

No multi-hop, just simple L1-> L2.

- Alice sends ether at blockheight 100.
- Bob sends ether at blockheight 100.

Carol (relayer) is an EOA who processes the messages on the destination chain and it obviously waits till L2 anchor syncs the L1 state root for (at least) l1 blockHeight 100 -> Then Carol is about to process indeed those messages with the appropriate proofs.
 
Carol has the right to NOT cache any data when first processes Alice's sending. But in such case, next time he also has to pay (again) for the full merkle proof (account + storage) data on-chain plus the calculation of the storage root in LibTrieProof.sol

So instead Carol (and other relayers for the same blockheight - which is now already proven tho not cached) pays X + Y amount of fees (where X> Y), Carol now pays X + X.

Why not making caching default and not optional ? 🤔  In such case, Carol (and every other relayer) would save money if they are submitting proofs for same blockheight.

I understand that there is an extra storage write during caching, which adds up to 4.8% of the original X cost (current ether bridge costs 430K for the relayer/EOA to process that message), so i guess that is the reason, right ? 
Maybe with this flexible option EOAs/relayers can configure if they want to, because maybe they are only processing 1-2 messages per 5-10 minutes so does not worth that extra 4.8% cost and dont want to pay tht so that others could bridge cheaper ?

In any case, maybe official taiko relayers could/should be 'smart relayers' and indeed use caching when load/demand is high?

dantaik:
> Ok, after going through the code i have a question, with an example:
> 
> No multi-hop, just simple L1-> L2.
> 
> * Alice sends ether at blockheight 100.
> * Bob sends ether at blockheight 100.
> 
> Carol (relayer) is an EOA who processes the messages on the destination chain and it obviously waits till L2 anchor syncs the L1 state root for (at least) l1 blockHeight 100 -> Then Carol is about to process indeed those messages with the appropriate proofs.
> 
> Carol has the right to NOT cache any data when first processes Alice's sending. But in such case, next time he also has to pay (again) for the full merkle proof (account + storage) data on-chain plus the calculation of the storage root in LibTrieProof.sol
> 
> So instead Carol (and other relayers for the same blockheight - which is now already proven tho not cached) pays X + Y amount of fees (where X> Y), Carol now pays X + X.
> 
> Why not making caching default and not optional ? 🤔 In such case, Carol (and every other relayer) would save money if they are submitting proofs for same blockheight.
> 
> I understand that there is an extra storage write during caching, which adds up to 4.8% of the original X cost (current ether bridge costs 430K for the relayer/EOA to process that message), so i guess that is the reason, right ? Maybe with this flexible option EOAs/relayers can configure if they want to, because maybe they are only processing 1-2 messages per 5-10 minutes so does not worth that extra 4.8% cost and dont want to pay tht so that others could bridge cheaper ?
> 
> In any case, maybe official taiko relayers could/should be 'smart relayers' and indeed use caching when load/demand is high?

Dani, I think you've got it right. "Smart relayers" will certainly try to figure out what to cache on the destination chain to save gas, but this requires some extra logics in the relayer code and even our own relayer is not ready now due to other priorities.

packages/protocol/contracts/signal/SignalService.sol
Brechtpd: Main question still remains though: in what case is caching the state root more efficient so that somebody would choose to cache the state root over the signal root (or both)? There is a 1:1 mapping between the state root and the signal root, and for any signal that needs to be verified onchain we always need to prove the signal root for a certain state root on any chain (and any hops used). So caching the state root doesn't seem useful to me for proving signals, because for any signal that is being proven, there already needs to be an account proof onchain that verifies the signal root as well to do anything with it. And if only the state root is cached, that same account proof would have to be put onchain again and verified in a following signal proof, so would be less efficient in all cases.

Of course, if the argument is that the state root being cached could in theory be useful for other dapps that need to read data from other contracts that are not the signal contract, it could indeed be useful! So if that is the expected use case of being able to cache state roots than fair enough! But then there's still the problem that the caching is only done through proving signals, so only people that are interested in making the proving of signals more efficient will actually cache any data, so at least for now there doesn't seem to be any incentive to cache any state roots even in that case as far as I can see.

dantaik:
> > Dani, I think you've got it right. "Smart relayers" will certainly try to figure out what to cache on the destination chain to save gas, but this requires some extra logics in the relayer code and even our own relayer is not ready now due to other priorities.
> 
> I guess some potential use case for optional caching that could make things more efficient is to not cache certain intermediate hops because only certain chains have more signals that need to be proven. Indeed caching is not that significant of a cost, but as long as the code difference between caching and not caching is small like it currently is, it could be worth it for smart relayers so they can take advantage of these types of smart optimizations that do squeeze some extra gas savings out of this option.
> 
> I guess one benefit of the approach in Dani's PR is that the smart contract automatically skipped proving the account proof if the signal root was already cached. So even with "dumb" relayers, we were able to save the account proof from being verified onchain. But of course, that was still less than ideal because the relayer would still put the account proof onchain and pay the gas cost for that, so the relayer should really just be made smarter to really make things more efficient in the best way.

I actually gave some thought to the idea of auto-optimizing the onchain verification logics even if the proof is redudant, but later I told myself: lets not do it in the smart contract. If the relayer can be optimized to do it, it should be done offchain.

adaki2004 (approved):

Brechtpd (approved):
Optimistic approval. I don't think anybody will cache state roots (https://github.com/taikoxyz/taiko-mono/pull/15761#discussion_r1489316737) so would still like to hear the expected use case, but in any case it doesn't hurt to have the option.

PR #15770, opened 2024-02-13
fix(protocol)!: add message owner parameter to vault operations

Brechtpd:
Addresses https://github.com/taikoxyz/taiko-mono/pull/15740#discussion_r1486381661

There's now a separate owner on both the source (`srcOwner`) and destination chain (`destOwner`). This is necessary because the `srcOwner` (`msg.sender`) that does the bridge operation on the source chain may not own the same address on the destination chain. That's where `destOwner` comes in, that's an address on the destination chain that the `srcOwner` also needs to control. So we have to be careful to only use `srcOwner` on the source chain and `destChain` on the destination chain to make sure the message owner actually does have access to the message on both chains as expected.

Please also double check that srcOwner is only used on the source chain, and destOwner is only used on the destination chain.

## Summary by CodeRabbit

- **New Features**
	- Introduced distinctions between source and destination owners in various contracts to enhance clarity and functionality.
- **Refactor**
	- Updated contract logic and events to accommodate the separation of `srcOwner` and `destOwner`.
	- Modified message processing and token transfer operations to reflect owner distinctions.
- **Bug Fixes**
	- Corrected permissions and gas limit checks to ensure they are based on the appropriate owner roles.


dantaik (approved):

dantaik:
@KorbinianK @cyberhorsey @xiaodino please also review this PR as there is a minor breaking change that bridge shall be aware of.

Brechtpd:
Oh was double checking the owner use, and recall still uses `message.owner` on the source chain, `message.owner` should only be used on the destination chain. Will fix.

Brechtpd:
Fixed, but please also double check that `srcOwner` is only used on the source chain, and `destOwner` is only used on the destination chain.

dantaik (approved):

dantaik:
> Fixed, but please also double check that `srcOwner` is only used on the source chain, and `destOwner` is only used on the destination chain.

Doubled checked, looks good.

d1onys1us:
@coderabbitai review

adaki2004 (approved):

PR #15788, opened 2024-02-14
Update .gitignore

damiano68:
guess better no ?

## Summary by CodeRabbit

- **Chores**
	- Updated the `.gitignore` file to better exclude specific directories from version control, enhancing project cleanliness and reducing unnecessary file tracking.


KorbinianK:
There is no need to add trailing slashes to exclude directories, not worth a PR. Each package also has their own `.gitignore` on top of this global one and we do not track any unnecessary files.

PR #15791, opened 2024-02-14
feat(protocol): option1 - different getSyncedSnippet behavior

dantaik:
The `ICrossChainSync` is not named and is confusing. Assuming in one L1 tx, we verified block 100, 101, ... 105. The Snippets of these blocks are all available via the `getSyncedSnippet` function in TaikoL1 and TaikoL2 contract. However, in TaikoL1, only the stateRoot of block 105 (the last L2 block verified in the L1 tx) is relayed (through the signal service) and verifiable for the bridge (mainly to save gas on L1). If a bridging proof uses state root of block 100, ... 104, it will fail.

However, before this PR, `getSyncedSnippet(104)` will return the block hash and state root without failure. With this PR, now if you query `getSyncedSnippet(104)`, it will now revert with a custom error. 

But if you always use `getSyncedSnippet(0)` to always use the latest verified block height, it will not revert.

@cyberhorsey I need to talk to you about this :)


cyberhorsey:
OK, I understand.

PR #15797, opened 2024-02-15
feat(protocol):  remove getSyncedSnippet and onchain state-root indexing (need discussion)

dantaik:
There is a bug in the current code - the return value of `getSyncedSnippet` doesn't guarantee the state root can be used to verify cross-chain bridging message as `relayChainData` is NOT called for each and every L2 block -- only the state root of the last verified block in a verifyBlocks() is relayed to save gas.

Another bridging example is chainA->chainB->chainC. If we cache one of chainA's block state root on chainC. `getSyncedSnippet` won't return any data associated with chainA because getSyncedSnippet is desiged to track only L1<->L2 data sync.

Take a look at the above example with 3 chains and 2 hops. The L3 can can cache state roots or signal roots of blocks from L1 and/or L2, in orders that we cannot predict (some caching is triggered by TaikoL1/TaikoL2 automatically and some are triggered by user cache options). 

On L3, we can index all these other-chain's state roots or signal roots onchain, then provide a funciton similiar to getSyncedSnippet(), but this is very expensive and not necessary. My point is that this should be done totally off-chain. Otherwise, each signal root/state root will at least have 2 storage writes per chain on which it is relayed to or cached.

Now, by subscribing to the following event of the Signal Service on each chain, you can build tables off chain:

```
  event ChainDataRelayed(
        uint64 indexed chainid,
        uint64 indexed blockId,
        bytes32 indexed kind,
        bytes32 data,
        bytes32 signal
    );
```

Each table can have the following field:

- localchainID
- localchain block height (this is available in the event's receipt)
- localchain time (also available in the event's recipt)
- remote chain  ID (multiple remote chains, not just one)
- remote chain data type: state root or signal root
- remote chain's data = state root or signal root
- remote chain block id that the data belong
- the corresponding signal on this chain
- the signal's corresponding slot can also be calculated off chain.

You can also use one table to merge all data from all chains. This is the cross-chain relay table. Then you can offer a API to the bridge relayer frontend, this API replaces `getSyncedSnippet`.


adaki2004:
Would be curious why you think it is better ? I understand that relayers could theoretically listen to events, but current system is battle tested with a relatively great performance.
If you need to rely on off-chain services (indexers/listeners) to catch events and put everything in DB given the extra complexity of 2-step bridges introduces, this will result easily in a disastrous bridging experience and a nightmare for Jeff.

dantaik:
> Would be curious why you think it is better ? I understand that relayers could theoretically listen to events, but current system is battle tested with a relatively great performance. If you need to rely on off-chain services (indexers/listeners -> which can experience heavy downtimes) to catch events and put everything in DB given the extra complexity of 2-step bridges introduces, this will result easily in a disastrous bridging experience and a nightmare for Jeff.

There is a bug in the current code - the return value of `getSyncedSnippet` doesn't guarantee the state root can be used to verify cross-chain bridging message as relayChainData is NOT called for each and every L2 block -- only the state root of the last verified block in a verifyBlocks() is relayed to save gas.

adaki2004:
> > Would be curious why you think it is better ? I understand that relayers could theoretically listen to events, but current system is battle tested with a relatively great performance. If you need to rely on off-chain services (indexers/listeners -> which can experience heavy downtimes) to catch events and put everything in DB given the extra complexity of 2-step bridges introduces, this will result easily in a disastrous bridging experience and a nightmare for Jeff.
> 
> There is a bug in the current code - the return value of `getSyncedSnippet` doesn't guarantee the state root can be used to verify cross-chain bridging message as relayChainData is NOT called for each and every L2 block -- only the state root of the last verified block in a verifyBlocks() is relayed to save gas.
> 
> Another example is chainA->chainB->chainC. If we cache one of chainA's block state root on chainC. `getSyncedSnippet` won't return any data associated with chainA because getSyncedSnippet is desiged to track only L1<->L2 data sync.

It is easily solveable if the code in `getSyncedSnippet` as you write in this [PR](https://github.com/taikoxyz/taiko-mono/pull/15791) do this:
`However, before this PR, getSyncedSnippet(104) will return the block hash and state root without failure. With this PR, now if you query getSyncedSnippet(104), it will now revert with a custom error.` or return false, or something.

This way we dont need to develop a whole new infra.

dantaik:
> Another example is chainA->chainB->chainC. If we cache one of chainA's block state root on chainC. `getSyncedSnippet` won't return any data associated with chainA because getSyncedSnippet is desiged to track only L1<->L2 data sync.

No, it doesn't solve this issue:

> Another example is chainA->chainB->chainC. If we cache one of chainA's block state root on chainC. getSyncedSnippet won't return any data associated with chainA because getSyncedSnippet is desiged to track only L1<->L2 data sync.

dantaik:
Take a look at the above example with 3 chains and 2 hops. The L3 can can cache state roots or signal roots of blocks from L1 and/or L2, in orders that we cannot predict (some caching is triggered by TaikoL1/TaikoL2 automatically and some are triggered by user cache options). 

On L3, we can index all these other-chain's state roots or signal roots onchain, then provide a funciton similiar to getSyncedSnippet(), but this is very expensive and not necessary. My point is that this should be done totally off-chain. Otherwise, each signal root/state root will at least have 2 storage writes per chain on which it is cached.

adaki2004:
> > Another example is chainA->chainB->chainC. If we cache one of chainA's block state root on chainC. `getSyncedSnippet` won't return any data associated with chainA because getSyncedSnippet is desiged to track only L1<->L2 data sync.
> 
> No, it doesn't solve this issue:
> 
> > Another example is chainA->chainB->chainC. If we cache one of chainA's block state root on chainC. getSyncedSnippet won't return any data associated with chainA because getSyncedSnippet is desiged to track only L1<->L2 data sync.

Well, IMO we should only cache SignalService storage roots and the associated block height, nothing else.
Not state roots -> This is the real cost saving. (See LibMerkleTrie, we need to build - cost ineffectively - the storage root of the signal service.).

So in case you example:
`chainA->chainB->chainC. If we cache one of chainA's block state root on chainC. getSyncedSnippet won't return any data associated with chainA because getSyncedSnippet is desiged to track only L1<->L2 data sync.`


Okay, so i think best if we break down the workflow first and then from the common ground we proceed further (what do we need what not):

A multi-hop workflow example you described:

1.
On Chain A:
Alice wants to send ether to Chain C at blockheight `100`. 
Bob wants to send ether to Chain B at blockheight `100` too.

2.
On Chain B: 
Relayer checks if the blockheight `100` (or higher blockheight - because as you said not always all of the block height will be directly synced as you said, so for the sake of simplicity blockheight `101` is synced) of `chain A` synced in `chain B` and if true, then it query's Chain B's height, let's say it is `2000`.

3. 
On Chain C:
Relayer waits until ChainB's blockheight `2000` (or a higher blockNumber, let's say `2001`, bc. `2000` was skipped) is synced. If this is true, then :

1. Relayer will query full merkle proof from ChainA's `101` and full proof proof from ChainB's `2001`. It will submit Alice's transaction and cache both of the chain ID **signal service storage roots** (not the state roots of the blockchain  - because caching blockhcain state root, you still need to recreate SS storage proof anyways - please see what we could save in red circle in the below diagram.) (And this is implemented here btw.)

NOW if point 1 cached both of the chain IDs **signal service storage roots**, we can skip a lot  (see picture in the bottom).

Back to **removing snippet**:  This would make relayer life much harder not easier, becasue we need to have a completely new infra for that (reading, writing to backend instead of query the smart contracts) so a `getSnippet()` is pretty handy i think.

So my opinions are: 
- we dont need to cached state roots. We need to cache **signal service storage roots**
-  we should not rush out changes and 2-4 new PRs/day in the signal service (and design) like there is no tomorrow, this does not help to understand the mechanisms. A best would be a collaborative meeting to discuss together what might be needed what not. We are planning to launch very soon and every change to things which are battle tested is a huge risk.


Currently:
kép


Real saving from caching signal service storage root:
kép


@dantaik @cyberhorsey @Brechtpd @davidtaikocha Please let me know if my above logic is false, i might not understand something and i reserve the right to be horribly wrong :)

dantaik:
> Well, IMO we should only cache SignalService storage roots and the associated block height, nothing else.
Not state roots -> This is the real cost saving. (See LibMerkleTrie, we need to build - cost ineffectively - the storage root of the signal service.).

Why did we remove signal service from the core protocol in the first place after A6?

adaki2004:
> > Well, IMO we should only cache SignalService storage roots and the associated block height, nothing else.
> > Not state roots -> This is the real cost saving. (See LibMerkleTrie, we need to build - cost ineffectively - the storage root of the signal service.).
> 
> Why did we remove signal service from the core protocol in the first place after A6?

I guess arhcive node dependency, no ? But not cahcing via anchor. Caching via bridging i meant.

dantaik:
Screenshot 2024-02-15 at 14 44 08

I made this diagram to show you that the original ICrossChainSync is designed fo cross-chain exchange and keep state root (and block hash) between TaikoL1 contract and TaikoL2 contract. In a 2-hop/3 chain set up. There are 2 TaikoL1 deployed and 2 TaikoL2 deployed.  On L3, calling `getSyncedSnippet` won't return any data associated with the L1 chain, so it doens't work.

adaki2004:
> I made this diagram to show you that the original ICrossChainSync is designed fo cross-chain exchange and keep state root (and block hash) between TaikoL1 contract and TaikoL2 contract. In a 2-hop/3 chain set up. There are 2 TaikoL1 deployed and 2 TaikoL2 deployed. On L3, calling `TaikoL2.getSyncedSnippet` won't return any data associated with the L1 chain, so it doens't work if you want to verify bridged message sent from L1 to L3.

ICrossChainSync does not indeed and in my proposal the purpose of that one is not to store things related to storage proofs, but to give relayer information at which height we are synced with!

What does the caching is this - in the SignalService:
 ```
    /// @dev Meaning, for this chainId A, signal service contract B, in given snapshot world state
    /// root 'C', then here is a storage proof for that signal service.
    mapping(
        uint256 signalServiceChainId
            => mapping(
                address signalServiceAddress
                    => mapping(bytes32 worldStateRoot => bytes32 signalStorageRoot)
            )
    ) public verifiedStorageRoots;
```

**Comment by dantaik** [comment]
> > I made this diagram to show you that the original ICrossChainSync is designed fo cross-chain exchange and keep state root (and block hash) between TaikoL1 contract and TaikoL2 contract. In a 2-hop/3 chain set up. There are 2 TaikoL1 deployed and 2 TaikoL2 deployed. On L3, calling `TaikoL2.getSyncedSnippet` won't return any data associated with the L1 chain, so it doens't work if you want to verify bridged message sent from L1 to L3.
> 
> ICrossChainSync does not indeed and in my proposal the purpose of that one is not to store things related to storage proofs, but to give relayer information at which height we are synced with!
> 
> What does the caching is this - in the SignalService:
> 
> ```
>    /// @dev Meaning, for this chainId A, signal service contract B, in given snapshot world state
>    /// root 'C', then here is a storage proof for that signal service.
>    mapping(
>        uint256 signalServiceChainId
>            => mapping(
>                address signalServiceAddress
>                    => mapping(bytes32 worldStateRoot => bytes32 signalStorageRoot)
>            )
>    ) public verifiedStorageRoots;
> ```

If you don't agree the current design is better, I'm fine. But I won't try to convince you.

**Comment by adaki2004** [comment]
> If you don't agree the current design is better, I'm fine. But I won't try to convince you.

I'm fine with the design (general) as i approved the PR, i just wanted to sparkle a conversation. :)

On the other hand, i think here, in this pr, removing the `getSyncedSnippet` will make our lives much much harder - but let's really speak to Jeff and see what he suggests - i totally accept i could be wrong with my take.

**Comment by Brechtpd** [comment]
I don't believe I know the necessary requirements, so I'm just sharing this because it may or may not be useful/interesting.

If the synced snippet things should work only for data that is actually available in the signal service, the synced snippet functionality should just move there. Then it's possible to mostly get the same data directly from the signal service by making 2 changes:
- Allow signals to store data other than `1`: I don't think there's actually any reason not to also allow storing other kinds of data at the signal slot. So for the chain data we can actually store the state/signal root onchain in the slot.
- Let the signal service track what is the last synced blockId for each chain and data type

I think this achieves most things of the snippet functionality without much extra cost (only extra cost compared to vanilla signal service is the storing of the latest blockId per chain/kind when a newer blockId is used). It also allows getting the sync state for any chain, so does away with the limitation here of the original `getSyncedSnippet`. It also only allows getting the data that is actually synced for the chain because instead of depending on data stored in the rollup, we only depend on data that is actually available in the signal contract.

One limitation is that there is no (or not much) extra cost as long as only the state/signal root needs to be known (fits in 32 bytes). If more data is required per block, like for example `syncedInBlock`), more data would need to be stored/loaded at `slot + 1` for example. So definitely possible, but at a cost.

How this compares to the pure offchain method described here, I do not know and I do not know what is better. Just sharing this different approach for other people that do know.

```
contract SignalService {
    event SnippetRelayed(
        uint64 indexed chainid, uint64 indexed blockid, bytes32 indexed kind, bytes32 data
    );

   mapping(uint64 blockId => mapping(bytes32 bytes32=> uint64) private lastSyncedBlockId;

    function relayChainData(
        uint64 chainId,
        uint64 blockId,
        bytes32 kind,
        bytes32 data
    )
        external
        returns (bytes32 slot)
    {
        if (!isRelayerAuthorized[msg.sender]) revert SS_UNAUTHORIZED();
        return _relayChainData(chainId, blockId, kind, data);
    }

    function signalForChainData(
        uint64 chainId,
        uint64 blockId,
        bytes32 kind
    )
        public
        pure
        returns (bytes32)
    {
        return keccak256(abi.encode(chainId, blockId, kind));
    }

    function _relayChainData(
        uint64 chainId,
        uint64 blockId,
        bytes32 kind,
        bytes32 data
    )
        internal
        returns (bytes32 slot)
    {
        // Store the latest verified state per chain
        if (blockId > lastSyncedBlockId[chainId][kind]) {  
            lastSyncedBlockId[chainId][kind] = blockId;
        }
        bytes32 signal = signalForChainData(chainId, blockId, kind);
        emit SnippetRelayed(chainId, blockId, kind, data);
        return _sendSignal(address(this), signal, data);
    }

    function _sendSignal(address sender, bytes32 signal, bytes32 data) internal returns (bytes32 slot) {
        if (signal == 0) revert SS_INVALID_SIGNAL();
        if (data == 0) revert SS_INVALID_SIGNAL();
        slot = getSignalSlot(uint64(block.chainid), sender, signal);
        bytes32 currentData;
        assembly {
            currentData := sload(slot)
            sstore(slot, data)
        }
        require(currentData == bytes32(0) || currentData == data, "invalid signal");
    }

    function getSignalSent(address app, bytes32 signal) public view returns (bytes32 value) {
        if (signal == 0) revert SS_INVALID_SIGNAL();
        if (app == address(0)) revert SS_INVALID_APP();
        bytes32 slot = getSignalSlot(uint64(block.chainid), app, signal);
        assembly {
            value := sload(slot)
        }
    }

    function getSyncedSnippet(
        uint64 chainId,
        uint64 blockId,
        bytes32 kind,
    )
        external
        view
        returns (bytes32)
    {
        uint64 _blockId = blockId == 0 ? lastVerifiedBlockId[chainId][kind] : blockId;
        bytes32 signal = signalForChainData(chainId, _blockId, kind);
        bytes32 data = getSignalSent(address(this), signal);
        require(data != bytes32(0), "block state not synced");
        return data;
    }
}
```

dantaik:
Brecht, if keeping the snippet of the highest block for each chain works, I would do it as you suggested. I'll have to confirm if this is sufficient for the relayer. Otherwise, I still think relayer should index data offchain.

PR #15799, opened 2024-02-15
chore(protocol): rename a SignalService event and change the return of relayChainData()

dantaik:


packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -67,7 +67,7 @@ contract SignalService is EssentialContract, ISignalService {
     )
         external
         onlyFromNamed("taiko")
-        returns (bytes32 slot)
+        returns (bytes32 signal)
```
dantaik, line 70: with the signal, we can get the slot with getSignalSlot

adaki2004 (approved):

PR #15802, opened 2024-02-15
fix(protocol): tstore is not suppported on L2 now

dantaik:
According to David C, L2 is not on the same hardfork.
We can somehow use tstore for L1 contracts. But not sure what's the best way of doing it.

See https://github.com/foundry-rs/foundry/issues/6099

## Summary by CodeRabbit

- **Refactor**
	- Renamed and updated logic for reentry handling in `OwnerUUPSUpgradable.sol`.
	- Introduced a new constant and updated context functions in `Bridge.sol` for improved functionality.
- **Chores**
	- Enhanced deployment script with EVM version specification for better compatibility and testing.
- **New Features**
	- Adjusted contract configuration generation to include new property settings, enhancing the contract's capabilities.

### DIFF
--- a/packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
+++ b/packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
@@ -30,7 +30,7 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
     bytes32 private constant _REENTRY_SLOT =
         0xa5054f728453d3dbe953bdc43e4d0cb97e662ea32d7958190f3dc2da31d9721a;
 
-    uint8 private _reentryDeprecated; // slot 1
+    uint8 private _reentry; // slot 1
     uint8 private _paused;
     uint256[49] private __gap;
 
@@ -94,15 +94,17 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
 
     // Stores the reentry lock
     function _storeReentryLock(uint8 reentry) private {
-        assembly {
-            tstore(_REENTRY_SLOT, reentry)
-        }
+         _reentry = reentry;
+        // assembly {
+        //     tstore(_REENTRY_SLOT, reentry)
+        // }
     }
 
     // Loads the reentry lock
     function _loadReentryLock() private view returns (uint8 reentry) {
-        assembly {
-            reentry := tload(_REENTRY_SLOT)
-        }
+        reentry = reentry;
+        // assembly {
+        //     reentry := tload(_REENTRY_SLOT)
+        // }
     }
 }
--- a/packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
+++ b/packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
@@ -97,8 +97,8 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
     }
 
     // Loads the reentry lock
-    function _loadReentryLock() internal view virtual returns (uint8 /*reentry*/ ) {
-        return _reentry;
+    function _loadReentryLock() internal view virtual returns (uint8 reentry) {
+        reentry = _reentry;
         // assembly {
         //     reentry := tload(_REENTRY_SLOT)
         // }
--- a/packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
+++ b/packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
@@ -88,23 +88,23 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
         _paused = _FALSE;
     }
 
-    function _inNonReentrant() internal view returns (bool) {
-        return _loadReentryLock() == _TRUE;
-    }
-
     // Stores the reentry lock
-    function _storeReentryLock(uint8 reentry) private {
-         return reentry;
+    function _storeReentryLock(uint8 reentry) internal virtual {
+        return reentry;
         // assembly {
         //     tstore(_REENTRY_SLOT, reentry)
         // }
     }
 
     // Loads the reentry lock
-    function _loadReentryLock() private view returns (uint8 reentry) {
+    function _loadReentryLock() internal view virtual returns (uint8 reentry) {
         _reentry = reentry;
         // assembly {
         //     reentry := tload(_REENTRY_SLOT)
         // }
+    }
+
+    function _inNonReentrant() internal view returns (bool) {
+        return _loadReentryLock() == _TRUE;
     }
 }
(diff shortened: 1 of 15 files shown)


packages/protocol/foundry.toml
```diff
@@ -11,6 +11,8 @@ optimizer_runs = 200
 ffi = true
 gas_limit = '18446744073709551615'
 memory_limit = 2073741824
+
+# David C: do we need to remove this as well?
```
davidtaikocha, line 15: My opinion is let's remove this, which may make others think we have the same version of EVM on both L1 and L2.

davidtaikocha (approved):

packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
```diff
@@ -88,21 +88,23 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
...
-    function _inNonReentrant() internal view returns (bool) {
-        return _loadReentryLock() == _TRUE;
-    }
-
     // Stores the reentry lock
-    function _storeReentryLock(uint8 reentry) private {
-        assembly {
-            tstore(_REENTRY_SLOT, reentry)
-        }
+    function _storeReentryLock(uint8 reentry) internal virtual {
```
dantaik, line 92: so when you deploy on L1 for mainnet, maybe you want to use the tstore manually. @davidtaikocha

Brechtpd:
- Would it be possible to switch on `block.chainid == 1` to enable/disable the use of the tstore version? Because we'll want to support cancun as soon as possible on L2 I assume (even for mainnet), so having both code paths running seems like a good test (and this part being part of the audit)?
- `Bridge.sol` also uses transient storage for the context, so will also have to be updated to remove it/conditionally disable it.

dantaik:
> * block.chainid == 1

A good idea, but I wonder if it will work.
Will give it a try and let Dave to test.

davidtaikocha:
> * Would it be possible to switch on `block.chainid == 1` to enable/disable the use of the tstore version? Because we'll want to support cancun as soon as possible on L2 I assume (even for mainnet), so having both code paths running seems like a good test? (and this code path being part of the audit as well)
> * `Bridge.sol` also uses transient storage for the context, so will also have to be updated to remove it/conditionally disable it.

I dont think https://github.com/taikoxyz/taiko-mono/pull/15810 will work, since the compiled bytecode will still contain the `TSTORE` opcode... right?

Brechtpd:
> I dont think #15810 will work, since the compiled bytecode will still contain the `TSTORE` opcode... right?

Yes but that is currently okay right I think? Because currently the code is not checked yet when it is deployed (maybe in the future it will), so if some random opcode/byte value  is part of the bytecode but it is not executed, I think it should be fine.

dantaik:
I don't think it will work. Please see https://github.com/foundry-rs/foundry/issues/6099

dantaik:
> Bridge.sol also uses transient storage for the context, so will also have to be updated to remove it/conditionally disable it.

Good point, I think we need also to remove it.

Brechtpd:
> I don't think it will work. Please see [foundry-rs/foundry#6099](https://github.com/foundry-rs/foundry/issues/6099)

Oh right this is a good point. We do have to compile with 0.8.24 for both the L1 and the L2 contracts, so if for some reason the compiler also chooses to use some other functionality internally that is only supported by cancun for some other code, it will also not work. But I think currently that is not the case and only explicit use of assembly opcode will make it have a dependcy on cancun. Will reread the release notes to check. But it would indeed be something to be careful for.

dantaik:
@Brechtpd I think for Bridge and OwnerUUPSUpgradable.sol, maybe we can extract the code that use transient storage into a small solidity file respectively then have a non-transient version. Then we write a preprocessing script to swap these versions depends using a flag.

Brechtpd:
image

> Note that code compiled with this flag may not work as desired on any network where the hard fork is not yet implemented.

> It is important to note that at the moment the code generator does not automatically make use of the new opcodes to improve the emitted code and the optimizer support is minimal. Code generation and optimizer improvements will follow in subsequent releases.

So currently it should be fine to do it! But future versions of solidity it may be a problem.

dantaik:
OK, seems we have a solution but PR still needs to be updated to change the Bridge.

- we keep two implementation, one with transient storage and one without, using 'block.chainid == 1' as the condition.
- For L2 contracts, we use cancun to compile, but the transitent storage code will never run, so it should be fine.

Brecht, if you think this is a good solution, could you also modify the Bridge code?

Brechtpd:
> @Brechtpd I think for Bridge and OwnerUUPSUpgradable.sol, maybe we can extract the code that use transient storage into a small solidity file respectively then have a non-transient version. Then we write a preprocessing script to swap versions by renaming files, based on an env var.

How long do we think it will take to update our L2 to cancun? Is the plan to do it before mainnet or not? If we will do it for mainnet, then maybe this seems more work than necessary, and the conditional trick should be okay I believe. If not, than indeed some longer term solution needs to be found and then something like this seems like a reasonable approach. Or maybe just use inheritance and have different contracts for L1 and L2?

davidtaikocha:
> > I dont think #15810 will work, since the compiled bytecode will still contain the `TSTORE` opcode... right?
> 
> Yes but that is currently okay right I think? Because currently the code is not checked yet when it is deployed (maybe in the future it will), so if some random opcode/byte value is part of the bytecode but it is not executed, I think it should be fine.



> @Brechtpd I think for Bridge and OwnerUUPSUpgradable.sol, maybe we can extract the code that use transient storage into a small solidity file respectively then have a non-transient version. Then we write a preprocessing script to swap versions by renaming files, based on an env var.
> 
> @davidtaikocha do you think that will work?

Tested in go-ethereum, it works. (compiled the code by a cancun compile, then let it run inside a shanghai evm)

Brechtpd:
> Brecht, if you think this is a good solution, could you also modify the Bridge code?

Okay will do!

dantaik:
> > @Brechtpd I think for Bridge and OwnerUUPSUpgradable.sol, maybe we can extract the code that use transient storage into a small solidity file respectively then have a non-transient version. Then we write a preprocessing script to swap versions by renaming files, based on an env var.
> 
> How long do we think it will take to update our L2 to cancun? Is the plan to do it before mainnet or not? If we will do it for mainnet, then maybe this seems more work than necessary, and the conditional trick should be okay I believe. If not, than indeed some longer term solution needs to be found and then something like this seems like a reasonable approach. Or maybe just use inheritance and have different contracts for L1 and L2?

 I would say later, definitely not now.

Brechtpd:
> > > @Brechtpd I think for Bridge and OwnerUUPSUpgradable.sol, maybe we can extract the code that use transient storage into a small solidity file respectively then have a non-transient version. Then we write a preprocessing script to swap versions by renaming files, based on an env var.
> > 
> > 
> > How long do we think it will take to update our L2 to cancun? Is the plan to do it before mainnet or not? If we will do it for mainnet, then maybe this seems more work than necessary, and the conditional trick should be okay I believe. If not, than indeed some longer term solution needs to be found and then something like this seems like a reasonable approach. Or maybe just use inheritance and have different contracts for L1 and L2?
> 
> I would say later, definitely not now.

Okay, then conditional trick and just sticking to 0.8.24 compiler is actually fine as well. We can see with future compiler versions if there would be problems.

Brechtpd (approved):

PR #15859, opened 2024-02-16
feat(protocol)!: improve signal service and remove ICrossChainSync

dantaik:
The following changes are included in this PR after synchronizing with Dani and Jeff with recommendations from Brecht.

- ICrossChainSync is deleted and TaikoL1/L2 no longer offers getSyncedSnippet nor emits events
- SignalService now use the signal as its default value instead of constant value `1`. 
- SignalService use relayed chaindata (state root/signal root) as the signal value.
- SignalService now always track the largest blockID for each (chainId, kind) pair and offers a new function `getChainData` which relayers can use
- SignalService will emit the following event for each chain data relayed:

```solidity
    event ChainDataRelayed(
        uint64 indexed chainid,
        uint64 indexed blockId,
        bytes32 indexed kind,
        bytes32 data,
        bytes32 signal
    );
```
Bridge relayers can subscribe to this event to build a indexed db.

Screenshot 2024-02-16 at 12 30 14

## Summary by CodeRabbit

- **New Features**
    - Introduced a limit on unrelayed blocks for better synchronization control.
    - Added a new event `ChainDataRelayed` to signal successful relay of chain data.
    - Implemented new contract functions for more granular control over block synchronization and signal relaying.
- **Improvements**
    - Simplified contract inheritance and removed unnecessary interfaces for clearer contract structure.
    - Enhanced logic for block synchronization and relay, including adjustments for block ID usage over height.
    - Refined signal service functionality with additional parameters and updated error handling.
- **Bug Fixes**
    - Corrected error message handling within signal validation logic.
- **Refactor**
    - Consolidated test functions and removed deprecated tests related to removed functionalities.
    - Updated variable names and logic to align with new synchronization strategies.
- **Documentation**
    - Updated error message documentation to reflect changes in signal validation logic.


packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -237,24 +233,24 @@ library LibVerifying {
...
+                        b.unrelayedBlocks = uint8(unrelayedBlocks);
+                    } else {
+                        // Store the L2's state root as a signal to the local signal
+                        // service to allow for multi-hop bridging.
+                        ISignalService(resolver.resolve("signal_service", false)).relayChainData(
+                            config.chainId, b.lastVerifiedBlockId, LibSignals.STATE_ROOT, stateRoot
+                        );
+                        b.unrelayedBlocks = 0;
+                    }
+                }
```
Brechtpd, line 251: I think could be made a bit more efficient if we just query the latest synced state from the signal service
```
ISignalService signalService = ISignalService(resolver.resolve("signal_service", false));
uint64 lastSyncedBlockId = signalService.topBlockId(config.chainId, LibSignals.STATE_ROOT);
if (b.lastVerifiedBlockId > lastSyncedBlockId + config.maxUnrelayedBlocks) {
    signalService.relayChainData(config.chainId, b.lastVerifiedBlockId, LibSignals.STATE_ROOT, stateRoot);
}
```

Reading the data is more expensive (but also makes the call to `relayChainData` cheaper if it's needed because the signal service contract is already warm), but no need to update `b.unrelayedBlocks` which is even more expensive.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -64,6 +64,8 @@ library TaikoData {
         uint256 ethDepositGas;
         // The maximum fee allowed for an ETH deposit.
         uint256 ethDepositMaxFee;
+        // The max number of L2 blocks that can stay unsyned on L1
+        uint8 maxUnrelayedBlocks;
```
Brechtpd, line 68: Perhaps something like `relayBlockThreshold` is more clear, because `maxUnrelayedBlocks` is a bit confusing because the value could be 10 but there could actually be 15 unrelayed blocks when multiple blocks are verified.

packages/protocol/contracts/signal/ISignalService.sol
```diff
@@ -50,24 +69,39 @@ interface ISignalService {
...
-    /// @return True if the signal has been sent, otherwise false.
-    function isSignalSent(address app, bytes32 signal) external view returns (bool);
+    /// @notice Returns the latest relayed chain data.
+    /// @param blockId The chain data's corresponding blockId
+    /// @param kind A value to mark the data type.
+    /// @return blockId The block id.
+    /// @return chainData The relayed chain data.
+    function getLatestChainData(
+        uint64 chainId,
+        bytes32 kind
```
Brechtpd, line 102: Without being able to specify the `blockId` it will be trickier to use this function onchain because transactions may quickly get outdated because different data will be queried each time the top blockId changes.

Should be okay as is if onchain querying of data from other smart contracts is not an expected usecase of this function.

Brechtpd (approved):

dantaik:
@Brechtpd take another look please

Brechtpd (approved):

adaki2004 (approved):

PR #15864, opened 2024-02-16
fix(protocol): fix OZ bump side effects

adaki2004:
Changes due to OZ 4.9.5 version bump: 
1. in new version `GovernorUpgradeable` got also a public `cancel()` function so we need to override that one.
2. A possible small `supportsInterface` bug (in 4.9.5.OZ) codebase. Since it seems they removed the support for simple `IGovernorUpgradeable`, which is an abstract class, so it shall be default supported (See screenhosts below old version vs. new version) - now i'm adding the support in `TaikoGovernor.sol` (Sent the questions to OZ guys about it to confirm).
3. No more externally visible `admin()` function in proxy contract so query admin is only possible via storage slot

Support interface 4.8.3.:
Pasted Graphic 1

Supports interface 4.9.5.:
Pasted Graphic 2

### DIFF
--- a/packages/protocol/contracts/L1/gov/TkoGv.sol
+++ b/packages/protocol/contracts/L1/gov/TkoGv.sol
@@ -22,7 +22,7 @@ import
     "@openzeppelin/contracts-upgradeable/governance/extensions/GovernorTimelockControlUpgradeable.sol";
 import "../../common/OwnerUUPSUpgradable.sol";
 
-contract TaikoGovernor is
+contract TkoGv is
     OwnerUUPSUpgradable,
     GovernorCompatibilityBravoUpgradeable,
     GovernorVotesQuorumFractionUpgradeable,
@@ -38,8 +38,7 @@ contract TaikoGovernor is
         initializer
     {
         __OwnerUUPSUpgradable_init();
-        __Governor_init("TaikoGovernor");
-        __GovernorCompatibilityBravo_init();
+        __Governor_init("TkoGv");
         __GovernorVotes_init(_token);
         __GovernorVotesQuorumFraction_init(4);
         __GovernorTimelockControl_init(_timelock);
--- a/packages/protocol/contracts/L1/gov/TaikoGovernor.sol
+++ b/packages/protocol/contracts/L1/gov/TaikoGovernor.sol
@@ -22,7 +22,7 @@ import
     "@openzeppelin/contracts-upgradeable/governance/extensions/GovernorTimelockControlUpgradeable.sol";
 import "../../common/OwnerUUPSUpgradable.sol";
 
-contract TkoGv is
+contract TaikoGovernor is
     OwnerUUPSUpgradable,
     GovernorCompatibilityBravoUpgradeable,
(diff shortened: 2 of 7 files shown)


adaki2004:
@dantaik @Brechtpd 
Any suggestion what else to remove to get rid of this ?
`TaikoGovernor is above the contract size limit (24630 > 24576).`

It seems this OZ version bump had these side effects.

adaki2004:
> @dantaik @Brechtpd Any suggestion what else to remove to get rid of this ? `TaikoGovernor is above the contract size limit (24630 > 24576).`
> 
> It seems this OZ version bump had these side effects.

Okay, latest commit:
- renamings (unfrotunate but helpful since every character is 1 byte,
- removal of an `__init` which has no effect we are exactly fitting to the code size with no extra space :) 

But at least works.


kép

dionysuzx (approved):

packages/protocol/contracts/L1/gov/TkoGv.sol
```diff
@@ -14,21 +14,17 @@
...
     "@openzeppelin/contracts-upgradeable/governance/compatibility/GovernorCompatibilityBravoUpgradeable.sol";
-import "@openzeppelin/contracts-upgradeable/governance/extensions/GovernorVotesUpgradeable.sol";
 import
     "@openzeppelin/contracts-upgradeable/governance/extensions/GovernorVotesQuorumFractionUpgradeable.sol";
 import
     "@openzeppelin/contracts-upgradeable/governance/extensions/GovernorTimelockControlUpgradeable.sol";
 import "../../common/OwnerUUPSUpgradable.sol";
 
-contract TaikoGovernor is
+contract TkoGv is
```
dantaik, line 25: Hmm... not sure we should reduce the size by renaming the contract...

dantaik: How other people did it.

adaki2004: Yeah, but please help me out how can we, bc. we are now on the edge.

adaki2004: > How other people did it.

I see multiple folks running into these type of issues - if we aggregating to many voting schemes (Compound style bravo upgrade support, etc.)
e.g.:
https://forum.openzeppelin.com/t/if-i-use-the-optimizer-to-optimize-the-code-is-there-any-problem-in-using-all-the-functions-provided-by-openzeppelins-governor/30298
https://forum.openzeppelin.com/t/governor-contract-code-size-is-32070-bytes-and-cannot-be-deployed/36408/8

With optimizer disbaled, we can rename it back.. tho.. Will look into it a bit more.

adaki2004: Was able to rename it back to TaikoGovernor on the price of having "Taiko" in `__Governor_init` for the domain separator, i think that is a fair compromise (for now), thanks Brecht for the extra pair of eyes.

Brechtpd (approved):

PR #15898, opened 2024-02-18
fix(protocol): update amounts emitted to match length with tokenIds

n1punp:
The `TokenReceived` event should emit `amounts` of the same length as the `tokenIds`, even if it is unused for ERC721. 

I would also recommend enforcing `amount = 1` for each ERC721 send in `sendToken` and updating the `TokenReceived` event to have `amounts` just be an array of 1s. The upside is that it is clear that you are transferring the whole ERC721, instead of unused `amount` field. The downside is gas usage.


dantaik (approved):

n1punp:
@dantaik This will slightly increase the gas usage due to nonzero calldata bytes, but I think it is more correct to require amount = 1 for ERC721. What do you think?

davidtaikocha (approved):

PR #15910, opened 2024-02-18
fix(bridge-ui-v2): facilitate reset of recipient address back to sender address

shak58:
Hey :)

I just changed the bridge interface a bit. You know how sometimes you set a custom recipient address for your tokens and then want to switch back to your sender address? Before, you'd have to copy and paste your own sender address back in which is a bit of a hassle.

Now with this fix if you want to reset the recipient address to your own, just hit the little 'x' in the address field to clear it out, and then click 'Confirm'. It'll automatically set the recipient address back to the default, which is your sender address.

1. Click the 'x' to clear the custom address:
   Clear Custom Address
2. Then hit 'Confirm' to reset to your default address:
   Confirm Reset
 
 And now you recipient address gets reset!

### DIFF
--- a/packages/bridge-ui-v2/src/components/Bridge/SharedBridgeComponents/RecipientStep/Recipient.svelte
+++ b/packages/bridge-ui-v2/src/components/Bridge/SharedBridgeComponents/RecipientStep/Recipient.svelte
@@ -36,6 +36,14 @@
     addressInput.focus();
     addEscKeyListener();
   }
+  
+  function confirmModal() {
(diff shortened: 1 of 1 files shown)


KorbinianK:
@shak58 valid point regarding the "reset", but "confirming" an empty input to reset something is not the correct behaviour. Also cancelling a modal should not "confirm" it.

Please create an issue describing your feature and we might look into adding a "reset" functionality in the future, but this is just a workaround.

PR #15914, opened 2024-02-19
feat(protocol): revert revert update open-zeppelin contracts (#15896)

dionysuzx:
This reverts commit 994e29e67f68b2478c2e79ce28c9542fd048dc3c.


dantaik:
The new version doesn't work?

d1onys1us:
> The new version doesn't work?

yea i had to cancel the build. it's stuck in an infinite loop, see:
image

adaki2004:
> The new version doesn't work?
This PR aims to set back the "new" one.

So the problem was (in chronology):
- Dave suggested to use a newer one (4.9.5)
- We merged it (but has issues with DeployOnL1 workflow in CICD)
- We reverted it so taht it does not cause any issues on main
- We eventually need this so it is a 2nd attempt to add it back and debug first that `DeployOnL1` thing

dantaik:
@d1onys1us I think we are OK not to proceed with this upgrade.

d1onys1us:
> @d1onys1us I think we are OK not to proceed with this upgrade.

https://github.com/advisories/GHSA-93hq-5wgc-jc82

i think we do need to update so long as we are using `GovernorCompatibilityBravo`, but we can do `4.8.3`, `4.9.3`, or `5.0.1`, but we should patch the security issue IMO.

dantaik:
Suggest to close this PR given https://github.com/taikoxyz/taiko-mono/pull/15947

PR #15921, opened 2024-02-19
chore(relayer): add TargetBlock in indexer

xiaodino:


packages/relayer/processor/process_single.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func (p *Processor) processSingle(ctx context.Context) error {
-	slog.Info("processing tx", "estimateGas", common.Hash(*p.targetTxHash).Hex())
```
xiaodino, line 17: It's a typo introduced in the last PR.

packages/relayer/pkg/queue/rabbitmq/queue.go
```diff
@@ -163,8 +163,6 @@ func (r *RabbitMQ) Ack(ctx context.Context, msg queue.Message) error {
 
 	err := rmqMsg.Ack(false)
 
-	slog.Info("attempted acknowledge rabbitmq message")
-
```
xiaodino, line 167: To reduce the logs.

packages/relayer/indexer/config.go
```diff
@@ -40,6 +40,7 @@ type Config struct {
 	SyncMode                            SyncMode
 	WatchMode                           WatchMode
 	NumLatestBlocksToIgnoreWhenCrawling uint64
+	TargetBlock                         *uint64
```
mask-pp, line 43: Use uint64 and just check `== 0` can be more simplier.

xiaodino: It won't work . ```== 0``` will set ```TargetBlock``` to 0 and the indexer will only process block 0.

cyberhorsey:
why do we need this, when we have it in processor? it is only intended to be run locally or very sparingly to process a specific stuck message. do you have a use case for this?

dantaik (approved):

xiaodino:
> why do we need this, when we have it in processor? it is only intended to be run locally or very sparingly to process a specific stuck message. do you have a use case for this?

My use case is to index one specified block and debug the transactions.

packages/relayer/indexer/config.go
```diff
@@ -70,6 +71,14 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
 		SyncMode:                            SyncMode(c.String(flags.SyncMode.Name)),
 		ETHClientTimeout:                    c.Uint64(flags.ETHClientTimeout.Name),
 		NumLatestBlocksToIgnoreWhenCrawling: c.Uint64(flags.NumLatestBlocksToIgnoreWhenCrawling.Name),
+		TargetBlock: func() *uint64 {
+			if c.IsSet(flags.TargetBlock.Name) {
+				value := c.Uint64(flags.TargetBlock.Name)
+				return &value
+			} else {
```
cyberhorsey, line 78: we don't need this else, since the if statement returns. it can just be return nil.

actually this doesn't need to be a function at all. IMO, if you can just do this check above the struct?

packages/relayer/indexer/indexer.go
```diff
@@ -95,6 +95,8 @@ type Indexer struct {
 
 	numLatestBlocksToIgnoreWhenCrawling uint64
 
+	targetBlock *uint64
+
```
cyberhorsey, line 99: this should may be called "targetBlockNumber"

packages/relayer/indexer/indexer.go
```diff
@@ -387,7 +398,7 @@ func (i *Indexer) filter(ctx context.Context) error {
 
 	if i.processingBlockHeight < latestBlockIDToCompare {
 		slog.Info("header has advanced",
-			"processingBlockHeight", i.processingBlockHeight,
+			"startBlockID", i.processingBlockHeight,
 			"latestBlock", latestBlockIDToCompare,
```
cyberhorsey, line 402: this log looks incorrect? it should not be changed?

cyberhorsey (changes requested):

packages/relayer/indexer/indexer.go
xiaodino: Updated.

packages/relayer/indexer/config.go
xiaodino: Removed the else. 

It's to check if `TARGET_BLOCK_NUMBER` is set. If it's not, then `TargetBlockNumber` is set to nil. 

`c.Uint64(flags.TargetBlock.Name)` will return 0 if `TARGET_BLOCK_NUMBER` is not set, which doesn't work.

mask-pp (approved):

packages/relayer/indexer/config.go
cyberhorsey: you can use if c.IsSet above during the config creation though, and still set it to nil. this doesn't need to be a function. but if you want it to be a function then okay

cyberhorsey (approved):

PR #15923, opened 2024-02-19
fix(protocol): assembly Error fix

cryptoryda:
```javascript
Error:
Compiler run failed:
Error (4619): Function "mcopy" not found.
   --> contracts/automata-attestation/utils/BytesUtils.sol:273:13:
    |
273 |             mcopy(dest, src, len)
    |             ^^^^^
````
> We can solve this using :
```javascript
 function memcpy(uint256 dest, uint256 src, uint256 len) private pure {
        assembly {
            for {
                let i := 0
            } lt(i, len) {
                i := add(i, 32)
            } { 
                mstore(add(dest, i), mload(add(src, i)))
            }
        }
    }
 ```
**This loop copies 32 bytes at a time until the specified length is reached. Note that this implementation assumes that the length is a multiple of 32.**

### DIFF
--- a/packages/protocol/contracts/automata-attestation/utils/BytesUtils.sol
+++ b/packages/protocol/contracts/automata-attestation/utils/BytesUtils.sol
@@ -270,7 +270,13 @@ library BytesUtils {
 
     function memcpy(uint256 dest, uint256 src, uint256 len) private pure {
         assembly {
-            mcopy(dest, src, len)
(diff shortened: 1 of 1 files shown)


dantaik:
I don't have to error. Are you using the right compiler version?

cryptoryda:
Yes the right complier version is being used

PR #15928, opened 2024-02-19
chore(bridge-ui-v2): fix sample configuration file name for README instructions

shak58:
### Description

While setting up the local environment before by following the README instructions for copying the configuration examples, I encountered an issue with the `configuredCustomToken.example` file. The README commands assume the file is named `configuredCustomTokens.example`, but the actual file in the repository was named `configuredCustomToken.example` (without the plural 's'). This issue causes the command:

```sh
cp config/sample/configuredCustomTokens.example config/configuredCustomTokens.json
```

to fail with the error:

```
cp: cannot stat 'config/sample/configuredCustomTokens.example': No such file or directory
```

I just called instead `cp config/sample/configuredCustomToken.example config/configuredCustomTokens.json` but renaming the example file to include the plural 's' seems more appropriate than changing the README.

### Changes Proposed

- Rename the file `config/sample/configuredCustomToken.example` to `config/sample/configuredCustomTokens.example`


KorbinianK:
There are no changes. Pull the main branch again, this has been fixed already.

KorbinianK:
@shak58 you are actually correct, my bad! 

I've reopened a PR that came shortly before yours fixing the same problem. Thanks for double checking!

PR #15947, opened 2024-02-20
fix(protocol): fix governor bravo vulnerability

adaki2004:
So one (valid) argument to upgrade contracts to OZ 4.8.3 is the vulnerability described here:
https://github.com/taikoxyz/taiko-mono/security/dependabot/114

But moving to a newer version could be troublesome, as we saw (with breaking changes, not working deployment scripts) so here i'm cherry-picking the v4.8.3 patch.

This way we can still be (safely) compatible with compound.


adaki2004:
@dantaik Removing `GovernorCompatibilityBravoUpgradeable` not wise - as it implements some functions of `IGovernorUpgradeable` what `GovernorUpgradeable` does not implement.. So we would see more headache doing that.

packages/protocol/contracts/L1/gov/TaikoGovernor.sol
```diff
@@ -62,6 +62,30 @@ contract TaikoGovernor is
...
+        string[] memory signatures,
+        bytes[] memory calldatas,
+        string memory description
+    )
+        public
+        virtual
+        override(GovernorCompatibilityBravoUpgradeable)
+        returns (uint256)
+    {
+        require(signatures.length == calldatas.length, "GovernorBravo: invalid signatures length");
```
dantaik, line 83: To avoid potential contract-code-too-large issue and since this is now our code, lets use a custom error?

dantaik (approved):

dantaik (approved):

davidtaikocha (approved):

PR #15968, opened 2024-02-21
feat(protocol): make getInvocationDelays return non-zero values for base chains

dantaik:
We need to test this behavior in next testenet.

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -52,6 +52,7 @@ contract Bridge is EssentialContract, IBridge {
         0xe4ece82196de19aabe639620d7f716c433d1348f96ce727c9989a982dbadc2b9;
     // Place holder value when not using transient storage
     uint256 internal constant PLACEHOLDER = type(uint256).max;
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -463,19 +463,23 @@ contract Bridge is EssentialContract, IBridge {
...
+            block.chainid == 1 // Ethereum mainnet
+        ) {
+            // 384 seconds = 6.4 minutes = one ethereum epoch
+            return (6 hours, 384 seconds);
+        } else if (
+            block.chainid == 2 // Ropsten
+                || block.chainid == 4 // Rinkeby
+                || block.chainid == 5 // Goerli
+                || block.chainid == 42 // Kovan
+                || block.chainid == 11_155_111 // Sepolia
```
davidtaikocha, line 476: Shall we also add Holesky network id here?

dantaik: Done

davidtaikocha (approved):

adaki2004 (approved):

PR #15969, opened 2024-02-21
feat(protocol): extend SignalService interface 

adaki2004:
@KorbinianK @cyberhorsey 
You can query if we need to have ful proof or not if calling:

`getSyncedChainData`(`cahinId`, **`LibSignals.SIGNAL_ROOT`**, `blockId`)

If you supply 0 for blockId, it will return you the `topBlockId` where the **`LibSignals.SIGNAL_ROOT`** was cached and you dont need to have full proof (for that block height!).

If you supply non-0 for blockId, it will specifically search for that block id and revert if not cached.

### DIFF
--- a/packages/protocol/contracts/signal/ISignalService.sol
+++ b/packages/protocol/contracts/signal/ISignalService.sol
@@ -122,21 +122,4 @@ interface ISignalService {
         external
         pure
         returns (bytes32 signal);
-
-    /// @notice Returns true if a signal root is ached. If storage slot value is not bytes32(0),
-    /// than signal root is cached, so no need for account proof.
-    /// @dev If return values _cahced false and _blockId 0, it means there is no cached signal
-    /// service root with give chainId).
-    /// @param chainId Indenitifer of the chain for we are quering it's cached signal service root.
-    /// @param blockId The block id in question. If 0, indicates the query is targetting the latest
-    /// synced height.
-    /// @return _blockId blockId or highest synced block id.
-    /// @return _cached True if cached, false if not.
-    function isSignalRootCached(
-        uint64 chainId,
-        uint64 blockId
-    )
-        external
-        view
-        returns (uint64 _blockId, bool _cached);
 }
--- a/packages/protocol/contracts/signal/ISignalService.sol
+++ b/packages/protocol/contracts/signal/ISignalService.sol
@@ -93,6 +93,7 @@ interface ISignalService {
         returns (bool);
 
     /// @notice Returns the given block's  chain data.
+    /// @param chainId Indenitifer of the chainId.
     /// @param kind A value to mark the data type.
     /// @param blockId The chain data's corresponding block id. If this value is 0, use the top
     /// block id.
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -194,6 +194,24 @@ contract SignalService is EssentialContract, ISignalService {
         }
     }
 
+    /// @inheritdoc ISignalService
+    function isSignalRootCached(
```
dantaik, line 198: We already have `getSyncedChainData`.

d1onys1us:
🚨👮‍♂️
This is a backwards-compatible feature / minor bump, so the title can say "feat(protocol): *"

packages/protocol/contracts/signal/SignalService.sol
dantaik: "signal root", "state_root" are not something we expose through the signal service interface. They are values for the "kind" parameter.

packages/protocol/contracts/signal/ISignalService.sol
```diff
@@ -106,4 +107,36 @@ interface ISignalService {
...
+    /// @param blockId The chain data's corresponding block id. If this value is 0, use the top
+    /// block id.
+    /// @return signal The signal used for caching slot creation.
+    function signalForChainData(
+        uint64 chainId,
+        bytes32 kind,
+        uint64 blockId
+    )
+        external
+        view
```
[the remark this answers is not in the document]
dantaik: please also run `pnpm lint:sol` if you accept this suggested change.

packages/protocol/contracts/signal/SignalService.sol
adaki2004: @dantaik do you think we shall close this then ?

adaki2004: Removed redundant.

dantaik: you can keep this one `signalForChainData`

packages/protocol/contracts/signal/ISignalService.sol
```diff
@@ -106,4 +107,36 @@ interface ISignalService {
...
+    /// @notice Returns true if a signal root is ached. If storage slot value is not bytes32(0),
+    /// than signal root is cached, so no need for account proof.
+    /// @dev If return values _cahced false and _blockId 0, it means there is no cached signal
+    /// service root with give chainId).
+    /// @param chainId Indenitifer of the chain for we are quering it's cached signal service root.
+    /// @param blockId The block id in question. If 0, indicates the query is targetting the latest
+    /// synced height.
+    /// @return _blockId blockId or highest synced block id.
+    /// @return _cached True if cached, false if not.
+    function isSignalRootCached(
```
dantaik, line 135: but remove this function.

dantaik (approved):

davidtaikocha (approved):

PR #15979, opened 2024-02-22
chore(protocol): remove unnecessary reserved slots

dantaik:
`Transition` struct is never persisted on chain.


davidtaikocha (approved):

adaki2004 (approved):

Brechtpd:
I though this was added for onchain compatibility (so dependent contracts would still remain compatible even if extra values are added). I guess not.

PR #15980, opened 2024-02-22
fix(bridge-ui): fix wrong balance updates on network switch

shak58:
### Description:

There's a race condition in balance updates when switching networks. The `updateBalance` function is concurrently called by `onNetworkChange` and `onAccountChange` handlers leading to incorrect balance displays. This was because of the asynchronous execution of `updateBalance`, which when invoked by `onNetworkChange`, would update to the new network's balance. However, a simultaneous call from `onAccountChange` would still be running with data from the old network, sometimes resulting in the balance being updated with the outdated balance.
Due to being a race condition, the issue only happens some of the time when switching the network.

### Steps to Reproduce:

1. Navigate to the network switch interface where you can select between different blockchain networks.
2. Observe the current network's displayed balance.
3. Use the network dropdowns to switch to a different network.
4. Switch back and forth between the two networks multiple times.
5. Notice that occasionally the balance label does not update accordingly and remains displaying the balance from the previous network despite the network change being reflected in the UI.

#### Resolution:

To address this issue, I added a check within the `onAccountChange` handler. Now, it verifies whether `onNetworkChange` already triggeres a `updateBalance` call due to a network switch. Thus `updateBalance` is not called with stale data on network change.

#### Screenshot of the issue:

The screenshot bellow demonstrates the issue. It shows the user interface when executing the network switch, and how the balance was erroneously updated to 1 ETH instead of 10 ETH.

Balance Update Issue and Fix


KorbinianK (approved):
We will remove the onNetwork util, as with wagmi v2 it is all handled via account events. But good spot :)

dantaik (approved):

PR #15996, opened 2024-02-22
fix(protocol): mandate bridge message only calls onMessageInvocation

dantaik:
Screenshot 2024-02-22 at 17 47 10

`onMessageInvocation` is designed to be callable by the local bridge, guard each `onMessageInvocation` with `onlyFromNamed("bridge")` make it more explicit. Using `onlyFromNamed("bridge")` also prevents one `onMessageInvocation` from calling another `onMessageInvocation`, which is a security issue.


adaki2004 (approved):

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -528,6 +531,14 @@ contract Bridge is EssentialContract, IBridge {
 
         _storeContext({ msgHash: msgHash, from: message.from, srcChainId: message.srcChainId });
 
+        if (
+            message.data.length >= 4 // msg can be empty
+                && bytes4(message.data) != IMessageInvocable.onMessageInvocation.selector
+                && message.to.isContract()
```
Brechtpd, line 537: Very niche case, but you could think of some attack where `message.to` is not yet a contract, and then somebody does a bridge transaction and in the mean time `message.to` would be a contract and then any funds of the bridge transaction would be locked because this reverts. I think safer to just let this return false instead of throwing so it can be recalled.

packages/protocol/contracts/bridge/IBridge.sol
```diff
@@ -80,3 +80,9 @@ interface IRecallableSender {
         external
         payable;
 }
+
+/// @title IMessageInvocable
+/// @notice An interface that all bridge message receiver shall implement
+interface IMessageInvocable {
+    function onMessageInvocation(bytes calldata data) external payable;
+}
```
Brechtpd, line 88: The `CrossChainOwned` contract still needs to be updated to work with this new interface.

Brechtpd (commented):
The example case given seems pretty questionable because there is no reason why somebody would set approvals on the bridge contract or even trust it in any way more than any other contract (except of course contracts that explicitly interact with the bridge contract).

But I do see the benefit of trying to limit flexibility of the bridge to limit the amount of cases to think about and the number of ways things could be exploited, so I still think this is still a good change.

packages/protocol/contracts/bridge/IBridge.sol
dantaik: Good catch!

packages/protocol/contracts/L2/CrossChainOwned.sol
```diff
@@ -24,23 +24,24 @@ import "../bridge/IBridge.sol";
...
-    error XCO_INVALID_TX_ID();
     error XCO_INVALID_OWNER_CHAINID();
+    error XCO_INVALID_TX_ID();
     error XCO_PERMISSION_DENIED();
     error XCO_TX_REVERTED();
 
-    function executeCrossChainTransaction(uint64 txId, bytes calldata txdata) external {
-        if (txId != nextTxId) revert XCO_INVALID_TX_ID();
-
+    function onMessageInvocation(bytes calldata data) external payable nonReentrant whenNotPaused {
```
Brechtpd, line 39: I don't think this one can be `nonReentrant` because this contract needs to call into other `nonReentrant` functions.

Brechtpd (approved):

PR #16019, opened 2024-02-23
chore(protocol): remove unused third-party lib code

dantaik:


adaki2004:
Just thinking out loud, we brought this file from optimism 1:1, right ? Even the folder name is `thirdparty/optimism`, shall not we just leave the file untouched ? 🤔 even if we dont use some of the functions ?

Brechtpd (approved):
I think because this version of the protocol will be frozen without any more development except fixes, having the unused code there doesn't really matter, so removing unused stuff makes sense to me to make it easier for us and auditors to know what is important and in scope.

adaki2004 (approved):

adaki2004:
Why we seeing this in CICD ?
kép

PR #16020, opened 2024-02-23
chore(protocol): remove code related to PSE zkEVM

dantaik:


packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -210,7 +210,7 @@ contract TaikoL1 is EssentialContract, ITaikoL1, ITierProvider, TaikoEvents, Tai
     function getConfig() public view virtual override returns (TaikoData.Config memory) {
         // All hard-coded configurations:
         // - treasury: the actual TaikoL2 address.
-        // - blockMaxTxs: 80 (limited by the PSE zkEVM circuits)
+        // - blockMaxTxs: 80
```
davidtaikocha, line 213: Then shall we just remove `blockMaxTxs` + `blockMaxGasLimit` limits from our protocol?  or even allow a larger `blockMaxGasLimit` than 30m?

dantaik: blockMaxTxs was removed already. I shall delete this whole line

dantaik: Regarding blockMaxGasLimit I'm not sure, maybe we keep it for now as we'll use zkVM as well.

davidtaikocha (approved):

Brechtpd (approved):

packages/protocol/contracts/L1/TaikoL1.sol
Brechtpd: A higher `blockMaxGasLimit` I think would be very beneficial to reduce onchain costs, but if we do it I think we first have to test it out how the node/provers behave under big blocks.

PR #16023, opened 2024-02-23
feat(protocol): support for ERC1155 airdrop

arularmstrong:
This PR adds support for ERC1155 token airdrops, enhancing its versatility alongside existing ERC20 and ERC721 functionalities. This addition enhances the protocol, accommodating a wider range of token standards for airdrop distribution.


arularmstrong:
Yeah sure, I thought it would be a good extension to have it under protocol when we have an existing support for ERC721 & ERC20.

PR #16029, opened 2024-02-24
feat(protocol): use Ownable2StepUpgradeable for better security

dantaik:
This is to address the findings from Quill Audit:

## Description: 
Using single step ownership transfer is not secure, the existing owner can assign any new address as owner, and lose the ownership if the assigned address is incorrect address which can’t perform actions.
[Ownable2StepUpgradable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/access/Ownable2StepUpgradeable.sol) provides added safety where the pending owner can accept the ownership due to its two-step process to transfer the ownership.

## Recommendation(s): 
Consider using Ownable2StepUpgradable instead of OwnableUpgradable.


davidtaikocha (approved):

dantaik:
@davidtaikocha do you need to modify any scripts?

Brechtpd (approved):

PR #16054, opened 2024-02-24
fix(protocol): remove an unused event 

cryptoryda:
###  `event Debug(bytes32 x)`;  
-  Remove this line if not used
- Event is not emitted in anywhere of codebase. If you want to debug the code after times you can use it but now it takes gas-cost.

-  **compiler setting added to understand the solidity compiler directly.**


adaki2004:
Build process failing 🤔

dantaik (approved):

packages/protocol/contracts/automata-attestation/utils/SHA1.sol
```diff
@@ -6,7 +6,6 @@ pragma solidity 0.8.24;
 // https://github.com/ensdomains/solsha1/blob/master/contracts/SHA1.sol
 
 library SHA1 {
-    event Debug(bytes32 x);
```
dantaik, line 9: you need to delete this empty line

adaki2004 (approved):

PR #16063, opened 2024-02-25
feat(protocol): change cooldown and proving window to minutes

dantaik:
Previously the cooldown and proving window are in seconds, with uint16, these value can be 18.2 hours. This PR change these two values to use minutes as their unit.

Also added a mainnet tier provider file.


davidtaikocha (approved):

packages/protocol/contracts/L1/tiers/MainnetTierProvider.sol
```diff
@@ -0,0 +1,85 @@
...
+        tiers = new uint16[](3);
+        tiers[0] = LibTiers.TIER_SGX;
+        tiers[1] = LibTiers.TIER_SGX_ZKVM;
+        tiers[2] = LibTiers.TIER_GUARDIAN;
+    }
+
+    function getMinTier(uint256 rand) public pure override returns (uint16) {
+        // 0.1% require SGX + ZKVM; all others require SGX
+        if (rand % 1000 == 0) return LibTiers.TIER_SGX_ZKVM;
+        else return LibTiers.TIER_SGX;
```
Brechtpd, line 83: I would suggest removing randomness: https://github.com/taikoxyz/taiko-mono/pull/15568#issuecomment-1910538272. In practice I believe it will be avoided so in the end it doesn't really matter what the code says here, but it does have UX downsides and also makes proposing blocks more complicated because proposers will need to have some code to avoid this randomness (and if they don't, they either are unprofitable or simply can't compete with proposers that do take it into account to avoid it).

dantaik: as you wish.

packages/protocol/contracts/L1/tiers/MainnetTierProvider.sol
```diff
@@ -0,0 +1,85 @@
...
+import "../../common/EssentialContract.sol";
+import "./ITierProvider.sol";
+
+/// @title MainnetTierProvider
+/// @dev Labeled in AddressResolver as "tier_provider"
+/// @dev Assuming liveness bound is 250TKO.
+// Taiko token's total supply is 1 billion. Assuming block time is 2 second, and
+// the cool down period is 2 days. In 2 days, we can have (2*86400/2)=86400
+// blocks. Assuming 10% tokens are used in bonds, then each block may use up to
+// these many tokens: 1,000,000,000 * 10% / 86400=1157 TOK per block, which is
```
adaki2004, line 26: Not sure if USD amount is correct, or we shall mention valuation like this at all ? (Also true for `TestnetTierProvider.sol`)

packages/protocol/contracts/L1/tiers/MainnetTierProvider.sol
Brechtpd: hmmm even though now there is no randomness, there is still a fee spike every 1/1000 block which could still make things quite tricky I think because it still has the same UX and proposer implementation difficulties. Except now not because the more expensive proof can be avoided, but because the more expensive cannot be avoided. The proposer still needs to be profitable for each block it submits (in a based rollup it is not a valid approach to ever lose money when proposing blocks, because there is no benefit in doing so to be able to submit future blocks, so being profitable "on average" doesn't work). So either
- Every 1/1000 the user fees need to spike (with transactions that normally would be profitable in the mempool being stuck). This seems pretty unlikely to me because of how EIP-1559 fees work with a max tip, so these transactions would really have to be made for these specific blocks with higher prover fees otherwise the users would lose always overpay.
- The user fees don't spike (or at least not at much). The proposer now needs a multi-block approach for profitability and so also needs to submit blocks in a batch (multiple blocks in the same tx!). User fees can remain largely the same, it's just that the proposer will have to offset the extra costs for the block with higher prover costs with other following blocks that have the cheaper proof. The proposer needs to batch these up in a single tx because that is the only it can work for a based rollup, the proposer cannot simply propose these blocks in different txs.

So I think the problem still remains. Actually, it may have gotten more difficult because avoiding certain randomness seems easier than the problem above.

In any case, approving this PR because I think this problem will be very short lived, I'm currently optimistic that we'll have cheap zk proofs ready around mainnet!

Brechtpd (approved):

packages/protocol/contracts/L1/tiers/MainnetTierProvider.sol
dantaik: Then I prefer the previous random number based approach.

packages/protocol/contracts/L1/tiers/MainnetTierProvider.sol
dantaik: deleted

PR #16066, opened 2024-02-25
fix(bridge-ui): truncate selected token name to 5 characters

leppaludi:
* Large token names overflow the pill showing the selected token name

Before:

image

After:

image


KorbinianK:
@leppaludi Good job. Can you add it to the balance above as well?

leppaludi:
Updated to truncate balance and the dropdown name as well.

KorbinianK (approved):

xiaodino (approved):

PR #16069, opened 2024-02-25
feat(protocol): create SharedTaikoTokenVault.sol

dantaik:


adaki2004:
Seems pretty straightforward but i'll add couple of tests today.

adaki2004:
Did not we discuss last week this will be a multisig ? 🤔

dantaik:
> Did not we discuss last week this will be a multisig ? 🤔

Ah, you are right.

adaki2004:
> Seems pretty straightforward but i'll add couple of tests today.

@dantaik modified the tests to use this contract.

PR #16070, opened 2024-02-25
fix(bridge-ui): enable continue button for scanned nft imports

### LINKED ISSUE #16064: Bug when trying to bridge nft from Katla to Holesky
### Describe the bug

When i click 'Scan for NFT' i got some js code appear on the front end
And then 'Continue' button is not active
Clear all cache and cookie doesn't help
```
0
validBalance false
$enteredAmount gt 0 false
has amount false
(nftHasAmount ? $enteredAmount gt 0 : true) true
false
```
image
image

### Steps to reproduce

Steps to reproduce here.

### Spam policy

- [x] I verify that this issue is NOT SPAM and understand SPAM issues will be closed and reported to GitHub, resulting in ACCOUNT TERMINATION.

leppaludi:
Fixes https://github.com/taikoxyz/taiko-mono/issues/16064

* Continue button is disabled when trying to bridge scanned NFTs

Before

image

After

image

### DIFF
--- a/packages/bridge-ui/src/components/Bridge/NFTBridgeComponents/ImportStep/ScannedImport.svelte
+++ b/packages/bridge-ui/src/components/Bridge/NFTBridgeComponents/ImportStep/ScannedImport.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-  import { onDestroy } from 'svelte';
   import { t } from 'svelte-i18n';
 
   import { enteredAmount, selectedNFTs } from '$components/Bridge/state';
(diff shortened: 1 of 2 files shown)


packages/bridge-ui/src/components/Bridge/NFTBridgeComponents/ImportStep/ScannedImport.svelte
```diff
@@ -61,10 +60,6 @@
   } else {
     canProceed = false;
   }
-
-  onDestroy(() => {
-    $selectedNFTs = [];
-  });
```
leppaludi, line 67: This component is only used in `packages/bridge-ui/src/components/Bridge/NFTBridgeComponents/ImportStep/ImportStep.svelte` and that component sets `$selectedNFTs = [];` when it is mounted.

KorbinianK:
That is unfortunately not the right way. There are a couple of other things I spotted, will create a separate PR

PR #16071, opened 2024-02-25
fix(protocol): set initial owner in the init() function without `acceptOwnership`

dantaik:
The PR looks extensive, but it's mostly grammatical changes.

Now, all contracts are Ownable2StepUpgradeable for better security. But when they are initialized (using `init(...)`), the initial owner doesn't have to call `acceptOwnership` to become the current owner (all new owners must do that though). These changes will make David C's deployment much easier.


packages/protocol/contracts/common/OwnerUUPSUpgradable.sol
```diff
@@ -84,7 +84,8 @@ abstract contract OwnerUUPSUpgradable is UUPSUpgradeable, OwnableUpgradeable {
     /// @notice Initializes the contract with an address manager.
     // solhint-disable-next-line func-name-mixedcase
     function __OwnerUUPSUpgradable_init() internal virtual onlyInitializing {
-        __Ownable_init();
+        // Sets owner = msg.sender directly without the pending-accept process.
+        _transferOwnership(msg.sender);
```
adaki2004, line 88: I think `OwnerUUPSUpgradeable.sol `is OZ contract, which we now modifying ? 
Here, there shall be no pending-accept process (imo) because this is inheriting from `OwnableUpgradeable` not `Ownable2StepUpgradeable`, no ? 🤔  i think we should leave this contract untouched.

dantaik: > I think OwnerUUPSUpgradeable.sol is OZ contract, which we now modifying ?
OwnerUUPSUpgradeable.sol is not part of OZ lib.

dantaik: Now we have 

```solidity
  function __Essential_init(
        address _owner,
        address _addressManager
    )
        internal
        virtual
        onlyInitializing
    {
        _transferOwnership(_owner == address(0) ? msg.sender : _owner);
        __AddressResolver_init(_addressManager);
        _paused = _FALSE;
    }
``` 

and prevously we have:

```  function __Essential_init(
        address _owner,
        address _addressManager
    )
        internal
        virtual
        onlyInitializing
    {
        _transferOwnership(msg.sender); // = __Ownable_init_unchained
        __AddressResolver_init(_addressManager);
        _paused = _FALSE;
    }
```

We need to give the contract a chance to set the initial owner to an address other than `msg.sender` to avoid many calls to `acceptOwnership`

davidtaikocha (approved):

adaki2004 (approved):

PR #16096, opened 2024-02-27
ci(repo): skip vercel previews on dependabot PRs

dionysuzx:
fixes those dependabot failures.

### DIFF
--- a/.github/workflows/bridge-ui-preview.yml
+++ b/.github/workflows/bridge-ui-preview.yml
@@ -13,6 +13,7 @@ on:
 jobs:
   Deploy-Preview:
     runs-on: [taiko-runner]
+    if: github.actor != 'dependabot[bot]'
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


dantaik (approved):

.github/workflows/bridge-ui-preview.yml
```diff
@@ -13,6 +13,7 @@ on:
 jobs:
   Deploy-Preview:
     runs-on: [taiko-runner]
+    if: github.actor != 'dependabot[bot]'
```
dantaik, line 16: Dave, won't incompatible dependency upgrades break the build?

KorbinianK (approved):

.github/workflows/bridge-ui-preview.yml
d1onys1us: i didn't know i can have secrets for dependabot in github settings. i added them now, so this job should not fail anymore. i agree it's nice to build a preview as part of an "e2e test" after dependency upgrades.

PR #16153, opened 2024-02-28
chore(protocol): oz - remove unused enums

dantaik:


packages/protocol/contracts/automata-attestation/lib/EnclaveIdStruct.sol
```diff
@@ -22,7 +22,7 @@ library EnclaveIdStruct {
     }
 
     enum EnclaveIdStatus {
-        OK,
+        OK, // placeholder for value 0
```
dantaik, line 25: @smtmfft SGX_ENCLAVE_REPORT_ISVSVN_REVOKED is not used in the code to be assigned to a variable as its value. Do we expect the user to supply the value?

PR #16157, opened 2024-02-28
chore(protocol): improve imports

dantaik:


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -474,7 +474,7 @@ contract Bridge is EssentialContract, IBridge {
         ) {
             // For Taiko mainnet
             // 384 seconds = 6.4 minutes = one ethereum epoch
-            return (6 hours, 384 seconds);
+            return (1 hours, 384 seconds);
```
dantaik, line 477: 1 hour after yesterday's discussion

davidtaikocha (approved):

packages/protocol/contracts/bridge/Bridge.sol
arcxteam: so long to return, can this improve codes

adaki2004 (approved):

packages/protocol/contracts/bridge/Bridge.sol
dantaik: Improve security.

PR #16174, opened 2024-02-28
Fix(protocol): use of keccak256 abi encode0 instead of blobhash

cryptoryda:
### The function blobhash(0) is not defined or imported in the provided code. use keccak256(abi.encode(0)) to generate a hash based on the number 0.


KorbinianK:
@cryptoryda maybe a quick Google search would have cleared things up ;)

PR #16198, opened 2024-02-29
chore(fork-diff): update fork-diff

RogerLamTd:
update forkdiff to geth 1.13.14


dantaik:
@davidtaikocha do you think it's a good time to update the diff now? If there are still considerable amount of work on the client, then I'd suggest postponing the diff.

davidtaikocha:
> @davidtaikocha do you think it's a good time to update the diff now? If there are still considerable amount of work on the client, then I'd suggest postponing the diff.

it's geth diff only, but since there is still a new geth repo feature (saving the state trie of the latest verified block) to be merged in a few days, I would suggest let's postpone it for a while.

RogerLamTd:
will update again after state trie feature is merged

PR #16230, opened 2024-03-01
feat(bridge-ui): update to protocol 1.0, two step bridging

KorbinianK:
- Supports protocol 1.0 changes
- Adds two step claim dialogs for NFT, ERC20 etc.
- Improve code coverage
- Small bugfixes


adaki2004 (approved):
Pretty huge PR - would just rather accept it, then after deployment check via QA if we have any issues.

KorbinianK:
@adaki2004 yes, could have split it up to be honest

dionysuzx (approved):

PR #16255, opened 2024-03-02
fix(protocol): Require amount=1 when transferring ERC721 instead of 0

n1punp:
This change is in response with #15898. Transferring ERC721 should require the amount to be 1 instead of 0, so the transfer will not be ambiguous for offchain tracking. While it may cost slightly more gas to require amount=1 instead of 0, but I still think it makes sense to do so.


adaki2004:
Thank you but we are all set for now and dont want to change any interface (UI), just in case this is absolutely necessary ror there is a bug.

PR #16274, opened 2024-03-03
fix(eventindexer): fix chartByTask query bindings and cast to date for start & end date in chart

arif-x:
#16276


bigrayfx:
yes can't do it

cyberhorsey:
sorry, what do we need this for?
Also this PR doesn't update bindings but it says it does in the title.

arif-x:
> sorry, what do we need this for? is the chart API not working as expected? Also this PR doesn't update bindings but it says it does in the title.
I just knowing when I call chartByTask endpoint, I try to check this query and binding. Because swagger not have start and end params, so this query cant run

arif-x:
> sorry, what do we need this for? is the chart API not working as expected? Also this PR doesn't update bindings but it says it does in the title.

I just knowing when I call chartByTask endpoint, I try to check this query and binding. Because swagger not have start and end params, so this query cant run.
I try to update that query with if start and end param is not set condition, and this query can fecth data. And because date column is using varchar, I try to parse date column into date datatype in select statement.

arif-x:
Sorry my english is bad

packages/eventindexer/docs/swagger.json
```diff
@@ -1,311 +1,359 @@
...
-    },
-    "eventindexer.Stat": {
-      "type": "object",
-      "properties": {
-        "averageProofReward": {
-          "type": "string"
-        },
-        "averageProofTime": {
-          "type": "string"
+    "swagger": "2.0",
```
xiaodino, line 2: Is this change not related?

packages/eventindexer/docs/docs.go
```diff
@@ -52,7 +52,7 @@ const docTemplate = `{
                 }
             }
         },
-        "/chartByTask": {
+        "/chart/chartByTask": {
```
xiaodino, line 55: Do we also need to change this https://github.com/taikoxyz/taiko-mono/blob/main/packages/eventindexer/pkg/http/routes.go#L25 ?

xiaodino (changes requested):
What is the motivation for this PR?

packages/eventindexer/pkg/repo/chart.go
```diff
@@ -38,43 +40,80 @@ func (r *ChartRepository) Find(
...
+	}
 
-	tx = r.getDB().Raw(q, task, start, end)
+	var q string
+	var args []interface{}
+	if feeTokenAddress == "" && tier == "" {
+		if start != "" && end != "" {
+			q = `SELECT * FROM time_series_data
+			WHERE task = ? AND STR_TO_DATE(date, '%Y-%m-%d') BETWEEN ? AND ?
+			ORDER BY STR_TO_DATE(date, '%Y-%m-%d')`
```
xiaodino, line 66: Does it mean the query will only query by date?

packages/eventindexer/pkg/repo/chart.go
```diff
@@ -38,43 +40,80 @@ func (r *ChartRepository) Find(
...
+		}
+	}
 
-	tx = r.getDB().Raw(q, task, start, end)
+	var q string
+	var args []interface{}
+	if feeTokenAddress == "" && tier == "" {
+		if start != "" && end != "" {
+			q = `SELECT * FROM time_series_data
+			WHERE task = ? AND STR_TO_DATE(date, '%Y-%m-%d') BETWEEN ? AND ?
```
xiaodino, line 65: Does it mean the query will only filter data by date?

packages/eventindexer/pkg/repo/chart.go
xiaodino, line 66: Why use ```ORDER BY STR_TO_DATE(date, '%Y-%m-%d')```, rather than ```ORDER BY date```?

cyberhorsey:
This chart code does not need to be changed, and runs fine in production already. All we need to do is update the documentation, as yes, the route is `/chart/chartByTask`, which is outdated in swagger.

However, this is only used for our status page.

packages/eventindexer/pkg/repo/chart.go
arif-x: because in native mysql need to be casted at date data type

packages/eventindexer/pkg/repo/chart.go
arif-x: I add option if start & end date is filled or not

arif-x: I add option if start & end date is filled or not

packages/eventindexer/docs/docs.go
arif-x: I change this because on https://eventindexer-swagger.taiko.xyz/#/ has not found result

PR #16290, opened 2024-03-03
update merge_contracts

Emir236070:
I observed that the file path required for the code to run is not complete and therefore it cannot run.


Emir236070:
fix:

PR #16300, opened 2024-03-04
chore(deps-dev): upgrade scripts and docker version and solc version

mask-pp:


davidtaikocha:
Looks like the workflow is not passed @mask-pp

mask-pp:
> Looks like the workflow is not passed @mask-pp

Ok, I will fix it.

dantaik (approved):

davidtaikocha (approved):

PR #16321, opened 2024-03-05
fix(protocol)!: fix custom coinbase `transferFrom` issue

dantaik:
In AssignmentHook, we have 

```solidity
IERC20(assignment.feeToken).safeTransferFrom(_meta.coinbase,_blk.assignedProver,proverFee)
```

If `meta.coinbase` can be set to any value, then the current block proposer can maliciously consume fee token of another address that have set up the allowance. See  https://github.com/crytic/slither/wiki/Detector-Documentation#arbitrary-from-in-transferfrom

We can fix this issue by not allowing the `coinbase` value to be set and always using `msg.sender`; or we can request a signature from `params.coinbase` unless it is address(0). This PR choose the first approach.


davidtaikocha (approved):

adaki2004:
I tried to follow the logic, but (maybe i dont see something)i think it is not possible to set an arbitrary `meta.coinbase` by the proposer.

Becase `ProverAssignment` is part of the signature. `metaHash` is part of `ProverAssignment`. So it has to be signed (off-chain) beforehand. In case the proposer injects an arbitrary address, the signature will not match.

dantaik:
Suppose I'm a proposer, I can set my block's `params.coinbase` to your address, can't I? Then your token will be used by me.

adaki2004:
> Suppose I'm a proposer, I can set my block's `params.coinbase` to your address, can't I? Then your token will be used by me.

Yes,that variable you can, but the PR description says:
`If meta.coinbase can be set to any value,`

And you also mention this line:
`IERC20(assignment.feeToken).safeTransferFrom(_meta.coinbase,_blk.assignedProver,proverFee)`

This uses `_meta.coinbase` - not `params.coinbase`.

Maybe i still dont see the issue here.

Another issue i see with `msg.sender` is that what is a malicious proposer scrapes the mempool and frontruns ?

PR #16327, opened 2024-03-05
fix(protocol): fix custom coinbase `transferFrom` issue

dantaik:
In AssignmentHook, we have 

```solidity
IERC20(assignment.feeToken).safeTransferFrom(_meta.coinbase,_blk.assignedProver,proverFee)
```

If `meta.coinbase` can be set to any value, then the current block proposer can maliciously consume fee token of another address that have set up the allowance. See  https://github.com/crytic/slither/wiki/Detector-Documentation#arbitrary-from-in-transferfrom


dantaik:
This may be a better fix compared with https://github.com/taikoxyz/taiko-mono/pull/16321

Brechtpd (approved):

adaki2004 (approved):

PR #16330, opened 2024-03-05
feat(relayer): document indexer with comments, fix broken crawler logic, add new metrics

cyberhorsey:
The crawler does not work because of a bug introduced in `indexer.go`:
`endBlockID = i.numLatestBlocksToIgnoreWhenCrawling`

This PR fixes it, plus it adds new metrics, and documents the indexer. It also adds an actual test for the encoding of the hop proofs.


packages/relayer/indexer/handle_no_events_in_batch.go
```diff
@@ -10,8 +10,8 @@ import (
 	"github.com/taikoxyz/taiko-mono/packages/relayer"
 )
 
-// handleNoEventsInBatch is used when an entire batch call has no events in the entire response,
-// and we need to update the latest block processed
+// handleNoEventsInBatch is used when there are no events remaining ian a batch, or
```
RogerLamTd, line 13:
```suggestion
// handleNoEventsInBatch is used when there are no events remaining in a batch, or
```

packages/relayer/indexer/indexer.go
```diff
@@ -275,31 +308,49 @@ func (i *Indexer) filter(ctx context.Context) error {
...
 		return i.subscribe(ctx, i.srcChainId, i.destChainId)
 	}
 
+	// the end block is the latest header.
 	endBlockID := header.Number.Uint64()
 
-	// ignore latest N blocks, they are probably in queue already
-	// and are not "missed".
+	// ignore latest N blocks if we are crawling past blocks, they are probably in queue already
+	// and are not "missed", have justd not been processed.
```
RogerLamTd, line 340:
```suggestion
	// and are not "missed", have just not been processed.
```

RogerLamTd (approved):

dantaik (approved):

PR #16345, opened 2024-03-06
feat(relayer):  rmq: add dead letter exchange, add retry capability

cyberhorsey:
This also removes the fake "onlyOwner" status from events, and adds Recalled.

This PR adds retry capability to the rabbitmq messages, ensuring if we have an issue processing,it will be negatively acknowledged, then bound to a dead letter queue in rabbitMQ. It will then be re-routed back to the original queue, and a retry will be incremented.

We also add a large amount of  additional metrics to be better able to monitor our queues, and set alerts.


RogerLamTd (approved):

dantaik:
@cyberhorsey @RogerLamTd there are lint issues.

davidtaikocha (approved):

PR #16346, opened 2024-03-06
fix(bridge-ui): preserve custom processing fee selection across components

CodeDragonVN:
## Description
This pull request solves an issue where the processing fee selection is not preserved when navigating between components in the bridge. Specifically, when a user selected 'None' as the processing fee, it would revert to the recommended instead of showing '0 ETH | Customized' after returning to the Import step, unlike when a custom fee was selected and remain at the Import Step.
This happened because `hasEnoughEth` would be set the it's false declaration in ProcessingFee.svelte, since the `ProcessingFee` component created in TokenInput.svelte wasn't binding the `hasEnoughEth` prop, which would then trigger `unselectNoneIfNotEnoughETH($processingFeeMethod, hasEnoughEth)` and set the `processingFeeMethod` to recommended.

## Changes
- Ensured the `hasEnoughEth` prop is correctly bound and passed through to the `ProcessingFee` component within [TokenInput.svelte](https://github.com/taikoxyz/taiko-mono/blob/main/packages/bridge-ui/src/components/Bridge/FungibleBridgeComponents/ImportStep/TokenInput/TokenInput.svelte).

## How to Reproduce the Issue
1. Navigate to the 'Review' step of the bridge.
2. Select 'None' for the processing fee.
3. Return to the 'Import' step.
4. Observe that previously, the fee selection would reset to the recommended, but with this fix, it should now correctly show zero.

## Screenshots

| Description | Screenshot |
| --- | --- |
| The issue where the processing fee reverted to the recommended. | Broken Fee Selection |
| The custom fee selection being preserved as expected. | Custom Fee Selection |
|  The fixed state showing the '0 ETH \| Customized' fee preserved correctly. | Fixed Zero Fee |
|  Console warning indicating the missing `hasEnoughEth` prop. | Console Warning |


KorbinianK:
@CodeDragonVN 
Valid point, it's indeed missing the binding to reset correctly.
Feel free to fix the issue. It should reset fully when going back to the start.
You can keep this PR open I suppose and just add the changes, but up to you.

KorbinianK (approved):

davidtaikocha (approved):

PR #16353, opened 2024-03-07
chore(repo): add blob-catcher to monorepo

dionysuzx:
## summary
- add blob-catcher to monorepo
- add steps to adding a repo to monorepo to CONTRIBUTING.md


packages/blob-storage/internal/logic/config.go
```diff
@@ -0,0 +1,94 @@
...
+	MongoDB  MongoDBConfig
+	Server   ServerConfig
+}
+
+// GetConfig loads the configuration from environment variables or a config file.
+func GetConfig() (*Config, error) {
+	cfg := &Config{}
+	cfg.Networks = []NetworkConfig{
+		{
+			RPCURL:      "wss://l1ws.internal.taiko.xyz",
```
mask-pp, line 52: Do these configuration items don't need to be changed?

packages/blob-storage/internal/logic/config.go
```diff
@@ -0,0 +1,94 @@
...
+			BeaconURL:   "https://l1beacon.internal.taiko.xyz/eth/v1/beacon/blob_sidecars/", // Set the beacon URL here
+			NetworkName: "L2A",
+			IndexedEvents: []IndexedEvent{
+				{
+					Contract:  "0xC069c3d2a9f2479F559AD34485698ad5199C555f",
```
mask-pp, line 57: ditto

packages/blob-storage/internal/logic/config.go
```diff
@@ -0,0 +1,94 @@
...
+					EventHash: "0xa62cea5af360b010ef0d23472a2a7493b54175fd9fd2f9c2aa2bb427d2f4d3ca",
+					EventName: "BlockProposed",
+					Callback:  BlockProposedCallback,
+				},
+			},
+		},
+		{
+			RPCURL:      "wss://l1ws.internal.taiko.xyz",
+			BeaconURL:   "https://l1beacon.internal.taiko.xyz/eth/v1/beacon/blob_sidecars/", // Set the beacon URL here
```
mask-pp, line 66: ditto

packages/blob-storage/internal/logic/config.go
```diff
@@ -0,0 +1,94 @@
...
+					Callback:  BlockProposedCallback,
+				},
+			},
+		},
+	}
+
+	// Same DB for the blobs (L2A, L2B, etc.)
+	cfg.MongoDB = MongoDBConfig{
+		Host:     "localhost",
+		Port:     27017,
```
mask-pp, line 82: ditto

packages/blob-storage/internal/indexer/indexer_factory.go
```diff
@@ -0,0 +1,24 @@
...
+func InitFromConfig(cfg *logic.Config, pastEvents bool, startBlockNumber *big.Int) error {
+	// Initialize listeners for each network
+	for _, network := range cfg.Networks {
+		for _, event := range network.IndexedEvents {
+			indexer := NewIndexer(network.RPCURL, network.BeaconURL, network.NetworkName, event.Contract, pastEvents, startBlockNumber)
+
+			indexer.SubscribeEvent(event.EventHash, event.Callback)
+
+			// Run the event listener for each network
+			go indexer.Start()
```
mask-pp, line 19: Please handle the goroutine.

dantaik:
@adaki2004 We need README.md to start with...

packages/blob-storage/internal/indexer/indexer.go
```diff
@@ -0,0 +1,95 @@
...
+
+	if pastEvents == false {
+		header, err := client.HeaderByNumber(context.Background(), nil)
+		if err != nil {
+			log.Fatal(err)
+		}
+		startBlockNumber = header.Number
+	}
+
+	fmt.Println(client.Client())
```
mask-pp, line 41:
```suggestion
```

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,185 @@
...
+		KzgCommitmentHex []byte `json:"-"`
+	} `json:"data"`
+}
+
+// Callback functions
+
+// BlockProposedCallback is a callback for the "BlockProposed" event.
+func BlockProposedCallback(rpcURL, beaconURL, networkName string, log types.Log) {
+
+	contractAbi, err := abi.JSON(strings.NewReader(taikol1.TaikoL1ABI))
```
mask-pp, line 42:
```suggestion
	contractAbi, err := taikol1.TaikoL1MetaData.GetAbi()
```

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	// url := fmt.Sprintf("https://l1beacon.internal.taiko.xyz/eth/v1/beacon/blob_sidecars/%s", blockID)
+	url := fmt.Sprintf("%s/%s", beaconURL, blockID)
+	response, err := http.Get(url)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	body, err := ioutil.ReadAll(response.Body)
```
mask-pp, line 105:
```suggestion
	body, err := io.ReadAll(response.Body)
```

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,185 @@
...
+			return err
+		}
+
+		// Comparing the hex strings of meta.blobHash (blobHash)
+		if calculateBlobHash(data.KzgCommitment) == blobHashInMeta {
+			n := new(big.Int)
+
+			blockNrBig, ok := n.SetString(blockID, 10)
+			if !ok {
+				fmt.Println("SetString: error")
```
mask-pp, line 128: Please use `log.Println`.

packages/blob-storage/internal/logic/database.go
```diff
@@ -0,0 +1,51 @@
...
+		return nil, err
+	}
+
+	// Check the connection
+	err = client.Ping(context.Background(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Println("Connected to MongoDB")
```
mask-pp, line 40: ditto

packages/blob-storage/cmd/blob_catcher/main.go
```diff
@@ -0,0 +1,41 @@
...
+	"math/big"
+	"strconv"
+
+	"github.com/taikoxyz/taiko-mono/packages/blob-storage/internal/indexer"
+	"github.com/taikoxyz/taiko-mono/packages/blob-storage/internal/logic"
+)
+
+func main() {
+	// Define command-line flags
+	pastEvents := flag.Bool("past_events", false, "Enable indexing past events")
```
mask-pp, line 15: Recommended to use `github.com/urfave/cli/v2` to be consistent with other binaries.

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,185 @@
...
+	// Some debug logs for now.
+	// fmt.Println("Block id is:")
+	// fmt.Println(log.Topics[1])
+
+	if eventData.Meta.BlobUsed {
+		// in LibPropose we assign block.height-1 to l1Height, which is the parent block.
+		l1BlobHeight := eventData.Meta.L1Height + 1
+		blobHash := hex.EncodeToString(eventData.Meta.BlobHash[:])
+
+		storeBlob(rpcURL, beaconURL, networkName, strconv.Itoa(int(l1BlobHeight)), blobHash)
```
mask-pp, line 63: Handle error.

mask-pp (commented):
@d1onys1us Need more test cases.

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	var responseData Response
+	if err := json.Unmarshal(body, &responseData); err != nil {
+		return err
+	}
+
+	for _, data := range responseData.Data {
+
+		data.KzgCommitmentHex, err = hex.DecodeString(data.KzgCommitment[2:])
```
mask-pp, line 117:
```suggestion
		data.KzgCommitmentHex, err = hex.DecodeString(data.KzgCommitment[2:])
```

adaki2004:
> @adaki2004 We need README.md to start with...

Isn't it good, or what kind of readme.md you want to see ?
https://github.com/taikoxyz/taiko-mono/blob/4bde8d59e01b7a99eda6c26c850e6d34f5d58fde/packages/blob-storage/README.md

packages/blob-storage/internal/logic/config.go
adaki2004: I think it can change but rare occasions, which would just mean to restart the application (if changed - if ever).

packages/blob-storage/internal/logic/config.go
adaki2004: ditto my answer above.

packages/blob-storage/internal/logic/config.go
adaki2004: ditto my answer above.

packages/blob-storage/internal/logic/config.go
adaki2004: ditto my answer above. + as said in the readme:
_"- small refinements and DevOps (prod-grade DB with creditentials, proper containerization)"_ -> if you can help me with that, i'd appreciate.

mask-pp: So, do we have to specify that Mongodb port to `27017 ` when deploying MongoDB?

packages/blob-storage/internal/logic/config.go
mask-pp: I don't think this is a wise move. Once the URL changed you have to modify the code.

cyberhorsey:
I was requested to review this but don't know what I'm reviewing. @d1onys1us any context? It prints blobs out from blocks? who wrote it originally

packages/blob-storage/internal/logic/callbacks.go
cyberhorsey: specifically either use the now go standard "slog" package or the go-ethereum "log" package which implements structured logging

packages/blob-storage/internal/indexer/indexer.go
```diff
@@ -0,0 +1,94 @@
...
+	beaconURL         string
+	networkName       string
+	contractAddress   common.Address
+	eventHash         common.Hash
+	startHeight       *big.Int
+	eventCallbackFunc func(string, string, string, types.Log)
+}
+
+// NewIndexer creates a new Indexer instance.
+func NewIndexer(rpcURL, beaconURL, networkName, contractAddress string, pastEvents bool, startBlockNumber *big.Int) *Indexer {
```
cyberhorsey, line 26: should be a Opts struct instead of so many separate initialize variables

packages/blob-storage/internal/indexer/indexer.go
```diff
@@ -0,0 +1,94 @@
...
+	client, err := ethclient.Dial(rpcURL)
+	if err != nil {
+		log.Fatal("Failed to connect to the Ethereum client:", err)
+	}
+
+	if pastEvents == false {
```
cyberhorsey, line 32: if !pastEvents

packages/blob-storage/internal/indexer/indexer.go
```diff
@@ -0,0 +1,94 @@
...
+		header, err := client.HeaderByNumber(context.Background(), nil)
+		if err != nil {
+			log.Fatal(err)
+		}
+		startBlockNumber = header.Number
+	}
+
+	slog.Info(client.Client())
+
```
cyberhorsey, line 41: don't log out entire structs, also, slog package format isn't like this. it's like

slog.Info("logging client", "client", client.Client())

the first parameter is a context string, then a pair of values with a string identifier and a valur

packages/blob-storage/internal/indexer/indexer.go
```diff
@@ -0,0 +1,94 @@
...
+		rpcURL:          rpcURL,
+		beaconURL:       beaconURL,
+		networkName:     networkName,
+		contractAddress: common.HexToAddress(contractAddress),
+		startHeight:     startBlockNumber,
+	}
+}
+
+// SubscribeEvent subscribes to the specified event.
+func (indexer *Indexer) SubscribeEvent(eventHash string, callbackFunc func(string, string, string, types.Log)) {
```
cyberhorsey, line 52: callbackFunc should be of type CallbackFunc, which is defined but not used here. callbackFunc is also a non descriptive name - "onEvent" or something is better, to show when the function is used

packages/blob-storage/internal/indexer/indexer.go
```diff
@@ -0,0 +1,94 @@
...
+		FromBlock: indexer.startHeight,
+	}
+
+	logsCh := make(chan types.Log)
+	sub, err := client.SubscribeFilterLogs(context.Background(), query, logsCh)
+	if err != nil {
+		log.Fatal("Failed to subscribe to logs:", err)
+	}
+
+	log.Println("Indexer for ", indexer.networkName, " started.")
```
cyberhorsey, line 75: use slog package

packages/blob-storage/internal/indexer/indexer.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Indexer struct holds the configuration and state for the Ethereum chain listener.
+type Indexer struct {
+	rpcURL            string
+	beaconURL         string
+	networkName       string
+	contractAddress   common.Address
+	eventHash         common.Hash
+	startHeight       *big.Int
+	eventCallbackFunc func(string, string, string, types.Log)
```
cyberhorsey, line 22: nondescriptive as to what the three things are - you could provide named parameters here

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// Callback functions
+
+// BlockProposedCallback is a callback for the "BlockProposed" event.
+func BlockProposedCallback(rpcURL, beaconURL, networkName string, log types.Log) {
+
+	contractAbi, err := abi.JSON(strings.NewReader(taikol1.TaikoL1ABI))
+	if err != nil {
+		slog.Error("Could not initiate reader")
```
cyberhorsey, line 45: if we encounter an error here, we don't return, only log,  so the rest of the function will continue. also, we aren't logging what error occurred

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,188 @@
...
+			if !ok {
+				slog.Info("SetString: error")
+				return errors.New("SetString: error")
+			}
+			blockTs, err := getBlockTimestamp(rpcURL, blockNrBig)
+
+			if err != nil {
+				slog.Info("TIMESTAMP issue")
+				return errors.New("TIMESTAMP issue")
+			}
```
cyberhorsey, line 139: don't return a new error, return the error that actually happened

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,188 @@
...
+			blockNrBig, ok := n.SetString(blockID, 10)
+			if !ok {
+				slog.Info("SetString: error")
+				return errors.New("SetString: error")
+			}
+			blockTs, err := getBlockTimestamp(rpcURL, blockNrBig)
+
+			if err != nil {
+				slog.Info("TIMESTAMP issue")
+				return errors.New("TIMESTAMP issue")
```
cyberhorsey, line 138: you're using slog.Info, should be slog.Error and log trh actual error

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,188 @@
...
+		"blob_hash":      blobHashInMeta,
+		"kzg_commitment": kzgCommitment,
+		"timestamp":      blockTs,
+		"blob_data":      blob, // Assuming this is the blob data field
+	})
+	if err != nil {
+		return err
+	}
+
+	log.Println("Blob data inserted into MongoDB successfully")
```
cyberhorsey, line 183: mix of slog and log, please use slog

packages/blob-storage/internal/logic/config.go
```diff
@@ -0,0 +1,94 @@
+package logic
+
+import (
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+// CallbackFunc represents the type of the callback function for handling events.
+type CallbackFunc func(string, string, string, types.Log)
```
cyberhorsey, line 8: define custom types for the strings, or name the parameters

cyberhorsey (changes requested):

packages/blob-storage/internal/logic/config.go
adaki2004: To be honest, i'd leave the decision to you or DevOps ppl, i dont know best practices there.

packages/blob-storage/internal/logic/config.go
adaki2004: Yeah, i dont have a strong opinion, but the URL for example is used once, when you initialize the indexer, so.. how you can modify the indexer on the go ? I think you initialize it, and once changes you stop and restart with a new, no ?
But i'd leave it for you guys.

adaki2004:
> I was requested to review this but don't know what I'm reviewing. @d1onys1us any context? It prints blobs out from blocks? who wrote it originally

Hey @cyberhorsey @mask-pp !

This subfolder includes 2 executables - i was doing it previously - a `"blob-catcher"` and a `"server"`.

Blob-catcher does nothing else but listening on BlockProposed events, and if it sees a blob is used (`blobUsed`) inside  , then it will get the `blobHash`, queries lighhouse for sidecar (blobs) with blockId and constructs the blobHash from the commitment, and compares it. If it matches, it stores it into DB (MongoDB). This way we are sure we store Taiko related blobs in our DB.

Server on the other hands just tries to retrieve the blob based on `blobHash(es)`.

This is an MvP/PoC, it works but if it does not fit into the styling, maybe it is the best if you guys do the finishing touches (logs / command line arguments, etc.) because most probably it would take me much more time. I guess it is not the highest prio (for anyone) since it is proven to be working - if config vars set correctly, currently the even hash is wron, because it has changed lately - but it is definitely working. (via the readme there is a small howto).

Please help me get this into shape, where it fits with the current go code, as i know you are the go-masters here.

packages/blob-storage/internal/indexer/indexer.go
```diff
@@ -0,0 +1,94 @@
...
+		log.Fatal("Failed to connect to the Ethereum client:", err)
+	}
+
+	query := ethereum.FilterQuery{
+		Addresses: []common.Address{indexer.contractAddress},
+		FromBlock: indexer.startHeight,
+	}
+
+	logsCh := make(chan types.Log)
+	sub, err := client.SubscribeFilterLogs(context.Background(), query, logsCh)
```
davidtaikocha, line 70: Geth subscription is not promising: https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub#create-subscriptions 

image

Maybe we need to scan all L1 blocks one by one, and we have the same need in taiko-client, so not 100% sure but probly maybe u can reuse this package: https://github.com/taikoxyz/taiko-client/blob/main/pkg/chain_iterator/event_iterator/block_proposed_iterator.go

packages/blob-storage/internal/logic/callbacks.go
```diff
@@ -0,0 +1,188 @@
...
+		KzgCommitment    string `json:"kzg_commitment"`
+		KzgCommitmentHex []byte `json:"-"`
+	} `json:"data"`
+}
+
+// Callback functions
+
+// BlockProposedCallback is a callback for the "BlockProposed" event.
+func BlockProposedCallback(rpcURL, beaconURL, networkName string, log types.Log) {
+
```
davidtaikocha, line 42:
```suggestion
```

packages/blob-storage/internal/indexer/indexer.go
adaki2004: TBH i'd leave it to @cyberhorsey to not waste to much time with theses. I think it would take a pro-go dev. `x` time, but for me maybe `2x` or even `3x`

cyberhorsey: done, we filter on an event loop in batches now

packages/blobstorage/indexer/indexer.go
```diff
@@ -0,0 +1,357 @@
...
+	}()
+
+	return nil
+}
+
+// eventLoop runs on an interval ticker, every N seconds we will check
+// the latest processed block, and the latest header, and filter every block in between
+// for BlockProposed events, if we are not already filtering. This lets us avoid
+// unreliable subscription issues.
+func (i *Indexer) eventLoop(ctx context.Context, startBlockID uint64) error {
```
mask-pp, line 112: startBlockID is not used.

cyberhorsey: ah I should add the linter to this repo, I didn't notice it wasn't added before. it will catch things like this. I will add the linter tonight

mask-pp: Ok, I already gave a [commit](https://github.com/taikoxyz/taiko-mono/pull/16353/commits/6c0d98fcedc0b25b9a7f5d0386aa0968d93794aa) and removed this one. Later you can add it again.

mask-pp: Also, I don't think `filtering logic` is necessary, because `filter function` is not running Individually.
image

cyberhorsey: but now the code will run filter every time the ticker ticks. what if the previous filter is still running?

cyberhorsey (approved):

adaki2004 (approved):

PR #16361, opened 2024-03-08
chore(protocol): avoid copying any data when send Ether

dantaik:
I've confirmed that our implementation doesn't utilize the `returndata` for any ether-sending operations. This PR eliminates the data copying back to the caller, reducing gas costs. Additionally, I've recognized that some functions can tolerate failures, while others must ensure the successful transfer of ether. To address this, I've divided `sendEther` into two distinct functions: `sendEther` for standard transfers and `sendEtherAndVerify` for transfers that require confirmation of success, enhancing code clarity.


packages/protocol/contracts/libs/LibAddress.sol
```diff
@@ -36,7 +37,31 @@ library LibAddress {
...
+    /// @dev Sends Ether to the specified trusted address.
+    /// @param _to The recipient address.
+    /// @param _amount The amount of Ether to send in wei.
+    /// @param _gasLimit The max amount gas to pay for this transaction.
+    function sendEther(address _to, uint256 _amount, uint256 _gasLimit) internal {
+        // Check for zero-address transactions
+        if (_to == address(0)) revert ETH_TRANSFER_FAILED();
+
+        // Attempt to send Ether to the recipient address
+        (bool success,) = payable(_to).call{ value: _amount, gas: _gasLimit }("");
```
Brechtpd, line 57: Is it necessary to use `excessivelySafeCall` if the return data is never read like here? I would think no `returndatacopy` instruction is generated and so no attack is possible, though I haven't looked at the generated bytecode to confirm.

dantaik: According to [this issue](https://github.com/ethereum/solidity/issues/12306), the data is copied even if it's not used.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -112,7 +112,7 @@ contract AssignmentHook is EssentialContract, IHook {
         // Ether or ERC20 tokens.
         if (assignment.feeToken == address(0)) {
             // Paying Ether
-            _blk.assignedProver.sendEther(proverFee, MAX_GAS_PAYING_PROVER);
+            _blk.assignedProver.sendEtherAndVerify(proverFee, MAX_GAS_PAYING_PROVER);
```
Brechtpd, line 115: should be allowed to fail I think, especially because of the gas limit

dantaik: Emitting an event in case this payment fails.

Brechtpd (approved):

adaki2004 (approved):

PR #16363, opened 2024-03-08
feat(protocol): make TaikoGovernor codesize smaller

adaki2004:
The reason behind this change is, that we basically do not use any of the feature EssentialContract offers, except the ownership, so why not just use pure ownership here ?

### DIFF
--- a/packages/protocol/contracts/L1/gov/TaikoGovernor.sol
+++ b/packages/protocol/contracts/L1/gov/TaikoGovernor.sol
@@ -1,22 +1,20 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.24;
 
-import "@openzeppelin/contracts-upgradeable/governance/GovernorUpgradeable.sol";
 import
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/L1/gov/TaikoGovernor.sol
```diff
@@ -58,33 +55,6 @@ contract TaikoGovernor is
...
-        override(GovernorCompatibilityBravoUpgradeable)
-        returns (uint256)
-    {
-        if (_signatures.length != _calldatas.length) revert TG_INVALID_SIGNATURES_LENGTH();
-
-        return GovernorCompatibilityBravoUpgradeable.propose(
-            _targets, _values, _signatures, _calldatas, _description
-        );
-    }
-
```
adaki2004, line 87: This is removed because with the new version it is already patched, we no need to make this patch here.

dantaik (approved):

PR #16390, opened 2024-03-11
chore(protocol): remove unused SGX code

dantaik:
Are we supposed to use some of the code in the future? I think we should keep only code that we used.


Brechtpd (approved):
Makes sense to me, but unsure why the code is there in the first place @smtmfft? Do you think it will ever be used?

yueawang:
Yes, the RSA part can be removed totally, the parsing part before is to compatible with the legacy path (i.e., parse quote string online), as long as we don't care the legacy path, these can be also removed. The use case is, if RA is not working, we switch back to admin register mode.

smtmfft (approved):

smtmfft:
> Makes sense to me, but unsure why the code is there in the first place @smtmfft? Do you think it will ever be used?

No, we will not parse quote online again.

PR #16393, opened 2024-03-11
chore(protocol): code4rena automated bot findings

adaki2004:
Non-breaking , minor changes.

Changes are from: https://github.com/code-423n4/2024-03-taiko/blob/main/4naly3er-report.md

Suggested changes evaluation: https://www.notion.so/taikoxyz/gas-code4rena-8c82409063704d25a7f5e3b714349c2a

### DIFF
--- a/packages/protocol/contracts/L1/TaikoToken.sol
+++ b/packages/protocol/contracts/L1/TaikoToken.sol
@@ -44,12 +44,12 @@ contract TaikoToken is EssentialContract, ERC20SnapshotUpgradeable, ERC20VotesUp
     /// @notice Burns tokens from the specified address.
     /// @param _from The address to burn tokens from.
     /// @param _amount The amount of tokens to burn.
-    function burn(address _from, uint256 _amount) public onlyOwner {
+    function burn(address _from, uint256 _amount) public payable onlyOwner {
         _burn(_from, _amount);
     }
 
     /// @notice Creates a new token snapshot.
-    function snapshot() public onlyFromOwnerOrNamed("snapshooter") {
+    function snapshot() public payable onlyFromOwnerOrNamed("snapshooter") {
         _snapshot();
     }
 
--- a/packages/protocol/contracts/L1/libs/LibProving.sol
+++ b/packages/protocol/contracts/L1/libs/LibProving.sol
@@ -191,7 +191,7 @@ library LibProving {
         if (isTopTier) {
             // A special return value from the top tier prover can signal this
             // contract to return all liveness bond.
-            bool returnLivenessBond = blk.livenessBond > 0 && _proof.data.length == 32
+            bool returnLivenessBond = blk.livenessBond != 0 && _proof.data.length == 32
                 && bytes32(_proof.data) == RETURN_LIVENESS_BOND;
 
             if (returnLivenessBond) {
@@ -251,7 +251,7 @@ library LibProving {
                 // doesn't have any significance.
                 ts.contestBond = tier.contestBond;
                 ts.contester = msg.sender;
-                ts.contestations += 1;
+                ts.contestations = ts.contestations + 1;
 
                 emit TransitionContested({
                     blockId: blk.blockId,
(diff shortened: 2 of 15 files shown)


packages/protocol/contracts/L1/TaikoToken.sol
```diff
@@ -44,12 +44,12 @@ contract TaikoToken is EssentialContract, ERC20SnapshotUpgradeable, ERC20VotesUp
     /// @notice Burns tokens from the specified address.
     /// @param _from The address to burn tokens from.
     /// @param _amount The amount of tokens to burn.
-    function burn(address _from, uint256 _amount) public onlyOwner {
```
dantaik, line 47: There are other instances where we shall add 'payable' if this practice is applied. But prefer keeping checking if ether shall be accepted rather than save gas by removing it.

dantaik (changes requested):

dantaik:
Screenshot 2024-03-11 at 23 14 27

I like this one :D

packages/protocol/contracts/L1/TaikoToken.sol
dantaik: @Brechtpd what's your take here?

Brechtpd: I think I also prefer not having functions marked as payable if they do not need to accept ETH, feels strangely inconsistent. Functions marked as `onlyOwner` are also likely not to be called often so the couple of gas saved doesn't seem to make much of a difference either.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -251,7 +251,7 @@ library LibProving {
                 // doesn't have any significance.
                 ts.contestBond = tier.contestBond;
                 ts.contester = msg.sender;
-                ts.contestations += 1;
+                ts.contestations = ts.contestations + 1;
```
Brechtpd, line 254: I'm not the biggest fan of this because now you have to check that both values on both sides are the same. Which isn't a big thing, but the presumably 16 gas saved also isn't a big thing either hehe. In any case, fine for me. :)

dantaik: Lets revert this change, @adaki2004 and also ignore the `unchecked {++i}` one.

dantaik:
Based on the discussion in this PR, we only need to take care of:


- GAS-13 | Use of this instead of marking as public an external function
- [GAS-15](https://github.com/code-423n4/2024-03-taiko/blob/main/4naly3er-report.md#GAS-15)Use != 0 instead of > 0 for unsigned integer comparison
- 
- GAS-9 | ++i costs less gas compared to i++ or i += 1 (same for --i vs i-- or i -= 1)

packages/protocol/contracts/L1/libs/LibProving.sol
adaki2004: Ok

adaki2004:
>     * GAS-14 | Increments/decrements can be unchecked in for-loops

With .8.22 or above, the for loops are optimized by default - this is just from an auto-bot finding, most probably the bot is not updated.

https://soliditylang.org/blog/2023/10/25/solidity-0.8.22-release-announcement/

kép

dantaik:
> > ```
> > * GAS-14 | Increments/decrements can be unchecked in for-loops
> > ```
> 
> With .8.22 or above, the for loops are optimized by default - this is just from an auto-bot finding, most probably the bot is not updated.
> 
> https://soliditylang.org/blog/2023/10/25/solidity-0.8.22-release-announcement/
> 
> kép

Good to know

adaki2004:
>     * GAS-13 | Use of this instead of marking as public an external function

This automated findings is also wrong, because `this.onMessageInvocation` is not an invocation but encoding some `bytes4` (function selector) into the data. I tested with `this.onMessageInvocation` and `IBridge.onMessageInvocation` -> no effect.

PR #16394, opened 2024-03-11
fix(protocol): add address manager to taiko token

adaki2004:
Without this, for example the function `function snapshot() public onlyFromOwnerOrNamed("snapshooter")` is not interpretable, since there is no address manager to look the address up.


davidtaikocha (approved):

dantaik:
Screenshot 2024-03-11 at 23 14 27

I like this one :D

adaki2004:
> Screenshot 2024-03-11 at 23 14 27
> 
> I like this one :D

Wrong PR, but yeah, not all valid findings, this is why i filtered. But that is in https://github.com/taikoxyz/taiko-mono/pull/16393

dantaik (approved):

PR #16399, opened 2024-03-11
feat(repo): refactor blob storage pkg

cyberhorsey:
This Pr brings Dani's blob-catcher software up to speed with our existing architectural patterns., fixed some bugs, and cleans up the code, as well as fixes logging.

It initializes an ethclient / db connection once, rather than for every event it finds, and as well, it gets events in batches (as before, it was getting infinite number of block data at once, which would cause it to crash if too many blocks). Then, when it is caught up, it subscribes to new blocks.

@adaki2004 can i take out MongoDB and make it MySQL? we already use MySQL in production, dont see the need for two different database. Or is there a reason we use Mongo here?


m3shm:
What u mean by WIP

cyberhorsey:
> What u mean by WIP

work in progress. who are you?

davidtaikocha:
> @adaki2004 can i take out MongoDB and make it MySQL? we already use MySQL in production, dont see the need for two different database. Or is there a reason we use Mongo here?

imo we should also use MySQL for this service, can't any reason to maintain a new type of database?

packages/blob-storage/indexer/indexer.go
```diff
@@ -0,0 +1,388 @@
...
+
+		opts := &bind.FilterOpts{
+			Start:   processingBlockHeight,
+			End:     &filterEnd,
+			Context: i.ctx,
+		}
+
+		group, groupCtx := errgroup.WithContext(i.ctx)
+
+		events, err := i.taikoL1.FilterBlockProposed(opts, nil, nil)
```
davidtaikocha, line 152: Not sure if it is simpler for this repo, shall we just `import "github.com/taikoxyz/taiko-client/pkg/chain_iterator/event_iterator"` package here, and reuse `BlockProposedIterator`?

packages/blob-storage/indexer/indexer.go
```diff
@@ -0,0 +1,388 @@
...
+		if err != nil {
+			return err
+		}
+
+		for events.Next() {
+			event := events.Event
+			group.Go(func() error {
+				if err := i.storeBlob(groupCtx, event); err != nil {
+					slog.Error("error storing blob", "error", err)
```
davidtaikocha, line 161: Shall we return the error / retry here? since we want to save every txlist blob into database?

packages/blob-storage/indexer/indexer.go
```diff
@@ -0,0 +1,388 @@
...
+func (i *Indexer) subscribe(ctx context.Context) error {
+	defer func() {
+		i.wg.Done()
+	}()
+
+	slog.Info("subscribing to new events")
+
+	errChan := make(chan error)
+
+	go i.subscribeBlockProposed(ctx, errChan)
```
davidtaikocha, line 191: Since the event subscription is not promising: https://github.com/taikoxyz/taiko-mono/pull/16353#discussion_r1519176189 , shall we keep using the new blocks scanning logic here?

cyberhorsey:
> > @adaki2004 can i take out MongoDB and make it MySQL? we already use MySQL in production, dont see the need for two different database. Or is there a reason we use Mongo here?
> 
> imo we should also use MySQL for this service, can't any reason to maintain a new type of database?

don't think so, it was just started with Mongo. I will swap it out for MySQL.

packages/blob-storage/indexer/indexer.go
cyberhorsey: I didn't know that. We use subscriptions in the relayer and eventindexer as well. I should change it for both of them as well, then.  
  
I will do it for this PR first.

packages/blob-storage/indexer/indexer.go
cyberhorsey: I'll take a look tomorrow, perhaps we wait until Taiko client is merged into this codebase otherwise the imports will break when we delete the Taiko client repo

packages/blob-storage/indexer/indexer.go
cyberhorsey: retry and error handled

packages/blob-storage/indexer/indexer.go
cyberhorsey: I updated it to run on an event loop ticker instead, and crawl blocks. Please take a look. No more subscriptions.

I will do the same for our other services.

PR #16404, opened 2024-03-12
fix(protocol): fix bridge prove message issue using staticcall

dantaik:
Previously `proveSignalReceived` in signalService is a `view` function, but now it can cache data so it's no longer a `view` function. Therefore, in Bridge.sol, we can no longer use `staticcall`.

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -572,10 +572,12 @@ contract Bridge is EssentialContract, IBridge {
         bytes calldata _proof
     )
         private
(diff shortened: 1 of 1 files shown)


adaki2004:
> Previously `proveSignalReceived` in signalService is a `view` function, but now it can cache data so it's no longer a `view` function. Therefore, in Bridge.sol, we can no longer use `staticcall`.

How could it work even on an internal devnet ? Or simply we never cached the data, that was it ?

adaki2004 (approved):

davidtaikocha (approved):

PR #16425, opened 2024-03-13
feat(protocol): allow L2 contracts to read L2's parent block timestamp

dantaik:
This is a small feature but it may be very useful for us to incentivize block proposers for liveness. For example, we can [allow a fair-mint token](https://github.com/taikoxyz/taiko-mono/pull/16408) that only can be minted by `block.coinbase` for each L2 block but only when the L2's block is x seconds later than the parent's block timestamp. Without being able to read the parent L2 block's timestamp, this is impossible.

This tiny feature imposes a very small amount of additional gas cost.

### DIFF
--- a/packages/protocol/contracts/L2/TaikoL2.sol
+++ b/packages/protocol/contracts/L2/TaikoL2.sol
@@ -49,6 +49,12 @@ contract TaikoL2 is CrossChainOwned {
     /// @notice The last synced L1 block height.
     uint64 public lastSyncedBlock;
 
+    /// @notice The parent block's timestamp.
+    uint64 public parentTimestamp;
+
+    /// @notice The current block's timestamp.
+    uint64 private __currentBlockTimestamp;
+
     uint256[47] private __gap;
 
     /// @notice Emitted when the latest L1 block details are anchored to L2.
@@ -95,6 +101,7 @@ contract TaikoL2 is CrossChainOwned {
 
         gasExcess = _gasExcess;
         (publicInputHash,) = _calcPublicInputHash(block.number);
+        __currentBlockTimestamp = uint64(block.timestamp);
     }
 
     /// @notice Anchors the latest L1 block details to L2 for cross-layer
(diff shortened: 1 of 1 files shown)


davidtaikocha (approved):

adaki2004:
> This is a small feature but it may be very useful for us to incentivize block proposers for liveness. For example, we can allow a fair-mint token that only can be minted by `block.coinbase` for each L2 block but only when the L2's block is x seconds later than the parent's block timestamp. Without being able to read the parent L2 block's timestamp, this is impossible.
> 
> This tiny feature imposes a very small amount of additional gas cost.

While i understand the mechanics you described, i'd be interested in what would be that "free" or "fair" mint token ?

AFAIK TKO will have constant supply (no deflationary, no inflationary) so it must be something else. If something else, what would it be and who would back that token ? or rather what would it give value - so that it incentivizes the liveness ? Should not liveness be incentivized simply by the collected fees ? If that is not enough, proposers should simply wait until they are profitable, even tho txns would be a bit delayed - but this is a known 'issue'.

dantaik:
> > This is a small feature but it may be very useful for us to incentivize block proposers for liveness. For example, we can allow a fair-mint token that only can be minted by `block.coinbase` for each L2 block but only when the L2's block is x seconds later than the parent's block timestamp. Without being able to read the parent L2 block's timestamp, this is impossible.
> > This tiny feature imposes a very small amount of additional gas cost.
> 
> While i understand the mechanics you described, i'd be interested in what would be that "free" or "fair" mint token ?
> 
> AFAIK TKO will have constant supply (no deflationary, no inflationary) so it must be something else. If something else, what would it be and who would back that token ? or rather what would it give value - so that it incentivizes the liveness ? Should not liveness be incentivized simply by the collected fees ? If that is not enough, proposers should simply wait until they are profitable, even tho txns would be a bit delayed - but this is a known 'issue'.
> 
> Or it could also be used for off-chain mechanisms (airdrop of TKO) by the DAO to active proposers. (?)

It is not the TKO token for sure, but it can ben NFTs or another reward token that we can promise will receive some airdrop as part of our trailblazer campaign.

adaki2004 (approved):

Brechtpd:
>  Without being able to read the parent L2 block's timestamp, this is impossible.

It is pretty much already possible to read the parent's block timestamp by reading it from the parent block's blockhash. And because all L2 blockhashes are stored, that way allows even reading it from any L2 block compared to just the current block. Of course, this does depend on putting the blockheader data corresponding to the blockheader onchain to make it available to be used in a smart contract.

PR #16428, opened 2024-03-13
chore(protocol): improve test coverage

adaki2004:
Goal is to improve test coverage.

Current main:
kép

This PR:
kép


packages/protocol/contracts/L1/TaikoToken.sol
```diff
@@ -51,8 +51,8 @@ contract TaikoToken is EssentialContract, ERC20SnapshotUpgradeable, ERC20VotesUp
     }
 
     /// @notice Creates a new token snapshot.
-    function snapshot() public onlyFromOwnerOrNamed("snapshooter") {
-        _snapshot();
+    function snapshot() public onlyFromOwnerOrNamed("snapshooter") returns (uint256) {
```
dantaik, line 54: BridgedERC20 also has a `snapshot` function that you need to change. If this is a bug, then we should have a fix(protocol)... PR.

dantaik: See https://github.com/taikoxyz/taiko-mono/pull/16436

dantaik (approved):

packages/bridge-ui/src/components/TokenDropdown/TokenDropdown.svelte
```diff
@@ -221,7 +221,7 @@
         <span class="title-subsection-bold text-base text-secondary-content">{$t('token_dropdown.label')}</span>
       {:else if value}
         <div class="flex f-space-between space-x-2 items-center text-secondary-content">
-          <!-- Only match icons to configurd tokens -->
+          <!-- Only match icons to configured tokens -->
```
d1onys1us, line 224: please dont include this change, or include it as part of a separate PR. PR should only modify a single package ideally.

davidtaikocha (approved):

Issue #16429, opened 2024-03-13
Transaction panel Dropdown menu not reversing


whitetechna:
### The transaction panel is not reversing in Bridge UI. 

![taiko transaction-1](https://github.com/taikoxyz/taiko-mono/assets/153755991/dedadaba-befc-43d6-82da-170aa931ac6c)

![taiko issue-1](https://github.com/taikoxyz/taiko-mono/assets/153755991/f4ceb561-1c96-4c32-abb9-320638c0d1db)

**Just reverse the panel in Bridge Transaction and in NFT transaction section.**

**- solution:** 
- create reverse technique in icon of transaction.
- make condition in `StatusFilterDropdown.svelte` to open that toggle menu.

I have taken the liberty to fix this issue and you can find the subsequent PR below 

https://github.com/taikoxyz/taiko-mono/pull/16430

### Spam policy

- [X] I verify that this issue is NOT SPAM and understand SPAM issues will be closed and reported to GitHub, resulting in ACCOUNT TERMINATION.

courage-ugw:
Hi @KorbinianK has this been fixed ?

dantaik:
@courage-ugw please verify if the bug has been fixed, if not, you are welcome to work on it.

courage-ugw:
> @courage-ugw please verify if the bug has been fixed, if not, you are welcome to work on it.

@dantaik I have attempted to do a fix for it here: https://github.com/taikoxyz/taiko-mono/pull/16609. Kindly review. Let me know if there is any issue. Thanks

KorbinianK:
Fixed in our next testnet UI version

PR #16430, opened 2024-03-13
Fix(Transaction-UI) : toggle menu-change with conditional dropdown and close.

whitetechna:
#16429

This PR is the solution to the issue raised in the above link 

"The transaction panel is not reversing in the Bridge UI"


cryptoryda:
PS. This PR is the fix for https://github.com/taikoxyz/taiko-mono/issues/16429 
Where The issue was that the transaction panel is not reversing in The Bridge UI as shown in the issue

KorbinianK:
This does not fix the problem and just by looking at the change this should be obvious

PR #16431, opened 2024-03-13
Fix(GitHub) : spell-env fixes in the workflow files.

whitetechna:
Spelling issue in the code
     
     SKIP_ENV_VALDIATION: "true" 
Changed to 
          SKIP_ENV_VALIDATION: "true"

"VALIDATION" being the misspelt word

### DIFF
--- a/.github/workflows/bridge-ui.yml
+++ b/.github/workflows/bridge-ui.yml
@@ -37,7 +37,7 @@ jobs:
 
       - name: Build Svelte app
         env:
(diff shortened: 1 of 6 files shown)


KorbinianK:
While we appreciate your effort, we are not accepting simple spelling fixes of variables. 
As this is a variable that is just used during the build job, it is not even actively used in the application.

Also you did not follow our Contribution guidelines reg. opening PRs.

The spelling mistake will be fixed in an upcoming bigger PR, so thank you for raising it.

PR #16432, opened 2024-03-13
chore(protocol): apply a minor bridge improvement

dantaik:
if `gasLImit` is zero, we skip the underlying `call` to save gas and consider the call fails.


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -278,7 +278,7 @@ contract Bridge is EssentialContract, IBridge {
                 // use remaining gas
                 uint256 gasLimit = msg.sender == _message.destOwner ? gasleft() : _message.gasLimit;
 
-                if (_invokeMessageCall(_message, msgHash, gasLimit)) {
+                if (gasLimit != 0 && _invokeMessageCall(_message, msgHash, gasLimit)) {
```
adaki2004, line 281: I think it is not necessary, because in `_invokeMessageCall()` we already revert if gasLimit is actually 0.

PR #16439, opened 2024-03-14
fix(protocol): remove the blob-reuse feature

dantaik:
Based on our discussion in the previous all-hands, it seems reuse-blob is not a feature that we will use.

Also removed the txListSize from Block metadata as it is not necessary and not used by the client at all.

BEGIN_COMMIT_OVERRIDE
fix(protocol): remove the blob-reuse feature
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -75,7 +75,6 @@ library TaikoData {
         address assignedProver;
         address coinbase;
         bytes32 extraData;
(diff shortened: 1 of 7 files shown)


adaki2004:
As discussed offline, we dont need the blob-resue if we:
- increase gas limit on L2
- confident we can utilize blobs efficiently (maybe a check in the proposer sw)

adaki2004 (approved):

davidtaikocha (approved):

PR #16464, opened 2024-03-16
fix(blobstorage): fix command instructions and missing local_docker folder in packages/blobstorage

SamiAlHassan:
I noticed that the current commands in the README.md to start the indexer and server are not correctly formatted and don't work. They are missing the `go` keyword to run Go programs. I have updated these commands to the proper format.

Also, the README.md references a `local_docker` directory for starting MongoDB with `docker-compose up -d`, but this directory does not exist in the current repository structure so I can't start the indexer & server. Maybe it wasn't staged and commited?

If the `local_docker` directory and the associated `docker-compose.yml` file are available but have not been pushed to the repository, could they please be added?

### **UPDATE:**

I've made the necessary updates to reflect the migration to a mysql database setup. Including:

- Adding a `docker-compose.yml` file to facilitate starting the MySQL database.
- Updating all relevant configurations to accurately reflect the migration from MongoDB to mysql.
- Revising the README to accurately represent these changes and provide correct command instructions.


cyberhorsey:
that readme is a hangover from when we used MongoDB which we no longer do. they can be removed from the readme

packages/docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "3.8"
```
RogerLamTd, line 1: Shouldn't this docker-compose folder not be a separate package? File structure should be blobstorage/docker-compose/docker-compose.yml?

cyberhorsey: yes

RogerLamTd: @SamiAlHassan please alter it to match the relayer then!

RogerLamTd (approved):

cyberhorsey (approved):

PR #16465, opened 2024-03-16
fix(relayer): restart indexer and crawler when they exit due to failed API requests

xiaodino:
Use https://relayer.katla.taiko.xyz/events?event=MessageSent&page=0&size=100&address=0x9e47FBb2a2A27B3b02E4a63b3eF5A3dC863c0223 as the example.

Before, the API response only returns 16 items. With the change, the API returns 19 items.

When a crawler is restarted, it will start from the last processing block height, rather than always from the genesis block.

### DIFF
--- a/packages/relayer/indexer/indexer.go
+++ b/packages/relayer/indexer/indexer.go
@@ -218,10 +218,6 @@ func (i *Indexer) Start() error {
 	i.wg.Add(1)
 
 	go func() {
-		defer func() {
-			i.wg.Done()
-		}()
-
 		if err := i.filter(i.ctx); err != nil {
 			slog.Error("error filtering blocks", "error", err.Error())
 		}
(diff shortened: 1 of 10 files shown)


davidtaikocha (approved):

packages/relayer/indexer/indexer.go
```diff
@@ -218,28 +218,38 @@ func (i *Indexer) Start() error {
...
-
-		if err := i.filter(i.ctx); err != nil {
-			slog.Error("error filtering blocks", "error", err.Error())
+		if err := backoff.Retry(func() error {
+			err := i.filter(i.ctx)
+			if err != nil {
+				slog.Error("filter failed, will retry", "error", err)
+			}
+			return err
+		}, backoff.NewConstantBackOff(5*time.Second)); err != nil {
```
mask-pp, line 227:
```suggestion
		}, backoff.WithContext(backoff.NewConstantBackOff(5*time.Second)), i.ctx); err != nil {
```

packages/relayer/indexer/indexer.go
```diff
@@ -218,28 +218,38 @@ func (i *Indexer) Start() error {
...
 
 	go func() {
 		if err := backoff.Retry(func() error {
-			return i.queue.Notify(i.ctx, i.wg)
+			err := i.queue.Notify(i.ctx, i.wg)
+			if err != nil {
+				slog.Error("i.queue.Notify failed, will retry", "error", err)
+			}
+			return err
 		}, backoff.NewConstantBackOff(5*time.Second)); err != nil {
```
mask-pp, line 251: ditto

mask-pp (approved):

PR #16475, opened 2024-03-18
fix(bridge-ui): correct display of forward arrow and handling of invalid pagination input

courage-ugwuanyi:
**This pull request addresses:**

1. Abug related to pagination in the taiko-mono project. Previously, the forward arrow button was incorrectly displayed even beyond the last page, leading to confusion for users. This bug has been fixed to ensure that the forward arrow button is displayed only when there are additional pages to navigate to.

2. Previously, when users input a number greater than the total number of pages in the pagination input field, the application would attempt to navigate to the specified page, leading to unexpected behavior. This is addressed by adding validation in the handleKeydown function to check if the input number is within a valid range i.e. between 0 and the total number of pages. If the input is invalid, the application ignores the input.

**How to Test:**

1. Navigate to the transaction records page.
2. Verify that the forward arrow button is not displayed on the last page.
3. Verify that when a number outside the valid range (between 0 and total pages) that the page remains unchanged
4. Input a number greater than the total number of pages and ensure that the forward arrow button remains hidden.


KorbinianK:
@courage-ugwuanyi please fix the linting errors :)

courage-ugwuanyi:
> @courage-ugwuanyi please fix the linting errors :)

Please if the error still exists

KorbinianK:
@courage-ugwuanyi it still fails. Run `pnpm svelte:check && pnpm lint:fix`

PR #16483, opened 2024-03-19
refactor(guardian-prover-health-check-ui):  use reduce to avoid for in loop and eliminate temporary variables.

hyperbolic-versor:
### Description

`getGuardianProverIdsPerBlockNumber.ts`:  avoid ` for in ` loop ,  use Object.keys to obtain the keys of the blocks object, ensuring that is processed in the order of insertion.

`signedBlocksPerGuardian.ts`:   eliminate temporary variables , `count` can be removed and avoid using the increment operator.


canter1:
```typescript
struct Some {
  uint256 foo; /* This is foo */
}
```

KorbinianK (approved):

davidtaikocha (approved):

PR #16502, opened 2024-03-22
chore(repo): preview builds for internal project

KorbinianK:
WIP

With the change we get the preview builds on the internal vercel project and can then manually promote those to production.

That way we can assign a fixed domain and permissions while staying entirely separate from the public UI:


.github/workflows/bridge-ui-preview.yml
```diff
@@ -12,22 +13,25 @@ on:
 
 jobs:
   Deploy-Preview:
-    runs-on: [taiko-runner]
+    runs-on: ubuntu-latest
```
d1onys1us, line 16: why not [taiko-runner]

dionysuzx (approved):

.github/workflows/bridge-ui-preview.yml
KorbinianK: fixed

davidtaikocha (approved):

PR #16505, opened 2024-03-23
feat(relayer): unprofitable queue + use txmgr

cyberhorsey:
This PR introduces a new queue for unprofitable messages, and updates our tx sender to use the same on as the new taiko-client.

It is likely on Mainnet, we will have messages that are unprofitable to process. We don't want to discard these messages, nor do we want to process them and lose money. Intead, we should publish them to a queue to "sit and die" for a little bit. A delayed Queue.

Then, when they expire in that queue, they should be re-routed to the original queue via a dead-letter mechanism, to see if they are yet profitable, and enter the normal workflow of message processing again.

It also adds new metrics to track this.

This has been deployed onto our latest internal devnet as well.


davidtaikocha (approved):

packages/relayer/cmd/flags/txmgr.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+var (
+	NumConfirmations = &cli.Uint64Flag{
+		Name:     "tx.numConfirmations",
+		Usage:    "Number of confirmations which we will wait after sending a transaction",
+		Value:    1,
+		Category: txmgrCategory,
+	}
+	SafeAbortNonceTooLowCount = cli.Uint64Flag{
```
RogerLamTd, line 16:
```suggestion
	SafeAbortNonceTooLowCount = &cli.Uint64Flag{
```

packages/relayer/cmd/flags/txmgr.go
```diff
@@ -0,0 +1,99 @@
...
+		Name:     "rpc.timeout",
+		Usage:    "Timeout in seconds for RPC calls",
+		Category: commonCategory,
+		Value:    12 * time.Second,
+	}
+)
+
+var TxmgrFlags = []cli.Flag{
+	NumConfirmations,
+	&SafeAbortNonceTooLowCount,
```
RogerLamTd, line 88:
```suggestion
	SafeAbortNonceTooLowCount,
```

xiaodino (approved):

PR #16531, opened 2024-03-25
ci(blobstorage): deploy blobs swagger api

RogerLamTd:
https://blobstorage-swagger.vercel.app/


.github/workflows/blobstorage-preview.yml
```diff
@@ -0,0 +1,33 @@
+env:
+  VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+  VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID_BLOBS }}
```
d1onys1us, line 3: better to follow the same naming convention. i think that would be `VERCEL_PROJECT_ID_BLOBSTORAGE`

packages/blobstorage/.gitignore
```diff
@@ -19,3 +19,4 @@
 
 # Go workspace file
 go.work
+.vercel
```
d1onys1us, line 22: we can just add this to the root gitignore because other vercel projects will benefit from it.

d1onys1us:
@RogerLamTd we need one for `blobstorage-production.yml` as well i believe. it will be very similar, you can copy the other examples.

dionysuzx (approved):

d1onys1us:
LGTM

dantaik (approved):

PR #16545, opened 2024-03-28
fix(protocol): fix bridge bug caused by incorrect check of `receivedAt` (by OZ)

dantaik:


Brechtpd (approved):
Approved because I think the code is correct, but as I said elsewhere I think more checks than really required.

adaki2004 (approved):

PR #16547, opened 2024-03-28
fix(relayer): func WaitConfirmations should not wait forever for ethereum.NotFound

xiaodino:
WaitConfirmations used in processor should not wait forever for ethereum.NotFound. It will cause the processor stuck.


davidtaikocha (approved):

dantaik (approved):

cyberhorsey:
@xiaodino we don't need this, we have a context cancel via timeout when this function is called. This should be unmerged

PR #16548, opened 2024-03-28
fix(protocol): fix suspend message

adaki2004:
Same as https://github.com/taikoxyz/taiko-mono/pull/16545 but a different way.
My last attempt to address comments from finding: https://github.com/taikoxyz/taiko-mono-private/pull/31#issuecomment-2024345097


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -89,6 +91,14 @@ contract Bridge is EssentialContract, IBridge {
         uint64 _timestamp = _suspend ? type(uint64).max : uint64(block.timestamp);
         for (uint256 i; i < _msgHashes.length; ++i) {
             bytes32 msgHash = _msgHashes[i];
+
+            // If not proven -> Don't do anything
```
dantaik, line 95: which part of the change in this PR that you don't lik?
https://github.com/taikoxyz/taiko-mono/pull/16545/files

adaki2004: > which part of the change in this PR that you don't lik? https://github.com/taikoxyz/taiko-mono/pull/16545/files

There is no such change i dont like there. I even accepted that (on private repo), but my take is: the simpler the better if reaching the same end- goal, no ? 

If this one does not solve a real issue, which the other does ! I'm fully OK closing this.

On the other hand, if this is the same as the other (which i think it is, or pretty close!) then i'd vouch for this simple one.

dantaik: Thanks, please review mine please.

PR #16550, opened 2024-03-28
test(protocol): calculate what gasExcess value will push basefee to 1gwei then 10gwei

dantaik:
In the modified test, you'll see if we sell 327 * (gasTarget = 60_000_000) gas, then the basefee will be pushed to 1gwei. This means the current math works.


davidtaikocha (approved):

packages/protocol/genesis/test_config.js
```diff
@@ -111,7 +111,7 @@ module.exports = {
     };
   },
   param1559: {
-    gasExcess: 1,
+    gasExcess: 20000000000,
```
Brechtpd, line 114: hmmm but gasExcess can go down to `1`, so when that happens it will take much longer for it to reach this much higher value again no? If so we probably do want to enforce a gasExcess minimum to whatever results in a 1 wei basefee.

dantaik: I also think it's a good idea to apply a min value for gasExcess.

Brechtpd: I guess at that point maybe https://github.com/taikoxyz/taiko-mono/pull/16551 is indeed just simpler if we also confirmed that works.

dantaik: See https://github.com/taikoxyz/taiko-mono/pull/16555 please

dantaik: If possible, please approve this PR to merge it first to main.

Brechtpd (approved):

davidtaikocha (approved):

PR #16554, opened 2024-03-28
fix(relayer): revert func `WaitConfirmations` should not wait forever for ethereum.NotFound

xiaodino:
Reverts taikoxyz/taiko-mono#16547

Synced with @cyberhorsey  offline. 
```
The context has a timeout
It will hit the case <-ctx.Done(): and return, after the contet's timeout
not the ethereum RPC's timeout.
ctx, cancelFunc := context.WithTimeout(ctx, time.Duration(p.confTimeoutInSeconds)*time.Second)
```


cyberhorsey (approved):

cyberhorsey:
needs title change

davidtaikocha (approved):

PR #16555, opened 2024-03-28
feat(protocol): apply a min value for gas excess

dantaik:
This guarantees a min base fee value.

Based on the current config, the min base fee is 0.099627953 wei or 0.099gwei. 

At the current moment, the Ethereum base fee is 40gwei:
 
Screenshot 2024-03-29 at 02 17 57


adaki2004:
> Based on the current config, the min base fee is 0.099627953 wei or 0.99gwei.

One question : i think the 99627953 is 0.099 gwei, not 0.99 gwei.

Is it ok and intended ? Beside that looks good to me.

Brechtpd (approved):

davidtaikocha (approved):

PR #16591, opened 2024-04-02
feat(protocol): add a view function isSignalReceived for Bridge relayer/UI

dantaik:
`isSignalReceived` has the same implementation as `proveSignalReceived` except it doesn't transact to cache data.

### DIFF
--- a/packages/protocol/contracts/signal/ISignalService.sol
+++ b/packages/protocol/contracts/signal/ISignalService.sol
@@ -106,6 +106,23 @@ interface ISignalService {
     )
         external;
 
+    /// @notice Verifies if a signal has been received on the target chain.
+    /// This is the "readonly" version of proveSignalReceived.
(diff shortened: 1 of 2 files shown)


Brechtpd (approved):
```
function isSignalReceived(
        uint64 _chainId,
        address _app,
        bytes32 _signal,
        bytes calldata _proof) 
  external 
  view
{
  bytes memory call_data = abi.encodePacked(SignalService.proveSignalReceived.selector, msg.data[4:]);
  (bool success, ) = address(this).staticcall(call_data);
  require(success);
}
```

Something like this should also work (though I have only tested with dummy functions) so the code doesn't have to be duplicated, though it would require the caller to disable any caching manually instead of the current way where it is simply ignored no matter the value.

cyberhorsey (approved):

PR #16593, opened 2024-04-02
fix(protocol): add 1 to _REENTRY_SLOT in EssentialContract

dantaik:
Change _REENTRY_SLOT from `0xa5054f728453d3dbe953bdc43e4d0cb97e662ea32d7958190f3dc2da31d9721a` to `0xa5054f728453d3dbe953bdc43e4d0cb97e662ea32d7958190f3dc2da31d9721a + 1`

Based on https://github.com/code-423n4/2024-03-taiko-findings/issues/11

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -16,10 +16,10 @@ contract Bridge is EssentialContract, IBridge {
     using LibAddress for address;
     using LibAddress for address payable;
 
-    /// @dev The slot in transient storage of the call context.
-    /// keccak256("bridge.ctx_slot") + 1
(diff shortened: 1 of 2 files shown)


Brechtpd (approved):
Approved, but we do assume the string being used is unique to already prevent collisions from happening. I guess a good general recommendation to be super super sure, but wouldn't call this an issue.

davidtaikocha (approved):
Emmm looks like we need to upgrade every contract we deployed both on L1 / L2 for this change...

dantaik:
> Emmm looks like we need to upgrade every contract we deployed both on L1 / L2 for this change...

We don't have to apply the change anytime soon. The previous code works fine.

PR #16599, opened 2024-04-02
fix(bridge-ui): fix token balance display

mintcnn:
There's a token balance displaying incorrectly when bridging following the step below
1. Select a token (TTKOk or Horse)
2. Select tab NFT
3. Select tab Token
4. Switch chain
5. The balance will display incorrectly
Screenshot 2567-04-01 at 22 43 34

So that I found that `tokenBalance` is not changed in `reset()`


KorbinianK:
Thanks, this is already coming with a PR later today, will be fixed then.

PR #16602, opened 2024-04-02
chore(protocol): index more events

dantaik:
Based on a report from code4rena - https://github.com/code-423n4/2024-03-taiko-findings/blob/main/data/0x11singh99-Q.md#n-11-event-is-missing-indexed-fields

> Index event fields make the field more quickly accessible to off-chain tools that parse events. However, note that each index field costs extra gas during emission, so it's not necessarily best to index the maximum allowed per event (three fields). Each event should use three indexed fields if there are three or more fields, and gas usage is not particularly of concern for the events in question. If there are fewer than three fields, all of the fields should be indexed.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoEvents.sol
+++ b/packages/protocol/contracts/L1/TaikoEvents.sol
@@ -77,9 +77,11 @@ abstract contract TaikoEvents {
         uint16 tier
     );
 
-    /// @dev Emitted when proving has been paused
-    /// @param paused True if paused, false if unpaused.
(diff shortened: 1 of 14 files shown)


Brechtpd (approved):
Approved but I don't have any opinion on what fields should or should not be indexed, it also seems like our own relayer did not need these indexed fields for the bridge for example and it does increase the gas cost a little bit.

dantaik:
after a second thought, maybe we don't need this PR :D

PR #16603, opened 2024-04-02
chore(protocol): emit event before calling external contract

dantaik:
https://github.com/code-423n4/2024-03-taiko-findings/blob/main/data/0x11singh99-Q.md#l-02-emit-events-before-external-call


Brechtpd (approved):
Okay having read a bit about it there does seem to be some benefit in doing so.

davidtaikocha (approved):

PR #16609, opened 2024-04-02
fix(bridge-ui): fix dropdown menu behavior and chevron icon in transactions status-filter-drop-down

courage-ugw:
**This pull request:**

1. Fixes the behavior where the dropdown menu did not close upon clicking again, and the chevron icon remained stagnant.
2. Resolves the issue where clicking elsewhere on the page while the dropdown menu was open failed to reverse the chevron icon to its previous state.

**Changes Made:**

1. Added logic to toggle the dropdown menu state and update the chevron icon accordingly.
2. Implemented event listeners to detect clicks outside the dropdown menu, allowing for automatic menu closure.
3. Updated the component behavior to ensure consistent dropdown menu functionality.

**How to Test:**

1. Open the transactions filter component in the application (https://bridge.katla.taiko.xyz/transactions)
2. Click on the dropdown button to open the menu.
3. Verify that the dropdown menu opens and displays the available options.
4. Click on the dropdown button again to close the menu.
5. Ensure that the dropdown menu closes, and the chevron icon reverts to its initial state.
6. Click anywhere outside the dropdown menu to close it.
7. Confirm that the dropdown menu closes, and the chevron icon flips back accordingly.

### DIFF
--- a/packages/bridge-ui/src/components/Transactions/Filter/StatusFilterDropdown.svelte
+++ b/packages/bridge-ui/src/components/Transactions/Filter/StatusFilterDropdown.svelte
@@ -9,7 +9,6 @@
 
   export let selectedStatus: MessageStatus | null = null;
 
-  let flipped = false;
   let menuOpen = false;
   let uuid = `dropdown-${uid()}`;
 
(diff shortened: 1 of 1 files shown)


KorbinianK:
While this may work, introducing a new click listener is not the correct fix. Removing the binding and replacing it with the function call is also unnecessary as svelte binding works just fine. A fix will be added in an upcoming PR.

PR #16610, opened 2024-04-02
fix(protocol): allow recall message to fail

adaki2004:
As per description: https://github.com/code-423n4/2024-03-taiko-findings/issues/279

So first of all, this "less-invasive" solution, than the author proposed is enough, since:
- We will not support USDC right away
- We won't expect recall() to be called oftern
- We won't expect that our bridge allowance - once approved by Circle - would be reduced to 0 (or to an exact numnber).

So investigating why an `onMessageRecalled` via a specific event might be more useful, and later on we can upgrade our contracts to reward the users if their message status is RECALL_FAILED.

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -209,7 +209,6 @@ contract Bridge is EssentialContract, IBridge {
 
         if (block.timestamp >= invocationDelay + receivedAt) {
             delete proofReceipt[msgHash];
-            messageStatus[msgHash] = Status.RECALLED;
 
             // Some cases recall might fail (e.g. USDC shuts down minting allowance)
             bool success = true;
@@ -235,8 +234,10 @@ contract Bridge is EssentialContract, IBridge {
 
             if (success) {
                 emit MessageRecalled(msgHash);
+                messageStatus[msgHash] = Status.RECALLED;
             } else {
                 emit MessageRecallFailed(msgHash, data);
+                messageStatus[msgHash] = Status.RECALL_FAILED;
             }
         } else if (isNewlyProven) {
             emit MessageReceived(msgHash, _message, true);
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -209,24 +209,36 @@ contract Bridge is EssentialContract, IBridge {
...
                 _message.srcOwner.sendEtherAndVerify(_message.value);
             }
-            emit MessageRecalled(msgHash);
+
+            if (success) {
+                emit MessageRecalled(msgHash);
+                messageStatus[msgHash] = Status.RECALLED;
+            } else {
+                emit MessageRecallFailed(msgHash, data);
+                messageStatus[msgHash] = Status.RECALL_FAILED;
```
dantaik, line 240: what's the benefit of setting the status to RECALL_FAILED?

adaki2004: > what's the benefit of setting the status to RECALL_FAILED?

So the benefit is: we will be able to reward people somehow later-on (source or dest) and can be trustless. (On-chain proveable) agains a specific messageStatus.

> Without this PR, if a message failed in recallMessage, people can try it again, now they cannot as the check here will fail:

Yes, but:
1. according to the issue description it can (and indeed) forever stuck. So people can do 100x it would always revert.
2. low chance that we every see a `recallMessage` let alone a failing recall message.

dantaik: my counter argument:
- rewarding based on failed tx history is a very centralized decision, so using events is good enough.
- recallMessage shall rarely fail, in case of failure, for example, a vault has a bug, I think people will prefer to have a chance to try it again after the bug is fixed, rather than making it fail permanetly.

adaki2004: Okay, i closing it then for now.

PR #16611, opened 2024-04-02
fix(protocol): fix signature reuse bug

dantaik:


adaki2004:
Well, it is OK, but i dont see any security concern here, because basically it is just a "delegation".

So Bob delegates to Alice to withdran on behalf of him, but with this solution it guards it with a nonce (so one-time delegation) ? Ok, maybe enhanced security, so I'm OK with it.

packages/protocol/contracts/team/TimelockTokenPool.sol
```diff
@@ -169,18 +174,25 @@ contract TimelockTokenPool is EssentialContract, EIP712Upgradeable {
...
+        _withdraw(account, _to);
     }
 
     /// @notice Gets the hash to be signed to authorize an withdrawal.
     /// @param _to The destination address.
+    /// @param _nonce The withdrawal nonce.
     /// @return The hash to be signed.
-    function getWithdrawalHash(address _to) public view returns (bytes32) {
-        return _hashTypedDataV4(keccak256(abi.encode(TYPED_HASH, _to)));
+    function getWithdrawalHash(address _to, uint256 _nonce) public view returns (bytes32) {
```
adaki2004, line 194: to be cross-chain fixed, maybe block.chainId (?).

dantaik: its part of _hashTypedDataV4 I believe.

adaki2004: > its part of _hashTypedDataV4 I believe.

Ah yeah, false alarm!

adaki2004 (approved):

Brechtpd (approved):
I guess the owner may indeed want to stop using the `to` address.

PR #16613, opened 2024-04-02
fix(protocol): check 63/64 gasleft() not smaller than specified gaslimit

dantaik:
See https://github.com/code-423n4/2024-03-taiko-findings/issues/180


packages/protocol/contracts/libs/LibAddress.sol
```diff
@@ -33,14 +35,23 @@ library LibAddress {
     {
         // Check for zero-address transactions
         if (_to == address(0)) revert ETH_TRANSFER_FAILED();
+
+        uint256 gasLimit;
+        if (_gasLimit == 0) {
+            gasLimit = gasleft();
+        } else {
+            // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md
+            if (((gasleft() * 63) >> 6) < _gasLimit) revert NOT_ENOUGH_GASLEFT();
```
Brechtpd, line 44: To be 100% correct, this will require some extra gas to be taken into account for the gas consumed between the `gasleft` call here and the actual gas value left to be sent into the call. So also the Safe contract: https://github.com/safe-global/safe-smart-account/blob/1cd7568769128717c1a6862d22fe34873d7c79c8/contracts/Safe.sol#L160

Brechtpd (commented):
Ah this discussion, I remember it still from the Loopring smart wallet days! For me this is still a UI issue because the gas limit can simply be 1/64 higher to avoid it if it would ever be a problem. But the check is valid, just depends on what you expect the smart contract to check and I think this check goes pretty far because the gas limit is already just an abstract number for users.

dantaik:
Simplified the changes:

- We only check `gasleft()*64/64 >= gasLimit` when processMessage is called by someone other than the owner. In all other cases (including recall the message), failure due to gas will not cause any issue for the msg owner.
- No extra gas addition is applied to the above check (as in the Loopring code), if after the external call, the tx failed due to out-of-gas, the whole transaction will revert so no harm will be done to the msg.owner as well.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -299,9 +300,25 @@ contract Bridge is EssentialContract, IBridge {
...
-                // Use the remaining gas if called by a the destOwner, else
-                // use the specified gas limit.
-                uint256 gasLimit = msg.sender == _message.destOwner ? gasleft() : _message.gasLimit;
+                uint256 gasLimit;
+                if (msg.sender == _message.destOwner) {
+                    // Use the remaining gas if called by a the destOwner, else
+                    // use the specified gas limit.
+                    gasLimit = gasleft();
+                } else {
+                    // he "1/64th rule" refers to the gasleft at the time the call is made. When a
```
davidtaikocha, line 309:
```suggestion
                    // The "1/64th rule" refers to the gasleft at the time the call is made. When a
```

davidtaikocha (approved):

adaki2004 (approved):

Brechtpd (approved):

davidtaikocha (approved):

PR #16616, opened 2024-04-02
chore(bridge-ui): improve dropdown ux 

Pitchanai:
- Improve the UX of the `ChainsDropdown`
  - The problem where the dropdown wouldn't disappear when the dropdown is showing has been fixed. This was done by adding a `setTimeout` to the `<ul />` callback, which made the toggle feature work properly again.


KorbinianK:
@Pitchanai What problem are you referring to? To me this looks more like a hacky workaround than a fix, but I don't even understand what's it trying to fix. Please create a detailed bug issue first.

PR #16621, opened 2024-04-03
chore(protocol): slightly improve NFT vaults

dantaik:


davidtaikocha (approved):

packages/protocol/contracts/tokenvault/BridgedERC1155.sol
```diff
@@ -50,6 +48,9 @@ contract BridgedERC1155 is EssentialContract, IERC1155MetadataURIUpgradeable, ER
         // for them instead.
         LibBridgedToken.validateInputs(_srcToken, _srcChainId);
         __Essential_init(_owner, _addressManager);
+
+        // The token URI here is not important as the client will have to read the URI from the
+        // canonical contract to fatch meta data.
```
adaki2004, line 53:
```suggestion
        // canonical contract to fetch meta data.
```

adaki2004 (approved):

adaki2004:
@dantaik you commented here: https://github.com/code-423n4/2024-03-taiko-findings/issues/215 but then you removed the metadata component.
Currently there is no `tokenURI` set at all n BridgedERC1155. (Since the base is not set).
I'd say we should use the same way as ERC721, and this is also the same way as the issue auther (link above) suggests.

dantaik:
> @dantaik you commented here: [code-423n4/2024-03-taiko-findings#215](https://github.com/code-423n4/2024-03-taiko-findings/issues/215) but then you removed the metadata component. Currently there is no `tokenURI` set at all n BridgedERC1155. (Since the base is not set). I'd say we should use the same way as ERC721, and this is also the same way as the issue auther (link above) suggests.

I commented on your PR. I'm not sure we need the ID to be appended to the URL. I think sometime ago I commented on @Brechtpd 's PR who added the URL. @Brechtpd what's your take on this bug report?

PR #16625, opened 2024-04-03
chore(protocol): change SGX INSTANCE_EXPIRY to 365 days

dantaik:
The idea is still "buy us time" in case we are very busy with other stuff.

@Brechtpd do you think 365 days is acceptable?


Brechtpd (approved):
I guess not really that big of a jump because already very long. If some vulnerability is known, we'll probably have to switch SGX verifier contracts and start from scratch, so I think no difference there directly when using these long time frames.

dionysuzx (approved):

PR #16626, opened 2024-04-03
fix(protocol): add uri override

adaki2004:
As per code4rena finding too, and hance we dont have any tokenURI information (or even how to find it) for 1155, reusing the erc721 way of consturcting the URI, this way any 3rd party dApps could know where to find the canonical one.

### DIFF
--- a/packages/protocol/contracts/tokenvault/BridgedERC1155.sol
+++ b/packages/protocol/contracts/tokenvault/BridgedERC1155.sol
@@ -2,15 +2,13 @@
 pragma solidity 0.8.24;
 
 import "@openzeppelin/contracts-upgradeable/token/ERC1155/ERC1155Upgradeable.sol";
-import
-    "@openzeppelin/contracts-upgradeable/token/ERC1155/extensions/IERC1155MetadataURIUpgradeable.sol";
 import "../common/EssentialContract.sol";
 import "./LibBridgedToken.sol";
 
 /// @title BridgedERC1155
 /// @notice Contract for bridging ERC1155 tokens across different chains.
 /// @custom:security-contact security@taiko.xyz
-contract BridgedERC1155 is EssentialContract, IERC1155MetadataURIUpgradeable, ERC1155Upgradeable {
+contract BridgedERC1155 is EssentialContract, ERC1155Upgradeable {
     /// @notice Address of the source token contract.
     address public srcToken;
 
@@ -50,6 +48,9 @@ contract BridgedERC1155 is EssentialContract, IERC1155MetadataURIUpgradeable, ER
         // for them instead.
         LibBridgedToken.validateInputs(_srcToken, _srcChainId);
         __Essential_init(_owner, _addressManager);
+
+        // The token URI here is not important as the client will have to read the URI from the
+        // canonical contract to fatch meta data.
         __ERC1155_init(LibBridgedToken.buildURI(_srcToken, _srcChainId));
 
         srcToken = _srcToken;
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/tokenvault/BridgedERC1155.sol
```diff
@@ -122,6 +122,27 @@ contract BridgedERC1155 is EssentialContract, ERC1155Upgradeable {
...
+    /// @return The source token's address.
+    /// @return The source token's chain ID.
+    function source() public view returns (address, uint256) {
+        return (srcToken, srcChainId);
+    }
+
+    /// @notice Returns the token URI.
+    /// @param _tokenId The token id.
+    /// @return The token URI following EIP-681.
+    function uri(uint256 _tokenId) public view override returns (string memory) {
```
dantaik, line 135: I understand this differently: missing _tokenId in the URL seems fine based on the 1155 standard, as long as the app knows the value, it can replace {id}.

adaki2004: But in our contract there is nothing which set the _uri in `ERC1155Upgradeable`, so if we leave it as it, it will always return "" (empty string).

packages/protocol/contracts/tokenvault/BridgedERC1155.sol
```diff
@@ -122,6 +122,27 @@
...
+        // https://github.com/crytic/slither/wiki/Detector-Documentation#abi-encodePacked-collision
+        // The abi.encodePacked() call below takes multiple dynamic arguments. This is known and
+        // considered acceptable in terms of risk.
+        return string(
+            abi.encodePacked(
+                LibBridgedToken.buildURI(srcToken, srcChainId), Strings.toString(_tokenId)
+            )
+        );
+    }
```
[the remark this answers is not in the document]
Brechtpd: You mean the ABI encodePacked Collision  alert? Should be fine, the code is only used to concatenate strings so no security impact or any way to manipulate in a meaningful way.

Brechtpd (dismissed):

packages/protocol/contracts/tokenvault/BridgedERC1155.sol
dantaik: line 54 does it `__ERC1155_init(LibBridgedToken.buildURI(_srcToken, _srcChainId));`

Inside ERC1155Upgradeable.sol, there is:

```
  function __ERC1155_init(string memory uri_) internal onlyInitializing {
        __ERC1155_init_unchained(uri_);
    }

    function __ERC1155_init_unchained(string memory uri_) internal onlyInitializing {
        _setURI(uri_);
    }
    
    function _setURI(string memory newuri) internal virtual {
        _uri = newuri;
    }
    
     function uri(uint256) public view virtual override returns (string memory) {
        return _uri;
    }
```

adaki2004: Right, i missed that buildURI, but still, it is incorrect for ERC1155.

To conform the erc1155, you should return literally `{id}.json`.

Please see it here, i did this contract, query the uri() function:
https://polygonscan.com/token/0x88883145049dec03c00cb7708cbc938058802520#readContract

adaki2004: But i'm not sure if we shall add that at this point then.. Because we using another standard to build this URI.. If we add `{id}.json` -> it might have no effect, since the dApps could easily query this information cross-chain.

adaki2004:
Closing as the standard we using to build the URI is a clear path how to get the original tokenURI from the canonical chain.

PR #16634, opened 2024-04-03
feat(protocol): remove ETH deposit fees

dantaik:
This is based on a feedback from an auditor - he is not a fun of the ETH deposit fee design. 

How do you guys like the idea of completely remove the fee and lift the min ETH deposit amount?

BTW, I think these changes are backward compatible so Hekla doesn't need to regenesis. @Brechtpd @davidtaikocha

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -40,10 +40,6 @@ library TaikoData {
         uint64 ethDepositMaxCountPerBlock;
         // The minimum amount of ETH required for a deposit.
         uint96 ethDepositMinAmount;
-        // The maximum amount of ETH allowed for a deposit.
-        uint96 ethDepositMaxAmount;
-        // The gas cost for processing an ETH deposit.
-        uint256 ethDepositGas;
         // The max number of L2 blocks that can stay unsynced on L1 (a value of zero disables
         // syncing)
         uint8 blockSyncThreshold;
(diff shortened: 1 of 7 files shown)


packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -35,17 +35,11 @@ library TaikoData {
         // The size of the ETH deposit ring buffer.
         uint256 ethDepositRingBufferSize;
         // The minimum number of ETH deposits allowed per block.
-        uint64 ethDepositMinCountPerBlock;
+        uint64 ethDepositsPerBlock;
```
adaki2004, line 38: I think we lose backward compatiblity with Hekla with the removal of these:
`ethDepositMaxAmount`, `ethDepositGas`, `ethDepositMaxFee`

because the `blockSyncThreshold` position within the struct slots will be changed, and we use that in LibVerifying.sol:

`if (_lastVerifiedBlockId > lastSyncedBlock + _config.blockSyncThreshold) `

Brechtpd: If all smart contracts are updated should be good I guess because will point to the correct location?

Brechtpd (approved):

packages/protocol/contracts/L1/TaikoData.sol
adaki2004: Ah ok, since `memory` object, right !

adaki2004 (approved):

packages/protocol/contracts/L1/TaikoData.sol
dantaik: @Brechtpd, @adaki2004 do you prefer the entire EthDeposit feature is removed instead? Now we have our Bridge and it seems nobody used the EthDeposit feature, and the current implementation of this feature also assume the existence of the Bridge.

dantaik:
Closed as I'm in favor of https://github.com/taikoxyz/taiko-mono/pull/16638

PR #16641, opened 2024-04-04
fix(protocol): use signature check to verify if msg.sender is EOA

dantaik:
After the next hardfork, tx.origin will always be ms.sender with [EIP 7645](https://ethereum-magicians.org/t/eip-7645-alias-origin-to-sender/19047/4)

@davidtaikocha This change may break compatibility.


davidtaikocha (approved):

davidtaikocha (approved):

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -55,7 +55,7 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
         LibVerifying.init(state, getConfig(), _genesisBlockHash);
     }
 
-    function init2() external reinitializer(2) {
+    function init2() external onlyOwner reinitializer(2) {
```
adaki2004, line 58: As effectively this `reinitializer` is fully harmless, and that `onlyOwner` might be `Timelock` (?), i'm not sure if we need to guard `onlyOwner`, what do you think ?

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -198,7 +198,8 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
             // Taiko blocks proposed per Ethereum block is smaller than 1.
             blockMaxGasLimit: 240_000_000,
             livenessBond: 250e18, // 250 Taiko token
-            blockSyncThreshold: 16
+            blockSyncThreshold: 16,
+            onlyEOACanUseCalldataForDA: true
```
adaki2004, line 202: We using this flag for not breaking tests ?
Bc. I think this flag is not really needed because we need EOAs now too. (even tho different check what we using now:  `msg.sender != tx.origin`)

dantaik: > We using this flag for not breaking tests ?
Yes

dantaik: > Bc. I think this flag is not really needed because we need EOAs now too. (even tho different check what we using now: msg.sender != tx.origin)

Didn't get your point, please elaborate a bit.

packages/protocol/contracts/L1/TaikoL1.sol
dantaik: In this case, onlyOwner is not really necessary, but I think maybe it's a good practice to ensure reinitializers are only called by owners so the owner controls the timing, which is probably important in case the protocol needs to be upgraded first, then the node is updated, the a feature is activated, then the reinitializer shall be called.

packages/protocol/contracts/L1/TaikoL1.sol
adaki2004: I meant that basically even now (pre-Petra) we dont want contracts to be the proposers, because this way, retrieving calldata back is hard (from client perspective).
So in testnet/mainnet environment this flag will always has to be true.

adaki2004: If it is only needed to distinguish between test suite (foundry test) vs. runnig testnet/mainnet environment, maybe we can reuse the `livenessBond == 1 TKO` -> which is 1 in our `TaikoL1LibProvingWithTiers.t.sol `and `TaikoL1.t.sol` and remove this flag completely. What do you think ?

dantaik: Using `livenessBond == 1` is like a hidden rule, using a dedicated flag makes it more explicit.

Besides, we may want to skip EOA check and set livenessBond to a value other than 1, then the trick will stop working.

dantaik: Fixed it differently, please see the new commit.

adaki2004 (approved):

PR #16647, opened 2024-04-04
chore(protocol):  apply small improvements

dantaik:
See https://github.com/code-423n4/2024-03-taiko-findings/issues/255


packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
```diff
@@ -1,129 +0,0 @@
-// SPDX-License-Identifier: MIT
-pragma solidity 0.8.24;
-
-import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
dantaik, line 4: deleted because we probably won't need it, plus there is this bug: https://github.com/code-423n4/2024-03-taiko-findings/blob/main/data/Shield-Q.md#l-23-user-can-not-claim-the-same-amount-twice-using-the-erc20airdropclaimfunction

adaki2004 (approved):
Approved, but still think ERC20Airdrop2 is usable in Trailblazer hence it smoothens down the sell pressure with that gradual unlock. But eventually we can get it back, if needed.

dantaik:
> Approved, but still think ERC20Airdrop2 is usable in Trailblazer hence it smoothens down the sell pressure with that gradual unlock. But eventually we can get it back, if needed.

Lets try to fix the bug then.

davidtaikocha (approved):

PR #16654, opened 2024-04-05
feat(protocol): add `lastSyncedBlockId ` for L2 DAO vote aggregation

dantaik:
This small feature is added to support future DAO. The DAO will aggregate votes/vetos on L2 which will be bridged to L1. Suppose a DAO proposal has passed its voting period on L1. We need to wait for L2-to-L1 synchronization to happen for at least once before the proposal's state can be finalized. If for some reason L2 blocks is no longer verified, then all DAO proposals will have to wait.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -135,7 +135,7 @@ library TaikoData {
     struct SlotA {
         uint64 genesisHeight;
         uint64 genesisTimestamp;
-        uint64 lastSyncedBlocKId;
+        uint64 lastSyncedBlockId;
         uint64 __reservedA2;
     }
 
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -134,7 +134,7 @@ library TaikoData {
     struct SlotA {
         uint64 genesisHeight;
         uint64 genesisTimestamp;
-        uint64 __reservedA1;
+        uint64 lastSyncedAt;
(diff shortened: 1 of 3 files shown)


Brechtpd:
I think this will check that **some** L2 block has synced after the deadline, but it could be a very old L2 block, especially if the block verification is behind (currently always at least 1 day behind I guess). But I think the intention is that all L2 blocks with timestamp < dao.voteDeadlineOnL1 are actually synced so all votes/vetos that happened on L2 before that timestamp are also taken into account? Maybe you could some extra 1 day or something on the DAO side to take this into account, but without checking the L2 block timestamp not sure if you could do it in an accurate way.

dantaik:
> I think this will check that **some** L2 block has synced after the deadline, but it could be a very old L2 block, especially if the block verification is behind (currently always at least 1 day behind I guess). But I think the intention is that all L2 blocks with timestamp < dao.voteDeadlineOnL1 are actually synced so all votes/vetos that happened on L2 before that timestamp are also taken into account? Maybe you could some extra 1 day or something on the DAO side to take this into account, but without checking the L2 block timestamp not sure if you could do it in an accurate way.

Good feedback. I change to use `lastSyncedBlockId` so with we can change the way to verify if a proposal's status can be finalized as:

```
taikoL2.getBlock(lastSyncedBlockId).proposedAt >  dao.voteDeadlineOnL1 &&
taikoL2.getBlock(lastSyncedBlockId).proposedAt <=  dao.voteDeadlineOnL1 + 2days
```

davidtaikocha (approved):

smtmfft (approved):

PR #16659, opened 2024-04-05
fix(protocol): check no loops in multi-hop in Bridge

dantaik:
This PR mitigates the concerns in this [issue](https://github.com/code-423n4/2024-03-taiko-findings/issues/125) , but I'm having a hard time to construct an exploiting attack transaction with a bridge loop that has security impact.

### DIFF
--- a/packages/protocol/contracts/signal/SignalService.sol
+++ b/packages/protocol/contracts/signal/SignalService.sol
@@ -21,6 +21,16 @@ contract SignalService is EssentialContract, ISignalService {
 
     uint256[48] private __gap;
 
+    struct CacheAction {
+        bytes32 rootHash;
+        bytes32 signalRoot;
+        uint64 chainId;
+        uint64 blockId;
+        bool isFullProof;
+        bool isLastHop;
+        CacheOption option;
+    }
+
     error SS_EMPTY_PROOF();
     error SS_INVALID_HOPS_WITH_LOOP();
     error SS_INVALID_SENDER();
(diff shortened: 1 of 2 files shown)


dantaik:
> Thinking thru, i think it is possible to embeded signals, but i dont think there is a security concern here because of 2 reasons:
> 
> * In order to be malicious you need to somewhat "mock" a call to `SignalService` on behalf of the `Bridge` -> otherwise the `sendSignal()` will use a different slot `_sendSignal(`**msg.sender**`, _signal, _signal)`. So you definitely can call `SignalService` `sendSignal`, but later during verification the slot is always used with one of the Bridge contracts (so proof will be invalid).
> * For app we always use our own `address signalService = resolve(chainId, "signal_service", false);`
> 
> So given the 2 factors, i dont think a self-referential hop is safety critical.

I tend to agree, but TBH, preventing loops makes me feel more comfortable given the cost is really small.

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -148,23 +160,35 @@ contract SignalService is EssentialContract, ISignalService {
         HopProof[] memory hopProofs = abi.decode(_proof, (HopProof[]));
         if (hopProofs.length == 0) revert SS_EMPTY_PROOF();
 
+        uint256 lenLessOne;
+        unchecked {
+            lenLessOne = hopProofs.length - 1;
+        }
+        uint64[] memory trace = new uint64[](lenLessOne);
+
```
adaki2004, line 168: Here (`verifySignalReceived`) i think there is no need to check these things as this is a helper view function to the relayer and not used for proving.

dantaik: good point

dantaik: I think verifySignalReceived should behavice the same so if it returns true, then proveSignalReceived should succeed.

Brechtpd (approved):

adaki2004 (approved):

PR #16660, opened 2024-04-05
fix(protocol): check invocation gas limit also in `retryMessage`

adaki2004:
https://github.com/code-423n4/2024-03-taiko-findings/issues/177

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -353,7 +353,7 @@ contract Bridge is EssentialContract, IBridge {
         }
 
         // We check _message.gasLimit > gasleft() to make sure we not only need to bridge invocation
-        // call to succeeed, we also need it to succeed with a gas limit no smaller than the
+        // call to succeed, we also need it to succeed with a gas limit no smaller than the
         // message's gasLimit.
         if (
             (
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -355,11 +355,11 @@ contract Bridge is EssentialContract, IBridge {
         // We check _message.gasLimit > gasleft() to make sure we not only need to bridge invocation
         // call to succeeed, we also need it to succeed with a gas limit no smaller than the
         // message's gasLimit.
-        // Also if caller is not the destOwner, we need to be sure there is enough gas for rest of
-        // the code to finish execution.
         if (
-            (_message.gasLimit != 0 && _message.gasLimit > gasleft())
-                || (msg.sender != _message.destOwner && _message.gasLimit > (gasleft() * 63) >> 6)
+            (
+                _message.gasLimit != 0 && msg.sender != _message.destOwner
+                    && _message.gasLimit > (gasleft() * 63) >> 6
+            )
         ) {
             revert B_NOT_ENOUGH_GASLEFT();
         }
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -353,7 +353,7 @@ contract Bridge is EssentialContract, IBridge {
         }
 
         // We check _message.gasLimit > gasleft() to make sure we not only need to bridge invocation
-        // call to succeeed, we also need it to succeed with a gas limit no smaller than the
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -352,6 +352,13 @@ contract Bridge is EssentialContract, IBridge {
             revert B_NON_RETRIABLE();
         }
 
+        // We check _message.gasLimit > gasleft() to make sure we not only need to bridge invocation
+        // call to succeeed, we also need it to succeed with a gas limit no smaller than the
+        // message's gasLimit.
+        if (_message.gasLimit != 0 && _message.gasLimit > gasleft()) {
```
dantaik, line 358: Dani, I think the 1/64 also apply here. Assuming `gasleft() == gasLimit`, then only `63/64 * gasleft()` can be forward to the destination contract.

So maybe we check `_message.gasLimit > (63* gasleft())>> 6)`

dantaik: If this make sense, maybe we extract a small function for the 1/64 rule.

adaki2004: Okay, i can do, but there is a reason i did not put it there. Because simple if that is the case and code reverts because there is no more gas to execute thre resot of the functions (emit event, etc.) then there is no effect, right ? Simply people just retry with higher gas. Not really safety critical and we dont expect that to happen (too often) ?

dantaik: if gasleft() == gasLimit, then 63*gasLimit/64 gas will be forwarded to the dest app to call the function, but we do not want this call to succeed.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -352,6 +352,18 @@ contract Bridge is EssentialContract, IBridge {
             revert B_NON_RETRIABLE();
         }
 
+        // We check _message.gasLimit > gasleft() to make sure we not only need to bridge invocation
+        // call to succeeed, we also need it to succeed with a gas limit no smaller than the
+        // message's gasLimit.
+        // Also if caller is not the destOwner, we need to be sure there is enough gas for rest of
+        // the code to finish execution.
+        if (
+            (_message.gasLimit != 0 && _message.gasLimit > gasleft())
```
dantaik, line 361: if gasLimit == 0, msg.sender must be the destOwner, so the checks here an be simplified:

packages/protocol/contracts/bridge/Bridge.sol
adaki2004: added the check. (Also not created a separate function but inlined, since it is a lil bit cheaper)

packages/protocol/contracts/bridge/Bridge.sol
dantaik:
```
if (_message.gasLimit != 0 &&
msg.sender != _message.destOwner &&
_message.gasLimit > (gasleft() * 63) >> 6) {
```

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -352,6 +352,18 @@ contract Bridge is EssentialContract, IBridge {
             revert B_NON_RETRIABLE();
         }
 
+        // We check _message.gasLimit > gasleft() to make sure we not only need to bridge invocation
+        // call to succeeed, we also need it to succeed with a gas limit no smaller than the
```
dantaik, line 356:
```suggestion
        // call to succeed, we also need it to succeed with a gas limit no smaller than the
```

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -352,6 +352,18 @@ contract Bridge is EssentialContract, IBridge {
             revert B_NON_RETRIABLE();
         }
 
+        // We check _message.gasLimit > gasleft() to make sure we not only need to bridge invocation
+        // call to succeed, we also need it to succeed with a gas limit no smaller than the
+        // message's gasLimit.
+        if (
+            (
```
dantaik, line 359: please change "((...))" to "(...)"

dantaik (approved):

dantaik (approved):

Brechtpd (approved):

PR #16665, opened 2024-04-06
fix(protocol): fix potential 1271 signature replay if proposers are smart contracts

dantaik:
Screenshot 2024-04-06 at 11 29 44

https://mirror.xyz/curiousapple.eth/pFqAdW2LiJ-6S4sg_u1z08k4vK6BCJ33LcyXpnNb8yU


packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -143,33 +145,45 @@ contract AssignmentHook is EssentialContract, IHook {
...
                 _assignment.tierFees
             )
         );
+
+        return keccak256(
+            abi.encodePacked(
+                "PROVER_ASSIGNMENT",
+                ITaikoL1(_taikoL1Address).getConfig().chainId,
+                _taikoL1Address,
+                _blockProposer,
```
adaki2004, line 180: I see the recommendation from the bottom of the screenshot (currently cannot access the audit) to include the `_blockProposer` in the hash, but it is already part of it already indirectly via metaHash (since it is in BlockMetadata.sender) no ?

dantaik: Not really, since `assignment.metaHash` can be 0 if the assigned prover doesn't want to check metahash.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -143,33 +145,45 @@ contract AssignmentHook is EssentialContract, IHook {
...
+        address _blockProposer,
+        address _assignedProver,
         bytes32 _blobHash
     )
         public
         view
         returns (bytes32)
     {
-        return keccak256(
+        bytes32 hash = keccak256(
```
Brechtpd, line 163: split up into two parts because of stack too deep?

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -143,33 +145,45 @@ contract AssignmentHook is EssentialContract, IHook {
...
         );
+
+        return keccak256(
+            abi.encodePacked(
+                "PROVER_ASSIGNMENT",
+                ITaikoL1(_taikoL1Address).getConfig().chainId,
+                _taikoL1Address,
+                _blockProposer,
+                _assignedProver,
+                address(this),
```
Brechtpd, line 182: a dumb thing but would move `address(this)` to below `_taikoL1Address` again because it's constant data, the rest is non-constant.

Brechtpd (approved):

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
dantaik: Yes

adaki2004 (approved):

PR #16666, opened 2024-04-06
feat(protocol): use 35000 as gas limit for sending Ether in Brdge

dantaik:
This is based on OZ's feedback:

Screenshot 2024-04-06 at 11 46 33

BTW, according to ChatGPT:

> When you send Ether to an EOA with a zero balance, you are changing the state from zero to a non-zero balance, which has a higher gas cost due to the additional work required to update the state. The gas cost for sending Ether to an account with a zero balance is 21,000 gas for the transaction itself, plus an additional 20,000 gas for changing the state from zero to non-zero, totaling 41,000 gas.

@Brechtpd do you think 60000 is a good number?


Brechtpd:
image

As far as I understand, the cost for most ETH transfer related costs is charged on the call itself, not inside the call. So the gas passed into I think could be set much lower. There's also already an automatic stipend of 2300 gas.

My feeling is that 60k gas is quite high, on the other hand it's also quite critical that people can count on the refund to work even if the refund address is a smart contract that may contain some fancy logic. Taking that into account I think 60k is reasonable.

I think we shouldn't make it a general gas limit though for all ETH transfers everywhere, if a limit makes sense for the refund I think we should just specify a gas limit there.

Brechtpd:
Adding this limit to the refund ETH transfer I think makes sense because something the relayer does have to pay the cost for so a risk (depending on how sophisticated the relayer is). I'm just not sure about adding this limit to all ETH transfers because those transfers mostly do not have the same concern. But I'm not against it to keep things simple and uniform.

Brechtpd (approved):

adaki2004 (approved):

PR #16680, opened 2024-04-06
fix(bridge-ui): fix ERC721 and ERC1155 detection in NFT bridge

ricecodekhmer:
I wanted to try out the NFT bridging feature, so I created my own NFT, which is an ERC-721 contract on holesky, but I encountered an issue where the 'Scan for NFT' button wasn't working as expected. So, attempting to use the "Add Manually" option, I encountered an error message saying "You must be the owner of all tokens" when entering the address and token id for my NFT.

After setting up the bridge locally and looking at the code, I came to the implementation of the `isERC721` function in `packages/bridge-ui/src/libs/token/detectContractType.ts`, and I realised that the `readContract` call would always throw an exception if the specified token id didn't match the owner's token id, regardless of whether the contract was ERC-721 or not.

To fix this, I have altered both `isERC721` and `isERC1155` functions to properly check things by using the standard identifiers (`0xd9b67a26` for [ERC-1155](https://eips.ethereum.org/EIPS/eip-1155) and `0x80ac58cd` for [ERC-721](https://eips.ethereum.org/EIPS/eip-721)) which adhere to the EIP specifications for the contracts. 

# Screenshots 
## Before

Screenshot from 2024-04-06 20-40-29

## After

Screenshot from 2024-04-06 20-42-05

## Testing

I made sure to test with ERC-20, ERC-1155 and ERC-721 to check if all contracts are identified properly.


packages/bridge-ui/src/libs/token/detectContractType.ts
```diff
@@ -13,34 +13,43 @@ const log = getLogger('detectContractType');
...
       chainId,
     });
-    return true;
+
+    if (isERC721) {
+      return true;
+    }
   } catch (err) {
     // we expect this error to be thrown if the token is a ERC721 and the tokenId is invalid
     return (err as Error)?.message?.includes('ERC721: invalid token ID') ?? false;
```
KorbinianK, line 29: This check is not needed anymore.

KorbinianK (changes requested):

KorbinianK:
@ricecodekhmer Thanks for this enhancement, been meaning to change it but tbh. it worked, `supportsInterface` is just checking it on contract level and requires more than a single function, so certainly the correct approach.

**But:** I am curious how this changes your ownership check. If the contract type check reverts or fails, you get `This is not a valid NFT address on the current network` not the one in your screenshot. 
That one should only appear if `ownerOf(tokenID)` fails. Can you send me your verified contract address so I can have a look?


Regardless of this, please fix the comment and adjust the unit tests if you want to have this merged :)

KorbinianK (approved):

davidtaikocha (approved):

PR #16681, opened 2024-04-06
chore(protocol): small fixes in documentation

adaki2004:
Still cannot access the audit in defender but Brecht exported me these:

```
N-07 Missing or Misleading Documentation

Status: unresolved

Severity: Note

Last updated: 18 hours ago

The following opportunities to improve the clarity of the documentation were found:

    The structs defined in the TaikoData library would benefit from some added documentation on what the different fields represent.
    The EthDeposit struct is incorrectly documented as using 1 slot but actually uses 2 slots as it is 40 bytes long.
    meta.coinbase is documented as being the L2 block proposer, but it is only chosen by the block proposer.
    In the proveBlock function, a comment mentions that if the transition does not exist the tid will be set to 0. However, it is set to 1 in practice.
    A conditional branch for the TIER_OP tier is present in the code but could be documented as only being used for testnet.
    The contest bond is documented as being burned from the prover, but it is only transferred and can in fact be regained if the transition is proven to be invalid.
    The _overrideWithHigherProof function would benefit from better documentation overall (e.g., when ts.contester == 0 for the first transition, there is technically no existing contest but the code still uses this branch).
    The NatSpec around the getBasefee function indicates that the function can be used to "get the basefee and gas excess [...]". However, the function only returns the base fee.
    The NatSpec around the __ctx variable in the Bridge contract indicates that it fits in 3 slots, whereas it only occupies 2.
    The NatSpec around the proveMessageReceived function is incorrect as it was duplicated from the function above it.
    The NatSpec around the getInvocationDelays function has an extra "and" in the phrase "a message can be executed since and the time it was received".
    The processDeposits function could use some additional documentation (e.g., around the fee logic).
    The Message struct in the IBridge interface could document that the gasLimit of a message can be set to 0 so that only the destOwner can process the message on the destination chain. Additionally, the gasLimit could be documented as not being respected on message retries or when processed by the owner.
    A comment around the gasLimit is reversed, as the remaining gas is used if called by the owner.

Consider addressing the above instances to improve the clarity and readability of the codebase.
```


dantaik (approved):

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -37,22 +37,25 @@ library TaikoData {
...
     }
 
-    /// @dev Struct representing prover assignment
+    /// @dev Struct representing prover fees per given tier
     struct TierFee {
         uint16 tier;
         uint128 fee;
     }
 
+    /// @dev Each tier id with it's own proof
```
RogerLamTd, line 46:
```suggestion
    /// @dev A proof and the tier of proof it belongs to
```

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -19,7 +19,9 @@ library LibProving {
     /// @notice Keccak hash of the string "RETURN_LIVENESS_BOND".
     bytes32 public constant RETURN_LIVENESS_BOND = keccak256("RETURN_LIVENESS_BOND");
 
-    /// @notice The tier name for optimistic proofs.
+    /// @notice The tier name for optimistic proofs - expected to only present for testnets. For
```
RogerLamTd, line 22:
```suggestion
    /// @notice The tier name for optimistic proofs - expected to only being used for testnets. For
```

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -354,7 +356,8 @@ library LibProving {
         }
     }
 
-    /// @dev Handles what happens when there is a higher proof incoming
+    /// @dev Handles what happens when either the first transition is being proven or there is a
+    /// higher proof incoming
```
RogerLamTd, line 360:
```suggestion
    /// higher tier proof incoming
```

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -506,8 +506,7 @@ contract Bridge is EssentialContract, IBridge {
     /// @notice Returns invocation delay values.
     /// @dev Bridge contract deployed on L1 shall use a non-zero value for better
     /// security.
-    /// @return The minimal delay in second before a message can be executed since and the time it
-    /// was received on the this chain.
+    /// @return The minimal delay in second between message execution and proving.
```
RogerLamTd, line 509:
```suggestion
    /// @return The minimal delay in seconds between message execution and proving.
```

RogerLamTd (approved):

packages/protocol/contracts/bridge/IBridge.sol
```diff
@@ -38,7 +38,8 @@ interface IBridge {
         // Processing fee for the relayer.
         uint256 fee;
         // gasLimit to invoke on the destination chain. If this value is zero, only destOwner can
-        // process the message.
+        // process the message. this value will not be respected on retries or
```
Brechtpd, line 41: I guess with the latest change it will actually mostly be respected when retrying (at least as gas > gas_limit, not as an exact limit)

Brechtpd (approved):

packages/protocol/contracts/bridge/IBridge.sol
RogerLamTd: Should we phrase it as "This value will mostly be respected when retrying (gas > gas_limit) but not when processed by the owner" then?

dantaik (approved):

PR #16684, opened 2024-04-07
fix(protocol): fix liveness bond related tokenomics issue

dantaik:
BEGIN_COMMIT_OVERRIDE
fix(protocol): return liveness bond only to assigned prover
feat(protocol): allow assigned prover to prove blocks outside proving window (liveness bond not returned)
feat(protocol): remove `contestations` from `TransitionState` and events (it's buggy)
feat(protocol): add `TaikoL1.getTransition(blockId, transitionID)` function
refactor(protocol): use a `Local` struct in LibProving to cache state variables and get around of stack-too-deep error
test(protocol): add more tests to verify bonds are handled correctly
END_COMMIT_OVERRIDE

---

## The liveness bug
The bug was identified by([OpenZeppelin](https://defender.openzeppelin.com/v2/#/audit/306ca593-6cb7-4b7f-92d4-ba1e41ac9e56/issues/L-06)) and [Code4rena](https://github.com/code-423n4/2024-03-taiko-findings/issues/227).

### Scenario (quoted from the bug report)

- Henry is assigned to prove a block, and sends 250 TKO as liveness bond to the TaikoL1 contract. He correctly proves the state transition with an SGX proof, providing another 250 TKO as validity bond.
- Mallory contests the state transition, providing 500 TKO as contest bond to the TaikoL1 contract.
- In the same transaction, Mallory re-proves the block and confirms Henry's transition using a zk proof. In doing so, Henry earns 1/4 * contestBond = 125 TKO, and gets his original validity bond of 250 TKO back. Mallory gets 1/4 * contestBond = 125 TKO as she is the new prover. She provides a new validity bond of 500 TKO.
- The block is verified, and Mallory as the final prover gets back half of the liveness bond as well as her validity bond, 1/2*250 + 500 = 625 TKO in total.

As a result of the above, Henry has lost his initial liveness bond but got 1/4th of the contest bond, a net loss of 125 TKO. Mallory has lost 3/4th of her contest bond but got back half of the liveness bond, a net loss of 250 TKO.

### Result
- Henry incurs a net loss of 125 TKO, losing the liveness bond but gaining a quarter of the contest bond.
- Mallory incurs a net loss of 250 TKO, losing three-quarters of her contest bond but regaining half of the liveness bond.

### Fix

The bug was in the handling of the liveness bond, which was returned to `ts.prover` instead of `blk.assignedProver`. The liveness bond was returned upon block verification, and only if the assigned prover created and used the first transition to finalize the block.

Now, the liveness bond is returned to the assigned prover immediately upon the creation of the first transition, provided it's done before the proving window expires. Otherwise, `ts.livenessBond` is set to `0`.  `test_taikoL1_group_2_case_1` covers the scenario above.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -103,7 +103,6 @@ library TaikoData {
         uint96 contestBond;
         uint64 timestamp; // slot 6 (90 bits)
         uint16 tier;
-        uint8 contestations;
     }
 
     /// @dev Struct containing data required for verifying a block.
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -103,6 +103,7 @@ library TaikoData {
         uint96 contestBond;
         uint64 timestamp; // slot 6 (90 bits)
         uint16 tier;
+        uint8 __reserved1;
     }
 
     /// @dev Struct containing data required for verifying a block.
--- a/packages/protocol/contracts/L1/libs/LibProving.sol
+++ b/packages/protocol/contracts/L1/libs/LibProving.sol
@@ -306,6 +306,7 @@ library LibProving {
             ts_.contestBond = 1; // to save gas
             ts_.timestamp = _blk.proposedAt;
             ts_.tier = 0;
+            ts_.__reserved1 = 0;
 
             if (tid_ == 1) {
                 // This approach serves as a cost-saving technique for the
--- a/packages/protocol/contracts/L1/libs/LibProving.sol
+++ b/packages/protocol/contracts/L1/libs/LibProving.sol
@@ -371,7 +371,6 @@ library LibProving {
     {
         // Higher tier proof overwriting lower tier proof
         uint256 reward; // reward to the new (current) prover
-        uint96 livenessBondToValidityBond;
 
         if (_ts.contester != address(0)) {
             if (_sameTransition) {
@@ -397,7 +396,9 @@ library LibProving {
             uint96 livenessBond = _blk.livenessBond;
             if (livenessBond != 0) {
                 if (_blk.assignedProver == msg.sender) {
-                    livenessBondToValidityBond = livenessBond;
+                    unchecked {
+                        reward += livenessBond;
+                    }
                 }
                 _blk.livenessBond = 0;
             }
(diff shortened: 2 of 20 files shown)


packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -103,7 +103,6 @@ library TaikoData {
         uint96 contestBond;
         uint64 timestamp; // slot 6 (90 bits)
         uint16 tier;
-        uint8 contestations;
```
dantaik, line 106: @davidtaikocha notice this change and let me know if it's a problem for node/client.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -393,21 +389,29 @@ library LibProving {
...
             if (_sameTransition) revert L1_ALREADY_PROVED();
+
             // Contest the existing transition and prove it to be invalid. The new prover get all
             // rewards.
             reward = _rewardAfterFriction(_ts.validityBond);
-            _ts.contestations += 1;
+
+            uint96 livenessBond = _blk.livenessBond;
+            if (livenessBond != 0) {
+                if (_blk.assignedProver == msg.sender) {
```
Brechtpd, line 399: Could you explain this if statement? I don't think I follow. I'm not sure why or how the assigned prover would be `msg.sender` here (also together with some limitations on assignedProver in `_checkProverPermission`).

dantaik: I guess you asked the question mainly due to the incorrect comment in this section (updated).

When the block is proposed, `blk.livenessBond` is nonzero; if the assigned prover managed to prove the block in the proving window, it also enters this `_overrideWithHigherProof`'s second if-statement, then `msg.sender (the prover) == blk.assignedProver` will be true; if another prover proves the block out of the proving window, the if-statement will be false.

Brechtpd (approved):

packages/protocol/contracts/L1/TaikoData.sol
davidtaikocha: Its ok on client side

davidtaikocha (approved):

dantaik:
@Brechtpd please take another look at the last commit - suggested by David C.

davidtaikocha (approved):

adaki2004 (approved):
Changes LGTM, looks it has changed a lot but `local` variable is causing most of the diffs.
A friendly reminder that this code (especially contestation and it's bond) part is still complex, so best would be 
- to pay high attention when modify and review
- in case we will have reliable ZK - just switch to a one, combined tier (ZK /sg/Guardian) for the sake of quick finalization

PR #16691, opened 2024-04-08
chore(protocol): apply some optimization recommendations from OpenZeppelin

adaki2004:
As per the below issue (some left out due to being done already OR out of interest, if savings is on the guardian addiiton/deletion, since they wont be called frequently).

```
N-04 Gas Optimization

Status: unresolved

Severity: Note

Last updated: 18 hours ago

The following opportunities for gas optimizations were found:

    Resolved names such as "proposer", "tier_provider", and "taiko_token", domain separators such as "PROVER_ASSIGNMENT" or "SIGNAL", and verifier names such as "tier_sgx" could be constant bytes32 contract variables. -> "tier_sgx" is currently in the tier provider, which might be changed - I left that one out.
    During during block verification, the tko token address could be cached, similar to the tier provider.
    In the GuardianProver contract:
        The Approved and the GuardianApproval events are always emitted consecutively and have some duplicate information. Consider only emitting one.
        The indexation of the guardian could use the loop iterator i + 1 instead of guardians.length to avoid reading from storage in the loop. -> It's ok as it will be called by owner, infrequently
        The minGuardians storage variable read in the for loop during approvals could be cached to memory, similar to guardians.length.-> It's ok as it will be called by owner, infrequently
        The address[] memory _newGuardians argument could be stored in calldata to avoid copying the array to memory.
    In the sendEther function, no returned value is necessary and the allocation of 64 bytes of memory could be avoided.
    In the anchor function, the blockhash(parentId) and the state variable gasExcess could be cached to memory to save gas during the Anchored event emission. -> Done already other pr
    In the onBlockProposed function of hooks, the _blk, _meta and _data arguments could be in calldata rather than memory by default to save gas. -> Done already other PR

Consider updating the identified instances to save gas during the operation/deployment of the protocol.
```


packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -19,6 +19,11 @@ library LibProving {
     /// @notice Keccak hash of the string "RETURN_LIVENESS_BOND".
     bytes32 public constant RETURN_LIVENESS_BOND = keccak256("RETURN_LIVENESS_BOND");
 
+    /// @notice Keccak hash of the string "tier_provider".
+    bytes32 private constant TIER_PROVIDER = bytes32("tier_provider");
```
dantaik, line 39: @dantaik how we have a LibConstants.sol to make include all these constants there?

dantaik: lets only put string constants there, other constants should stay where they are.

In LibConstants (maybe LibStrings?):, we can have naming convension, 

for a string "str", the variable is named S_STR, with "str" converted to upper case with a leading "S_"?

adaki2004: We dont really have string constants, except 2: `PROVER_ASSIGMENT` or `SIGNAL`. Rest is byte32 representation of string.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -41,6 +41,11 @@ contract AssignmentHook is EssentialContract, IHook {
     /// the prover's job to make sure it can get paid within this limit.
     uint256 public constant MAX_GAS_PAYING_PROVER = 50_000;
 
+    /// @notice bytes32 representation of the string "taiko_token".
+    bytes32 private constant TKO = bytes32("taiko_token");
+    /// @notice Constant string "PROVER_ASSIGNMENT".
+    string private constant PROVER_ASSIGNMENT = "PROVER_ASSIGNMENT";
```
dantaik, line 47: how about we also convert this to bytes32?

adaki2004: We can - no issue, but we canot do with "SIGNAL", (bc. of tests)

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -19,6 +19,13 @@ library LibProposing {
     bytes32 private constant _EMPTY_ETH_DEPOSIT_HASH =
         0x569e75fc77c1a856f6daaf9e69d8a9566ca34aa47f9133711ce065a571af0cfd;
 
+    /// @notice bytes32 representation of the string "tier_provider".
+    bytes32 private constant TIER_PROVIDER = bytes32("tier_provider");
+    /// @notice bytes32 representation of the string "proposer".
+    bytes32 private constant PROPOSER = bytes32("proposer");
+    /// @notice bytes32 representation of the string "taiko_token".
+    bytes32 private constant TKO = bytes32("taiko_token");
```
dantaik, line 27: should we enforce some name convention here for **bytes32 strings**:

- name them STR_SOME_THING = "some Thing" if it is public or internal, same for "SOME THING"
- name them _STR_SOME_THING if it is private.

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
dantaik: Oh, then keep them as-is not to break upgradability then.

packages/protocol/contracts/L1/libs/LibProving.sol
dantaik: Same, same.

packages/protocol/contracts/signal/LibSignals.sol
```diff
@@ -9,4 +9,7 @@ library LibSignals {
 
     /// @notice Keccak hash of the string "SIGNAL_ROOT".
     bytes32 public constant SIGNAL_ROOT = keccak256("SIGNAL_ROOT");
+
+    /// @notice Constant string "SIGNAL".
+    string public constant SIGNAL = "SIGNAL";
```
dantaik, line 14: how about
- SIGNAL_ROOT => HASH_STR_SIGNAL_ROOT
- SIGNAL => STR_SIGNAL

packages/protocol/contracts/L1/libs/LibProposing.sol
adaki2004: OK, tho in practice they will be public (so that other Libraries could see) or if libraries inheriting from libraries it could be private, but i dont think inheritance is necessary.

packages/protocol/contracts/signal/LibSignals.sol
adaki2004: Ok, will do.

packages/protocol/contracts/common/EssentialContract.sol
```diff
@@ -21,11 +21,11 @@ abstract contract EssentialContract is UUPSUpgradeable, Ownable2StepUpgradeable,
...
         0xa5054f728453d3dbe953bdc43e4d0cb97e662ea32d7958190f3dc2da31d9721b;
 
-    /// @dev Slot 1.
-    uint8 private __reentry;
-
-    uint8 private __paused;
-
+    /// @dev Slot 1&2 - not optimized if we store them in 1 slot, because they neither read nor
+    /// written at the same time.
+    uint256 private __reentry;
```
dantaik, line 26: Dani, please revert this change as it will require us to reset the current testnet.
We are not optimizing, but it's still OK to have them use 1 slot.

adaki2004: Ok, will do !

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: I think internal should be good enough as these library will all be inlined to TaikoL1.sol now upon deployment. There is no public/external methods now.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -35,6 +35,11 @@ library LibProving {
     /// @notice Keccak hash of the string "RETURN_LIVENESS_BOND".
     bytes32 public constant RETURN_LIVENESS_BOND = keccak256("RETURN_LIVENESS_BOND");
 
+    /// @notice Keccak hash of the string "tier_provider".
```
dantaik, line 38: `RETURN_LIVENESS_BOND` can be `internal` now.

Brechtpd (approved):

packages/protocol/contracts/L1/libs/LibConstStrings.sol
```diff
@@ -0,0 +1,26 @@
...
+/// @custom:security-contact security@taiko.xyz
+library LibConstStrings {
+    /// @notice bytes32 representation of the string "tier_provider".
+    bytes32 internal constant BYTES32_STR_TIER_PROVIDER = bytes32("tier_provider");
+
+    /// @notice bytes32 representation of the string "proposer".
+    bytes32 internal constant BYTES32_STR_PROPOSER = bytes32("proposer");
+
+    /// @notice bytes32 representation of the string "taiko_token".
+    bytes32 internal constant BYTES32_STR_TKO = bytes32("taiko_token");
```
dantaik, line 14: BYTES32_STR_TKO => BYTES32_STR_TAIKO_TOKEN

packages/protocol/contracts/L1/libs/LibConstStrings.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.24;
+
+/// @title LibConstStrings
+/// @custom:security-contact security@taiko.xyz
+library LibConstStrings {
```
dantaik, line 6: If we use this name, it means all constants here are strings, so we may remove "_STR" from the var names.

packages/protocol/contracts/L1/libs/LibConstStrings.sol
dantaik: How about other strings "taiko", "bridge", "xxx_vault", ..., I think there are more.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -8,13 +8,15 @@ import "../../libs/LibAddress.sol";
...
 import "../tiers/ITierProvider.sol";
+import "./LibConstStrings.sol";
 
 /// @title LibProposing
 /// @notice A library for handling block proposals in the Taiko protocol.
 /// @custom:security-contact security@taiko.xyz
 library LibProposing {
     using LibAddress for address;
 
+    /// @notice Leaving this here to not forget about that this is a removed feature
```
dantaik, line 19: I prefer this new notice removed as it is not a string, it shall be here.

packages/protocol/contracts/signal/LibSignals.sol
```diff
@@ -5,8 +5,11 @@ pragma solidity 0.8.24;
 /// @custom:security-contact security@taiko.xyz
 library LibSignals {
     /// @notice Keccak hash of the string "STATE_ROOT".
-    bytes32 public constant STATE_ROOT = keccak256("STATE_ROOT");
+    bytes32 internal constant HASH_STR_STATE_ROOT = keccak256("STATE_ROOT");
```
dantaik, line 8: how about we delete this file and move the constants to  LibConstStrings?

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
```diff
@@ -6,6 +6,7 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
 import "../../common/EssentialContract.sol";
 import "../../libs/LibAddress.sol";
+import "../libs/LibConstStrings.sol";
```
dantaik, line 9: Please also use this in DeployL1.s.sol.

packages/protocol/contracts/L1/libs/LibConstStrings.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.24;
+
+/// @title LibConstStrings
```
dantaik, line 4: please move this file to contracts/common/

packages/protocol/contracts/L1/hooks/AssignmentHook.sol
adaki2004: Ok, but in `DeployOnL1.s.sol` it can be used but only 2 times because the functions (`deployProxy`, `register` and `copy_register`) we are supplying the `“xxxx_yyy”` are `string` literals, not `bytes32` representation of them, so it would require change the input parameters , which not handy, as it is used for debugging (printing the string) purposes. But i changed that 2 params.

packages/protocol/contracts/common/LibConstStrings.sol
```diff
@@ -0,0 +1,87 @@
...
+    bytes32 internal constant HASH_STATE_ROOT = keccak256("STATE_ROOT");
+
+    /// @notice Keccak hash of the string "SIGNAL_ROOT".
+    bytes32 internal constant HASH_SIGNAL_ROOT = keccak256("SIGNAL_ROOT");
+
+    /// @notice Constant string "SIGNAL".
+    string internal constant SIGNAL = "SIGNAL";
+
+    /// @notice bytes32 representation of the string "chain_pauser".
+    bytes32 internal constant BYTES32_CHAIN_PAUSER = bytes32("chain_pauser");
```
dantaik, line 17: @adaki2004 I still suggest that we use shorter leading string, "H_" for hash, "B_" for bytes32, "S_" for string. I even suggest to rename this to "Strings.sol" to make it shorter.

packages/protocol/contracts/common/LibConstStrings.sol
```diff
@@ -0,0 +1,87 @@
...
+    bytes32 internal constant BYTES32_BRIDGED_ERC721 = bytes32("bridged_erc721");
+
+    /// @notice bytes32 representation of the string "bridge_watchdog".
+    bytes32 internal constant BYTES32_BRIDGE_WATCHDOG = bytes32("bridge_watchdog");
+
+    /// @notice bytes32 representation of the string "rollup_watchdog".
+    bytes32 internal constant BYTES32_ROLLUP_WATCHDOG = bytes32("rollup_watchdog");
+
+    /// @notice Keccak hash of the string "RETURN_LIVENESS_BOND".
+    bytes32 internal constant HASH_RETURN_LIVENESS_BOND = keccak256("RETURN_LIVENESS_BOND");
```
dantaik, line 68: lets group srings, hashes, and bytes.

packages/protocol/contracts/common/LibConstStrings.sol
adaki2004: Done, but used `LibString` instead, since `Strings` bc. it could have been misleading. (OZ's Strings module)

dantaik (approved):

PR #16692, opened 2024-04-08
feat(relayer): relayer small fixes + logs

cyberhorsey:
fixes an unprofitable queue message issue with publishing the wrong body, and correctly parses a context.Done error as needing to be requeued.


dantaik:
> `bridge.isMessageReceived` is not working for L2-L1, it is returning false, but removing it allows the relayer to process the message. Unsure if this is a smart contract bug or otherwise right now, removed until figured out. It works L1-L2.

@adaki2004 could you please help Jeff to figure out why the function is not working? The code is the same and the implementation doesn't differentiate L1/L2.

adaki2004:
@cyberhorsey It can be related to caching and/or block verification ?
What kind of proof you are supplying for that L2 ? I guess the full proof, right ? (Not storage only).
If so, it might be that the L2 block is not yet verified, so you have to wait `lastVerifiedBlockId <= x`.
Also (theoretically calls the same functions under the hood so should have no impact) could you try out `proveMessageReceived` as underlying function, instead of `isMessageReceived`  - as input params, and retVal is the same ?

xiaodino (approved):

RogerLamTd (approved):

cyberhorsey:
> @cyberhorsey It can be related to caching / block verification / what kind of proof you supply.
> 
> What kind of proof you are supplying for that L2 ? I guess the full proof, right ? (Not storage only). If so, it might be that the L2 block is not yet verified, so you have to wait `lastVerifiedBlockId <= x`. Also (theoretically calls the same functions under the hood so should have no impact) could you try out `proveMessageReceived` as underlying function, instead of `isMessageReceived` - as input params, and retVal is the same ?

I will check it in a diff pr. the block is definitely verified because I can submit the proof and process the message. its only this external call that fails before hand, and returns false. the block is 100% verified.

PR #16698, opened 2024-04-09
fix(Protocol): use Permanent storage functions.

whitetechna:
### using permanent storage functions `sstore` and `sload`.


packages/protocol/contracts/common/EssentialContract.sol
```diff
@@ -119,7 +119,7 @@ abstract contract EssentialContract is UUPSUpgradeable, Ownable2StepUpgradeable,
     function _storeReentryLock(uint8 _reentry) internal virtual {
         if (block.chainid == 1) {
             assembly {
-                tstore(_REENTRY_SLOT, _reentry)
```
dantaik, line 122: why？

whitetechna: sstore is used for storing persistent data. while tstore is used for storing temporary data. we can use both of them as a storing part. but sstore and sload is helpful when you need to update state variables that persist across transactions.

PR #16712, opened 2024-04-11
feat(protocol): interpret bridge `message.fee` as the max fee

dantaik:
BEGIN_COMMIT_OVERRIDE
feat(protocol): interpret bridge message.fee as the max fee and refund unused fee
refactor(protocol): deprecate bridge message's 'refundTo' field
END_COMMIT_OVERRIDE

---

In this PR, we measure the actual gas used by message invocation `x`, calculate an actual fee per gas `p`, then allow the relayer to charge  actual fee  of `(x+o)*p` from `message.fee`, the rest are refunded to `message.destOwner`. Note `o` is the message processing overheads, whose value depends on whether the message is processed in one step or two.

We consider `message.fee/message.gasLimit` the max fee per gas, lets call it `m`, we ensure that:

- if `m < block.basefee`, `p = m`
- else `p = (m + block.basefee)/2`

--- 

@KorbinianK On our app, this is how the message.fee shall be calculated:
- estimate the max gas fee used on the destination chain to invoke the message call, use 0 if there is no message call at all (only Ether transfer), call it X
- add Bridge.TWO_STEP_PROCESSING_GAS_OVERHEAD() to X to get Y
- multiple the user-selected base fee per gas B with Y to get Z. We can also suggest a value = 2*current dest chain base fee.
- Z is the fee.


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -19,6 +19,9 @@ contract Bridge is EssentialContract, IBridge {
     using LibAddress for address;
     using LibAddress for address payable;
 
+    uint256 public constant ONE_STEP_PROCESSING_GAS_OVERHEAD = 40_000;
```
dantaik, line 22: these values are to be measured.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -308,29 +321,41 @@ contract Bridge is EssentialContract, IBridge {
...
             } else {
+                // gas now represents the actual gas used
+                uint256 fee;
+                {
+                    gasAmount -= gasleft();
+                    gasAmount += isNewlyProven
+                        ? ONE_STEP_PROCESSING_GAS_OVERHEAD
+                        : TWO_STEP_PROCESSING_GAS_OVERHEAD;
+
+                    uint256 maxFee = _message.fee * gasAmount / _message.gasLimit;
```
adaki2004, line 349: _messsage.fee (in previous versions at least) represented the  the "whole" fee the relayer pocket. In this context it is more `fee_per_gas`, no ?

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -308,29 +321,41 @@ contract Bridge is EssentialContract, IBridge {
...
+                    uint256 actualFee = block.basefee * gasAmount / _message.gasLimit;
```
Brechtpd, line 350:
```suggestion
                    uint256 actualFee = block.basefee * gasAmount;
```

I think this is the correct formula because basefee is per gas

packages/protocol/contracts/bridge/Bridge.sol
Brechtpd: I think it would still be the whole fee, and `* (gasAmount / _message.gasLimit)` just takes the fraction of the max total fee.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: You are right, good catch.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: _message.fee: max fee amount in ETH
_message.gasLimit: the tx's max gas limit including the overhead
_message.gasLimit - TWO_STEP_PROCESSING_OVERHEAD: the function call's max gas limit
_message.fee/_message.gasLimit: the max fee per gas.

dantaik:
This PR is ready to be reviewed again for merge.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -719,4 +759,36 @@ contract Bridge is EssentialContract, IBridge {
...
+    )
+        private
+        view
+        returns (uint256)
+    {
+        uint256 gasUsed = _gasStartValue - gasleft();
+        gasUsed += _isNewlyProven // use different overhead value
+            ? ONE_STEP_PROCESSING_GAS_OVERHEAD
+            : TWO_STEP_PROCESSING_GAS_OVERHEAD;
+        uint256 actualCost = block.basefee * gasUsed; // tip is not considered
```
adaki2004, line 783: Are we sure it is the desired calc (without tip) ? 

The realyer gets only this:
     `msg.sender.sendEtherAndVerify(fee, _SEND_ETHER_GAS_LIM)`  (which is on line 95).

But the fee is calculated with no tip 🤔 Maybe i need to wrap my head around it a little bit more.

Brechtpd: The fee is added below and it's that fee that is the `fee` value you posted.

Brechtpd (approved):

packages/protocol/contracts/bridge/Bridge.sol
dantaik: If we use gas price not base fee, then the relayer can use a really high tip to take all the fees -- imagine the relayer is a validator on L1.

Brechtpd: Oh that tip, seems like I misunderstood Dani's question. Yeah the relayer will have to make up for the tip he paid to the L1 validator in his L1 tx using the 50% he gets from the max fee set in the message.

cyberhorsey:
@dantaik I think we need to record the basefee of the first step of the two step invocation so the relayer can be profitable considering both steps.

dantaik:
> @dantaik I think we need to record the basefee of the first step of the two step invocation so the relayer can be profitable considering both steps. However this still doesn't let us know whether it's worth it for the relayer to even send the first tx because we don't know if the second tx baseFee at that point.
> 
> is there no way to do both steps at once and block proposers check for messages to be released/executed similiar to how eth deposits worked before, and proposers get a small% of fee?
> 
> Or when a bridge message is sent, they pay to finalize previous messages? IDK just spitballing here but it seems this still doesn't fully solve the "how to make two transactions guaranteed to be profitable so I know I should send the first tx"
> 
> Or perhaps best of all, one step mode can be activated with a signature from the watchdog, which could be verified onchain and the relayer can collect this before submitting.

I‘m afraid this current implementation is the best we can have for 2-step message execution. Hopefully we can disable the 2-step execution months after the genesis if there is no bug identified.

dantaik:
> @dantaik I think we need to record the basefee of the first step of the two step invocation so the relayer can be profitable considering both steps. However this still doesn't let us know whether it's worth it for the relayer to even send the first tx because we don't know if the second tx baseFee at that point.
> 
> is there no way to do both steps at once and block proposers check for messages to be released/executed similiar to how eth deposits worked before, and proposers get a small% of fee?
> 
> Or when a bridge message is sent, they pay to finalize previous messages? IDK just spitballing here but it seems this still doesn't fully solve the "how to make two transactions guaranteed to be profitable so I know I should send the first tx"
> 
> Or perhaps best of all, one step mode can be activated with a signature from the watchdog, which could be verified onchain and the relayer can collect this before submitting.

I think we can pay the two steps separately:
- step 1:  https://github.com/taikoxyz/taiko-mono/pull/16722
- step 2: still working on it

PR #16715, opened 2024-04-11
feat(protocol): trigger simultaneous recurring TKO snapshots

dantaik:
To allow voting on both L1 and L2, we need to prevent the same token from voting twice on both L1 and L2.  Making sure snapshots of TKO are taken in the same time is one of the conditions to prevent double voting.


Brechtpd (approved):

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -35,6 +35,8 @@ library TaikoData {
         // The max number of L2 blocks that can stay unsynced on L1 (a value of zero disables
         // syncing)
         uint8 blockSyncThreshold;
+        // The number of L1 blocks after which a TKO snapshot should be taken.
+        uint32 tkoSnapshotInterval;
```
adaki2004, line 39: Hmm, i'm not sure if this is a good way/practice to have snapshots taking at a predictable time. (Id ask Aragon maybe?)

If the tokens can be borrowed at a low interest rate OR a quick flash-loan can squeeze in right before a snapshot is taken, attacker can inflate his voting power with no (or low) extra cost.

dantaik: Stopped the auto merge. I do think we should talk to Aragon about this approach. 

Flashloan won't work as a way of stealing voting power as only a selected few addresses can take snapshot. But it is true that one can monitor pending propose-block transactions then buy a lot of taiko token, and sell them right after the snapshot is taken. However, people can always monitor snapshot transactions to front-run it to buy taiko tokens. No solution to my knowledge,so I do not see this is an issue that we need to solve. 

Let's discuss with Aragon before deciding if we should merge this PR.

packages/protocol/contracts/common/LibAutoSnapshot.sol
```diff
@@ -0,0 +1,29 @@
...
+    /// @param id The snapshot id.
+    event TaikoTokenSnapshotTaken(address tkoAddress, uint256 id);
+
+    // Take a snapshot every 200,000 L1 blocks which is roughly 27 days.
+    function autoSnapshot(IAddressResolver _resolver, uint256 _blockId) internal {
+        if (_blockId % 200_000 != 0) return;
+
+        address taikoToken = _resolver.resolve(LibStrings.B_TAIKO_TOKEN, true);
+        if (taikoToken != address(0)) {
+            uint256 id = ISnapshot(taikoToken).snapshot();
```
dantaik, line 25: will snapshot() also snapshot the voting power?

adaki2004 (approved):

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -33,8 +33,11 @@ contract BridgedERC20 is
     error BTOKEN_CANNOT_RECEIVE();
     error BTOKEN_UNAUTHORIZED();
 
-    modifier onlyOwnerOrSnapshooter() {
-        if (msg.sender != owner() && msg.sender != snapshooter) {
+    modifier onlyAuthorizedForSnapshot() {
+        if (
+            msg.sender != owner() && msg.sender != snapshooter
+                && msg.sender != resolve(LibStrings.B_TAIKO, true)
```
adaki2004, line 39: So actually `TaikoL2` contract can make a snapshot in any bridged contracts, right ? While indeed it does not doing it automatically just for TKO, right ? Isn't it a risk to any, arbitrary bridged contracts that technically taiko can make a snapshot - and sabotage their DAO ?

adaki2004: Maybe a check here like:
```
...
if ( address(this) == resolve(LibStrings.B_TAIKO_TOKEN, false)
{do this check}
else
{do regular check}
...
```

adaki2004 (approved):

PR #16732, opened 2024-04-14
change some internal design in bridge

dantaik:
BEGIN_COMMIT_OVERRIDE
feat(protocol): remove preferred message executer and extra invocation delay
feat(protocol): allow gasLimit to be 0 in `Bridge._invokeMessageCall` (by returning false)
feat(protocol): deprecate Bridge message's `refundTo` field
feat(protocol): replace `MessageRetried` event with `B_RETRY_FAILED` error
refactor(protocol): refactor Bridge code for readability
feat(protocol): interpret `message.fee` as the max fee
END_COMMIT_OVERRIDE

The above changes are included in this PR. 

`message.fee` will be used as the max message handling fee to pay for both receiving and processing the message. If a message is handled in two steps (receiving, then processing), the receiving part will also incur a fee and this fee is recorded in `receipt.feePaid`. At the end of message handling, any unused fee will be refunded to `message.destOwner`.


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -19,6 +19,18 @@ contract Bridge is EssentialContract, IBridge {
     using LibAddress for address;
     using LibAddress for address payable;
 
+    /// @dev Gas overhead if a message is received but not processed in the same transaction.
+    // The measured value  is 104047
+    uint256 public constant GAS_RECEIVING = 110_000 + 21_000;
```
dantaik, line 24: Brecht, if I measure the gas used inside the functions, shall I add 21000 to it?

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -19,6 +19,18 @@ contract Bridge is EssentialContract, IBridge {
     using LibAddress for address;
     using LibAddress for address payable;
 
+    /// @dev Gas overhead if a message is received but not processed in the same transaction.
+    // The measured value is 104047
+    uint256 public constant GAS_RECEIVING = 130_000;
```
dantaik, line 24: @Brechtpd shall I added 21000 gas to the measure gas amount used?

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -251,13 +267,22 @@ contract Bridge is EssentialContract, IBridge {
...
-                // if (!transactedByOwner) {
-                //     receipt.feePaid = 0;
-                //     msg.sender.sendEtherAndVerify(receipt.feePaid, _SEND_ETHER_GAS_LIMIT);
-                // }
+                if (!transactedByOwner) {
+                    receipt.feePaid = uint160(
+                        _calcFee(
+                            _message.fee,
+                            _message.gasLimit,
+                            GAS_RECEIVING,
```
Brechtpd, line 275: I think having a fixed gas cost here can be very inaccurate because merkle proofs can have very different costs. I don't know in practice how much they actually differ in costs at roughly the same tree size (very much depends which data and how the tree looks like), but the longer the signal contract is used the larger the proofs will get on average

The other difficulty is that a large part of the gas cost will be related to the calldata cost of the proof, not the processing cost inside the smart contract. Not sure if `GAS_PROCESSING` takes that into account, but seems like a pretty low number (though because of the small Merkle proofs when not a lot of data is already stored in the smart contract, maybe accurate because of that reason). 

For an accurate cost for the receiving, I would measure the gas used for verifying the merkle proof, and additional charge `proof.length * 16` gas for the proof calldata.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -328,8 +355,14 @@ contract Bridge is EssentialContract, IBridge {
         refundAmount += remainingFee;
 
         if (!transactedByOwner) {
-            refundAmount -= remainingFee;
-            msg.sender.sendEtherAndVerify(remainingFee, _SEND_ETHER_GAS_LIMIT);
+            uint256 fee = _calcFee(
+                _message.fee,
+                _message.gasLimit,
+                processInTheSameTx ? GAS_RECEIVING_AND_PROCESSING : GAS_PROCESSING,
```
Brechtpd, line 361: Seems like this does not take into account the actual gas used for processing the user's call? Don't see how a fixed gas cost could work here?

dantaik: if GAS_RECEIVING cannot be a fixed number, then I'll have to reconsider the solution.

dantaik:
Please take another look at the new code.

PR #16739, opened 2024-04-15
feat(protocol): redesign Bridge fee & gasLimit and remove 2-step processing

dantaik:
## Objectives

- Allow users to specify the message's fee to a large value without worrying about that all the fees will be taken by the relayer. The relayer should only make a small profit if the block's base fee is smaller than the message's specified max fee per gas, which is `message.fee /message.gasLimit`; the relayer will also only charge for the actual gas used.

- Allow the relayer to make simple decision regarding whether it should process a message: if `block.basefee <= message.fee /message.gasLimit`, process the message, otherwise, no.

- Allow the relayer to use a really large gas limit to transact `processMessage` without worrying about the actual invocation will eat all gas. The relayer can also simply just use `message.gasLimit` as the gas limit for his `processMessage` call, or with a reasonable overhead.

## Modifications

- removed two step processing to simplify design/impl, remove suspend message feature.
- `message.refundTo` is deprecated, the `destOwner` field is used instead.
-  `gasLimit` no longer means the invocation gas limit, it covers the entire `processMessage` call. The gas limit for the message invocation is calculated as:
   
   ```solidity 
    invocationGasLimit = message.gasLimit - gasDeductedForCallData - _GAS_RESERVE
   ```

   - `gasDeductedForCallData` is the call data gas reseve for the message struct itself, we reserve 32 gas per byte as a message may be processed in two steps. 
   - `_GAS_RESERVE` is a constant value, all processing gas cost will use this part of gas reserve, including the calldata for proof.

     The new implementation ensures even if the relayer specific a large gas limit for his `processMessage` transaction, the actual invocation will only use up to `invocationGasLimit` gas. 
  
     @cyberhorsey  To calculate the proper `message.gasLimit`, the relayer can use `Bridge.getMessageMinGasLimit(message) + expectedInvocationGasLimit`.

- refund 20000 gas per cache operation.


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -238,106 +262,106 @@ contract Bridge is EssentialContract, IBridge {
...
 
-        uint64 receivedAt = proofReceipt[msgHash].receivedAt;
-        if (receivedAt == type(uint64).max) revert B_MESSAGE_SUSPENDED();
+            receipt = ProofReceipt(uint64(block.timestamp), 0, 0);
 
-        (uint256 invocationDelay, uint256 invocationExtraDelay) = getInvocationDelays();
+            if (local.invocationDelay != 0) {
+                if (local.notProcessedByOwner) {
+                    receipt.gasUsed =
+                        uint32(local.gas - gasleft() + GAS_OVERHEAD_RECEIVING + _proof.length >> 4);
```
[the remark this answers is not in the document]
Brechtpd: `abi.decode` allows trailing bytes, so this attack would currently be possible. I'm not sure how this problem can be fixed. In smart wallets I think this is mostly a non-issue because most of the data is either fixed or needs to be signed by the wallet owner, in this case however the proof is just sent by the relayer, so it's tricky.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -238,106 +262,106 @@ contract Bridge is EssentialContract, IBridge {
...
-            delete proofReceipt[msgHash];
+        if (
+            _message.to == address(0) || _message.to == address(this)
+                || _message.to == local.signalService
+        ) {
+            // Handle special addresses that don't require actual invocation but
+            // mark message as DONE
+            local.refundAmount = _message.value;
+            _updateMessageStatus(local.msgHash, Status.DONE);
+        } else if (_invokeMessageCall(_message, local.msgHash)) {
```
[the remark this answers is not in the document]
cyberhorsey: noted. this is better than having message.gasLimit just be for the invocation, in terms of handling it offchain.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -238,106 +262,106 @@ contract Bridge is EssentialContract, IBridge {
...
+                ? GAS_OVERHEAD_RECEIVING_AND_INVOKING
+                : GAS_OVERHEAD_INVOKING;
 
-            // Determine the refund recipient
-            address refundTo =
-                _message.refundTo == address(0) ? _message.destOwner : _message.refundTo;
+            uint256 fee = _calcFee(
+                _message.fee, //
+                _message.gasLimit,
+                local.gas - gasleft() + overhead + _proof.length >> 4,
```
Brechtpd, line 356: will need to require here that the proof is empty for two step withdrawals to make sure that the relayer doesn't charge the user extra for useless data

packages/protocol/contracts/bridge/Bridge.sol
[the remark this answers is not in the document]
Brechtpd: It's actually even more complex than I thought because of the caching. The proof verification cost is now paid by the user, which makes sense, but the caching which is decided by the relayer is also paid by the user now. So there isn't much reason for the relayer to use the caching unless that somehow increases its profit (which could still be the case if it allows extracting a higher fee from the user).

So that also removed the benefit of the relayer being able to lower the cost of all bridge transactions by batching multiple togethe, lowering the cost of all bridge transactions in the batch. Now this caching cost (if even done) would be paid by the bridge transaction doing the caching (instead of, in the best case, the cost being charged spread out over the multiple message).

I think the main cause of these problems is the shared gas limit. There are two very distinct things the user wants the relayer to do, which are the two different steps. I think we need to have a gas limit/max fee per step to be able to handle the two steps correctly (even when not done in two separate steps):

- Step 1 - verifying the merkle proof + proof data: fixed fee not taking the gas used in account (still based on the base fee, so lower base fee results in less of the max fee being used, the fee paid is `message.maxFeeProofVerification * (block.basefee / message.maxBaseFee)`. This ensures that the relayer will want to optimize the proof verification as much as possible (making a profit when doing so), and also makes sure that the user can pay a fee that forces the relayer to optimize the processing of its bridge transactions with caching by setting the max fee low.
- Step 2 - message call invocation: basically the system in this PR based on gas used.

I think something like this would work, but needs some more thinking and eyes on it.

dantaik:
@Brechtpd please take another look.

I'm considering the removal two-step processing. The goal of introducing this feature is to handle bridge bugs, but its very existence may introduce its own bugs....

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -238,106 +209,53 @@ contract Bridge is EssentialContract, IBridge {
...
-                }
-
-                if (_invokeMessageCall(_message, msgHash, gasLimit)) {
-                    _updateMessageStatus(msgHash, Status.DONE);
-                } else {
-                    _updateMessageStatus(msgHash, Status.RETRIABLE);
-                }
-            }
+        if (msg.sender != _message.destOwner) {
+            uint256 gasUsed = _calcGasToCharge(gas, GAS_OVERHEAD, _proof.length, numCacheOps);
```
adaki2004, line 250: Hmm, why we need to calculate specifically with `_calcGasToCharge` with taking into account the number of caching ?

Isn't (psuedo) would to the job - so actually calculating the gas consumed:
```
uint256 gasWhenFunctionStarted = gasleft(); // same as like line 212
...
uint256 gasUsed = (gasWhenFunctionStarted - gasleft()) + gasOverhead
```

dantaik: Basically caching only benefit the relayer, not the message owner, therefore, if there is cache actions, the corresponding gas should be paid by the relayer, not the user.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -707,16 +565,57 @@ contract Bridge is EssentialContract, IBridge {
...
         private
         view
-        returns (bool)
+        returns (uint256)
+    {
+        if (_msgFee == 0 || _msgGasLimit == 0) return 0;
+
+        uint256 maxFee = _msgFee * _gasUsed / _msgGasLimit;
+        uint256 baseFee = block.basefee * _gasUsed;
+        return _msgFee.min(baseFee >= maxFee ? maxFee : (maxFee + baseFee) >> 1);
```
adaki2004, line 581: Just throwing it in here (helps with understanding)
Example 1.:

Let's say `maxFee` is 99.
Let's say `baseFee` is 100.

In this case the returned value (fee) is 99 - while the baseFee (so the actual, minimum real cost) would be higher (`100`), no ? This fee is then sent to `msg.sender` (aka `relayer`) -> But since the actual cost is higher, `relayer` will not process.

Example 2.:
Let's say `maxFee` is 110.
Let's say `baseFee` is 100.

In this case, the fee would be `110` `MIN` `((110+100) / 2 = 105 )` --> 105, so the relayer would process.

This part looks about good to me.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -32,11 +49,12 @@ contract Bridge is EssentialContract, IBridge {
...
 
     /// @dev Place holder value when not using transient storage
-    uint256 internal constant PLACEHOLDER = type(uint256).max;
+    uint256 private constant _PLACEHOLDER = type(uint256).max;
 
     /// @notice The next message ID.
     /// @dev Slot 1.
-    uint128 public nextMessageId;
+    uint64 private __reserved1;
+    uint64 public nextMessageId;
```
adaki2004, line 57: Is there any particular reason using `uint64` over `uint128` ?
As `nextMessageId` is used in `struct Message: id`, and `message` is either `calldata` or `memory`, there won't be any optimization gained from data-packing (bc. there is no data packing there).

Memory layout spec:
kép


Calldata layout spec:
kép

Link: https://docs.soliditylang.org/en/latest/internals/layout_in_memory.html

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -707,16 +565,57 @@ contract Bridge is EssentialContract, IBridge {
...
+    )
+        private
+        view
+        returns (uint32)
+    {
+        unchecked {
+            uint256 gasUsed = _gasStart - gasleft() + _gasOverhead
+                + _calldataBytes.min(_MAX_PROOF_BYTES_TO_CHARGE) >> 4;
+
+            uint256 refund = _numCacheOps * _GAS_REFUND_PER_CACHE_OPERATION;
```
Brechtpd, line 598: The user doing the caching will still end up paying the onchain verification cost for the additional merkle proof verification path. Which I guess is okay, the relayer will pick a message that has a max fee that is able to pay this cost for him, but if not the relayer will still be forced to eat the cost of this caching because the max fee for the message will be hit, and then he will have to make up for the cost of this caching he paid himself using the profit of other bridge transactions that are now more efficient/profitable to process (because the reduced cost of the merkle proof verification now allows enough of the fee to be used to actually process the call while still being profitable.

The dynamics here between the two different steps get quite tricky which is why I think it's easier to seem them as completely separate transactions. But the system in this PR where both can still influence each other I think is likely also okay but harder to wrap your head around.

Brechtpd:
> I'm considering the removal two-step processing. The goal of introducing this feature is to handle bridge bugs, but its very existence may introduce its own bugs....

The caching itself is fairly simple I think because it's just storing something and then checking if it was stored, but fair enough.

adaki2004 (approved):

adaki2004:
Generally dont see any breaking thing , so i’m accepting, maybe Jeff / Korbi during real integration can see something we cant. 🤷🏻‍♂️

cyberhorsey:
without any comment on the actual implementation, the offchain implications are very, very nice, and very simple to use from a relayer and bridge UI perspective.

cyberhorsey:
regarding this:
```ts
@cyberhorsey To calculate the proper message.gasLimit, the relayer can use Bridge.getMessageMinGasLimit(message) + expectedInvocationGasLimit.
```

Relaeyr will not be setting the message.gasLimit, the bridge-ui or sender or the the message will be. This is somethign for @KorbinianK

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -548,6 +398,16 @@ contract Bridge is EssentialContract, IBridge {
         return _msgHash ^ bytes32(uint256(Status.FAILED));
     }
 
+    /// @notice Returns the minimal gas limit required for sending a given message.
+    /// @param _message The message.
+    /// @return The minimal gas limit required for sending this message.
+    function getMessageMinGasLimit(Message calldata _message) public pure returns (uint32) {
+        unchecked {
+            uint256 calldataCost = (_message.data.length + 192) >> 4;
+            return uint32(GAS_RESERVE + calldataCost + 1);
```
Brechtpd, line 407: I think this effectively reserves `GAS_RESERVE` of the gas limit mainly for the proof verification. So two things can happen:
- `GAS_RESERVE` < actual proof verification cost: the relayer runs the risk to lose money because gas used > gas limit, so depending on the block basefee, the fee paid by the user may be capped and the relayer will maye have spent more money processing the bridge message than it got because the call invocation has a gas limit that allows more gas to be used than the max gas limit set by the user.
- `GAS_RESERVE` > actual proof verification cost: this gas got reserved for the proof verification but will be refunded to the user because not usable. Only downside I guess is that the user has to send more ETH to the destination chain than necessary (but that will always be the case, but maybe a bit worse than required depending on this value).

Compared to just using whatever gas is left seems to be that the gas limit for the invoked call is deterministic and not gameable by the relayer. So in this way they can be seen as separate steps each with its own fee payment, but with a single shared fee pool by shifting some of the risk/complexity of profitability to the relayer, and depending on how smart the relayer is it could also be beneficial to the user.

I think it works.

Brechtpd, line 407: What is the purpose of the +1 here?

Brechtpd (approved):
Optimistically approving.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: In example one, I expect the relayer NOT to process the message.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: If we don't consider the dynamic-size bytes in Message, the previous Message definition uses 320bytes, the new definition uses 192 bytes. he padding to 32 bytes doesn't apply to each field.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -707,16 +565,57 @@ contract Bridge is EssentialContract, IBridge {
...
+        uint256 _calldataBytes,
+        uint256 _numCacheOps
+    )
+        private
+        view
+        returns (uint32)
+    {
+        unchecked {
+            uint256 gasUsed = _gasStart - gasleft() + _gasOverhead
+                + _calldataBytes.min(_MAX_PROOF_BYTES_TO_CHARGE) >> 4;
```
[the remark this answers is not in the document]
dantaik: Agreed. I think we should make sure the GAS_OVERHEAD is large (but reasonable) enough so it will cover most proof data cost. I change the event GasChargedByRelayer to "log" more data so we can calculate GAS_RESERVE and GAS_OVERHEAD more accurately.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: Screenshot 2024-04-17 at 11 48 25

Actually caching doesn't trigger additional merkle proof verification path.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: I think we need to modify this contract later after gathering some data to fine-tune some constants.

dantaik: Removed.

packages/protocol/contracts/bridge/Bridge.sol
adaki2004: > If you have a struct with 2 uint128 fields, it will be packed into a 32 bytes, not 64 bytes.

Only in case of storage there is data packing savings.
kép

 I did not comment because i think there is an issue here, just wanted to say, that we dont optimize here anything, because `message` (which has the `id` field  as `nextMessageId`)struct (is memory or calldata) so data packing will not be done. 
 
 But anyway, we also dont lose anything here, so all good!

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -244,15 +243,26 @@ contract Bridge is EssentialContract, IBridge {
...
-            refundAmount -= fee;
-            msg.sender.sendEtherAndVerify(fee, _SEND_ETHER_GAS_LIMIT);
+            if (msg.sender != _message.destOwner && _message.gasLimit != 0) {
+                unchecked {
+                    uint256 gasUsed = gas - gasleft();
+                    emit GasChargedByRelayer(gasUsed, GAS_OVERHEAD);
 
-            emit GasChargedByRelayer(gasUsed);
+                    uint256 refund = numCacheOps * _GAS_REFUND_PER_CACHE_OPERATION;
+                    gasUsed = (GAS_OVERHEAD + gasUsed).max(refund) - refund;
```
adaki2004, line 252: I think here :
`gasUsed = (GAS_OVERHEAD + gasUsed).max(refund) - refund;`

`refund` is always smaller than `(GAS_OVERHEAD+gasUsed)` since `gasUsed` already captures the amount of gas consumed with caching.

So maybe just this:
`gasUsed = (GAS_OVERHEAD + gasUsed) - refund;`

packages/protocol/contracts/bridge/Bridge.sol
Brechtpd: It doesn't trigger additional merkle proof verification, but the bridge message the relayer uses to do the caching does end up paying to verify the bigger merkle proof. Bridge messages that come afterwards where the relayer makes use of the cached roots then have to verify less merkle proof paths so it will be cheaper for them. So it's not additional, but the the bridge message that did the caching is the sole payer normally of this "normal" bigger merkle proof verification.

packages/protocol/contracts/bridge/Bridge.sol
Brechtpd: The relayer could choose to cache a value that was already cached I think, in which case the SSTORE would be much cheaper than 20k (though the merkle proof verification is still extremely likely to be > 20k gas). 

But good to have the check there because it cannot be ruled out I think.

Brechtpd (approved):
Will go over the code again later today, but everything looks good so approving so I'm not blocking anything.

adaki2004 (approved):

PR #16744, opened 2024-04-16
feat(protocol): remove two step processing

dantaik:
- removed "refundTo" and "memo" from messages and vaults.
- reordered field in bridge messagen and vault's op structs to make it use fewer slots.


dantaik:
- removed "refundTo" and "memo" from messages and vaults.
- reordered field in bridge messagen and vault's op structs to make it use fewer slots

PR #16750, opened 2024-04-16
chore(protocol): add realistic testcase to simulate behavior

adaki2004:
@dantaik 
This is to address the fear that a token is unlocked (withdrawable) the same time when the grant grants Alice her 100%. No, it all depends on the parameters we supply in, so fully dynamic, but behavior is 100% compliant with grant letters (for labs people and taiko grants too).

Please read the testcase documentation example carefully.

To run the test:
`forge test --match-test test_realistic_scenario -vv`

### DIFF
--- a/packages/protocol/test/team/TimelockTokenPool.t.sol
+++ b/packages/protocol/test/team/TimelockTokenPool.t.sol
@@ -547,4 +547,125 @@ contract TestTimelockTokenPool is TaikoTest {
         assertEq(tko.balanceOf(Alice), 10_000e18);
         assertEq(usdc.balanceOf(Alice), 1_000_000_000e6 - paidUsdc);
     }
+
+    // Let's calculate at which date, how much is unlocked vs. withdrawable.
+    function test_realistic_scenario() public {
+        // Let's assume Alice started at Taiko 2023.02.01. 00:00:00 -> Unix timestamp is 1675206000
+        uint64 grantStart = 1_675_206_000;
+        // Full grant period of her is 4 years -> So tokens fully owned (but not unlcoked yet!) by
+        // 2027.02.01. 00:00:00
+        uint32 grantPeriod = 4 * 365 days;
+        // Half year is the cliff for "owning" tokens
+        uint64 grantCliff = grantStart + 180 days;
+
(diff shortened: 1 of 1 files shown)


packages/protocol/test/team/TimelockTokenPool.t.sol
```diff
@@ -547,4 +547,125 @@ contract TestTimelockTokenPool is TaikoTest {
         assertEq(tko.balanceOf(Alice), 10_000e18);
         assertEq(usdc.balanceOf(Alice), 1_000_000_000e6 - paidUsdc);
     }
+
+    // Let's calculate at which date, how much is unlocked vs. withdrawable.
+    function test_realistic_scenario() public {
```
dantaik, line 552: The test doesn't really cover the real world grant settings. Assume we hire Alice at TGE, granting her 1 million TKO.

- grantStart = TGE
- grantCliff: 90 days
- grantPeriod: 4 years
- unlockPeriod: 4 years.

Note that:

- there is no shared unlock start data. For each token (wei), its unlocking starts when this token is vested.
- Alice will own all these tokens at TGE+4 years
- Alice can withdraw all these tokens at TGE+8 years (not TGE+4 years).

dantaik:
Screenshot 2024-04-17 at 12 38 28
We can talk about it later.

packages/protocol/test/team/TimelockTokenPool.t.sol
adaki2004: This is in the token grant letter (or maybe it is in just mine?)
`Tokens that are allocated to the Company at Token Launch will only be available for delivery to you as and when such Tokens are released from the mandatory lockup restriction ("Token Unlock"), which is expected to occur from time to time during the period commencing on the date of Token Launch and ending on the fourth anniversary of the date of Token Launch`

So for Alice (in my example above) , her lockup starts (with some cliff ofc) at the TGE.

In your example, why you say Alice can only withdraw first her tokens after 4 years ? (and 8 years the last portion).

If this is the case, it has very serious implicaitons on token grants.

adaki2004: In you case the only thing we need to do is to set the unlock cliff to 1 year, and then that is it.

PR #16751, opened 2024-04-16
fix(protocol): renounce timelock_admin role from msg.sender

adaki2004:
As per OZ finding:
kép


dantaik (approved):

dantaik:
> @davidtaikocha just my only question is that `msg.sender` is preserving the same context during foundry scripts, because `renounceRole(role, address)` can only work if `address` = `msg.sender `

If it doesn't work, this function will revert.

davidtaikocha (approved):

PR #16758, opened 2024-04-17
feat(protocol): re-implement token grant contract

dantaik:
The calculation of the unlocked tokens in LibVestUnlock.sol can be illustrated using the diagrams below.

Screenshot 2024-04-17 at 23 44 56
Screenshot 2024-04-17 at 23 45 34


packages/protocol/contracts/team/TokenGrant.sol
```diff
@@ -0,0 +1,116 @@
...
+        grantAmount = _grantAmount;
+        startedAt = _startedAt;
+        vestDuration = _vestDuration;
+        unlockDuration = _unlockDuration;
+        costPerTko = _costPerTko;
+
+        emit GrantCreated(memo);
+    }
+
+    function withdraw() external whenNotPaused nonReentrant {
```
Brechtpd, line 73: Maybe can specify the withdrawal amount here? May help some people when the tokens need to bought and enough funds to be supplied. And doesn't seem like it would complicate the logic much.

packages/protocol/contracts/team/LibTokenGrant.sol
```diff
@@ -0,0 +1,67 @@
...
+        returns (uint256)
+    {
+        if (_vestDuration == 0) return _grantAmount;
+        return _portion(_grantAmount, _time, _vestDuration);
+    }
+
+    function calcUnlockedAmount(
+        uint256 _grantAmount,
+        uint256 _vestDuration,
+        uint256 _unlockDuration,
```
Brechtpd, line 23: Is there some explanation of how the unlocking is supposed to work? Going on the tests, I don't think I understand.

Brechtpd (commented):
Code is simple so that looks good, but I don't know the logic behind the unlocking so I can't say anything about that.

packages/protocol/contracts/team/LibTokenGrant.sol
dantaik: Screenshot 2024-04-17 at 23 44 56

Let me explain using the diagram above.

The green area represents the token grant total amount. The x-axis represents the time.

- ``a: the vesting duration
- `b`: the unlocking duration
- `a + b` then represents when all tokens will be vested AND unlocked.

The idea is that if some tokens are vested, these tokens' unlock period starts immediately at the vesting time:
- tokens vested at the first second (time 0) will have a unlocking period `[0, b]`;
- tokens vested at the last second (time a) will have a unlocking period `[a, a+b]`

packages/protocol/contracts/team/LibTokenGrant.sol
```diff
@@ -0,0 +1,67 @@
...
+    )
+        private
+        pure
+        returns (uint256)
+    {
+        if (a > b) revert INVALID_PARAMS();
+        if (t <= tMin) return 0;
+        uint256 _t = t - tMin;
+        return z * _t * _t / b / a / 2;
+    }
```
adaki2004, line 66: Here is an example based on this math:

Bob grant is 100K tokens. It vests 4 years, and unlocks also 4 years.

Bob start date (grant start): 2023.01.01
At 2025.01.01 the math returns in `calcUnlockedAmount()` is the following:
```

       return _triagleArea(_grantAmount, a, b, _time, 0)
            + _triagleArea(_grantAmount, a, b, _time, a + b)
            - _triagleArea(_grantAmount, a, b, _time, a) //
            - _triagleArea(_grantAmount, a, b, _time, b);
```

Only first line returns a value, rest is 0, so let's just calculate the unlocked amount:

`return z * _t * _t / b / a / 2;`

100k * 2 * 2 / 4 / 4 /2 = 12.5K tokens.

So far so "good", seems kind of intentional.

So Bob started 2 years ago, an (now) at this point is able to withdraw 12.5% of his tokens.

Let see with different dates, how much is unlcoked at year 3 ?
100k * 3 * 3 / 4 / 4 / 2 = 28.125

 after 3 years spent this is 28.125%
 
 etc.

adaki2004 (commented):
All in all i think the code is buggy at least 2 places:
 - termination
 - no real blocking point of withdrawing tokens after TGE as cliff is only applied at grantStart (which is employee start date),
 
 Also think this change for the worse. Now, to everyone get his / her tokens, it takes 8 years since 4 vesting and 4 unlock is somwhat "adding up" (even tho not linerarily. as you can see my 2year 3 year calculation: https://github.com/taikoxyz/taiko-mono/pull/16758/files#r1570514036) and in my opinion it goes against this sentence in the token grant letter:
 
` Tokens that are allocated to the Company at Token Launch will only be available for delivery to you as and when such Tokens are released from the mandatory lockup restriction ("Token Unlock"), which is expected to occur from time to time during the period commencing on the date of Token Launch and ending on the fourth anniversary of the date of Token Launch.`

Interpretation is: Unlocking ends on the 4th anniversary of the date of the token launch. You may still have unvested tokens(!) at year 4 post TGE, (if you joined after TGE) but the lockup should not exsist.

Personally, previous version is the version i would go with.

packages/protocol/contracts/team/TokenGrant.sol
```diff
@@ -0,0 +1,124 @@
...
+        Config memory c = config();
+        uint256 cost = amountToWithdraw * c.costPerTko / 1e18;
+        if (cost != 0) {
+            IERC20(feeToken).safeTransferFrom(msg.sender, owner(), cost);
+        }
+
+        emit GrantWithdrawn(amountToWithdraw, cost);
+    }
+
+    function terminate() external onlyOwner {
```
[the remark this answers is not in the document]
dantaik: Good feedback. Will change the impl.

packages/protocol/contracts/team/TokenGrants.sol
```diff
@@ -0,0 +1,30 @@
...
+contract TokenGrant3m4y4y1c is TokenGrant {
+    function config() public pure override returns (Config memory c) {
+        // 3 months
+        c.vestCliff = 90 days;
+        // 4 years
+        c.vestDuration = 4 * 365 days;
+        // 4 years
+        c.unlockDuration = 4 * 365 days;
+        // 1 cent, assuming the stable coin used has 6 decimals.
+        c.costPerTko = 10_000;
```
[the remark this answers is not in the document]
dantaik: Bridged ERC20's decimals settings is exactly the same as the in the canonical tokens.

PR #16760, opened 2024-04-18
feat(protocol): add back EventProcessed event and improve gas logging in Bridge

dantaik:
@cyberhorsey please take a look if this simple event will work for you.

@KorbinianK if you need any change in the protocol, please let me know.


adaki2004 (approved):

cyberhorsey:
I need the Message added to the event, like in MessageSent, to make the watchdog as reliable/simple as possible.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -198,6 +206,9 @@ contract Bridge is EssentialContract, IBridge {
     }
 
     /// @inheritdoc IBridge
+    /// @dev This transaction's gas limit must not be smaller than:
+    /// `(message.gasLimit - GAS_RESERVE) * 24 / 23 + GAS_RESERVE`,
```
Brechtpd, line 210:
```suggestion
    /// `(message.gasLimit - GAS_RESERVE) * 64 / 63 + GAS_RESERVE`,
```
I think these are the numbers, otherwise not sure how this formula was created (some doubling/approximation somewhere maybe? because the message data limit is also not taken into account here?).

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -198,6 +206,9 @@ contract Bridge is EssentialContract, IBridge {
     }
 
     /// @inheritdoc IBridge
+    /// @dev This transaction's gas limit must not be smaller than:
+    /// `(message.gasLimit - GAS_RESERVE) * 64 / 63 + GAS_RESERVE`,
+    /// Or we can use a simplified rule: `tx.gaslimit = message.gaslimit * 105%`.
```
dantaik, line 211:
```suggestion
    /// Or we can use a simplified rule: `tx.gaslimit = message.gaslimit * 102%`.
```

Brechtpd (approved):

cyberhorsey (approved):

PR #16766, opened 2024-04-19
feat(protocol): remove caching from signal service

dantaik:
Why? I think no relayer will be willing to cache anything as now the fee is no longer a fixed one so caching will only enable users to pay less fees. If caching is a feature we don't expect to be used, we should remove it.

I'd like to get your feedback @Brechtpd @cyberhorsey regarding this change, what's your preference?


Brechtpd:
The fee for the proof calldata is still fixed if I understand correctly (it's part of `GAS_RESERVE`), so the relayer should still be able to make use of caching to reduce the calldata cost for himself, while the user gets the benefit of not having to verify the additional proof logic.

Even though I proposed also making the verification cost fixed for the user, I think it still works out with the current fee system because the extra calldata for the account proof should be significant (the 20k gas for caching is only ~1250 bytes of calldata), so the caching cost should be worthwhile for a relayer when he can batch even maybe just two bridge messages (though not really sure anymore how big these merkle proofs were in practice, but I vaguely remember ~6k bytes but I could be wrong).

dantaik:
Alright, lets keep the feature for now.

PR #16770, opened 2024-04-19
refactor(protocol,eventindexer,relayer): merge GasLog into MessageProcessed

dantaik:
Jeff, after a bit of research, I realized that emitting this MessageProcessed event with the whole message is expensive (4224gas) and unnecessary, as the message itself has already been included in the transaction's param list, you can decode the transaction calldata to get the message.

Here is an example code snippet to decode transaction input params with a transaction hash. Please give it a try, if this works, I prefer applying this change.

```golang
import (
    "context"
    "fmt"
    "log"

    "github.com/ethereum/go-ethereum"
    "github.com/ethereum/go-ethereum/accounts/abi"
    "github.com/ethereum/go-ethereum/common"
    "github.com/ethereum/go-ethereum/core/types"
    "github.com/ethereum/go-ethereum/ethclient"
)

func main() {
    client, err := ethclient.Dial("YOUR_INFURA_URL")
    if err != nil {
        log.Fatalf("Failed to connect to the Ethereum client: %v", err)
    }

    txHash := common.HexToHash("your_transaction_hash_here")
    transaction, isPending, err := client.TransactionByHash(context.Background(), txHash)
    if err != nil {
        log.Fatalf("Failed to fetch transaction: %v", err)
    }
    fmt.Printf("Is Pending: %t\n", isPending)

    // Define your contract ABI
    const contractABI = `[
        {
            "constant": false,
            "inputs": [
                {"name": "_arg1", "type": "uint256"},
                {"name": "_arg2", "type": "string"}
            ],
            "name": "yourFunction",
            "outputs": [],
            "type": "function"
        }
    ]`

    // Parse the ABI
    abi, err := abi.JSON(strings.NewReader(contractABI))
    if err != nil {
        log.Fatal(err)
    }

    // Decode the transaction input
    if transaction.To() != nil { // Ensure this is a contract call
        method, err := abi.MethodById(transaction.Data())
        if err != nil {
            log.Fatalf("Error finding method ID: %v", err)
        }

        inputs := make(map[string]interface{})
        err = method.Inputs.UnpackIntoMap(inputs, transaction.Data()[4:])
        if err != nil {
            log.Fatalf("Error unpacking inputs: %v", err)
        }

        fmt.Println("Decoded inputs:")
        for name, input := range inputs {
            fmt.Printf("%s: %v\n", name, input)
        }
    }
}

```


cyberhorsey (approved):

Brechtpd (approved):

Brechtpd:
Very important to note here is that the way of getting the data does NOT work in the general case. This only works if an EOA calls `processMessage` directly, it does not work if some other contract calls it instead, because that contract can have a different function signature or even just creates the Message data dynamically inside that intermediate contract.

Because Jeff also needs this data for the watchdog and not just for `processMessage` calls it does itself (which the relayer could all through an EOA of course), it is not a reliable way to get the message data.

The only easy and reliable way to get the data is to just put the message in the event I think.

dantaik:
@cyberhorsey since Message object is still needed in an event, I think we can drop this PR entirely. What do you think?

cyberhorsey (approved):

PR #16785, opened 2024-04-21
fix(relayer): shell script update

chillpill91:


cyberhorsey:
what does this fix? abigen already reports an error directly, there is no need to wrap it

chillpill91:
Just added these simple checks that allows user to verify the commands are run successfully

cyberhorsey:
it will already fail with a given error, this isnt needed tbh

PR #16792, opened 2024-04-22
feat(protocol): delete ERC20 native token adaptor

dantaik:
The adaptor design introduces an issue. Using Native USDC token as one example, the adaptor is now the bridged token (for ERC20Vault), but the underlying native USTC token is not, so users can chose to bridge native USDC token as if it is a canonical token which creates confusion.

We shall deploy the Native USDC contracts, then use its `configureMinter` function to config `ERC20Vault` as one of the minters, then later add Circle's address(es) as minter(s) as well.


adaki2004 (changes requested):
I'm ok deleting it, but there are several issues with just deleting this - and that was the reason we introduced `USDCAdapter`.
Let's say Bob wants to bridge back USDC (from L2) to L1. Now, he calls `sendsMessage()`. Now, Bridge will always fail, since it cannot burn USDC (on L2), because:
- USDC mint() funciton signature is different than this: `IBridgedERC20(_token).burn(msg.sender, _amount)`.
The correct one for USDC is this:
`function burn(uint256 _amount)`
- only `minterRoles` can burn tokens
- only if `msg.sender` owns the tokens can burn the tokens.

So we also have to make ERC20Vault:
- has minterRole
- When it is USDC -> It handles burn witha  special case, which is: 1. transfers the tokens from user to ERC20Vault (also approve) and then burns is (see deleted USDCAdapter's `burnToken` function)

dantaik:
> I'm ok deleting it, but there are several issues with just deleting this - and that was the reason we introduced `USDCAdapter`. Let's say Bob wants to bridge back USDC (from L2) to L1. Now, he calls `sendsMessage()`. Now, Bridge will always fail, since it cannot burn USDC (on L2), because:
> 
> * USDC mint() funciton signature is different than this: `IBridgedERC20(_token).burn(msg.sender, _amount)`.
>   The correct one for USDC is this:
>   `function burn(uint256 _amount)`
> * only `minterRoles` can burn tokens
> * only if `msg.sender` owns the tokens can burn the tokens.
> 
> So we also have to make ERC20Vault:
> 
> * has minterRole
> * When it is USDC -> It handles burn witha  special case, which is: 1. transfers the tokens from user to ERC20Vault (also approve) and then burns is (see deleted USDCAdapter's `burnToken` function)

I think ` which is: 1. transfers the tokens from user to ERC20Vault (also approve) and then burns is (see deleted USDCAdapter's `burnToken` function)` works. Could you please follow up by working on this PR, @adaki2004

adaki2004:
> > I'm ok deleting it, but there are several issues with just deleting this - and that was the reason we introduced `USDCAdapter`. Let's say Bob wants to bridge back USDC (from L2) to L1. Now, he calls `sendsMessage()`. Now, Bridge will always fail, since it cannot burn USDC (on L2), because:
> > 
> > * USDC mint() funciton signature is different than this: `IBridgedERC20(_token).burn(msg.sender, _amount)`.
> >   The correct one for USDC is this:
> >   `function burn(uint256 _amount)`
> > * only `minterRoles` can burn tokens
> > * only if `msg.sender` owns the tokens can burn the tokens.
> > 
> > So we also have to make ERC20Vault:
> > 
> > * has minterRole
> > * When it is USDC -> It handles burn witha  special case, which is: 1. transfers the tokens from user to ERC20Vault (also approve) and then burns is (see deleted USDCAdapter's `burnToken` function)
> 
> I think `which is: 1. transfers the tokens from user to ERC20Vault (also approve) and then burns is (see deleted USDCAdapter's`burnToken` function)` works. Could you please follow up by working on this PR, @adaki2004

@dantaik 
I think it is fixed now, but also a complete E2E test need to be done - and also on Holesky - Hekla context. (So writing unit tests for this special case i dont think is necessary, since we would be mocking a lot of things anyways (USDC contract, etc) and given the fact Circle will require a testnet example, i'd propose the following:

1. If you agree with this functional change, deploy USDC with it's proxy on L2. (Hekla)
2. Add `minterRole` to `ERC20Vault` in USDC contract
3. call `changeBridgedToken` on Hekla, to set the newly deployed USDC (proxy)
4. bridge over some USDC from Holesky -> Hekla and see if it works

Prerequisite: We need to have an USDC contract on Holesky - but i think we have as I did - but maybe it was on Sepolia, i gotta check this out.

cc. @davidtaikocha  for helping with deployments/contract calls.

dantaik:
> > > I'm ok deleting it, but there are several issues with just deleting this - and that was the reason we introduced `USDCAdapter`. Let's say Bob wants to bridge back USDC (from L2) to L1. Now, he calls `sendsMessage()`. Now, Bridge will always fail, since it cannot burn USDC (on L2), because:
> > > 
> > > * USDC mint() funciton signature is different than this: `IBridgedERC20(_token).burn(msg.sender, _amount)`.
> > >   The correct one for USDC is this:
> > >   `function burn(uint256 _amount)`
> > > * only `minterRoles` can burn tokens
> > > * only if `msg.sender` owns the tokens can burn the tokens.
> > > 
> > > So we also have to make ERC20Vault:
> > > 
> > > * has minterRole
> > > * When it is USDC -> It handles burn witha  special case, which is: 1. transfers the tokens from user to ERC20Vault (also approve) and then burns is (see deleted USDCAdapter's `burnToken` function)
> > 
> > 
> > I think `which is: 1. transfers the tokens from user to ERC20Vault (also approve) and then burns is (see deleted USDCAdapter's`burnToken` function)` works. Could you please follow up by working on this PR, @adaki2004
> 
> @dantaik I think it is fixed now, but also a complete E2E test need to be done - and also on Holesky - Hekla context. (So writing unit tests for this special case i dont think is necessary, since we would be mocking a lot of things anyways (USDC contract, etc) and given the fact Circle will require a testnet example, i'd propose the following:
> 
> 1. If you agree with this functional change, deploy USDC with it's proxy on L2. (Hekla). Add that proxy address to `addressManager` with this tag:
> 
> ```
>     /// @notice bytes32 representation of the string "bridged_native_usdc".
>     bytes32 internal constant B_BRIDGED_NATIVE_USDC = bytes32("bridged_native_usdc");
> ```
> 
> 2. Add `minterRole` to `ERC20Vault` in USDC contract
> 3. call `changeBridgedToken` on Hekla, to set the newly deployed USDC (proxy)
> 4. bridge over some USDC from Holesky -> Hekla and see if it works
> 
> Holesky (L1) USDC contract: https://holesky.etherscan.io/address/0x6436d7b483a35516f7cd19379a392ce9bd0b2852
> 
> cc. @davidtaikocha for helping with deployments/contract calls.

Please proceed with these changes.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -355,7 +360,17 @@ contract ERC20Vault is BaseVault {
         // If it's a bridged token
         if (bridgedToCanonical[_token].addr != address(0)) {
             ctoken_ = bridgedToCanonical[_token];
-            IBridgedERC20(_token).burn(msg.sender, _amount);
+
+            bytes[] memory encodedCalls = _getBurnLowLevelCalls(_token, _user, _amount);
```
dantaik, line 364: I prefer to handle all Bridged token the same way if possible. Rather than burning the token, I think transfer the token to this address then burn it is a much better idea.

PR #16810, opened 2024-04-23
feat(repo): add `taikoon` and `supplementary-contracts` sub packages

dantaik:
Misc contracts that are not part of the core rollup protocol, including token airdrops, token grant, etc, I hope to move them to this new package eventually.


Brechtpd (approved):
`opcontracts` might be a bit confusing because it's used as a prefix for optimism's tech (op-stack, op-node, ...). That's also what I thought when first reading the name here that this would be a folder for the op contracts that we use in the protocol. :)

Not sure what a better name would be though, maybe just `contracts`?

image

Brechtpd (approved):

adaki2004 (approved):

PR #16812, opened 2024-04-23
feat(protocol): add a script to use native USDC token as the bridged token

dantaik:


packages/protocol/script/SetupUSDCBridging.s.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.24;
+
+import "../test/DeployCapability.sol";
+import "../contracts/tokenvault/ERC20Vault.sol";
+
+/// @dev This script shall be run on Taiko chain to hook up a native USDC deployment and the USDC
+/// deployed on EThereum.
```
adaki2004, line 8:
```suggestion
/// deployed on Ethereum.
```

packages/protocol/script/SetupUSDCBridging.s.sol
```diff
@@ -0,0 +1,46 @@
...
+/// To deploy a native USDC contract on Taiko, run:
+/// https://github.com/taikoxyz/USDC/blob/main/script/DeployUSDC.s.sol
+contract SetupUSDCBridging is DeployCapability {
+    address public constant USDC = 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48;
```
adaki2004, line 12: This is indeed USDC (FiatTokenProxy) on ethereum, but just flagging this address is an EOA on holesky, so only use this script on Mainnet.

adaki2004 (approved):

packages/protocol/script/SetupUSDCBridging.s.sol
dantaik: Take another look

adaki2004 (approved):

Brechtpd (approved):

PR #16814, opened 2024-04-23
fix(bridge-ui): update disabled for chainselector

monleru:
**Select error before connecting the wallet.**

Before update:
Select opens but the change is not active which looks like a bug.
Any actions are not active until the wallet is connected.
I think it's not supposed to work like that - https://www.youtube.com/watch?v=1GyqVcW-Xuc

image

After the update:
Select is active only after the wallet is connected.
image

### DIFF
--- a/packages/bridge-ui/src/components/ChainSelectors/ChainPill/ChainPill.svelte
+++ b/packages/bridge-ui/src/components/ChainSelectors/ChainPill/ChainPill.svelte
@@ -9,10 +9,10 @@
   import { classNames } from '$libs/util/classNames';
   import { truncateString } from '$libs/util/truncateString';
   import { uid } from '$libs/util/uid';
+  import { account } from '$stores';
 
   export let value: Maybe<Chain> | null = null;
   export let label = '';
-  export let readOnly = false;
   export let selectChain: (event: CustomEvent<{ chain: Chain; switchWallet: boolean }>) => Promise<void>;
 
   export let switchWallet = false;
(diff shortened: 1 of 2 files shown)


monleru:
> Before update: Select opens but the change is not active which looks like a bug.
> 
> image
> After the update:
> 
> image

packages/bridge-ui/src/components/ChainSelectors/ChainPill/ChainPill.svelte
```diff
@@ -34,6 +34,8 @@
       modalOpen = true;
     }
   };
+
+  $: disabled = !$account || !$account.isConnected;
```
KorbinianK, line 38:
```suggestion
  $: disabled = !$account || !$account.isConnected || readOnly;
```

packages/bridge-ui/src/components/ChainSelectors/ChainPill/ChainPill.svelte
```diff
@@ -9,10 +9,10 @@
   import { classNames } from '$libs/util/classNames';
   import { truncateString } from '$libs/util/truncateString';
   import { uid } from '$libs/util/uid';
+  import { account } from '$stores';
 
   export let value: Maybe<Chain> | null = null;
   export let label = '';
```
KorbinianK, line 15:
```suggestion
  export let label = '';
  export let readOnly = false;
```

Even though `readOnly` is currently unused, please do not remove this functionality.

KorbinianK (changes requested):

KorbinianK (approved):

davidtaikocha (approved):

PR #16824, opened 2024-04-24
feat(supplementary-contracts): add TokenUnlocking to supplementary

adaki2004:
Added the first version of `TokenUnlocking` contract.

Pretty simple, but there are some `todo`, please search for that text.

So my question (i also asked in discord DM):

Base situation, per those quarterly/half-yearly unvesting and it’s excersise :
1. Bob gets purchase notice from company (for the actual amount)
2. Bob decides to buy
3. If buy happened that amount of TKO token is deposited into the contract.

But the question here is: Is Bob really paying OR he acknowledges that he wants to get those tokens and he will pay once he will withdraw ?

1. Bob is paying for his tokens 'upfront', right. In case int the smart contract there is no need to maintain a `costToWithdraw` and `costToken` variables, because the purchase notice has to be paid before the vested amount is deposited into the contract. (If so, i have to remove them and contract gets simpler).
OR 
2. payment still remains the same ? During withdrawal,  Bob pays for it.

And in the situation above steps nr. 1 & 2 is just a "YES-I-ACKNOWEDGE-I-WANT-MY-TOKENS", but Bob does not have to pay immediately (just at withdrawal).

Somewhat i'm kind of in favor of the second version, because of the following reason:

- Bob knows TKO price NOW, but since those vested tokens not immediately unlocked, he does not know the price at time of unlock. So would be uncertain, if it worth to get those tokens or not.
- In case of grant nr. 1, it is OK since price is max 0.01 usd per token, but in case of grant nr. 2, we wont know what will be the price at unlocking...


adaki2004:
@dantaik No idea why it shows so many diffs.. I based my branch off of this:
kép

dantaik:
> @dantaik No idea why it shows so many diffs.. I based my branch off of this, could you please resolve this issue or let me know how can i ? kép

You should create a new PR off the main branch, copy your new files there and compile. We shall also use the npm to manage contracts dependency without using forge's `/lib` folder.

PR #16830, opened 2024-04-24
feat(supplementary-contracts): add TokenUnlocking

adaki2004:
Added the first version of `TokenUnlocking` contract.

Pretty simple, but there are some `todo`, please search for that text.

So my question (i also asked in discord DM):

Base situation, per those quarterly/half-yearly unvesting and it’s excersise :
1. Bob gets purchase notice from company (for the actual amount)
2. Bob decides to buy
3. If buy happened that amount of TKO token is deposited into the contract.

But the question here is: Is Bob really paying OR he acknowledges that he wants to get those tokens and he will pay once he will withdraw ?

1. Bob is paying for his tokens 'upfront', right. In case int the smart contract there is no need to maintain a `costToWithdraw` and `costToken` variables, because the purchase notice has to be paid before the vested amount is deposited into the contract. (If so, i have to remove them and contract gets simpler).
OR 
2. payment still remains the same ? During withdrawal,  Bob pays for it.

And in the situation above steps nr. 1 & 2 is just a "YES-I-ACKNOWEDGE-I-WANT-MY-TOKENS", but Bob does not have to pay immediately (just at withdrawal).

Somewhat i'm kind of in favor of the second version, because of the following reason:

- Bob knows TKO price NOW, but since those vested tokens not immediately unlocked, he does not know the price at time of unlock. So would be uncertain, if it worth to get those tokens or not.
- In case of grant nr. 1, it is OK since price is max 0.01 usd per token, but in case of grant nr. 2, we wont know what will be the price at unlocking...


Brechtpd:
I would also say generally option 2 makes the most sense, because if people have to pay upfront it's not really usable as an alternative way for income:
- People first have to be able to pay for those tokens, which would mean they have a significant amount of disposable money already (unless the amount of tokens here would be small in the future compared to fiat wages, but that is currently not the case so seems problematic for the current employees in any case). I would assume this is not the typical employee.
- People would be able to lose money on this because of the long time delays, which makes it risky and feels strange

I think option 1 is interesting for higher risk/higher reward, but doesn't seem like a great fit for normal employees unless the price/amount of the tokens is low enough.

dantaik:
The token grant agreement mandates option 1 (upfront payment). The token purchase action must be completed with full payment before tokens will be deposited to the smart contracts. 

To ensure people are not losing money, we should only charge grantees the cost after they have sold the token... We simply cannot remove all risks here.

packages/supplementary-contracts/contracts/TokenUnlocking.sol
```diff
@@ -1,3 +1,332 @@
...
+/// those will be deposited of course).
+///
+/// We should deploy multiple instances of this contract per grantee and per grant. So each person
+/// should have the same amount of contract deployed as grants granted (grant 1, grant 2, etc.)
+/// @custom:security-contact security@taiko.xyz
+contract TokenUnlocking is OwnableUpgradeable, ReentrancyGuardUpgradeable {
+    using SafeERC20 for IERC20;
+
+    // Variables, which shall be same for everybody, it is part of the GrantInfo struct.
+    struct UnlockTiming {
```
dantaik, line 32: I would suggest to remove struct UnlockTiming and put its data fields directly in to the contract.

We can also make these fields public constant variables and give them some fix values, then in a sub-contract of this contract, we override these fixed values with values that can decided after TGE.

packages/supplementary-contracts/contracts/TokenUnlocking.sol
```diff
@@ -1,3 +1,332 @@
...
+        /// @notice Shall be TGE date.
+        uint64 unlockStartDate;
+        /// @notice The end date of cliff period. Shall be 1 year post TGE (date).
+        uint64 unlockCliffDate;
+        /// @notice Shall be 4 years (period)
+        uint32 unlockPeriod;
+    }
+
+    // Variables, which shall be unique per grantee
+    struct GrantInfo {
```
dantaik, line 42: Since each grantee will have his/her own deployment, we can also get rid of this struct and flatten it into the contract's storage layout directly.

packages/supplementary-contracts/contracts/TokenUnlocking.sol
```diff
@@ -1,3 +1,332 @@
...
+        /// @notice Represents how many tokens withdrawn already and helps with: withdrawable
+        /// amount.
+        // - The current "withdrawable amount" is determined by the help of this variable =
+        // (amountVested *(% of unlocked) ) - amountWithdrawn
+        uint128 amountWithdrawn;
+        /// @notice The address of the recipient.
+        address grantRecipient;
+        // This can define which grant is this (grant 1, grant 2, etc.). Only informative
+        // purposes.
+        uint8 grantId;
```
dantaik, line 57: I think we can get rid of this ID, each deposit is a new vesting.

packages/supplementary-contracts/contracts/TokenUnlocking.sol
```diff
@@ -1,3 +1,332 @@
...
+
+    // Holding any grant information
+    GrantInfo public grantInfo;
+
+    /// @notice The Taiko token address.
+    address public taikoToken;
+
+    /// @notice The shared vault address, from which tko token deposits will be triggered by the
+    /// depositToGrantee() function
+    address public sharedVault;
```
dantaik, line 68: Since the grantee has paid, lets transfer the TKO token directly from "msg.sender" to this contract without holding purchased tokens in the shared vault.

packages/supplementary-contracts/contracts/TokenUnlocking.sol
```diff
@@ -1,3 +1,332 @@
...
+
+    /// @notice Mapping of recipient address to the next withdrawal nonce.
+    mapping(address recipient => uint256 withdrawalNonce) public withdrawalNonces;
```
dantaik, line 71: Lets remove this map so this contract is not shared by multiple recipients.

packages/supplementary-contracts/contracts/TokenUnlocking.sol
```diff
@@ -1,3 +1,332 @@
...
+        uint64 unlockCliffDate,
+        uint32 unlockPeriod,
+        uint8 grantId
+    );
+
+    /// @notice Emitted when a grant is made.
+    /// @param recipient The grant recipient address.
+    /// @param currentDeposit The current deposited tko amount.
+    /// @param totalVestedAmount The total vested amount so far.
+    event DepositTriggered(
```
dantaik, line 93: how about TokenVested, and change "deposit" to "vestToken"

dantaik (changes requested):

packages/supplementary-contracts/contracts/TokenUnlocking.sol
dantaik: Overriding constant variables may not be possible with solidity, so maybe define a virtual function getTGETimestamp() then always assume the unlocking is 1y + 3y.

packages/supplementary-contracts/contracts/TokenUnlocking.sol
adaki2004: It is for granting access to withdraw if Bob grants his signature to Alice (to withdraw on behalf of Bob).

packages/supplementary-contracts/contracts/TokenUnlocking.sol
adaki2004: We still need a "trigger deposit" and that deposit shall be coming from a vault but thru this contract. 
Otherwise we cannot keep track of deposited ERC20, since there is no fallback or receive() funciton which is called. So let's say TaikoLabs can say, it deposited 100K tokens, but indeed it did not.
All i'm saying is: this `depositToGrantee` should trigger a deposit into this contract (so that we can keep track of the indeed vested amount).

packages/supplementary-contracts/contracts/TokenUnlocking.sol
adaki2004: Left a storage var for `tgeTimestamp` and be it initailizable during `init`. It can be overriden tho later with `getTgeTimestamp()` in subcontracts if needed.

dantaik (approved):

packages/supplementary-contracts/contracts/TokenUnlocking.sol
adaki2004: I can remove for simplicity for sure.
Please do not merge this until i do testing tomorrow.

packages/supplementary-contracts/contracts/TokenUnlocking.sol
```diff
@@ -1,3 +1,237 @@
...
+    address grantRecipient;
+    /// @notice For tests or sub-contracts, getTgeTimestamp() can be overridden.
+    uint64 tgeTimestamp;
+    /// @notice The Taiko token address.
+    address public taikoToken;
+    /// @notice The shared vault address, from which tko token deposits will be triggered by the
+    /// depositToGrantee() function
+    address public sharedVault;
+    /// @notice Mapping of recipient address to the next withdrawal nonce.
+    mapping(address recipient => uint256 withdrawalNonce) public withdrawalNonces;
```
Brechtpd, line 49: Can be removed now

Brechtpd (approved):

Brechtpd (approved):

dantaik (approved):

PR #16839, opened 2024-04-25
fix(relayer): dont return errunprofitable if gasUsed > gas Limit

cyberhorsey:
we can tweak this later during testnet, for exact gas usage. Right now, just the current profitability check above is fine, and we we will log the estimated gas used for later tweaking.


davidtaikocha (approved):

RogerLamTd (approved):

cyberhorsey:
We can close this since we fix the GAS_RESERVE on chain.

PR #16862, opened 2024-04-26
fix(taikoon): fix taikoon contract test dependency issues

dantaik:


packages/taikoon/package.json
bearni95, line 1: So it was `@openzeppelin/foundry-upgrades` causing the test issues?

bearni95 (dismissed):
Seems like the `@openzeppelin/foundry-upgrades` package was behind the test issues

bearni95 (approved):

davidtaikocha (approved):

PR #16864, opened 2024-04-26
chore(protocol): recover the out-of-data tcb info error.

smtmfft:
Missed 1 recoverable error type OUT_OF_DATE, update tcb file for testnet workaound.

PR #16867, opened 2024-04-26
feat(protocol): put automata dcap v3 ra behind proxy

adaki2004:
As per request. https://github.com/taikoxyz/taiko-mono/pull/16864#issuecomment-2078786482


dantaik:
and who is the owner to perform upgrade? The idea was deploy non-upgradable contract to use address manager to switch to a new verifier if needed.

adaki2004:
> and who is the owner to perform upgrade? The idea was deploy non-upgradable contract to use address manager to switch to a new verifier if needed.

Owner to upgrade i guess is the proxy owner.
But good question, i let it be answered by John, dont know who shall maintain this.

dantaik:
If this contract is stateless, then maybe we just deploy a new version.

adaki2004:
> If this contract is stateless, then maybe we just deploy a new version.

Please decide, in our current design:
A: deploy a new and upgrade the lookup in AddressManager
or 
B: upgrade

both can work. I have no strong opinion, maybe B is more handy and easier (?). Please you decide.

davidtaikocha (approved):

dantaik:
@johntaiko @smtmfft 

The reason I think we should keep this contract *not upgradeable* is to ensure we can sync with the official repo's code easily, otherwise, the official repo is not upgradeable and ours is, then we always have to manually copy code and make necessary changes to our copy to ensure it works.

dantaik:
> @johntaiko @smtmfft
> 
> The reason I think we should keep this contract _not upgradeable_ is to ensure we can sync with the official repo's code easily, otherwise, the official repo is not upgradeable and ours is, then we always have to manually copy code and make necessary changes to our copy to ensure it works.

Given that this verifier has state data, I'm OK with upgradeability

dantaik (approved):

PR #16871, opened 2024-04-26
fix(bridge-ui): make if else statements more readable

hkey0:
A small fix to improve code readability.


KorbinianK:
How are chained ternary expressions more readable?

PR #16899, opened 2024-04-28
feat(protocol): allow first block proposer to skip EOA check signature

dantaik:
This is to avoid the first block proposer (you know who) having to sign the txList to avoid unnecessary interaction.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -140,7 +140,7 @@ library LibProposing {
             // We cannot rely on `msg.sender != tx.origin` for EOA check, as it will break after EIP
             // 7645: Alias ORIGIN to SENDER
             if (
-                _checkEOAForCalldataDA
+                _checkEOAForCalldataDA && meta_.id != 1
                     && ECDSA.recover(meta_.blobHash, params.signature) != msg.sender
             ) {
                 revert L1_INVALID_SIG();


adaki2004 (approved):

adaki2004 (approved):

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -140,7 +140,7 @@ library LibProposing {
             // We cannot rely on `msg.sender != tx.origin` for EOA check, as it will break after EIP
             // 7645: Alias ORIGIN to SENDER
             if (
-                _checkEOAForCalldataDA
+                _checkEOAForCalldataDA && meta_.id != 1
```
adaki2004, line 143: Just wondering, if this way EOA will still be the proposer, right (`Bob` for block nr. 1) ? Not a smart contract relaying things ?
Also in this case how some could be a proposer if the meta `sender: msg.sender`, so this `Bob` is the sender physically, OR he is not msg.sender, just someone.. who put together the block data off-chain ?

Brechtpd: Should be okay I think:
- First proposer is still checked with msg.sender
- First proposer address is hard coded and could be made to make sure it's an EOA (though if it's not, also not that big of a deal though the node may have some troubles getting the data at first)

Brechtpd (approved):

PR #16900, opened 2024-04-28
chore(eventindexer): remove INDEX_NFTS

xiaodino:


cyberhorsey:
why?

xiaodino:
> why? when we run this on mainnet L1 we will not index nfts, it will massively bloat the database, and we will use an actual nft indexer API like moralis or alchemy or quicknode for the bridge. We need to keep this variable.

If we keep this variable, we should assume that NFT on index may not be always available. cc @KorbinianK

cyberhorsey:
> > why? when we run this on mainnet L1 we will not index nfts, it will massively bloat the database, and we will use an actual nft indexer API like moralis or alchemy or quicknode for the bridge. We need to keep this variable.
> 
> If we keep this variable, we should assume that NFT on index may not be always available. cc @KorbinianK

we just set it to true for our testnets, and we should have every NFT from genesis onwards in our L2, and every NFT from the point taikoL1 was deployed on Holesky.  
  
then on mainnet we set it to false for L1 and true for L2.

dantaik (approved):

PR #16902, opened 2024-04-28
fix: remove unnecessary exit() command

eminmtas:
This pull request removes an unnecessary exit() command from the end of the script, allowing the graph plotting code to execute properly. The changes ensure that the script functions as intended without any hindrance to its execution flow.


dantaik:
@eminmtas Thank you for your contribution. This script, however, is a utility that is not part of our production code. I think it's ok to leave it as-is.

PR #16926, opened 2024-04-30
refactor(supplementary-contracts): suggest additional changes.

dantaik:
a few renaming and simplification. I didn't run the scirpt, so please make sure it works before merging.


packages/supplementary-contracts/script/tokenunlocking/Vest.s.sol
```diff
@@ -0,0 +1,55 @@
...
+    struct VestingItem {
+        bytes32 name; // Conversion from json "string" to bytes32 will take place in foundry,
+            // cannot use string here, as json parser cannot interpret string from json, everything
+            // is bytes-chunks. It is more of informational to script executor anyways.
+        address recipient;
+        address proxy;
+        uint256 vestAmount;
+    }
+
+    uint256 public PRIVATE_KEY = vm.envUint("PRIVATE_KEY"); // owner
```
adaki2004, line 24: We dont need this PRIVATE_KEY in this implementation then. Because we use the `msg.sender `(`line 39`) we need to call the script in anyways with the `--private-key` flag with `forge script`.

packages/supplementary-contracts/script/tokenunlocking/Vest.s.sol
```diff
@@ -0,0 +1,55 @@
...
+
+            require(TokenUnlocking(proxy).owner() == msg.sender, "msg.sender not owner");
+            require(
+                TokenUnlocking(proxy).recipient() == items[i].recipient, "inconsistent recipient"
+            );
+
+            uint128 vestAmount = uint128(items[i].vestAmount * 1e18);
+            require(tko.balanceOf(msg.sender) >= vestAmount, "insufficient TKO balance");
+
+            vm.startBroadcast(PRIVATE_KEY);
```
adaki2004, line 47: We can just simply use `vm.startBroadcast()` at the beginning, since no need to have `PRIVATE_KEY` in env var, as we need to call the `forge script` with `--private-key`

dantaik: feel free to merge and modify as you wish, Dani.

PR #16932, opened 2024-04-30
feat(protocol): implement timestamp based checkpoints for TKO & BridgedERC20

dantaik:


packages/protocol/contracts/L1/TaikoToken.sol
```diff
@@ -84,4 +84,8 @@ contract TaikoToken is EssentialContract_, ERC20VotesUpgradeable {
         if (_to == address(this)) revert TKO_INVALID_ADDR();
         return super.transferFrom(_from, _to, _amount);
     }
+
+    function clock() public view override returns (uint48) {
+        return SafeCastUpgradeable.toUint48(block.timestamp);
```
dantaik, line 89: shall I use `SafeCastUpgradeable.toUint48(block.timestamp/ SOME_CONSTANT)` instead?

adaki2004: We needCLOCK_MODE() override too:
Something like:

```
// solhint-disable-next-line func-name-mixedcase 
function CLOCK_MODE() public pure override returns (string memory) { 
    return "mode=timestamp"; 
}
```

brickpop: @dantaik I would suggst keeping the best practices from OpenZeppelin and keep the consistency check: 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/utils/Votes.sol#L68-L71

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -93,6 +93,16 @@ contract BridgedERC20 is BridgedERC20Base_, ERC20VotesUpgradeable {
         return (srcToken, srcChainId);
     }
 
+    function clock() public view override returns (uint48) {
+        return SafeCastUpgradeable.toUint48(block.timestamp);
+    }
+
+    // solhint-disable-next-line func-name-mixedcase
+    function CLOCK_MODE() public view override returns (string memory) {
+        // See https://eips.ethereum.org/EIPS/eip-6372
```
brickpop, line 102: Same as above: 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/utils/Votes.sol#L68-L71

packages/protocol/contracts/L1/TaikoToken.sol
dantaik: Thank you, Jordi, but since the clock() function is no longer a virtual function, I think removing the check is fine.

brickpop (approved):

adaki2004 (dismissed):

dantaik:
@davidtaikocha we need to deploy this new version and perform a L1 upgrade to the TKO token contract.

Brechtpd (approved):

davidtaikocha (approved):

PR #16943, opened 2024-05-01
fix(eventindexer): nft balance changes should be wrapped in db transaction

the-laziest:
Increase and subtract of nft balance in db should be done in transaction to prevent corrupted data in case of error in second query


xiaodino (dismissed):
@the-laziest Could you explain more what the PR is for and a way to reproduce the issue?

the-laziest:
> @the-laziest Could you explain more what the PR is for and a way to reproduce the issue?

Sure, @xiaodino,  if there will be an error for some reason [here](https://github.com/taikoxyz/taiko-mono/blob/215a6001c2c0dc0e8d2d20a438be9889792ff6ef/packages/eventindexer/indexer/index_nft_transfers.go#L135) (related to db connection or smth like that), then in database NFT balance will be increased for `To` address, but not decreased for `From` address

cyberhorsey:
Two notes:

1) I'm happy to merge this PR, however, we will not use any of this code, and it will be deleted for mainnet (since we will use professional NFT indexing APIs that already exist), so it is not robust on purpose.

2) We can delete the existing `IncreaseBalance` and `SubtractBalance` methods now, since they are unused, so if this PR is to be merged, we should do that.

the-laziest:
@cyberhorsey, updated all mentioned things

packages/eventindexer/indexer/index_nft_transfers.go
```diff
@@ -116,35 +116,34 @@ func (i *Indexer) saveERC721Transfer(ctx context.Context, chainID *big.Int, vLog
...
 		Address:         to,
 		TokenID:         tokenID,
 		ContractAddress: vLog.Address.Hex(),
 		ContractType:    "ERC721",
 		Amount:          1, // ERC721 is always 1
-	})
-	if err != nil {
-		return err
 	}
+	descreaseOpts := eventindexer.UpdateNFTBalanceOpts{}
```
cyberhorsey, line 128: typo, should be decrease not descrease

packages/eventindexer/nft_balance.go
```diff
@@ -30,8 +30,11 @@ type UpdateNFTBalanceOpts struct {
 
 // NFTBalanceRepository is used to interact with nft balances in the store
 type NFTBalanceRepository interface {
-	SubtractBalance(ctx context.Context, opts UpdateNFTBalanceOpts) (*NFTBalance, error)
-	IncreaseBalance(ctx context.Context, opts UpdateNFTBalanceOpts) (*NFTBalance, error)
+	IncreaseAndSubtractBalancesInTx(
+		ctx context.Context,
+		increaseOpts UpdateNFTBalanceOpts,
+		subtractOpts UpdateNFTBalanceOpts,
+	) (*NFTBalance, *NFTBalance, error)
```
cyberhorsey, line 37: returning two pointers to the same strict here is not clear. we should either give them names return parameters, or wrap them in a parent struct and give each a named member, like "IncreasedNFTBalance" and "DecreasedNFTBalance" on the parent struct of "NFTBalances"

cyberhorsey (changes requested):

cyberhorsey:
Left two comments.

the-laziest:
@cyberhorsey , could you please take a look on it?

cyberhorsey (approved):

cyberhorsey:
LGTM. Thanks for doing the requested changes.

dantaik (approved):

the-laziest:
@cyberhorsey , can I please ask for contributor poap, if it's applicable to this pr

RogerLamTd:
@gitpoap-bot please give @the-laziest a poap

PR #16949, opened 2024-05-01
refactor(protocol): improve contracts based on suggestions from OpenZeppelin

adaki2004:
BEGIN_COMMIT_OVERRIDE
refactor(protocol): improve contracts based on suggestions from OpenZeppelin
chore(protocol): delete ERC20Airdrop2 contract
END_COMMIT_OVERRIDE

---

(A blue checkmark does not mean it is implemented but rather that it is answered in defender -> which in some cases it is a NOT implement / acknowledge, please check defender response !)

- [x] N11
- [x] N12
- [ ] N13 Missing or misleading documentation (Would love to get help from dev. rels if possible on those).
- [x] N14
- [x] N15
- [x] N16
- [x] N17
- [x] N18


packages/protocol/contracts/tokenvault/BridgedERC1155.sol
```diff
@@ -94,21 +94,21 @@ contract BridgedERC1155 is EssentialContract, ERC1155Upgradeable {
...
-    /// @param _amount Amount of tokens to burn.
-    function burn(
+    /// @param _ids Array of IDs of the tokens to burn.
+    /// @param _amounts Amount of tokens to burn respectively.
+    function burnBatch(
         address _account,
-        uint256 _tokenId,
-        uint256 _amount
+        uint256[] memory _ids,
+        uint256[] memory _amounts
```
dantaik, line 104: shall we change this function to external, then we can change 'memory' to 'calldata' to save some gas?

packages/protocol/contracts/common/EssentialContract.sol
```diff
@@ -100,7 +100,7 @@ abstract contract EssentialContract is UUPSUpgradeable, Ownable2StepUpgradeable,
         __AddressResolver_init(_addressManager);
     }
 
-    function __Essential_init(address _owner) internal virtual {
+    function __Essential_init(address _owner) internal virtual onlyInitializing {
```
dantaik, line 103: we should remove `onlyInitializing` from the other `__Essential_init` then.

packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
```diff
@@ -36,7 +36,7 @@ contract ERC20Airdrop2 is MerkleClaimable {
     /// @notice Event emitted when a user withdraws their tokens.
     /// @param user The address of the user.
     /// @param amount The amount of tokens withdrawn.
-    event Withdrawn(address user, uint256 amount);
+    event Withdrawn(address indexed user, uint256 amount);
```
dantaik, line 39: Dani, how about we delete this file.

dantaik:
@davidtaikocha I think we only need to upgrade these deployments:

- BridgedERC1155
- ERC1155Vault
- SgxVerifier
- BridgedERC20

packages/protocol/contracts/team/airdrop/ERC20Airdrop2.sol
dantaik: I'm assuming ERC20Airdrop is the one we'll use.

adaki2004: Yes, true. I'm removing now.

packages/protocol/contracts/tokenvault/BridgedERC20Base.sol
```diff
@@ -62,7 +64,7 @@ abstract contract BridgedERC20Base is EssentialContract, IBridgedERC20 {
         address _migratingAddress = migratingAddress;
         if (msg.sender == _migratingAddress) {
             // Inbound migration
-            emit MigratedTo(_migratingAddress, _account, _amount);
+            emit MigratedTo(msg.sender, _account, _amount, migratingInbound);
```
dantaik, line 67: I think the first address in MigratedTo is the token's address, not the user's address.

Brechtpd: This is the case where `msg.sender == _migratingAddress` so seems okay?

Though I think it's clearer to use `_migratingAddress` like it was. But the event name seems a bit confusing because `_migratingAddress` is actually the address that is migrated from (unless I misunderstand, calling the event `Migrated` now that it also contains the `migratingInbound` maybe more clear if I understand correctly at least.

dantaik, line 67:
```suggestion
            emit MigratedTo(_migratingAddress, _account, _amount, migratingInbound);
```

dantaik: Lets change it to `Migrated`

packages/protocol/contracts/common/EssentialContract.sol
```diff
@@ -22,8 +22,8 @@ abstract contract EssentialContract is UUPSUpgradeable, Ownable2StepUpgradeable,
         0xa5054f728453d3dbe953bdc43e4d0cb97e662ea32d7958190f3dc2da31d9721b;
 
     /// @dev Slot 1.
-    uint8 private __reentry;
-    uint8 private __paused;
+    uint8 public __reentry;
```
dantaik, line 25: instead of making these public:

- change `_inNonReentrant` to `inNonReentrant` and make it public
- we already have `paused` as a public method

dantaik (approved):

davidtaikocha (approved):

davidtaikocha (approved):

PR #16950, opened 2024-05-01
feat(protocol): improve TaikoToken, BridgedERC20, and ERC20Vault

dantaik:
BEGIN_COMMIT_OVERRIDE
refactor(protocol): merge `BridgedERC20Base`into `BridgedERC20`for simplicity
feat(protocol):  allow contract owner to mint/burn bridged ERC20 tokens (besides ERC20Vault)
feat(protocol): disallow bridged token contract owner to trigger migration directly (must go through ERC20Vault)
feat(protocol): disallow migration within 90 days post the previous migration
feat(protoocl): make BridgedERC20 no longer inherit `ERC20VotesUpgradeable`
feat(protocol): add `BridgedTaikoToken` that inherits `ERC20VotesUpgradeable`
END_COMMIT_OVERRIDE

This is to respond this feedback from OZ

---

Tokens inheriting from the [BridgedERC20Base](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/tokenvault/BridgedERC20Base.sol) contract can be [migrated](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/tokenvault/BridgedERC20Base.sol#L43) by the owner or by calling the [changeMigrationStatus](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/tokenvault/BridgedERC20Base.sol#L36) function of the ERC20Vault. If a migration is triggered, token owners can [burn](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/tokenvault/BridgedERC20Base.sol#L76-L82) their tokens to mint the same amount of tokens on the new contract. If the migration is triggered by the bridge, old tokens are [added to a blacklist](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/tokenvault/ERC20Vault.sol#L181) and have to be migrated before being able to bridge.

However, there are multiple migration scenarios that could cause accidental user losses:

- If a new migration is triggered before some users have migrated from a previous migration, then these users would effectively lose their tokens as they would [no longer be able to call mint()](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/tokenvault/BridgedERC20Base.sol#L61-L63) on the new contract nor bridge their tokens.
- If a [migration](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/tokenvault/BridgedERC20Base.sol#L43) is triggered by the owner instead of the ERC20Vault, the vault would no longer be able to call burn or mint on the migrated token. This would prevent users from bridging and could cause losses as the new migrated token would not be bridging to the same contract when [bridging through the vault](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/tokenvault/ERC20Vault.sol#L207).
- If a migration happens to a token that doesn't implement the IBridgedERC20 interface, user losses could happen as well.
- If such concerns are shared, consider addressing the above instances to protect users' funds from errors during migrations. Potential solutions include adding a delay to migrations during which no new migration can be done, preventing the owner from calling changeMigrationStatus, and validating that the new token implements the IBridgedERC20 using [ERC-165](https://eips.ethereum.org/EIPS/eip-165) during migrations.


dantaik:
Screenshot 2024-05-01 at 23 03 13


@davidtaikocha TaikoToken's storage layout stays the same!

Brechtpd (dismissed):

packages/protocol/contracts/tko/BridgedTaikoToken.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.24;
+
+import "../tokenvault/IBridgedERC20.sol";
+import "./TaikoTokenBase.sol";
+
+/// @title BridgedTaikoToken
+/// @notice The TaikoToken on L2 to support checkpoints and voting. For testnets, we do not need to
+/// use this contract.
```
adaki2004, line 9: Wondering why we need this (special) for TaikoToken ? The idea is now, to remove Votes from bridged tokens ? (Except bridged TKO)

packages/protocol/contracts/tokenvault/BridgedERC20.sol
```diff
@@ -1,39 +1,56 @@
...
-    uint256[50] private __slots_previously_used_by_ERC20SnapshotUpgradeable;
-}
 
 /// @title BridgedERC20
 /// @notice An upgradeable ERC20 contract that represents tokens bridged from
 /// another chain.
+/// Note this contract offers timestamp-based checkpoints and voting functions.
 /// @custom:security-contact security@taiko.xyz
-contract BridgedERC20 is BridgedERC20Base_, ERC20VotesUpgradeable {
+contract BridgedERC20 is EssentialContract, IBridgedERC20, ERC20Upgradeable, IERC165Upgradeable {
```
adaki2004, line 16: Ok, so the idea is not to have Votes anymore in Bridged tokens ? If so the note above:
`Note this contract offers timestamp-based checkpoints and voting functions.` -> Seems to be not correct as we dont have voting.
Also how do we ensure the upgrade from bridged tokens (alread on hekla) ? Or we simply dont upgrade the old tokens, but from X point on (X is where the impl. is deployed on Hekla) new tokens bridged from Hoelsky -> Hekla will use this deployed implementation behind the proxy ?

packages/protocol/contracts/tko/BridgedTaikoToken.sol
dantaik: yes

packages/protocol/contracts/tokenvault/BridgedERC20.sol
dantaik: The comment removed.
Correct, for deployed bridged tokens on Hekla, we leave them as-is.

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -159,7 +165,10 @@ contract ERC20Vault is BaseVault {
         nonReentrant
         returns (address btokenOld_)
     {
-        if (_btokenNew == address(0) || bridgedToCanonical[_btokenNew].addr != address(0)) {
+        if (
+            _btokenNew == address(0) || bridgedToCanonical[_btokenNew].addr != address(0)
+                || !_btokenNew.supportsInterface(type(IBridgedERC20).interfaceId)
```
adaki2004, line 170: `_btokenNew.supportsInterface(type(IBridgedERC20).interfaceId)` -> How the USDC contract (which we plan to deploy right after genesis block) will pass this check ?
This script will fail, if not mistaken:
https://github.com/taikoxyz/taiko-mono/pull/16812/files

dantaik: Good finding. Removed this and IERC165 related functions.

adaki2004 (approved):

packages/protocol/contracts/common/EssentialContract.sol
```diff
@@ -101,6 +111,7 @@ abstract contract EssentialContract is UUPSUpgradeable, Ownable2StepUpgradeable,
     }
 
     function __Essential_init(address _owner) internal virtual onlyInitializing {
+        __Context_init();
```
Brechtpd, line 114: Do we need to call this here?

Brechtpd (approved):

packages/protocol/contracts/common/EssentialContract.sol
dantaik: Not really, it's a empty function ...

davidtaikocha (approved):

PR #16652, opened 2024-04-05
fix(eventindexer, blobstorage): improved configuration and README documentation

SamiAlHassan:
This pull request introduces updates to the eventindexer, providing clearer setup guidance.

**Key Changes:**
- Updated `README.md` to include detailed setup instructions.
- Created docker-compose.yml for the mysql db.
- Updated `.env` files (`.default.env`, `.l1.env`, and `.l2.env`) with the database credentials and the [new contract addresses](https://docs.taiko.xyz/network-reference/addresses/).

However, the files in `taiko-mono/packages/eventindexer/contracts/` and in `taiko-mono/packages/blobstorage/api` need to be regenerated with the new contracts, as they are with the old addresses neither package works as both return `no contract code at given address`. Could someone guide me on how to do this, please?

PR #16958, opened 2024-05-01
chore(main): update CI to ignore checks in audit files

SamiAlHassan:
### Description

This PR makes changes to the typo checking setup by adding the external audits folder `"packages/protocol/audit/**"` to the exclude list, otherwise these will be flagged in every PR as in #16652.

I'm assuming it's better not to fix typos in the audit files because altering them could compromise their (legal?) integrity, and those typos don't show up in any of the products or docs.


cyberhorsey:
this has already been done in a PR that was merged, but thank you!

PR #16966, opened 2024-05-02
ci(repo,protocol): commit contract layout file in protocol workflow

dionysuzx:
I decided to use a github action instead of a post-commit hook, because this is better to ensure any changes to "protocol/*" will come with the updated layouts.md.


d1onys1us:
@dantaik oops, i forgot the base branch was not protected by approvals, so i merged this.

if you don't like it, you can revert the commit on your branch. thanks

PR #16969, opened 2024-05-03
chore(protocol): delete buggy Lib4844.sol

dantaik:
The implementation is buggy and we'll not use it in zkVM proofs.

## from OZ:

The [evaluatePoint](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/libs/Lib4844.sol#L30C14-L30C27) function is used to validate that the opening of a KZG Commitment (corresponding to a blob in practice) at a point x is equal to the purported y by calling Ethereum's [point evaluation precompile](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/libs/Lib4844.sol#L43-L45). This precompile expects its commitment and proof arguments to be [48 bytes each](https://www.evm.codes/precompiled#0x0a?fork=cancun).

However, the _commitment and _pointProof arguments are of type bytes1[48], and passed by calling [abi.encodePacked](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/libs/Lib4844.sol#L44). The packed encoding of a bytes1[48] is done in-place without the length but with each element [padded to 32 bytes](https://docs.soliditylang.org/en/v0.8.24/abi-spec.html#abi-packed-mode). This means that the commitment and proof arguments sent to the precompile are 32 * 48 = 1536 instead of 48 bytes long, and thus the first padded 3 bytes of the commitment are erroneously interpreted as representing both the commitment and the proof. Because these are padded and mostly 0s, they are unlikely to correspond to a valid commitment/proof pair.

In practice, this means that any normal call to the point evaluation precompile would revert. Additionally, gas could be [saved](https://docs.soliditylang.org/en/v0.8.24/types.html#bytes-and-string-as-arrays) by changing the type of these [arguments](https://github.com/taikoxyz/taiko-mono/blob/b47fc34cb0e7fe9b7ebd9416b3051a067483b860/packages/protocol/contracts/libs/Lib4844.sol#L34-L35) to bytes and validating their length separately. We note that while the Lib4844 library is in the scope of this audit, it is not currently used by the contracts in scope.

Consider changing the _commitment and _pointProof arguments to be of type bytes. Additionally, if the Lib4844 is intended to be used in practice, consider adding tests targeting this library.


dantaik:
@Brechtpd I don't think this lib will every be used in zkVM proofs, but let me know otherwise.

Brechtpd (approved):
There is a small chance we may still use the precompile as an optimization at some point, but that's getting unlikelier by the day. So I also don't see a reason to keep it in the main repo.

dantaik:
> There is a small chance we may still use the precompile as an optimization at some point, but that's getting unlikelier by the day. So I also don't see a reason to keep it in the main repo.

OK, then lets delete it for know. We can revert this commit then fix the bug if we change our mind.

davidtaikocha (approved):

PR #16970, opened 2024-05-03
feat(protocol): add bridge rate limiter for ETH and ERC20s

dantaik:
Following open zeppelin's suggestion, this PR introduces a feature to manage cross-chain quota. For each token, we can config the daily quota. 

Note that the quota is only consumed and checked on the destination chain (process message and retry message) and on the source chain (recall message but NOT send message). So if I set 1000 ether daily quota on L1, on L2, people can still withdraw more than 1000 ether, but on L1, they can NOT withdraw more than 1000 ether per day, so bridge and UI doesn't need to add new features to check available quota when messages are sent.

The relayer backend does need to check the quota before processing messages, otherwise, the relayer will not get paid.

Lets vote on this. I'm ok with or without this PR to be merged.
- thumb up for yes
- thumb down for no


dantaik:
Need to add some tests later

packages/protocol/contracts/common/LibStrings.sol
```diff
@@ -4,31 +4,32 @@ pragma solidity 0.8.24;
...
 /// @custom:security-contact security@taiko.xyz
 library LibStrings {
-    bytes32 internal constant B_CHAIN_WATCHDOG = bytes32("chain_watchdog");
-    bytes32 internal constant B_WITHDRAWER = bytes32("withdrawer");
-    bytes32 internal constant B_PROPOSER = bytes32("proposer");
-    bytes32 internal constant B_PROPOSER_ONE = bytes32("proposer_one");
-    bytes32 internal constant B_SIGNAL_SERVICE = bytes32("signal_service");
-    bytes32 internal constant B_TAIKO_TOKEN = bytes32("taiko_token");
-    bytes32 internal constant B_TAIKO = bytes32("taiko");
+    bytes32 internal constant B_AUTOMATA_DCAP_ATTESTATION = bytes32("automata_dcap_attestation");
```
dantaik, line 7: I sorted the strings alphabetily and added B_RATE_LIMITER

packages/protocol/contracts/common/AddressResolver.sol
```diff
@@ -26,6 +26,17 @@ abstract contract AddressResolver is IAddressResolver, Initializable {
         _;
     }
 
+    /// @dev Modifier that ensures the caller is the resolved address of a given
+    /// name.
```
adaki2004, line 30:
```suggestion
    /// @dev Modifier that ensures the caller is a resolved address to either _name1 or _name2
    /// name.
```

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -437,4 +442,11 @@ contract ERC20Vault is BaseVault {
             ctokenDecimal: ctoken.decimals
         });
     }
+
+    function _checkRateLimit(address _ctoken, uint256 _amount) private {
+        address rateLimiter = resolve(LibStrings.B_RATE_LIMITER, true);
+        if (rateLimiter != address(0)) {
+            IRateLimiter(rateLimiter).consumeAmount(_ctoken, _amount);
+        }
+    }
```
adaki2004, line 451: Function name `_checkRateLimit` a little bit ambigous to me, since it callse consumeAmount, which i guess eventually will deduct some balance from something, so not only 'checking'.

dantaik: renamed

packages/protocol/contracts/bridge/QuotaManager.sol
```diff
@@ -0,0 +1,75 @@
...
+        onlyFromNamedEither(LibStrings.B_BRIDGE, LibStrings.B_ERC20_VAULT)
+    {
+        uint256 available = availableQuota(_token);
+        if (available == type(uint256).max) return;
+        if (available < _amount) revert QM_OUT_OF_QUOTA();
+
+        unchecked {
+            available -= _amount;
+        }
+        tokenQuota[_token].available = available.toUint104();
```
adaki2004, line 63:
```suggestion
        tokenQuota[_token].available = uint104(available);
```

packages/protocol/script/DeployL1QuotaManager.s.sol
```diff
@@ -0,0 +1,36 @@
...
+        QuotaManager qm = QuotaManager(
+            deployProxy({
+                name: "quota_manager",
+                impl: address(new QuotaManager()),
+                data: abi.encodeCall(QuotaManager.init, (owner, addressManager))
+            })
+        );
+
+        // L2-to-L1 Ether per day: 25_000 Ether
+        qm.updateDailyQuota(address(0), 25_000 ether);
```
Brechtpd, line 31: seems quite high, this is already $75M so a loss of this magnitude is already very significant. Maybe we can start lower?

packages/protocol/contracts/bridge/QuotaManager.sol
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /// @notice Returns the available quota for a given token.
+    /// @param _token The token address with Ether represented by address(0).
+    function availableQuota(address _token) public view returns (uint256) {
+        Quota memory q = tokenQuota[_token];
+        if (q.dailyQuota == 0) return type(uint256).max;
+        if (q.updatedAt == 0) return q.dailyQuota;
+
+        uint256 issuance = q.dailyQuota * (block.timestamp - q.updatedAt) / 24 hours;
```
Brechtpd, line 80: Maybe good to discuss what the timeframe should be. Maybe 24 hours good to start with , but with the bridge watchdog working reliably maybe we can lower it to shorter times. This way we can reduce the amount that can be stolen (bridge watchdog paused the contracts when that happens), while still allowing more money to be withdrawn per day.

Brechtpd (approved):

packages/protocol/contracts/bridge/QuotaManager.sol
dantaik: Good feedback. I think shorter quota period is probably more suitable, changing from 25000 per day to 1000 per hour or 500 per 30 minutes can lower the max amount to be stolen.

Brechtpd (approved):

cyberhorsey:
@dantaik given the relayer sends multiple messages at a time (ie: it has multiple pending transactions simultaneously), what would be the best way to ensure we don't waste gas on reverted transactions, given this new rate limit cap?

Brechtpd:
> @dantaik given the relayer sends multiple messages at a time (ie: it has multiple pending transactions simultaneously), what would be the best way to ensure we don't waste gas on reverted transactions, given this new rate limit cap?

Not specifically for this but in general the relayer will have the same problem as people proposing blocks in a based rollup: So I will again shill hooking into PBS also for this like with https://docs.flashbots.net/flashbots-protect/overview (though I don't know what the best service is to do this in practice). There is no other way (except maybe some of the account abstraction stuff for revert protection but I don't think that's directly usable now and maybe also not sufficient) to prevent both other parties stealing transactions and also ensuring no money is wasted on failed/reverted transactions.

Based rollups can't exist without hooking into PBS for the same reason. The problem here where messages can be processed by anyone is the same, though here keeping the transaction private is less of a concern but you still need to have protection against failed transactions so have to make the L1 tx fee payment conditional on the successful execution somehow. Otherwise as soon as there are multiple relayers processing messages there will be (as vitalik called it) total anarchy and everybody losing money on failed transactions.

cyberhorsey:
> > @dantaik given the relayer sends multiple messages at a time (ie: it has multiple pending transactions simultaneously), what would be the best way to ensure we don't waste gas on reverted transactions, given this new rate limit cap?
> 
> Not specifically for this but in general the relayer will have the same problem as people proposing blocks in a based rollup: So I will again shill hooking into PBS also for this like with https://docs.flashbots.net/flashbots-protect/overview (though I don't know what the best service is to do this in practice). There is no other way (except maybe some of the account abstraction stuff for revert protection but I don't think that's directly usable now and maybe also not sufficient) to prevent both other parties stealing transactions and also ensuring no money is wasted on failed/reverted transactions.
> 
> Based rollups can't exist without hooking into PBS for the same reason. The problem here where messages can be processed by anyone is the same, though here keeping the transaction private is less of a concern but you still need to have protection against failed transactions so have to make the L1 tx fee payment conditional on the successful execution somehow. Otherwise as soon as there are multiple relayers processing messages there will be (as vitalik called it) total anarchy and everybody losing money on failed transactions.

:|

dantaik:
- There is a QuotaUpdated event, it will be emitted when the quota changed by the admin.
- `availableQuota` can be used to query the currently available quota.

dantaik:
BTW, we'll deploy a QuotaManager on only L1 for Ether and USDC/USDT initially.

cyberhorsey:
> BTW, we'll deploy a QuotaManager on only L1 for Ether and USDC/USDT initially.

so Quota manager should be optional in the relayer. Got it, will update my PR accordingly.

adaki2004 (approved):

davidtaikocha (approved):

PR #16992, opened 2024-05-04
fix(protocol): fix two bridge bugs with additional tests

dantaik:
## Bugs

- When `message.destOwner` calls `processMessage`,  `gasleft()` should be used to invoke the message call, not `_invocationGasLimit(_message, true)`.

-  If the following statement is evaluated to be `true`, it will always be `true`, thus `retryMessage` won't work. The fix makes sure we send ether to `message.destOwner` instead.

  ```
   _message.data.length >= 4
  && bytes4(_message.data) != IMessageInvocable.onMessageInvocation.selector
  && _message.to.isContract()
  ```

## Improvements

- checks src and dest chain IDs more carefully

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -229,15 +229,7 @@ contract Bridge is EssentialContract, IBridge {
             _proveSignalReceived(signalService, msgHash, _message.srcChainId, _proof);
 
         uint256 refundAmount;
-        if (
-            _message.to == address(0) || _message.to == address(this)
-                || _message.to == signalService
-                || (
-                    _message.data.length >= 4
-                        && bytes4(_message.data) != IMessageInvocable.onMessageInvocation.selector
-                        && _message.to.isContract()
-                )
-        ) {
+        if (_unableToInvokeMessageCall(_message, signalService)) {
             // Handle special addresses that don't require actual invocation but
             // mark message as DONE
             refundAmount = _message.value;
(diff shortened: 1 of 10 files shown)


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -226,25 +224,30 @@ contract Bridge is EssentialContract, IBridge {
...
+        stats.numCacheOps =
             _proveSignalReceived(signalService, msgHash, _message.srcChainId, _proof);
-        if (!received) revert B_SIGNAL_NOT_RECEIVED();
 
         uint256 refundAmount;
         if (
             _message.to == address(0) || _message.to == address(this)
                 || _message.to == signalService
+                || (
+                    _message.data.length >= 4
```
dantaik, line 235: bug fix 2

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -226,25 +224,30 @@ contract Bridge is EssentialContract, IBridge {
...
         ) {
             // Handle special addresses that don't require actual invocation but
             // mark message as DONE
             refundAmount = _message.value;
             _updateMessageStatus(msgHash, Status.DONE);
         } else {
-            Status status = _invokeMessageCall(
-                _message, msgHash, _invocationGasLimit(_message, true)
-            ) ? Status.DONE : Status.RETRIABLE;
+            uint256 gasLimit = msg.sender == _message.destOwner
```
dantaik, line 245: bug fix 1

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -605,4 +595,22 @@ contract Bridge is EssentialContract, IBridge {
             IQuotaManager(quotaManager).consumeQuota(address(0), _amount);
         }
     }
+
+    function _unableToInvokeMessageCall(
```
dantaik, line 599: bug fix 2

Brechtpd (approved):

adaki2004 (approved):

PR #16996, opened 2024-05-05
fix(eventindexer): fix Dockerfile problems

edvenukko:
There are two problems in the current `Dockerfile` setup which can lead to issues when building and running
### Building problem

Currently `ARG PACKAGE=eventindexer` is after `FROM golang:1.21.0 as builder`, when building with `docker build . -t taiko-mono`, only `builder` stage can correctly get the `PACKAGE=eventindexer`, while stage with `alpine` will get empty `PACKAGE`, causing the problem:
```
Step 13/14 : COPY --from=builder /taiko-mono/packages/$PACKAGE/bin/$PACKAGE /usr/local/bin/
COPY failed: stat taiko-mono/packages//bin/: file does not exist
```
This can be fixed by moving `ARG PACKAGE=eventindexer` to the first line of Dockerfile, and add `ARG PACKAGE` in each stage.

### Running problem

Currently Dockerfile is with `ENTRYPOINT ["$PACKAGE"]`, however in https://docs.docker.com/reference/dockerfile/#shell-and-exec-form we know this form(called "exec form") 
> makes it possible to avoid shell string munging, and to invoke commands using a specific command shell,
>
> This means that normal shell processing, such as variable substitution, doesn't happen.

So using ENTRYPOINT like this will cause the following problem while running:
```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "${PACKAGE}": executable file not found in $PATH: unknown.
```

I've added `ENV PACKAGE=${PACKAGE}`  and changed to `ENTRYPOINT /usr/local/bin/${PACKAGE}`, now the image can be run with commands like `docker run -it taiko-mono PACKAGE=eventindexer`.


edvenukko:
Ah, finally, the title is correct now. 😅

xiaodino (approved):

davidtaikocha (approved):

PR #17003, opened 2024-05-06
fix(protocol): fix metadata retrieval in vaults

dantaik:
try/catch does not catch abi decoding issues, so when 
```
try meta.symbol() returns (string memory _symbol) {
           symbol = _symbol;
       } catch { }
```
is called on a token like MKR whose symbol is a bytes32, it reverts with EvmError (quick test [here](https://gist.github.com/0xmp/ae133fa842031a4777070dee092135b6) for a repro). To support these, the most failsafe way to do this would be to do a low level call, and on success decode the result with a custom function (Examples [here](https://github.com/0xPolygonHermez/zkevm-contracts/blob/main/contracts/PolygonZkEVMBridge.sol#L835-L860) and [here](https://github.com/OffchainLabs/token-bridge-contracts/blob/b3894ecc8b6185b2d505c71c9a7851725f53df15/contracts/tokenbridge/arbitrum/StandardArbERC20.sol#L50-L61).). Sorry I should have been clearer in the recommendation here that try/catch is not the maximally compatible option


adaki2004 (approved):

packages/protocol/contracts/tokenvault/ERC20Vault.sol
```diff
@@ -453,4 +442,22 @@ contract ERC20Vault is BaseVault {
...
+        (bool success, bytes memory data) =
+            address(_token).staticcall(abi.encodeCall(IERC20Metadata.symbol, ()));
+        return success ? data.toString() : "";
+    }
+
+    function _safeName(address _token) private view returns (string memory) {
+        (bool success, bytes memory data) =
+            address(_token).staticcall(abi.encodeCall(IERC20Metadata.name, ()));
+        return success ? data.toString() : "";
+    }
```
adaki2004, line 456: Maybe these 2 (name, symbol) can be put into BaseNFTVault too ?

adaki2004 (approved):

packages/protocol/contracts/tokenvault/ERC20Vault.sol
dantaik: Done

davidtaikocha (approved):

PR #17009, opened 2024-05-06
refactor(protocol): improve base-fee math and remove base fee min value

dantaik:
This PR is based on suggestions from OZ. See below.

## Maxim: My feedback:
 
First and second points are resolved IMO
About the third point, is your opinion that the base fee should not be allowed to go below 0,1 gwei on Taiko L2 ? If so, is there a reasoning behind this number ? It seems a bit arbitrary, for example I see that the last tx on Base has a < 0.1 gwei gas price (see below). My issue was that the computation was basefee(x) = 1/A * exp( x ), and you force basefee(x) >= 1 (1 wei). So I was saying that the multiplication by 1/A could be removed, that way it would naturally be bounded by 1 wei to the downside which would be simpler and would save precision in the exponential. If you want to bound it by `C`, you could do `C * exp(x)`. To be clear it was a nitpick, you likely won't need this precision in practice, but it’s also a bit clearer (and what Ethereum is doing). 

---
## Changes

- Removed the min base-fee design so now in theory, L2 base fee can be as small as 1 wei. We introduced the min base-fee design so that we can witness some changes in base fees on testnets, but maybe a arbitrary value for min base-fee is not a good idea and in practice, a base fee of 0.01 gwei is no different than a base fee of 1 wei for the rollup or our token holders.

- Also changed the base fee math of calculating the base fee from ` basefee(x) = 1/A * exp( x )` to ` basefee(x) = exp( x )`. As said in the deleted comment in packages/protocol/contracts/L2/Lib1559Math.sol, the new math is equivalent to the previous one but offers better precision.

With a test, I can verify that:
- base fee will reach 0.01 gwei if gasExcess is greater than 130 x gasTargetPerL1Block
- base fee will reach 0.1 gwei if gasExcess is greater than 149 x gasTargetPerL1Block
- base fee will reach 1 gwei if gasExcess is greater than  167 x gasTargetPerL1Block
- base fee will reach 10 gwei if gasExcess is greater than  186 x gasTargetPerL1Block
- base fee will reach 100 gwei if gasExcess is greater than  204 x gasTargetPerL1Block


adaki2004 (approved):

Brechtpd (approved):
> Removed the min base-fee design so now in theory, L2 base fee can be as small as 1 wei. We introduced the min base-fee design so that we can witness some changes in base fees on testnets, but maybe a arbitrary value for min base-fee is not a good idea and in practice, a base fee of 0.01 gwei is no different than a base fee of 1 wei for the rollup or our token holders.

There is one potentially good reason for a higher min basefee: a protocol tax to ensure the protocol has some income even when there is not congestion. Now I'm not saying that's good because it increases tx fees for users, but it could be used for that. :)

PR #17023, opened 2024-05-07
chore(bridge-ui): ternary refactor

kirillov13:
Ternary refactor

### DIFF
--- a/packages/bridge-ui/src/components/Bridge/NFTBridgeComponents/IDInput/IDInput.svelte
+++ b/packages/bridge-ui/src/components/Bridge/NFTBridgeComponents/IDInput/IDInput.svelte
@@ -14,7 +14,7 @@
   export let limit = 1;
   // export let state: State = State.DEFAULT;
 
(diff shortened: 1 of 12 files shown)


KorbinianK:
@kirillov13 have you actually tested your code? Because the approval checks are not working correctly...

Adding function calls to reactive statements in svelte 4 is not always ideal. 
It's a bit clunky right now, I know, but once svelte 5 is out, things get better with runes and we can refactor then.

kirillov13:
@KorbinianK 

Yes, I just tested it again and the code works correctly

I don't think there could be any problems with the functions, but I can do refactor without functions if you think it will be better

kirillov13:
@KorbinianK 

Oops, sorry, didn't know about some of the light issues.

There is an option to pass data to a prop by calling a function, what do you think of this solution?

I've tested it now and it seems to work.

image


You can try it in my fork - https://github.com/kirillov13/taiko-mono

PR #17027, opened 2024-05-08
fix(protocol): fix bridge quota processing and make `processMessage` return data

dantaik:
BEGIN_COMMIT_OVERRIDE
feat(protocol): make bridge processMessage return message's status and reason
fix(protocol): fix bridge quota processing in processMessage
END_COMMIT_OVERRIDE

### Before this PR
If the bridge is out of ETH quota, the `processMessage` transaction will fail without marking the message as "retryable", which will keep the message unprocessable and ether locked on the source chain, because this message has to wait for enough quota to be available and cannot be recalled to unlock the funds on the source chain.

### After this PR
If there is not enough quota, the message is marked as retrievable, however, no fees are paid to the bridge processor ,otherwise, a very high `message.fee` can be used to pay` message.destOwner`  to get around of the ETH quota verification. The message can be retried or recalled to unlock funds.


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -228,52 +230,55 @@ contract Bridge is EssentialContract, IBridge {
...
             _proveSignalReceived(signalService, msgHash, _message.srcChainId, _proof);
 
-        uint256 refundAmount;
-        if (_unableToInvokeMessageCall(_message, signalService)) {
-            // Handle special addresses that don't require actual invocation but
-            // mark message as DONE
-            refundAmount = _message.value;
-            _updateMessageStatus(msgHash, Status.DONE);
+        if (!_consumeEtherQuota(_message.value + _message.fee)) {
+            stats.status = Status.RETRIABLE;
```
Brechtpd, line 242: Maybe we should also only allow the owner to make it retriable? Even though whoever processes the message doesn't get any fees so normally it won't happen anyway, but making it retriable would be quite inconvenient so maybe can't hurt to protect against it.

Brechtpd (approved):

packages/protocol/contracts/bridge/Bridge.sol
dantaik: Done. please take another look.

Brechtpd (approved):

davidtaikocha (approved):

PR #17037, opened 2024-05-08
fix(deps): tries to fix Guardian UI CI/CD using secrets

syedcherry:
Currently Guardian UI CI/CD is always failing and I suspect it's related to mis-use of `with` combined with `secrets`.

Reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idsecrets


kimo-ice:
Issue addressed in another PR. 
Efforts appreciated.

PR #17043, opened 2024-05-08
feat(protocol): make bridge processMessage return message's status and reason

dantaik:
@KorbinianK please confirm if you need this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,6 @@ __pycache__/
 .vscode/launch.json
 packages/protocol/config.js
 .vercel
+
+# Idea
+.idea/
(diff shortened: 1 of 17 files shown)


KorbinianK:
@dantaik I need the return value, otherwise I cannot determine whether a transaction will go to retriable via simulateContract. It does not emit events during simulation (`eth_call` does not return events).

But even with the return of message status, I won't know the reason for it. Should we return the reason for the status change as well?

dantaik:
> @dantaik I need the return value, otherwise I cannot determine whether a transaction will go to retriable via simulateContract. It does not emit events during simulation (`eth_call` does not return events).
> 
> But even with the return of message status, I won't know the reason for it. Should we return the reason for the status change as well?

Added a reason to the return of `processMessage`

KorbinianK (approved):

PR #17049, opened 2024-05-08
feat(protocol): change to transfer-and-burn pattern with NFT vaults

adaki2004:
To remove additional trust assumption, and do not allow vaults (if compromised) to arbitrarily burn tokens. Adapt the ‘transfer and burn’ pattern for NFTs as well (as we did for ERC20).


dantaik (changes requested):

dantaik:
@KorbinianK @cyberhorsey I think you guys need to know about this change (similar to what we did to BridgedERC20 vaults). Now you always have to get approvals first before burning bridged NFTs, just like what you do with canonical NFTs.

adaki2004:
> @KorbinianK @cyberhorsey I think you guys need to know about this change (similar to what we did to BridgedERC20 vaults). Now you always have to get approvals first before burning bridged NFTs, just like what you do with canonical NFTs.

@KorbinianK 
Yes, so for bridged tokens, while bridging back we now require `approve()` for `erc721` and `setApprovalForAll()` for `erc1155`.

packages/protocol/contracts/tokenvault/BridgedERC721.sol
```diff
@@ -65,19 +65,15 @@ contract BridgedERC721 is EssentialContract, ERC721Upgradeable {
...
-        address _account,
-        uint256 _tokenId
-    )
+    function burn(uint256 _tokenId)
         external
         whenNotPaused
         onlyFromNamed(LibStrings.B_ERC721_VAULT)
         nonReentrant
     {
         // Check if the caller is the owner of the token.
```
dantaik, line 75:
```suggestion
        // Check if the caller is the owner of the token. Somehow this is not done inside the _burn() function below.
```

dantaik (approved):

davidtaikocha (approved):

PR #17067, opened 2024-05-09
refactor(protocol): remove bridged token impl dependencies from vaults

dantaik:
With a few new interfaces, vaults no longer depend on any bridged token implementation but their interfaces, which will be helpful to avoid mistakes like https://github.com/taikoxyz/taiko-mono/commit/cfc8fa8f00b1b7fc7283cce0fd936e0d8bfe0f8c


packages/protocol/contracts/tokenvault/ERC1155Vault.sol
```diff
@@ -43,7 +43,7 @@ contract ERC1155Vault is BaseNFTVault, ERC1155ReceiverUpgradeable {
             if (_op.amounts[i] == 0) revert VAULT_INVALID_AMOUNT();
         }
         // Check token interface support
-        if (!_op.token.supportsInterface(ERC1155_INTERFACE_ID)) {
+        if (!_op.token.supportsInterface(type(IERC1155).interfaceId)) {
```
dantaik, line 46: verified this works!

Brechtpd (approved):

davidtaikocha (approved):

PR #17078, opened 2024-05-10
feat(protocol): allow resetting genesis hash on L1 before 1st block is proposed

dantaik:
We need to reset genesis in TaikoL1 on Ethereum to indirectly upgrade its L2 contracts

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibVerifying.sol
+++ b/packages/protocol/contracts/L1/libs/LibVerifying.sol
@@ -61,8 +61,9 @@ library LibVerifying {
 
     function resetGenesisHash(TaikoData.State storage _state, bytes32 _genesisBlockHash) internal {
         // Do not revert as this function is called inside a reinitializer function.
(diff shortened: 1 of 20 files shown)


packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -83,6 +83,26 @@ library LibVerifying {
         });
     }
 
+    function resetGenesisHash(TaikoData.State storage _state, bytes32 _genesisBlockHash) internal {
```
dantaik, line 86: we do not revert in this function so that if we have another reinitializer, the sequence number won't mess up.

davidtaikocha (approved):

adaki2004 (approved):
I think it makes sense, but only in case if we want to deploy all contracts while still being able to do last minute change - which we should not imo at this point.

My .02: we should have a process for upgrading, and documented. we cannot do these kind of random, rushed merges/changes at mainnet either. We should practice some relax.

PR #17099, opened 2024-05-12
feat(protocol) :add Dockerfile for packages/protocol use for future testing

quangtuyen88:
- Purpose reduce step install pnpm, nodejs, forge 
- Better for future testing
- Can be used for quick on-chain RA .


quangtuyen88:
@smtmfft @d1onys1us Can you help review this later ? Thanks

PR #17100, opened 2024-05-12
feat(protocol): add Dockerfile/docker compose for packages/protocol 

quangtuyen88:
- Purpose reduce step install pnpm, nodejs, forge
- Better for future testing
- Can be used for quick on-chain RA .

For onchain RA can run with 
```
export FORK_URL=your_holesky_rpc
export QUOTE=your_quote
export PRIVATE_KEY=0xyour_private_key_start
docker compose up
```
I have tested working as expect 
Screenshot 2024-05-12 at 20 16 27


quangtuyen88:
@smtmfft  @d1onys1us Can you help review this later ? Thanks

packages/protocol/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
...
+
+services:
+  ra:
+    build: .
+    container_name: onchain-ra
+    environment:
+      - FORK_URL=${FORK_URL}
+      - PRIVATE_KEY=${PRIVATE_KEY}
+      - QUOTE=${QUOTE}
+    entrypoint: [ "/bin/bash", "-c", "PRIVATE_KEY=${PRIVATE_KEY} ./script/config_dcap_sgx_verifier.sh --quote ${QUOTE}"]
```
dantaik, line 11: so this is only for testing the sgx verifier? If so, shall we create a sub folder for this docker-compose file? Keeping it in the project's root folder doesn't make sense to me.

PR #17110, opened 2024-05-13
fix(repo): re-add the lockfile

KorbinianK:
Lockfiles should always be committed, not ignored or removed. 

Just run pnpm install and there are no conflicts when merging. In fact, the opposite may happen when we don't have it.

Reference: https://pnpm.io/git#lockfiles

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,6 @@ typings/
 
 # Optional npm cache directory
 .npm
-pnpm-lock.yaml
 
 # Optional eslint cache
 .eslintcache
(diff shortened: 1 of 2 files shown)


dantaik:
Shall we check in only sub-package lock files but not the root one?

KorbinianK:
> Shall we check in only sub-package lock files but not the root one?

@dantaik a mono repo has one single lockfile.

dantaik (approved):

bearni95 (approved):

PR #17114, opened 2024-05-13
docs(protocol): prettify formulas

Sheisuka:
# Prettify formulas in the description of EIP1559
I have made some formulas look better and a bit more readable
You can see changes below (a picture with a blue background is a previous version)
## First change
prev1
new1

## Second change
prev2
new2


dantaik:
I don't think this change is necessary and I prefer to keep the change log cleaner.

PR #17120, opened 2024-05-13
feat(relayer): check event status one more time before sending call

cyberhorsey:
its possible the message could be processed befoer we wait for confirmations, get proof, and do all necessary checks. We should make sure the message is still processable right before we send the call, to avoid reverts, especially on mainnet.

Also fixes a bug with the retry headers, and fixes broekn CICD.


packages/relayer/processor/process_message.go
```diff
@@ -431,6 +431,25 @@ func (p *Processor) sendProcessMessageCall(
 		estimatedCost = gasUsed * (baseFee.Uint64() + gasTipCap.Uint64())
 	}
 
+	// we should check event status one more time, after we have waiting for
+	// confirmations, and after we have gnerated proof. its possible another relayer
```
RogerLamTd, line 435:
```suggestion
	// confirmations, and after we have generated proof. its possible another relayer
```

RogerLamTd (approved):

k-kaddal (approved):

k-kaddal (approved):

davidtaikocha (approved):

PR #17169, opened 2024-05-14
feat(protocol): refactor tier providers and added minority-guardian provers to all providers

davidtaikocha:
Note that tier providers are no longer upgradeable.


packages/protocol/contracts/L1/tiers/DevnetTierProvider.sol
```diff
@@ -30,6 +30,17 @@ contract DevnetTierProvider is EssentialContract, ITierProvider {
             });
         }
 
+        if (_tierId == LibTiers.TIER_GUARDIAN_MINORITY) {
```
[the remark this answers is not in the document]
dantaik: Maybe also delete TierProviderV1?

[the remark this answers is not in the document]
davidtaikocha: Deleted `TierProviderV1 ` and `HeklaTierProvider`, they are duplicated, but for this one, the lowest tier (`TIER_OPTIMISTIC`) is different from `TierProviderV2`(`TIER_SGX`), we still need it in some cases.

dantaik (approved):

dantaik (approved):

packages/protocol/contracts/L1/tiers/DevnetTierProvider.sol
```diff
@@ -30,6 +30,17 @@ contract DevnetTierProvider is EssentialContract, ITierProvider {
...
+            return ITierProvider.Tier({
+                verifierName: LibStrings.B_TIER_GUARDIAN_MINORITY,
+                validityBond: 500 ether, // TKO
+                contestBond: 3280 ether, // =500TKO * 6.5625
+                cooldownWindow: 60, //1 hours
```
Brechtpd, line 38: I think this is low because the minority one isn't so secure and this doesn't leave a lot of time

dantaik: Brecht, do you mean the cooldown window? If we change this value, then we should apply the same change to all other Provider files. I checked all cooldownWindow for guardian provers (minority and majority) are 1 hour. 

What value do you think makes more sense?

Brechtpd: yeah just the cooldown window. Because for the minority prover, if only 1 or 2 guardians are compromised there's only 1 hour for the majority guardians to override it, which I think is a very short amount of time because still the risk of the L1 tx of the majority guardians to be censored (or be subject to other congestion) so it may take a while for the tx to get included.

For majority guardian I think 1 hour is fine, but for the minority guardian I believe we should be closer to the other normal proof cooldown periods, I think at least 6 hours would be good.

dantaik: Lets change it to 6h then.

dantaik (approved):

Brechtpd (approved):

PR #17170, opened 2024-05-14
fix(protocol): fix tier ID check issue in `GuardianProver.approve()`

davidtaikocha:


adaki2004 (approved):

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -76,10 +76,6 @@ contract GuardianProver is IVerifier, Guardians {
         nonReentrant
         returns (bool approved_)
     {
-        if (_proof.tier != LibTiers.TIER_GUARDIAN) {
```
dantaik, line 79: I think we should remove this check.

dantaik (approved):

dantaik (approved):

adaki2004 (approved):

PR #17194, opened 2024-05-15
feat(supplementary-contracts): allow any users to withdraw to recipient

dantaik:
This change is made for VCs. Some of them are not comfortable to give their private keys to internal personale to perform the withdrawal. This change will allow any address to withdraw tokens to the recipient address.

### DIFF
--- a/packages/supplementary-contracts/contracts/tokenUnlocking/TokenUnlocking.sol
+++ b/packages/supplementary-contracts/contracts/tokenUnlocking/TokenUnlocking.sol
@@ -93,12 +93,16 @@ contract TokenUnlocking is OwnableUpgradeable, ReentrancyGuardUpgradeable {
 
     /// @notice Withdraws all withdrawable tokens.
     /// @param _to The address the token will be sent to.
-    function withdraw(address _to) external onlyRecipient nonReentrant {
+    function withdraw(address _to) external nonReentrant {
         uint256 amount = amountWithdrawable();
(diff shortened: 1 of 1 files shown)


davidtaikocha (approved):

Brechtpd (approved):

packages/supplementary-contracts/contracts/tokenUnlocking/TokenUnlocking.sol
```diff
@@ -93,12 +93,16 @@ contract TokenUnlocking is OwnableUpgradeable, ReentrancyGuardUpgradeable {
...
-    function withdraw(address _to) external onlyRecipient nonReentrant {
+    function withdraw(address _to) external nonReentrant {
         uint256 amount = amountWithdrawable();
         if (amount == 0) revert NOT_WITHDRAWABLE();
 
-        amountWithdrawn += amount;
         address to = _to == address(0) ? recipient : _to;
+        if (to != recipient && msg.sender != recipient) {
+            revert PERMISSION_DENIED();
+        }
```
adaki2004, line 103: What is the modification compared to the previous version ? I think it is a bug.

Now the onlyRecipient modifier removed, which actually checks this:
`msg.sender != recipient`

So how any users can withdraw ?

dantaik: anyone can help withdraw to the recipient address. if the to is not the recipient address, then it must be the recipient address to transact.

PR #17215, opened 2024-05-16
feat(protocol): add succint's on-chain verifier

adaki2004:
SP1VerifierBase's code directly copied from here:
https://github.com/succinctlabs/sp1-project-template/blob/main/contracts/src/SP1Verifier.sol

Only changes:
- renamed from `SP1Verifier` to `SP1VerifierBase`
- pargma changed from .8.20 to .8.24

Template implementation used as a pattern:
https://github.com/succinctlabs/sp1-project-template/blob/main/contracts/src/Fibonacci.sol

@Brechtpd please have a look before proceeding further, there is a at line73.


packages/protocol/contracts/verifiers/SP1Verifier.sol
```diff
@@ -0,0 +1,71 @@
...
+        if (!isProgramTrusted[programVKey]) {
+            revert SP1_INVALID_PROGRAM_VKEY();
+        }
+
+        uint64 chainId = ITaikoL1(resolve(LibStrings.B_TAIKO, false)).getConfig().chainId;
+        bytes32 hash = LibPublicInput.hashPublicInputs(
+            _tran, address(this), address(0), _ctx.prover, _ctx.metaHash, chainId
+        );
+
+        // @Brecht: Is 'hash' var the public value ? OR the input params of the
```
Brechtpd, line 66: Currently it is neither directly because we also output the calculated block header so we can directly check all the calculated block parameters. So we don't want to actually do this for a proof, but currently it is nice to have it there for debugging purposes. So not sure yet how we can have the best of both worlds.

But I think it makes sense to let `hashPublicInputs` return the bytes instead of already the keccak hashed value, gives some more flexibility on how the public inputs are encoded.

PR #17224, opened 2024-05-17
chore(bridge-ui): suggest stable coin message change

dantaik:
Joaquin, our liquidity campaign with startgate is only related to USDC or it also include USDT?

@KorbinianK In case it's USDC, I think we should only display this warning for USDC. I also think we should change the message as I suggested. "Currently we are not supporting a native 1:1 conversion" is confusing and kind of scary.


KorbinianK:
@dantaik The warning will only show up for tokens we configured with the attributes: `{"stablecoin": true, "supported": false }` So it can be used for others, but if we only defined USDC, then it's the only one with this message.

We could rephrase it and say  something like this?

> "Please be sure that the dapp on Taiko, that you want to use your stablecoin with, supports this version: < link to USDC.t >. Currently we run a campaing with Stargate, which uses a different contract. Use < link > if your dapp requires their version."

PR #17251, opened 2024-05-19
feat(protocol,supplementary-contracts): relocate & allow TokenUnlock to deploy and own ProverSets

dantaik:
Allow Taiko's investors to use their locked tokens as bonds.


Brechtpd:
Wouldn't this allow people to withdraw a big chunk of their tokens by getting slashed and then getting the reward?

 I guess would have to think about the specifics and some protocol details, but for example if you can prove a block wrongly and directly also contest and prove it correctly in the same tx you would be able to extract a big chunk without any risks of anybody else getting the tokens. So sure you still lose a big chunk, but you also don't have to wait 4 years to get your tokens.

dantaik:
> Wouldn't this allow people to withdraw a big chunk of their tokens by getting slashed and then getting the reward?
> 
> I guess would have to think about the specifics and some protocol details, but for example if you can prove a block wrongly and directly also contest and prove it correctly in the same tx you would be able to extract a big chunk without any risks of anybody else getting the tokens. So sure you still lose a big chunk, but you also don't have to wait 4 years to get your tokens.

Good points. I believe technically it is possible. However, I think practically this won't happen for two reasons:

- 12.5% token will be lost during the contest-then-re-prove process.
- all of our blocks requires SGX, they must be able to generate incorrect but verifiable sgx proofs, then overwrite them with valid ones.

Brechtpd:
Yeah true that if SGX is required for each block it shouldn't be possible, so should be good.

Brechtpd (approved):

davidtaikocha (approved):

Brechtpd:
I'm not 100% sure about the behavior for this in the protocol right now, but the liveness bond slashing could also be a potential way to avoid the locking. But that one is probably a lot trickier to pull off risk free.

dantaik:
> liveness

As of now liveness bond is the smallest bond compared with all other bonds. When liveness bond is returned, another bond must be placed.

packages/protocol/contracts/team/proving/ProverSet.sol
```diff
@@ -1,63 +1,92 @@
...
 
         emit ProverEnabled(_prover, _isProver);
     }
 
-    /// @notice Withdraws Taiko tokens back to the owner address.
-    function withdraw(uint256 _amount) external onlyOwner {
-        IERC20(resolve(LibStrings.B_TAIKO_TOKEN, false)).transfer(owner(), _amount);
+    /// @notice Withdraws Taiko tokens back to the admin address.
+    function withdrawToAdmin(uint256 _amount) external onlyAuthorized {
+        IERC20(resolve(LibStrings.B_TAIKO_TOKEN, false)).transfer(admin, _amount);
```
adaki2004, line 75: I guess we need to put this function in the `TokenUnlock` too, if we want to allow unlocking people calling it ?

packages/protocol/contracts/team/proving/ProverSet.sol
```diff
@@ -1,63 +1,92 @@
...
-        address assignmentHook = resolve(LibStrings.B_ASSIGNMENT_HOOK, false);
-        tko.approve(assignmentHook, type(uint256).max);
+        IERC20 tko = IERC20(resolve(LibStrings.B_TAIKO_TOKEN, false));
+        tko.approve(resolve(LibStrings.B_TAIKO, false), type(uint256).max);
+        tko.approve(resolve(LibStrings.B_ASSIGNMENT_HOOK, false), type(uint256).max);
     }
 
     /// @notice Enables or disables a prover.
-    function enableProver(address _prover, bool _isProver) external onlyOwner {
+    function enableProver(address _prover, bool _isProver) external onlyAuthorized {
```
adaki2004, line 66: I guess we need to put this function in the `TokenUnlock` too, if we want to allow unlocking people calling it.

packages/protocol/contracts/team/proving/ProverSet.sol
```diff
@@ -1,63 +1,92 @@
...
+    function proveBlock(uint64 _blockId, bytes calldata _input) external onlyProver nonReentrant {
         emit BlockProvenBy(msg.sender, _blockId);
         ITaikoL1(resolve(LibStrings.B_TAIKO, false)).proveBlock(_blockId, _input);
     }
 
+    /// @notice Delegates token voting right to a delegatee.
+    /// @param _delegatee The delegatee to receive the voting right.
+    function delegate(address _delegatee) external onlyAuthorized nonReentrant {
+        ERC20VotesUpgradeable(resolve(LibStrings.B_TAIKO_TOKEN, false)).delegate(_delegatee);
+    }
```
adaki2004, line 88: This one too, shall be put also to `TokenUnlocking` (to allow calling into this function) ?

packages/protocol/contracts/team/proving/ProverSet.sol
dantaik: No need, the `onlyAuthorized` allows recipient to call this function.

packages/protocol/contracts/team/proving/ProverSet.sol
dantaik: why the unlocking needs to know about prover enabling...

packages/protocol/contracts/team/proving/ProverSet.sol
dantaik: Sorry I don't understand this feedback, why duplicate logics in two contracts?

packages/protocol/contracts/team/proving/ProverSet.sol
adaki2004: Yes, okay i see the modifier is like that:
`msg.sender != IHasRecipient(admin).recipient()`

packages/protocol/contracts/team/proving/ProverSet.sol
adaki2004: Yes, okay i see the modifier is like that:
`msg.sender != IHasRecipient(admin).recipient()`

packages/protocol/contracts/team/proving/ProverSet.sol
adaki2004: Yep, i missed the modifier change. Seems no issue indeed.

adaki2004 (approved):

PR #17253, opened 2024-05-20
feat(taiko-client): integrate `ProverSet` in client

davidtaikocha:


packages/taiko-client/pkg/rpc/utils.go
```diff
@@ -158,7 +159,8 @@ func GetBlockProofStatus(
 		}, nil
 	}
 
-	if proverAddress == transition.Prover {
+	if proverAddress == transition.Prover ||
+		(proverSetAddress != ZeroAddress && proverAddress == proverSetAddress) {
```
YoGhurt111, line 163: Would this explain the block's proof has already been submitted？

davidtaikocha: Yep if a prover is behind a proverSet, the proof will be sent by: https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/team/proving/ProverSet.sol#L79

packages/taiko-client/prover/event_handler/block_proposed.go
```diff
@@ -31,6 +31,7 @@ var (
 type BlockProposedEventHandler struct {
 	sharedState           *state.SharedState
 	proverAddress         common.Address
+	proverSetAddress      common.Address
```
mask-pp, line 34: `proverSetAddress` is not initialized.

mask-pp (approved):

YoGhurt111 (approved):

PR #17272, opened 2024-05-20
feat(protocol): deploy `ProverSet` in `DeployOnL1` script

davidtaikocha:


dantaik:
David, I think we can simply deploy a ProverSet by `new ProverSet` without proxy as impl. Later the TokenUnlock contract will deploy a proxy before the impl.

davidtaikocha:
> David, I think we can simply deploy a ProverSet by `new ProverSet` without proxy as impl. Later the TokenUnlock contract will deploy a proxy before the impl.

After the discussion, we will still deploy `ProverSet` behind a proxy.

dantaik (approved):

smtmfft (approved):

PR #17275, opened 2024-05-20
ci(repo): revert and fix workflow triggers

dionysuzx:
This reverts my change [here](https://github.com/taikoxyz/taiko-mono/pull/17247) to remove the pull_request event trigger and makes it better.

@davidtaikocha would you prefer that we build a docker image on PR commits for taiko-client as well? This is the way it is set up in the eventindexer and relayer. if you would want this change, I can add it, but it will increase the amount of pushes to the docker registry, I don't know what the limit is or if this is preferred.

Right now it's as it was before, only on main, and integration tests on PR.

### DIFF
--- a/.github/workflows/blobstorage.yml
+++ b/.github/workflows/blobstorage.yml
@@ -5,6 +5,9 @@ on:
     branches: [main]
     paths:
       - "packages/blobstorage/**"
+  pull_request:
+    paths:
+      - "packages/blobstorage/**"
 
 jobs:
   test-blobstorage:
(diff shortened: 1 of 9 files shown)


RogerLamTd (approved):

cyberhorsey (approved):

davidtaikocha:
> @davidtaikocha would you prefer that we build a docker image on PR commits for taiko-client as well? This is the way it is set up in the eventindexer and relayer. if you would want this change, I can add it, but it will increase the amount of pushes to the docker registry, I don't know what the limit is or if this is preferred.

emmm for client, i think its ok to leave the build process as it is for now.

PR #17281, opened 2024-05-21
feat(taiko-client): improve progress tracker

### LINKED ISSUE #17160: feat(taiko-client): stop triggering another round of beacon sync when the node is still syncing

davidtaikocha:


packages/taiko-client/driver/chain_syncer/beaconsync/progress_tracker.go
```diff
@@ -162,16 +163,34 @@ func (t *SyncProgressTracker) ClearMeta() {
...
 	defer t.mutex.RUnlock()
 
+	// If the beacon sync has not been triggered yet, we will simply trigger it.
 	if !t.triggered {
 		return true
 	}
 
-	return t.lastSyncedBlockID != nil && t.lastSyncedBlockID != newID
+	if t.lastSyncedBlockID == nil {
+		return false
```
mask-pp, line 180:
```suggestion
		return true
```

mask-pp:
Actually `lastSyncedBlockID` is used for recording the current syncing `blockID`, how about rename to `lastSyncingBlockID`?

packages/taiko-client/driver/chain_syncer/beaconsync/progress_tracker.go
YoGhurt111: Yep, it should be `true`

YoGhurt111 (approved):

packages/taiko-client/driver/chain_syncer/beaconsync/progress_tracker.go
davidtaikocha: changed

mask-pp (approved):

PR #17284, opened 2024-05-21
fix(protocol): fix bridge bugs in `getMessageMinGasLimit`

dantaik:
Fix two bugs in Bridge.getMessageMinGasLimit reported by OZ:

> The abi encoding of A = (Message calldata msg) is actually 10 * 32 bytes (not 7 * 32 as elements are not packed) + 32 bytes (A is a dynamic tuple, offset to first elements) + 32 bytes (offset to last bytes element of Message ) + 32 bytes (padded encoding of length of Message.data + dataLength (padded to 32 bytes) = 13 * 32 + (dataLength // 32 * 32) + 32. 

> For example, if dataLength = 4, the actual abi encoding is 448 bytes long, whereas dataLength + 256 which is the formula used in the code gives only 260 bytes long (code [here](https://gist.github.com/0xmp/de96284548f62228735116db5dc57e28))

> I don't understand the [division by 16](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L452C60-L452C64): I assume the 16 comes from the gas cost of a nonzero calldata byte, but then why divide the number of bytes by 16 instead of multiplying ?


adaki2004:
Seems OK, but we shall immediately try it out on testnet to see in action - as currently (even tho seems inaccurate, it works :) )

adaki2004 (approved):

dantaik:
> Seems OK, but we shall immediately try it out on testnet to see in action - as currently (even tho seems inaccurate, it works :) )
> 
> And 2 tests (related to gas) are failing.

Fixed

cyberhorsey:
> Seems OK, but we shall immediately try it out on testnet to see in action - as currently (even tho seems inaccurate, it works :) )
> 
> And 2 tests (related to gas) are failing.

it works in testnet because we are multiplying the return value offchain

cyberhorsey (approved):

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -448,9 +448,16 @@ contract Bridge is EssentialContract, IBridge {
...
-            return GAS_RESERVE + uint32((dataLength + 256) >> 4);
+            // The abi encoding of A = (Message calldata msg) is 10 * 32 bytes
+            // + 32 bytes (A is a dynamic tuple, offset to first elements)
+            // + 32 bytes (offset to last bytes element of Message)
+            // + 32 bytes (padded encoding of length of Message.data + dataLength (padded to 32
+            // bytes)
+            // = 13 * 32 + (dataLength / 32 * 32) + 32.
+            // non-zero calldata cost per byte is 16.
+
+            uint256 dataCost = (dataLength / 32 * 32 + 448) << 4;
```
Brechtpd, line 459: Does using a bitshift actually save gas? I would think the compiler just handles it

Brechtpd (approved):

packages/protocol/contracts/bridge/Bridge.sol
dantaik: I think it does, but not verified. We also got such optimization suggestions from code4rena auditors.

PR #17285, opened 2024-05-21
refactor(protocol): merge Guardians.sol into GuardianProver.sol

dantaik:
This is to prepare another feature that David C and myself discussed today: enabling pause block proving if two guardian provers submitted conflicting proofs.

This PR has no logical change.


packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -3,23 +3,37 @@ pragma solidity 0.8.24;
...
+
+    /// @notice The version of the guardians
+    /// @dev Slot 4
+    uint32 public version;
 
-    uint256[50] private __gap;
+    /// @notice The minimum number of guardians required to approve
+    uint32 public minGuardians;
+
+    uint256[46] private __gap;
```
adaki2004, line 36: Hmm, will we be fully storage compatible if we leave out `GuardianProver` contract's 50 uint256-long slot ? I guess so since they were unsued, just flagging here.

dantaik: Yes, fully compatiable as long as there is no sub-contract of this contract deployed/used.

adaki2004 (approved):

davidtaikocha (approved):

PR #17286, opened 2024-05-21
feat(protocol): allow guardian prover to pause block proving & verification

dantaik:
Enabled guardian prover -- once set as the chain pauser -- to pause chain proving if two guardians submits different proofs for the same block in one proving session. If a proof has enough guardian approval and had been used to prove a block, the session history are cleared, so different proof can be submitted again by the guardians for the same block in another "session".

@davidtaikocha  I think we should keep this feature disabled (do not set chain pauser to the guardian prover until we fully test this behavior in Hekla.


packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -153,16 +170,38 @@ contract GuardianProver is IVerifier, EssentialContract {
...
-            ITaikoL1(resolve(LibStrings.B_TAIKO, false)).proveBlock(
-                _meta.id, abi.encode(_meta, _tran, _proof)
-            );
+        if (conflicting) {
+            latestProofHash[_version][_meta.id] = proofHash;
+            emit ConflictingProofs(_meta.id, msg.sender, currProofHash, proofHash, pauseProving);
+        }
+
+        if (pauseProving) {
+            ITaikoL1(resolve(LibStrings.B_TAIKO, false)).pauseProving(true);
```
Brechtpd, line 192: I think we should also allow any guardian to directly pause as well because I think it will be useful to support all reasons to pause the bridge (for example too many guardian proofs necessary because proofs are broken). They basically already have the power to so by using (or abusing) this system so I think should be okay to do so.

Brechtpd (approved):

packages/protocol/contracts/L1/provers/GuardianProver.sol
dantaik: Can do. Please take a look at https://github.com/taikoxyz/taiko-mono/pull/17293

dantaik: If https://github.com/taikoxyz/taiko-mono/pull/17293 is good, I perfer to merge it later after this one is merged to main

dantaik: Done, take a look.

packages/protocol/contracts/L1/provers/GuardianProver.sol
```diff
@@ -153,17 +181,51 @@ contract GuardianProver is IVerifier, EssentialContract {
...
+    /// @notice Pauses or unpauses the chain proving.
+    /// @param _pause true to pause, false to unpause.
+    function pauseTaikoProving(bool _pause) external whenNotPaused {
+        if (guardianIds[msg.sender] == 0) revert GP_INVALID_GUARDIAN();
+
+        if (address(this) != resolve(LibStrings.B_CHAIN_WATCHDOG, true)) {
+            revert GV_PERMISSION_DENIED();
+        }
+
+        ITaikoL1(resolve(LibStrings.B_TAIKO, false)).pauseProving(_pause);
```
Brechtpd, line 228: Maybe we shouldn't allow unpausing?

dantaik: Good point, actually the chain_watchdog can only unpause. I'll change the code to make sure unpausing is now an option here.

dantaik: Done

Brechtpd (approved):

davidtaikocha (approved):

PR #17341, opened 2024-05-25
fix(protocol): add post genesis L2 scripts

dantaik:


packages/protocol/script/PostMainnetLaunch.s.sol
```diff
@@ -0,0 +1,40 @@
...
+        vm.startBroadcast();
+
+        ERC20Vault.CanonicalERC20 memory canonical;
+        canonical.chainId = 1;
+
+        canonical.addr = 0x10dea67478c5F8C5E2D90e5E9B26dBe60c54d800;
+        canonical.decimals = 18;
+        canonical.symbol = "TKO";
+        canonical.name = "Taiko Token";
+        ERC20Vault(erc20Vault).changeBridgedToken(canonical, bridgedTKO);
```
adaki2004, line 28: Daniel i think we have this script embeded into the deployment BridgedTKO.
`DeployAndChangeTaikoTokenOnL2` 

https://github.com/taikoxyz/taiko-mono/blob/fa481c1b56ea63c51fc1d79313d0e192fa7501e8/packages/protocol/script/DeployAndChangeTaikoTokenOnL2.s.sol#L40

packages/protocol/script/PostMainnetLaunch.s.sol
```diff
@@ -0,0 +1,40 @@
...
+
+        canonical.addr = 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48;
+        canonical.decimals = 6;
+        canonical.symbol = "USDC";
+        canonical.name = "USD Coin";
+        ERC20Vault(erc20Vault).changeBridgedToken(canonical, bridgedUSDC);
```
adaki2004, line 34: Also USDC deployment script has it:
https://github.com/taikoxyz/USDC/blob/c58a12efbfc61ecc07c01fa311942b4c06e6f6f1/script/DeployUSDC.s.sol#L39

packages/protocol/script/PostMainnetLaunch.s.sol
adaki2004: And actually BridgedTKO gotta / or shall be second as that one also `unpausing` the bridge.

adaki2004 (approved):
lgtm

PR #17356, opened 2024-05-27
feat(protocol): enable Tier Provider to accept blockId as input

dantaik:
No need to deploy/upgrade soon.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -156,7 +156,7 @@ library LibProposing {
 
         // Use the difficulty as a random number
         meta_.minTier = ITierProvider(_resolver.resolve(LibStrings.B_TIER_PROVIDER, false))
-            .getMinTier(uint256(meta_.difficulty));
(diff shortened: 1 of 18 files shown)


dantaik:
If we need to upgrade our protocol, for example changing EVM from shanghai to cancun, we need to support different tier configurations based on block numbers. This is current missing.

Brechtpd:
I think for our provers we will always support all different HF versions like a node does (because we'll reuse the same code), and in the proof the correct HF will be enabled based on the L2 block number/timestamp. If that is the case, do we need multiple tiers based on block number? I think before we update we just update the verifiers so they support both the current and new HF, and so not really any other changes required I think.

dantaik:
> I think for our provers we will always support all different HF versions like a node does (because we'll reuse the same code), and in the proof the correct HF will be enabled based on the L2 block number/timestamp. If that is the case, do we need multiple tiers based on block number? I think before we update we just update the verifiers so they support both the current and new HF, and so not really any other changes required I think.

"If that is the case, do we need multiple tiers based on block number" may actually be a better idea.

PR #17389, opened 2024-05-28
feat(nfts): mainnet deployments

bearni95:
- [x] Mainnet deployment scripts for `taikoon` and `snaefell`
- [x] Blacklist adaptations for `snaefell`


adaki2004 (approved):
LGTM but only could view from mobile as i’m not in feomt of my PC

dantaik (approved):

PR #17421, opened 2024-05-29
fix(protocol): mitigate Bridge proof size memory expansion attack

dantaik:
Please see [L-22 Diff-Audit: Message Processors Can Control Gas and Fee Consumption](https://defender.openzeppelin.com/v2/#/audit/306ca593-6cb7-4b7f-92d4-ba1e41ac9e56/issues/L-22).

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -19,9 +19,6 @@ contract Bridge is EssentialContract, IBridge {
     ///@dev The max proof size for a message to be processable by a relayer.
     uint256 public constant PROOF_BYTES_THREAHOLD = 240_000;
 
-    ///@dev The max message.data size for a message to be processable by a relayer.
-    uint256 public constant MESSAGE_DATA_THRESHOLD = 8192;
-
     using Address for address;
     using LibMath for uint256;
     using LibAddress for address;
@@ -92,7 +89,6 @@ contract Bridge is EssentialContract, IBridge {
     error B_INVALID_VALUE();
     error B_INSUFFICIENT_GAS();
     error B_MESSAGE_NOT_SENT();
-    error B_MESSAGE_DATA_TOO_LARGE();
     error B_OUT_OF_ETH_QUOTA();
     error B_PERMISSION_DENIED();
     error B_PROOF_TOO_LARGE();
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -16,6 +16,8 @@ import "./IQuotaManager.sol";
 /// L1 and L2 may be different.
 /// @custom:security-contact security@taiko.xyz
 contract Bridge is EssentialContract, IBridge {
+    uint256 public constant MAX_PROOF_BYTES = 240_000;
```
dantaik, line 19: Please do not merge, we still need to figure out this value.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -16,6 +16,9 @@ import "./IQuotaManager.sol";
 /// L1 and L2 may be different.
 /// @custom:security-contact security@taiko.xyz
 contract Bridge is EssentialContract, IBridge {
+    ///@dev The max proof size for a message to be processable by a relayer.
+    uint256 public constant PROOF_BYTES_THREAHOLD = 240_000;
```
dantaik, line 20: need to config this constant

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -627,73 +641,43 @@ contract Bridge is EssentialContract, IBridge {
...
-
-        // The previous implementation is  `return gasleft() * 63) >> 6 >= gasLimit_`.
-    }
-
-    function _checkStatus(bytes32 _msgHash, Status _expectedStatus) private view {
-        if (messageStatus[_msgHash] != _expectedStatus) revert B_INVALID_STATUS();
-    }
+            // https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/vm/gas.py#L128
+            uint256 words = _dataLength / 32 + 1;
+            uint256 memoryGasCost = words * 3 + words * words / 512;
```
Brechtpd, line 665: I don't think we need to add this check here because we don't know if the memory expansion is required for this data inside the invocation call, and it's fully controlled by the user so nothing can be manipulated.

Also the `words * words / 512` part at least should be removed because that part should be done on the total memory size when it's done which is unknown here, so can't calculate it at this point.

Brechtpd (commented):
Not sure if we should add this because it's almost impossible to exploit, but I guess some simple check on the proof size also doesn't complicate the code much.

packages/protocol/contracts/bridge/Bridge.sol
dantaik: Sorry @Brechtpd , could you move this comment to https://github.com/taikoxyz/taiko-mono/pull/17422. I'm closing this issue in favor of another one.

PR #17422, opened 2024-05-29
feat(protocol): improve Bridge `gasleft()` validation

dantaik:
Resolving  [L-23 Diff-Audit: Inaccurate Gas Limit on Message Invocation](https://defender.openzeppelin.com/v2/#/audit/306ca593-6cb7-4b7f-92d4-ba1e41ac9e56/issues/L-23)

--- 
Users can send cross-chain messages by calling [sendMessage](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L126C14-L126C25) on the Bridge contract. An optional gas limit can be defined for this message. In that case, if a user wants their message to be [executed with at least A gas](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L488), message.gasLimit has to be at least A + minGas, where minGas depends on the [message length plus a flat gas amount](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L452).

When a message is processed by a third party, the gas remaining before invoking the user's message is validated to be [greater than](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L611-L613) 64 / 63 * A. This is because of [EIP-150](https://eips.ethereum.org/EIPS/eip-150), which silently caps the amount of gas sent in external calls to 63 / 64 * gasleft(). If such a check was not present, it would be possible for the gas to be capped by EIP-150, executing the call with less than A gas.

However, while the intention was right, a gas limit of A is not guaranteed. This is due to the gas expenses that accumulate between the [EIP-150 check](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L611-L613) and the actual [message execution on the target](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L488):

costs of memory expansion
account access
transfers
opcode costs
Because 1 / 64 * gasleft() has to be enough to execute the [rest of the processMessage function](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L265-L296) without running out of gas, a gas limit below A on the target is only possible for high gas limits. A [proof of concept](https://gist.github.com/0xmp/4da10ab59c04f71cdb9557ff320f9fc2) shows the issue in practice.

Consider addressing the above to build a more predictable bridge for users and projects sending cross-chain messages. What follows is an example of how this could be achieved.

The goal of the computation is to ensure that EIP-150 does not silently cap the amount of gas sent with the external call to be less than A. We note:

memory_cost: the amount of gas needed to expand the memory when storing the inputs and outputs of the external call.
access_gas_cost: the gas cost of accessing the message.to account. This currently corresponds to 2600 gas if the account is cold, and 100 otherwise.
transfer_gas_cost: the cost of transferring a nonzero msg.value. This cost is currently 9000, but provides a 2300 gas stipend to the called contract.
create_gas_cost: the cost of creating a new account, currently 25000. This only applies if message.value != 0, message.to.nounce == 0, message.to.code == b"" and message.to.balance == 0. Because message.to is [checked to have code](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L645), this cost can be ignored here.
We thus want to check that the following:

63 / 64 * (gasleft() - memory_cost - access_gas_cost - transfer_gas_cost)
is at least as much as A ([reference implementation](https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/vm/gas.py#L230)). The sum of access_gas_cost and transfer_gas_cost can be upper bounded with 2_600 + 9_000 - 2_300 = 9_300. The [memory_cost](https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/vm/gas.py#L128) is cumbersome to compute in practice, but by estimating the costs through tests, we can upper bound the cost of memory expansion for up to 10_000 bytes of message.data in the context of the call with the formula 1_200 + 3 * message.data.length / 32.

Thus, it would be possible to validate that the call will have enough gas by checking the following condition right before the external call is made:

63 * gasleft() >= 64 * A + 63 * (9_300 + 1_200 + 3 * message.data.length / 32) + small_buffer
This would be accurate for messages with up to 10_000 bytes of message.data. Any message above this limit could be required to have a gas limit of zero. This would force it to be [processed by its destOwner](https://github.com/taikoxyz/taiko-mono/blob/dd8725f8d27f835102fa3c5a013003090268357d/packages/protocol/contracts/bridge/Bridge.sol#L233-L235). A similar approach, without a constraint on the data size, has been adopted by [Optimism](https://github.com/ethereum-optimism/optimism/blob/8dd18f0efd9719762961897bf93ef1dd8ce702a7/packages/contracts-bedrock/src/libraries/SafeCall.sol#L51-L81) and can be used for inspiration.

----

Note that I didn't put a hard limit as suggested to `message.data.length` as we don't know yet in practice  how large `message.data` can be.


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -582,6 +571,39 @@ contract Bridge is EssentialContract, IBridge {
         }
     }
 
+    /// @notice Consumes a given amount of Ether from quota manager.
+    /// @param _amount The amount of Ether to consume.
+    /// @return true if quota manager has unlimited quota for Ether or the given amount of Ether is
+    /// consumed already.
+    function _consumeEtherQuota(uint256 _amount) private returns (bool) {
```
dantaik, line 578: `_consumeEtherQuota` and `_loadContext` were moved in the file but their implementations remain eactly the same. No need to review them.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -608,44 +630,38 @@ contract Bridge is EssentialContract, IBridge {
...
-        returns (uint256 gasLimit_)
+        returns (bool result_)
     {
         unchecked {
-            uint256 minGasRequired = getMessageMinGasLimit(_message.data.length);
-            gasLimit_ = minGasRequired.max(_message.gasLimit) - minGasRequired;
-        }
+            // https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/vm/gas.py#L128
+            uint256 words = _dataLength / 32 + 1;
+            uint256 memoryGasCost = words * 3 + words * words / 512;
```
Brechtpd, line 645: I don't think we need to add this check here because we don't know if the memory expansion is required for this data inside the invocation call, and it's fully controlled by the user so nothing can be manipulated.

Also the `words * words / 512` part at least should be removed because that part should be done on the total memory size when it's done which is unknown here, so can't calculate it at this point.

Brechtpd: I misunderstood the problem and though this was about something else, if this is the initial memory expansion this is the correct formula.

Brechtpd (approved):

adaki2004 (approved):

PR #17463, opened 2024-06-02
feat(taiko-client): remove useless function and correct erorr handling

mask-pp:


packages/taiko-client/pkg/rpc/beaconclient.go
```diff
@@ -95,13 +95,17 @@ func (c *BeaconClient) GetBlobs(ctx context.Context, time uint64) ([]*blob.Sidec
 		return nil, err
 	}
 
-	var sidecars *blob.SidecarsResponse
 	resBytes, err := c.Get(ctxWithTimeout, c.BaseURL().Path+fmt.Sprintf(sidecarsRequestURL, slot))
 	if err != nil {
 		return nil, err
 	}
 
-	return sidecars.Data, json.Unmarshal(resBytes, &sidecars)
```
mask-pp, line 104: If `json.Unmarshal(resBytes, &sidecars)` failed, then `sidecars` is nil so `sidecars.Data` will cause panic.

packages/taiko-client/prover/event_handler/block_proposed.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 var (
-	errL1Reorged         = errors.New("L1 reorged")
+	errL1Reorged         = errors.New("l1 reorged")
```
davidtaikocha, line 26: lets use `L1` instead of `l1`, its intended

packages/taiko-client/prover/event_handler/block_proposed.go
```diff
@@ -212,7 +212,7 @@ func (h *BlockProposedEventHandler) checkL1Reorg(
 		)
 
 		return fmt.Errorf(
-			"L1 block hash mismatch due to L1 reorg: %s != %s",
+			"l1 block hash mismatch due to L1 reorg: %s != %s",
```
YoGhurt111, line 215:
```suggestion
			"L1 block hash mismatch due to L1 reorg: %s != %s",
```

mask-pp: Error string should not be capitalized or end with punctuation mark.

YoGhurt111: `L1` is the proper noun.

YoGhurt111 (approved):

davidtaikocha (approved):

PR #17465, opened 2024-06-03
refactor(protocol): remove BlockAssigned event

dantaik:
@davidtaikocha could you confirm if the `BlockAssigned` event is used at all by node/client? It seems to be equivalent to `BlockProposed`.


smtmfft (approved):

YoGhurt111 (approved):

mask-pp (approved):

davidtaikocha (approved):
Its not used.

PR #16764, opened 2024-04-18
feat(protocol): update airdrop deploy scripts

2manslkh:
Updated airdrop flow:

1. Set AIRDROP_START_TIME=1717588800, AIRDROP_END_TIME=1720180800, AIRDROP_MERKLE_ROOT, BRIDGED_TKO_ADDRESS in .env file
2. Transfer ETH to deployer address (0.01 ETH is sufficient)
3. Deploy commands
- For testnet: `forge script ./script/DeployERC20Airdrop.s.sol --rpc-url https://rpc.hekla.taiko.xyz --broadcast`
- For mainnet: `forge script ./script/DeployERC20Airdrop.s.sol --rpc-url https://rpc.taiko.xyz --broadcast`
4. Ownership will be transferred to `0xf8ff2AF0DC1D5BA4811f22aCb02936A1529fd2Be`
5. Transfer TKO to ERC20Airdrop on Taiko Mainnet
6. Withdraw function added to ERC20Airdrop to withdraw excess TKO after claimEndDate

### Hekla Deployments
```
TKO
✅  [Success]Hash: 0xd7b43ff10d354b9bf9f86d2273e9b25801ee767618ccad9192e08d98586f2c29
Contract Address: 0x02aE3B816a35896537D31c54f4b04E8C35866a6d
Block: 212006
Paid: 0.001094118000729412 ETH (729412 gas * 1.500000001 gwei)

ERC20Airdrop Implementation
✅  [Success]Hash: 0xca550d540fb52d2874bf92c6aed37e65d1bee92a9d311afb0081cc0832d25ad9
Contract Address: 0x8153DaAE24De4154f561Fb9E7e0e29d44f0E46cc
Block: 212006
Paid: 0.002439724501626483 ETH (1626483 gas * 1.500000001 gwei)

ERC20Airdrop Proxy
✅  [Success]Hash: 0x14590709fe0462c98c9285f8e6fb4c7dc7bc356026bf404a2188101dfeb53ac9
Contract Address: 0x7aDe26cf21ab4f11D076AE99580ADD39C2AaA218
Block: 212006
Paid: 0.000415180500276787 ETH (276787 gas * 1.500000001 gwei)

Mint tokens to Airdrop contract
✅  [Success]Hash: 0x10636272e3ba8ace24c66de279ef76af68c915293fea69192e486a7ea86c1ad6
Block: 212006
Paid: 0.000105783000070522 ETH (70522 gas * 1.500000001 gwei)

Set new config
[...]

PR #17468, opened 2024-06-03
chore(protocol): airdrop contract emits ConfigChanged event

dantaik:


adaki2004:
Can you please target this branch with name: `airdrop-contracts` ?
https://github.com/taikoxyz/taiko-mono/pull/16764

It is a separate branch from main as we modified the deployment and also the that the ERC20Airdrop acts as a vault, etc.

PR #17487, opened 2024-06-05
feat(protocol): reduce gas cost by skipping reading storage for `delegates()`

dantaik:
Screenshot 2024-06-05 at 16 42 46

As you can see from the image above, for everyone TAIKO transfer, this function is always called:

Screenshot 2024-06-05 at 16 43 48

Reading `delegates(from)` and `delegates(to)` cost 2100 each, which sucks. This PR make sure  `delegates` always return 0 for two special address: TaikoL1 and ERC20Vault to reduce gas cost, as these addresses never delegate.

A better but my trouble "fix" is to wrap TAIKO into a BOND token which is a pure ERC20 token for bonds.

### DIFF
--- a/packages/protocol/contracts/tko/TaikoToken.sol
+++ b/packages/protocol/contracts/tko/TaikoToken.sol
@@ -27,9 +27,9 @@ contract TaikoToken is TaikoTokenBase {
         _mint(_recipient, 1_000_000_000 ether);
     }
 
-    function delegates(address account) public view virtual returns (address) {
+    function delegates(address account) public view virtual override returns (address) {
         // Special check to avoid reading from storage slots
         if (account == _TAIKO_L1 || account == _ERC20_VAULT) return address(0);
-        else return _delegates[account];
+        else return super.delegates(account);
     }
 }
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/tko/TaikoToken.sol
```diff
@@ -22,4 +26,10 @@ contract TaikoToken is TaikoTokenBase {
         // Mint 1 billion tokens
         _mint(_recipient, 1_000_000_000 ether);
     }
+
+    function delegates(address account) public view virtual override returns (address) {
+        // Special checks to avoid reading from storage slots
+        if (account == _TAIKO_L1 || account == _ERC20_VAULT) return address(0);
+        else return super.delegates(account);
+    }
```
adaki2004, line 34: Should not we apply it on L2's (`BridgedTaikoToken`) ERC20Vault too ? I know gas is cheaper there.. just for the sake of synchronicity.

adaki2004 (approved):
Beside my comment, LGTM.

davidtaikocha (approved):

packages/protocol/contracts/tko/TaikoToken.sol
dantaik: I think it's not necessary.

PR #17488, opened 2024-06-05
fix(protocol): fix getTierProvider and apply additional gas optimization

dantaik:
optimization: read TAIKO_TOKEN, and TIER_ROUTER only once.
bug fix: in block verification, tier_provider is wrongly cached.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoL1.sol
+++ b/packages/protocol/contracts/L1/TaikoL1.sol
@@ -83,9 +83,7 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
         returns (TaikoData.BlockMetadata memory meta_, TaikoData.EthDeposit[] memory deposits_)
     {
         TaikoData.Config memory config = getConfig();
-        (meta_, deposits_) = LibProposing.proposeBlock(
-            state, config, this, _params, _txList, _checkEOAForCalldataDA()
-        );
+        (meta_, deposits_) = LibProposing.proposeBlock(state, config, this, _params, _txList);
 
         if (!state.slotB.provingPaused) {
             LibVerifying.verifyBlocks(state, config, this, config.maxBlocksToVerifyPerProposal);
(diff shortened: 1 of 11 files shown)


packages/protocol/contracts/L1/libs/LibVerifying.sol
```diff
@@ -122,18 +121,18 @@ library LibVerifying {
                 // It's not possible to verify this block if either the
                 // transition is contested and awaiting higher-tier proof or if
                 // the transition is still within its cooldown period.
+                uint16 tier = ts.tier;
+
                 if (ts.contester != address(0)) {
                     break;
                 } else {
-                    if (tierProvider == ITierProvider(address(0))) {
```
dantaik, line 128: we cannot assume the tier provider for block N is also the same one as block N+1 in the future.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -83,12 +83,12 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
         returns (TaikoData.BlockMetadata memory meta_, TaikoData.EthDeposit[] memory deposits_)
     {
         TaikoData.Config memory config = getConfig();
-        (meta_, deposits_) = LibProposing.proposeBlock(
-            state, config, this, _params, _txList, _checkEOAForCalldataDA()
-        );
+        IERC20 tko = IERC20(resolve(LibStrings.B_TAIKO_TOKEN, false));
```
dantaik, line 86: resolve TAIKO_TOKEN once in proposeBlock

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -112,10 +112,10 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
         if (_blockId != meta.id) revert L1_INVALID_BLOCK_ID();
 
         TaikoData.Config memory config = getConfig();
+        IERC20 tko = IERC20(resolve(LibStrings.B_TAIKO_TOKEN, false));
```
dantaik, line 115: resolve TAIKO_TOKEN once in proveBlock

Brechtpd (approved):

adaki2004 (approved):

PR #17494, opened 2024-06-06
refactor(protocol): avoid a SLOAD in Signal Service when a block is proposed

dantaik:
Screenshot 2024-06-06 at 10 16 25

### DIFF
--- a/packages/protocol/contracts/signal/ISignalService.sol
+++ b/packages/protocol/contracts/signal/ISignalService.sol
@@ -152,7 +152,7 @@ interface ISignalService {
     /// @param _chainId Identifier of the chainId.
     /// @param _kind A value to mark the data type.
     /// @return blockId_ The largest synced block id.
-    function getLargestSyncedBlockId(
+    function getSyncedChainHeight(
         uint64 _chainId,
         bytes32 _kind
     )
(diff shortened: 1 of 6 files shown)


packages/protocol/contracts/signal/ISignalService.sol
```diff
@@ -148,6 +148,18 @@ interface ISignalService {
         view
         returns (bool);
 
+    /// @notice Returns the given block's  chain data.
```
dantaik, line 151:
```suggestion
    /// @notice Returns the given block's chain height.
```

packages/protocol/contracts/signal/SignalService.sol
```diff
@@ -13,7 +13,7 @@ import "./ISignalService.sol";
 contract SignalService is EssentialContract, ISignalService {
     /// @notice Mapping to store the top blockId.
     /// @dev Slot 1.
-    mapping(uint64 chainId => mapping(bytes32 kind => uint64 blockId)) public topBlockId;
+    mapping(uint64 chainId => mapping(bytes32 kind => uint64 blockId)) private _topBlockId;
```
adaki2004, line 16: None of our services (UI, BE) using this ? If so i guess, fine.

dantaik: will revert this change.

YoGhurt111 (approved):

Brechtpd (approved):

PR #17497, opened 2024-06-06
feat(protocol): allow QuotaManager to set quota period

dantaik:


adaki2004 (approved):
LGTM, just a friendly reminder: we may need to increase also the quota (for eth).

davidtaikocha (approved):

PR #17503, opened 2024-06-06
fix(protocol): take calldata into account when calculating gas charge

adaki2004:
As Maxim wrote + Seppi reminded:
(It is a rounded-up approach, favoring the relayer/processor a bit, by not checking which byte is 0 which not, but assuming non-zero.. This way relayer can have a bit more buffer).

kép

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -286,7 +286,11 @@ contract Bridge is EssentialContract, IBridge {
                     // - need to have a buffer/small revenue to the realyer since it consumes
                     // maintenance and infra costs to operate
                     uint256 refund = stats.numCacheOps * _GAS_REFUND_PER_CACHE_OPERATION;
-                    stats.gasUsedInFeeCalc = uint32(GAS_OVERHEAD + gasStart - gasleft());
+                    // Taking into account the encoded message calldata cost, and can count with 16
+                    // gas per bytes (vs. checking each and every byte if zero or non-zero)
+                    uint32 msgCalldataGas = uint32(abi.encode(_message).length) << 4;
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -286,7 +286,11 @@ contract Bridge is EssentialContract, IBridge {
                     // - need to have a buffer/small revenue to the realyer since it consumes
                     // maintenance and infra costs to operate
                     uint256 refund = stats.numCacheOps * _GAS_REFUND_PER_CACHE_OPERATION;
-                    stats.gasUsedInFeeCalc = uint32(GAS_OVERHEAD + gasStart - gasleft());
+                    // Taking into account the encoded message calldata cost, and can count with 16
```
dantaik, line 289: could you add a TODO in the L1 contract log file to upgrade the Bridge on both L1 and L2.

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -286,7 +286,11 @@ contract Bridge is EssentialContract, IBridge {
...
+                    // gas per bytes (vs. checking each and every byte if zero or non-zero)
+                    uint32 msgCalldataGas = uint32(abi.encode(_message).length) << 4;
```
dantaik, line 291: the message only have the data file as a variable length data, all the other fields are fixed size. Maybe we can use _message.data.lenght + a constant value to avoid `abi.encode`?

adaki2004: Done. Needed to get rid of this tmp variable tho, due to stack too deep.

dantaik (approved):

Brechtpd (approved):

PR #17517, opened 2024-06-07
feat(protocol): allow quota whitelisting

dantaik:
Shall we consider adding this feature. Please vote!


Brechtpd:
Doing this to temporarily unblock bridge transactions that are now stuck because of the quota I think would be acceptable (though in that case whitelisting on specific message hashes would be even better). As a general long lived feature though I don't think we should do this (I also don't really see a reason if the UI warns against too large withdrawals, people should just split them up in smaller ones).

PR #17521, opened 2024-06-08
fix(protocol): fix an issue for `ProverSet.proposeBlock`

davidtaikocha:


dantaik (approved):

adaki2004 (approved):

packages/protocol/contracts/team/proving/ProverSet.sol
```diff
@@ -83,7 +83,7 @@ contract ProverSet is EssentialContract, IERC1271 {
         bytes calldata _txList
     )
         external
-        onlyProver
```
Brechtpd, line 86: Don't we need this check because of this: https://github.com/taikoxyz/taiko-mono/blob/d3037ad6551b5ae6353a360ae3677ec17cd00ec9/packages/protocol/contracts/L1/libs/LibProposing.sol#L184-L186

So now anybody could propose a block through this contract, set no hooks, and use the TAIKO in this contract as a bond for the proving no?

davidtaikocha: They still need an enabled prover to sign their blob at first?

Brechtpd: You mean this check: https://github.com/taikoxyz/taiko-mono/blob/d3037ad6551b5ae6353a360ae3677ec17cd00ec9/packages/protocol/contracts/L1/libs/LibProposing.sol#L138 ? This check is only done when no blob is used, so then still a problem when a blob is used I think.

hmmm also does ECDSA.recover even support contracts? Looking at https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol it doesn't seem to call `isValidSignature` anywhere for contract addresses.

davidtaikocha: I think you are right, we may also need to whitelist the proposer addresses in proverSet @dantaik

PR #17525, opened 2024-06-08
chore(relayer): prevent indexing halts and add error monitoring metrics

xiaodino:
Updated error handling in the block indexing code to use slog.Error for logging errors instead of wrapping and returning them.

This ensured that if the indexer encounters an error with a specific transaction, it logs the error after retries and continues processing the remaining transactions. This prevents the indexing process from halting due to individual transaction errors.

### DIFF
--- a/packages/relayer/cmd/flags/common.go
+++ b/packages/relayer/cmd/flags/common.go
@@ -105,12 +105,14 @@ var (
 		Usage:    "Retry interval in seconds when there is an error",
 		Category: processorCategory,
 		Value:    12,
+		EnvVars:  []string{"BACKOFF_RETRY_INTERVAL"},
 	}
 	BackOffMaxRetrys = &cli.Uint64Flag{
 		Name:     "backoff.maxRetrys",
 		Usage:    "Max retry times when there is an error",
 		Category: processorCategory,
-		Value:    3,
+		Value:    5,
+		EnvVars:  []string{"BACKOFF_MAX_RETRYS"},
 	}
 )
 
--- a/packages/relayer/indexer/indexer.go
+++ b/packages/relayer/indexer/indexer.go
@@ -384,7 +384,7 @@ func (i *Indexer) filter(ctx context.Context) error {
 		switch i.eventName {
 		case relayer.EventNameMessageSent:
 			if err := i.withRetry(func() error { return i.indexMessageSentEvents(ctx, filterOpts) }); err != nil {
-				return errors.Wrap(err, "i.indexMessageSentEvents")
+				slog.Error("i.indexMessageSentEvents", "error", err)
 			}
 
 			// we dont want to watch for message status changed events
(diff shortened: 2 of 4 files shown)


davidtaikocha (approved):

cyberhorsey:
why?

dantaik (approved):

xiaodino:
> Updated error handling in the block indexing code to use slog.Error for logging errors instead of wrapping and returning them.
> 
> This ensured that if the indexer encounters an error with a specific transaction, it logs the error after retries and continues processing the remaining transactions. This prevents the indexing process from halting due to individual transaction errors.

Updated the description: 
Updated error handling in the block indexing code to use slog.Error for logging errors instead of wrapping and returning them.
This ensured that if the indexer encounters an error with a specific transaction, it logs the error after retries and continues processing the remaining transactions. This prevents the indexing process from halting due to individual transaction errors.

xiaodino:
> why?

Updated the description:
Updated error handling in the block indexing code to use slog.Error for logging errors instead of wrapping and returning them.
This ensured that if the indexer encounters an error with a specific transaction, it logs the error after retries and continues processing the remaining transactions. This prevents the indexing process from halting due to individual transaction errors.

cyberhorsey:
> > why?
> 
> Updated the description: Updated error handling in the block indexing code to use slog.Error for logging errors instead of wrapping and returning them. This ensured that if the indexer encounters an error with a specific transaction, it logs the error after retries and continues processing the remaining transactions. This prevents the indexing process from halting due to individual transaction errors.

But now we don't know about errors and can miss messages, because we will index blocks past the one that contained the error. Simply logging it isn't enough. If you want to make this more robust, you need to log it and update a Prometheus metric then create an alert tied to the metric with the message type that is being errored.

With this change, we could miss any number of messages, and never know unless we happen to be actively reading logs.

xiaodino:
> > > why?
> > 
> > 
> > Updated the description: Updated error handling in the block indexing code to use slog.Error for logging errors instead of wrapping and returning them. This ensured that if the indexer encounters an error with a specific transaction, it logs the error after retries and continues processing the remaining transactions. This prevents the indexing process from halting due to individual transaction errors.
> 
> But now we don't know about errors and can miss messages, because we will index blocks past the one that contained the error. Simply logging it isn't enough. If you want to make this more robust, you need to log it and update a Prometheus metric then create an alert tied to the metric with the message type that is being errored.
> 
> With this change, we could miss any number of messages, and never know unless we happen to be actively reading logs.
> 
> With this change, what is the current plan for reindexing blocks where messages had errors? What is our plan for being notified about them?


Thank you for the feedback. The primary goal of this PR is to prevent the indexing process from halting due to individual transaction errors. Here's how we address your concerns:

Key Changes:
1. Error Logging and Metrics:
- Added Prometheus metrics to track errors for each event type:
```message_sent_events_after_retry_error_count```
```message_status_changed_events_after_retry_error_count```
```chain_data_synced_events_after_retry_error_count```
```message_processed_events_after_retry_error_count```
These metrics increment when errors are logged after retries.

2. Alerting:
We will set up Prometheus alerts based on the new error metrics. This ensures we are promptly notified of errors and can take action.

3. Reindexing Strategy:
We have an existing crawler that reindexes data. When an error is logged, it is tracked by Prometheus. This ensures any missed messages are eventually processed.

Next Steps:
1. Implement Alerts:
- Set up Prometheus alerts based on the new error metrics to ensure timely notifications.
Ensure Crawler Reindexing:
2. Ensure Crawler Reindexing:
- Ensure the crawler to reindex blocks.

Let me know if you have any further suggestions or concerns.

cyberhorsey (approved):

PR #17529, opened 2024-06-08
fix(protocol): fix Bridge forwarded gas check and `getMessageMinGasLimit`

dantaik:
Please wait for auditor's feedback.


packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -666,7 +666,10 @@ contract Bridge is EssentialContract, IBridge {
 
             // Also need to add the cost of transferring a nonzero msg.value. This cost is currently
             // 9000, but provides a 2300 gas stipend to the called contract.
-            result_ = gasleft() * 63 < _minGas * 64 + (memoryGasCost + 9300) * 63;
+
+            // Also need to add the cost of the external CALL, which is 5700.
```
Brechtpd, line 670: Where does the 5700 gas come from? Just the call itself is 100 gas, and then there are some additional costs which are already taken into account I think, though we are pessimistic here because we don't check if there is a msg.value. A transfer to an empty account will cost an additional 25k gas, but because the call requires implementing an interface I think this case cannot be triggered.

I don't know how accurate this formula here needs to be.

dantaik: I think being more accurate is always better. I updated the code and comment and listed all extra cost that we need to take into consideration. Please take another look.

adaki2004 (approved):
lgtm

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -661,12 +661,14 @@ contract Bridge is EssentialContract, IBridge {
...
             uint256 memoryGasCost = words * 3 + words * words / 512;
 
-            // Need to add the gas cost of accessing the message.to account. This currently
-            // corresponds to 2600 gas if the account is cold, and 100 otherwise.
-
-            // Also need to add the cost of transferring a nonzero msg.value. This cost is currently
-            // 9000, but provides a 2300 gas stipend to the called contract.
-            result_ = gasleft() * 63 < _minGas * 64 + (memoryGasCost + 9300) * 63;
+            // Need to add the cost of the worst case scenario of the external CALL, which include
+            // 700 gas for `CALL`, 2,600 gas for `address_access_cost`, 9,000 for
```
Brechtpd, line 665: there is no 700 gas for the call itself as far as I know, the minimum gas cost for a CALL is 100 gas which is just the warm address access cost.

Brechtpd (approved):

packages/protocol/contracts/bridge/Bridge.sol
dantaik: According to https://www.evm.codes/, it does seem the cost of CALL opcode is 100 gas.

dantaik: Updated the code and comment, but keep using 40000

Brechtpd: > According to https://www.evm.codes/, it does seem the cost of CALL opcode is 100 gas.

I don't think even the 100 gas is a fixed cost for the opcode, that's just the minimum gas cost for the whole opcode:
> static_gas = 0
> dynamic_gas = memory_expansion_cost + code_execution_cost + address_access_cost + positive_value_cost + value_to_empty_account_cost

The static gas is 0, so no fixed cost for the opcode itself. The minimum 100 gas comes from the warm `address_access_cost` I believe.

dantaik: Fixed by our auditor, please take another look.

packages/protocol/foundry.toml
```diff
@@ -48,5 +48,8 @@ multiline_func_header = "all"
 number_underscore = "thousands"
 wrap_comments = true
 
+[profile.default.optimizer_details]
+yul = false
```
dantaik, line 52: this is necessary for `msize` opcode to work

packages/protocol/contracts/bridge/Bridge.sol
```diff
@@ -658,15 +644,24 @@ contract Bridge is EssentialContract, IBridge {
...
-            uint256 memoryGasCost = words * 3 + words * words / 512;
-
-            // Need to add the gas cost of accessing the message.to account. This currently
-            // corresponds to 2600 gas if the account is cold, and 100 otherwise.
 
-            // Also need to add the cost of transferring a nonzero msg.value. This cost is currently
-            // 9000, but provides a 2300 gas stipend to the called contract.
-            result_ = gasleft() * 63 < _minGas * 64 + (memoryGasCost + 9300) * 63;
+            // We can get the actual memory allocated with `msize()`, but that will require
+            // yul optimizer turned off, which will make other contract's code size too large.
```
Brechtpd, line 649: So we're trying to calculate the gas cost to expand the memory to setup the call on the bridge side right? I previously assumed it was the potential memory expansion inside the call, but that actually makes less sense.

dantaik: We are trying to measure the memory cost of *setting up* the call, not the memory cost inside the call.

adaki2004:
I think Maxim has no rights to comment the PR, but he posted this in Slack.

```
function _hasInsufficientGas(
	uint256 _minGas,
	uint256 _dataLength
    )
        private
        view
        returns (bool result_)
{
	unchecked {
		// https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/vm/gas.py#L128
		uint256 currentMemSize; 
		assembly { currentMemSize := mload(0x40) } // Get size in bytes of currently allocated memory
		uint256 oldTotalMemWords = (currentMemSize + 31) / 32;
		uint256 newTotalMemWords = (currentMemSize + _dataLength + 31) / 32; 

		uint256 alreadyPaid = oldTotalMemWords * 3 + oldTotalMemWords * oldTotalMemWords / 512;
		uint256 newlyPaid = newTotalMemWords * 3 + newTotalMemWords * newTotalMemWords / 512;
		uint256 memoryGasCost = newlyPaid - alreadyPaid;

		// Need to add the cost of the worst case scenario of the external CALL, which includes
		// 2,600 gas for `address_access_cost`, 9,000 for `positive_value_cost` (but provides 
		// a 2300 gas stipend to the called contract), and 25,000 for `value_to_empty_account_cost`.
		// See EIP-2929.

		// Therefore the total additional cost is 2600 + 9000 - 2300 + 25000 = 34300.
		// We use 40000, not 34300, below as a safer buffer.
		result_ = gasleft() * 63 < _minGas * 64 + (memoryGasCost + 40_000) * 63;
	}
}
```

dantaik:
Updated the implementation again, please review. @Brechtpd @adaki2004

adaki2004 (approved):

Brechtpd (approved):

adaki2004 (approved):

Brechtpd (approved):

PR #17532, opened 2024-06-08
refactor(protocol): optimize storage reads/writes in proveBlock

dantaik:
Most `proveBlock` transactions submit  a new state transition for blocks. Previously we would fetch and compare storage values from the state transition ring buffer with the prover's values, updating each slot one by one.

With the new optimization, we bypass fetching these values for new state transitions (because the slots' current values don't matter), instead opting to perform a direct storage save at the end.

This optimisation reduces gas by 20K.


Brechtpd (approved):
Looks okay, but there's a lot of logic related to these state transitions  and it's too much for me to keep in my head just reviewing these PRs without going through all the logic again so I don't know for sure.

dantaik:
> Looks okay, but there's a lot of logic related to these state transitions and it's too much for me to keep in my head just reviewing these PRs without going through all the logic again so I don't know for sure.

Don't worry, we need to test this PR in hekla to make sure it truly works.

dantaik:
> Looks okay, but there's a lot of logic related to these state transitions and it's too much for me to keep in my head just reviewing these PRs without going through all the logic again so I don't know for sure.

Brecht, please take another look. I made some new changes.

Brechtpd (approved):

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -376,7 +367,7 @@ library LibProving {
     // 6.5625.
     function _overrideWithHigherProof(
         TaikoData.Block storage _blk,
-        TaikoData.TransitionState storage _ts,
+        TaikoData.TransitionState memory _ts,
```
adaki2004, line 370: I'm not sure this works !

In this function we shall persist some data on-chain - but obviously with `memory` we dont:

```
   _ts.validityBond = _local.tier.validityBond;
        _ts.contester = address(0);
        _ts.prover = msg.sender;
        _ts.tier = _proof.tier;

        if (!_local.sameTransition) {
            _ts.blockHash = _tran.blockHash;
            _ts.stateRoot = _tran.stateRoot;
        }
```

adaki2004: Ok, i think this is what made me sceptical:
kép

But i verified that `memoy struct` is passed by value to private functions, so so far seems good !

adaki2004 (approved):

PR #17538, opened 2024-06-09
feat(eventindexer): indexing nft metadata

k-kaddal:
### Indexing NFT Metadata

This PR us index NFT metadata by fetching and storing it in the db when an NFT transfer. Implemented an endpoint to retrieve the stored NFT metadata by contract address and token ID.

`GET /nftMetadata?contractAddress={contract_address}f&tokenID={token_id}`

Key Changes:

1. Database: Added migration to create the nft_metadata table.
2. Repository:
  - SaveNFTMetadata: Saves NFT metadata.
  - GetNFTMetadata: Retrieves NFT metadata by contract address and token ID.
  - FindByContractAddress: Paginates NFT metadata by contract address.
3. Fetching Metadata: Implements functions to fetch and parse metadata for ERC-1155 and ERC-721 tokens.
4. API: Added an endpoint to retrieve NFT metadata by contract address and token ID.
5. Indexing Integration: Integrated metadata indexing in the index_nft_transfer function.


packages/eventindexer/indexer/fetch_nft_metadata.go
```diff
@@ -0,0 +1,167 @@
...
+
+	err = contractABI.UnpackIntoInterface(&tokenURI, methodName, result)
+	if err != nil {
+		return nil, err
+	}
+
+	url, err := resolveMetadataURL(tokenURI)
+	if err != nil {
+		if errors.Is(err, eventindexer.ErrInvalidURL) {
+			slog.Warn("Skipping metadata due to invalid URI")
```
cyberhorsey, line 60: can we update this log? `slog.Warn("Skipping metadata due to invalid URI", "tokenID", tokenID.String(), "contractAddress", contractAddress)`.. Would help us look at some that encounter this.

packages/eventindexer/indexer/fetch_nft_metadata.go
```diff
@@ -0,0 +1,167 @@
...
+			return nil, nil
+		}
+
+		return nil, err
+	}
+
+	//nolint
+	resp, err := http.Get(url)
```
cyberhorsey, line 68: we should be passing the context in, so if the application is shut down, it responds to context cancellations.
ie:
```go

func fetchURL(ctx context.Context, url string) (*http.Response, error) {
	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
	if err != nil {
		return nil, err
	}

	resp, err := http.DefaultClient.Do(req)
	if err != nil {
		return nil, err
	}

	return resp, nil
}
```

cyberhorsey (changes requested):

cyberhorsey (approved):

dantaik:
@k-kaddal this PR is taking very long. Will you get it ready or closed?

xiaodino (approved):

PR #17544, opened 2024-06-10
fix(protocol): fix seemingly quota issue

adaki2004:
The false-positive revert message is: `QM_OUT_OF_QUOTA`, but the reason is: `_token` is `address(0)` and used before assignment.

e.g.: https://etherscan.io/tx/0xaef434ded5c098f42cac27e9d4ecbb028afb672ba1b45429b95a7295752886e1


KorbinianK (approved):

davidtaikocha (approved):

Brechtpd (approved):

Brechtpd:
I guess it doesn't matter much because we only use the quotas on L1, but just always using `_ctoken.addr` as the address also seems to make sense to me.

adaki2004:
> I guess it doesn't matter much because we only use the quotas on L1, but just always using `_ctoken.addr` as the address also seems to make sense to me.

I think it could result in a posiisble address clash.
E.g.: `Token X` is deployed as canonical on `TaikoL2B`.

When it is a Bridged token on Ethereum (L1), it might be technically possible that a token address on TaikoL2B (e.g.: `Token X`) has a token already deployed on Ethereum. So i'd leave it as is.

dantaik (approved):

PR #17552, opened 2024-06-10
feat(docs-site): update multiple articles

RogerLamTd:


dantaik (approved):

adaki2004 (commented):
To the **excalidraw diagrams**:
1. maybe the first title is a bit misleading: `"Process message on source chain" ` - becaue we have `processMessage()` call on the destination chain.. Maybe `"Handle a message on source chain"` or `"Send message on source chain"` could be better, since we anyways call those as `sendMessage()` ?
2. We no longer have EtherVault, could you remove from the excxalidraw ? And also maybe if something mentioning it in the text ? (Only EtherVault seems incorrect). Now we take custody of `Ether` in the `Bridge`.


Also now we dont have `receiveToken()` but instead `onMessageInvocation()`, can you exchange that in the text ?

adaki2004 (approved):

PR #17553, opened 2024-06-11
feat(protocol): optimize protocol for gas reduction and simplicity

dantaik:
BEGIN_COMMIT_OVERRIDE
feat(protocol): verify blocks less frequently but more efficiently
feat(protocol): persist and compare stateRoot only once per 16 blocks
feat(protocol): require assigned prover to be the block proposer itself
feat(protocol): remove hook support completely
END_COMMIT_OVERRIDE

In theory, when the network is stable, the number of blocks verified and the number of new blocks proposed are roughly equal. This implies that, on average, there is only 0.5 block verified per `proposeBlock`/`proveBlock` transaction. In most cases, this results in either 0 or 1 block being verified per `proposeBlock`/`proveBlock` transaction. Notably, verifying a block involves an `sstore` operation for a `verifiedTransitionId`, which costs 5000 gas.

This PR reduces the number of verification attempts triggered by `proposeBlock`/`proveBlock`. Now, `proposeBlock` will attempt to verify blocks when `blockId % 8 == 0`, and `proveBlock` will do so when `blockId % 8 == 4`. This adjustment means each attempt will, on average, verify 4 blocks instead of 0.5 blocks. Additionally, we now only write the `verifiedTransitionId` for the last block verified. With these two changes, this PR reduces the gas cost by approximately 5325 gas ((5000+2100) * ¾).

We now also save stateRoot into storage only when `blockId % 16 == 15` to avoid one storage write and read for 15 blocks every 16 blocks.


dantaik:
@davidtaikocha @cyberhorsey @xiaodino 
Tthe `BlockVerified` event has been modified, please verify if deleting `stateRoot` from the event works for you?

cyberhorsey:
> @davidtaikocha @cyberhorsey @xiaodino Tthe `BlockVerified` event has been modified, please verify if deleting `stateRoot` from the event works for you?

Yes, I think the only thing we need now is to take the eventindexer, and be able to use a different Abi binding depending on the time. For instance, before the upgrade, we need the old ABI to index these events - once the upgrade is performed, we cancel the previous Abi, and use a new one, essentially "forking".   
  
Relayer will be unaffected

packages/protocol/contracts/L1/TaikoEvents.sol
```diff
@@ -30,13 +30,13 @@ abstract contract TaikoEvents {
...
     /// @param blockHash The hash of the verified block.
-    /// @param stateRoot The block's state root.
+    /// @param transitionId The transition that used to verify this block.
     /// @param tier The tier ID of the proof.
     event BlockVerified(
         uint256 indexed blockId,
         address indexed prover,
         bytes32 blockHash,
-        bytes32 stateRoot,
+        uint32 transitionId,
```
davidtaikocha, line 39: I think we should keep the type here, or all clients will break. This will be an ABI break change.

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -104,7 +103,6 @@ library LibProving {
         TaikoData.TierProof memory _proof
     )
         internal
-        returns (uint8)
```
davidtaikocha, line 107: I think this will also lead to another abi break change, shall we keep it? or all prover clients will break after upgrade.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -87,8 +86,9 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
 
         (meta_, deposits_) = LibProposing.proposeBlock(state, tko, config, this, _params, _txList);
 
-        if (!state.slotB.provingPaused) {
-            LibVerifying.verifyBlocks(state, tko, config, this, config.maxBlocksToVerifyPerProposal);
+        uint256 freq = config.verificationFrequencyFactor;
```
adaki2004, line 89: I think the `freq` (or rather `config.verificationFrequencyFactor`) variable shall has the same value as `maxBlocksToVerify`, so we can reuse.
Why ?

Let's say `verificationFrequencyFactor` is 16 but `maxBlocksToVerify` 8. In a world where proposing and verification are roughly equal as PR desc. suggest, it would mean that the verification will be a bottleneck, as we only trying to verify 8, after every 16 block proposed (or proven).

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -230,6 +234,9 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
             blockMaxGasLimit: 240_000_000,
             livenessBond: 250e18, // 250 Taiko token
             blockSyncThreshold: 32,
+            // Setting this to 16 means on average, on average 8 blocks will be verified in each
+            // verifyBlocks().
+            verificationFrequencyFactor: 16,
```
adaki2004, line 239: Maybe we can omit this extra config value and reuse the `maxBlocksToVerify` ?
https://github.com/taikoxyz/taiko-mono/pull/17553/files#r1635177191

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -106,6 +106,7 @@ library TaikoData {
         uint96 contestBond;
         uint64 timestamp; // slot 6 (90 bits)
         uint16 tier;
+        uint8 __reserved1;
```
dantaik, line 110: revert a previous change that breaks client

packages/protocol/contracts/L1/tiers/ITierProvider.sol
```diff
@@ -11,7 +11,6 @@ interface ITierProvider {
         uint96 contestBond;
         uint24 cooldownWindow; // in minutes
         uint16 provingWindow; // in minutes
-        uint8 maxBlocksToVerifyPerProof;
```
davidtaikocha, line 14: This change will also break the client.

packages/protocol/contracts/L1/tiers/TierProviderBase.sol
```diff
@@ -22,7 +22,7 @@ abstract contract TierProviderBase is ITierProvider {
                 contestBond: 500 ether, // TKO
                 cooldownWindow: 1440, //24 hours
                 provingWindow: 30, // 0.5 hours
-                maxBlocksToVerifyPerProof: 16
+                maxBlocksToVerifyPerProof: 0
```
adaki2004, line 25: Proving a block would not trigger verification from this point on ?

adaki2004 (approved):
generally lgtm, changes tho seems massive (mainly because of the renaming of variable `local` to `l`, and i really hope i did not overlook any renaming with a real change! That can be a real fear, maybe `local` vs. `l` does not really matter, so i might have left that variable named like that, also more descriptive.)
But if works on hekla, than shall be OK.

dantaik:
https://dashboard.tenderly.co/tx/holesky/0x06341628cd5de3318a4be969b060ca24b81a328536e04dd87ff23aeddf9b1d3c/gas-usage (https://holesky.etherscan.io/tx/0x06341628cd5de3318a4be969b060ca24b81a328536e04dd87ff23aeddf9b1d3c)

Here is a hekla Guardian's `proveBlock` transaction that verified 16 blocks.

packages/protocol/contracts/L1/tiers/TierProviderBase.sol
dantaik: No, but this field is not used. We use TaikoL1.getConfig().maxBlocksToVerify instead regardless of tiers.

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -87,8 +87,9 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
 
         (meta_, deposits_) = LibProposing.proposeBlock(state, tko, config, this, _params, _txList);
 
-        if (!state.slotB.provingPaused) {
-            LibVerifying.verifyBlocks(state, tko, config, this, config.maxBlocksToVerifyPerProposal);
+        uint256 oddBase = config.maxBlocksToVerify >> 1;
+        if (oddBase != 0 && meta_.id % oddBase == 0 && !state.slotB.provingPaused) {
```
Brechtpd, line 91: why this code with oddBase here and in the proveBlock? I don't think I understand why this just doesn't do the modulo directly with some value.

Brechtpd:
> Additionally, we now only write the verifiedTransitionId for the last block verified.

So all other blocks will have invalid data now? Seems important to mention somewhere if so.

dantaik:
> > Additionally, we now only write the verifiedTransitionId for the last block verified.
> 
> So all other blocks will have invalid data now? Seems important to mention somewhere if so.

Updated the code with some comments. Let me if your questions got answered in the comment.

packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -123,4 +123,38 @@ library LibUtils {
...
+        uint256 segmentSize = _config.maxBlocksToVerify / 2;
+
+        if (segmentSize <= 1) return true;
+
+        uint256 remainder;
+        unchecked {
+            remainder = _isBlockProposed ? segmentSize - 1 : segmentSize / 2 - 1;
+        }
+
+        return _blockId % segmentSize == remainder;
```
Brechtpd, line 178: just `_blockId % segmentSize == 0;` not the same?

packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -143,4 +143,38 @@ library LibUtils {
...
+        if (_config.maxBlocksToVerify == 0) return false;
+
+        // Consider each segment of 8 blocks, verification is attempted either on block 3 if it has
+        // been
+        // proved, or on block 7 if it has been proposed. Over time, the ratio of blocks to
+        // verification attempts averages 4:1, meaning each verification attempt typically covers 4
+        // blocks. However, considering worst cases caused by blocks being proved out of order, some
+        // verification attempts may verify few or no blocks. In such cases, additional
+        // verifications are needed to catch up. Consequently, the `maxBlocksToVerify` parameter
+        // should be set high enough to allow for efficient catch-up.
```
dantaik, line 165:
```suggestion
        // should be set high enough, for example 16, to allow for efficient catch-up.
```

packages/protocol/contracts/L1/libs/LibUtils.sol
dantaik: Statistically yes, but we don't want a block's proposer AND prove to pay for verifications.

dantaik:
> > Additionally, we now only write the verifiedTransitionId for the last block verified.
> 
> So all other blocks will have invalid data now? Seems important to mention somewhere if so.

Added a comment in the code for `verifiedTransitionId`.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -118,6 +120,9 @@ library TaikoData {
         uint64 proposedAt; // timestamp
         uint64 proposedIn; // L1 block number, required/used by node/client.
         uint32 nextTransitionId;
+        // The ID of the transaction that is used to verify this block. However, if
```
dantaik, line 123: @RogerLamTd Previous for every block verified, this verifiedTransitionId will be non-zero, now, only the last block verified in a batch will have verifiedTransitionId non-zero. Do you think you can update related documentation?

RogerLamTd: > @RogerLamTd Previous for every block verified, this verifiedTransitionId will be non-zero, now, only the last block verified in a batch will have verifiedTransitionId non-zero. Do you think you can update related documentation?

will update description of how we verify blocks in batches now in documentation and tag this PR

davidtaikocha (approved):

PR #17570, opened 2024-06-12
feat(protocol): transfer bonds back only when the recipient's balance is non-zero

dantaik:
This is to force proposer/provers/contesters to have non-zero token to reduce token transfer cost. In theory, one prover can prove a block, then move all his TAIKO to another addresses. When this block is verified, his token will be returned, now it takes 21000 gas for this token transfer, paid by another prover or proposer.

Is it a good idea at all?


packages/protocol/contracts/L1/libs/LibUtils.sol
```diff
@@ -22,6 +22,14 @@ library LibUtils {
...
     error L1_UNEXPECTED_TRANSITION_ID();
 
+    /// @notice Transfer ERC20 tokens to the recipient only if the recipient's current token balance
+    /// is not zero.
+    function conditionallyTransferTo(IERC20 token, address recipieint, uint256 amount) internal {
+        if (amount != 0 && token.balanceOf(recipieint) != 0) {
+            token.transfer(recipieint, amount);
+        }
+    }
+
```
adaki2004, line 32: I'm not sure if this is desireable given our permissionless/decentralize nature, just to save some gas. 
Like, i understand the balance can be just 1 wei.. But if there is a hack (a phising site where the malicious actor cannot get access to the PK) and the funds are drained from the user, it might happen that he cannot get back the bond.

dantaik: I welcome these kind of feedback.

In your case above, all bonds returned are also going to be drained by the hacker.

PR #17582, opened 2024-06-13
feat(protocol): introduce breaking changes for gas optimization

dantaik:
The objective of this change is to optimize the gas cost of the `proposeBlock`, `proveBlock`, and `verifyBlock` functions. On mainnet, even small reductions in gas usage can lead to significant Ether savings over time.

### `proposeBlock`

- Block proposer must be the same as the block prover. Custom contracts can handle proposer-prover settlements, as seen in the ProverSet contract. This removes the proposer-prover assignment hook from the protocol.
- Moved `livenessBond` from storage to block metadata as it is static per block. Added a boolean state variable to mark if the `livenessBond` is returned.

These changes reduce the number of storage writes in `proposeBlock` from three to two.

### `proveBlock`

- Defined a block segment as 16 consecutive blocks. For the first 15 blocks in a segment, `stateRoot` is forced to be zero, skipping its verification and storage writes. The 16th block writes `stateRoot` to storage. This change saves 15/16 of storage writes and reads per block.
- The best segment size is yet to be determined. With a 30-second interval between blocks, synchronizing the L2 state root to L1 every 8 minutes (16 blocks) seems reasonable, larger values could delay synchronization.

### `verifyBlock`

- Due to changes in `proveBlock`, `stateRoot` is synchronized only every 16 blocks. For the first 15 blocks in a segment, `stateRoot` is not read as it is zero.
- Changed the odds of attempting to verify blocks. Now, `proposeBlock` transactions for blocks 0 and 8 and proveBlock transactions for blocks 4 and 12 trigger verification attempts. Each verification attempt can verify up to 16 blocks across segment boundaries.

This change makes verification attempts more efficient, as only the last block's verifying transition ID is written to storage, instead of writing for each verified block.

## Early Feedback

Before diving into fixing the implementation bugs in this PR, I would appreciate your feedback on these changes. The value of gas reduction is significant, as every sstore and sload operation translates to real dollar savings. 

## TODOs

- [ ] Fixing existing test failures
- [ ] Add more tests for protocol fork
- [ ] Figure out how the client can handle this breaking change


Brechtpd:
> Block proposer must be the same as the block prover. Custom contracts can handle proposer-prover settlements, as seen in the ProverSet contract. This removes the proposer-prover assignment hook from the protocol.

[Taiko Simplified](https://github.com/Brechtpd/taiko-mono/pull/1) also does this: https://github.com/Brechtpd/taiko-mono/blob/7516eab250e111d80d75d5621c2d59ed2e1bde0f/packages/protocol/contracts/L1/actors/ProverPayment.sol. And many more simplifications that make things simpler (and save gas). And I think Taiko Simplified currently doesn't even go far enough for certain things, more things to be done!

> The value of gas reduction is significant, as every sstore and sload operation translates to real dollar savings.

That's also why I'm always advocating doing as little as possible onchain! Keeps things simple (fewer chance of bugs) and either is already efficient by not doing anything or makes it easier to optimize when necessary. I did not feel that was the direction chosen with the current smart contract implementations, which is one of the reasons I made Taiko Simplified.

adaki2004:
In short: I feel it very risky.

It is nice that we want to remove some extra complexity and lowering the gas consumption, but - tho it could have been done before launching mainnet, i remember kind of these discussions back in the day about Assigment Hooks necessity - we need to ask some one important questions: what can we gain and what can we lose ? 
We can gain:
-  some `$$$`, which I agree is not insignificant, but only in case if we can pull it off with a technically properly executed fork, that does not cause any hiccups
We can lose:
- trust if we dont execute it properly, without major fuck ups

The fact that mainnet already diverged from Hekla a bit (so not 100% equivalency) would also require to maintain 2 separate branches for these upgrades, one for Hekla one for mainnet and not only on protocol level but possibly in the client makes it even more risky. 

So more pairs of eyes, maybe an extra audit could be necessary (to stamp it with a "safe" upgrade), which would increase the manpower and possibly slow down the process.

I'd leave these kind of breaking changes to BBR honestly.

dantaik:
> > Block proposer must be the same as the block prover. Custom contracts can handle proposer-prover settlements, as seen in the ProverSet contract. This removes the proposer-prover assignment hook from the protocol.
> 
> [Taiko Simplified](https://github.com/Brechtpd/taiko-mono/pull/1) also does this: https://github.com/Brechtpd/taiko-mono/blob/7516eab250e111d80d75d5621c2d59ed2e1bde0f/packages/protocol/contracts/L1/actors/ProverPayment.sol. And many more simplifications that make things simpler (and save gas). And I think Taiko Simplified currently doesn't even go far enough for certain things, more things to be done!
> 
> > The value of gas reduction is significant, as every sstore and sload operation translates to real dollar savings.
> 
> That's also why I'm always advocating doing as little as possible onchain! Keeps things simple (fewer chance of bugs) and either is already efficient by not doing anything or makes it easier to optimize when necessary. I did not feel that was the direction chosen with the current smart contract implementations, which is one of the reasons I made Taiko Simplified.

Every design has different assumptions and objectives, such as if proposing and proving blocks shall be permissionless or permissioned. I hope to see BBR can achieve simplification AND permissionless at the same time.

dantaik:
> In short: I feel it very risky.
> 
> It is nice that we want to remove some extra complexity and lowering the gas consumption, but - tho it could have been done before launching mainnet, i remember kind of these discussions back in the day about Assigment Hooks necessity - we need to ask the most important question: what can we gain and what can we lose ?
> 
> We can gain:
> 
> * some `$$$`, which I agree is not insignificant, but only in case if we can pull it off with a technically properly executed fork, that does not cause any hiccups
> 
> We can lose:
> 
> * trust if we dont execute it properly, without major fuck ups
> 
> The fact that mainnet already diverged from Hekla a bit (so not 100% equivalency) would also require to maintain 2 separate branches for these upgrades, one for Hekla one for mainnet and not only on protocol level but possibly in the client makes it even more risky.
> 
> So more pairs of eyes, maybe an extra audit could be necessary (to stamp it with a "safe" upgrade), which would increase the manpower and possibly slow down the process.
> 
> I'd leave these kind of breaking changes to BBR honestly.

I agree this PR is risky, we definitely need to add more unit tests and then use private devnet to ensure the fork is thoroughly tested before rolling it out to Hekla and testnet (BTW, hekla has been upgraded to the latest code and is using the same code as the mainnet).

I'm also considering a followup one week additional audit, I've talked about it with OZ last night. But this shall be started only when the code is well tested already.

dantaik:
Damn, accidentally merged!

PR #17590, opened 2024-06-14
feat(protocol): change guardian prover cooldown windows

dantaik:
Change guardian minority cooldown window from 24 hours to 4 hours. If one guardian is hacked and his proof is incorrect, guardian majority will override it, and guardian majority's cooldown is changed from 1 hour 24 hours so there will be more time for monitoring and necessary actions.


davidtaikocha (approved):

adaki2004:
cc @KorbinianK for bridge UI (message) and @cyberhorsey for relayer queue monitoring time.

adaki2004 (approved):

Brechtpd:
> If one guardian is hacked and his proof is incorrect, guardian majority will override it

4 hours is certainly on the low side for censorship resistance attacks, the majority of guardians now need to get an L1 transaction included in only 4 hours (each), which is lower than is currently seen as safe as far as I know (though that safe value involves quite a bit of hand waving).

PR #17609, opened 2024-06-17
feat(taiko-client): introduce mev pool

YoGhurt111:


Brechtpd:
Can you provide a bit of context? I'm not familiar enough with the geth codebase to immediately see what this PR does.

YoGhurt111:
> Can you provide a bit of context? I'm not familiar enough with the geth codebase to immediately see what this PR does.

We want to optimize the gas setting and send private TX to protect the proposer's profit, so we introduce Blocknative's API.
For the gas price estimate, you can see this https://docs.blocknative.com/gas-prediction/gas-platform.
For the mev production, you can see https://docs.blocknative.com/blocknative-mev-protection/transaction-boost.

packages/taiko-client/pkg/rpc/client.go
```diff
@@ -62,19 +65,38 @@ func NewClient(ctx context.Context, cfg *ClientConfig) (*Client, error) {
 		l1BeaconClient *BeaconClient
 		l2CheckPoint   *EthClient
 		err            error
+		mevPool        *MevPool
```
dantaik, line 68: maybe rename "mevPool" to "privateTxPool"

YoGhurt111: OK

Brechtpd:
> > Can you provide a bit of context? I'm not familiar enough with the geth codebase to immediately see what this PR does.
> 
> We want to optimize the gas setting and send private TX to protect the proposer's profit, so we introduce Blocknative's API. For the gas price estimate, you can see this https://docs.blocknative.com/gas-prediction/gas-platform. For the mev production, you can see https://docs.blocknative.com/blocknative-mev-protection/transaction-boost.

Oh nice, batch send to all builders with revert protection.

packages/taiko-client/pkg/rpc/private_tx_pool.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+const (
+	BlockPriceURL = "/gasprices/blockprices"
+)
+
+type BlocknativePrivateTxPool struct {
+	apiEndpoint string
+	apiKey      string
+	rpcClient   *ethclient.Client
```
mask-pp, line 20:
```suggestion
	rpcClient   *EthClient
```

mask-pp: And `SuggestGasTipCap(ctx context.Context) (*big.Int, error)` in `BlocknativePrivateTxPool` so that we can keep `EthClient` clean.

YoGhurt111: EthClient includes BlocknativePrivateTxPool, which would be better since we have different types of PrivateTxPool

mask-pp: The `SuggestGasTipCap` function in `BlocknativePrivateTxPool` will Shadow the one in `EthClient`.

mask-pp: > EthClient includes BlocknativePrivateTxPool, which would be better since we have different types of PrivateTxPool

It doesn't matter. It's similar to inheritance.

YoGhurt111: That's we want

mask-pp: @davidtaikocha What's your point?

mask-pp (approved):
LGTM, just left a carp.

PR #17620, opened 2024-06-18
feat(taiko-client): skip L2 blob tx when proposing a L2 block

YoGhurt111:


packages/taiko-client/driver/txlist_decompressor/txlist_decompressor.go
```diff
@@ -72,6 +73,16 @@ func (v *TxListDecompressor) TryDecompress(
 		return []byte{}
 	}
 
+	// Re-encode txs when txs contains invalid tx
+	for _, tx := range txs {
+		// Skip invalid L2 blob tx
+		if tx.Type() == types.BlobTxType {
```
smtmfft, line 79: if tx.Type() == types.BlobTxType  && fork <= SHANGHAI ??

YoGhurt111: We won't support l2 blob tx after cancun either.

mask-pp: filter blob tx belongs to `taiko-geth` function, it seems inappropriate to decide how geth should do.

davidtaikocha: Actually don't think this is necessary? taiko-geth will handle it?

YoGhurt111: If taiko-geth will do this case, taiko-geth has to do it now.
It is a breaking bug and needs to be quickly fixed.

davidtaikocha:
Closed due to we want to apply it in taiko-geth

PR #17627, opened 2024-06-20
docs(docs-site): update proof tier config for testnet

RogerLamTd:


dionysuzx (approved):
In the future we could try using Mermaid syntax and programmatically render the config, but out of scope for this PR ofc.

RogerLamTd:
> In the future we could try using Mermaid syntax and programmatically render the config, but out of scope for this PR ofc.

I can look into it as my next task! Good tip

dantaik (approved):

PR #17631, opened 2024-06-20
fix(protocol): support multicall context in `DelegateOwner` 

adaki2004:
In `DelegateOwner` we have the `setAdmin` function, which works as expected if we call it in a "normal", but currently does not support `multicall3` way of calling itself.

So in case of this scenario for example:

`onMessageInvocation` (through the bridge) by calling `aggregate3` on `multicall3`, to perform 2 inner transactions:
- one of them is to set this `DelegateOwner`'s admin address to a new value
- the other is to call an owned contract to perform a upgrade

So the path of calling would be this:
Bridge -> DelegateOwner -> Multicall -> 1. DelegateOwner (setAdmin)
                                                               -> 2. Another contract to upgrade the impl contract.
 
 The problem with the 1st path is, that the `msg.sender` in `setAdmin`, will be DelegateOwner contract address - in this `multicall3` way.

### DIFF
--- a/packages/protocol/contracts/L2/DelegateOwner.sol
+++ b/packages/protocol/contracts/L2/DelegateOwner.sol
@@ -104,7 +104,10 @@ contract DelegateOwner is EssentialContract, IMessageInvocable {
     /// @dev Updates the admin address.
     /// @param _admin The new admin address.
     function setAdmin(address _admin) external {
-        if (msg.sender != owner() && msg.sender != admin) revert DO_PERMISSION_DENIED();
+        IBridge.Context memory ctx = IBridge(msg.sender).context();
+        if (msg.sender != owner() && msg.sender != admin && ctx.from != realOwner) {
+            revert DO_PERMISSION_DENIED();
+        }
         if (admin == _admin) revert DO_INVALID_PARAM();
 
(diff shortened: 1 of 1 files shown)


adaki2004:
@Brechtpd @dantaik 
Honestly i dont think it is a good idea to "multicall" into SELF into DelegateOwner -> Cause it requires changes into the code, as it is currently not supported.

`DelegateOwner` + `Multicall` can work for other contracts (without modification) but not into self - without changes. Given the fact, that DelegateOwner itself has only 1-2 functions, i dont think it shall be a test case:
```

onMessageInvocation (through the bridge) by calling aggregate3 on multicall3, to perform 2 inner transactions:

    one of them is to set this DelegateOwner's admin address to a new value
    the other is to call an owned contract to perform a upgrade
```

If you agree, i'd close this PR and not bother with multicall into self (DelegateOwner).

dantaik:
@adaki2004 Thank you for identiying an issue with `setAdmin`. I prefer a fix in this PR: https://github.com/taikoxyz/taiko-mono/pull/17633/files with some design simplicity. Now only `onMessageInvocation` is the function that shall be called by local admin or a remote owner. All other functions must be called inside `onMessageInvocation`.

PR #17633, opened 2024-06-20
fix(protocol): fix an issue in DelegateOwner then refactor the code

dantaik:


packages/protocol/contracts/L2/DelegateOwner.sol
```diff
@@ -72,63 +83,49 @@ contract DelegateOwner is EssentialContract, IMessageInvocable {
...
+        _invokeCall(_data, false);
+        revert DO_DRYRUN_SUCCEEDED();
     }
 
     /// @dev Updates the admin address.
     /// @param _admin The new admin address.
-    function setAdmin(address _admin) external {
-        if (msg.sender != owner() && msg.sender != admin) revert DO_PERMISSION_DENIED();
-        if (admin == _admin) revert DO_INVALID_PARAM();
+    function setAdmin(address _admin) external nonReentrant onlyFromSelf {
```
adaki2004, line 111: I think onlyFromSelf only true in case it is a `delegateCall` via `multicall`, right ?
But what if the "remoteOwner" or admin wants to set this ? Maybe we dont want "removeOwner" to set this, but i guess admin can do it indeed, without the need of a multicall, but "simple bridge call" ?

dantaik: - you can see delegatecall as borrowing code from another contract into this (DelegateOwner) contract. 
- calling `setAdmin` in the multicall3 (as a delegate call) will make `msg.sender` to be `address(this)`.

The change in this PR is to make sure the only entrance to call anything is `onMessageInvocation`, and enforce `setAdmoin` and `acceptOwnership` to be called as inner transations.

adaki2004: >     * calling `setAdmin` in the multicall3 (as a delegate call) will make `msg.sender` to be `address(this)`.
> The change in this PR is to make sure the only entrance to call anything is `onMessageInvocation`, and enforce `setAdmin` and `acceptOwnership` to be called as inner transations.

Yes, this is what i'm questioning/asking. Do we want to allow to **only** be called via bridge+multicall ? So no direct call ever, neither 'simple' bridge transaction (without multicall) ?

dantaik: Direct calling `onMessageInvocation` by the local admin is still allowed.

adaki2004 (approved):

davidtaikocha (approved):

PR #17635, opened 2024-06-20
fix(relayer): add sanitizeString to fix invalid CanonicalTokenSymbol

xiaodino:


xiaodino:
Use https://github.com/taikoxyz/taiko-mono/pull/17637

PR #17641, opened 2024-06-21
feat(protocol): allow contract proposers to use calldata as well

dantaik:
Only merge this PR when nodes/clients have started to use CalldataTxList event.


Brechtpd:
At 8 gas per byte in an event, this is actually quite expensive which is why events normally or not used for this. Though I guess it's an easy way to mitigate the restriction for small tx lists without too much cost.

Other ways could be shadow events but I have not looked closely into this (next to the more obvious just re-executing the txs to fetch the data which I guess is kind of what shadow events do anyway).

dantaik:
> At 8 gas per byte in an event, this is actually quite expensive which is why events normally or not used for this. Though I guess it's an easy way to mitigate the restriction for small tx lists without too much cost.
> 
> Other ways could be shadow events but I have not looked closely into this (next to the more obvious just re-executing the txs to fetch the data which I guess is kind of what shadow events do anyway).

Shadow events ERC is still in draft, I don't think it will be available anytime soon.

Calldata DA is already very expensive most of the time compared to blobs for L2s even without this new event. But for L3s, I think emiting the event may be fine.

dantaik:
Another idea I have is that the inner transaction (`proposeBlock`) is provided with two inputs: `offset` and `size` , that can be used to locate a segment of the outmost transaction's calldata. And use that segment as the txList. Inside `proposeBlock`, we can then emit an event with `offset` and `size` only.

Do you know to locate the segment from within an inner transaction?

Brechtpd:
> Another idea I have is that the inner transaction (`proposeBlock`) is provided with two inputs: `offset` and `size` , that can be used to locate a segment of the outmost transaction's calldata. And use that segment as the txList. Inside `proposeBlock`, we can then emit an event with `offset` and `size` only.
> 
> Do you know to locate the segment from within an inner transaction?

I don't think you can do it in a smart contract, but I think you can make it work. Because this is not too dissimilar to what optimism does, which simply sends a tx to an EOA and the data is directly read form the tx data (without the need for a range of course in their case) and the data needs to be validated against the tx data directly because it's not possible to calculate the hash of the data anymore in the smart contract. 

This would require some more substantial changes to how the data needs to be proven. Verifying this data requires access to the originating L1 tx, which the only way I can currently think of would require checking that tx against the L1 block hash in which the block was proposed. We could check how optimism verifies the data should be very similar to that, but requires quite a bit of work I think.

dantaik:
Given all the information above and based on my own research, I think we should go with this PR for simplicity.

PR #17646, opened 2024-06-21
docs(protocol): update Hekla deployment

davidtaikocha:


dantaik (approved):

packages/protocol/deployments/hekla-contract-logs.md
```diff
@@ -142,6 +143,14 @@
 - logs:
   - deployed on Jun 17, 2024 at commit `b7e12e3`
 
+### prover_set_contester
+
+- proxy: `0x335103c4fa2F55451975082136F1478eCFeB84B9`
```
dantaik, line 148: @davidtaikocha can we use the same prover_set contract instead?

davidtaikocha: No, if we do so, contester will consider that that block is assigned to him...(so in mainnet, we also use two different prover sets)

Brechtpd (approved):

Issue #14793, opened 2023-09-24
feat(protocol): Longer term predictability of block parameters (for fast tx confirmation)


Brechtpd:
### Describe the feature request

Currently the window a proposer knows how a block will execute is limited to 12 seconds, because the L2 block directly depends on some L1 values that are only known during the span that L1 block is being created. Most notably: 
- The timestamp
- The L1 block height
- The L1 block hash

If we want to make the L2 block building window deterministic for a longer time, that means it needs to be possible for the parameters used for a block to be known for a longer period somehow.

The easiest way to achieve this is to allow the block proposer some flexibility on which values will be used for these block parameters. Because the previous L1 block hash that is immediately available to the L1 smart contract is limited to only the last 256 blocks, this seems to put a pretty reasonable hard limit on how long we'd want this window to be without introducing any extra caching mechanisms.

And so the proposer could decide on what L1 block to base the L2 block upon, as long as
- The L1 block is equal or newer than the L1 block used in the previous L2 block
- The L1 block is not older than 256 blocks

This could however open up L2 block building to more manipulation because it gives more flexibility to the proposer. However, I don't think this will be problematic.

### Describe alternatives you've considered

Description of the alternatives you've considered here.

### Additional context

Additional context here.

dantaik:
Brecht, do you think 12 second is too short, what use cases demand "the L2 block building window deterministic for a longer time"?

Brechtpd:
> Brecht, do you think 12 second is too short, what use cases demand "the L2 block building window deterministic for a longer time"?

Mostly fast soft confirmations with an L1 validator that can propose the L1 block in one of the future slots. Could also be helpful maybe to have blocks with an immediate proof attached.

> I thought the change above may work, but it doesn't -- it's still the same 12 second window.

Something like that I think would work for the case above because the exact moment the L2 block will be inserted in the L1 block is known. For cases where that may not be the case, some more flexibility would be required.

PR #17654, opened 2024-06-23
feat(protocol): add preconfirmation support [helder]

Brechtpd:
For now these changes are only to support the upcoming preconfirmation testnet!

- Implement https://github.com/taikoxyz/taiko-mono/issues/14793
- Add support for proposer selection mechanisms through `ISequencerRegistry`. Currently the testnet needs some very simple whitelisting implementation, this will be replaced when the actual registry can be used.

Tried to minimize changes with how `ISequencerRegistry` is implemented using a callback inside the propose function, otherwise I would probably add an additional contract in front of TaikoL1 but that seems like a bigger change for multiple reasons.

Untested.


packages/protocol/contracts/L1/SequencerRegistry.sol
```diff
@@ -0,0 +1,44 @@
...
+    function setSequencers(
+        address[] memory _sequencers,
+        bool[] memory _enabled
+    )
+        external
+        onlyOwner
+    {
+        require(_sequencers.length == _enabled.length, "invalid input data");
+        for (uint256 i = 0; i < _sequencers.length; i++) {
+            sequencers[_sequencers[i]] = _enabled[i];
```
dantaik, line 36: emit an event?

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -64,6 +64,7 @@ library TaikoData {
         bytes32 parentMetaHash;
         HookCall[] hookCalls; // DEPRECATED, value ignored.
         bytes signature;
+        uint32 l1StateBlockNumber;
```
dantaik, line 67: As this will break our node in mainnet, lets merge this code and other potential changes to a preconf branch so that main branch can still be used to upgrade mainnnet.

packages/protocol/contracts/L1/TaikoErrors.sol
```diff
@@ -35,4 +35,5 @@ abstract contract TaikoErrors {
     error L1_TRANSITION_NOT_FOUND();
     error L1_UNEXPECTED_PARENT();
     error L1_UNEXPECTED_TRANSITION_ID();
+    error L1_INVALID_PROPOSER();
```
dantaik, line 38: sort strings

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -124,6 +125,7 @@ library TaikoData {
         // this block is not verified as the last block in a batch, verifiedTransitionId
         // will remain zero.
         uint32 verifiedTransitionId;
+        uint64 l1StateBlockNumber;
```
dantaik, line 128: What's the meaning of l1StateBlockNumber?

packages/protocol/contracts/L1/SequencerRegistry.sol
```diff
@@ -0,0 +1,44 @@
...
+
+import "../common/EssentialContract.sol";
+import "./ISequencerRegistry.sol";
+
+/// @title SequencerRegistry
+/// A dummy implementation that only whitelist some trusted addresses. A real
+/// implementation would only allow a single proposer address to propose a block
+/// using some selection mechanism.
+/// @custom:security-contact security@taiko.xyz
+contract SequencerRegistry is EssentialContract, ISequencerRegistry {
```
dantaik, line 12: Sequencer => Proposer?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -74,35 +75,47 @@ library LibProposing {
...
 
         // Check if parent block has the right meta hash. This is to allow the proposer to make sure
         // the block builds on the expected latest chain state.
-        if (params.parentMetaHash != 0 && parentMetaHash != params.parentMetaHash) {
+        if (params.parentMetaHash != 0 && parentBlock.metaHash != params.parentMetaHash) {
             revert L1_UNEXPECTED_PARENT();
         }
 
+        // Verify the passed in L1 state block number.
+        // We only allow the L1 block to be 2 epochs old.
```
dantaik, line 88: is "2 epoch" a special value? 

If we use epoch, shall we check `params.l1StateBlockNumber/32 >= block.number/32 - 2`

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -74,35 +75,47 @@ library LibProposing {
...
+        }
+
         // Initialize metadata to compute a metaHash, which forms a part of
         // the block data to be stored on-chain for future integrity checks.
         // If we choose to persist all data fields in the metadata, it will
         // require additional storage slots.
         unchecked {
             meta_ = TaikoData.BlockMetadata({
-                l1Hash: blockhash(block.number - 1),
+                l1Hash: blockhash(params.l1StateBlockNumber),
```
dantaik, line 105: I think even if we keep `l1Hash: blockhash(block.number - 1)` it should still work, right?

dantaik: Same question for `timestamp`, `l1Height` and `difficulty` below.  If in Raiko we need to access `l1StateBlockNumber`, we can still keep`timetamp` and `difficulty` as-is and only change `l1Height`. 12 seconds per L1 block is not accurate.

packages/protocol/contracts/L1/TaikoData.sol
dantaik, line 128: this will use the 4th slot

Brechtpd: The L1 block that will be used in the anchor tx. So the L1 block for which the state will be available on L2. It's the name I also used in raiko, haven't though of a better one.

packages/protocol/contracts/L1/SequencerRegistry.sol
Brechtpd: I'm following the API spec from limechain for the naming here, potentially will be close to the actual name used sometime later.

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: No real special value, but this window should be large enough for a preconfer to always be able to select an L1 block that he can later propose in his own slot. That should only be up to 2 epochs of L1 slots.

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: The thing necessary for preconfirmations is that the whole L2 block is fully deterministic within a reasonable window to give some extra flexibility to when the block can be proposed without changing how the block will execute.

`timestamp`, `l1Height`,`l1BlockHash`, and `difficulty` can all impact the execution of transactions on L2, and so they also need to be some predictable value even if the block is being proposed at some later L1 slot.

packages/protocol/contracts/L1/TaikoData.sol
dantaik: Maybe add a comment for it, something like `l1StateBlockNumber will be the value for _l1BlockId in TaikoL2's anchor() function`.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -74,35 +75,47 @@ library LibProposing {
...
             revert L1_UNEXPECTED_PARENT();
         }
 
+        // Verify the passed in L1 state block number.
+        // We only allow the L1 block to be 2 epochs old.
+        // The other constraint is that the L1 block number needs to be larger than or equal the one
+        // in the previous block.
+        if (
+            params.l1StateBlockNumber < block.number - 64
+                || params.l1StateBlockNumber >= block.number
```
dantaik, line 93: so previously before this PR, anchor's `_l1BlockId` is automatically selected, after this PR, the proposer can choose a value?

dantaik: Is it ok to have `params.l1StateBlockNumber == block.number`?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -74,35 +75,47 @@ library LibProposing {
...
                 blobHash: 0, // to be initialized below
                 extraData: params.extraData,
                 depositsHash: _EMPTY_ETH_DEPOSIT_HASH,
                 coinbase: params.coinbase,
                 id: b.numBlocks,
                 gasLimit: _config.blockMaxGasLimit,
-                timestamp: uint64(block.timestamp),
-                l1Height: uint64(block.number - 1),
+                // Use the timestamp of the L1 state block
+                timestamp: uint64(block.timestamp - 12 * (block.number - params.l1StateBlockNumber)),
```
dantaik, line 114: if `params.l1StateBlockNumber == block.number` is not allowed, L2's timestamp is always at least 12 second smaller than L1's `block.timestamp`. Shall we use `uint64(block.timestamp - 12 * (block.number  - 1 - params.l1StateBlockNumber))` instead?

dantaik: Since L2's block time is not strictly 12 seconds, calculating L2's timestamp using l1StateBlockNumber is always not accurate, maybe we can simply just use `block.timestamp`?

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -74,35 +75,47 @@ library LibProposing {
...
-        if (params.parentMetaHash != 0 && parentMetaHash != params.parentMetaHash) {
+        if (params.parentMetaHash != 0 && parentBlock.metaHash != params.parentMetaHash) {
             revert L1_UNEXPECTED_PARENT();
         }
 
+        // Verify the passed in L1 state block number.
+        // We only allow the L1 block to be 2 epochs old.
+        // The other constraint is that the L1 block number needs to be larger than or equal the one
+        // in the previous block.
+        if (
```
dantaik, line 91: can we have a config flag to allow `l1StateBlockNumber` to be zero to fallback to the previous logic?

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: > so previously before this PR, anchor's `_l1BlockId` is automatically selected, after this PR, the proposer can choose a value?

yes!

> Is it ok to have `params.l1StateBlockNumber == block.number`?

Not possible because the blockhash of the current block is not available (also the reason we include the previous L1 block now).

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: > Since L2's block time is not strictly 12 seconds, calculating L2's timestamp using l1StateBlockNumber is always not accurate, maybe we can simply just use `block.timestamp`?
 
L1s block time is always 12 seconds, so I think it's the best way we can calculate the latest timestamp while still making sure that the value is predictable for the proposer who starts building an L2 block using `l1StateBlockNumber `. Using `block.timestamp` would make it so the preconfer HAS to include the L2 block being in some specific L1 block to make sure the L2 block execution goes as expected (which the preconfer cannot always do).

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: Can do!

dantaik (approved):

dantaik:
Brecht, my understanding is that if we have two preconfers on slot 5 and 31, only the first perconfer can propose blocks in slot 0 to slot 5, and only the second preconfer can propose block in slot 6 to 31 (this is Justin's original proposal I believe). The `l1StateBlockNumber` in your PR is to help these preconfers to propose blocks with new restrictions so that L2 blocks proposed by a preconfer will not be included in other *out-of-range* L1 blocks unexpected.

In the implementation of `isEligibleSigner`, should we also check if the msg.sender can propose block for `block.number % 32`?

packages/protocol/contracts/L1/libs/LibProposing.sol
adaki2004: Is it like a "64 block L1 block buffer, to find L1 slot for your transaction?"
So on L2, we might just mean, regarding Bob's transaction:
"Ok, your 1eth swap to WBTC will be included and you can consider it valid on L2" but we will propose it somewhere in the next 64 (L1) blocks ?

Also this kind of indicates that honest people (preconferms) can be only whitelised, otherwise they might just front-run and take advantage of that TXN ? And if one does so, we need some kind of slashing, but how to detect this behaviour off-chain then slash on-chain ?

(I guess it is just a POC tho, so for testnet we shall iterate from it).

Brechtpd:
Two more changes in latest version:
- Also allow the L2 block's timestamp to be chosen by the proposer with similar constraints as the `l1StateBlockNumber`. This is to solve potential problems if there are missed slots which made the previous calculation not deterministic (thanks gd for noticing this!)
- Saw that the difficulty calculation still contained a dependency on `block.number` which also made it not deterministic. It is calculated like this: `difficulty = keccak256(abi.encodePacked(blockhash(params.l1StateBlockNumber), b.numBlocks)); `

@davidtaikocha @dantaik @YoGhurt111

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: > "Ok, your 1eth swap to WBTC will be included and you can consider it valid on L2" but we will propose it somewhere in the next 64 (L1) blocks ?

yep!

> Also this kind of indicates that honest people (preconferms) can be only whitelised (because of the fallback 0), otherwise they might just front-run and take advantage of that TXN ? And if one does so, we need some kind of slashing, but how to detect this behaviour off-chain then slash on-chain ?

Yeah they will need to register in the sequencer registry, where they will have show that they have some stake that can get slashed when it can be proven they didn't fulfill their preconfirmation. When the are registered, the sequencer registry will select a single preconfer per L1 slot so there can not be any frontrunning between them.

> (I guess it is just a POC tho, so for testnet we shall iterate from it).

Currently the testnet will not enforce anything except blank whitlist of preconfers that indeed all have to trust each other and users also have to trust them because nothing at stake.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -64,6 +64,8 @@ library TaikoData {
         bytes32 parentMetaHash;
         HookCall[] hookCalls; // DEPRECATED, value ignored.
         bytes signature;
+        uint32 l1StateBlockNumber;
```
dantaik, line 67: in order to merge this PR to main in the future, I think we should rename this struct to BlockParams2 and keep BlockParams as-is so we can support hardfork logics in the protocol.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -161,7 +195,9 @@ library LibProposing {
             // For a new block, the next transition ID is always 1, not 0.
             nextTransitionId: 1,
             // For unverified block, its verifiedTransitionId is always 0.
-            verifiedTransitionId: 0
+            verifiedTransitionId: 0,
+            l1StateBlockNumber: params.l1StateBlockNumber,
```
dantaik, line 199: Maybe we don't need to write l1StateBlockNumber or timestamp to the storage, as long as these two values are available in the metadata. proveBlock also has metadata as calldata input.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -74,35 +88,58 @@ library LibProposing {
...
+        if (params.parentMetaHash != 0 && parentBlock.metaHash != params.parentMetaHash) {
             revert L1_UNEXPECTED_PARENT();
         }
 
+        // Verify the passed in L1 state block number.
+        // We only allow the L1 block to be 2 epochs old.
+        // The other constraint is that the L1 block number needs to be larger than or equal the one
+        // in the previous L2 block.
+        if (
+            params.l1StateBlockNumber < block.number - 64
```
dantaik, line 105:
```suggestion
            params.l1StateBlockNumber +64 < block.number
```
```suggestion
            params.l1StateBlockNumber < block.number - 64
```
```suggestion
            params.l1StateBlockNumber < block.number - 64
```

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -74,35 +88,58 @@ library LibProposing {
...
+        ) {
+            revert L1_INVALID_L1_STATE_BLOCK();
+        }
+
+        // Verify the passed in timestamp.
+        // We only allow the timestamp to be 2 epochs old.
+        // The other constraint is that the timestamp needs to be larger than or equal the one
+        // in the previous L2 block.
+        if (
+            params.timestamp < block.timestamp - 64 * 12 || params.timestamp > block.timestamp
```
dantaik, line 117:
```suggestion
            params.timestamp + 64 * 12 < block.timestamp || params.timestamp > block.timestamp
```

packages/protocol/contracts/L1/SequencerRegistry.sol
adaki2004: Should not we need (mayber dummy tho) `eligibleCountAt` and `sequencerByIndex`, `activate`, `register`, `withdraw`, `stake` functions too ? (Not sure if still needed, but have seen in the PoC proposal).

YoGhurt111 (approved):

Brechtpd:
Increased the time window for the L1 data to 4 epochs (from 2 epochs). The reason is that reorgs are possible up to 2 epochs in ethereum, and the preconfer will need to use L1 data that is 100% safe if he doesn't want to risk getting slashed.

The window is now very long though, we'll have to see if this is actually needed or not.

Brechtpd:
For the block building part, this is what  a preconfer normally needs to do:

- preconfer waits until it's his turn
- the preconfer chooses the L1 data:
  -  `l1StateBlockNumber` by checking the current tip of the L1 chain block number, and subtracting 64 (to make sure the selected L1 block cannot be reorged out)
  - the preconfer chooses a `timestamp` by checking the current tip of the L1 chain block timestamp, and subtracting 12*64
- the preconfer starts building a block
   - sets block parameters:
     - set the block timestamp to the chosen `timestamp` above
     - set the block difficulty to `keccak256(abi.encodePacked(blockhash(l1StateBlockNumber), parent_l2_block.header.number + 1))`
     - set the block gas_limit to `config.blockMaxGasLimit`
     - all other values are default ethereum values
   - adds the anchor tx by generating it using the `l1StateBlockNumber` data above using `AssembleAnchorTx`:
	    - `l1Height` = `l1StateBlockNumber`
	    - `l1Hash` = `blockhash(l1StateBlockNumber)`,
	    - `l2Height` = `parent_l2_block.header.number + 1`,
	    - `baseFee` =  `calc_new_basefee()`
	    - `parentGasUsed` = `parent_l2_block.header.gas_used`
   - Listens to preconfirmation requests, for each tx
      - validate if the transaction is executable, if not reject
      - add tx to the tip of the block
      - generate receipt for user
   - when block is full, or time is up, stop
- the preconfer proposes the block
  - if in his own L1 slot: make sure it's included in his own L1 block
  - if in sombody else's L1 slot: create an L1 tx and try to propose it
    - keeping trying until it's the preconfer's own L1 slot

davidtaikocha:
Looks like there is a deployment error: `TaikoL1` is above the contract size limit (25459 > 24576).

Brechtpd:
Disabled some checks/hardcoded the mintier to make it deployable so nothing needs to be changed to the deployment itself.

PR #17679, opened 2024-06-26
feat(protocol): prepare `ontake` fork to support new params & metadata

dantaik:
BEGIN_COMMIT_OVERRIDE
feat(protocol): allow contract proposers to use calldata as well
feat(protocol): add preconfirmation support based on [Brecht's PR](https://github.com/taikoxyz/taiko-mono/pull/17654) (with some renaming) to implement https://github.com/taikoxyz/taiko-mono/issues/14793 
feat(protocol): add `proposeBlock2` in TaikoL1.sol and `approve2` in GuardianProver.sol to be used post a fork
refactor(protocol): avoid writing livenessBond to storage to remove unnecessary SSTORE and SLOAD.
refactor(protocol): enable AddressManager to cache address(0) without reading from storage
END_COMMIT_OVERRIDE

## TODO
- [x] merge Brecht's preconfirmation changes to this PR.
- [x] Testing the hardfork and proposeBlock2 function


packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -153,7 +175,7 @@ library LibProposing {
         // Ethereum block, we choose to introduce a salt to this random
         // number as the L2 mixHash.
         meta_.difficulty =
-            keccak256(abi.encodePacked(block.prevrandao, local.b.numBlocks, block.number));
+            keccak256(abi.encodePacked(local.params.anchorBlockId, local.b.numBlocks, block.number));
```
Brechtpd, line 178: This reverts back to the dependency on block.number that I fixed in my PR: https://github.com/taikoxyz/taiko-mono/pull/17654#issuecomment-2189462000. Using `blockhash(params.l1StateBlockNumber)` also indirectly does depend on a randao instead of just using `l1StateBlockNumber`.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -124,6 +149,8 @@ library TaikoData {
         // this block is not verified as the last block in a batch, verifiedTransitionId
         // will remain zero.
         uint32 verifiedTransitionId;
+        uint64 anchorBlockId; // slot 4
+        uint64 anchorTimestamp;
```
Brechtpd, line 153: Not saying my names are necessarily better, but anchorX names are also a bit confusing haha. Because `anchorTimestamp` is never used in the anchor, it's just the timestamp put in the L2 block header. And `anchorBlockId` is not the anchor call's block id, but I guess this one somehow could make more sense to people than my L1 state block name. But I think `anchorTimestamp` can just remain `timestamp`.

packages/protocol/contracts/L1/libs/LibProposing.sol
dantaik: I update the code but it is still different from your original change. Please take another look.

packages/protocol/contracts/L1/TaikoData.sol
dantaik: `anchorTimestamp` renamed to `timestamp`.

packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -68,55 +70,101 @@ library LibProposing {
...
+            // We only allow the L1 block to be 2 epochs old.
+            // The other constraint is that the L1 block number needs to be larger than or equal the
+            // one
+            // in the previous L2 block.
+            if (
+                params.anchorBlockId + 64 < block.number //
+                    || params.anchorBlockId >= block.number
+                    || params.anchorBlockId < parentBlk.anchorBlockId
+            ) {
+                revert L1_INVALID_ANCHOR_BLOCK();
```
adaki2004, line 124: Out of curiosity. This 2-epoch check is to have a "foreseeable" timeframe in which we can confidently say the preconfers will include L2 blocks for sure.
But what happens in case on L2 we execute, but (for whatever reason) we skip 64 L1 blocks in the meantime (system downtime, wtwr) and no proposal happens on L1 ? In such case those L2 blocks cannot be proposed anymore.. (?) all of those blocks are not valid and TaikoL2 needs to have a reorg, right ? Proposer needs to propose with `anchorBlockId == 0` to be able to pass this check but the `not-yet-proposed-but-preconfed-txns` (within this 64 slots, which was missed) are gone.
Should not we need a failsafe mechanism or that is the nature of this and we are OK with that ?

dantaik: > In such case those L2 blocks cannot be proposed anymore.. 

Even if there is a long time, much longer than 64 blocks, on L1 without any L2 blocks proposed, proposers can still propose new blocks.

dantaik:
> > ## TODO
> > ```
> > * [x]  merge Brecht's preconfirmation changes to this PR.
> > ```
> 
> Just flagging `SequencerRegistry` is missing (but i guess first you just want to make it compatible, then do the interface/function updates).

True, didn't want to replace the previous PR.

packages/protocol/contracts/L1/TaikoData.sol
Brechtpd: Oh yes yes it's the value of `_l1BlockId`! Just that you people may think that `anchorBlockId` is the block in which the anchor is included, but I guess always different interpretations possible. :)

packages/protocol/contracts/L1/libs/LibProposing.sol
Brechtpd: Yep, somebody just needs to build a block with more recent L1 data.

adaki2004: Sure i meant, that the "promised" (preconfirmed TXN) L2 blocks are gone, but chain is operational!

packages/protocol/contracts/L1/access/IProposerAccess.sol
```diff
@@ -0,0 +1,13 @@
...
+/// @title IProposerAccess
+/// @dev An interface to check if a proposer is eligible to propose blocks in a specific Ethereum
+/// block.
+/// @custom:security-contact security@taiko.xyz
+interface IProposerAccess {
+    /// @notice Checks if a proposer can propose block in the current Ethereum block.
+    /// @param _proposer The proposer.
+    /// @return true if the proposer can propose blocks, false otherwise.
+    function isProposerEligible(address _proposer) external view returns (bool);
+}
```
adaki2004, line 13: Just for client interface point of view, would be good to use the exact same namings like [here](https://github.com/taikoxyz/taiko-mono/blob/015ed3499f28002f42e27164812bfd4c077f7c0b/packages/protocol/contracts/L1/ISequencerRegistry.sol), in [this](https://github.com/taikoxyz/taiko-mono/pull/17654) PR,  if we wanna reuse the backend, OR use this naming there.. But definitely would be cool to avoid duplicated work later.

dantaik: I don't like the word `sequencer`. And I also believe going forward, ISequencerRegistry will also change due to the dependency on restaking platforms/protocols.

Brechtpd: `ISequencerRegistry` very likely stable because how the implementation selects the current proposer does not matter to taiko.

Sequencer vs proposer is subjective, I'm fine with both but people do tend to use the word sequencer here to stress that they are sequencing the blocks and there's only one.

dantaik: We can change it later. I think we just care about the permission checking function rather than the registration logics.

PR #17693, opened 2024-06-27
feat(eventindexer): add graceful termination

bennettyong:


packages/eventindexer/cmd/utils/subcommand_action.go
```diff
@@ -35,27 +36,47 @@ func SubcommandAction(app SubcommandApplication) cli.ActionFunc {
...
-		if err := startMetrics(); err != nil {
-			slog.Error("Starting metrics server error", "error", err)
+		metricsErrCh := make(chan error, 1)
+
+		go func() {
+			if err := startMetrics(); err != nil {
+				slog.Error("Starting metrics server error", "error", err)
+				metricsErrCh <- err
+			} else {
+				metricsErrCh <- nil
```
mask-pp, line 48: This line will never be executed.

packages/eventindexer/pkg/metrics/metrics.go
```diff
@@ -20,14 +19,6 @@ func Serve(ctx context.Context, c *cli.Context) (*echo.Echo, func() error) {
 	e := echo.New()
 	p.SetMetricsPath(e)
 
-	go func() {
```
mask-pp, line 23: What's the reason for deleting this code?

mask-pp:
Can you point out what specific bug this PR is trying to address? All I see are some coding style changes.

PR #17705, opened 2024-06-29
fix(protocol): allow `srcOwner` to have unlimited gas limit on dest retry invocation

adaki2004:
Beside `destOwner`, also allow `srcOwner` to call `retryMessage` with non restrictive gas limit. (This is due to the fact of [discord messages ](https://discord.com/channels/984015101017346058/1206245579647229984/1256492505835372655) in question.)

The only downside of this is a very rare possibility of: someone on the destination chain, control the `srcOwner` address and make the message failed -> but even in this case, highly unlikely that anyone would set up a sniper bot for this action since it would result in 0 capital gains, as the funds were be available to re-claim it on the source chain, by the original `srcOwner`, who is indeed in charge on the source chain.

Given the fact we ran into an edge case with contracts being the destination owners, but lacking the relay function to `retryMessage` with unlimited gas, mght be worth considering this addition 

OR

If the option is to let them upgrade their contracts (as per discord message link suggestions above), than it's OK to close and not merge this PR.

### DIFF
--- a/packages/protocol/contracts/bridge/Bridge.sol
+++ b/packages/protocol/contracts/bridge/Bridge.sol
@@ -332,12 +332,12 @@ contract Bridge is EssentialContract, IBridge {
             succeeded = _message.destOwner.sendEther(_message.value, _SEND_ETHER_GAS_LIMIT, "");
         } else {
             uint256 invocationGasLimit;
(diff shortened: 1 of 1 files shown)


dantaik:
@adaki2004 What do you think we simply use `gaslef()` in `retryMessage()`.

PR #17708, opened 2024-06-30
fix(protocol): use gasleft() in Bridge's retryMessage function

dantaik:


adaki2004 (approved):
LGTM, might be good to go out with the tomorrow update - so that the issue in subject (SC is the destOwner and cannot raise gas limit) can be resolved.

Brechtpd (approved):

PR #17725, opened 2024-07-03
feat(protocol): allow TAIKO token bonds deposits and withdrawal

dantaik:
This PR implements the old idea of managing in-contract token balance to avoid:
1) resolving Taiko token address
2) multiple balance modifications in TaikoToken contract which also reads delegates and modifies voting powers.

Screenshot 2024-07-03 at 11 35 39

///

Screenshot 2024-07-03 at 11 36 14


davidtaikocha (approved):

adaki2004 (approved):
LGTM, 1 remark: Even tho we are not using the `batchTransfer() `in protocol, but i think it is still a useful feature to the token when someone wants to batch send for whatever reason.

dantaik:
> LGTM, 1 remark: Even tho we are not using the `batchTransfer() `in protocol, but i think it is still a useful feature to the token when someone wants to batch send for whatever reason.

Added it back.

davidtaikocha (approved):

PR #17742, opened 2024-07-05
chore(blobstorage): fix some carps

mask-pp:


davidtaikocha:
I think `blobstorage` is deprecated: https://github.com/taikoxyz/taiko-mono/pull/17278

PR #17745, opened 2024-07-05
feat(taiko-client): preconfirmations taiko client support contracts

cyberhorsey:


packages/taiko-client/driver/chain_syncer/blob/syncer.go
```diff
@@ -248,16 +248,41 @@ func (s *Syncer) onBlockProposed(
...
+
+	var txListBytes []byte
+	calldataTxListFound := false
+
+	// Iterate through the receipt logs
+	for _, vLog := range receipt.Logs {
+		// Use the generated binding's method to parse the log
+		event, err := s.rpc.TaikoL1.ParseCalldataTxList(*vLog)
+		if err == nil {
+			log.Info("CalldataTxList event found", "blockID", event.BlockId.String())
```
Brechtpd, line 266: Because now there can be multiple blocks being proposed in a single tx, we probably need to make sure that the block id here matches the one in the block proposed event we're handling?

cyberhorsey: pushed up!

davidtaikocha (approved):

Brechtpd (approved):

YoGhurt111 (approved):

PR #17762, opened 2024-07-07
chore(protocol): give more slots for verified blocks in ring buffer

dantaik:
Previous, we only keep the last 512 verified block's data in the ring buffer. This may be problematic if a third party app, such as a prover pool, tries to settle prover-proposer payments based on who actually proved the blocks. 

This PR allows up to 7200 * 5 = 36000 recently verified blocks to have data in the ring buffer.


Brechtpd (approved):

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -233,10 +238,13 @@ contract TaikoL1 is EssentialContract, ITaikoL1 {
...
-            // new blocks without any verification.
-            blockMaxProposals: 324_000, // = 45*86400/12, 45 days, 12 seconds avg block time
-            blockRingBufferSize: 324_512,
+            // If we have 1 block per 12 seconds, then each day there will be 86400/12=7200 blocks.
+            // We therefore use 7200 as the base unit to configure blockMaxProposals and
+            // blockRingBufferSize.
+            blockMaxProposals: 324_000, // = 7200 * 45
+            // We give 7200 * 5 = 36000 slots for verifeid blocks in case third party apps will use
+            // their data.
+            blockRingBufferSize: 360_000, // = 7200 * 50
```
davidtaikocha, line 247: For Hekla, we may need to change it to a larger number? since rn hekla already has ~425480 blocks?

adaki2004 (approved):

packages/protocol/contracts/L1/TaikoL1.sol
dantaik: Added TaikoL1Hekla.sol

davidtaikocha (approved):

PR #17787, opened 2024-07-12
feat(protocol): propose multiple blocks in one call

cyberhorsey:
this pr allows you to propose multiple blocks in one call, three different ways:
1) calldata (arrays of txLists and params as inputs to the function)
2) multiple blocks per blob (by passing in offsets and lengths as members on the blockParams struct)
3) as multiple blobs (by passing in blobIndex on the blockParams struct)

### DIFF
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -90,6 +90,8 @@ library TaikoData {
         bool blobUsed;
         bytes32 parentMetaHash;
         address sender; // a.k.a proposer
+        uint256 blobTxListOffset;
+        uint256 blobTxListLength;
     }
 
     /// @dev Struct representing transition to be proven.
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -174,7 +174,7 @@ library TaikoData {
         mapping(
             uint64 blockId_mod_blockRingBufferSize
                 => mapping(uint32 transitionId => TransitionState ts)
-        ) transitions;
+            ) transitions;
         // Ring buffer for Ether deposits
         bytes32 __reserve1;
         SlotA slotA; // slot 5
--- a/packages/protocol/contracts/L1/TaikoData.sol
+++ b/packages/protocol/contracts/L1/TaikoData.sol
@@ -177,7 +177,7 @@ library TaikoData {
         mapping(
             uint64 blockId_mod_blockRingBufferSize
                 => mapping(uint32 transitionId => TransitionState ts)
-            ) transitions;
+        ) transitions;
         // Ring buffer for Ether deposits
         bytes32 __reserve1;
         SlotA slotA; // slot 5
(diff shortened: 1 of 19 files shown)


packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -89,13 +89,18 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
...
-        (meta_, deposits_) = LibProposing.proposeBlock(state, tko, config, this, _params, _txList);
+        for (uint256 i = 0; i < _params.length; i++) {
+            (meta_, deposits_) =
+                LibProposing.proposeBlock(state, tko, config, this, _params[i], _txLists[i]);
 
-        if (LibUtils.shouldVerifyBlocks(config, meta_.id, true) && !state.slotB.provingPaused) {
-            LibVerifying.verifyBlocks(state, tko, config, this, config.maxBlocksToVerify);
+            if (LibUtils.shouldVerifyBlocks(config, meta_.id, true) && !state.slotB.provingPaused) {
+                LibVerifying.verifyBlocks(state, tko, config, this, config.maxBlocksToVerify);
+            }
```
Brechtpd, line 103: I think a bit more efficient to move this out of the loop and use `config.maxBlocksToVerify * _params.length` as max blocks to verify.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -66,6 +66,9 @@ library TaikoData {
         bytes signature;
         uint32 l1StateBlockNumber;
         uint64 timestamp;
+        uint256 blobTxListOffset;
+        uint256 blobTxListLength;
```
Brechtpd, line 70: Also add these two to `BlockMetadata` because the prover will need to read the slice info from there.

packages/protocol/contracts/L1/TaikoL1.sol
cyberhorsey: will do!

packages/protocol/contracts/L1/TaikoData.sol
cyberhorsey: kk

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -89,13 +89,20 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
             }
         }
 
+        require(_params.length == _txLists.length, "mismatched params length");
```
adaki2004, line 92: I tend to think, it might be better like this below if we wanna support `calldata` efficiently. 

```
        if(txLists.length != 0) {
            require(params.length == _txLists.length, "mismatched params length"));
        }
```

So currently protocol can handle both, but with the above change, we would have to create empty `_txLists` -matched to the length of `_params`, even if we use blob OR some empty if we use blob + calldata both, separately, paralel with one L1 TXN.

I dont think it is practical to use different data avail. layer for multiple L2 blocks but one L1 TXN. (Maybe i'm wrong in here, and in this case inefficiency is i guess negligible).

cc. @Brechtpd  as it is more of a question arised from Gwyneth adaptation.

PR #17734, opened 2024-07-04
feat(protocol): apply multiple improvements for "ontake" fork

dantaik:
BEGIN_COMMIT_OVERRIDE
feat(protocol): allow contract proposers to use calldata for DA
feat(protocol): add preconfirmation support based on https://github.com/taikoxyz/taiko-mono/pull/17654 (with some renaming) (https://github.com/taikoxyz/taiko-mono/issues/14793)
feat(protocol): add proposeBlock2 in TaikoL1.sol and approve2 in GuardianProver.sol
refactor(protocol): avoid writing `livenessBond`, `proposedAt`, and `proposedIn` to storage
feat(protocol): relocate L2 base fee parameters to L1 configuration
END_COMMIT_OVERRIDE

PR #17788, opened 2024-07-13
feat(taiko-client): muliple blocks in one proposeBlock call, three ways

cyberhorsey:
1) array of txList and params
2) multiple blocks in one blob
3) multiple blocks in multiple blobs


packages/taiko-client/preconfapi/builder/blob.go
```diff
@@ -34,40 +35,79 @@ func NewBlobTransactionBuilder(
...
+
+		params := &encoding.BlockParams{
+			Coinbase:           common.HexToAddress(opt.Coinbase),
+			ExtraData:          rpc.StringToBytes32(opt.ExtraData),
+			Signature:          []byte{}, // no longer checked
+			L1StateBlockNumber: opt.L1StateBlockNumber,
+			Timestamp:          opt.Timestamp,
+		}
+
+		if opts.MultipleBlobs {
```
Brechtpd, line 69: Is there a reason to have this option instead of always trying to pack as many tx lists in a single blob? Maybe easier code for now so for now not trying to be smart?

packages/taiko-client/preconfapi/builder/calldata.go
```diff
@@ -32,30 +33,41 @@ func NewCalldataTransactionBuilder(
...
-	encodedParams, err := encoding.EncodeBlockParams(&encoding.BlockParams{
-		ExtraData:          extraData,
-		Coinbase:           coinbase,
-		Signature:          []byte{}, // no longer checked
-		L1StateBlockNumber: l1StateBlockNumber,
-		Timestamp:          timestamp,
-	})
-	if err != nil {
-		return nil, err
+	txLists := make([][]byte, 0)
```
Brechtpd, line 40: Just future improvement maybe, but I always found it a bit strange how this is a completely separate path from the blobs. Could this be merged with the blob path eventually where the code just picks one or the other depending on gas/blob gas cost eventually?

cyberhorsey: yeah definitely, right now we basically just start our proposer in one or the other mode, it can definitely just estimate fee and pick the smarter one later

packages/taiko-client/preconfapi/builder/blob.go
cyberhorsey: yeah it's just a lazy way for MVPs for testnet right now, next iteration can just check if the max bytes for a blob has been exceeded or not

cyberhorsey: pushed up a change

packages/taiko-client/preconfapi/builder/blob.go
```diff
@@ -56,33 +64,54 @@ func (b *BlobTransactionBuilder) BuildUnsigned(
...
 
-		params := &encoding.BlockParams{
-			Coinbase:           common.HexToAddress(opt.Coinbase),
-			ExtraData:          rpc.StringToBytes32(opt.ExtraData),
-			Signature:          []byte{}, // no longer checked
-			L1StateBlockNumber: opt.L1StateBlockNumber,
-			Timestamp:          opt.Timestamp,
+		// if we have exceeded max blob data size, we can make a blob with all the data
+		// *except* this one.
+		if len(totalWithCurrentTxList) >= eth.MaxBlobDataSize {
```
Brechtpd, line 71: `>` instead of `>=` ?

packages/taiko-client/preconfapi/builder/blob.go
dantaik: I would strongly prefer not to merge this feature into our main branch.

Brechtpd: Why?

cyberhorsey: > I would strongly prefer not to merge this feature into our main branch.

any reason why? it was an ask from a preconf team, but I can tell them no if we have a good reason.

YoGhurt111:
lgtm, but still need to fix unit test.

YoGhurt111 (approved):

packages/taiko-client/preconfapi/builder/blob.go
```diff
@@ -34,40 +35,101 @@ func NewBlobTransactionBuilder(
...
+		compressedTxListBytes, err := utils.Compress(txListBytes)
+		if err != nil {
+			return nil, err
+		}
+
+		totalWithCurrentTxList := append(totalBytes, compressedTxListBytes...)
+
+		// if we have exceeded max blob data size, we can make a blob with all the data
+		// *except* this one.
+		if len(totalWithCurrentTxList) >= eth.MaxBlobDataSize {
```
mask-pp, line 71: Use `len(totalBytes) + len(compressedTxListBytes) >=` instead of `len(totalWithCurrentTxList)`?

packages/taiko-client/preconfapi/builder/blob.go
```diff
@@ -34,40 +35,101 @@ func NewBlobTransactionBuilder(
...
+			var blob = &eth.Blob{}
+			if err := blob.FromData(totalBytes); err != nil {
+				return nil, err
+			}
+
+			blobs = append(blobs, blob)
+
+			// clear the bytes array and increment the blob index for next iterations of the loop
+			totalBytes = []byte{}
```
mask-pp, line 80: Why clean `totalBytes` at there but not when `blob.FromData(totalBytes)` is called?

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -89,14 +89,12 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents, TaikoErrors {
             }
         }
 
-        require(_params.length == _txLists.length, "mismatched params length");
-
         TaikoData.Config memory config = getConfig();
         TaikoToken tko = TaikoToken(resolve(LibStrings.B_TAIKO_TOKEN, false));
 
-        for (uint256 i = 0; i < _params.length; i++) {
+        for (uint8 i = 0; i < _params.length; i++) {
```
mask-pp, line 95: If we use `uint8` we should make sure `_params.length` is less than uint8.Max value.

packages/taiko-client/preconfapi/builder/interface.go
```diff
@@ -2,18 +2,54 @@ package builder
...
-		extraData [32]byte,
+		opts BuildUnsignedOpts,
 	) (*types.Transaction, error)
 }
+
+func signedTransactionsToTxListBytes(txs []string) ([]byte, error) {
+	var transactions types.Transactions
+
+	for _, signedTxHex := range txs {
+		signedTxHex = strings.TrimPrefix(signedTxHex, "0x")
```
mask-pp, line 34: Use `hexutil.Decode`, then don't need to trim prefix `0x`.

packages/taiko-client/preconfapi/builder/blob.go
cyberhorsey: we do, it's right above this

packages/protocol/contracts/L1/TaikoL1.sol
cyberhorsey: good point, I don't think anyone would do 256 blocks at once because it would run out of gas though

packages/taiko-client/preconfapi/builder/interface.go
cyberhorsey: will do!

cyberhorsey: the user is passing these in, we don't know if they have 0x or not

mask-pp: `hexutil.Decode` function will do this kind of inspection work.

packages/taiko-client/preconfapi/builder/blob.go
mask-pp: Put them together be easier to understand？

packages/protocol/contracts/L1/TaikoL1.sol
mask-pp: There is no harm in us thinking about this situation, but when it happens, it is fatal.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -66,9 +66,9 @@ library TaikoData {
         bytes signature;
         uint32 l1StateBlockNumber;
         uint64 timestamp;
-        uint256 blobTxListOffset;
-        uint256 blobTxListLength;
-        uint256 blobIndex;
+        uint64 blobTxListOffset;
+        uint64 blobTxListLength;
+        uint8 blobIndex;
```
mask-pp, line 71: The choice to use `uint8` may iincrease gas payment, the same to `uint64`.

packages/taiko-client/preconfapi/builder/blob.go
dantaik: If in one Ethereum transaction,  the same blob is used for more than one L2 block, then the proposer shall simply propose one larger block instead of two to avoid unnecessary cost overhead. If these two blocks are not sharing the same blob, I'd prefer each block is proposer separately using 'proposeBlock' to avoid chaning 'proposeblock' to 'proposeBlocks' (or another wrapper function can be added either in the contract or in a wrapper contract).

packages/protocol/contracts/L1/TaikoData.sol
dantaik: FYI, modifying the code this way is OK for a fresh new testnet, but it will break the current mainnet upgradability, therefore, the code shall never be merged into main.

cyberhorsey: > FYI, modifying the code this way is OK for a fresh new testnet, but it will break the current mainnet upgradability, therefore, the code shall never be merged into main as-is, new BlockParams struct must be defined, please see #17734 to learn how to make breaking changes.

Yep, until something is "locked in" we will just do it like this, to avoid wasting time over engineering.

packages/taiko-client/preconfapi/builder/blob.go
cyberhorsey: what?

packages/taiko-client/preconfapi/builder/blob.go
cyberhorsey: > If in one Ethereum transaction, the same blob is used for more than one L2 block, then the proposer shall simply propose one larger block instead of two to avoid unnecessary cost overhead. If these two blocks are not sharing the same blob, I'd prefer each block is proposer separately using 'proposeBlock' to avoid chaning 'proposeblock' to 'proposeBlocks' (or another wrapper function can be added either in the contract or in a wrapper contract).

OK, so you are fine with adding "proposeBlocks" function separately, and to remove logic of multiple blocks in one blob.

packages/protocol/contracts/L1/TaikoL1.sol
cyberhorsey: Logic will be changed and moved outside to a different function and I will change to handle this scenario

packages/taiko-client/preconfapi/builder/blob.go
mask-pp: Use it and clean it in the same place.

packages/taiko-client/preconfapi/builder/blob.go
dantaik: Yes, as long as `proposeBlocks` simply calls '_proposeBlock' multiple times without different logics :D

packages/protocol/contracts/L1/TaikoData.sol
dantaik: Great, later lets figure out together when/how to do a upgrade, then cherrypick your changes into a new fork-PR.

packages/taiko-client/preconfapi/builder/blob.go
cyberhorsey: it's literally the previous line of code? 
Screenshot_20240715-204258

packages/taiko-client/preconfapi/builder/blob.go
Brechtpd: Multiple blocks inside a single blob is also a feature request of nethermind. Their preconfs would commit to the tx trie root which is only available in a block header, and they will do so every 3 seconds. I think in theory this does not necessarily require multiple blocks, but it certainly makes checking the preconf condition very easy (check the value in the block header), they will be creating multiple L2 blocks per L1 block in any case. Blob data slices also makes it possible to share blob data with other based rollups.

The overhead of multiple blocks can also be avoided, though that it is a big change so unlikely to be done for the current protocol version.

packages/taiko-client/preconfapi/builder/blob.go
mask-pp: Actually, Line 73 already exists one.

packages/protocol/contracts/L1/TaikoData.sol
```diff
@@ -66,9 +66,9 @@ library TaikoData {
         bytes signature;
         uint32 l1StateBlockNumber;
         uint64 timestamp;
-        uint256 blobTxListOffset;
-        uint256 blobTxListLength;
-        uint256 blobIndex;
+        uint64 blobTxListOffset;
```
dantaik, line 69: uint32 is good enough?

cyberhorsey:
closed, added to preconfirmations branch

PR #17792, opened 2024-07-14
fix(protocol): reduce MainnetTaikoL1 code size

dantaik:
MainnetTaikoL1's code is too large. In this PR, I changed a library function to public, added back TaikoEvents (otherwise client cannot handle public library events), and changed deploy scripts to test deploying MainnetTaikoL1.

### DIFF
--- a/packages/protocol/contracts/L1/TaikoEvents.sol
+++ b/packages/protocol/contracts/L1/TaikoEvents.sol
@@ -1,0 +1,85 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.24;
+
+import "./TaikoData.sol";
+
+/// @title TaikoEvents
+/// @notice This abstract contract provides event declarations for the Taiko
+/// protocol, which are emitted during block proposal, proof, verification, and
+/// Ethereum deposit processes.
+/// @dev The events defined here must match the definitions in the corresponding
+/// L1 libraries.
+/// @custom:security-contact security@taiko.xyz
+abstract contract TaikoEvents {
+    /// @notice Emitted when a block is proposed.
+    /// @param blockId The ID of the proposed block.
+    /// @param assignedProver The address of the assigned prover.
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/L1/TaikoEvents.sol
```diff
@@ -0,0 +1,85 @@
...
+import "./TaikoData.sol";
+
+/// @title TaikoEvents
+/// @notice This abstract contract provides event declarations for the Taiko
+/// protocol, which are emitted during block proposal, proof, verification, and
+/// Ethereum deposit processes.
+/// @dev The events defined here must match the definitions in the corresponding
+/// L1 libraries.
+/// @custom:security-contact security@taiko.xyz
+abstract contract TaikoEvents {
```
dantaik, line 13: this file is added back to ensure libraries with public function can be used.

packages/protocol/script/DeployOnL1.s.sol
```diff
@@ -245,6 +245,20 @@ contract DeployOnL1 is DeployCapability {
         copyRegister(rollupAddressManager, _sharedAddressManager, "signal_service");
         copyRegister(rollupAddressManager, _sharedAddressManager, "bridge");
 
+        deployProxy({
+            name: "mainnet_taiko",
```
Brechtpd, line 249: Do we still need the TaikoL1 deployment below?

davidtaikocha (approved):

packages/protocol/script/DeployOnL1.s.sol
dantaik: Maybe not, but TaikoL1 is "cleaner" (without the address caching), so I keep using it in case the contact is actually called. MainnetTaikoL1 cannot be invoked as the cached addresses don't have code deployed.

Brechtpd (approved):

PR #17802, opened 2024-07-16
fix(taiko-client): avoid seting nil value to `GuardianProverHeartbeater`

mask-pp:


YoGhurt111 (approved):

packages/taiko-client/prover/init.go
```diff
@@ -218,7 +218,7 @@ func (p *Prover) initEventHandlers() error {
 		ContesterMode:         p.cfg.ContesterMode,
 		ProveUnassignedBlocks: p.cfg.ProveUnassignedBlocks,
 	}
-	if p.IsGuardianProver() {
+	if p.IsGuardianProver() && p.guardianProverHeartbeater != nil {
```
davidtaikocha, line 221: If we are running a Guardian prover without a heartbeater, we still need to call `NewBlockProposedEventGuardianHandler`?

davidtaikocha (approved):

PR #17806, opened 2024-07-17
refactor(protocol): name address manager param clearer

dantaik:
Given that we have two address manages (shared and rollup), naming parameters this way is less confusing.


smtmfft (approved):

swarna1101:
this looks good, and the parameter naming is much more clearer now

swarna1101 (approved):

PR #17819, opened 2024-07-19
docs(repo, docs-site): add release info links

dionysuzx:
This PR adds dynamic buttons that display the latest versions of our subpackages in the taiko-mono README. It also adds a new page "node releases" to our documentation site. Also some other small updates.

@RogerLamTd maybe we should also extend the node releases page to include version info about the latest deployed protocol versions that are live in each testnet. this would be a bit trickier though, because testnet/mainnet are not guaranteed to be running the same versions so a manual step might be necessary. or we can read it from some deployment artifact elsewhere. let me know your thoughts.

image
image


RogerLamTd (approved):

RogerLamTd:
@d1onys1us can merge for now, we can work out the artifact thing or just manually update as we go!

davidtaikocha (approved):

d1onys1us:
@davidtaikocha @RogerLamTd I am putting this back into draft mode as Roger rose a concern about how we may release a tag, use it in testnet, but not recommend it for mainnet yet. We may need to come up with a different solution (pre-releases, for example). I will re-open after implementing a good solution. in a follow-up PR we can add protocol version tags.

RogerLamTd (approved):

PR #17822, opened 2024-07-21
feat(protocol): relocate L2 basefee parameters to L1 configuration

dantaik:
This pull request introduces significant improvements to our L2's base fee mechanics and gas issuance, addressing several key issues and aligning our approach more closely with Ethereum's design:

### 1. Gas Issuance Per L2 Block:
- Implement gas issuance of `blockGasIssuance` per L2 block.
- Deprecate the `_gasTargetPerL1Block` configuration item.
- Align with Ethereum's design principles.
- Mitigate the issue of block proposers manipulating base fees by choosing minimal valid anchoring block heights (which prevents new L2 gas issuance).

### 2. L1 Configuration for Base Fee Adjustment:
- Move `basefeeAdjustmentQuotient` to L1 configuration.
- Inject both `basefeeAdjustmentQuotient` and `blockGasLimit` into the new `anchorV2` transaction.
- Enable immediate adoption of configuration changes without coordinated L1-L2 hard forks.

### 3. Enhanced Base Fee Prediction:
- Implement new `calculateBaseFee` function to predict base fees for multiple blocks.
- Function is pure and does not rely on previous fee-updating writes to predict future block base fees.

### 4. Base fee sharing
- Introduced a new configuration item `basefeeSharingPctg` to dictate how much base fee should be shared with block.coinbase.

@Brechtpd please review again.


Brechtpd (approved):

Brechtpd:
Setting the target to `blockMaxGasLimit/2` may impact the efficiency of the protocol (at least in the current protocol), because the gas target is now directly related to the block size, and so we can't set the block size to a very high number anymore depending on what the provers support because that now automatically increases the max sustained throughput the chain needs to support.

dantaik:
> Setting the target to `blockMaxGasLimit/2` may impact the efficiency of the protocol (at least in the current protocol), because the gas target is now directly related to the block size, and so we can't set the block size to a very high number anymore depending on what the provers support because that now automatically increases the max sustained throughput the chain needs to support.

dantaik:
> Setting the target to `blockMaxGasLimit/2` may impact the efficiency of the protocol (at least in the current protocol), because the gas target is now directly related to the block size, and so we can't set the block size to a very high number anymore depending on what the provers support because that now automatically increases the max sustained throughput the chain needs to support.

I see. I think it's more flexible to keep `blockMaxGasLimit` and `blockGasIssuance` separate. Updated.

PR #17823, opened 2024-07-21
fix(protocol): fix BridgedERC20V2.sol initializer logic

dantaik:
BridgedERC20V2 could not be used to upgrade BridgedERC20 without the new `init2()` function. I also changed `init()` to mark implementation as version `2` so contract upgraded from `BridgedERC20` will have the same version number as newly deployed `BridgedERC20V2`.


packages/protocol/contracts/tokenvault/BridgedERC20V2.sol
```diff
@@ -41,7 +42,7 @@ contract BridgedERC20V2 is BridgedERC20, IERC20PermitUpgradeable, EIP712Upgradea
         external
         virtual
         override
-        initializer
+        reinitializer(2)
```
adaki2004, line 45: I was aware of the issue with the previous V2, but i thought we are not upgrading the previous tokens but going forward we will be using this in the bridge -> this case it would have been ok.

now, you need 2 transactions then in the bridge when a new token is bridged. One init, and an another one which initializes the eip712, or am i missing something ?

dantaik: > now, you need 2 transactions then in the bridge when a new token is bridged. One init, and an another one which initializes the eip712, or am i missing something ?

For new deployment, we only need to call `init` once, once it is called, `init2` can NOT be called as version is 2 after `init` is called. But for upgraded contracts, they need to call `init2` to change version from 1 to 2.

adaki2004: > For new deployment, we only need to call `init` once, once it is called, `init2` can NOT be called as version is 2 after `init` is called. But for upgraded contracts, they need to call `init2` to change version from 1 to 2.

Yes but init() is in normal BridgedERC20, no ? Does ir initializes the eip712 ? (From mobile and cannot see the code changes properly on small screen).

dantaik: Sorry, I don't understand the question...

adaki2004: Sorry, i did not see properly on mobile an cannot type my question. But after reviewing this solution, i think it works for retrospecitve obviously + future deployments, so all good it seems.

adaki2004 (approved):

Brechtpd (approved):

davidtaikocha (approved):

PR #17824, opened 2024-07-21
docs(protocol): typo fix

engn33r:


RogerLamTd:
thank you for your contribution but we don't include typos!

PR #17827, opened 2024-07-22
chore(relayer): combine bindings interface

mask-pp:


cyberhorsey:
these arent really `bindings` though, theyre interafaces in domain-driven design / clean architecture, `bindings` is an IMPLEMENTATION of these interfaces.

PR #17828, opened 2024-07-22
feat(protocol): incorporate L1 base fees into L2 fee structure

dantaik:
This PR proposes incorporating L1 base fees on top of the native L2 EIP-1559 base fee. The key challenge addressed is handling L1 base fees in the context of preconfirmation, where real-time access to L1 fees in the protocol is not feasible due to multiple L2 blocks being built for later submission.

To solve this, I implemented a caching mechanism for L1 base fees. This ensures that regardless of the anchoring block chosen by the proposer, the L1 base fees used in L2 blocks remain consistent and manipulation-resistant.

The change aims to create a more holistic fee structure reflecting both L1 and L2 costs. However, if the current epoch has very high base fees and the next epoch has lower base fees, the current implementation is buggy in that users pay more gas fee to the proposers in the next epoch, not the current epoch. The impact of L1 base fees to L2 base fees are always delayed by 1 epoch.

Optimism's approach (https://docs.optimism.io/stack/transactions/fees) actually breaks our type-1 claim as calldata cost and computational cost are decoupled and they use different gas-price (L1 cost contributes to L2 call data gas price). I'm not against such an approach, but I would argue that we need a lot more changes to go-ethereum and taiko-reth.

@Brechtpd @dantaik @davidtaikocha Please let me know if you have/prefer another approach.


dantaik:
@Brechtpd could you share some feedback/ideas?

Brechtpd:
No feedback yet because I wanted at least some concrete answer. :) 

How fees should be paid is a tricky thing, so hard to find good rules of something the protocol can enforce, because things can quickly change if certain things are optimized.
- Proposing cost ~= data publishing costs (which in an efficient rollup would ideally be the biggest part of the cost) would not really be related to the L1 base/blob fee * gas_used, instead it would be blob basefee * bytes used by the transaction. And even the bytes used by the transaction is dependent on how data availability is achieved (how much bytes the tx will actually use after compression, but also higher level tx list vs state delta if this would change). Some normal normal gas would also be used here, but should be very minimal, though because blobs are basically free now it currently could still be the bulk of the cost depending on how blocks are proposed.
- Proving cost: gas used here is also not an accurate way to approximate the actual cost (especially for zk proofs where the proving cost/gas is different depending on what the block does) and with proof aggregation the onchain cost calculation also gets complicated quickly.
- L2 basefee: measures congestion, so as long as it goes up and down based on how much gas the L2 uses should be fine
- L2 tip: within an L2 block, there could still be a tip based gas auction to get included in a block, so how much the tip should be is still hard to predict.
- In the future, there may even be a separate preconfirmation tip.

And then on top you have that these costs often need to be predicted in advance, maybe up to 32 L1 blocks in advance in the worst case.

So currently all these costs are part of the tip. And this tip somehow needs to pay for the combination of ~5 independent markets going on for the actual L2 tx cost: L1 gas market, L1 blob market, proving market, L1 tip market, L1/L2 preconfirmation tip market. For sure this already breaks the normal fee market of an L1 tx (but it was already broken because of MEV anyway, so no way back).

Some open questions:
- Is gas_used * some L1 basefee a good estimation that solves a problem? Is a single offer from the user not good enough that the proposer can reject/accept? I don't see people still waiting for >12 seconds for their transaction to be included on an L2 and hope for the best that the basefee goes down.
- Is it fair to let users pay a higher L2 basefee to the protocol just because the L1 basefee goes up? 
- What problem are we trying to solve that currently cannot be handled by the tip? 
- Can the protocol enforce a reasonable cost or should it be done offchain somehow? And if it can enforce it, should it be done on that level?

To me it seems almost impossible to have some protocol enforced mechanism that guarantees "fair" pricing to users/proposers. Maybe you can get a bit closer to it, but seems like a losing battle where truly efficient builders will just have to work around any protocol enforced mechanism. It could very well be that less is more in this case.

This doesn't mean that users have to be fully dependent of the builders now however! For example I really like this recent work of flashbots that will allow users to get a refund for their transaction if it turned out that they paid too much, and the code that decides this is run in a TEE so no need to even trust flashbots to do this, and the code is open source so people know what they will be getting: https://x.com/SheaKetsdever/status/1810680518360314190

As an L2, we could even enforce builders to build blocks like this (though perhaps that is too limiting to builders). And this code could be used to 100% accurately price the actual cost of a tx to ensure fairness. The only slight improvement needed on the L2 side is to have a separate field for actual tip used, so we don't have to always do max tip payment and then allow users to withdraw their refunds, but that's only a small improvements because tip estimation could also run in a TEE.

So this is a well of text, sorry. But my current conclusion: The current fee payment mechanism we have on L2 right now might be close to optimal on the protocol level exactly because it does not enforce anything it does not need to enforce. It just needs to be used well offchain in some way where we have all the data available to accurately calculate costs.

PR #17835, opened 2024-07-22
docs(docs-site): new tab component

### LINKED ISSUE #17180: Open External Links in a New Tab

Description of the feature request here.
Looking through the [Receive token Guide](https://docs-mxctb0to5-taikoxyz.vercel.app/guides/receive-tokens/) and i wanted to navigate to the Bridge which is an external link but it just opens in that same tab instead of openning in a new tab.

You can fix this by using `target="_blank" rel="noopener noreferrer"` this also prevents tabnabbing
The rel attribute sets the relationship between the current page and the linked URL. Setting it to noopener noreferrer is to prevent a type of phishing known as [tabnabbing](https://en.wikipedia.org/wiki/Tabnabbing).

### Spam policy

- [X] I verify that this issue is NOT SPAM and understand SPAM issues will be closed and reported to GitHub, resulting in ACCOUNT TERMINATION.

RogerLamTd:
closes #17180

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,10 @@ Get started with Taiko:
 
 ## Packages
 
+<!-- uncomment after release docs-site
 > [!TIP]
 > Make sure your node is using the latest version tags for taiko-client and taiko-geth. Check out the [node releases page](https://docs.taiko.xyz/network-reference/node-releases)!
+-->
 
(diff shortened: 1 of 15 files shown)


RogerLamTd:
breaks user expectation

PR #17841, opened 2024-07-24
fix(protocol): decide a block's min tier in the next block

dantaik:
This PR addresses a vulnerability identified in previous audits regarding the manipulability of min-tier selection by block proposers. The new approach leaves the min-tier undecided until the next block proposal. This change mitigates direct manipulation by current block proposers and introduces a delay mechanism in min-tier selection. While some manipulation risk remains when consecutive blocks are proposed by the same entity, this solution reduces overall vulnerability compared to the current implementation.

TODO:

- [ ] fix tests
- [ ] modify TierRounter in test to return either ITierProvider or ITierProviderV2


Brechtpd:
Seems problematic to postpone it till the next block because the proposer now does not know that the block it is proposing is profitable?

From a UX perspective, I still think any randomness is problematic because of the spike in fees (and if there is no need for a fee increase the proof should always be required because cheap enough).

dantaik:
> Seems problematic to postpone it till the next block because the proposer now does not know that the block it is proposing is profitable?

It can be an issue, but may be minor.

A similar issue in the current protocol is that blocks are verified in a batch triggered by `proposeBlock` and `proveBlock`, if a proposer is unlucky, he will spend more gas on L1 to verify 16 block (fee spikes). In theory, when base fee and tips are both small, even the biggest block cannot yield enough fees to cover the gas cost on L1. In world where all proposers are perfectly rational, nobody will propose the next block, until more block can be constructed whose fees, combined with all previous block's fees, can cover all the gas cost of all these blocks on L1. One way to solve this is to ask each proposers to deposit some ETH in the protocol and when his block is verified, we simply deduct some Ether from his account, so all verified blocks' proposers share the settlement cost. But these operation will increase the over all gas cost, so this is not a good idea.

I think these are minor issues if we believe the cost of different possible proofs are very small compared to L2 fees to be collected in a block.

dantaik:
Given that the benefit of these changes is unknown, I'm dropping this PR.

PR #17861, opened 2024-07-29
feat(protocol): add SP1 verification support

adaki2004:
This PR is aiming to integrate SP1's on-chain verifier (gateway). 

Functional wise it is identical to this PR/branch: https://github.com/taikoxyz/taiko-mono/pull/17215 the key difference is, that SP1 now is managing the verifier deployment themselves, we just need to call into their contracts. So no need to incorporate the veriying logic into our own contracts.

Veriifer deployments: https://github.com/succinctlabs/sp1-contracts/tree/main/contracts/deployments

If accepted, i'll do a mock `VerifierGateway` contract for unit tests.

### DIFF
--- a/packages/protocol/contracts/verifiers/IVerifier.sol
+++ b/packages/protocol/contracts/verifiers/IVerifier.sol
@@ -28,3 +28,28 @@ interface IVerifier {
     )
         external;
 }
+
+/// @title SP1 Verifier Interface
+/// @author Succinct Labs
+/// @notice This interface is for the deployed SP1 Verifier and 100% brought over from :
+/// https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1Verifier.sol
+interface ISP1Verifier {
+    /// @notice Verifies a proof with given public values and vkey.
+    /// @dev It is expected that the first 4 bytes of proofBytes must match the first 4 bytes of
+    /// target verifier's VERIFIER_HASH.
+    /// @param programVKey The verification key for the RISC-V program.
+    /// @param publicValues The public values encoded as bytes.
+    /// @param proofBytes The proof of the program execution the SP1 zkVM encoded as bytes.
+    function verifyProof(
+        bytes32 programVKey,
+        bytes calldata publicValues,
+        bytes calldata proofBytes
+    )
+        external
+        view;
+}
+
+interface ISP1VerifierWithHash is ISP1Verifier {
+    /// @notice Returns the hash of the verifier.
+    function VERIFIER_HASH() external pure returns (bytes32);
+}
--- a/packages/protocol/contracts/verifiers/IVerifier.sol
+++ b/packages/protocol/contracts/verifiers/IVerifier.sol
@@ -28,28 +28,3 @@ interface IVerifier {
     )
         external;
 }
-
-/// @title SP1 Verifier Interface
-/// @author Succinct Labs
-/// @notice This interface is for the deployed SP1 Verifier and 100% brought over from :
-/// https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1Verifier.sol
-interface ISP1Verifier {
-    /// @notice Verifies a proof with given public values and vkey.
-    /// @dev It is expected that the first 4 bytes of proofBytes must match the first 4 bytes of
-    /// target verifier's VERIFIER_HASH.
-    /// @param programVKey The verification key for the RISC-V program.
-    /// @param publicValues The public values encoded as bytes.
-    /// @param proofBytes The proof of the program execution the SP1 zkVM encoded as bytes.
-    function verifyProof(
-        bytes32 programVKey,
-        bytes calldata publicValues,
-        bytes calldata proofBytes
-    )
-        external
-        view;
-}
-
-interface ISP1VerifierWithHash is ISP1Verifier {
-    /// @notice Returns the hash of the verifier.
-    function VERIFIER_HASH() external pure returns (bytes32);
-}
--- a/packages/protocol/contracts/verifiers/SP1Verifier.sol
+++ b/packages/protocol/contracts/verifiers/SP1Verifier.sol
@@ -88,10 +88,11 @@ contract SP1Verifier is EssentialContract, IVerifier {
 
         uint64 chainId = ITaikoL1(resolve(LibStrings.B_TAIKO, false)).getConfig().chainId;
 
-        bytes memory encodedPublicInput = LibPublicInput.abiEncodePublicInputs(
+        // Need to be converted from bytes32 to bytes
+        bytes32 hashedPublicInput = LibPublicInput.hashPublicInputs(
             _tran, address(this), address(0), _ctx.prover, _ctx.metaHash, chainId
         );
 
-        ISP1Verifier(verifier).verifyProof(programVKey, encodedPublicInput, proof);
+        ISP1Verifier(verifier).verifyProof(programVKey, abi.encode(hashedPublicInput), proof);
     }
 }
--- a/packages/protocol/contracts/verifiers/SP1Verifier.sol
+++ b/packages/protocol/contracts/verifiers/SP1Verifier.sol
@@ -64,7 +64,7 @@ contract SP1Verifier is EssentialContract, IVerifier {
         );
 
         // _proof.data[32:] is the succinct's proof position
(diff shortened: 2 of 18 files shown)


packages/protocol/contracts/verifiers/SP1Verifier.sol
```diff
@@ -0,0 +1,97 @@
...
+
+        if (!isProgramTrusted[programVKey]) {
+            revert SP1_INVALID_PROGRAM_VKEY();
+        }
+
+        uint64 chainId = ITaikoL1(resolve(LibStrings.B_TAIKO, false)).getConfig().chainId;
+
+        bytes memory encodedPublicInput = LibPublicInput.abiEncodePublicInputs(
+            _tran, address(this), address(0), _ctx.prover, _ctx.metaHash, chainId
+        );
```
Brechtpd, line 93: I think currently we actually do `sha256(keccak(public_inputs)`), so this would also just have to call `hashPublicInput`.

adaki2004: And then convert bytes32 to bytes `abi.encode(hashedPublicInput)` ? Because `SPVerifier` expects `bytes`. But then we are where we are now :) So i'm not sure i follow your comment.

adaki2004: Done the "regular" way.

packages/protocol/contracts/verifiers/SP1Verifier.sol
```diff
@@ -0,0 +1,98 @@
...
+        (bytes32 programVKey, bytes memory proof) = abi.decode(_proof.data, (bytes32, bytes));
+
+        if (!isProgramTrusted[programVKey]) {
+            revert SP1_INVALID_PROGRAM_VKEY();
+        }
+
+        uint64 chainId = ITaikoL1(resolve(LibStrings.B_TAIKO, false)).getConfig().chainId;
+
+        // Need to be converted from bytes32 to bytes
+        bytes32 hashedPublicInput = LibPublicInput.hashPublicInputs(
```
CeciliaZ030, line 92: The current `pi_hash` for proof public input is this https://github.com/taikoxyz/raiko/blob/e19ad01116cd39c825ee8ce2e118e877873b6ad1/lib/src/protocol_instance.rs#L166
@adaki2004 does `LibPublicInput.hashPublicInputs` return the same things?

Brechtpd: It should now

CeciliaZ030:
Also if we want to integrate blob verification altogether see https://github.com/taikoxyz/raiko/issues/292#issuecomment-2174901136
Does `TaikoData.TierProof` has the kzg proof needed to call the point precompile?

adaki2004:
> @adaki2004 I think for the testnet, if we are not interacting with [Sp1 contracts](https://github.com/succinctlabs/sp1-contracts/tree/main) directly you have two ways of doing this:
> 
> 1. Directly route to a deployment of [v1.0.5 Sp1 Verifier exported from ZkVM](https://github.com/taikoxyz/raiko/blob/282e4e2411e411e8f6d6899f71d327757f38260d/provers/sp1/contracts/src/PlonkVerifier.sol#L21)
> 2. Deploy their whole [Gateway contract ](https://github.com/succinctlabs/sp1-contracts/blob/9dc34e8eca42a0678727c3c7b2d86c8300946532/contracts/src/SP1VerifierGateway.sol#L13) to our testnet and register [v1.0.5 Sp1 Verifier exported from ZkVM](https://github.com/taikoxyz/raiko/blob/282e4e2411e411e8f6d6899f71d327757f38260d/provers/sp1/contracts/src/PlonkVerifier.sol#L21) to the gateway. On our end, route to the gateway.
> 
> Option 2 is more compatible with their setting on mainnet, option 1 is more convenient.

I think option 1 is more close what setup we will have anyways on mainnet (since we wont be responsible for veritication + deployment anyways if not mistaken) but that’s up to who deploys these contracts on e.g.: holesky (hekla), simply just sets the SP1 router correctly and that is it. (As said they have it on Holesky Sepolia, etc.).

I was referring “tests” in unit test context for coverage.

Brechtpd (approved):

adaki2004:
Will add deployment script + unit test Tuesday (plz dont merge till).

packages/protocol/contracts/verifiers/IVerifier.sol
```diff
@@ -28,3 +28,28 @@ interface IVerifier {
     )
         external;
 }
+
+/// @title SP1 Verifier Interface
+/// @author Succinct Labs
+/// @notice This interface is for the deployed SP1 Verifier and 100% brought over from :
+/// https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1Verifier.sol
+interface ISP1Verifier {
```
dantaik, line 36: put in /thirdparty/?

packages/protocol/contracts/verifiers/SP1Verifier.sol
```diff
@@ -0,0 +1,98 @@
...
+
+/// @title SuccinctVerifier
+/// @custom:security-contact security@taiko.xyz
+contract SP1Verifier is EssentialContract, IVerifier {
+    /// @notice The address of the SP1 verifier contract.
+    /// @dev This can either be a specific SP1Verifier for a specific version, or the
+    ///      SP1VerifierGateway which can be used to verify proofs for any version of SP1.
+    ///      For the list of supported verifiers on each chain, see:
+    ///      https://github.com/succinctlabs/sp1-contracts/tree/main/contracts/deployments
+    address public verifier;
```
dantaik, line 18: use address manager to resolve this address so we can cache it.

packages/protocol/contracts/verifiers/SP1Verifier.sol
```diff
@@ -0,0 +1,98 @@
...
+        TaikoData.TierProof calldata _proof
+    )
+        external
+        view
+    {
+        // Do not run proof verification to contest an existing proof
+        if (_ctx.isContesting) return;
+
+        // Decode will throw if not proper length/encoding
+        (bytes32 programVKey, bytes memory proof) = abi.decode(_proof.data, (bytes32, bytes));
```
dantaik, line 83: I think you can use bytes slicing to avoid copying to memory.

packages/protocol/contracts/verifiers/SP1Verifier.sol
```diff
@@ -0,0 +1,98 @@
...
+
+        if (!isProgramTrusted[programVKey]) {
+            revert SP1_INVALID_PROGRAM_VKEY();
+        }
+
+        uint64 chainId = ITaikoL1(resolve(LibStrings.B_TAIKO, false)).getConfig().chainId;
```
dantaik, line 89: please also define a function so we can override it to return a constant without reading from storage. Please see https://github.com/taikoxyz/taiko-mono/blob/a962b06ee23da3b56a69b13ebb7c76e78dc3f7c0/packages/protocol/contracts/mainnet/MainnetSgxVerifier.sol#L18

dantaik (changes requested):

dantaik (approved):

packages/protocol/contracts/verifiers/SP1Verifier.sol
```diff
@@ -0,0 +1,82 @@
...
+            revert SP1_INVALID_PROGRAM_VKEY();
+        }
+
+        // Need to be converted from bytes32 to bytes
+        bytes32 hashedPublicInput = LibPublicInput.hashPublicInputs(
+            _tran, address(this), address(0), _ctx.prover, _ctx.metaHash, taikoChainId()
+        );
+
+        // _proof.data[32:] is the succinct's proof position
+        (bool success,) = (resolve(LibStrings.B_SP1_REMOTE_VERIFIER, false)).staticcall(
```
dantaik, line 67: @adaki2004 we can also define something like this:

```
function sp1RemoteVerifier() public view virtual returns (address) {
        return resolve(LibStrings.B_SP1_REMOTE_VERIFIER, false);
}
```

It has two benefits:

- we can use this function to view the address on explorer
- we can cache it the same way as `taikoChainId` in MainnetSP1Verifier

adaki2004: Thanks! Done.

dantaik (approved):

PR #17866, opened 2024-07-30
feat(taiko-client): preconfirmation indexing

cyberhorsey:
+ exposing an API that will let exporers query the transaction by hash, following the expected TX format of an existing Blockscout transaction.

Super hacky, very ugly, just something to use for testnets rn.

### DIFF
--- a/packages/taiko-client/cmd/flags/preconfapi.go
+++ b/packages/taiko-client/cmd/flags/preconfapi.go
@@ -29,6 +29,13 @@ var (
 		Value:    "/tmp/badgerdb",
 		EnvVars:  []string{"DB_PATH"},
 	}
+	CORSOrigins = &cli.StringSliceFlag{
+		Name:     "preconfapi.corsORigins",
+		Usage:    "Cors Origins",
+		Category: preconfAPICategory,
+		EnvVars:  []string{"CORS_ORIGINS"},
+		Required: true,
+	}
 )
 
 // PreconfAPIFlags contains all preconfirmations API flags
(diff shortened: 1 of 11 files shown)


packages/taiko-client/cmd/flags/preconfapi.go
```diff
@@ -13,6 +15,27 @@ var (
...
 		EnvVars:  []string{"PRECONFAPI_PORT"},
 	}
+	PollingInterval = &cli.DurationFlag{
+		Name:     "preconfapi.pollingInterval",
+		Usage:    "Interval at which to poll",
+		Category: preconfAPICategory,
+		Value:    1 * time.Second,
+		EnvVars:  []string{"POLLING_INTERVAL"},
+	}
+	DBPath = &cli.StringFlag{
```
YoGhurt111, line 25: Is this change to the DB just for testing or does it need to be merged into the main network?

cyberhorsey: all of this so far is just for testing, this is currently just using. badger DB for an in memory / file based DB to index pre confirmed transactions and expose them via an API for the explorer to query, via polling. If we decide to use this approach in the future for real, it will be refactoref to accept a more complicated and robust database as an option via an interface instead of hard coded badgerdb in the indexer. If it's not needed for mainnet/etc, it will be removed.

YoGhurt111 (approved):

packages/taiko-client/cmd/flags/preconfapi.go
YoGhurt111: ok

PR #17871, opened 2024-08-01
fix(protocol): refine EIP-1559 basefee calculation and gas excess adjustment in TaikoL2

dantaik:
This PR introduces two changes to improve our implementation of the EIP-1559. 

### 1. Refined EIP-1559 Base Fee Calculation

   - Previous: `basefee = eth_qty(excess_gas_issued)`
   - New: `basefee = eth_qty(excess_gas_issued) / (TARGET * ADJUSTMENT_QUOTIENT)`

   This modification aligns our implementation more closely with Vitalik Buterin's proposal. While both approaches are functional, the new calculation allows for a more gradual adjustment of the base fee.

### 2. Gas Excess Adjustment in TaikoL2

   Implemented a gas excess adjustment mechanism  when `TARGET * ADJUSTMENT_QUOTIENT` in a block changes to keep the base fee the same. Without this mechanism, the base fee may spike, as demonstrated in the `test_change_of_quotient_and_gips` test.

### 3. Set 5m gas per second

   Configured the L2 to issue 5 million gas per second (Ethereum issues 2.5 million gas per second which translates into 2.5*12=30 million block gas target).

### DIFF
--- a/packages/protocol/contracts/L2/Lib1559Math.sol
+++ b/packages/protocol/contracts/L2/Lib1559Math.sol
@@ -57,7 +57,7 @@ library Lib1559Math {
             ratio = f * _newGasTarget / _gasTarget;
         }
 
-        if (ratio > type(int256).max) revert EIP1559_INVALID_PARAMS();
+        if (ratio > uint256(type(int256).max)) revert EIP1559_INVALID_PARAMS();
 
         int256 lnRatio = LibFixedPointMath.ln(int256(ratio));
 
--- a/packages/protocol/contracts/L2/Lib1559Math.sol
+++ b/packages/protocol/contracts/L2/Lib1559Math.sol
@@ -15,8 +15,7 @@ library Lib1559Math {
     error EIP1559_INVALID_PARAMS();
 
     function calc1559BaseFee(
-        uint32 _gasTarget,
-        uint8 _adjustmentQuotient,
+        uint256 _gasTarget,
         uint64 _gasExcess,
         uint64 _gasIssuance,
         uint32 _parentGasUsed
@@ -35,7 +34,7 @@ library Lib1559Math {
         // bonding curve, regardless the actual amount of gas used by this
         // block, however, this block's gas used will affect the next
         // block's base fee.
-        basefee_ = basefee(gasExcess_, uint256(_adjustmentQuotient) * _gasTarget);
+        basefee_ = basefee(gasExcess_, _gasTarget);
     }
 
     /// @dev Returns the new gas excess that will keep the basefee the same.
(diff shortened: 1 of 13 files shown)


packages/protocol/contracts/L2/TaikoL2.sol
```diff
@@ -305,8 +323,9 @@ contract TaikoL2 is EssentialContract {
         l2Hashes[parentId] = parentHash;
 
         publicInputHash = newPublicInputHash;
-        gasExcess = newGasExcess;
+        parentGasExcess = newGasExcess;
         parentTimestamp = uint64(block.timestamp);
+        parentGasTarget = newGasExcess;
```
Brechtpd, line 328:
```suggestion
        parentGasTarget = newGasTarget;
```

dantaik:
Update:

@Brechtpd I updated this PR such that when (TARGET * ADJUSTMENT_QUOTIENT) changes, we adjust the value of `parentGasExcess` to maintain the same basefee value instead of setting `parentGasExcess` to zero.

packages/protocol/contracts/L2/Lib1559Math.sol
```diff
@@ -36,41 +36,57 @@ library Lib1559Math {
...
-    /// @param _adjustmentFactor The product of gasTarget and adjustmentQuotient
-    function basefee(
-        uint256 _gasExcess,
-        uint256 _adjustmentFactor
+    /// @dev Returns the new gas excess that will keep the basefee the same.
+    /// `_newGasTarget * ln(_newGasTarget / _target) + _gasExcess * _newGasTarget / _target`
+    function adjustExcess(
+        uint64 _gasExcess,
+        uint64 _gasTarget,
+        uint64 _newGasTarget
```
Brechtpd, line 46: Maybe slightly easier to take as input the current basefee (which we can already calculate)? Then the formula should just be `excess_gas = _newGasTarget * ln(basefee * _newGasTarget)`.

Is the reason for the unchecked code to avoid over/under flow? Because normally this code should be trigger very rarely I think so saving some gas shouldn't be important.

dantaik: The `uncheck` can be removed as the function shall be rarely used.

dantaik: > Maybe slightly easier to take as input the current basefee (which we can already calculate)? Then the formula should just be `excess_gas = _newGasTarget * ln(basefee * _newGasTarget)`.

This will work. However, to calculate the parent block's basefee, we need both the parent's `gasIssuancePerSecond` and `basefeeAdjustmentQuotient` which are not available in the current block. While using the current implementation we only need their product (`parentGasTarget`).

Brechtpd (approved):

davidtaikocha (approved):

PR #17894, opened 2024-08-08
feat(protocol): tee and zk tier router

adaki2004:
This is a draft PR to introduce routing features to specific TEE (SGX or TDX) and ZKVM (SP1, Risc0) verifiers as shown in the diagram below.

kép

### DIFF
--- a/packages/protocol/contracts/verifiers/TeeAndZkVerifier.sol
+++ b/packages/protocol/contracts/verifiers/TeeAndZkVerifier.sol
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+//  _____     _ _         _         _
+// |_   _|_ _(_) |_____  | |   __ _| |__ ___
+//   | |/ _` | | / / _ \ | |__/ _` | '_ (_-<
+//   |_|\__,_|_|_\_\___/ |____\__,_|_.__/__/
+//
+//   Email: security@taiko.xyz
+//   Website: https://taiko.xyz
+//   GitHub: https://github.com/taikoxyz
+//   Discord: https://discord.gg/taikoxyz
+//   Twitter: https://twitter.com/taikoxyz
+//   Blog: https://mirror.xyz/labs.taiko.eth
+//   Youtube: https://www.youtube.com/@taikoxyz
+
+pragma solidity 0.8.24;
+
+import "../common/EssentialContract.sol";
+import "../L1/TaikoData.sol";
+import "../L1/tiers/ITierProvider.sol";
+import "./IVerifier.sol";
+import "@optimism/packages/contracts-bedrock/src/libraries/Bytes.sol";
+
+/// @title TeeAndZkVerifier
+/// @notice See the documentation in {IVerifier}.
+contract TeeAndZkVerifier is EssentialContract, IVerifier {
+    // Some public constants related to specific verifier constraints (e.g.: due to slicing)
+    uint8 public constant SGX_PROOF_SIZE = 89;
+
+    uint256[50] private __gap;
+
+    /// @notice Initializes the contract.
+    /// @param _owner The owner of this contract. msg.sender will be used if this value is zero.
+    /// @param _rollupAddressManager The address of the {AddressManager} contract.
+    function init(address _owner, address _rollupAddressManager) external initializer {
+        __Essential_init(_owner, _rollupAddressManager);
+    }
+
+    /// @inheritdoc IVerifier
+    function verifyProof(
+        Context calldata ctx,
+        TaikoData.Transition calldata tran,
+        TaikoData.TierProof calldata proof
+    )
+        external
+        onlyFromNamed("taiko")
+    {
+        TaikoData.TierProof memory _proof;
+
+        // Based on the tier id - we shall conduct the rooting!
+        // For example in LibTiers, we will have something like this:
+        // SGX + risc0 ZKVM tier ID. -> uint16 public constant TIER_SGX_RISC0 = 301;
+        if (_proof.tier == LibTiers.TIER_SGX_RISC0) {
+            // Verify the SGX part
+            _proof.data = Bytes.slice(proof.data, 0, SGX_PROOF_SIZE);
+            IVerifier(resolve("tier_sgx", false)).verifyProof(ctx, tran, _proof);
+
+            // Verify the risc_zero ZK part
+            _proof.data =
+                Bytes.slice(proof.data, SGX_PROOF_SIZE, (proof.data.length - SGX_PROOF_SIZE));
+            IVerifier(resolve("tier_risc_zero", false)).verifyProof(ctx, tran, _proof);
+        } else if (_proof.tier == LibTiers.TIER_SGX_SP1) {
+            // Verify the SGX part
+            _proof.data = Bytes.slice(proof.data, 0, SGX_PROOF_SIZE);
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/verifiers/TeeAndZkVerifier.sol
```diff
@@ -0,0 +1,72 @@
...
+    )
+        external
+        onlyFromNamed("taiko")
+    {
+        TaikoData.TierProof memory _proof;
+
+        // Based on the tier id - we shall conduct the rooting!
+        // For example in LibTiers, we will have something like this:
+        // SGX + risc0 ZKVM tier ID. -> uint16 public constant TIER_SGX_RISC0 = 301;
+        if (_proof.tier == LibTiers.TIER_SGX_RISC0) {
```
smtmfft, line 52: I think no need to specify RISC0 or SP1 here, we eventually have a tier_zk to support RISC0, SP1 & the other future vendors. so we can call
`IVerifier(resolve("tier_zk")).verifyProof(...)` and it will know which type it should use, i.e, these 301/302 are differentiated inside tier_zk instead of tier_sgx_zk. Meanwhile I think maybe little bit better to have these types in a map instead of pre-def in code, but the diff is not big, i.e., call the contract call update_zk_verifier() (in tier_zk if exists) vs. proxy update the whole tier_zk.

adaki2004: Not sure i understand what you mean here, as you sent me the diagram above.
I thought we will require TEE+ZK with some block proofs, submitted together (as we planned originally some months ago).
From that i assumed that we have a top tier, which acts as a router/distributor. If not, would be cool to have a bit clarity then.

Now if i'm correct we do need:
- A separate router for both ZK (Risc0, SP1) and TEE (not now, but maybe later, when we have TDX, or others)

Question: So we dont require TEE (e.g.g: SGX) + ZK (e.g. SP1) submitted together ? Because if so, then the above diagram (and implementation) is much more gas efficient, then separate router for both.

smtmfft: > Question: So we dont require TEE (e.g.g: SGX) + ZK (e.g. SP1) submitted together ? Because if so, then the above diagram (and implementation) is much more gas efficient, then separate router for both.

Yes, we do, but not SGX+SP1 but TEE+ZK, the SGX is one of TEE, and SP1 is one of ZK. the use case could be like a block in tier tee+zk, and we can submit any two proofs  as long as 1 from TEE group [TDX, SGX, etc] and the other from ZK group [RISC0, SP1, etc]. so far there are just 2 options as you listed here, however, I just think you don't need to explicitly make the combination here, maybe better to outsource to low level (in image) TEE & ZK verifier, and tee/zk verifier knows how to verify any one in TEE/ZK field. Otherwise in future you need to make a big permutation list here.

Brechtpd: I guess the best design indeed depends on how we think we will use it. If we have TDX and SGX proofs, will we require only one? Verifying TEEs is very cheap so simply always requiring all of them feels good to me. For zk some router contract of a way for a verifier to return what type it is would be nice.

adaki2004: > Yes, we do, but not SGX+SP1 but TEE+ZK, the SGX is one of TEE, and SP1 is one of ZK. the use case could be like a block in tier tee+zk, and we can submit any two proofs as long as 1 from TEE group [TDX, SGX, etc] and the other from ZK group [RISC0, SP1, etc]. so far there are just 2 options as you listed here, however, I just think you don't need to explicitly make the combination here, maybe better to outsource to low level (in image) TEE & ZK verifier, and tee/zk verifier knows how to verify any one in TEE/ZK field. Otherwise in future you need to make a big permutation list here.

I got it now, but i still think the best way is to have 1 router contract (this one) doing the distribution. Let me explain why.
So the input of the `verifyProof()` is `TaikoData.TierProof`. The struct looks like this:
```
struct TierProof {
        uint16 tier;
        bytes data;
    }
```

In this case, where we require SGX+SP1 or as if Brecht says, let's say all TEE proofs required to be together, so e.g.: TDX+SGX+SP1, we encode / concat those proof bytes to be one big chunk of bytes.

So let's say SGX is 89 bytes, TDX also 89 bytes and SP1 requires 500 bytes, this would mean:
proof_bytes[0..88] = SGX proof 
proof_bytes[89..176] = TDX proof
proof_bytes[177..676] = SP1 proof

**If we dont care about slicing and arranging the data** in this top level router contract and we proceed the way you mentions, we can do the slicing logic in each verifier (TEE and ZK) but that would be:
1. Highly gas inefficient. We need at least +2 extra external callse (2x external call + 2x data submission)
1. complex because we the same data-slicing in ZK and TEE verifier

**If we do arrange tha data*  (meaning we slice the byte stream into chunk as the above contract already does) in those combinations in this contract, we already have the TEE+ZK "logic" contract implemented here, so no neeed an extra level abstraction (TEE verifier and ZK verifier), just directly call the each individual verifiers.

Realistically also, i dont know if we have in the near future 5x ZK proof providers, but if the TEEs (SGX and TDX) will be required together, then it is not a big of a combination even with 10 ZK proofs, beause it is just 10 if-else structure.

@Brechtpd @smtmfft let me know how you wanna proceed, because it seems it still not decided, but i'd vouch for this type (and no extra level abstraction and complexity) - if it is fixed, i can help more.

smtmfft: For me, I guess at least in near a few months there are no more new types. perhaps that hierarchy design is of a little bit over @Brechtpd ?

smtmfft: But here another problem is if like SGX_SP1 is 302, SGX_RISC0 is 301, the here first 302 has higher tier weight so the 301 type is challengeable by 302, but by design all ZK are equal. Maybe we need to make 301 & 302 all with a tier weight equal to 300.

adaki2004: > But here another problem is if like SGX_SP1 is 302, SGX_RISC0 is 301, the here first 302 has higher tier weight so the 301 type is challengeable by 302, but by design all ZK are equal. Maybe we need to make 301 & 302 all with a tier weight equal to 300.

True true. It is just a draft PoC from me. 
But to overcome the issue, we can for example encode the combination on the first 1byte (0..255) in the `_proof.data` section for example. 
(0: allTEE + R0)
(1: allTEE+ SP1)
etc.

I can adjust these if we agree on a final design. So would be cool to get some kind of an agreement if we need other, underlying router contracts (for TEE + ZK)or not. Personally i'd vouch for a no, bc. of the aforementioned reasons.

Brechtpd: I kind of also lean to not do underlying router contracts for now because you could also implement things differently

smtmfft: OK, so if we can make all ZKs equal, then plain list is good enough for me. otherwise, the outside prover need to prepare both risc0 & sp1 services as they can't know the tier-selected is which one.

PR #17896, opened 2024-08-09
fix(protocol): fix bug in `adjustExcess` and improve L2 basefee calculation

dantaik:
BEGIN_COMMIT_OVERRIDE 
fix(protocol): fix bug in adjustExcess
feat(protocol): improve L2 basefee calculation 
END_COMMIT_OVERRIDE


YoGhurt111 (approved):

YoGhurt111 (approved):

YoGhurt111 (approved):

YoGhurt111 (approved):

davidtaikocha (approved):

adaki2004 (approved):

packages/protocol/contracts/L2/Lib1559Math.sol
```diff
@@ -61,8 +61,15 @@ library Lib1559Math {
         int256 lnRatio = FixedPointMathLib.lnWad(int256(ratio)); // may be negative
 
         uint256 newGasExcess;
+
         assembly {
-            newGasExcess := sdiv(add(mul(lnRatio, _newGasTarget), mul(ratio, _gasExcess)), f)
+            // x = (_newGasTarget * lnRatio + _gasExcess * ratio)
+            let x := add(mul(_newGasTarget, lnRatio), mul(_gasExcess, ratio))
+
+            // If x < 0, set newGasExcess to 0, otherwise calculate newGasExcess = x / f
```
Brechtpd, line 69: If it can still be 0, maybe easier to just accept and just always set it back to 0?

dantaik: Not sure what you meant, but x (the new gas excess) can be zero. Here we need to set it to 0 if it is negative before we perform `x/f`.

packages/nfts/foundry.toml
```diff
@@ -48,7 +48,7 @@ libs = ["node_modules", "lib"]
 runs = 200
 
 [fmt]
-bracket_spacing = true
+bracket_spacing = false
```
dantaik, line 51:
```suggestion
bracket_spacing = true
```

packages/protocol/foundry.toml
```diff
@@ -48,7 +48,7 @@ ignored_error_codes = [2394, 3860, 5574, 5740]
 runs = 200
 
 [fmt]
-bracket_spacing = true
+bracket_spacing = false
```
dantaik, line 51:
```suggestion
bracket_spacing = true
```

packages/protocol/contracts/L2/Lib1559Math.sol
Brechtpd: Oh I just mean that the math to calculate the `newGasExcess` is quite complex, and does not seem to guarantee that we can calculate the correct value to maintain the current basefee in all cases, so if it's still sometimes reset to 0 then I guess okay to always do it to make the code simpler in general?

YoGhurt111:
@dantaik Every time https://github.com/taikoxyz/taiko-mono/pull/17896/commits/a29d17c269dceb667fe297f6fbe85b914161541b formats, it seems https://github.com/taikoxyz/taiko-mono/pull/17896/commits/146854a98a0d5b46d9466994b8a44cab29b74ab8 change it back.

packages/protocol/contracts/L2/Lib1559Math.sol
dantaik: I see. If the math is even harder to implement and test, I would agree to simply set the gasExcess to 0, but since the math has ben actually figured out, I think we should still maintain base fee unchanged (or slightly changed) when configurations changed.

Brechtpd (approved):

YoGhurt111 (approved):

dantaik:
> @dantaik Every time [a29d17c](https://github.com/taikoxyz/taiko-mono/commit/a29d17c269dceb667fe297f6fbe85b914161541b) formats, it seems [146854a](https://github.com/taikoxyz/taiko-mono/commit/146854a98a0d5b46d9466994b8a44cab29b74ab8) change it back.

sometimes you have to run 'foundryup' in your terminal. the github workflow always run this command first.

packages/protocol/contracts/L2/Lib1559Math.sol
```diff
@@ -61,8 +61,15 @@ library Lib1559Math {
         int256 lnRatio = FixedPointMathLib.lnWad(int256(ratio)); // may be negative
 
         uint256 newGasExcess;
+
         assembly {
-            newGasExcess := sdiv(add(mul(lnRatio, _newGasTarget), mul(ratio, _gasExcess)), f)
+            // x = (_newGasTarget * lnRatio + _gasExcess * ratio)
```
dantaik, line 66:
```suggestion
            // compute x = (_newGasTarget * lnRatio + _gasExcess * ratio)
```

davidtaikocha (approved):

PR #17897, opened 2024-08-09
chore(taiko-client): add hive tests to workflow

mask-pp:
run `make hive_test`
```
sh integration_test/hive_test.sh
```

equal to

```
git clone git@github.com:taikoxyz/hive.git
cd hive && make hive
./build/bin/hive --client taiko2/geth,taiko2/prysm-bn,taiko2/prysm-vc --sim taiko2 --sim.limit "eth2-deneb-testnet/test-deneb-genesis"
```


.github/workflows/taiko-client--docker.yml
```diff
@@ -51,3 +51,7 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+      - name: hive test
```
RogerLamTd, line 55: could we make this change in a separate PR and mark it with feat(repo): ?

mask-pp: When a new Docker image is generated, the Hive test needs to be triggered. I haven't found an effective way to achieve this yet. Do you have any better suggestions?

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Push docker image to GAR"
 
 on:
   push:
-    branches: [main]
+    branches: [hive_tests_]
```
mask-pp, line 5: Temporarily change it in order to trigger the hive test.

.github/workflows/taiko-client--docker.yml
RogerLamTd: We can keep this solution, I just want to migrate the change to a separate PR for documentation clarity!

mask-pp: `taiko-client--hive_test.yml` is added.

packages/taiko-client/integration_test/hive_tests.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func TestHiveHandler(t *testing.T) {
+	baseDir := os.Getenv("HIVE_BASE_DIR")
+	if baseDir == "" {
+		t.SkipNow()
+	}
+	handler, err := hivesim.NewHiveFramework(&hivesim.HiveConfig{
+		BaseDir:        baseDir,
+		SimPattern:     "taiko2",
```
davidtaikocha, line 19: Let's use `taiko` instead

mask-pp: OK, will be fixed in the hive repo and then change here.

davidtaikocha (approved):

RogerLamTd (approved):

PR #17915, opened 2024-08-14
feat(taiko-client): introduce `TierZkVMRisc0ID`

YoGhurt111:


mask-pp (dismissed):

davidtaikocha (dismissed):

packages/taiko-client/prover/init.go
```diff
@@ -114,7 +114,8 @@ func (p *Prover) initProofSubmitters(
 				Dummy:               p.cfg.Dummy,
 				RaikoRequestTimeout: p.cfg.RaikoRequestTimeout,
 			}
-		case encoding.TierSgxAndZkVMID:
+		case encoding.TierZkVMRisc0ID:
+			// TODO: Temporarily use SgxAndZKvmProofProducer to generate zkvm proofs
```
davidtaikocha, line 118: So we will introduce the `ZKvmProofProducer` in another PR?

YoGhurt111: No, after we finish the sgx + zkvm verifier contract, we will remove `TierZkVMRisc0ID` and add sgx logic in `SgxAndZKvmProofProducer`.
In my understanding, we don't need pure zk proof.
If it is wrong, plz let me know.

davidtaikocha: @smtmfft , could u confirm that? or we will keep having pure zkvm tier in hekla?

smtmfft: eventually we will have sgx+zk, but I am not sure how long the pure zk will last

davidtaikocha: Then lets create a separated `ZKvmProofProducer `? otherwise this part of code even can't be used for mainnet? Since in mainnet, we only have zk + sgx?

YoGhurt111: OK

davidtaikocha (approved):

mask-pp (approved):

PR #17917, opened 2024-08-15
feat(protocol): introduce `HeklaTierProvider`

davidtaikocha:


packages/protocol/contracts/hekla/HeklaTierProvider.sol
```diff
@@ -0,0 +1,30 @@
...
+        tiers_[0] = LibTiers.TIER_OPTIMISTIC;
+        tiers_[1] = LibTiers.TIER_SGX;
+        tiers_[2] = LibTiers.TIER_ZKVM_RISC0;
+        tiers_[3] = LibTiers.TIER_GUARDIAN_MINORITY;
+        tiers_[4] = LibTiers.TIER_GUARDIAN;
+    }
+
+    /// @inheritdoc ITierProvider
+    function getMinTier(uint256 _rand) public pure override returns (uint16) {
+        if (_rand % 1000 == 0) {
```
YoGhurt111, line 21:
```suggestion
        if (_rand % 1000 == 1) {
```

YoGhurt111: Avoid occupancy of the probability of sgx.

davidtaikocha: Since SGX is in `else if`, its ok to put `0` here?

YoGhurt111 (approved):

PR #17938, opened 2024-08-17
feat(protocol): proof verification aggregation

Brechtpd:
Proof aggregation minimizes the onchain costs for proof verification, which is very important for zk proof verfication. Without aggregation the onchain costs would be very expensive, with proof aggregation we mostly only pay the offchain proof generation cost which is already pretty cheap.

Currently this PR just outlines one easy way to get proof aggregation working with minimal changes.

raiko PR: https://github.com/taikoxyz/raiko/pull/347

### DIFF
--- a/packages/protocol/contracts/L1/TaikoL1.sol
+++ b/packages/protocol/contracts/L1/TaikoL1.sol
@@ -149,7 +149,7 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents {
     function proveBlocks(
         uint64[] calldata _blockIds,
         bytes[] calldata _inputArr,
-        bytes proof
+        bytes calldata _proof
     )
         external
         whenNotPaused
@@ -163,14 +163,25 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents {
 
         TaikoData.Config memory config = getConfig();
 
+        // Run over all blocks and collect all blocks that need to be verified
         IVerifier.Context[] memory ctxs = new IVerifier.Context[](_blockIds.length);
-        TaikoData.Transition[] memory trans = new TaikoData.Transition[](_blockIds.length);
-
         for (uint256 i; i < _blockIds.length; ++i) {
-            (ctxs[i], trans[i]) = _proveBlock(_blockIds[i], _inputArr[i], config);
+            ctxs[i] = _proveBlock(_blockIds[i], _inputArr[i], config);
+            // We have to make sure that we only batch verify the same type of proof
+            if (
+                ctxs[i].verifier != ctxs[0].verifier || ctxs[i].isContesting != ctxs[0].isContesting
+            ) {
+                revert L1_INVALID_PARAMS();
+            }
         }
 
-        IVerifier(verifier).verifyProof(ctxs, trans, proof);
+        // Batch verify the blocks
+        // Do not run proof verification to contest an existing proof
+        if (!ctxs[0].isContesting) {
+            IVerifier(ctxs[0].verifier).verifyProofs(
+                ctxs, abi.decode(_proof, (TaikoData.TierProof))
+            );
+        }
     }
 
     /// @inheritdoc ITaikoL1
@@ -348,8 +359,9 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents {
         TaikoData.Config memory _config
     )
         internal
+        returns (IVerifier.Context memory ctx_)
     {
-        LibProving.proveBlock(state, _config, this, _blockId, _input);
+        ctx_ = LibProving.proveBlock(state, _config, this, _blockId, _input);
 
         if (LibUtils.shouldVerifyBlocks(_config, _blockId, false)) {
             LibVerifying.verifyBlocks(state, _config, this, _config.maxBlocksToVerify);
--- a/packages/protocol/contracts/L1/TaikoL1.sol
+++ b/packages/protocol/contracts/L1/TaikoL1.sol
@@ -148,7 +148,8 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents {
     /// @inheritdoc ITaikoL1
     function proveBlocks(
         uint64[] calldata _blockIds,
-        bytes[] calldata _inputArr
+        bytes[] calldata _inputArr,
+        bytes proof
     )
         external
         whenNotPaused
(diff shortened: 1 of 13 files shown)


packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -148,7 +148,8 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents {
     /// @inheritdoc ITaikoL1
     function proveBlocks(
         uint64[] calldata _blockIds,
-        bytes[] calldata _inputArr
+        bytes[] calldata _inputArr,
+        bytes proof
```
dantaik, line 152: if `proof` is "", we can still call the verifyProof inside LibProving.sol so we support the current non-aggregation mode?

packages/protocol/contracts/verifiers/SgxVerifier.sol
```diff
@@ -154,19 +155,35 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
         if (_ctx.isContesting) return;
 
+        if (_ctx.length != _tran.length) {
+            revert SGX_INVALID_INPUT();
+        }
+
         // Size is: 89 bytes
-        // 4 bytes + 20 bytes + 65 bytes (signature) = 89
-        if (_proof.data.length != 89) revert SGX_INVALID_PROOF();
+        // 4 bytes + 20 bytes + 20 bytes + 65 bytes (signature) = 109
```
dantaik, line 163: Since the current SGX prover doesn't support multiple blocks, can we ensure the length of `_ctx` and `_tran` are both 1 for now?

packages/protocol/contracts/L1/libs/LibProving.sol
```diff
@@ -136,6 +136,7 @@ library LibProving {
         bytes calldata _input
     )
         public
+        returns (IVerifier.Context memory ctx, TaikoData.Transition memory tran)
```
dantaik, line 139: trailing "_" for return parameters please

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -162,9 +163,14 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents {
 
         TaikoData.Config memory config = getConfig();
 
+        IVerifier.Context[] memory ctxs = new IVerifier.Context[](_blockIds.length);
```
dantaik, line 166: Does it make sense to make Transition part of ctx struct?

packages/protocol/contracts/L1/TaikoL1.sol
```diff
@@ -162,9 +163,14 @@ contract TaikoL1 is EssentialContract, ITaikoL1, TaikoEvents {
...
+        TaikoData.Transition[] memory trans = new TaikoData.Transition[](_blockIds.length);
+
         for (uint256 i; i < _blockIds.length; ++i) {
-            _proveBlock(_blockIds[i], _inputArr[i], config);
+            (ctxs[i], trans[i]) = _proveBlock(_blockIds[i], _inputArr[i], config);
         }
+
+        IVerifier(verifier).verifyProof(ctxs, trans, proof);
```
adaki2004, line 173: I have a compilation error, due to this outter `verifyProof()`, the `verifier` variable is not know. It is coming from the tier configuration itself (now, in our current design).

And also this yields into another problem, with an example:
The block tiers (and hence the tier verifiers) can be for example:
block nr. 102: SGX only
block nr. 103: SGX+RISC0
block nr. 104: SGX+SP1
block nr. 105: SGX only again

And thus, the verifier contract address itself differs from block to block. so aggregating multiple, continuous blocks will only work (in the current design at least!) if we have the exact same logic (contract) to verify (so same tiers), otherwise no.

0xjunger:
Is this aggregation between zk proofs(r0, sp1) inside Raiko? Or it’s mean compatibility with proof aggregation layers?

packages/protocol/contracts/L1/TaikoL1.sol
dantaik: I think you can only aggregate proofs for the same tier together, but the proofs may correspond to blocks that are not consecutive.

adaki2004: Yes, but then with ever 500-100th zk tier blocks in practice it will be useful for tee only. (At least for now.)

packages/protocol/contracts/verifiers/SgxVerifier.sol
Brechtpd: The inputs into the verification are still changed so a bit tricky. If length is 1 we could use the old verification path, but then this would be a special case and proof aggregation on a single proof would not be valid. Seems easier if aggregation is always used even if there's only a single proof, otherwise we need multiple verification paths.

packages/protocol/contracts/L1/TaikoL1.sol
Brechtpd: I think I have now added the required checks to make sure only the same proofs can be aggregrated.

Brechtpd:
> Is this aggregation between zk proofs(r0, sp1) inside Raiko? Or it’s mean compatibility with proof aggregation layers?

It is aggregation of the same proof types within raiko (so a batch of SP1 proofs OR a batch of SGX proofs OR ...)

Brechtpd:
Ok I will leave it here 🫡

packages/protocol/contracts/verifiers/RiscZeroVerifier.sol
```diff
@@ -47,39 +47,48 @@ contract RiscZeroVerifier is EssentialContract, IVerifier {
...
+        for (uint256 i = 0; i < _ctx.length; i++) {
+            public_inputs[i + 1] = LibPublicInput.hashPublicInputs(
+                _ctx[i].tran,
+                address(this),
+                address(0),
+                _ctx[i].prover,
+                _ctx[i].metaHash,
+                taikoChainId()
+            );
+            emit ProofVerified(_ctx[i].metaHash, public_inputs[i + 1]);
```
smtmfft, line 78: not yet verified here??

packages/protocol/contracts/verifiers/SgxVerifier.sol
```diff
@@ -143,30 +143,43 @@ contract SgxVerifier is EssentialContract, IVerifier {
...
-            _proof.data[24:]
-        );
+        address oldInstance = address(bytes20(_proof.data[4:24]));
+        address newInstance = address(bytes20(_proof.data[24:44]));
+        bytes memory signature = _proof.data[44:];
+
+        // Collect public inputs
+        bytes32[] memory public_inputs = new bytes32[](_ctx.length + 1);
+        // First public input is the current instance public key
+        public_inputs[0] = bytes32(bytes20(oldInstance));
```
smtmfft, line 165: Is this necessary? I mean as oldInstance is always recovered from signature, why we need explicit list oldInstance here?

packages/protocol/contracts/verifiers/RiscZeroVerifier.sol
adaki2004: Technically true only but:
- this is the only way to emit events per transition (one by one) since we are going thru the aggregated proofs.
- if something fails, the TXN reverted anyways

Brechtpd: Assumption indeed that all the proofs are verified correctly in some way afterwards.

packages/protocol/contracts/verifiers/SgxVerifier.sol
Brechtpd: We need to have it available in the aggregation proof, because that's the address we need to verify against for the first block proof. We need to get it from somewhere to be able to correctly verify that proof, because we don't have it available anymore in the SGX instance itself (could be many blocks ago, and we only store the latest key pair in SGX).

PR #17951, opened 2024-08-19
refactor

dantaik:


packages/protocol/contracts/verifiers/compose/ZkVMVerifier.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.24;
+
+import "./ComposeVerifier.sol";
+
+/// @title ZkVMVerifier
+/// @notice This contract is a verifier for the Mainnet ZkVM that composes RiscZero and SP1
+/// Verifiers.
+/// @custom:security-contact security@taiko.xyz
+contract ZkVMVerifier is ComposeVerifier {
```
smtmfft, line 10: Just think maybe also impl TEEVerifier in this PR since the framework code is exactly the same.

smtmfft (approved):

packages/protocol/contracts/verifiers/compose/ComposeVerifier.sol
```diff
@@ -0,0 +1,97 @@
...
+        Context calldata _ctx,
+        TaikoData.Transition calldata _tran,
+        TaikoData.TierProof calldata _proof
+    )
+        external
+    {
+        (address[] memory verifiers, uint256 threshold) = getSubVerifiersAndThreshold();
+
+        for (uint256 i; i < verifiers.length; ++i) {
+            // Store the value 1 in the temporary storage slot using inline assembly
```
adaki2004, line 44: Should not we have a safeguard to not possibly reuse verifiers ? 🤔
 
Something like:
```
address prevVerifier;
for (uint256 i; i < verifiers.length; ++i) {
    if (uint160(prevVerifier) >= uint160(verifiers[i])) {
                    revert L1_INVALID_VERIFIER();
    }
    ...
    prevVerifier = verifiers[i];
}
```

dantaik: The verifier is coded by us and is not dynamically configured by proposers/provers.

PR #17960, opened 2024-08-20
refactor(protocol): refactor tier provider & address caching and add ComposeVerifier

dantaik:
BEGIN_COMMIT_OVERRIDE
feat(protocol): add ComposeVerifier, TeeAnyVerifier, and ZkAnyVerifier
refactor(protocol): refactor TierProvider implementations
refactor(protocol): simplify mainnet address caching
END_COMMIT_OVERRIDE


davidtaikocha (approved):

packages/protocol/contracts/L1/tiers/TierProviderBase.sol
```diff
@@ -48,9 +48,20 @@ abstract contract TierProviderBase is ITierProvider {
             });
         }
 
-        if (_tierId == LibTiers.TIER_SGX2) {
+        if (_tierId == LibTiers.TIER_TDX) {
```
adaki2004, line 51: It seems (since `TIER_TDX == 200` as well ass `TIER_SGX` and `TIER_TEE_ANY`) that one is enough, also their fields are the same. 
Is there any reason having all defined ?

packages/protocol/contracts/verifiers/compose/ComposeVerifier.sol
```diff
@@ -0,0 +1,88 @@
...
+                    verifierFound = true;
+                    verifiers[j] = address(0);
+                }
+            }
+
+            if (!verifierFound) revert INVALID_VERIFIER();
+
+            (bool success, bytes memory returnData) = subproofs[i].verifier.call(
+                abi.encodeCall(
+                    IVerifier.verifyProof,
```
adaki2004, line 62: Calling `SgxVerifier`'s `verifyProof()` - if we will start using the TEE any, we need to modify the code because SgxVerifier.sol line 146, it has a guard:
`onlyFromNamed(LibStrings.B_TAIKO)`.

Maybe best to mention it as a todo.

packages/protocol/contracts/verifiers/compose/TeeAnyVerifier.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.24;
+
+import "../../common/LibStrings.sol";
+import "./ComposeVerifier.sol";
+
+/// @title TeeAnyVerifier
+/// @notice This contract is a verifier for the Mainnet ZkVM that composes RiscZero and SP1
```
adaki2004, line 8:
```suggestion
/// @notice This contract is a verifier for the Mainnet TEE proofs that composes SGX and TDX
```

packages/protocol/contracts/L1/tiers/TierProviderBase.sol
dantaik: With different names but same IDs, in TierProviderBase, we can configure all of them in the one single file so they have different verifiers.

packages/protocol/contracts/verifiers/compose/ComposeVerifier.sol
smtmfft: It was initially added by Dani for security concern I guess @adaki2004? And yes, I think it can be removed now.

adaki2004: Im not sure if we can / shall remove it.

From mobile ATM but there was an issue allowing calling the contract by anybody.

I think it was the replay issue. So if caller is not the TaikoL1 and the call succeeds, it already exchanges the keys, so that the call from TaikoL1 womt suceed, this is why we need here. (If not mistaken).

smtmfft: yes, but why a prover does that? it seems no harm for the others as he can not update others key, only negative impact to himself?

dantaik: I read the code again, it seems we can indeed remove the check. Without the check, basically if someone has access to an instance's private-key, he can then remove the instance (and potentially add a new one). But this will not affect others as this instance will never be used by others.

adaki2004: > yes, but why a prover does that? it seems no harm for the others as he can not update others key, only negative impact to himself?

No, (afaik) it is not about that.

Imagine Bob is an honest (sgx) prover. Alice is a malicious actor who wants to trick Bob into losing money.
Alice monitors the mempool and when she sees the proving transaction, she extracts the proof and submit directly to the SgxVerifier contract with higher gas price.
Bob’s proof then would be revertes and possibly would loose money (bond). This guard is to prevent that situation.
(Basically it would work all the time and can increase proving time so that causes harm to alk the SGX prover to pay up to “infonity” just to prove a block and get back a bond).

it was also flagged  by SigmaPrime audit.

packages/protocol/contracts/verifiers/SgxVerifier.sol
```diff
@@ -149,7 +149,6 @@ contract SgxVerifier is EssentialContract, IVerifier {
         TaikoData.TierProof calldata _proof
     )
         external
-        onlyFromNamed(LibStrings.B_TAIKO)
     {
```
adaki2004, line 152: I still think it is a security issue as referred to here. https://github.com/taikoxyz/taiko-mono/pull/17960#discussion_r1728417427

Please see SigmaPrime report. 
kép

packages/protocol/contracts/verifiers/compose/ComposeVerifier.sol
adaki2004: > I read the code again, it seems we can indeed remove the check. Without the check, basically if someone has access to an instance's private-key, he can then remove the instance (and potentially add a new one). But this will not affect others as this instance will never be used by others.

I'm referring to this audit finding.
kép

dantaik: Can Alice frontruns Bob it means Bob shared his private key with Alice so they are considered the same party, right? Otherwise, how can Alice create a signature for Bob's instance?


```
      address oldInstance = ECDSA.recover(
            LibPublicInput.hashPublicInputs(
                _tran, address(this), newInstance, _ctx.prover, _ctx.metaHash, taikoChainId()
            ),
            _proof.data[24:]
        );
```

adaki2004: > Can Alice frontruns Bob it means Bob shared his private key with Alice so they are considered the same party, right? Otherwise, how can Alice create a signature for Bob's instance?
> 
> ```
>       address oldInstance = ECDSA.recover(
>             LibPublicInput.hashPublicInputs(
>                 _tran, address(this), newInstance, _ctx.prover, _ctx.metaHash, taikoChainId()
>             ),
>             _proof.data[24:]
>         );
> ```
Maybe i'm wrong or i dont see if we changed something in between, but here is my idea:

1. Bob has to re-create this "payload" off-chain, right ? (to sign it)
```
LibPublicInput.hashPublicInputs(
                _tran, address(this), newInstance, _ctx.prover, _ctx.metaHash, taikoChainId()
            ),
```
            
2. Alice cannot have access to the private key of Bob BUT she **sees** Bob's transaction **in the mempool**. She can extract this information then submit the smart contract call with higher gas tip. (Without that guard).

3. So Alice take advantage of the fact she can replace Bob's with the new instance, so that Bob cannot verify the transition with the old anymore.

Imo this creates the security vulnerability. (Please let me know if it is not or we mitigated somehow.. But i still think we need the `onlyFromNamed` ).

Brechtpd: Because of the potential key switching it is indeed a problem and should be guarded.

dantaik: > 2\. Alice cannot have access to the private key of Bob BUT she **sees** Bob's transaction **in the mempool**. She can extract this information then submit the smart contract call directly to SgxVerifier, with higher gas tip. (Without that guard).

I see.

dantaik:
Based on the comments above:

- I deleted `ComposeVerifier.sol`.
- I changed the `verifyProof` function to **view** in `Risc0Verifier.sol` @smtmfft please verify.

packages/protocol/contracts/verifiers/Risc0Verifier.sol
```diff
@@ -53,6 +50,7 @@ contract RiscZeroVerifier is EssentialContract, IVerifier {
         TaikoData.TierProof calldata _proof
     )
         external
+        view
```
dantaik, line 53: @smtmfft please take a look at this.

adaki2004 (commented):
Honestly, i liked the `ComposeVerifier` and the other derived contracts, which acted like routes. 
Beside the fact we had that guard issue with `SGX` (i think easily solveable by just guarding all the calls from: `TaikoL1 -> ComposeVerifier -> Verifier` especially in SGX case), i think it was a good way to achieve what Yue wanted with that distribution diagram.
But let me copy the diagram again here, so we have that:
kép

Also this draft PR trying to achieve that too - not that dynamic like the `ComposeVerifier` tho -  but more lightweight and gas efficient, if we wish to proceed further with that, leaving it here: https://github.com/taikoxyz/taiko-mono/pull/17894

dantaik:
> Also this draft PR trying to achieve that too - not that dynamic like the `ComposeVerifier` tho - but more lightweight and gas efficient, if we wish to proceed further with that, leaving it here: #17894

@adaki2004  added back those contracts, and made some changes to the permission check in ComposeVerifier and SgxVerifier. Please take another look.

packages/protocol/contracts/verifiers/compose/ComposeVerifier.sol
```diff
@@ -0,0 +1,101 @@
...
+    /// @param _tran The transition to verify.
+    /// @param _proof The proof to verify.
+    function verifyProof(
+        Context calldata _ctx,
+        TaikoData.Transition calldata _tran,
+        TaikoData.TierProof calldata _proof
+    )
+        external
+    {
+        if (!isCallerAuthorized(msg.sender)) revert INVALID_CALLER();
```
adaki2004, line 46: Hmm, why not `onlyFromNamed(LibStrings.B_TAIKO)` ?

adaki2004 (approved):

packages/protocol/contracts/verifiers/compose/ComposeVerifier.sol
```diff
@@ -0,0 +1,101 @@
...
+                    verifierFound = true;
+                    verifiers[j] = address(0);
+                }
+            }
+
+            if (!verifierFound) revert INVALID_VERIFIER();
+
+            (bool success, bytes memory returnData) = subproofs[i].verifier.call(
+                abi.encodeCall(
+                    IVerifier.verifyProof,
```
Brechtpd, line 68: what was the reason again to not just call verifyProof directly on the address?

Brechtpd (approved):

packages/protocol/contracts/verifiers/compose/ComposeVerifier.sol
dantaik: Using the low level call may be an overkill here. I refactored the code to require the number of sub-proofs much match the threshold and each sub-proof must be verifiable.

packages/protocol/contracts/verifiers/compose/ComposeVerifier.sol
dantaik: isCallerAuthorized is to make testing easier.

Brechtpd (approved):

adaki2004 (approved):

packages/protocol/contracts/L1/tiers/ITierProvider.sol
```diff
@@ -24,6 +24,7 @@ interface ITierProvider {
     /// @dev Retrieves the IDs of all supported tiers.
     /// Note that the core protocol requires the number of tiers to be smaller
     /// than 256. In reality, this number should be much smaller.
+    /// Additionally, each tier's ID value must be unique.
```
dantaik, line 27:
```suggestion
```

davidtaikocha (approved):

YoGhurt111 (approved):

smtmfft (approved):

YoGhurt111 (approved):

PR #17962, opened 2024-08-21
feat(taiko-client): add optional `l1.private`

YoGhurt111:


packages/taiko-client/cmd/flags/common.go
```diff
@@ -28,6 +28,12 @@ var (
 		Category: commonCategory,
 		EnvVars:  []string{"L1_WS"},
 	}
+	L1PrivateEndpoint = &cli.StringFlag{
+		Name:     "l1.private",
```
dantaik, line 32: private_tx_endpoint

mask-pp (approved):

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Push docker image to GAR"
 
 on:
   push:
-    branches: [main]
+    branches: [feature/private-tx-endpoint]
```
davidtaikocha, line 5: Lets change it back?

packages/taiko-client/cmd/flags/common.go
```diff
@@ -28,6 +28,12 @@ var (
 		Category: commonCategory,
 		EnvVars:  []string{"L1_WS"},
 	}
+	L1PrivateTxEndpoint = &cli.StringFlag{
+		Name:     "l1.privateTx",
```
davidtaikocha, line 32:
```suggestion
		Name:     "l1.privateTxWs",
```

packages/taiko-client/cmd/flags/common.go
```diff
@@ -28,6 +28,12 @@ var (
...
+		Usage:    "Websocket RPC endpoint of a L1 private ethereum node",
+		Category: commonCategory,
+		EnvVars:  []string{"L1_PRIVATE_TX"},
```
davidtaikocha, line 35:
```suggestion
		EnvVars:  []string{"L1_PRIVATE_TX_WS"},
```

packages/taiko-client/proposer/config.go
```diff
@@ -105,5 +106,10 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
 			l1ProposerPrivKey,
 			c,
 		),
+		PrivateTxmgrConfigs: pkgFlags.InitTxmgrConfigsFromCli(
+			c.String(flags.L1PrivateTxEndpoint.Name),
```
davidtaikocha, line 110: This value might be an empty string, is that ok?

.github/workflows/taiko-client--docker.yml
YoGhurt111: Will change it back when all test are finished in hekla.

packages/taiko-client/cmd/flags/common.go
YoGhurt111: Both http and ws are supposed to be supported.

packages/taiko-client/cmd/flags/common.go
YoGhurt111: Both http and ws are supposed to be supported.

packages/taiko-client/cmd/flags/common.go
davidtaikocha: But in flag description it says websocket endpoint?

packages/taiko-client/proposer/config.go
YoGhurt111: It is ok . 
Added `if cfg.PrivateTxmgrConfigs != nil && len(cfg.PrivateTxmgrConfigs.L1RPCURL) > 0` when it was used

packages/taiko-client/cmd/flags/common.go
YoGhurt111: Let me fix the description!

davidtaikocha (approved):

PR #17978, opened 2024-08-26
fix(taiko-client): fix some issues about `calculateBaseFee`

YoGhurt111:


packages/taiko-client/driver/chain_syncer/blob/syncer.go
```diff
@@ -377,7 +377,7 @@ func (s *Syncer) insertNewHead(
 		}
 
 		var parentGasExcess uint64
-		if newGasTarget != parentGasTarget {
+		if newGasTarget != parentGasTarget && parentGasTarget != 0 {
```
dantaik, line 380: lets create a separate PR for this.

YoGhurt111: This is to solve the issue of inconsistency between client and contract.

davidtaikocha (approved):

smtmfft (approved):

mask-pp (approved):

PR #17986, opened 2024-08-27
feat(taiko-client): introduce `TxMgrSelector` for proposer / prover

davidtaikocha:


YoGhurt111 (approved):

packages/taiko-client/internal/utils/txmgr_selector.go
```diff
@@ -0,0 +1,67 @@
...
+		return s.txMgr, false
+	}
+
+	// If the private transaction manager has not failed, return it.
+	if s.privateTxMgrFailedAt == nil {
+		return s.privateTxMgr, true
+	}
+
+	// If the private transaction manager has failed, check if it is time to retry.
+	if time.Now().After(s.privateTxMgrFailedAt.Add(s.privateTxMgrRetryInterval)) {
```
mask-pp, line 55:
```suggestion
	if s.privateTxMgrFailedAt == nil || time.Now().After(s.privateTxMgrFailedAt.Add(s.privateTxMgrRetryInterval)) {
```

davidtaikocha: Its checked above.

mask-pp (approved):

PR #17998, opened 2024-08-28
feat(supplementary-contracts): bring over plonk verifier from raiko

adaki2004:
In order to have it in taiko-mono, where every contracts and deployments are, best to have it in here too.

Supplementary, because we only need it on devnet. We will not own Succinct contracts on mainnet, but this accelerates E2E testing.

source: https://github.com/taikoxyz/raiko/pull/350


packages/supplementary-contracts/contracts/plonk_verifier/ISP1Verifier.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+/// @title SP1 Verifier Interface
+/// @author Succinct Labs
+/// @notice This contract is the interface for the SP1 Verifier.
```
dantaik, line 6: Is it possible to `pnpm install` the SP1 repo as a dependencies to avoid copying the code?

dantaik: See https://github.com/taikoxyz/taiko-mono/pull/17878/files

adaki2004: Technically we can, but this is the only file needed from that repo. All the other files are our custom - to facilitate easier devnet E2E testing AFAIK.

dantaik: Previously we have a "third_party" folder, now it's all gone as all external file are dependencies managed by npm. I hope we don't have to bring "third_party" back.

adaki2004:
Closing as according to Yue, we already support that and do not need this specific type.

PR #18010, opened 2024-08-30
test(taiko-client): upgrade full sync and snap sync hive tests

mask-pp:


RogerLamTd (approved):

.github/workflows/taiko-client--hive_test.yml
```diff
@@ -2,7 +2,7 @@ name: "Taiko Client Hive Tests"
 
 on:
   push:
-    branches: [main]
+    branches: [hive_full_snap_upgrade_]
```
dantaik, line 5: so hive will only run on this branch?

dantaik: I think you can revert this change as now if you label a PR with "option.hive-test", it will run the hive tests.

mask-pp: Great, and reverted now.

smtmfft (approved):

PR #18029, opened 2024-09-03
fix(deps): add missing remapping

Akagi201:


packages/protocol/foundry.toml
```diff
@@ -18,6 +18,7 @@ remappings = [
   "openzeppelin/contracts/=node_modules/@openzeppelin/contracts/",
   "@risc0/contracts/=node_modules/risc0-ethereum/contracts/src/",
   "@solady/=node_modules/solady/",
+  "solady/=node_modules/solady/",
```
dantaik, line 21: please notice the previous line.

Akagi201: @dantaik They are different. I can't compile successfully. Can you?

PR #18048, opened 2024-09-09
feat(protocol): allow msg.sender to customize block proposer addresses

dantaik:
This is necessary for [PreconfTaskManager](https://github.com/NethermindEth/Taiko-Preconf-AVS/blob/master/SmartContracts/src/avs/PreconfTaskManager.sol) to assign the actual preconfers as the block proposers, not using the `PreconfTaskManager` contract itself.

### DIFF
--- a/packages/protocol/contracts/L1/libs/LibProposing.sol
+++ b/packages/protocol/contracts/L1/libs/LibProposing.sol
@@ -160,19 +160,19 @@ library LibProposing {
             if (_params.length != 0) {
                 local.params = abi.decode(_params, (TaikoData.BlockParamsV2));
             }
-            if (local.params.proposer == address(0)) {
-                local.params.proposer = msg.sender;
-            } else {
-                require(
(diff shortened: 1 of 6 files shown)


packages/protocol/contracts/L1/libs/LibProposing.sol
```diff
@@ -148,19 +150,33 @@ library LibProposing {
             revert L1_TOO_MANY_BLOCKS();
         }
 
+        address permittedProposer = _resolver.resolve(LibStrings.B_BLOCK_PROPOSER, true);
```
YoGhurt111, line 153: Should this `permittedProposer` come from `PreconfTaskManager `instead of an address from `AddressManager`?

dantaik: permittedProposer is PreconfTaskManager in this case.

YoGhurt111 (approved):

davidtaikocha (approved):

PR #18073, opened 2024-09-11
feat(repo): have certain workflow jobs ignore dependabot PRs

RogerLamTd:
The altered workflows all fail due to GAR / Docker / Vercel permissions on dependabot PRs, they should ignore them now and not burn through credits unnecessarily


.github/workflows/snaefell-ui.yml
```diff
@@ -2,6 +2,8 @@ name: Snaefell UI CI/CD
 
 on:
   push:
+    branches-ignore:
```
dantaik, line 5: do we need this, as dependabot will not be able to push to main using `snaefull-ui-v` tag.

.github/workflows/taikoon-ui.yml
```diff
@@ -2,6 +2,8 @@ name: Taikoon UI CI/CD
 
 on:
   push:
+    branches-ignore:
```
dantaik, line 5: this as well.

.github/workflows/snaefell-ui.yml
RogerLamTd: @dantaik I'm not sure if it's working as intended because you can see [here](https://github.com/taikoxyz/taiko-mono/actions/runs/10822019392/job/30025140228?pr=18074) that dependabot PRs are triggering the workflows even without the `snaefell-ui-v` tag.

dantaik: Then maybe we should disable all vercel actions for the bot in repo--vercel-deploy.yaml

dantaik: I made the above change in a upstream PR.

PR #18089, opened 2024-09-13
feat(protocol): preconfs

cyberhorsey:
based on latest main


packages/protocol/contracts/layer1/based/LibProposing.sol
```diff
@@ -260,27 +260,25 @@ library LibProposing {
 
         // Update certain meta fields
         if (meta_.blobUsed) {
-            if (!LibNetwork.isDencunSupported(block.chainid)) revert L1_BLOB_NOT_AVAILABLE();
+            // if (!LibNetwork.isDencunSupported(block.chainid)) revert L1_BLOB_NOT_AVAILABLE();
```
dantaik, line 263: @cyberhorsey you don't mean to check in these changes, do you?

cyberhorsey: Yes I do, this branch isnt made to be merged yet. This change is necessary, unless we also add in the L1 block Id of Helder to the `isDencunSupported` code.

dantaik: What's the chain id for helder, maybe we can add it to `isDencunSupported` to return false.

PR #18094, opened 2024-09-14
 feat(taiko-client): allow `--l1.beacon` to be optional when a blob server is given

mask-pp:
During `hive test`, I found that `l1.beacon` should be optional otherwise `driver` can never call `BLOB_SOCIAL_SCAN_ENDPOINT` and `BLOB_SERVER` in common situation.


packages/taiko-client/driver/config.go
```diff
@@ -64,11 +65,15 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
 		}
 	}
 
+	if beaconEndpoint == "" && blobServerEndpoint == nil && socialScanEndpoint == nil {
+		return nil, errors.New("empty L1 beacon endpoint, blob server and social scan endpoint")
```
davidtaikocha, line 69:
```suggestion
		return nil, errors.New("empty L1 beacon endpoint, blob server and Social Scan endpoint")
```

davidtaikocha (approved):

YoGhurt111 (approved):

PR #18102, opened 2024-09-16
feat(protocol): add blockId to bond events

dantaik:
Proposers and provers may want to track their bonds, but without blockIds in `BondCredited` and `BondDebited` event, it's really hard to track for which blocks their bonds are burned.


dantaik:
@davidtaikocha @YoGhurt111 will this change require a new node/client release?

davidtaikocha (approved):

davidtaikocha:
> @davidtaikocha @YoGhurt111 will this change require a new node/client release?

It won't, since right now client is not using those two events.

PR #18109, opened 2024-09-16
fix(taiko-client): fix blob server API URL when fetching blob data

mask-pp:


packages/taiko-client/pkg/rpc/blob_datasource.go
```diff
@@ -115,7 +118,7 @@ func (ds *BlobDataSource) getBlobFromServer(ctx context.Context, blobHash common
 		err        error
 	)
 	if ds.socialScanEndpoint != nil {
-		route = "/blob/" + blobHash.String()
+		route = "/blobs/" + blobHash.String()
```
mask-pp, line 121: During `hive-test` I found that calling `http://localhost:3001/blobs/xxx` is the right way to get blob content.

packages/taiko-client/pkg/rpc/blob_datasource.go
```diff
@@ -87,7 +88,9 @@ func (ds *BlobDataSource) GetBlobs(
 		sidecars, err = ds.client.L1Beacon.GetBlobs(ctx, timestamp)
 	}
 	if err != nil {
-		log.Info("Failed to get blobs from beacon, try to use blob server.", "error", err.Error())
+		if !errors.Is(err, pkg.ErrBeaconNotFound) {
```
mask-pp, line 91: Avoid unnecessary logging when `L1Beacon` is empty.

davidtaikocha (approved):

RogerLamTd (approved):

PR #18132, opened 2024-09-18
fix(protocol): reward non-assigned prover 7/8 liveness bond

dantaik:
Without this small fix, provers other than the assigned prover won't have an incentive to prove a block out of the proving window.


davidtaikocha (approved):

davidtaikocha (approved):

packages/protocol/contracts/layer1/based/LibProving.sol
```diff
@@ -607,10 +608,17 @@ library LibProving {
...
                 if (_returnLivenessBond(_local, _proof.data)) {
                     if (_local.assignedProver == msg.sender) {
-                        reward += _local.livenessBond;
+                        unchecked {
+                            reward += _local.livenessBond;
+                        }
                     } else {
                         LibBonds.creditBond(_state, _local.assignedProver, _local.livenessBond);
                     }
+                } else if (!_local.isTopTier) {
```
dantaik, line 617:
```suggestion
                } else {
```

packages/protocol/contracts/layer1/based/LibProving.sol
```diff
@@ -607,10 +608,17 @@ library LibProving {
...
                     if (_local.assignedProver == msg.sender) {
-                        reward += _local.livenessBond;
+                        unchecked {
+                            reward += _local.livenessBond;
+                        }
                     } else {
                         LibBonds.creditBond(_state, _local.assignedProver, _local.livenessBond);
                     }
+                } else {
+                    // Reward 50% of liveness bond to the actual prover, the other 50% burnt.
```
adaki2004, line 618: I think the comment is incorrect here. Example: 100 * 7 / 8 = 87.5% not 50%.

Brechtpd (approved):

adaki2004:
Also since the proposer is paying the liveness bonds (in LibProposing):
`LibBonds.debitBond(_state, _resolver, local.params.proposer, _config.livenessBond);`

Should not we take into consideration the proposer when it comes to returning it ? Maybe it changed in the meantime, but i was in the belief that prover pays the bond, not the proposer. 🤔  (Or proposer == prover ? but i dont see this enforced anywhere )

dantaik:
> Also since the proposer is paying the liveness bonds (in LibProposing): `LibBonds.debitBond(_state, _resolver, local.params.proposer, _config.livenessBond);`
> 
> Should not we take into consideration the proposer when it comes to returning it ? Maybe it changed in the meantime, but i was in the belief that prover pays the bond, not the proposer. 🤔 (Or proposer == prover ? but i dont see this enforced anywhere )

The proposer and assignedProver is the same address, we named them differently because of back-ward compatible reasons. They ARE the same.

adaki2004 (approved):
LGTM but for some reason L2 Genesis tests seems to be failing - so disabled the auto-merge.

PR #18137, opened 2024-09-18
fix(protocol): avoid invocation in Bridge message processing if calldata is "" and value is 0

dantaik:
While reviewing [this transaction](https://etherscan.io/tx/0x23fbbd9838f151d729ae5d8848af77ce7ae665cb9f88414aad139578bd0844ee), we noticed that the processMessage transaction succeeded, but the internal message invocation failed due to an out-of-gas error. This failure occurred because the inner invocation was assigned a gas limit of 1 (since 0 is invalid), despite the invocation being unnecessary.

This PR introduces a check to skip the internal message invocation if the calldata is empty (""), preventing unnecessary execution and gas consumption, and also mark such transactions as "succeeded" instead of "retryable".


packages/protocol/contracts/shared/bridge/Bridge.sol
```diff
@@ -495,7 +495,7 @@ contract Bridge is EssentialContract, IBridge {
         returns (bool success_)
     {
         assert(_message.from != address(this));
-
+        if (_message.data.length == 0) return true;
```
Brechtpd, line 498: If the call is a simple ETH transfer, than the message data length would be zero no, and the internal call would still have to be done?

dantaik: True, now making sure the value must also be zero!

Brechtpd (approved):

adaki2004 (approved):

PR #18138, opened 2024-09-18
ci(protocol): add step to install git in workflows and use new runner

kimo-ice:


.github/workflows/taiko-client--test.yml
```diff
@@ -12,8 +12,11 @@ jobs:
   lint:
     if: github.event.pull_request.draft == false
     name: Lint
-    runs-on: [ubuntu-latest]
+    runs-on: [arc-runner-set]
     steps:
+      - name: Install Git
+        run: sudo apt-get update && sudo apt-get install -y git
```
mask-pp, line 18:
```suggestion
        run: sudo apt-get update && sudo apt-get install -y git make
```

mask-pp (approved):

davidtaikocha (approved):

PR #18150, opened 2024-09-20
refactor(protocol): remove unused code post Ontake fork

dantaik:
There shouldn't be any breaking changes in this PR.


packages/protocol/contracts/layer1/provers/GuardianProver.sol
```diff
@@ -144,58 +142,39 @@ contract GuardianProver is IVerifier, EssentialContract {
...
     }
 
-    /// @dev Called by guardians to approve a guardian proof
-    /// @param _meta The block's metadata.
+    /// @notice Called by guardians to approve a guardian proof (version 2).
+    /// @param _metaV2 The block's metadata (version 2).
     /// @param _tran The valid transition.
     /// @param _proof The tier proof.
-    /// @return approved_ True if the minimum number of approval is acquired, false otherwise.
-    function approve(
```
YoGhurt111, line 180: `approveV2` is not yet implemented on the taiko-client side.
So maybe we need to keep this code or raise the priority of implementing `approveV2`.

davidtaikocha: I think its supported? https://github.com/taikoxyz/taiko-mono/blob/main/packages/taiko-client/prover/proof_submitter/transaction/builder.go#L127

YoGhurt111: Oh, I got this and https://github.com/taikoxyz/taiko-mono/issues/18162 mixed up...

packages/protocol/contracts/layer2/based/LibEIP1559.sol
```diff
@@ -52,7 +53,7 @@ library Lib1559Math {
         returns (uint64)
     {
         if (_gasTarget == 0) {
-            return _newGasTarget;
+            return _gasExcess;
```
YoGhurt111, line 56: Shall we open a new PR to change this gas setting instead of in the cleanup PR?
But it shouldn't get into this conditional judgment.

YoGhurt111 (approved):

packages/protocol/contracts/layer2/based/LibEIP1559.sol
dantaik: reverted this change

davidtaikocha (approved):

PR #18153, opened 2024-09-20
fix(protocol): update nc (netcat) pkg name and add error handling

kimo-ice:


davidtaikocha (approved):

.github/workflows/protocol.yml
```diff
@@ -20,7 +20,8 @@ jobs:
           access_token: ${{ github.token }}
 
       - name: Prepare environment
-        run: sudo apt-get update && sudo apt-get install -y git nc
+        continue-on-error: true
+        run: sudo apt-get update && sudo apt-get install -y git netcat
```
mask-pp, line 24:
```suggestion
        run: sudo apt-get update && sudo apt-get install -y git netcat-traditonal
```

mask-pp: I'm not sure if it's correct to install `netcat` in `arc-runner-set`, but if it's `apt-get` cmd source we need to install `netcat-traditonal`.

kimo-ice: netcat-openbsd (netcat) is the newer 'more powerful' version. It should work fine.

RogerLamTd (approved):

.github/workflows/protocol.yml
mask-pp: Better way, don't need to install nc, just replace line58 like this `until cast chain-id --rpc-url "htp://localhost:8545" 2> /dev/null; do` is ok .

mask-pp (approved):

PR #18159, opened 2024-09-22
ci(protocol): avoid installing `netcat` in action

mask-pp:
- Avoid installing netcat cmd.
- Ignore pnpm-lock.yml file.
- Update golangci.yml config to match the `v1.60.0` version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,6 +120,3 @@ __pycache__/
 
 # Idea
 .idea/
-
-# pnpm lock file
-pnpm-lock.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,6 @@ __pycache__/
 
 # Idea
 .idea/
+
+# pnpm lock file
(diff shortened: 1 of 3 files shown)


kimo-ice (approved):

.gitignore
```diff
@@ -120,3 +120,6 @@ __pycache__/
 
 # Idea
 .idea/
+
+# pnpm lock file
+pnpm-lock.yaml
```
davidtaikocha, line 125: Why ignore this `pnpm-lock.yaml`? Think this is necessary.

davidtaikocha (changes requested):

.gitignore
mask-pp: `pnpm-lock.yml` needs to be updated, after `pnpm install` this file is changed.

davidtaikocha (approved):

PR #18175, opened 2024-09-25
feat(protocol): introduce `getTransitionOrEmpty` to avoid revert

YoGhurt111:
When batch fetching the transitions, don't REVERT when meeting a not found case.


dantaik:
We should add a parameter to the existing functions.

YoGhurt111:
> I think you can simply just call `getTransition()` multiple times and catch errors. If the current `getTransitions()` function is not useful, please delete it. We should not add more functions.

ok

PR #18189, opened 2024-09-30
feat(protocol): check-in `HeklaTaikoToken`

davidtaikocha:
ref: https://holesky.etherscan.io/address/0x01bb2fd6d80942ce95b43c1322530fe690f2bc0e#code

### DIFF
--- a/packages/protocol/contracts/layer1/hekla/HeklaTaikoToken.sol
+++ b/packages/protocol/contracts/layer1/hekla/HeklaTaikoToken.sol
@@ -1,0 +1,145 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20SnapshotUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
+import "../../shared/common/EssentialContract.sol";
+
+/// @title HeklaTaikoToken
+/// @notice The HeklaTaikoToken (TTKOh), in the protocol is used for prover collateral
+/// in the form of bonds. It is an ERC20 token with 18 decimal places of
+/// precision.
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/layer1/hekla/HeklaTaikoToken.sol
```diff
@@ -0,0 +1,145 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20SnapshotUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
+import "../../shared/common/EssentialContract.sol";
+
+/// @title HeklaTaikoToken
+/// @notice The HeklaTaikoToken (TTKOh), in the protocol is used for prover collateral
```
dantaik, line 10: how about we also use TKO as the symbol name?

dantaik: We can just add a function symbol() and return "TAIKO"

dantaik: Add this:

/// @dev Due to historical reasons, the Taiko Token on Hekla has a different storage layout compared to the mainnet token contract. Therefore, we need to maintain this file.

davidtaikocha: Added

YoGhurt111 (approved):

packages/protocol/contracts/layer1/hekla/HeklaTaikoToken.sol
davidtaikocha: > Add this:
> 
> /// @dev Due to historical reasons, the Taiko Token on Hekla has a different storage layout compared to the mainnet token contract. Therefore, we need to maintain this file.

Done

dantaik (approved):

YoGhurt111 (approved):

PR #18200, opened 2024-10-02
fix(taiko-client): out of range for emptyTxLists

fbrv:
When calling /blocks/build, an index out of range error occurs because emptyTxList is initialized with size 0. The logic attempts to populate specific positions in the 2D array, causing the error. This is fixed by initializing emptyTxList with the length of encodedParams.


cyberhorsey:
we use the preconfs branch, I let him know in TG and will push that same fix there!

PR #18226, opened 2024-10-10
chore(repo): remove snaefull-ui

dionysuzx:
## Description of change

~~I do not believe we need this package anymore. The release-please job also takes way too long, I'm not sure why. There has been no activity on this subpackage for months. If it is needed, then I think we can revert this commit later or re-implement it nicely.~~

It seems like this package is still used in prod but perhaps we will sunset it soon. Feel free to close this PR at any time if it's causing noise. Thanks!


KorbinianK:
@d1onys1us The site is still live, but I suppose we can remove it from `release-please`. 
I'd keep the workflows and code, in case we want to update/fix something and create a new release. But removing it from `release-please-config.json` should do the trick, right?

d1onys1us:
> @d1onys1us The site is still live, but I suppose we can remove it from `release-please`. 
> 
> I'd keep the workflows and code, in case we want to update/fix something and create a new release. But removing it from `release-please-config.json` should do the trick, right?

Yea that would help. But do we plan to keep the site up indefinitely? If not, it could be better to have a sunset plan and totally remove it. Maybe can notify the community and say it's coming down in X days/weeks?

PR #18251, opened 2024-10-14
feat(taiko-client): improve `BlobTransactionBuilder` to support `BlobTxListOffset`

davidtaikocha:


YoGhurt111:
If the blob size is 10, block 1 blob size is 6, block 2 blob size is 10, block 3 blob size is 4.
The result is:
block 1 offset 0, blob index 0
block 2 offset 0, blob index 1
block 3 offset 0, blob index 2

rather than:
block 1 offset 0, blob index 0
block 2 offset 0, blob index 1
block 3 offset 6, blob index 0

mask-pp (approved):

packages/taiko-client/proposer/transaction_builder/blob.go
```diff
@@ -154,40 +154,74 @@ func (b *BlobTransactionBuilder) BuildOntake(
...
 		if err != nil {
 			return nil, err
 		}
 
 		encodedParamsArray = append(encodedParamsArray, encodedParams)
+
+		// If the current tx list bytes is the last one, then append the cached blob data to the blobs list.
+		if i == len(txListBytesArray)-1 {
+			var blob = &eth.Blob{}
+			if err := blob.FromData(txListBytesArray[i]); err != nil {
```
mask-pp, line 217:
```suggestion
			if err := blob.FromData(cachedBlobData); err != nil {
```

packages/taiko-client/proposer/transaction_builder/blob.go
```diff
@@ -154,40 +154,74 @@ func (b *BlobTransactionBuilder) BuildOntake(
...
+			BlobIndex:        uint8(len(blobs)),
 		})
 		if err != nil {
 			return nil, err
 		}
 
 		encodedParamsArray = append(encodedParamsArray, encodedParams)
+
+		// If the current tx list bytes is the last one, then append the cached blob data to the blobs list.
+		if i == len(txListBytesArray)-1 {
```
mask-pp, line 215: In this condition if `cachedBlobReused` is true blobs will lose all the collected `txListBytesArrays` in `cachedBlobData`.

packages/taiko-client/proposer/transaction_builder/blob.go
```diff
@@ -154,40 +154,74 @@ func (b *BlobTransactionBuilder) BuildOntake(
...
-		blobs = append(blobs, blob)
-
+		// ABI encode the TaikoL1.proposeBlocksV2 / ProverSet.proposeBlocksV2 parameters.
 		encodedParams, err := encoding.EncodeBlockParamsOntake(&encoding.BlockParamsV2{
 			Coinbase:         b.l2SuggestedFeeRecipient,
 			ParentMetaHash:   [32]byte{},
 			AnchorBlockId:    0,
 			Timestamp:        0,
-			BlobTxListOffset: 0,
+			BlobTxListOffset: uint32(offset),
```
mask-pp, line 204: And also the offset is not true.

mask-pp, line 204:
```suggestion
			BlobTxListOffset: uint32(cachedBlobOffset),
```

PR #18252, opened 2024-10-14
feat(protocol): update mainnet `ontakeForkHeight` config

davidtaikocha:
Assume the block time is 35s, then the chain height will be around `538_256` on 15th Nov.

image


packages/protocol/contracts/layer1/mainnet/rollup/MainnetTaikoL1.sol
```diff
@@ -35,8 +35,8 @@ contract MainnetTaikoL1 is TaikoL1, RollupAddressCache {
                 minGasExcess: 1_340_000_000, // correspond to 0.008847185 gwei basefee
                 maxGasIssuancePerBlock: 600_000_000 // two minutes: 5_000_000 * 120
              }),
-            ontakeForkHeight: 576_000 // = 7200 * 80
-         });
+            ontakeForkHeight: 538_256
```
YoGhurt111, line 38:
```suggestion
            ontakeForkHeight: 538_304
```
How about `538304`? Can be divided by 64.

YoGhurt111 (approved):

dantaik (approved):

PR #18253, opened 2024-10-16
feat(taiko-client): update `OntakeForkHeight` in mainnet

YoGhurt111:
ref: https://github.com/taikoxyz/taiko-mono/pull/18252

### DIFF
--- a/packages/taiko-client/bindings/encoding/protocol_config.go
+++ b/packages/taiko-client/bindings/encoding/protocol_config.go
@@ -55,8 +55,7 @@ var (
 		LivenessBond:          livenessBond,
 		StateRootSyncInternal: 16,
 		MaxAnchorHeightOffset: 64,
-		// TODO: update this value when mainnet fork height is decided
-		OntakeForkHeight: 9_000_000,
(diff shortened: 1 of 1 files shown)


packages/taiko-client/bindings/encoding/protocol_config.go
```diff
@@ -55,8 +55,7 @@ var (
 		LivenessBond:          livenessBond,
 		StateRootSyncInternal: 16,
 		MaxAnchorHeightOffset: 64,
-		// TODO: update this value when mainnet fork height is decided
-		OntakeForkHeight: 9_000_000,
+		OntakeForkHeight:      538_304,
```
mask-pp, line 58: format this line

mask-pp, line 58:
```suggestion
		OntakeForkHeight: 538_304,
```

davidtaikocha: We need to keep indent, why change?

davidtaikocha (approved):

mask-pp (approved):

PR #18256, opened 2024-10-16
docs(taiko-client): update readme how to do integration test

YoGhurt111:


packages/taiko-client/integration_test/README.md
```diff
@@ -1,4 +1,8 @@
 # How to debug test cases?
+- set up config
+```
+export L2_NODE=l2_geth
```
davidtaikocha, line 4: By default which value will be used? `l2_reth`?

YoGhurt111: This config is just for local debugging.
In CI/CD, they have different envs in different git workflows.

YoGhurt111: `l2_geth ` is the default choice in `make test`.

YoGhurt111: When we debug locally, we won't use `entrypoint.sh`, so there is no default value for our local debugging.

davidtaikocha: Then let's set a default value in `entrypoint.sh`?

davidtaikocha: And then we can avoid setting this value manually here

YoGhurt111: We did not use `entrypoint.sh` in local debugging, so setting a default value in `entrypoint.sh` is not helpful.
We set up the test environment manually and just run the test case we want to debug.

davidtaikocha (approved):

dantaik (approved):

PR #18259, opened 2024-10-17
feat(protocol): add anchor tx gaslimit to config

dantaik:
Suggested by @lorenzoXB from Gattaca:

 > Would it be possible to add the anchor tx gas limit (250k) to the onchain config? i think that's the only param affecting "consensus" that's not onchain currently


dantaik:
@davidtaikocha will this be compatible with current Taiko-geth?

davidtaikocha:
> @davidtaikocha will this be compatible with current Taiko-geth?

will break the current taiko-client (ABI decoding error).

dantaik:
After discussing with @davidtaikocha, we have decided not to proceed with the proposed change. If the value is part of the on-chain configuration, any modification would require a hard fork of the protocol, in addition to upgrading the Taiko Geth and prover software.

PR #18260, opened 2024-10-17
chore: Bridge Rebrand

smartprogrammer93:


dantaik:
Could you please provide some background for these changes?

smartprogrammer93:
Apologies, this was supposed to happen on CoreRollup repository and not here. I will explain more details in our meeting today

dantaik:
@smartprogrammer93 I think maybe it's a good idea to modify our code directly to support different branding (names, icons, images)? What do you think :D

smartprogrammer93:
@dantaik that sounds cool. We will have one of our front end engineers contribute here with the modification next week.

PR #18273, opened 2024-10-21
feat(taiko-client): soft block driver APIs

davidtaikocha:
Based on the latest design.

API preview: https://taikoxyz.github.io/taiko-mono/

taiko-geth changes: https://github.com/taikoxyz/taiko-geth/pull/342


packages/taiko-client/driver/preconf_server/api.go
```diff
@@ -0,0 +1,95 @@
...
+	ID               uint64                  `json:"groupId"`
+	TransactionsList types.Transactions      `json:"transactions"`
+	GroupStatus      PreconfBlockGroupStatus `json:"groupStatus"`
+	Signature        string                  `json:"signature"`
+
+	// Block parameters
+	Timestamp             uint64         `json:"timestamp"`
+	Random                common.Hash    `json:"prevRandao"`
+	SuggestedFeeRecipient common.Address `json:"suggestedFeeRecipient"`
+	BaseFeePerGas         uint64         `json:"baseFeePerGas"`
```
dantaik, line 32: do we need this `BaseFeePerGas`?
do we need the coinbase address, is it the same as `suggestedFeeRecipient`?

packages/taiko-client/driver/preconf_server/api.go
```diff
@@ -0,0 +1,95 @@
...
+	StatusFinalBlockGroup   PreconfBlockGroupStatus = "finalBlockGroup"
+	StatusFinalPreconfGroup PreconfBlockGroupStatus = "finalPreconfGroup"
+)
+
+// PreconfTransactionsGroup represents a preconfirmation block group.
+type PreconfTransactionsGroup struct {
+	BlockID          uint64                  `json:"blockId"`
+	ID               uint64                  `json:"groupId"`
+	TransactionsList types.Transactions      `json:"transactions"`
+	GroupStatus      PreconfBlockGroupStatus `json:"groupStatus"`
```
dantaik, line 25: `PreconfBlockGroupStatus` to `PreconfTxGroupStatus`

dantaik: or name 'Tx Group' to 'batch'?

packages/taiko-client/driver/preconf_server/api.go
davidtaikocha: `BaseFeePerGas`  is part of the block, we need to know it to build the block, and yes, its the `coinbase`, in geth block building params, its called `suggestedFeeRecipient`

packages/taiko-client/driver/preconf_server/api.go
davidtaikocha: > `PreconfBlockGroupStatus` to `PreconfTxGroupStatus`

changed

packages/taiko-client/driver/preconf_server/api.go
```diff
@@ -0,0 +1,95 @@
...
+// blocks to the backend L2 execution engine and return a success response.
+//
+//	@Summary	  Insert preconfirmation blocks by the given groups to the backend L2 execution engine, please note that
+//	            the AVS service should sort the groups and make sure all the groups are valid at first.
+//	@Param      body body CreateBlocksByGroupsRequestBody true "preconf blocks creation request body"
+//	@Accept			json
+//	@Produce		json
+//	@Success		200		{object} CreateBlocksByGroupsResponseBody
+//	@Router			/perconfBlocks [post]
+func (s *PreconfAPIServer) CreateBlocksByGroups(c echo.Context) error {
```
dantaik, line 56: I think a new block can be created or an existing block can be updated. So how about: CreateOrUpdateBlockFromBatch(...)

davidtaikocha: renamed

packages/taiko-client/driver/preconf_server/api.go
dantaik: We do need (at least) the following fields form `BlockParamsV2`:

- address coinbase
- uint64 anchorBlockId
- uint64 timestamp

dantaik: Now the driver calculates the base fee? I think this API should not request the basefee because the AVS node doesn't know how to calculate the basefee based on the previous block.

dantaik: I think we need these parameters.

Screenshot 2024-10-21 at 23 43 58

davidtaikocha: > We do need (at least) the following fields form `BlockParamsV2`:
> 
> * address coinbase
> * uint64 anchorBlockId
> * uint64 timestamp

I see, thought AVS would create an `anchorV2` transaction in group on their own, will add

dantaik: We need all these, but the config is global, it should be read from the contract, not from the caller (AVS node)

Screenshot 2024-10-21 at 23 45 28

dantaik (approved):

davidtaikocha:
> https://github.com/taikoxyz/taiko-mono/pull/18273#discussion_r1809073210

added

packages/taiko-client/driver/preconf_server/api.go
davidtaikocha: > We need all these, but the config is global, it should be read from the contract, not from the caller (AVS node)
> 
> Screenshot 2024-10-21 at 23 45 28

added

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,107 @@
...
+// L2 execution engine soft head.
+type RemoveSoftBlocksRequestBody struct {
+	NewHead uint64 `json:"newHead"`
+}
+
+// RemoveSoftBlocksResponseBody represents a response body when resetting the backend
+// L2 execution engine soft head.
+type RemoveSoftBlocksResponseBody struct {
+	CurrentHead  types.Header `json:"currentHead"`
+	HeadsRemoved uint64       `json:"headsRemoved"`
```
dantaik, line 82: NumSoftBlocksRemoved?

dantaik (changes requested):

packages/taiko-client/docs/swagger.yaml
```diff
@@ -0,0 +1,222 @@
...
+      batchType:
+        $ref: "#/definitions/softblocks.TransactionBatchMarker"
+      blockId:
+        type: integer
+      blockParams:
+        $ref: "#/definitions/softblocks.SoftBlockParams"
+      signature:
+        type: string
+      transactions:
+        type: string
```
mskrzypkows, line 54: What is the format of transactions string? Is it zlib compressed as it's done for the `TaikoL1.proposeBlock`?

linoscope:
The API itself LGTM. Still thinking of how we can performance impact of the implementation. @davidtaikocha How are you imagining the implementation would be like? Do we constantly reorg and reexecute at each soft-block proposal (simpler to implement but wasteful) or add "iterative" execution functionality to execution client (efficient but implementation-heavy)?

packages/taiko-client/docs/swagger.yaml
davidtaikocha: yes.

davidtaikocha:
> The API itself LGTM. Still thinking of how we can performance impact of the implementation. @davidtaikocha How are you imagining the implementation would be like? Do we constantly reorg and reexecute at each soft-block proposal (simpler to implement but wasteful) or add "iterative" execution functionality to execution client (efficient but implementation-heavy)?

I think at the first stage, we will choose the reexecute + reorg way to implement it, but make the interval of each batch longer, like 2-3s for each batch.

linoscope:
> > The API itself LGTM. Still thinking of how we can performance impact of the implementation. @davidtaikocha How are you imagining the implementation would be like? Do we constantly reorg and reexecute at each soft-block proposal (simpler to implement but wasteful) or add "iterative" execution functionality to execution client (efficient but implementation-heavy)?
> 
> I think at the first stage, we will choose the reexecute + reorg way to implement it, but make the interval of each batch longer, like 2-3s for each batch.

I agree that starting with reorgs and longer block times, especially given the potential complexity of modifying the execution client. We can run benchmarks to assess performance and also check how these reorgs are reflected in the indexer and wallet then.

CC: @dantaik

packages/taiko-client/docs/swagger.yaml
mskrzypkows: It would be nice to have it in the description 🙂

davidtaikocha: Added in comments, try adding in swagger now as well.

packages/taiko-client/driver/chain_syncer/blob/syncer.go
```diff
@@ -48,6 +49,7 @@ type Syncer struct {
 	reorgDetectedFlag   bool
 	maxRetrieveExponent uint64
 	blobDatasource      *rpc.BlobDataSource
+	mutex               *sync.Mutex
```
mask-pp, line 52:
```suggestion
	mutex               sync.Mutex
```

mask-pp: Then don't need to add code `new(sync.Mutex)` at line93.

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,305 @@
...
+	BlockID uint64 `json:"blockId"`
+	// @param batchId uint64 ID of this transaction batch
+	ID uint64 `json:"batchId"`
+	// @param transactions string zlib compressed RLP encoded bytes of a transactions list
+	TransactionsList []byte `json:"transactions"`
+	// @param batchType TransactionBatchMarker Marker of the transaction batch,
+	// @param either `end_of_block`, `end_of_preconf` or empty
+	BatchMarker TransactionBatchMarker `json:"batchType"`
+	// @param signature string Signature of this transaction batch
+	Signature string `json:"signature"`
```
mask-pp, line 52:
```suggestion
	Signature string `json:"signature" rlp:"-"`
```

mask-pp: Silence rlp encode and decode, so we don't need to copy the heavy struct at line59.

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	return crypto.PubkeyToAddress(*pubKey).Hex() == b.BlockParams.Coinbase.Hex(), nil
+}
+
+// BuildSoftBlockRequestBody represents a request body when handling
+// soft blocks creation requests.
+type BuildSoftBlockRequestBody struct {
+	// @param transactionBatch TransactionBatch Transaction batch to be inserted into the soft block
+	TransactionBatch TransactionBatch `json:"transactionBatch"`
```
mask-pp, line 79:
```suggestion
	TransactionBatch *TransactionBatch `json:"transactionBatch"`
```

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,305 @@
...
+		return c.JSON(http.StatusBadRequest, map[string]string{"error": "L2 execution engine is syncing"})
+	}
+
+	// Check if the softblock batch or the current preconf process is ended.
+	l1Origin, err := s.rpc.L2.L1OriginByID(
+		c.Request().Context(),
+		new(big.Int).SetUint64(reqBody.TransactionBatch.BlockID),
+	)
+	if err != nil && err.Error() != ethereum.NotFound.Error() {
+		return err
```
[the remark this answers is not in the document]
davidtaikocha: Changed

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	if l1Origin != nil {
+		if l1Origin.EndOfBlock {
+			return c.JSON(http.StatusBadRequest, map[string]string{"error": "soft block has already been marked as ended"})
+		}
+		if l1Origin.EndOfPreconf {
+			return c.JSON(http.StatusBadRequest, map[string]string{
```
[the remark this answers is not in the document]
davidtaikocha: Added a `returnError` helper method: https://github.com/taikoxyz/taiko-mono/pull/18273/commits/479711d6de4ed1d924dd035343ac1200a858ded4

packages/taiko-client/driver/soft_blocks/api.go
davidtaikocha: Is this the same with passing an empty string here?

packages/taiko-client/driver/soft_blocks/api.go
davidtaikocha: Changed

packages/taiko-client/driver/chain_syncer/blob/syncer.go
davidtaikocha: Changed

packages/taiko-client/driver/soft_blocks/api.go
mask-pp: That's it!

packages/taiko-client/driver/driver_test.go
```diff
@@ -344,6 +354,301 @@ func (s *DriverTestSuite) InitProposer() {
...
+	}
+
+	b, err := encodeAndCompressTxList([]*types.Transaction{signedTx})
+	s.Nil(err)
+
+	var marker softblocks.TransactionBatchMarker
+	if endOfBlock {
+		marker = softblocks.BatchMarkerEOB
+	} else if endOfPreconf {
+		marker = softblocks.BatchMarkerEOP
```
YoGhurt111, line 590: What is the difference between the EOB and EOP?

davidtaikocha: EoB means this soft block is ended, can't append new batches, EoP means this current preconfer is done, can't append new batches and new soft blocks.

packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -374,23 +374,50 @@ func (c *Client) GetPoolContent(
...
-	l1Head, err := c.L1.HeaderByNumber(ctx, nil)
+	// Get the latest L2 block header at first.
+	l2Head, err := c.L2.HeaderByNumber(ctx, nil)
 	if err != nil {
 		return nil, err
 	}
 
-	l2Head, err := c.L2.HeaderByNumber(ctx, nil)
+	l1Origin, err := c.L2.L1OriginByID(ctx, l2Head.Number)
+	if err != nil && err.Error() != ethereum.NotFound.Error() {
```
mask-pp, line 384: Shouldn't `err` be returned directly? Otherwise, panic is likely to occur below.

davidtaikocha: `l1Origin != nil` is checked below, don't think it will panic

davidtaikocha: And if there is no soft block, there will be `ethereum.NotFound` error, which is by design

packages/taiko-client/driver/chain_syncer/blob/soft_block.go
```diff
@@ -0,0 +1,277 @@
...
+
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/bindings/encoding"
+	softblocks "github.com/taikoxyz/taiko-mono/packages/taiko-client/driver/soft_blocks"
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/internal/utils"
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/pkg/rpc"
+)
+
+// InsertSoftBlockFromTransactionsBatch inserts a soft block into the L2 execution engine's blockchain
+// from the given transactions batch.
+func (s *Syncer) InsertSoftBlockFromTransactionsBatch(
```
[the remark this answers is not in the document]
davidtaikocha: the API is protected by JWT, the real AVS should do the preconfer check and then call the API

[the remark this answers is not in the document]
YoGhurt111: OK, we don't have an interface to AVS in taiko-geth yet?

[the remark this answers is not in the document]
davidtaikocha: The third-party should implement the AVS service,  we won't, we just provide the API with JWT

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+	}
+
+	// Check if the L2 execution engine is syncing from L1.
+	progress, err := s.rpc.L2ExecutionEngineSyncProgress(c.Request().Context())
+	if err != nil {
+		return c.JSON(http.StatusInternalServerError, map[string]string{"error": err.Error()})
+	}
+	if progress.IsSyncing() {
+		return s.returnError(c, http.StatusBadRequest, errors.New("L2 execution engine is syncing"))
```
mask-pp, line 167:
```suggestion
		return s.returnError(c, http.StatusBadRequest, errors.New("l2 execution engine is syncing"))
```

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// TransactionBatch represents a soft block group.
+type TransactionBatch struct {
+	// @param blockId uint64 Block ID of the soft block
+	BlockID uint64 `json:"blockId"`
+	// @param batchId uint64 ID of this transaction batch
+	ID uint64 `json:"batchId"`
+	// @param transactions string zlib compressed RLP encoded bytes of a transactions list
+	TransactionsList []byte `json:"transactions"`
```
mask-pp, line 47: Use `types.Transactions`?

packages/taiko-client/driver/soft_blocks/api.go
davidtaikocha: We should always use `L2` instead `l2`

packages/taiko-client/driver/soft_blocks/api.go
davidtaikocha: Because we want the data to be compressed at first before sending the request, to reduce the network load.

packages/taiko-client/driver/chain_syncer/blob/soft_block.go
```diff
@@ -0,0 +1,277 @@
...
+			new(big.Int).SetUint64(blockID),
+			baseFee,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create TaikoL2.anchorV2 transaction: %w", err)
+		}
+
+		txList = append([]*types.Transaction{anchorTx}, txList...)
+	} else {
+		prevSoftBlock, err := s.rpc.L2.BlockByNumber(ctx, new(big.Int).SetUint64(blockID))
```
mask-pp, line 98: Does here need to ensure `blockID` is the latest L2 block number?

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,307 @@
...
+//	  @Description	 ensuring the latest block ID does not exceed the given height. This method will fail if
+//	  @Description	 the block with an ID one greater than the specified height is not a soft block. If the
+//	  @Description	 specified block height is greater than the latest soft block ID, the method will succeed
+//	  @Description	 without modifying the blockchain.
+//		@Param      body body RemoveSoftBlocksRequestBody true "soft blocks removing request body"
+//		@Accept			json
+//		@Produce		json
+//		@Success		200	{object} RemoveSoftBlocksResponseBody
+//		@Router			/softBlocks [delete]
+func (s *SoftBlockAPIServer) RemoveSoftBlocks(c echo.Context) error {
```
mask-pp, line 243: Do we need to consider storing the transactions that belong to removed soft blocks?

davidtaikocha: Don't think so, AVS can impl that if they need

mask-pp: Do you mean the AVS node will handle the transactions?

davidtaikocha: If they want, they can store those removed txs before sending this request

packages/taiko-client/driver/chain_syncer/blob/soft_block.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		return err
+	}
+
+	lastVerifiedBlockHash, err := s.rpc.GetLastVerifiedBlockHash(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to fetch last verified block hash: %w", err)
+	}
+
+	fc := &engine.ForkchoiceStateV1{
```
mask-pp, line 277: We can only remove `softblocks`, so the `SafeBlockHash` and `FinalizedBlockHash` aren't changed and there is no need to set them. Besides the `SafeBlockHash` is not right, it should be equal to `canonicalHeadL1Origin.L2BlockHash`.

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	log.Info(
+		"New soft block removing request",
+		"newLastBlockId", reqBody.NewLastBlockID,
+		"canonicalHead", canonicalHeadL1Origin.BlockID.Uint64(),
+		"currentHead", currentHead.Number.Uint64(),
+	)
+
+	if reqBody.NewLastBlockID < canonicalHeadL1Origin.BlockID.Uint64() {
```
mask-pp, line 265: It doesn't need to contain the `SafeBlock` number.

mask-pp, line 265: If `reqBody.NewLastBlockID  == canonicalHeadL1Origin.BlockID` then `SafeBlock` will be removed, is it expected？

packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,304 @@
...
+		return s.returnError(
+			c,
+			http.StatusBadRequest,
+			errors.New("newLastBlockId must not be smaller than the canonical chain's highest block ID"),
```
mask-pp, line 269:
```suggestion
			errors.New("newLastBlockId must not be smaller than the canonical chain's safe block ID"),
```

packages/taiko-client/driver/chain_syncer/blob/soft_block.go
```diff
@@ -0,0 +1,291 @@
...
+			parent.GasUsed,
+			&protocolConfigs.BaseFeeConfig,
+			new(big.Int).SetUint64(blockID),
+			baseFee,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create TaikoL2.anchorV2 transaction: %w", err)
+		}
+
+		txList = append([]*types.Transaction{anchorTx}, txList...)
```
mask-pp, line 96:
```suggestion
		txList = append(txList, anchorTx)
```

packages/taiko-client/driver/chain_syncer/blob/soft_block.go
```diff
@@ -0,0 +1,291 @@
...
+				blockNums,
+			)
+		}
+
+		// Check the previous soft block status.
+		l1Origin, err := s.rpc.L2.L1OriginByID(ctx, prevSoftBlock.Number())
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch L1 origin for block %d: %w", blockID, err)
+		}
+		if l1Origin.BatchID.Uint64()+1 != batchID {
```
mask-pp, line 122:
```suggestion
		if l1Origin.BatchID == nil || l1Origin.BatchID.Uint64()+1 != batchID {
```

mask-pp, line 122: Plz use `if l1Origin != nil` to wrap the next code.

packages/taiko-client/driver/chain_syncer/blob/soft_block.go
davidtaikocha: Anchor tx must be the first one of each block

RogerLamTd (approved):

YoGhurt111 (approved):

YoGhurt111 (approved):

dantaik (approved):

PR #18281, opened 2024-10-23
suggested changes

dantaik:
@davidtaikocha I would suggest the following changes to your PR:

1. Use "Soft block" instead of "tentative block"
1. If we have to validate a parameter with a known correct value for building a soft block, we do not request the callers to provide the value, we use the correct value directly.
2. Only support building/updating one soft block per API

This PR is not ready to be merged and I will stop working on it. Feel free to bring some changes over to your PR.


packages/taiko-client/driver/soft_blocks/api.go
```diff
@@ -0,0 +1,112 @@
...
+// TransactionBatch represents a preconfirmation block group.
+type TransactionBatch struct {
+	BlockID          uint64                 `json:"blockId"`
+	ID               uint64                 `json:"batchId"`
+	TransactionsList types.Transactions     `json:"transactions"`
+	BatchMarker      TransactionBatchMarker `json:"batchType"`
+	Signature        string                 `json:"signature"`
+
+	// Block parameters
+	Timestamp             uint64         `json:"timestamp"`
```
dantaik, line 29: we only need to ask for values that the proposer can decide, not those our protocol contracts will calculate:

- params.timestamp
- params.anchorBlockId
- params.coinbase
- anchorStateRoot
- anchorBlockHash

It will be great if the server can query the right `anchorStateRoot`, and `anchorBlockHash` values to prevent a malicious preconfer to use the wrong values -- image the wrong values are used and Taikoscan display total bad blocks with these incorrect values.

PR #18285, opened 2024-10-24
test(taiko-client): introduce TestTxPoolContentWithMinTip test case

mask-pp:


dantaik (approved):

packages/taiko-client/proposer/proposer_test.go
```diff
@@ -119,6 +121,89 @@ func (s *ProposerTestSuite) SetupTest() {
 	s.cancel = cancel
 }
 
+func (s *ProposerTestSuite) TestTxPoolContentWithMinTip() {
+	if os.Getenv("L2_NODE") == "l2_reth" {
+		s.T().Skip()
+	}
+	skList := []string{
```
davidtaikocha, line 128:
```suggestion
	skList := []string{
```

what's the meaning of `sk` here?

packages/taiko-client/internal/testutils/helper.go
```diff
@@ -282,6 +282,50 @@ func SignatureFromRSV(r, s string, v byte) []byte {
 	return append(append(hexutil.MustDecode(r), hexutil.MustDecode(s)...), v)
 }
 
+func SendDynamicFeeTxWithNonce(
```
davidtaikocha, line 285: Seems this function name doesn't match what it will do? maybe `assembleTestTx()`?

packages/taiko-client/proposer/proposer_test.go
```diff
@@ -119,6 +121,89 @@ func (s *ProposerTestSuite) SetupTest() {
...
+		chainID = l2Cli.ChainID
+	)
+
+	for _, sk := range skList {
+		priv, err := crypto.ToECDSA(common.FromHex(sk))
+		s.Nil(err)
+		privs = append(privs, priv)
+	}
+
+	for _, priv := range privs {
```
davidtaikocha, line 149:
```suggestion
	for _, priv := range privKeys {
```

packages/taiko-client/proposer/proposer_test.go
```diff
@@ -119,6 +121,89 @@ func (s *ProposerTestSuite) SetupTest() {
...
+			testCase.blockMaxTxListBytes,
+			p.LocalAddresses,
+			testCase.maxTransactionsLists,
+			0,
+			p.chainConfig,
+		)
+		s.Nil(err)
+
+		s.GreaterOrEqual(int(testCase.maxTransactionsLists), len(res))
+		for i, txsLen := range testCase.txLengthList {
```
davidtaikocha, line 199: Let's also assert the nonce here?

packages/taiko-client/proposer/proposer_test.go
mask-pp: security key, abbreviation for cryptographic terms.

davidtaikocha: I think its just private key here right? but with Hex. lets use sth like `privetKeyHexList` /  `privateKeyStringList`?

packages/taiko-client/proposer/proposer_test.go
mask-pp: Ok, will fix it.

mask-pp: added.

davidtaikocha (approved):

PR #18312, opened 2024-10-25
refactor(protocol): set `blk.proposedIn` in genesis block

dantaik:


packages/protocol/contracts/layer1/based/LibUtils.sol
```diff
@@ -48,8 +48,7 @@ library LibUtils {
         TaikoData.BlockV2 storage blk = _state.blocks[0];
         blk.nextTransitionId = 2;
         blk.proposedAt = uint64(block.timestamp);
-        // TODO:
-        // blk.proposedIn = uint64(block.number);
+        blk.proposedIn = uint64(block.number);
```
adaki2004, line 51: hmm, i think it should go to similar `reinit`, no ? Cause `LibUtils.init()` is only called once at TaikoL1 init.

dantaik: For mainnet, the same slot in the ring buffer is already overwritten.

adaki2004 (approved):

YoGhurt111 (approved):

PR #18315, opened 2024-10-25
refactor(protocol): improve mapping syntax and modifier order for clarity and correctness

silentdea:
1. mapping(address prover => bool isProver) public isProver; is not valid because mapping definitions in Solidity do not allow a named key and value type (such as address prover and bool isProver) within the mapping definition itself. Simplifying it to mapping(address => bool) makes the code syntactically correct and more consistent.

2. In Solidity, modifiers need to be listed in the order they should be applied, and initializer should precede the nonZeroAddr modifier to ensure initialization is performed before nonZeroAddr is checked.

### DIFF
--- a/packages/protocol/contracts/layer1/provers/ProverSet.sol
+++ b/packages/protocol/contracts/layer1/provers/ProverSet.sol
@@ -14,6 +14,7 @@ interface IHasRecipient {
     function recipient() external view returns (address);
 }
 
+/// @dev Implements the contract for handling TKO tokens and relaying `proveBlock` calls.
 /// @title ProverSet
 /// @notice A contract that holds TKO token and acts as a Taiko prover. This contract will simply
 /// relay `proveBlock` calls to TaikoL1 so msg.sender doesn't need to hold any TKO.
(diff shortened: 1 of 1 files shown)


dantaik:
There will be a big PR which focus on NatSpec comments improvement only. Please hold off this PR.

silentdea:
Okay, Thank you where do you will announce it? Because I got many minor improvements, but don't want to fill the PR section :) I will hold off my PR till the big one is out.

dantaik:
> 1. mapping(address prover => bool isProver) public isProver; is not valid because mapping definitions in Solidity do not allow a named key and value type (such as address prover and bool isProver) within the mapping definition itself. Simplifying it to mapping(address => bool) makes the code syntactically correct and more consistent.

You should check the latest solc release.`mapping(address prover => bool isProver)` is preferred by us.

> 2. In Solidity, modifiers need to be listed in the order they should be applied, and initializer should precede the nonZeroAddr modifier to ensure initialization is performed before nonZeroAddr is checked.

We do not want to read from storage before checking `nonZeroAddr`

PR #18322, opened 2024-10-28
chore(protocol): change Hekla sharingPctg to 80% & gasIssuancePerSecond to 1000000

dantaik:
@davidtaikocha when we release 1.10.0, we can test that the fee sharing pctg will be 80%. Do you think taiko-geth already support this?

Maybe we should also change _gasIssuancePerSecond_ to a much smaller value for Hekla?

### DIFF
--- a/packages/protocol/contracts/layer1/hekla/HeklaTaikoL1.sol
+++ b/packages/protocol/contracts/layer1/hekla/HeklaTaikoL1.sol
@@ -23,7 +23,7 @@ contract HeklaTaikoL1 is TaikoL1 {
             baseFeeConfig: LibSharedData.BaseFeeConfig({
                 adjustmentQuotient: 8,
                 sharingPctg: 80,
-                gasIssuancePerSecond: 5_000_000,
+                gasIssuancePerSecond: 1_000_000,
                 minGasExcess: 1_340_000_000,
                 maxGasIssuancePerBlock: 600_000_000 // two minutes
              }),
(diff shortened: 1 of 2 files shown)


davidtaikocha:
Its supported.

davidtaikocha (approved):

YoGhurt111 (approved):

packages/protocol/contracts/layer1/hekla/HeklaTaikoL1.sol
```diff
@@ -22,7 +22,7 @@ contract HeklaTaikoL1 is TaikoL1 {
             maxAnchorHeightOffset: 64,
             baseFeeConfig: LibSharedData.BaseFeeConfig({
                 adjustmentQuotient: 8,
-                sharingPctg: 75,
+                sharingPctg: 80,
                 gasIssuancePerSecond: 5_000_000,
```
dantaik, line 26:
```suggestion
                gasIssuancePerSecond: 1_000_000,
```
??

davidtaikocha: Also supported

dantaik, line 26:
```suggestion
                gasIssuancePerSecond: 1_000_000,
```

YoGhurt111 (approved):

PR #18331, opened 2024-10-28
fix(protocol): revert a change to maintain taiko-geth compatibility 

dantaik:
The bug was introduced in https://github.com/taikoxyz/taiko-mono/pull/18308/files

### DIFF
--- a/packages/protocol/contracts/layer1/based/LibProposing.sol
+++ b/packages/protocol/contracts/layer1/based/LibProposing.sol
@@ -182,7 +182,8 @@ library LibProposing {
 
         unchecked {
             if (local.params.anchorBlockId == 0) {
-                local.params.anchorBlockId = uint64(block.number - 1);
+                local.params.anchorBlockId =
+                    local.postFork ? uint64(block.number - 1) : uint64(block.number);
             }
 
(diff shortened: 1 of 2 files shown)


davidtaikocha (approved):

packages/protocol/contracts/layer1/based/LibProposing.sol
```diff
@@ -182,7 +182,8 @@ library LibProposing {
 
         unchecked {
             if (local.params.anchorBlockId == 0) {
-                local.params.anchorBlockId = uint64(block.number - 1);
+                local.params.anchorBlockId =
+                    local.postFork ? uint64(block.number - 1) : uint64(block.number);
```
YoGhurt111, line 186: image
Why do we need to change it back here?

mask-pp, line 186: If `local.postFork` is false, we can't get `anchorBlockHash` at Line239.

davidtaikocha: changed

mask-pp (approved):

PR #18332, opened 2024-10-29
test(taiko-client): trigger client tests when there are changes in the protocol/contracts

mask-pp:


.github/workflows/taiko-client--test.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
+      - "packages/protocol/contracts/**"
```
dantaik, line 7:
```suggestion
      - "packages/protocol/contracts/layer1/based/**"
```

davidtaikocha (approved):

YoGhurt111 (approved):

PR #18349, opened 2024-10-30
feat(taiko-client): ignore not supported tier when init submitter

YoGhurt111:


packages/taiko-client/prover/init.go
```diff
@@ -134,7 +134,8 @@ func (p *Prover) initProofSubmitters(
 		case encoding.TierGuardianMajorityID:
 			producer = proofProducer.NewGuardianProofProducer(encoding.TierGuardianMajorityID, p.cfg.EnableLivenessBondProof)
 		default:
-			return fmt.Errorf("unsupported tier: %d", tier.ID)
+			log.Warn("unsupported tier", "id", tier.ID)
```
davidtaikocha, line 137: We also need to make sure that the unsupported tier won't be selected, otherwise, client will panic.

YoGhurt111: OK, I roughly checked that if there is no corresponding tier it will throw errTierNotFound.
But anyway, there seems to be no need to change it now, I will close this pr.

PR #18350, opened 2024-10-30
feat(taiko-client): client changes based on #18150

davidtaikocha:
ref: #18150


packages/taiko-client/bindings/encoding/protocol_config.go
```diff
@@ -19,7 +19,7 @@ var (
 		LivenessBond:          livenessBond,
 		StateRootSyncInternal: 16,
 		MaxAnchorHeightOffset: 64,
-		OntakeForkHeight:      2,
+		OntakeForkHeight:      0,
```
YoGhurt111, line 22: Did it change to 0 in taiko-geth as well?

YoGhurt111 (approved):

packages/taiko-client/bindings/encoding/protocol_config.go
davidtaikocha: Not yet, will do after this one merged into `main`

mask-pp (dismissed):

packages/taiko-client/bindings/encoding/protocol_config.go
YoGhurt111: Got it, so the CI won't succeed.

YoGhurt111 (dismissed):

smtmfft (approved):

mask-pp (approved):

PR #18368, opened 2024-11-02
chore(protocol): restore proving window changes

dantaik:
We did not announce these proving window changes in advance. Lets revert the deployment and use this new version with old values.

@RogerLamTd, let’s also document the configurations for the mainnet proving window and cooldown window on our website. Additionally, please add a link in `TierProviderBase.sol `pointing to the documentation for these configurations.


YoGhurt111 (approved):

packages/protocol/contracts/layer1/tiers/TierProviderBase.sol
```diff
@@ -64,22 +63,22 @@ abstract contract TierProviderBase is ITierProvider {
         pure
         returns (ITierProvider.Tier memory)
     {
-        // cooldownWindow is 4 hours and provingWindow is 60 minutes
-        return _buildTier(_verifierName, BOND_UNIT * 2, 4, 60);
+        // cooldownWindow is 1440 minutes and provingWindow is 60 minutes
```
davidtaikocha, line 66: Seems that the proving window is 5 hrs (`240mins + 60mins`) instead of `60mins`?

dantaik: Yes, lets change it later with an announcement.

davidtaikocha: But before ontake update, i think on mainnet its `60mins`

davidtaikocha (approved):

PR #18371, opened 2024-11-02
chore(protocol): change bond amounts, proving windows, and cooldown windows

dantaik:
We can launch these changes alongside the ZK proofs. However, we’ll need to draft an announcement (@dantaik) and update the documentation afterward (@RogerLamTd).

@smtmfft, please review to confirm if these values align as expected.


davidtaikocha (approved):

packages/protocol/contracts/layer1/tiers/TierProviderBase.sol
```diff
@@ -15,88 +15,86 @@ abstract contract TierProviderBase is ITierProvider {
     /// service may be paused if gas prices are excessively high. Since block proving is
     /// asynchronous, this grace period allows provers to defer submissions until gas
     /// prices become more favorable, potentially reducing transaction costs.
```
davidtaikocha, line 17: These comments of `GRACE_PERIOD` can also be removed?

smtmfft (approved):

PR #18378, opened 2024-11-03
feat(protocol): rename 3 fields in TaikoData

dantaik:
Rename `proposedAt` to `timestamp`, `proposedIn` to `anchorBlockId`, and `lastSynecdAt` for `lastSyncedAt` for improved clarity. This PR introduces a breaking change for taiko-geth. However, since we already have another breaking change in [PR #18344](https://github.com/taikoxyz/taiko-mono/pull/18344), taiko-geth could incorporate both changes simultaneously.

Also deleted `maxBlocksToVerifyPerProof` from ITierProvider.Tier 

@davidtaikocha if these changes are too troublesome for the client/node to adapt, then we can afford not doing it.


davidtaikocha:
> @davidtaikocha if these changes are too troublesome for the client/node to adapt, then we can afford not doing it.

Need to do some tests first.

PR #18438, opened 2024-11-04
docs(docs-site): update network config to match new upgrade

RogerLamTd:
merge when new tier router has been deployed from #18371


RogerLamTd:
outdated

PR #18442, opened 2024-11-05
feat(taiko-client): compatible changes for `lastProposedIn`

YoGhurt111:


packages/taiko-client/pkg/rpc/utils.go
```diff
@@ -31,6 +32,7 @@ var (
 		syscall.SIGTERM,
 		syscall.SIGQUIT,
 	}
+	ErrSlotBMarshal = errors.New("abi: cannot marshal in to go type: length insufficient 160 require 192")
```
mask-pp, line 35:
```suggestion
	ErrSlotBMarshal = errors.New("abi: cannot marshal into go type: length insufficient 160 require 192")
```

dantaik:
Will this change be compatible with 1.10.0 (without this new field)?

YoGhurt111:
> Will this change be compatible with 1.10.0 (without this new field)?

Yes, it is compatible with or without the new field.

packages/taiko-client/pkg/rpc/utils.go
YoGhurt111: It is copied from the error msg, so no need to change.

mask-pp (approved):

mask-pp:
Need to fix the failed ci.

dantaik (approved):

packages/taiko-client/pkg/rpc/utils.go
```diff
@@ -63,9 +65,22 @@ func GetProtocolStateVariables(
 	opts.Context, cancel = CtxWithTimeoutOrDefault(opts.Context, defaultTimeout)
 	defer cancel()
 
+	var slotBV1 bindings.TaikoDataSlotBV1
 	slotA, slotB, err := taikoL1Client.GetStateVariables(opts)
 	if err != nil {
-		return nil, err
+		if errors.Is(err, ErrSlotBMarshal) {
+			slotA, slotBV1, err = taikoL1Client.GetStateVariablesV1(opts)
```
davidtaikocha, line 72: Shall we also check the error here?

packages/taiko-client/pkg/rpc/utils.go
```diff
@@ -63,9 +65,22 @@ func GetProtocolStateVariables(
...
+				NumBlocks:           slotBV1.NumBlocks,
+				LastVerifiedBlockId: slotBV1.LastVerifiedBlockId,
+				ProvingPaused:       slotBV1.ProvingPaused,
+				LastProposedIn:      nil,
+				LastUnpausedAt:      slotBV1.LastUnpausedAt,
+			}
+		}
+		if err != nil {
+			return nil, err
+		}
```
davidtaikocha, line 83: Seems there are still some places are using the raw `c.TaikoL1.GetStateVariables`, lets also change them to use the method here?

YoGhurt111: Yes, but we also see the raw `GetStateVariables ` used in Indexer.
Maybe we should make Jeff noticed?

davidtaikocha: > Yes, but we also see the raw `GetStateVariables ` used in Indexer. Maybe we should make Jeff noticed?

Correct, cc @cyberhorsey

packages/taiko-client/pkg/rpc/utils.go
YoGhurt111: I think there is no need.
If we fail both ways, just throw an error!

davidtaikocha: If we don't check the `err` here, `slotBV1` / `slotA` will be `nil` and the code below will panic?

YoGhurt111:
I tested this pr with reverting https://github.com/taikoxyz/taiko-mono/pull/18379.
All ci passed.

packages/taiko-client/pkg/rpc/utils.go
YoGhurt111: Yes, fixed

davidtaikocha (approved):

PR #18451, opened 2024-11-06
feat(protocol): allow `local.params.parentMetaHash` to remain as 0

dantaik:
Previously if this parameter is zero, we set it to parent's hash, now we simply use 0. This will avoid a SLOAD for many blocks.

@davidtaikocha  @smtmfft  will this be a breaking change for client or prover? Note now the metadata may have `parentMetaHash` as 0.

Screenshot 2024-11-06 at 12 21 02


davidtaikocha:
Don't think its a break change for client, but might break zk cc @smtmfft

smtmfft:
we get parentMeta from blockpropose event, should be the same I think

smtmfft (approved):

davidtaikocha (approved):

PR #18455, opened 2024-11-06
test(taiko-client): build taiko-client@softblock docker image for integration test

mask-pp:


davidtaikocha:
actually u don't need to open this PR to trigger the docker build, just push it to the branch. So will close it for now.

PR #18494, opened 2024-11-16
perf(protocol): increase optimizer_runs to max value in foundry.toml

lagunovsky:
I'm not sure about the value of the optimizer_runs. But I think this is a good spot for optimization.


swarna1101:
This is not requires at the moment, hence closing the PR

PR #18508, opened 2024-11-23
feat(protocol): introduce ForkManager to improve protocol fork management

dantaik:
Currently, the TaikoL1 protocol required each fork's codebase to contain implementations for both the current fork and the next fork within the same contract. This created unnecessary code complexity and tight coupling between fork versions.

With the new ForkManager, we can:
- Split fork implementations into separate, focused contracts
- Maintain cleaner separation of concerns between fork versions
- Support multiple forks as long as their storage layouts remain compatible

This change improves code maintainability and makes the fork upgrade process more modular.

BTW, I'm aware of Diamond proxy design but I'm in favor of this ForkManager implementation for its simplicity.


dantaik:
The storage layout of ForkManager is compatible with the layout of TaikoL1.

Screenshot 2024-11-23 at 15 05 37

packages/protocol/contracts/layer1/fork/ForkManager.sol
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    function _authorizeUpgrade(address) internal virtual override onlyOwner { }
+
+    /// @notice Determines if the call should be routed to the old fork.
+    /// @dev This function is intended to be overridden in derived contracts to provide custom
+    /// routing logic.
+    /// @param  _selector The function selector of the call.
+    /// @return A boolean value indicating whether the call should be routed to the old fork.
+    function shouldRouteToOldFork(bytes4 _selector) internal pure virtual returns (bool) { }
```
YoGhurt111, line 70: If a function has the same selector in oldFork and newFork, how can we clarify?

packages/protocol/contracts/layer1/fork/ForkManager.sol
```diff
@@ -0,0 +1,71 @@
...
+/// calling upgradeTo, the proxy should always upgrade to a new ForkManager implementation, not an
+/// actual fork implementation.
+/// It is strongly advised to name functions differently for the same functionality across the two
+/// forks, as it is not possible to route the same function to two different forks.
+///
+///                         +--> newFork
+/// PROXY -> FORK_MANAGER --|
+///                         +--> oldFork
+
+contract ForkManager is UUPSUpgradeable, Ownable2StepUpgradeable {
```
davidtaikocha, line 23: Do we need an `init` method here for proxy deployment?

adaki2004, line 23: An instance of this `ForkManager` (even behind proxy) is not really upgradeable, or at least not the 2 main variables (oldFork, newFork). 
Since you placed here a constructor and not `init`, i suppose it is intentional from your side, but in this case we dont need the `Upgradeable` pattern. (Or if we do want to do that, just remove the immutability of the 2 fields).

packages/protocol/contracts/layer1/fork/ForkManager.sol
```diff
@@ -0,0 +1,71 @@
...
+
+    receive() external payable virtual {
+        _fallback();
+    }
+
+    function isForkManager() public pure returns (bool) {
+        return true;
+    }
+
+    function _fallback() internal virtual {
```
adaki2004, line 48: Should not we restrict reentrancy ? 🤔 Not sure tho, cause propose/prove already protected.

packages/protocol/contracts/layer1/fork/ForkManager.sol
dantaik: `oldFork` and `currFork` are not saved in storage as in a normal proxy, they are complied into the deployed code by the constructor and immutable variables..

packages/protocol/contracts/layer1/fork/ForkManager.sol
dantaik: I don't think we need to do it here.

packages/protocol/contracts/layer1/fork/ForkManager.sol
dantaik: The routing logic relies entirely on the `shouldRouteToOldFork` function. If it returns `false`, the new fork will be used. This means we should adopt different function names for the old and new forks—for example, `proveBlockV3` for the old fork and `proveBlockV4` for the new fork.

With this approach, if `shouldRouteToOldFork` returns `true` for `proveBlockV3`, we can seamlessly support both `proveBlockV3` and `proveBlockV4`. However, if both functions share the same name, only one fork can be supported.

packages/protocol/contracts/layer1/fork/ForkManager.sol
dantaik: No need, please take a look at the unit tests to see how ForkManager shall be used.

dantaik: `oldFork` and `currFork` are not saved in storage as in a normal proxy, they are complied into the deployed code by the constructor and immutable variables..

adaki2004 (approved):

davidtaikocha (approved):

davidtaikocha (approved):

YoGhurt111 (approved):

YoGhurt111 (approved):

PR #18524, opened 2024-11-26
feat(taiko-client): add more preconf logs

mask-pp:


packages/taiko-client/driver/chain_syncer/blob/soft_block.go
```diff
@@ -248,16 +250,31 @@ func (s *Syncer) InsertSoftBlockFromTransactionsBatch(
...
-		"transactions", len(payload.Transactions),
-		"baseFee", utils.WeiToGWei(header.BaseFee),
-		"withdrawals", len(payload.Withdrawals),
-		"endOfBlock", attributes.L1Origin.EndOfBlock,
-		"endOfPreconf", attributes.L1Origin.EndOfPreconf,
-	)
+	if batchID == 0 {
+		log.Info(
+			"⏰ New soft L2 block inserted",
+			"batchID", batchID,
```
davidtaikocha, line 256: don't think this is necessary, we can just use `⏰ New soft L2 block inserted` and check the `batchID`

davidtaikocha (approved):

PR #18545, opened 2024-12-02
L1StaticCall

dantaik:
This PR introduces an idea for performing an L1StaticCall from within an L2 transaction. The L1StaticCall[] and L1StaticCallResults are exposed and utilized during the construction of the L2 block. As a result, the post-state of an L2 block now depends on additional inputs (L1StaticCall[] and L1StaticCallResults), rather than relying solely on the previous state, txList, and metadata.

This change breaks the deterministic nature of L2 blocks, which could lead to issues with preconfirmations.

Additionally, other challenges arise, such as the L2 block builder not being fully aware of which L1StaticCall[] are necessary, especially since some L1StaticCall executions might be conditional.


dantaik:
replaced by

Issue #18085, opened 2024-09-13
feat(taiko-client): support switching to use calldata when the blob fee is too expensive after the `Ontake` fork


davidtaikocha:

mskrzypkows:
I've created PR: https://github.com/taikoxyz/taiko-mono/pull/18548 it solves the switching and adds proposing only when it's profitable

PR #18548, opened 2024-12-05
feat(taiko-client): proposer profitability check

mskrzypkows:


packages/taiko-client/proposer/proposer.go
```diff
@@ -456,24 +451,121 @@ func (p *Proposer) ProposeTxListOntake(
...
 		log.Warn("Failed to build TaikoL1.proposeBlocksV2 transaction", "error", encoding.TryParsingCustomError(err))
 		return err
 	}
 
+	if p.checkProfitability {
+		profitable, err := p.isProfitable(txLists, cost)
+		if err != nil {
+			return err
+		}
+		if !profitable {
```
YoGhurt111, line 465: For this issue https://github.com/taikoxyz/taiko-mono/issues/18085, we want to choose a cheaper way to propose txs rather than stopping the proposal.

packages/taiko-client/proposer/proposer.go
```diff
@@ -413,28 +423,13 @@ func (p *Proposer) ProposeTxListOntake(
...
-
-		compressedTxListBytes, err := utils.Compress(txListBytes)
-		if err != nil {
-			return err
-		}
-
-		txListsBytesArray = append(txListsBytesArray, compressedTxListBytes)
-		txNums = append(txNums, len(txs))
-		totalTxs += len(txs)
+	txListsBytesArray, totalTxs, err := p.compressTxLists(txLists)
```
YoGhurt111, line 426: Using `txNums` would be better to know each L2 block size.

packages/taiko-client/proposer/proposer.go
```diff
@@ -456,24 +451,121 @@ func (p *Proposer) ProposeTxListOntake(
...
+			return nil
+		}
+	}
+
 	if err := p.sendTx(ctx, txCandidate); err != nil {
 		return err
 	}
 
-	log.Info("📝 Batch propose transactions succeeded", "txs", txNums)
+	log.Info("📝 Batch propose transactions succeeded", "txs", totalTxs)
```
YoGhurt111, line 475: Using `txNums` would be better to know each L2 block size.

packages/taiko-client/proposer/proposer.go
```diff
@@ -456,24 +451,121 @@ func (p *Proposer) ProposeTxListOntake(
...
+
+	if calldataTxCost.Cmp(totalBlobCost) > 0 {
+		return txBlob, totalBlobCost, nil
+	}
+
+	return txCallData, calldataTxCost, nil
+}
+
+// compressTxLists compresses transaction lists and returns compressed bytes array and transaction counts
+func (p *Proposer) compressTxLists(txLists []types.Transactions) ([][]byte, int, error) {
```
YoGhurt111, line 541: Using `txNums` would be better to know each L2 block size.

packages/taiko-client/proposer/proposer.go
YoGhurt111: > For this issue #18085, we want to choose a cheaper way to propose txs rather than stopping the proposal.

Ignore it.

packages/taiko-client/proposer/proposer_test.go
```diff
@@ -361,50 +361,72 @@ func (s *ProposerTestSuite) TestName() {
 }
 
 func (s *ProposerTestSuite) TestProposeOp() {
-	// Propose txs in L2 execution engine's mempool
-	sink := make(chan *bindings.TaikoL1ClientBlockProposed)
+	testCases := []struct {
```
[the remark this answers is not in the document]
mskrzypkows: Added `TestProposeOpWithProfitabilityCheck`

packages/taiko-client/proposer/proposer.go
```diff
@@ -518,3 +610,118 @@ func (p *Proposer) sendTx(ctx context.Context, txCandidate *txmgr.TxCandidate) e
...
+			multiplier := new(big.Int).Add(tx.GasTipCap(), baseFee)
+			gasConsumed := new(big.Int).Mul(multiplier, baseFee)
+			totalGasConsumed.Add(totalGasConsumed, gasConsumed)
+		}
+	}
+
+	return totalGasConsumed, nil
+}
+
+func get75PercentOf(num *big.Int) *big.Int {
```
YoGhurt111, line 648: The 75 percent of base fee sharing is the config in contract, let's use the chain config instead of a constant value.

packages/taiko-client/proposer/proposer.go
```diff
@@ -518,3 +610,118 @@ func (p *Proposer) sendTx(ctx context.Context, txCandidate *txmgr.TxCandidate) e
...
+}
+
+func (p *Proposer) getTransactionCost(txCandidate *txmgr.TxCandidate) (*big.Int, error) {
+	// Get the current L1 gas price
+	gasPrice, err := p.rpc.L1.SuggestGasPrice(p.ctx)
+	if err != nil {
+		return nil, fmt.Errorf("getTransactionCost: failed to get gas price: %w", err)
+	}
+
+	estimatedGasUsage, err := p.rpc.L1.EstimateGas(p.ctx, ethereum.CallMsg{
```
YoGhurt111, line 662: @davidtaikocha I remembered that `EstimateGas` for blob tx is not very accurate. Did I misremember?

YoGhurt111:
Let me fix lint later.

packages/taiko-client/proposer/proposer.go
```diff
@@ -458,12 +453,23 @@ func (p *Proposer) ProposeTxListOntake(
 		return errors.New("insufficient prover balance")
 	}
 
-	txCandidate, err := p.txBuilder.BuildOntake(ctx, txListsBytesArray)
+	txCandidate, cost, err := p.buildCheaperOnTakeTransaction(ctx, txListsBytesArray)
```
YoGhurt111, line 456: I saw it was always get cheaper txCandidate from `buildCheaperOnTakeTransaction`.
How about making this feature optional?

davidtaikocha:
After an internal discussion, we decided to only impl the calldata / blob switch part at first: https://github.com/taikoxyz/taiko-mono/pull/18690 , anyway thanks for the contribution :)

mskrzypkows:
OK, so will it be needed in the future, or don't you want this kind of profitability check?

PR #18564, opened 2024-12-08
chore(protocol): test bond deposit and withdrawal

xiaodino:


packages/protocol/test/layer1/based/TaikoL1_EtherAsBond.t.sol
```diff
@@ -101,28 +99,9 @@ contract TaikoL1Test_EtherAsBond is TaikoL1TestBase {
         taikoL1.depositBond{ value: 0 }(depositAmount);
     }
 
-    function test_withdraw_more_than_bond_balance() external {
```
xiaodino, line 104: It's not removed. The github diff looks misleading with function renaming.

packages/protocol/test/layer1/based/TaikoL1_BondToken.t.sol
```diff
@@ -0,0 +1,200 @@
...
+            pacayaForkHeight: 0
+        });
+    }
+
+    function setUpOnEthereum() internal override {
+        super.setUpOnEthereum();
+        bondToken = deployBondToken();
+    }
+
+    function test_deposit_withdraw() external {
```
dantaik, line 36: can you change all test nam to "test_taikoL1_" so we can use test name to filter and run all taikoL1 tests?

dantaik (approved):

packages/protocol/test/layer1/based/TaikoL1_BondToken.t.sol
xiaodino: Updated.

PR #18569, opened 2024-12-08
feat(protocol): allow block propose to pass a bytes32 input to L2 anchor

dantaik:
This feature allows the block proposer to pass any bytes32 data as input to L2, which can then be consumed by transactions within the same block.

It is a general-purpose feature, leaving the definition and use of the input data entirely up to the block proposer. If larger data is required, its hash can be passed as the input. The original data can then be included as call data in L2 transactions, enabling the verification of the data’s integrity against the input (the hash).


packages/protocol/contracts/layer1/based/ITaikoL1.sol
```diff
@@ -180,6 +182,7 @@ interface ITaikoL1 {
     error ArraySizesMismatch();
     error BlobIndexZero();
     error BlobNotFound();
+    error BlobIndexZero();
```
myronrotter, line 185: Already exists in l183
```suggestion
```

smtmfft (approved):

packages/protocol/contracts/layer2/based/TaikoL2.sol
```diff
@@ -131,11 +134,13 @@ contract TaikoL2 is EssentialContract, IBlockHash, TaikoL2Deprecated {
     /// transaction, and any subsequent calls will revert with L2_PUBLIC_INPUT_HASH_MISMATCH.
     /// @param _anchorBlockId The `anchorBlockId` value in this block's metadata.
     /// @param _anchorStateRoot The state root for the L1 block with id equals `_anchorBlockId`.
+    /// @param _anchorInput An arbitrary bytes32 input chosen by the block proposer.
     /// @param _parentGasUsed The gas used in the parent block.
     /// @param _baseFeeConfig The base fee configuration.
     function anchorV2(
```
davidtaikocha, line 140: Shall we rename this `anchorV3`? since its a break change in params.

dantaik: renamed anchorV2 to anchorV3

davidtaikocha (approved):

davidtaikocha (approved):

PR #18570, opened 2024-12-09
feat(protocol): propose a batch blocks conditionally

YoGhurt111:


dantaik (approved):

packages/protocol/contracts/layer1/provers/ProverSet.sol
```diff
@@ -94,18 +94,22 @@ contract ProverSet is EssentialContract, IERC1271 {
...
     )
         external
         onlyProver
     {
         ITaikoL1 taiko = ITaikoL1(taikoL1());
-        // Ensure this block is the first block proposed in the current L1 block.
-        require(taiko.lastProposedIn() != block.number, NOT_FIRST_PROPOSAL());
-        taiko.proposeBlockV2(_params, _txList);
+
+        uint256 lastProposedBlock = taiko.lastProposedIn();
```
dantaik, line 108: I think the logic here is the same right? If so, I think we should revert changes in this function.

YoGhurt111: Yes, this change is made by @swarna1101 , maybe he can give some reasons.

> I think the logic here is the same right? If so, I think we should revert changes in this function.

davidtaikocha (approved):

davidtaikocha (approved):

dantaik (approved):

PR #18582, opened 2024-12-14
chore(protocol): fix issues in test:coverage

xiaodino:


packages/protocol/foundry.toml
```diff
@@ -67,6 +64,7 @@ test = "test/layer2"
 script = "script/layer2"
 out = "out/layer2"
 evm_version = "shanghai"
+fuzz.runs = 200
```
dantaik, line 67: should we also add this config to other prifiles?

packages/protocol/test/layer2/LibEIP1559.t.sol
```diff
@@ -98,23 +98,23 @@ contract TestLibEIP1559 is Layer2Test {
         assertEq(baselineBasefee, basefee);
     }
 
-    /// forge-config: layer2.fuzz.runs = 1000
-    /// forge-config: layer2.fuzz.show-logs = true
+    /// forge-config: layer_2.fuzz.runs = 1000
```
dantaik, line 101: why change to "layer_2"?

xiaodino: forge-config will convert layer2 to layer_2 in *.t.sol files when fuzz.runs is used.

xiaodino: Updated to `forge-config: default.fuzz_runs = 2000`

packages/protocol/foundry.toml
xiaodino: Updated to 
```
[profile.default]
fuzz_runs = 200
```

packages/protocol/test/layer2/LibEIP1559.t.sol
```diff
@@ -98,23 +98,23 @@ contract TestLibEIP1559 is Layer2Test {
         assertEq(baselineBasefee, basefee);
     }
 
-    /// forge-config: layer2.fuzz.runs = 1000
-    /// forge-config: layer2.fuzz.show-logs = true
+    /// forge-config: default.fuzz_runs = 1000
```
dantaik, line 101: I think we should also change `layer_2.fuzz_runs` to `default.fuzz_runs`, no?

xiaodino: Good point. Rename both `layer2.fuzz.runs` and `layer2.fuzz.show-logs` to `default.fuzz_runs` and `default.fuzz_runs_show_logs`

dantaik (approved):

dantaik (approved):

PR #18587, opened 2024-12-15
feat(protocol): allow same-slot L1-to-L2 signalling

dantaik:
Implementing https://ethresear.ch/t/same-slot-l1-l2-message-passing/21186


davidtaikocha (approved):

xiaodino (approved):

packages/protocol/contracts/layer1/based/TaikoL1.sol
```diff
@@ -591,6 +594,17 @@ abstract contract TaikoL1 is EssentialContract, ITaikoL1 {
...
         }
+
+        if (_params.signalSlots.length != 0) {
+            require(_params.signalSlots.length <= _maxSignalsToReceive, TooManySignals());
+
+            ISignalService signalService =
+                ISignalService(resolve(LibStrings.B_SIGNAL_SERVICE, false));
+
+            for (uint256 i; i < _params.signalSlots.length; ++i) {
+                require(signalService.isSignalSent(_params.signalSlots[i]), SignalNotSent());
```
AnshuJalan, line 605: Would it make sense to remove these force-pushed signals from the signal service storage giving a gas refund to the block proposer? I think that might impact message recall functionality in the bridge contract.
The refund was reduce the gas consumption to 200 from the current 2100 per signal read.

PR #18588, opened 2024-12-15
Update README.md

ahmeteme:
new updated readme.md


swarna1101:
We generally dont update the README unless something important needs to be added. I dont think this PR adds anything important, rather just enhancement to the existing one.

PR #18590, opened 2024-12-16
feat(taiko-client): propose blocks with revert protection

YoGhurt111:


packages/taiko-client/proposer/transaction_builder/blob.go
```diff
@@ -190,7 +190,7 @@ func (b *BlobTransactionBuilder) BuildOntake(
 	}
 	txListArray := make([][]byte, len(encodedParamsArray))
 	if b.proverSetAddress != rpc.ZeroAddress {
-		data, err = encoding.ProverSetABI.Pack("proposeBlocksV2", encodedParamsArray, txListArray)
+		data, err = encoding.ProverSetABI.Pack("proposeBlocksV2Conditionally", encodedParamsArray, txListArray)
```
davidtaikocha, line 193: Shall we make this feature as a new flag?

YoGhurt111: I think this is feature is benefit for all users. Without this feature, user would propose empty blocks and verify blocks with much gas and get no reward. If all users apply this feature, it will cost little gas and revert tx even thought without the support of revert protection from endpoint.

davidtaikocha: Emm but then it will be a force upgrade for all community proposers when we release a new version of client?

YoGhurt111: OK, will change it as an optional flag.

YoGhurt111: Done

davidtaikocha (approved):

PR #18618, opened 2024-12-19
fix: typos in documentation files

vtjl10:
Corrected `provide` to `provided`


davidtaikocha:
We generally dont update that third party doc unless something important needs to be added. I dont think this PR adds anything important.

PR #18658, opened 2024-12-27
chore(repo): bump frontend deps

dionysuzx:
I discovered it's a lot easier for maintainers to just do a `pnpm up -r` once a week (and let the weekly dependabot PRs be a "reminder"). This should close most if not all dependabot javascript PRs.

This doesn't bump any major versions, so it shouldn't break things but of course it could. So some testing, at least a sanity check would be good.

EDIT: Looks like some failing jobs that I need to address.


dionysuzx:
i'm going to try this again in a more step-by-step fashion.

PR #18664, opened 2024-12-27
fix(taiko-client): fix `GetBasefeeV2` usage

YoGhurt111:


packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -933,6 +933,19 @@ func (c *Client) calculateBaseFeeOntake(
 		uint32(l2Head.GasUsed),
 	)
 	if err != nil {
+		// If the `calculateBaseFee()` method is deprecated, we will use the new method `getBasefeeV2()`
+		// to calculate the base fee.
+		if strings.Contains(encoding.TryParsingCustomError(err).Error(), "L2_DEPRECATED_METHOD") {
+			baseFeeInfo, err := c.TaikoL2.GetBasefeeV2(
+				&bind.CallOpts{BlockNumber: l2Head.Number, Context: ctx},
+				0,
```
davidtaikocha, line 941:
```suggestion
				uint32(l2Head.GasUsed),
```
can't use `0` here neither?

YoGhurt111: I think it is the same logic as `GetBaseFee`.
https://github.com/taikoxyz/taiko-mono/blob/0f9bdde13bb8a321e1e0bd78cc222406b401851c/packages/taiko-client/pkg/rpc/methods.go#L297

YoGhurt111: Fixed

davidtaikocha (approved):

mask-pp (approved):

PR #18675, opened 2024-12-30
test(taiko-client): test generating signal proof

YoGhurt111:
I try to generate a signal proof for this tx and recall this message. https://holesky.etherscan.io/tx/0x086729ca21c2db5714560943a6522362c837c9cf6424dfcc7b8f15f132b7f934

So I wrote a test for generating the proof but it failed.
I only setup some necessary config for the message processor but not sure it is correct.
Could you have a look at this test? 
Your help is much appreciated！
@cyberhorsey @xiaodino


xiaodino (changes requested):
Not merge yet. Need some testing

PR #18686, opened 2025-01-01
fix(protocol): add timestamp as a new parameter to getBasefeeV2

YoGhurt111:
Not sure if the 1s diff with `block.timestamp` will have an effect on base fee.


packages/protocol/contracts/layer2/based/TaikoL2.sol
```diff
@@ -187,6 +187,7 @@ contract TaikoL2 is EssentialContract, IBlockHash, TaikoL2Deprecated {
     /// @return newGasExcess_ The new gas excess value.
     function getBasefeeV2(
         uint32 _parentGasUsed,
+        uint64 _curTimestamp,
```
dantaik, line 190:
```suggestion
        uint64 _blockTimestamp,
```

dantaik (approved):

davidtaikocha (approved):

PR #18718, opened 2025-01-06
chore(taiko-client): improve prover logs

davidtaikocha:


mask-pp (approved):

packages/taiko-client/prover/proof_producer/sgx_producer.go
```diff
@@ -137,6 +138,10 @@ func (s *SGXProofProducer) Aggregate(
 ) (*BatchProofs, error) {
 	log.Info(
 		"Aggregate sgx batch proofs from raiko-host service",
+		"batchSize", len(items),
+		"firstID", items[0].BlockID,
```
YoGhurt111, line 142: The first ID in the buffer is propbably not the smallest block id.

YoGhurt111: Which blockId first generates proof, which goes into the buffer first

YoGhurt111: OK

YoGhurt111 (approved):

PR #18722, opened 2025-01-06
chore(repo): update dependabot config

dionysuzx:
## Summary of changes
- Dependabot now only opens **three PRs max**, once a month for Go, JavaScript, and GitHub actions.
- It does major/minor/patch (ALL); if you want to exclude a dependency (eg. breaking change); you can do that explicitly and open a GitHub issue if needed to triage for later.
- I added @davidtaikocha, @RogerLamTd, @KorbinianK, and @bearni95 as dependabot owners here. Anyone works, but I think it's important the reviewers here stay on top of it.
- Turned off stale issues workflow, we don't need this anymore. This was from a time we had a lot of spam. Now it can get in the way of important tracking. Maintainers / contributors should close unnecessary issues.

I don't think we should automatically close these. It should be less noisy now, as you can see in the config, at the very most there will only be 3 PRs open once a month, which is quite forgiving. I think the bigger issue is we need to actually merge in these dependency updates for JavaScript projects.


dionysuzx:
@davidtaikocha feel free to update the interval for Go dependencies. In my projects I usually like to do once a week, but maybe once a month is also fine.

davidtaikocha (approved):

dantaik (approved):

PR #18737, opened 2025-01-09
fix(taiko-client): revert triggered relate changes

mask-pp:


davidtaikocha (approved):

packages/taiko-client/driver/chain_syncer/beaconsync/progress_tracker.go
```diff
@@ -157,6 +157,7 @@ func (t *SyncProgressTracker) ClearMeta() {
 
 	log.Debug("Clear sync progress tracker meta")
 
+	t.triggered = false
```
YoGhurt111, line 160: Why do we add it back?

mask-pp: It's revert, related to too many places.

YoGhurt111 (approved):

PR #18739, opened 2025-01-09
fix(taiko-client): check the `blockID` of the last verified block before using it as `FinalizedBlockHash`

davidtaikocha:
ref:
telegram-cloud-photo-size-5-6053379241615147620-w

Error in `taiko-geth`:
```
WARN [01-09|02:00:51.973] Served engine_forkchoiceUpdatedV2        conn=[::1]:51920 reqid=324196 duration=8.170036ms err="Invalid forkchoice state" errdata="{\"err\":\"final block not available in database\"}"
```

### DIFF
--- a/packages/taiko-client/driver/chain_syncer/blob/syncer.go
+++ b/packages/taiko-client/driver/chain_syncer/blob/syncer.go
@@ -409,23 +409,12 @@ func (s *Syncer) insertNewHead(
 	}
 
 	var lastVerifiedBlockHash common.Hash
-	if lastVerifiedBlockHash, err = s.rpc.GetLastVerifiedBlockHash(ctx); err != nil {
-		log.Debug("Failed to fetch last verified block hash", "error", err)
-
-		stateVars, err := s.rpc.GetProtocolStateVariables(&bind.CallOpts{Context: ctx})
-		if err != nil {
-			return nil, fmt.Errorf("failed to fetch protocol state variables: %w", err)
-		}
-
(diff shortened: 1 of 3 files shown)


packages/taiko-client/driver/chain_syncer/blob/syncer.go
```diff
@@ -409,23 +409,12 @@ func (s *Syncer) insertNewHead(
...
-			return nil, fmt.Errorf("failed to fetch last verified block: %w", err)
-		}
-
-		lastVerifiedBlockHash = lastVerifiedBlockHeader.Hash()
+	lastVerifiedBlockInfo, err := s.rpc.GetLastVerifiedBlock(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch last verified block: %w", err)
+	}
+	if payload.Number > lastVerifiedBlockInfo.BlockId {
+		lastVerifiedBlockHash = lastVerifiedBlockInfo.BlockHash
```
YoGhurt111, line 417: `lastVerifiedBlockHash` would be an empty hash if this conditiona is not matched

davidtaikocha: Yes thats what we need

YoGhurt111 (approved):

packages/taiko-client/driver/chain_syncer/blob/syncer.go
mask-pp: It's not normal if the condition is not matched.

mask-pp (approved):

PR #18742, opened 2025-01-09
fix(taiko-client): make sure beacon sync be executed exactly once

mask-pp:


davidtaikocha:
I think we still need to trigger beacon sync more than once based on the current impl.

PR #18743, opened 2025-01-09
feat(protocol): support smaller more frequent L2 blocks (replacing soft blocks)

dantaik:
## Overview  
This PR introduces support for smaller and more frequent L2 blocks, replacing the previously planned soft-blocks. The primary motivation for this change stems from @cyberhorsey's discovery that most existing Ethereum SDKs and tools (e.g., MetaMask) do not handle frequent block reorganizations gracefully. Instead of investing effort into improving these libraries and tools, adopting smaller and more frequent blocks appears to be the most pragmatic solution.

## Key Changes  
In this PR, we introduce the concept of a **"batch"**, which represents a collection of Taiko blocks. A batch serves as the smallest unit for block proposal, block proving, and cross-chain data synchronization. Block header data are derived from the same batch metadata to minimize on-chain footprint per block. This means that, given the same set of transactions within a batch, splitting them into 10 blocks versus 20 blocks does not significantly impact L1 gas costs—though there are still minor differences.  

However, creating more L2 blocks from the same set of transactions does increase gas costs for batch proposals (slightly) and proving costs (more noticeable). Specifically, proofs are now aggregated in two stages:  
1. From block proofs to batch proofs.  
2. From batch proofs to multi-batch proofs.  

More L2 blocks result in higher costs for the first level of aggregation.  

Additionally, the protocol does not support proposing multiple batches in a single transaction. Instead, it leverages multiple blobs within a single `proposeBatch` transaction. As a result, proposers are encouraged to propose fewer and larger batches whenever possible to optimize efficiency.  

## Review Tips

This is a large PR, but your primary focus should be on **TaikoInbox.sol**. The other changes are secondary and result from ripple effects.

## Todos  
- [ ] Consult Nethermind about this change and understand the impact on preconf
- [x] Remove soft block API - https://github.com/taikoxyz/taiko-mono/pull/18744
- [x] Need to verify "timestamp + timeThift" is within valid range
- [x] Document how block difficulty should be calculated in client and prover
- [x] Fix tests  
- [x] add more tests for verifying lastBlockId


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -90,94 +89,96 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, ITaiko {
...
+            parentBatch
+        );
+
+        // This section constructs the metadata for the proposed batch, which is crucial for
+        // nodes/clients to process the batch. The metadata itself is not stored on-chain;
+        // instead, only its hash is kept.
+        // The metadata must be supplied as calldata prior to proving the batch, enabling the
+        // computation and verification of its integrity through the comparison of the metahash.
+        unchecked {
+            meta_ = BatchMetadata({
```
davidtaikocha, line 122: Shall we comment how to calculate `block.difficulty` here?

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -17,76 +17,80 @@ import "src/shared/based/LibSharedData.sol";
...
     }
 
-    struct BlockMetadataV3 {
-        bytes32 difficulty;
+    struct BatchMetadata {
         bytes32 txListHash;
         bytes32 extraData;
         address coinbase;
-        uint64 blockId;
+        uint64 batchId;
```
YoGhurt111, line 41: In this design, when we propose blocks on L2,  do we propose the batch or the blocks in the batch?

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -90,94 +89,89 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, ITaiko {
...
-            }
-        } // end of for-loop
+        unchecked {
+            stats2.numBatches += 1;
+            stats2.lastProposedIn = uint56(block.number);
+        }
 
-        _debitBond(_proposer, config.livenessBond * _paramsArray.length);
-        emit BlocksProposedV3(metas_, calldataUsed, _txList);
+        _debitBond(_proposer, config.livenessBond);
```
davidtaikocha, line 162: So rn for each batch, we only charge `config.livenessBond` once, is it by design?

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
cyberhorsey: We propose a batch. We move holistically across the stack from block based to batch based in this design, from proposing to proving.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: Yes, as a batch is the smallest unit to be proved, so we only debit liveness bond once per batch.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -43,37 +45,34 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, ITaiko {
         __Taiko_init(_owner, _rollupResolver, _genesisBlockHash);
```
davidtaikocha, line 45: Client still needs `_state.slotA.genesisHeight` (https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/based/LibUtils.sol#L40) to know where to start scanning the L1 chain, seems its removed from the new `state` definition?

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -90,94 +89,100 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, ITaiko {
...
+            livenessBond: config.livenessBond,
+            proposedAt: uint64(block.timestamp),
+            proposedIn: uint64(block.number),
+            txListOffset: _batchParams.txListOffset,
+            txListSize: _batchParams.txListSize,
+            numBlobs: calldataUsed ? 0 : _batchParams.numBlobs,
+            anchorBlockId: updatedParams.anchorBlockId,
+            anchorBlockHash: blockhash(updatedParams.anchorBlockId),
+            signalSlots: _batchParams.signalSlots,
+            blocks: _batchParams.blocks,
```
davidtaikocha, line 138: When calldata is used, shall we make sure that only one block will be proposed?

dantaik: No, using calldata or blobs, multiple blocks can be proposed per batch.

PR #18745, opened 2025-01-10
feat(taiko-client): only check and trigger P2P sync progress right after starting

davidtaikocha:


mask-pp (approved):

packages/taiko-client/driver/chain_syncer/chain_syncer.go
```diff
@@ -88,6 +91,17 @@ func (s *L2ChainSyncer) Sync() error {
 	if err != nil {
 		return err
 	}
+
+	// If the L2 execution engine's chain is head of the block head to sync at the start,
+	// and the --p2p.sync flag is enabled, we should mark the beacon sync progress as finished.
+	if !s.p2pSyncChecked {
```
mask-pp, line 97: Just use `s.progressTracker.MarkFinished()` can be simpler?

YoGhurt111 (approved):

PR #18751, opened 2025-01-12
feat(protocol): redo ForkMananger (now ForkRouter)

dantaik:
Previous design use function selector to route between 2 fork implementations. The new idea is to query the new fork if it is active, if so, route to the new fork, otherwise, route to the old fork.


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -372,6 +373,10 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, ITaiko {
         return true;
     }
 
+    function isForkActive() external view override returns (bool) {
+        return state.stats2.numBlocks >= getConfigV3().forkHeights.pacaya;
```
davidtaikocha, line 377: Looks like this value will be changed to `numBatches`? Then we can't know the exact block numbers in protocol anymore?

dantaik: This is based on the previous PR, once the batch PR is merged, this var will be renamed.

davidtaikocha (approved):

PR #18753, opened 2025-01-13
fix(taiko-client): improve blob sync progress

mask-pp:


packages/taiko-client/bindings/metadata/interface.go
```diff
@@ -37,4 +37,5 @@ type TaikoBlockMetaData interface {
 	GetTxIndex() uint
 	GetTxHash() common.Hash
 	IsOntakeBlock() bool
+	GetL1BlockHash() common.Hash
```
YoGhurt111, line 40: Why do we change the bindings?

packages/taiko-client/driver/chain_syncer/blob/syncer.go
```diff
@@ -428,6 +432,13 @@ func (s *Syncer) insertNewHead(
 		return nil, fmt.Errorf("unexpected ForkchoiceUpdate response status: %s", fcRes.PayloadStatus.Status)
 	}
 
+	// Record the latest L1 event header.
+	l1Header, err := s.rpc.L1.HeaderByHash(ctx, meta.GetL1BlockHash())
+	if err != nil {
+		log.Debug("Failed to fetch L2 block header", "blockID", meta.GetBlockID(), "error", err)
+	}
+	s.latestL1EventHeader = l1Header
```
YoGhurt111, line 440: Do we have some mechanisms to check whether latestL1EventHeader is correct if reorg happened?

packages/taiko-client/bindings/metadata/interface.go
mask-pp: It exposes the l1 block hash, this PR needs it.

packages/taiko-client/driver/chain_syncer/blob/syncer.go
mask-pp: The code executed after reorg logic, so we don't need to care about it?

packages/taiko-client/bindings/metadata/interface.go
YoGhurt111: I think we already have `GetRawBlockHash` to expose the l1 block hash, so it is not need?

packages/taiko-client/driver/state/state.go
```diff
@@ -149,7 +149,7 @@ func (s *State) eventLoop(ctx context.Context) {
 			s.setHeadBlockID(e.BlockId)
 		case e := <-transitionProvedCh:
 			log.Info(
-				"✅ Transition proven",
+				"✅ transitionProved proven",
```
YoGhurt111, line 152: I think we don't need to change the log here.

packages/taiko-client/driver/chain_syncer/blob/syncer.go
```diff
@@ -428,6 +432,13 @@ func (s *Syncer) insertNewHead(
 		return nil, fmt.Errorf("unexpected ForkchoiceUpdate response status: %s", fcRes.PayloadStatus.Status)
 	}
 
+	// Record the latest L1 event header.
+	l1Header, err := s.rpc.L1.HeaderByHash(ctx, meta.GetL1BlockHash())
+	if err != nil {
+		log.Debug("Failed to fetch L2 block header", "blockID", meta.GetBlockID(), "error", err)
```
YoGhurt111, line 438:
```suggestion
		log.Debug("Failed to fetch L1 block header", "blockID", meta.GetRawBlockHeight(), "error", err)
```

packages/taiko-client/driver/chain_syncer/blob/syncer.go
YoGhurt111: I'm not sure, let's test it first

packages/taiko-client/driver/state/state.go
mask-pp: This log content will be confused with L160.

davidtaikocha: The English here is also confusing, i also prefer the previous log, and after onTake fork, there won't be any new event in `transitionProvedCh`

packages/taiko-client/bindings/metadata/interface.go
davidtaikocha: > I think we already have `GetRawBlockHash` to expose the l1 block hash, so it is not need?

I think so

packages/taiko-client/driver/chain_syncer/blob/syncer.go
mask-pp: I will add more tests later.

PR #18757, opened 2025-01-13
fix(taiko-client): check inner iterator errors in `BlockProposedIterator`

davidtaikocha:


packages/taiko-client/pkg/chain_iterator/event_iterator/block_proposed_iterator.go
```diff
@@ -170,6 +140,11 @@ func assembleBlockProposedIteratorCallback(
 			updateCurrentFunc(current)
 		}
 
+		// Check iterator errors.
+		if iterOntake.Error() != nil {
```
mask-pp, line 144: Directly return `iterOntake.Error()`?

davidtaikocha: Changed

mask-pp (approved):

YoGhurt111 (approved):

PR #18770, opened 2025-01-15
feat(protocol): encode proposer address in L2 block header

dantaik:
Blocks, before being proposed on L1, are now shared by the current proposer in Taiko's P2P network. Prior to this PR, there was no way to verify whether an L2 block originated from the current proposer simply by examining the block itself.

**However, I realized that this PR does not fully address the issue: a malicious node could still broadcast a block using the current proposer's address in the header.**

As a result, this PR will not be merged.

@davidtaikocha It seems in geth, the block must be signed by the preconfer before being broadcasted to p2p.


cyberhorsey:
Yes, the block must be signed and then the pubkey extracted from the message pulled from the p2p then verified that it is signed by a valid preconfer from the registry contract.

PR #18784, opened 2025-01-16
fix(protocol): fix bug about totalShift check

mask-pp:


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -676,7 +676,7 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, ITaiko, IFork {
                 totalShift += _params.blocks[i].timeShift;
             }
 
-            require(lastBlockTimestamp_ >= totalShift, TimestampTooSmall());
+            require(lastBlockTimestamp_ + totalShift >= block.timestamp, TimestampTooSmall());
```
dantaik, line 679: here we need to check:
- the first block's timestamp >=  block.timestamp - _maxAnchorHeightOffset * LibNetwork.ETHEREUM_BLOCK_TIME
- the last block's timestamp <= block.timestamp.

mask-pp: Ok, I see. And changed.

PR #18786, opened 2025-01-17
chore(taiko-client): improve `BlobFetcher`

davidtaikocha:


packages/taiko-client/driver/txlist_fetcher/blob.go
```diff
@@ -62,11 +64,27 @@ func (d *BlobFetcher) Fetch(
...
+			return b, nil
 		}
 	}
 
 	return nil, pkg.ErrSidecarNotFound
 }
+
+// sliceTxList returns the sliced txList bytes from the given offset and length.
+func sliceTxList(id *big.Int, b []byte, offset, length uint32) ([]byte, error) {
+	if offset+length > uint32(len(b)) {
```
mask-pp, line 84:
```suggestion
	if length > uint32(len(b[offset:])) {
```

mask-pp (approved):

YoGhurt111 (approved):

PR #18788, opened 2025-01-17
feat(protocol): add error selectors as comments for each error

YoGhurt111:


dantaik:
Why do we need to do this?

YoGhurt111:
> Why do we need to do this?

While testing contracts on devnet, we can quickly pinpoint the specific error type through error selectors.

dantaik:
> > Why do we need to do this?
> 
> While testing contracts on devnet, we can quickly pinpoint the specific error type through error selectors.

But this means when ever we add a new error, we have to manually compute its value. which is very inconvinient.

YoGhurt111:
> > > Why do we need to do this?
> > 
> > While testing contracts on devnet, we can quickly pinpoint the specific error type through error selectors.
> 
> But this means when ever we add a new error, we have to manually compute its value. which is very inconvinient.

I can submit a pr to compute the value when add a new error, but if you feel it is not suitable, can close this pr.

PR #18791, opened 2025-01-19
feat(protocol): add preconfirmation whitelist and router

AnshuJalan:
Added a preconfirmation whitelist and router. I have removed the old preconf contracts based on eigenlayer.


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -0,0 +1,40 @@
...
+    {
+        // Sender must be the selected operator for the epoch
+        address selectedOperator =
+            IPreconfWhitelist(resolve(LibStrings.B_PRECONF_WHITELIST, false)).getOperatorForEpoch();
+        require(msg.sender == selectedOperator, NOT_THE_OPERATOR());
+
+        // Force set the `proposer` field in the batch params to be the sender
+        ITaikoInbox.BatchParams memory batchParams =
+            abi.decode(_batchParams, (ITaikoInbox.BatchParams));
+        batchParams.proposer = msg.sender;
```
dantaik, line 34: Do you think we can simply check the proposer is `msg.sender` to avoid re-encoding `batchParams`?

dantaik: Or we can change `ITaikoInbox(taikoInbox).proposeBatch` to return the proposer address then after the `proposeBatch` call, we check the returned proposer is msg.sender? This will avoid the decoding in this method.

packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import "src/layer1/based/ITaikoInbox.sol";
+
+/// @title IPreconfRouter
+/// @custom:security-contact security@taiko.xyz
+interface IPreconfRouter {
+    error NOT_THE_OPERATOR();
```
dantaik, line 9:
```suggestion
    error NotTheOperator();
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -0,0 +1,34 @@
...
+    /// @param _operatorId The ID of the operator to be removed.
+    /// @dev Only callable by the owner or an authorized address.
+    /// @dev Reverts if the operator ID does not exist.
+    function removeOperator(uint256 _operatorId) external;
+
+    /// @notice Retrieves the address of the operator for the current epoch.
+    /// @dev Uses the beacon block root of the first block in the last epoch as the source
+    ///      of randomness.
+    /// @return The address of the operator.
+    function getOperatorForEpoch() external view returns (address);
```
dantaik, line 33: Do we need to support the next epoch?

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -0,0 +1,76 @@
...
+
+/// @title PreconfWhitelist
+/// @custom:security-contact security@taiko.xyz
+contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
+    // Tracks the total number of operators in the whitelist
+    uint256 public operatorCount;
+
+    // Maps operator index to their corresponding operator addresses
+    mapping(uint256 operatorIndex => address operator) public operatorIndexToOperator;
+
```
dantaik, line 18:
```suggestion
uint256[49] private __gap;
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -0,0 +1,40 @@
...
+import "src/shared/libs/LibStrings.sol";
+import "src/shared/common/EssentialContract.sol";
+
+/// @title PreconfRouter
+/// @custom:security-contact security@taiko.xyz
+contract PreconfRouter is EssentialContract, IPreconfRouter {
+    function init(address _owner, address _sharedResolver) external initializer {
+        __Essential_init(_owner, _sharedResolver);
+    }
+
```
dantaik, line 16:
```suggestion
uint256[50] private __gap?
```

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -0,0 +1,76 @@
...
+        address removedOperator = operatorIndexToOperator[_operatorIndex];
+
+        unchecked {
+            // Bring the last operator to this operator's index
+            address lastOperator = operatorIndexToOperator[_operatorCount - 1];
+            operatorIndexToOperator[_operatorIndex] = lastOperator;
+
+            // Remove the now duplicate operator at the last index
+            delete operatorIndexToOperator[_operatorCount - 1];
```
dantaik, line 54: Maybe we can delete this `delete` statement so removing one then adding one will end up wth a non-zero-to-non-zero SSTORE op?

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    // Maps operator index to their corresponding operator addresses
+    mapping(uint256 operatorIndex => address operator) public operatorIndexToOperator;
+
+    function init(address _owner, address _sharedResolver) external initializer {
+        __Essential_init(_owner, _sharedResolver);
+    }
+
+    /// @inheritdoc IPreconfWhitelist
+    function addOperator(address _operatorAddress)
```
dantaik, line 24: check _operatorAddress non zero?

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
AnshuJalan: I think we are already getting the metadata from taiko inbox along with the proposer, so I can check that directly so we don't have to encode or decode at all.

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
AnshuJalan: Not for whitelist.

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -0,0 +1,34 @@
...
+    /// @param operator The address of the operator that was removed.
+    event OperatorRemoved(address indexed operator);
+
+    error InvalidOperatorIndex();
+    error InvalidOperatorCount();
+
+    /// @notice Adds a new operator to the whitelist.
+    /// @param _operatorAddress The address of the operator to be added.
+    /// @dev Only callable by the owner or an authorized address.
+    function addOperator(address _operatorAddress) external;
```
dantaik, line 21: In the context of preconfirmation where blocks are constructed and shared within P2P network, to ensure these blocks originate from the current preconfirmer, the preconfirmer must sign the block's header. However, if the operator is a contract address, this would require a network operation to perform an `IERC1271` `isValidSignature` check. 

To optimize this process, it would be ideal for the operator to be represented as a pair of addresses `(preconfer, blockSigner)`, where:
- `preconfer` is the preconfirmer address responsible for calling `proposePreconfedBlocks`.
- `blockSigner` is the address used to sign block headers for preconfirmed blocks.

Let me know if you have alternative suggestions or additional thoughts!

dantaik: I initially considered modifying the protocol to include a `blockSigner` field in the metadata. However, this approach wouldn't work because the block is proposed *after* the shared block is validated in the P2P network. This means the `blockSigner` address must be known and available *early*, not just when a block or batch is proposed.

linoscope: > However, if the operator is a contract address, this would require a network operation to perform an IERC1271 isValidSignature check.

In what case would the operator want to be a contract address?

dantaik (approved):

packages/protocol/test/layer1/preconf/router/RouterTest.t.sol
```diff
@@ -0,0 +1,146 @@
...
+        address beaconBlockRootContract = LibPreconfConstants.getBeaconBlockRootContract();
+        vm.etch(beaconBlockRootContract, address(mockBeacon).code);
+        MockBeaconBlockRoot(payable(beaconBlockRootContract)).set(
+            epochOneStart + LibPreconfConstants.SECONDS_IN_SLOT, mockRoot
+        );
+
+        // Warp to arbitrary slot in epoch 2
+        vm.warp(epochTwoStart + 2 * LibPreconfConstants.SECONDS_IN_SLOT);
+
+        // Prank as David (not the selected oeprator) and propose blocks
```
dantaik, line 89:
```suggestion
        // Prank as David (not the selected operator) and propose blocks
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
```diff
@@ -0,0 +1,26 @@
...
+    error ProposerIsNotTheSender();
+
+    /// @notice Proposes a batch of blocks that have been preconfed.
+    /// @dev This function only accepts batches from an operator selected to preconf in a particular
+    ///      slot or epoch and routes that batch to the TaikoInbox.
+    /// @param _params ABI-encoded parameters for the preconfing operation.
+    /// @param _batchParams ABI-encoded parameters specific to the batch.
+    /// @param _batchTxList The transaction list associated to the batch.
+    /// @return meta_ BatchMetadata containing metadata about the proposed batch.
+    function proposePreconfedBlocks(
```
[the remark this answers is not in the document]
AnshuJalan: We want to keep a different name from the function in `TaikoInbox` since this is specifically for preconfed blocks and only called by authorised operators.

PR #18796, opened 2025-01-20
feat(protocol): fix the deployment script

YoGhurt111:


packages/protocol/script/layer1/based/DeployProtocolOnL1.s.sol
```diff
@@ -235,15 +249,36 @@ contract DeployProtocolOnL1 is DeployCapability {
                 TaikoInbox.init, (owner, rollupResolver, vm.envBytes32("L2_GENESIS_HASH"))
             )
         });
+        addressNotNull(vm.envAddress("TAIKO_INBOX"), "TAIKO_INBOX");
+        address oldFork = vm.envAddress("TAIKO_INBOX");
```
davidtaikocha, line 253: Let's call it `ONTAKE_TAIKO_INBOX` ?

YoGhurt111: How about `OLD_FORK_TAIKO_INBOX`?

davidtaikocha: Sounds good to me

YoGhurt111: For future fork upgrades, we can keep the variable name unchanged and simply update the version number.

davidtaikocha (approved):

PR #18800, opened 2025-01-20
feat(protocol): make resolver an Immutable

dantaik:
This PR makes the `resolver` address **immutable** to optimize gas usage by eliminating `SLOAD` operations. Two approaches are proposed:

- **Approach 1**: Deploy the resolver behind a proxy, keeping it upgradable.
- **Approach 2**: Use a non-upgradable resolver, requiring contract upgrades to change the resolver.

This is a big PR but there are only a few changes that you need to review (marked by me as "A change worth noticing."), all other changes are ripple effects.

@davidtaikocha, let’s proceed with Approach 1 for the Pacaya fork then use Approach 2 later.

- [x] check storage layout


packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -122,7 +122,7 @@ interface ITaikoInbox {
         /// @notice The chain ID of the network where Taiko contracts are deployed.
         uint64 chainId;
         /// @notice The maximum number of verifications allowed when a batch is proposed or proved.
-        uint64 maxBatchProposals;
+        uint64 maxUnverifiedBatches;
```
dantaik, line 125: worth a look

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -70,7 +65,7 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, ITaiko, IFork {
 
         unchecked {
             require(
-                stats2.numBatches < stats2.lastVerifiedBatchId + config.maxBatchProposals,
+                stats2.numBatches <= stats2.lastVerifiedBatchId + config.maxUnverifiedBatches,
```
dantaik, line 68: @mask-pp please take a look

davidtaikocha (approved):

PR #18616, opened 2024-12-19
feat(protocol): add solver support to enable fast bridging of erc20 tokens from L2 to L1

AnshuJalan:
This PR adds an intent-solver like functionality to process fast bridging of erc20 tokens from L2 to L1, without the need to wait for chain sync.

The inspiration is taken from: https://ethresear.ch/t/fast-and-slow-l2-l1-withdrawals/21161

PR #18805, opened 2025-01-21
feat(protocol): add solver support for l2 to l1 eth bridging

AnshuJalan:
This is an extension of #18616 for L2 -> L1 FnS ETH bridging.

- As of now, Ether bridging is facilitated directly via `Bridge.sol`. Thus, the Bridge contract itself acts as the “ether vault”.
- We might as well modify `Bridge.sol` to incorporate a solver condition, but that would fundamentally change the core of the message passing system. Therefore, a better alternative implemented in this PR is a separate `EtherBridgeWrapper.sol` that handles just the “Eth in” and “Eth out” part of bridging by including a solver condition, leaving the existing functionality of `Bridge.sol` and surrounding infra untouched.
- `Bridge.sol` can still be used for standard slow withdrawals.


dantaik:
@AnshuJalan I may need a few days on something else if that's OK.

AnshuJalan:
> @AnshuJalan I may need a few days on something else if that's OK.

Sure

packages/protocol/contracts/shared/bridge/EtherBridgeWrapper.sol
```diff
@@ -0,0 +1,235 @@
...
+import "../../shared/based/ITaiko.sol";
+import "src/layer1/based/ITaikoInbox.sol";
+import "src/shared/libs/LibAddress.sol";
+import "src/shared/common/EssentialContract.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+contract EtherBridgeWrapper is EssentialContract {
+    using Address for address;
+    using LibAddress for address;
+
```
dantaik, line 14: Lets reserve a gap of 50 slots here, just in case.

packages/protocol/contracts/shared/bridge/EtherBridgeWrapper.sol
```diff
@@ -0,0 +1,235 @@
...
+pragma solidity ^0.8.24;
+
+import "./Bridge.sol";
+import "../../shared/based/ITaiko.sol";
+import "src/layer1/based/ITaikoInbox.sol";
+import "src/shared/libs/LibAddress.sol";
+import "src/shared/common/EssentialContract.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+contract EtherBridgeWrapper is EssentialContract {
```
dantaik, line 11: Add Natspac comments please

packages/protocol/contracts/shared/bridge/EtherBridgeWrapper.sol
```diff
@@ -0,0 +1,235 @@
...
+        // Gas limit for the operation.
+        uint32 gasLimit;
+        // Amount of Ether to be sent.
+        uint256 amount;
+        // Added solver fee
+        uint256 solverFee;
+    }
+
+    /// @dev Represents an operation to solve an Ether bridging intent
+    struct SolverOp {
```
dantaik, line 34: Is it possible to reuse the SolverOp defines in ERC20Vault (maybe extract to another file) and use `0x0` for `EtherBridgeOp` to indicate the asset is Ether.

Same recommendation for `EtherBridgeOp`.

packages/protocol/contracts/shared/bridge/EtherBridgeWrapper.sol
```diff
@@ -0,0 +1,235 @@
...
+    /// @param _owner The owner of this contract. msg.sender will be used if this value is zero.
+    /// @param _sharedResolver The {IResolver} used by multipel rollups.
+    function init(address _owner, address _sharedResolver) external initializer {
+        __Essential_init(_owner, _sharedResolver);
+    }
+
+    /// @notice Sends Ether to another chain.
+    /// @param _op Options for sending Ether.
+    /// @return message_ The constructed message.
+    function sendToken(EtherBridgeOp calldata _op)
```
dantaik, line 87: sendEther?

packages/protocol/contracts/shared/bridge/EtherBridgeWrapper.sol
dantaik, line 11: How about we call it EtherVault? If this contract can be merged into ERC20Vault and we consider sending Ether is just another special instance of sending ERC20 tokens with a special token address  `0x0`, then I'd strongly prefer that implementation.

AnshuJalan:
Solver logic for eth has been merged into `ERC20Vault`. The formerly added `EtherBridgeWrapper` has been deleted.

packages/protocol/contracts/shared/tokenvault/ERC20Vault.sol
```diff
@@ -390,8 +399,11 @@ contract ERC20Vault is BaseVault {
             abi.decode(data, (CanonicalERC20, address, address, uint256, uint256, bytes32));
 
         // Transfer the ETH and tokens back to the owner
-        address token = _transferTokens(ctoken, _message.srcOwner, amount + solverFee);
-        _message.srcOwner.sendEtherAndVerify(_message.value);
+        uint256 amountToReturn = amount + solverFee;
+        address token = _transferTokens(ctoken, _message.srcOwner, amountToReturn);
```
dantaik, line 403: _transferTokens => _transferTokenOrEther, is it a better name?

packages/protocol/contracts/shared/tokenvault/ERC20Vault.sol
```diff
@@ -390,8 +399,11 @@ contract ERC20Vault is BaseVault {
...
+        _message.srcOwner.sendEtherAndVerify(
+            ctoken.addr == address(0) ? _message.value - amountToReturn : _message.value
+        );
```
dantaik, line 406: Maybe we can merge _transferTokens and sendEtherAndVerify into one function call to avoid sending back ether in 2 separate transfers?

packages/protocol/contracts/shared/tokenvault/ERC20Vault.sol
```diff
@@ -495,20 +518,20 @@ contract ERC20Vault is BaseVault {
...
-        // An identifier hash for the solver condition on destination chain
         bytes32 solverCondition;
 
-        // If it's a bridged token
-        CanonicalERC20 storage _ctoken = bridgedToCanonical[_op.token];
-        if (_ctoken.addr != address(0)) {
-            ctoken_ = _ctoken;
-            // Following the "transfer and burn" pattern, as used by USDC
+        if (_op.token == address(0)) {
+            balanceChangeAmount_ = _op.amount;
```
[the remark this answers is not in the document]
AnshuJalan: May I know what purpose this may serve for Ether? In the case of ERC20, I see this is required to choose between transferring a token if it is canonical or minting it if it is bridged.

packages/protocol/contracts/shared/tokenvault/ERC20Vault.sol
AnshuJalan: It would be conflicting with the logic in `onMessageInvocation` where the tokenRecipient could be different from the `to` entity:
```solidity
token = _transferTokensOrEther(ctoken, tokenRecipient, amountToTransfer);
            to.sendEtherAndVerify(
                ctoken.addr == address(0) ? msg.value - amountToTransfer : msg.value
            );
```

AnshuJalan:
@dantaik @cyberhorsey Would you mind managing the conflicts here? I am not sure where did these turn up from

packages/protocol/contracts/shared/tokenvault/ERC20Vault.sol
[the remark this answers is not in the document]
dantaik: The difference between Ether and ECR20 tokens is the address being 0 or not. But since we have checked _ctoken.chainId for ERC20 tokens, we shall do the same for Ether. Otherwise, `_ctoken.chainId == block.chainid` check shall also be removed for ERC20, but I think that will not be OK.

dantaik:
@AnshuJalan I helped you with the conflicts, but please double check if some of your changes are deleted by mistake :D

dantaik (approved):

YoGhurt111 (approved):

smtmfft (approved):

PR #18806, opened 2025-01-21
fix(protocol): register specific proof type verifiers

YoGhurt111:


packages/protocol/script/layer1/based/DeployProtocolOnL1.s.sol
```diff
@@ -272,7 +272,8 @@ contract DeployProtocolOnL1 is DeployCapability {
         address opVerifier = deployProxy({
             name: "op_verifier",
             impl: address(new OpVerifier(l2ChainId)),
-            data: abi.encodeCall(OpVerifier.init, (owner, rollupResolver))
```
dantaik, line 275: pleaser use branch `immutable_resolver` as your base

YoGhurt111: ok

davidtaikocha (approved):

PR #18807, opened 2025-01-21
test(taiko-client): fix integration test for pacaya fork

YoGhurt111:


YoGhurt111:
> I think no need to update the client side rn, let's implement the upgrade script in protocol at first.

OK

PR #18812, opened 2025-01-21
feat(protocol): whitelist addition + deploy script update

cyberhorsey:
deplying the preconf contracts is now tested with `pnpm run test:deploy:l1`, and as well deployable via our kubernetes scripts for a one-script deploy setup.

We also add a mapping to keep track of operators so we can't double-add by accident., and read offchain if someone is an operator already.

### DIFF
--- a/packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
+++ b/packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
@@ -19,7 +19,7 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
     // Maps operator index to their corresponding operator addresses
     mapping(uint256 operatorIndex => address operator) public operatorIndexToOperator;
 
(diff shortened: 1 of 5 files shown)


packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -13,6 +13,9 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
     // Tracks the total number of operators in the whitelist
     uint256 public operatorCount;
 
+    // Tracks whether an address is an operator
+    mapping(address operator => bool isOperator) public isOperator;
+
```
dantaik, line 18: `uint256[50] private __gap;` below should be changed to `uint256[47] private __gap;`

dantaik (changes requested):

dantaik (approved):

AnshuJalan (approved):

PR #18815, opened 2025-01-22
feat(protocol): support forced transactions as escape hatch  

dantaik:
This PR adds forced transaction supportas escape hatchwhen rollup proposers become overly centralized.  

## Key Implementation:

A block will now support 3 L2 transaction data sources at the same time:  

  - `forcedTxs`: Read from ForcedTransactionStore (empty if invalid decoding/non-existent).  
  - `txList`: Calldata from `proposeBatch` function.  
  - `blobs`: Data from 0+ blobs, concatenated before processing. 
 

## Data-to-Block Conversion Flow

1. Each source is zlib-decoded into transaction lists **A** (`forcedTxs`), **B** (`txList`), **C** (blob data). Invalid results become empty lists.  
2. Transactions are concatenated into final list **T** = **A** + **B** + **C** (notice the order)
3. L2 blocks are built from **T** by:  
   - Taking up to `blockParams.numTransactions` transactions per L2 block.
   - Removing invalid transactions (e.g., bad nonce, insufficient Ether balance)
   - Truncating transactions exceeding the block’s gas limit (discarded permanently, not reused by next block).  

This ensures forced transactions bypass centralized proposers while maintaining validity and gas constraints.

@davidcardenasus @smtmfft we need to talk about this PR ASAP.

## TODO
- [ ] copy the conversion flow as net-spec comments to TaikoInbox.sol.

### DIFF
--- a/packages/protocol/contracts/layer1/based/IForcedTransactionStore.sol
+++ b/packages/protocol/contracts/layer1/based/IForcedTransactionStore.sol
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.24;
 
-/// @title IForcedTransactionProvider
+/// @title IForcedTransactionStore
 /// @custom:security-contact security@taiko.xyz
-interface IForcedTransactionProvider {
+interface IForcedTransactionStore {
     function consumeForcedTransactions()
         external
         returns (bytes memory forcedTxs_);
--- a/packages/protocol/contracts/layer1/based/IForcedTransactionStore.sol
+++ b/packages/protocol/contracts/layer1/based/IForcedTransactionStore.sol
@@ -4,7 +4,5 @@ pragma solidity ^0.8.24;
 /// @title IForcedTransactionStore
 /// @custom:security-contact security@taiko.xyz
 interface IForcedTransactionStore {
-    function consumeForcedTransactions()
-        external
-        returns (bytes memory forcedTxs_);
+    function consumeForcedTransactions() external returns (bytes memory forcedTxs_);
 }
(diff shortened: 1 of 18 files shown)


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -103,6 +104,23 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, ITaiko, IFork {
...
+                if (store != address(0)) {
+                    forcedTxs = IForcedTransactionStore(store).consumeForcedTransactions();
+                }
+            }
+
+            // If forced txs is not empty, the first block must be big enough to include all these
+            // forced transactions.
+            if (forcedTxs.length != 0) {
+                require(
+                    params.blocks[1].numTransactions >= config.minForcedTxsBlockTxCount,
```
AnshuJalan, line 119: This is the second block (index 1). The comment states it should be the first block

packages/protocol/contracts/layer1/based/IForcedTransactionStore.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+/// @title IForcedTransactionStore
+/// @custom:security-contact security@taiko.xyz
+interface IForcedTransactionStore {
```
AnshuJalan, line 6: How are forced transactions added to this store? Are there additional functions yet to be written?

dantaik: The implementation is still missing.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik, line 119:
```suggestion
                    params.blocks[0].numTransactions >= config.minForcedTxsBlockTxCount,
```

dantaik:
After discussion with David, I think we can have a better design.

PR #18822, opened 2025-01-22
feat(nfts): out dir inclusion

bearni95:
Include the `out` dir into the tree to be used by external packages


dantaik:
I don't think it's a good idea to checkin output to git.

PR #18824, opened 2025-01-22
feat(protocol): router based forced tx inclusion

cyberhorsey:
offchain, in the gateway itself, we can:
in the gateway itself, we can:
1) add listeners for the events
2) call forcedTxList = preconfRouter.forcedTxLists[keccak256(event.txList)]
3) check forcedTxList.included
4) if !forcedTxList.included, propose a batch with that txList immediately to get it out of the way

if that isnt done, anyone can propose the block after `inclusionWindow` passes. TaikoInbox remains untouched.

Needs some more tests, and some are failing, but please look over @dantaik @AnshuJalan @davidtaikocha

### DIFF
--- a/packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
+++ b/packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
@@ -6,8 +6,23 @@ import "src/layer1/based/ITaikoInbox.sol";
 /// @title IPreconfRouter
 /// @custom:security-contact security@taiko.xyz
 interface IPreconfRouter {
+
+    struct ForcedTx {
+        bytes txList;
+        uint256 timestamp;
+        bool included;
+        uint256 stakeAmount;
+    }
+
+    error ForcedTxListAlreadyIncluded();
+    error ForcedTxListAlreadyStored();
+    error ForcedTxListHashNotFound();
+    error InsufficientStakeAmount();
     error NotTheOperator();
     error ProposerIsNotTheSender();
+
+
+    event ForcedTxStored(bytes indexed txHash, uint256 timestamp);
 
     /// @notice Proposes a batch of blocks that have been preconfed.
     /// @dev This function only accepts batches from an operator selected to preconf in a particular
@@ -19,8 +34,13 @@ interface IPreconfRouter {
     function proposePreconfedBlocks(
         bytes calldata _params,
         bytes calldata _batchParams,
-        bytes calldata _batchTxList
+        bytes calldata _batchTxList,
+        bool force
     )
         external
         returns (ITaikoInbox.BatchMetadata memory meta_);
+
+    function updateBaseStakeAmount(uint256 _newBaseStakeAmount) external;
+
+    function storeForcedTx(bytes calldata _txList) payable external;
 }
--- a/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
+++ b/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
@@ -4,33 +4,95 @@ pragma solidity ^0.8.24;
 import "../iface/IPreconfRouter.sol";
 import "../iface/IPreconfWhitelist.sol";
 import "src/layer1/based/ITaikoInbox.sol";
+import "src/shared/libs/LibAddress.sol";
 import "src/shared/libs/LibStrings.sol";
 import "src/shared/common/EssentialContract.sol";
 
 /// @title PreconfRouter
 /// @custom:security-contact security@taiko.xyz
 contract PreconfRouter is EssentialContract, IPreconfRouter {
-    uint256[50] private __gap;
+    mapping(bytes32 => ForcedTx) public forcedTxLists;
 
-    constructor(address _resolver) EssentialContract(_resolver) { }
+    uint256 public pendingForcedTxHashes;
+    
+    uint256 public inclusionWindow;
+
+    uint256 public baseStakeAmount;
+
+    uint256[46] private __gap;
+
+    constructor(address _resolver, uint256 _inclusionWindow, uint256 _baseStakeAmount) EssentialContract(_resolver) {
+        inclusionWindow = _inclusionWindow;
+        baseStakeAmount = _baseStakeAmount;
+    }
 
     function init(address _owner) external initializer {
         __Essential_init(_owner);
+    }
+
+    function updateBaseStakeAmount(uint256 _newBaseStakeAmount) external onlyOwner {
+        baseStakeAmount = _newBaseStakeAmount;
+    }
+
+    function getRequiredStakeAmount() public view returns (uint256) {
+        uint256 multiplier = pendingForcedTxHashes == 0 ? 1 : pendingForcedTxHashes;
+        return baseStakeAmount * multiplier;
+    }
+
+    function storeForcedTx(bytes calldata _txList) payable external {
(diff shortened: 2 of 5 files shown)


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -4,33 +4,95 @@ pragma solidity ^0.8.24;
...
+         if (force) {
+            require(forcedTxLists[forcedTxListHash].timestamp != 0, ForcedTxListHashNotFound());
+            require(!forcedTxLists[forcedTxListHash].included, ForcedTxListAlreadyIncluded());
+
+            // Pay out the stake to the proposer
+            LibAddress.sendEtherAndVerify(msg.sender, forcedTxLists[forcedTxListHash].stakeAmount);
+
+            pendingForcedTxHashes--;
+
+            forcedTxLists[forcedTxListHash].included = true;
```
dantaik, line 93: `forcedTxLists[forcedTxListHash].txList` is not used. I'm not sure we can simply replace `_batchTxList` with `forcedTxLists[forcedTxListHash].txList`, you will also need to construct a `batchParams` as:

```
BatchParams memory params;
params.blocks = new BlockParams[](1);
params.blocks[0].numTransactions = some_constant.
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -4,33 +4,95 @@ pragma solidity ^0.8.24;
...
     function init(address _owner) external initializer {
         __Essential_init(_owner);
     }
 
+    function updateBaseStakeAmount(uint256 _newBaseStakeAmount) external onlyOwner {
+        baseStakeAmount = _newBaseStakeAmount;
+    }
+
+    function getRequiredStakeAmount() public view returns (uint256) {
+        uint256 multiplier = pendingForcedTxHashes == 0 ? 1 : pendingForcedTxHashes;
```
dantaik, line 38:
```suggestion
        return baseStakeAmount * (pendingForcedTxHashes + 1);
```

But how about make it exponential:
```suggestion
        return (2 ** pendingForcedTxHashes).max(4096) * baseStakeAmount;
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -4,33 +4,95 @@ pragma solidity ^0.8.24;
...
+        return baseStakeAmount * multiplier;
+    }
+
+    function storeForcedTx(bytes calldata _txList) payable external {
+        uint256 requiredStake = getRequiredStakeAmount();
+        require(msg.value >= requiredStake, InsufficientStakeAmount());
```
dantaik, line 44:
```suggestion
        require(msg.value == requiredStake, InsufficientStakeAmount());
```

dantaik: Otherwise, please send back the overpaid fee.

packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
```diff
@@ -6,9 +6,24 @@ import "src/layer1/based/ITaikoInbox.sol";
 /// @title IPreconfRouter
 /// @custom:security-contact security@taiko.xyz
 interface IPreconfRouter {
+
+    struct ForcedTx {
+        bytes txList;
+        uint256 timestamp;
+        bool included;
+        uint256 stakeAmount;
```
dantaik, line 14: how about `priorityFee`

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
dantaik: Basically you need to call `ITaikoInbox(taikoInbox).proposeBatch` twice, one for the forced batch, the other one for the normal batch. The first batch always have 1 block with MAX_FORCED_TRANSACTIONS txs.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -4,33 +4,95 @@ pragma solidity ^0.8.24;
...
+        emit ForcedTxStored(_txList, block.timestamp);
+    }
+
     /// @inheritdoc IPreconfRouter
     function proposePreconfedBlocks(
         bytes calldata,
         bytes calldata _batchParams,
-        bytes calldata _batchTxList
+        bytes calldata _batchTxList,
+        bool force
```
dantaik, line 66: the preconfer can always use "false" for `force`, so we are not forcing?

dantaik: Maybe we can try to query if there is a forced txs request due, if so, force will be automatically true.

dantaik:
I really like the idea that forced inclusion of transactions/blocks are done out of the core protocol code. I would suggest to have this inheriting relationship:

TaikoInbox -> TaikoInboxWithForcedInclusion -> MainnetInbox

Then move all forced inclusion logics into TaikoInboxWithForcedInclusion and leave the router unchanged. TaikoInboxWithForcedInclusion shall always check if there is one inclusion request due, if so, it calls `proposeBatch` twice with the first one being the forced inclusion block. Forced inclusion shall be transparent to both TaikoInbox and the router.

dantaik (changes requested):

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
cyberhorsey: force is there so that a non preconfer can call this function with force = true, and then it will check the forced block inclusion. That way, the user doesn't rely on a preconfer to include it, nor does it block preconfer from proposing.

packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
cyberhorsey: yeah that's maybe a better name

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
cyberhorsey: will add

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
cyberhorsey: The idea if that you would call this with its own batch, and `_batchTxList` *is* `forcedTxLists[forcedTxListHash].txList`. It is gas inefficient and will likely never be used, since we don't plan on censoring any user's transactions, but it keeps it very simple.

This function is etiher called by the preconfing gateway, with it's own batch, and is profitable because of the `priorityFee` reward, or by the user themselves, and either way the input _batchTxList is indeed the forcedTxList, because `forcedTxListHash` is a hash of `_batchTxList`.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
dantaik: But this will ruin the preconfirmation - the preconfirmer may already shard some shared blocks before they are proposed on L1, now suddenly a user kicks in with a block which will reorg the chain.

PR #18826, opened 2025-01-23
feat(protocol): support delayed forced inclusion of txs

dantaik:
Initial implementation of delayed inbox (forced inclusion of transactions)

The original idea is from @cyberhorsey's PR: https://github.com/taikoxyz/taiko-mono/pull/18824/files#diff-bea5dd46ba3d231a238b7c76adc7d09dfa3f8ebd3bf552407f9098492f3ff8f5


dantaik:
@jeff, I had a call with our friends from Nethermind, here are some feedback/suggestions:

- In the `ForcedInclusionStore`, we should support using both calldata and a blob to submit their transactions (lets implement the blob support as the first step)
- A request should be processed with a 12*x seconds delay since 1) the time it is saved or 2) the time the previous request is processed or 3) the time the last batch is proposed. which ever is the biggest.

packages/protocol/contracts/layer1/based/TaikoInboxWithForcedTxInclusion.sol
```diff
@@ -0,0 +1,81 @@
...
+import "src/shared/libs/LibStrings.sol";
+import "src/shared/signal/ISignalService.sol";
+import "src/layer1/verifiers/IVerifier.sol";
+import "./IFork.sol";
+import "./ITaikoInbox.sol";
+import "./IForcedInclusionStore.sol";
+
+/// @title TaikoInboxWithForcedTxInclusion
+/// @custom:security-contact security@taiko.xyz
+contract TaikoInboxWithForcedTxInclusion is EssentialContract {
```
dantaik, line 19: need a better name

packages/protocol/contracts/layer1/based/IForcedInclusionStore.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+/// @title IForcedInclusionStore
+/// @custom:security-contact security@taiko.xyz
+interface IForcedInclusionStore {
```
dantaik, line 6: need a better name

packages/protocol/contracts/layer1/based/ForcedInclusionStore.sol
```diff
@@ -0,0 +1,97 @@
...
+    function consumeForcedInclusion(address _feeRecipient)
+        external
+        onlyFromNamed(LibStrings.B_TAIKO_FORCED_INCLUSION_INBOX)
+        returns (ForcedInclusion memory inclusion_)
+    {
+        // we only need to check the first one, since it will be the oldest.
+        uint64 _head = head;
+        ForcedInclusion storage inclusion = queue[_head];
+
+        if (inclusion.createdAt != 0 && block.timestamp >= inclusionDelay + inclusion.createdAt) {
```
Brechtpd, line 84: I'm kind of thinking that making the inclusion of these forced transactions "automatic" (as in the preconfer cannot decide easily when they are included) could be problematic for a preconfer. Because the preconfer will not always know in which L1 block its propose transaction will be included, but depending on when that happens, these forced transactions need to be inserted at the right place to be able to give preconfirmations after these are supposed to happen. It does seem to introduce a level of non-determinism that could be quite tricky for a preconfer to fully control easily. 

When the preconfer can decide on his own when these transactions actually get included within a window, that seems to make the preconfers job much easier to make things 100% predictable without any tricky boundaries enforced onchai, resulting with a hard deadline on when a propose tx actually needs to be included onchain to not mess with the expected order of the preconfer.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -33,9 +34,18 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
...
+        address forcedInclusionInbox = resolveAddress(LibStrings.B_TAIKO_FORCED_INCLUSION_INBOX, true);
+        if (forcedInclusionInbox == address(0)) {
+            // Call the proposeBatch function on the TaikoInbox
+            address taikoInbox = resolveAddress(LibStrings.B_TAIKO, false);
+            (, meta_) = ITaikoInbox(taikoInbox).proposeBatch(_batchParams, _batchTxList);
+        } else {
+            // Call the proposeBatchWithForcedInclusion function on the ForcedInclusionInbox
+            (, meta_) = IForcedInclusionInbox(forcedInclusionInbox).proposeBatchWithForcedInclusion(
+                _forcedInclusionParams, _batchParams, _batchTxList
+            );
```
Brechtpd, line 47: It seems that this is still only callable by the whitelisted preconfers right? If so, would say again that forced inclusion != inclusion. Wrote about that here before: https://github.com/taikoxyz/taiko-mono/pull/18815#issuecomment-2607335008

Inclusion can only be guaranteed when block proposing is made permissionless again (at least for the forced blocks, but may as well allow any block to be proposed when that happens because why not).

dantaik: I agree with Brecht's point (above) — in the current implementation, forced inclusion can only occur if there is at least one preconfer willing to propose blocks when it becomes the current preconfer.

To address this, one potential solution is to enable fully permissionless block proposing if no blocks have been proposed within the last N seconds (though some details still need to be ironed out). While this approach is related to forced inclusion, I believe it aligns more closely with the core protocol or preconfirmation design. Even without forced inclusion, this feature would be a valuable addition to the preconfirmation mechanism. Without it, if there are only a few preconfers and they all stop proposing blocks due to profitability issues, the chain could come to a halt.

packages/protocol/contracts/layer1/based/ForcedInclusionStore.sol
dantaik: Another good point. I think one way to solve this is to not use timestamp as the indicator whether a inclusion request is due, but to use `numBatches` in the ITaikoInbox contract. This basically means every N batch proposals, the oldest inclusion must be processed.

dantaik: See an improvement idea: https://github.com/taikoxyz/taiko-mono/pull/18842

dantaik:
This PR is ready for another round of review.

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
```diff
@@ -0,0 +1,113 @@
...
+            // MAX_FORCED_TXS_PER_FORCED_INCLUSION transactions
+            if (params.blocks.length == 0) {
+                params.blocks = new ITaikoInbox.BlockParams[](1);
+            }
+
+            if (params.blocks[0].numTransactions < MAX_FORCED_TXS_PER_FORCED_INCLUSION) {
+                params.blocks[0].numTransactions = MAX_FORCED_TXS_PER_FORCED_INCLUSION;
+            }
+
+            params.blobParams.blobHashes = new bytes32[](1);
```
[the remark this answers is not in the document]
dantaik: See line 94

```solidity
            if (!calldataUsed) {
                if (params.blobParams.blobHashes.length == 0) {
                    require(params.blobParams.numBlobs != 0, BlobNotSpecified());
                } else {
                    require(
                        params.blobParams.numBlobs == 0 && params.blobParams.firstBlobIndex == 0,
                        InvalidBlobParams()
                    );
                }
            }
            ```

PR #18830, opened 2025-01-24
feat(protocl): refactor based on Jeff's code

dantaik:
Summary:

- make the inclusion delay and fee immutable, we can upgrade the implementation to change these values.
- remove `id` from the ForcedInclusion struct.
- require time delay must be divide-able by 12 to match slot time.
- send Ether to the given recipient
- use blob index, not blob hash in function param

## TODO(jeff):

- Better contract/file names (and mv to sub-dir?)
- Add the two implementation file to the end of gen-layouts.sh


packages/protocol/contracts/layer1/based/ForcedInclusionStore.sol
```diff
@@ -4,91 +4,92 @@ pragma solidity ^0.8.24;
...
-            ForcedInclusion memory consumedInclusion = forcedInclusionQueue[head];
-            head++;
+        ForcedInclusion storage inclusion = queue[head];
 
-            emit ForcedInclusionConsumed(consumedInclusion);
-            
-            return consumedInclusion;
+        if (inclusion.createdAt != 0 && block.timestamp >= inclusionDelay + inclusion.createdAt) {
+            inclusion_ = inclusion;
+            _feeRecipient.sendEtherAndVerify(inclusion.fee);
```
cyberhorsey, line 85: you're sending the fee twice now, because you already send it outside of the store when it's consumed

packages/protocol/contracts/layer1/based/TaikoInboxWithForcedTxInclusion.sol
```diff
@@ -69,13 +74,13 @@ contract TaikoInboxWithForcedTxInclusion is EssentialContract {
...
-            params.blobParams.byteOffset = forcedInclusion.blobByteOffset;
-            params.blobParams.byteSize = forcedInclusion.blobByteSize;
+            params.blobParams.blobHashes[0] = inclusion.blobHash;
+            params.blobParams.byteOffset = inclusion.blobByteOffset;
+            params.blobParams.byteSize = inclusion.blobByteSize;
 
             inbox.proposeBatch(abi.encode(params), "");
-            msg.sender.sendEtherAndVerify(forcedInclusion.priorityFee, gasleft());
-            emit ForcedInclusionProcessed(forcedInclusion);
+            msg.sender.sendEtherAndVerify(inclusion.fee, gasleft());
```
cyberhorsey, line 82: this is where you're double sending the inclusion fee

packages/protocol/contracts/layer1/based/ForcedInclusionStore.sol
dantaik: fixed

PR #18842, opened 2025-01-27
feat(protocol): use num proposals to measure forced inclusion deadline

dantaik:
The following changes are implemented in this PR:

1. **Replaced timestamp-based deadlines with L2 proposal count**: The number of L2 proposals (batches) is now used to determine whether the oldest forced inclusion is due. This change addresses issues related to timestamp-based deadlines (see [Brecht's comment](https://github.com/taikoxyz/taiko-mono/pull/18826#discussion_r1929252805)).
   
2. **Enabled proactive processing of inclusion requests**: Proposers/preconfers are now allowed to proactively process the oldest inclusion request even before it is due.

@cyberhorsey Tests are broken and not yet fixed.


smartprogrammer93 (approved):
I like it 🙏 Apologies for the force push, just pressed rebase button on github 🤦‍♀️

AnshuJalan (commented):
@dantaik Will `ForcedInclusionInbox` only be callable from the `PreconfRouter` or can anyone permissionless-ly call the required function to clear the oldest forced inclusion?

dantaik:
Good question. I think we have at least two possible approaches:

1.	Independent of the Preconfirmation: The ForcedInclusionInbox could include an additional function allowing any address to permissionlessly propose a batch of blocks, provided that it includes a pending forced inclusion request that has exceeded a certain time threshold.

3.	Considering Preconfirmation: If the current preconfer stop proposing blocks, the `TaikoInbox` or `ForcedInclusionInbox` (I prefer to rename it to `TaikoInboxWrapper`) could permit any address to propose blocks/batches once a timeout occurs. This ensures not only that forced inclusion requests are processed, but also that normal blocks can also be proposed to keep the chain alive, when preconfer become inactive.

I believe option 2 is the better approach because, with option 1, we would still need a fallback mechanism to handle cases where preconfers stop working. Option 2 directly addresses this issue by ensuring a permissionless mechanism kicks in when necessary.

PR #18851, opened 2025-01-30
fix(protocol): check proposer param must be msg.sender in PreconfRouter

dantaik:
This was reported by OZ. Note that same issue also exist in our main branch, but the code will not be used in production for Hekla/Mainnet.

> Possible to Debit Bond Balances from Any Account
> Taiko developed the PreconfTaskManager and related smart contracts to address preconfirmation challenges in based rollups. In this approach, users that want to execute transaction pass them to a proposer, who provides a preconfirmation that can be used to penalize the proposer if the transaction is not included. At the start of each epoch, a proposer is designated for the next epoch. Once the PreconfTaskManager is deployed and the appropriate name is assigned in the resolver, the designated proposer calls the newBlockProposals[ function](https://github.com/PufferFinance/unifi-mono/blob/afd0c91dda95f56050fc36bb7188742a1402d267/packages/protocol/contracts/layer1/preconf/impl/PreconfTaskManager.sol#L89), triggering a block proposal in TaikoL1 after stake locking and other parameter validations. Notably, there is no validation of the blockParamsArr[argument](https://github.com/PufferFinance/unifi-mono/blob/afd0c91dda95f56050fc36bb7188742a1402d267/packages/protocol/contracts/layer1/preconf/impl/PreconfTaskManager.sol#L129) in this workflow.
In the TaikoL1[ contract](https://github.com/PufferFinance/unifi-mono/blob/afd0c91dda95f56050fc36bb7188742a1402d267/packages/protocol/contracts/layer1/based/TaikoL1.sol#L71-L83), LibProposing’s proposeBlocks function manages block proposals by calling _proposeBlock for each proposed block. After verifying the block number, it retrieves the preconfTaskManager[ address](https://github.com/PufferFinance/unifi-mono/blob/afd0c91dda95f56050fc36bb7188742a1402d267/packages/protocol/contracts/layer1/based/LibProposing.sol#L142-L146). If that address is set, only PreconfTaskManager is allowed to propose blocks, and the actual proposer is assumed to be encoded in the _params argument. This argument is decoded and checked in _validateParams. If _local.params.proposer[ is set to a non-zero address](https://github.com/PufferFinance/unifi-mono/blob/afd0c91dda95f56050fc36bb7188742a1402d267/packages/protocol/contracts/layer1/based/LibProposing.sol#L246-L251), the caller must be the designated proposer or _local.allowCustomProposer must be true (which it is when called through PreconfTaskManager). In case of default values, the contract will compute and assign the correct values on its own. The decoded proposer address [is written to the block metadata](https://github.com/PufferFinance/unifi-mono/blob/afd0c91dda95f56050fc36bb7188742a1402d267/packages/protocol/contracts/layer1/based/LibProposing.sol#L174) stored in the contract’s blocks, and the [proposer’s bond balance is debited](https://github.com/PufferFinance/unifi-mono/blob/afd0c91dda95f56050fc36bb7188742a1402d267/packages/protocol/contracts/layer1/based/LibProposing.sol#L226) by the livenessBond amount of 125 Taiko tokens. Because any address can be passed as the proposer, an external actor can effectively debit tokens from an arbitrary account. The actor can also provide invalid data in the proposed block to dispute it and potentially claim part of specified proposer’s bond balance.
Consider implementing verification of both the caller and the proposer address in the PreconfTaskManager contract and the LibProposing library to ensure only the genuine proposer’s bond is debited.

### DIFF
--- a/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
+++ b/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
@@ -27,6 +27,11 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
         external
         returns (ITaikoInbox.BatchMetadata memory meta_)
     {
+        // Make sure the proposer must be the msg.sender itself.
+        ITaikoInbox.BatchParams memory batchParams =
+            abi.decode(_batchParams, (ITaikoInbox.BatchParams));
+        require(msg.sender == batchParams.proposer, InvalidParams());
+
         // Sender must be the selected operator for the epoch
         address selectedOperator =
             IPreconfWhitelist(resolve(LibStrings.B_PRECONF_WHITELIST, false)).getOperatorForEpoch();
--- a/packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
+++ b/packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
@@ -6,6 +6,7 @@ import "src/layer1/based/ITaikoInbox.sol";
 /// @title IPreconfRouter
 /// @custom:security-contact security@taiko.xyz
 interface IPreconfRouter {
+    error InvalidParams();
     error NotTheOperator();
     error ProposerIsNotTheSender();
 
--- a/packages/protocol/test/layer1/based/InboxTest_BondToken.t.sol
+++ b/packages/protocol/test/layer1/based/InboxTest_BondToken.t.sol
@@ -97,7 +97,7 @@ contract InboxTest_BondToken is InboxTestBase {
         inbox.depositBond(depositAmount);
     }
 
-    function test_inbox_exceeding_balance() external {
+    function test_inbox_exceeding_token_balance() external {
         vm.warp(1_000_000);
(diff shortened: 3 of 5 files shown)


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
AnshuJalan, line 1: @dantaik Isn't this line already checking that? 

```solidity
  // Verify that the sender had set itself as the proposer
        require(meta_.proposer == msg.sender, ProposerIsNotTheSender());
```

I earlier did the decoding + encoding thing in my commit, but we planned to change it to use the returned metadata right? 🤔

dantaik:
It turns out this "fix" is not necessary as we have already verified the actual proposer must be msg.sender as the last statement:

```solidity
  // Verify that the sender had set itself as the proposer
        require(meta_.proposer == msg.sender, ProposerIsNotTheSender());
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
dantaik: Thank you for pointing that out.

PR #18854, opened 2025-01-31
fix(taiko-client): some client relate upgrades

mask-pp:


packages/taiko-client/bindings/encoding/input.go
```diff
@@ -591,8 +591,8 @@ func CalculatePacayaDifficulty(blockNum *big.Int) ([]byte, error) {
 	return packed, nil
 }
 
-// UnpackTxListBytes unpacks the input data of a TaikoL1.proposeBlock transaction, and returns the txList bytes.
-func UnpackTxListBytes(txData []byte) ([]byte, error) {
+// UnpackOntakeTxListBytes unpacks the input data of a TaikoL1.proposeBlock transaction, and returns the txList bytes.
+func UnpackOntakeTxListBytes(txData []byte) ([]byte, error) {
```
davidtaikocha, line 595:
```suggestion
func UnpackTxListBytesOntake(txData []byte) ([]byte, error) {
```

packages/taiko-client/bindings/encoding/input.go
```diff
@@ -618,6 +618,33 @@ func UnpackTxListBytes(txData []byte) ([]byte, error) {
 	return inputs, nil
 }
 
+// UnpackPacayaTxListBytes unpacks the input data of a TaikoL1.proposeBlock transaction, and returns the txList bytes.
+func UnpackPacayaTxListBytes(txData []byte) ([]byte, error) {
```
davidtaikocha, line 622:
```suggestion
func UnpackTxListBytesPacaya(txData []byte) ([]byte, error) {
```

packages/taiko-client/bindings/encoding/input.go
```diff
@@ -618,6 +618,33 @@ func UnpackTxListBytes(txData []byte) ([]byte, error) {
 	return inputs, nil
 }
 
+// UnpackPacayaTxListBytes unpacks the input data of a TaikoL1.proposeBlock transaction, and returns the txList bytes.
```
davidtaikocha, line 621:
```suggestion
// UnpackPacayaTxListBytes unpacks the input data of a TaikoInbox.proposeBatch transaction, and returns the txList bytes.
```

packages/taiko-client/bindings/encoding/input.go
davidtaikocha, line 595: After think twice, maybe we won't this method anymore? since we switched to fetch the txList from the event data?

mask-pp: Yeah, later this function won't be used anymore.

PR #18855, opened 2025-01-31
fix(taiko-client): make sure the length of txLists is not bigger than the maxBlocksPerBatch.

mask-pp:


packages/taiko-client/proposer/proposer.go
```diff
@@ -313,9 +313,15 @@ func (p *Proposer) ProposeTxLists(ctx context.Context, txLists []types.Transacti
 			return nil
 		}
 
-		if err := p.ProposeTxListPacaya(ctx, txLists); err != nil {
-			return err
+		// Make sure the tx list is not bigger than the maxBlocksPerBatch.
+		maxBlocksPerBatch, _ := p.protocolConfigs.MaxBlocksPerBatch()
```
davidtaikocha, line 317: Let's check the error here, or u don't need to make this function return an error.

packages/taiko-client/proposer/proposer.go
```diff
@@ -313,9 +313,15 @@ func (p *Proposer) ProposeTxLists(ctx context.Context, txLists []types.Transacti
...
+		for ; len(txLists) > 0; txLists = txLists[min(maxBlocksPerBatch, len(txLists)):] {
+			// If the current L2 chain is after ontake fork, batch propose all L2 transactions lists.
+			if err := p.ProposeTxListPacaya(ctx, txLists[:min(maxBlocksPerBatch, len(txLists))]); err != nil {
```
davidtaikocha, line 320: We need to pass the whole `txLists` to `ProposeTxListPacaya`, let's check the count inside `ProposeTxListPacaya`.

packages/taiko-client/proposer/proposer.go
mask-pp: No need to return error, and removed.

packages/taiko-client/proposer/proposer.go
mask-pp: Added check inside the `ProposeTxListPacaya `.

davidtaikocha (approved):

PR #18858, opened 2025-02-01
fix(protocol): avoid setting stateRoot = 0 in `ContextV2`

YoGhurt111:


packages/protocol/contracts/layer1/based/LibProving.sol
```diff
@@ -336,12 +338,12 @@ library LibProving {
...
                 ts.prover = msg.sender;
-                ts.blockHash = ctx_.tran.blockHash;
-                ts.stateRoot = ctx_.tran.stateRoot;
+                ts.blockHash = pendingTs.blockHash;
+                ts.stateRoot = pendingTs.stateRoot;
 
                 emit TransitionProvedV2({
                     blockId: local.blockId,
-                    tran: ctx_.tran,
+                    tran: pendingTs,
```
dantaik, line 346: so @smtmfft the issue is that in the events `stateRoot = 0`?

dantaik (approved):

smtmfft (approved):

packages/protocol/contracts/layer1/based/LibProving.sol
smtmfft: yes, here this ctx_.tran has 0 state because of the non-sync block check before verification.

davidtaikocha (approved):

PR #18864, opened 2025-02-03
 feat(taiko-client): check if the block is preconfirmed before calling `setHead `

davidtaikocha:
TODO:
- [x] Set `L1HeadOrigin`


packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/common.go
```diff
@@ -93,7 +118,11 @@ func createExecutionPayloads(
 		return nil, err
 	}
 
-	fc := &engine.ForkchoiceStateV1{HeadBlockHash: meta.ParentHash}
+	fc := &engine.ForkchoiceStateV1{
+		HeadBlockHash:      meta.ParentHash,
+		SafeBlockHash:      lastVerfiiedBlockHash,
```
mask-pp, line 123: Why add `SafeBlockHash` and `FinalizedBlockHash`?

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/common.go
```diff
@@ -31,35 +34,56 @@ func createPayloadAndSetHead(
...
-	)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create execution payloads: %w", err)
+	// If the Pacaya block is preconfirmed, we don't need to insert it again.
+	if meta.BlockID.Cmp(new(big.Int).SetUint64(rpc.PacayaClients.ForkHeight)) >= 0 {
+		header, err := isBlockPreconfirmed(ctx, rpc, meta, anchorTx)
+		if err != nil {
+			log.Debug("Failed to check if the block is preconfirmed", "error", err)
+		} else {
+			if header != nil {
```
mask-pp, line 43:
```suggestion
		} else if header != nil {
```

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/common.go
```diff
@@ -168,3 +197,46 @@ func createExecutionPayloads(
 
 	return payload, nil
 }
+
+// isBlockPreconfirmed checks if the block is preconfirmed.
+func isBlockPreconfirmed(
+	ctx context.Context,
+	rpc *rpc.Client,
+	meta *createPayloadAndSetHeadMetaData,
+	anchorTx *types.Transaction,
```
mask-pp, line 206: Unused `anchorTx `.

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/common.go
```diff
@@ -168,3 +197,46 @@ func createExecutionPayloads(
...
+		return nil, fmt.Errorf("header not found for block number %d", blockID)
+	}
+
+	var (
+		args = &miner.BuildPayloadArgs{
+			Parent:       meta.Parent.Hash(),
+			Timestamp:    header.Time,
+			FeeRecipient: header.Coinbase,
+			Random:       header.MixDigest,
+			Withdrawals:  []*types.Withdrawal{},
```
mask-pp, line 226:
```suggestion
			Withdrawals:  meta.Withdrawals,
```

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/common.go
davidtaikocha: changed

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/common.go
davidtaikocha: Think its a nice to have?

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/common.go
davidtaikocha: removed

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/common.go
mask-pp: I don't think so, it's an unnecessary addition.

davidtaikocha: changed: https://github.com/taikoxyz/taiko-mono/pull/18864/commits/15eb9be7a23b321b9254168dc3bb5b511113a518

mask-pp (approved):

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/pacaya.go
```diff
@@ -190,7 +190,7 @@ func (i *BlocksInserterPacaya) InsertBlocks(
 
 		// Get transactions in the block.
 		txs := types.Transactions{}
-		if txListCursor+int(blockInfo.NumTransactions) > len(allTxs) {
+		if txListCursor+int(blockInfo.NumTransactions) <= len(allTxs) {
```
davidtaikocha, line 193: Here is a bug fix @mask-pp @YoGhurt111 @cyberhorsey

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/common.go
```diff
@@ -30,32 +35,68 @@ func createPayloadAndSetHead(
...
+	if err != nil {
+		log.Error("Encode txList error", "blockID", meta.BlockID, "error", err)
+		return nil, err
+	}
+
+	// If the Pacaya block is preconfirmed, we don't need to insert it again.
+	if meta.BlockID.Cmp(new(big.Int).SetUint64(rpc.PacayaClients.ForkHeight)) >= 0 {
+		header, err := isBlockPreconfirmed(ctx, rpc, meta, txListBytes, anchorTx)
+		if err != nil {
+			log.Debug("Failed to check if the block is preconfirmed", "error", err)
```
YoGhurt111, line 50: Shall we throw the error?

davidtaikocha: No need, if its not precomfired or its an invalid preconfirmed block, we insert that block or reorg the chain normally.

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/pacaya.go
YoGhurt111: When `txListCursor+int(blockInfo.NumTransactions) > len(allTxs)`, it means there are invalid txs in `allTxs`, and even if there is just one invalid transaction, we will propose an empty block, is my understanding correct?

davidtaikocha: Yes, if `NumTransactions` is invalid, lets ignore that

YoGhurt111 (approved):

PR #18892, opened 2025-02-07
feat(taiko-client): check whether a submitted Pacaya batch proof is valid

davidtaikocha:


packages/taiko-client/prover/proof_submitter/transaction/builder.go
```diff
@@ -224,21 +224,24 @@ func (a *ProveBlockTxBuilder) BuildProveBatchesPacaya(batchProof *proofProducer.
 		for i, proof := range batchProof.ProofResponses {
 			metas[i] = proof.Meta
 			transitions[i] = pacayaBindings.ITaikoInboxTransition{
-				ParentHash: proof.Opts.PacayaOptions().Headers[i].ParentHash,
-				BlockHash:  proof.Opts.PacayaOptions().Headers[i].Hash(),
-				StateRoot:  proof.Opts.PacayaOptions().Headers[i].Root,
+				ParentHash: proof.Opts.PacayaOptions().Headers[0].ParentHash,
+				BlockHash:  proof.Opts.PacayaOptions().Headers[len(proof.Opts.PacayaOptions().Headers)-1].Hash(),
+				StateRoot:  proof.Opts.PacayaOptions().Headers[len(proof.Opts.PacayaOptions().Headers)-1].Root,
```
davidtaikocha, line 229: Here should also be a bug fix @YoGhurt111 @mask-pp

packages/taiko-client/prover/event_handler/assignment_expired.go
```diff
@@ -101,6 +101,10 @@ func (h *AssignmentExpiredEventHandler) Handle(
 
 	// If there is no contester, we submit a contest to protocol.
 	go func() {
+		if meta.IsPacaya() {
+			h.proofSubmissionCh <- &proofProducer.ProofRequestBody{Meta: meta}
```
YoGhurt111, line 105: Should we throw an ERROR instead of requesting another proof when encountering an invalid proof?

packages/taiko-client/prover/proof_submitter/transaction/builder.go
YoGhurt111: Why 'i' cannot be used as an index? I'm still trying to understand this.

packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -35,7 +35,7 @@ var (
 )
 
 // GetProtocolConfigs gets the protocol configs from TaikoInbox or TaikoL2 contract.
-func (c *Client) GetProtocolConfigs(opts *bind.CallOpts) (config.ProtocolConfigs, error) {
+func (c *Client) GetProtocolConfigs(opts *bind.CallOpts) (cfg config.ProtocolConfigs, err error) {
```
mask-pp, line 38:
```suggestion
func (c *Client) GetProtocolConfigs(opts *bind.CallOpts) (config.ProtocolConfigs, error) {
```

packages/taiko-client/prover/event_handler/assignment_expired.go
YoGhurt111: So should we rely on this event to start our manual fix workflow, rather than depending on specific error messages?
With this approach, since every user needs to submit a transaction, it would result in unnecessary gas cost

davidtaikocha: Yep, if there is an invalid, we need to pause the protocol instead of letting it verify, since it is really a rare case, the unnecessary cost is ok.

packages/taiko-client/prover/proof_submitter/transaction/builder.go
YoGhurt111: Sure, you are right!

YoGhurt111 (approved):

mask-pp (approved):

PR #18899, opened 2025-02-10
fix(taiko-client): avoid empty guardianProverHeartbeater

mask-pp:


YoGhurt111:
We don't need to make any more changes to the guardian.

PR #18903, opened 2025-02-10
fix(taiko-client): fix bug about pacaya block id

mask-pp:


packages/taiko-client/driver/chain_syncer/blob/syncer.go
```diff
@@ -315,7 +315,7 @@ func (s *Syncer) checkLastVerifiedBlockMismatch(ctx context.Context) (*rpc.Reorg
 		lastVerifiedBlockID = blockInfo.BlockId
 		lastVerifiedBlockHash = blockInfo.BlockHash
 	} else {
-		lastVerifiedBlockID = ts.BlockId
+		lastVerifiedBlockID = ts.BatchId
```
davidtaikocha, line 318: I think it should be `ts.BlockId`?

davidtaikocha (changes requested):

packages/taiko-client/driver/chain_syncer/blob/syncer.go
mask-pp: I'm checking and trying to make sure it.

PR #18921, opened 2025-02-12
fix(taiko-client): avoid null preconfBlockServer be used

mask-pp:


davidtaikocha:
Will be solved by https://github.com/taikoxyz/taiko-mono/pull/18920 ?

mask-pp:
> Will be solved by #18920 ?

The changes in that PR don't include this fix.

PR #18936, opened 2025-02-14
fix(taiko-client): fix the workflow to get `proof_verifier`

YoGhurt111:


packages/taiko-client/cmd/flags/common.go
```diff
@@ -89,6 +89,13 @@ var (
 		Category: commonCategory,
 		EnvVars:  []string{"TAIKO_TOKEN"},
 	}
+	ResolverAddress = &cli.StringFlag{
+		Name:     "resolver",
```
davidtaikocha, line 93: `c.PacayaClients.TaikoInbox.Resolver()` can also work, looks like there is no need to set a new flag here?

YoGhurt111: After using immutable values in `TaikoInbox`, there are no resolver exist in `TaikoInbox`.

davidtaikocha: Or since its only used by prover, let's get the address from `ComposeVerifier` ?

YoGhurt111: I think the `ComposeVerifier` will also remove the resolve method in the future, since calling resolve for each batch verification consumes significant gas as well.

davidtaikocha: We can also make the address an immutable value in the contract, trying to avoid the flag break change at first.

davidtaikocha (approved):

mask-pp (approved):

davidtaikocha (approved):

PR #18955, opened 2025-02-18
fix(taiko-client): fix and optimize force batch proving && unit test

YoGhurt111:


packages/taiko-client/prover/proof_submitter/proof_buffer.go
```diff
@@ -17,7 +17,7 @@ var (
 type ProofBuffer struct {
 	MaxLength     uint64
 	buffer        []*producer.ProofWithHeader
-	lastUpdatedAt time.Time
+	firstItemTime time.Time
```
davidtaikocha, line 20:
```suggestion
	firstItemAt time.Time
```

PR #18944, opened 2025-02-16
test(taiko-client): add more preconfirmation P2P network tests

davidtaikocha:

PR #18957, opened 2025-02-18
fix(taiko-client): some p2p fixes

mempirate:
Fixes issue with flag parsing, `p2p.disable` has to be explicitly 1. set and 2. set to `false` in order for P2P to be initialized, but for some reason this resulted in other flags not being parsed correctly. Just checks the bool value now instead.


davidtaikocha (approved):

YoGhurt111 (approved):

davidtaikocha:
These changes should have been included in: https://github.com/taikoxyz/taiko-mono/pull/18944 , thanks for the PR!

PR #18963, opened 2025-02-19
chore(protocol): keep `provingWindow` 5h for both TEE and ZK

YoGhurt111:


packages/protocol/contracts/layer1/tiers/TierProviderBase.sol
```diff
@@ -23,33 +23,33 @@ abstract contract TierProviderBase is ITierProvider {
...
             // cooldownWindow is 240 minutes and provingWindow is 60 minutes
-            return _buildTier(LibStrings.B_TIER_TEE_ANY, 2, 240, 60);
+            return _buildTier(LibStrings.B_TIER_TEE_ANY, 2, 240, 300);
         }
 
         // ZKVM Tiers: Allowing 120 minutes for proof aggregation.
         if (_tierId == LibTiers.TIER_ZKVM_RISC0) {
             // cooldownWindow is 240 minutes and provingWindow is 120 minutes
-            return _buildTier(LibStrings.B_TIER_ZKVM_RISC0, 3, 240, 120);
+            return _buildTier(LibStrings.B_TIER_ZKVM_RISC0, 3, 240, 300);
```
davidtaikocha, line 40: so 5 hrs also matches the current ZK window deployed on mainnet, right?

YoGhurt111: No, current zk proving window is 7h, but I think 5h is enough? It only effect ourself

davidtaikocha: See we said "keep it as it is", so lets just align every window?

YoGhurt111: OK

davidtaikocha (approved):

PR #18974, opened 2025-02-19
feat(taiko-client): handle blobCreatedIn being 0, missing basefee/prevrandao/extradata, anchor removal

cyberhorsey:
- in Taikoinbox, if TaikoWrapper != 0, `blob.createdIn` is left at 0. We need to handle this by using the block number of the proposal.

- missing BaseFee in executable data
 
- missing Prevrandao in executable data

- Missing Extradata in executable data

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,6 @@ __pycache__/
 # Idea
 .idea/
 packages/protocol/snapshots/InboxTest_ProposeAndProve.json
+
+packages/taiko-client/opnode_discovery_db
+packages/taiko-client/opnode_peerstore_db
--- a/.gitignore
+++ b/.gitignore
@@ -113,7 +113,6 @@ __pycache__/
 # backwards compatibility for past contributors
 .husky
 
-
 # VSCode
 .vscode/launch.json
 .vercel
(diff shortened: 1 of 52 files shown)


YoGhurt111 (approved):

packages/taiko-client/driver/txlist_fetcher/blob.go
```diff
@@ -92,12 +93,28 @@ func (d *BlobFetcher) FetchPacaya(
 		return nil, pkg.ErrBlobUnused
 	}
 
+	var blockNum uint64
+
+	if meta.GetBlobCreatedIn().Int64() == 0 {
```
mask-pp, line 98: Will the log code be removed later?

.gitignore
```diff
@@ -121,3 +121,6 @@ __pycache__/
 # Idea
 .idea/
 packages/protocol/snapshots/InboxTest_ProposeAndProve.json
+
+packages/taiko-client/opnode_discovery_db
+packages/taiko-client/opnode_peerstore_db
```
mask-pp, line 126:
```suggestion
packages/taiko-client/opnode_peerstore_db
packages/taiko-client/opnode_p2p_priv.txt
```

mask-pp (approved):

packages/taiko-client/driver/txlist_fetcher/blob.go
cyberhorsey: we can remove all log code now tbh, I only needed it to help figure out what was wrong.

davidtaikocha (approved):

mask-pp (approved):

PR #18980, opened 2025-02-21
fix(taiko-client): avoid using empty p2pSigner when start preconf p2p server

mask-pp:


packages/taiko-client/integration_test/test_env.sh
```diff
@@ -23,6 +23,13 @@ export TEST_ACCOUNT_PRIVATE_KEY=0x7c852118294e51e653712a81e05800f419141751be58f6
 export TREASURY=0x1670010000000000000000000000000000010001
 export VERBOSITY=${VERBOSITY-3}
 
+# Preconf p2p variables.
+export PRECONFIRMATION_P2P_DISCOVERY_PATH=memory
```
davidtaikocha, line 27: why need to add these extra ENVs?

packages/taiko-client/driver/driver.go
```diff
@@ -147,10 +148,11 @@ func (d *Driver) InitFromConfig(ctx context.Context, cfg *Config) (err error) {
 
 			log.Info("P2PNode", "Addrs", d.p2pNode.Host().Addrs(), "PeerID", d.p2pNode.Host().ID())
 
-			if !reflect2.IsNil(d.Config.P2PSignerConfigs) {
-				if d.p2pSigner, err = d.P2PSignerConfigs.SetupSigner(d.ctx); err != nil {
-					return err
-				}
+			if reflect2.IsNil(d.Config.P2PSignerConfigs) {
+				return errors.New("missing P2P signer setup")
```
davidtaikocha, line 152: If this is necessary for P2P gossip, let's move this check to `config.go`?

packages/taiko-client/integration_test/test_env.sh
mask-pp: Yeah, there's no error in Ci, if we use memory there will be no more folders(opnode_discovery_db) locally.

packages/taiko-client/integration_test/test_env.sh
```diff
@@ -23,6 +23,13 @@ export TEST_ACCOUNT_PRIVATE_KEY=0x7c852118294e51e653712a81e05800f419141751be58f6
...
+export PRECONFIRMATION_P2P_PEERSTORE_PATH=memory
+export PRECONFIRMATION_P2P_PRIV_RAW=0x3eafc84ef242d517ca50dc69b3f020579b403a454e2f9c5bec1c72c0061cee44
+export PRECONFIRMATION_P2P_SEQUENCER_KEY=$L1_PROPOSER_PRIV_KEY
```
davidtaikocha, line 30:
```suggestion
```

Also don't these are necessary, can init in unit tests when necessary

packages/taiko-client/driver/driver.go
mask-pp: The `P2PSignerConfigs` is only used when starting the p2p server, maybe checking here can be simpler?

davidtaikocha: I mean in `config.go`, if p2p is enabled, we check this, since its also required, right? Its more like a flag-related issue?

YoGhurt111 (approved):

packages/taiko-client/driver/driver.go
mask-pp: Fixed at this [commit](https://github.com/taikoxyz/taiko-mono/pull/18980/commits/cffca250a3b5412c0186c3b2891678461f07f20c).

packages/taiko-client/driver/config.go
```diff
@@ -126,6 +127,11 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
 		return nil, err
 	}
 
+	// Check if the P2P signer setup is empty.
+	if c.Uint64(flags.PreconfBlockServerPort.Name) > 0 && utils.IsNil(signerConfigs) {
+		return nil, errors.New("empty P2P signer setup")
```
davidtaikocha, line 132:
```suggestion
		return nil, errors.New("`--p2p.sequencer.key` flag is required for preconfirmation server")
```

Will this error message make more sense to users?

davidtaikocha:
After think twice, we might need to change the check another way, let me think about it first

davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/pull/18981

PR #18982, opened 2025-02-21
fix(taiko-client): check if preconfirmation whitelist address is set before fetching it

davidtaikocha:
Since `--preconfirmation.whitelist` is not a required flag.

### DIFF
--- a/packages/taiko-client/cmd/flags/driver.go
+++ b/packages/taiko-client/cmd/flags/driver.go
@@ -81,8 +81,8 @@ var (
 		EnvVars:  []string{"PRECONFIRMATION_SERVER_SIGNATURE_CHECK"},
 	}
 	PreconfWhitelistAddress = &cli.StringFlag{
-		Name:     "preconfWhitelist",
-		Usage:    "PreconfWhitelist L1 `address`",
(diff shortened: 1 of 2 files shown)


packages/taiko-client/cmd/flags/driver.go
```diff
@@ -81,8 +81,8 @@ var (
 		EnvVars:  []string{"PRECONFIRMATION_SERVER_SIGNATURE_CHECK"},
 	}
 	PreconfWhitelistAddress = &cli.StringFlag{
-		Name:     "preconfWhitelist",
-		Usage:    "PreconfWhitelist L1 `address`",
+		Name:     "preconfirmation.whitelist",
+		Usage:    "PreconfWhitelist contract L1 `address`",
 		Required: false,
 		Category: driverCategory,
 		EnvVars:  []string{"PRECONF_WHITELIST"},
```
davidtaikocha, line 88:
```suggestion
		EnvVars:  []string{"PRECONFIRMATION_WHITELIST"},
```

YoGhurt111 (approved):

mask-pp (approved):

PR #18992, opened 2025-02-25
feat(taiko-client): init proving workflow for pacaya

### LINKED ISSUE #18986: feat(taiko-client): zk any for pacaya fork

YoGhurt111:
## Major Updates
- Always enable the batches aggregation(even though the aggregation size is 1, still follow the aggregation workflow).
- Introduce the pivot proof. So the proving schema would be pivot + op/tee/zk.
- Currently, there is only one submitter in existence, rather than having one submitter per tier. 
- Make `RaikoHostEndpoint` required.


packages/taiko-client/prover/prover.go
```diff
@@ -559,9 +576,14 @@ func (p *Prover) submitProofOp(proofResponse *proofProducer.ProofResponse) error
...
 // submitProofsOp performs a batch proof submission operation.
 func (p *Prover) submitProofAggregationOp(batchProof *proofProducer.BatchProofs) error {
-	submitter := p.getSubmitterByTier(batchProof.Tier)
+	var submitter proofSubmitter.Submitter
+	if batchProof.IsPacaya {
+		submitter = p.proofSubmitterPacaya
+	} else {
+		submitter = p.getSubmitterByTier(batchProof.Tier)
+	}
 	if submitter == nil {
```
mask-pp, line 585:
```suggestion
	if utils.IsNil(submitter) {
```

packages/taiko-client/prover/config.go
```diff
@@ -124,6 +124,15 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
 		}
 	}
 
+	sgxBatchSize := c.Uint64(flags.SGXBatchSize.Name)
+	zkvmBatchSize := c.Uint64(flags.ZKVMBatchSize.Name)
+	if sgxBatchSize < 1 {
+		sgxBatchSize = 1
```
mask-pp, line 130: Is it better to expose the error message to let the user know the boundary value?

mask-pp: And also is there a reasonable maximum limit value?

packages/taiko-client/prover/proof_producer/proof_producer_pacaya.go
```diff
@@ -0,0 +1,367 @@
...
+			{
+				BatchID:                batchID,
+				L1InclusionBlockNumber: meta.GetRawBlockHeight(),
+			},
+		}
+	)
+
+	g := new(errgroup.Group)
+
+	g.Go(func() error {
```
mask-pp, line 86: Is it unnecessary code?

mask-pp, line 86: Is it unnecessary code?

mask-pp (dismissed):
LGTM, just some optional carps.

packages/taiko-client/prover/config.go
davidtaikocha: For these kinds of simple checks, shall we make the flag default value to `1` and use `Action` to check it?
image

packages/taiko-client/prover/init.go
```diff
@@ -160,22 +160,117 @@ func (p *Prover) initProofSubmitters(
 			p.proofSubmittersOntake = append(p.proofSubmittersOntake, submitter)
 		}
 	}
+	return p.initPacayaProofSubmitter(txBuilder)
+}
+
+func (p *Prover) initPacayaProofSubmitter(txBuilder *transaction.ProveBlockTxBuilder) error {
+	// Init verifiers for pacaya
```
davidtaikocha, line 167:
```suggestion
	// Init verifiers for Pacaya
```

plz use `Pacaya` in all comments instead of `pacaya`.

packages/taiko-client/prover/init.go
```diff
@@ -160,22 +160,117 @@ func (p *Prover) initProofSubmitters(
 			p.proofSubmittersOntake = append(p.proofSubmittersOntake, submitter)
 		}
 	}
+	return p.initPacayaProofSubmitter(txBuilder)
+}
+
+func (p *Prover) initPacayaProofSubmitter(txBuilder *transaction.ProveBlockTxBuilder) error {
```
davidtaikocha, line 166: Plz also make sure each new method is well commented.

packages/taiko-client/prover/init.go
```diff
@@ -160,22 +160,117 @@ func (p *Prover) initProofSubmitters(
...
+	// Init verifiers for pacaya
+	var (
+		baseLevelProver proofProducer.ProofProducer
+		zkvmProducer    proofProducer.ProofProducer
+		zkVerifiers     = make(map[string]common.Address, 2)
+		proofBuffers    = make(map[string]*proofProducer.ProofBuffer, 4)
+		proofTypes      = make([]string, 0, 4)
```
davidtaikocha, line 173: Also the `4` here, looks confusing.

packages/taiko-client/prover/init.go
```diff
@@ -160,22 +160,117 @@ func (p *Prover) initProofSubmitters(
...
+	)
+	pivotVerifier, err := p.rpc.GetPivotVerifierPacaya(&bind.CallOpts{Context: p.ctx})
+	if err != nil || pivotVerifier == transaction.ZeroAddress {
+		return fmt.Errorf("required pivot verifier not found: %w", err)
```
davidtaikocha, line 177: When `pivotVerifier == transaction.ZeroAddress` + `err == nil`, the output will be `"required pivot verifier not found: nil"`?

packages/taiko-client/prover/config.go
YoGhurt111: Since the default value of these 2 flags is 0, but the value 0 won't work after Pacaya fork, I don't want to make breaking change for user's config, not just for checks.
If there are some better ways to change the default value, let's choose to use it.

packages/taiko-client/prover/init.go
YoGhurt111: Sounds good

packages/taiko-client/prover/init.go
YoGhurt111: Sounds good

packages/taiko-client/prover/init.go
```diff
@@ -160,22 +160,117 @@ func (p *Prover) initProofSubmitters(
...
 	}
+	return p.initPacayaProofSubmitter(txBuilder)
+}
+
+func (p *Prover) initPacayaProofSubmitter(txBuilder *transaction.ProveBlockTxBuilder) error {
+	// Init verifiers for pacaya
+	var (
+		baseLevelProver proofProducer.ProofProducer
+		zkvmProducer    proofProducer.ProofProducer
+		zkVerifiers     = make(map[string]common.Address, 2)
```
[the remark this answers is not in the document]
YoGhurt111: Sounds good

packages/taiko-client/prover/init.go
YoGhurt111: Yes, I think the error message is optional, if `pivotVerifier == transaction.ZeroAddress`, I think the first part(`required pivot verifier not found`) already explains the reason

packages/taiko-client/prover/init.go
YoGhurt111: Sounds good

packages/taiko-client/prover/config.go
davidtaikocha: Is the code the same to set the default value to `1`?

packages/taiko-client/prover/proof_producer/proof_producer_pacaya.go
YoGhurt111: I have no idea what is your meaning about "unnecessary", could you provide a specific explanation?

packages/taiko-client/prover/init.go
```diff
@@ -168,8 +168,8 @@ func (p *Prover) initPacayaProofSubmitter(txBuilder *transaction.ProveBlockTxBui
 	var (
 		baseLevelProver proofProducer.ProofProducer
 		zkvmProducer    proofProducer.ProofProducer
-		zkVerifiers     = make(map[string]common.Address, 2)
-		proofBuffers    = make(map[string]*proofProducer.ProofBuffer, 4)
+		zkVerifiers     = make(map[string]common.Address, proofSubmitter.MaxNumSupportedZkTypes)
+		proofBuffers    = make(map[string]*proofProducer.ProofBuffer, proofSubmitter.MaxNumSupportedProofTypes)
 		proofTypes      = make([]string, 0, 4)
```
davidtaikocha, line 173: Shall we also replace the `4` here with a constant?

packages/taiko-client/prover/config.go
```diff
@@ -124,6 +124,15 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
...
+	}
+	if zkvmBatchSize < 1 {
+		zkvmBatchSize = 1
+	}
```
davidtaikocha, line 134: These can be removed since https://github.com/taikoxyz/taiko-mono/commit/657ef511f08b2ae1a37fc0bd32ea6857f6eeb48c ?

YoGhurt111: sure

packages/taiko-client/prover/init.go
YoGhurt111: sure

packages/taiko-client/prover/proof_producer/dummy_producer.go
```diff
@@ -19,23 +19,41 @@ func (o *DummyProofProducer) RequestProof(
...
+			ProofResponses: proofs,
+			BatchProof:     bytes.Repeat([]byte{0xbb}, 100),
+			Tier:           tier,
+		}, nil
+	} else {
+		return &BatchProofs{
+			ProofResponses: proofs,
+			BatchProof:     bytes.Repeat([]byte{0xbb}, 100),
+			ProofType:      proofType,
+		}, nil
```
davidtaikocha, line 57: Shall we just just use:
```go
 return &BatchProofs{
			ProofResponses: proofs,
			BatchProof:     bytes.Repeat([]byte{0xbb}, 100),
			ProofType:      proofType,
		}, nil
```

packages/taiko-client/prover/proof_producer/pivot_proof_producer.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/bindings/metadata"
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/internal/metrics"
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/pkg/rpc"
+)
+
+// PivotProofProducer generates a pivot proof for the given block.
+// TODO: Can be refactored like ProofProducerPacaya
+type PivotProofProducer struct {
+	Verifier            common.Address
+	RaikoHostEndpoint   string // a proverd RPC endpoint
```
davidtaikocha, line 27: Not sure what's the `proverd` here

packages/taiko-client/prover/proof_producer/pivot_proof_producer.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/ethereum/go-ethereum/log"
+
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/bindings/encoding"
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/bindings/metadata"
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/internal/metrics"
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/pkg/rpc"
+)
+
+// PivotProofProducer generates a pivot proof for the given block.
+// TODO: Can be refactored like ProofProducerPacaya
```
davidtaikocha, line 24: Still a TODO atm?

packages/taiko-client/prover/proof_producer/proof_producer_pacaya.go
```diff
@@ -0,0 +1,370 @@
...
+		if _, err := z.PivotProducer.RequestProof(ctx, opts, batchID, meta, requestAt); err != nil {
+			return err
+		}
+		return nil
```
davidtaikocha, line 90: shall we simply change to `return z.PivotProducer.RequestProof(ctx, opts, batchID, meta, requestAt)`?

packages/taiko-client/prover/proof_producer/proof_producer_pacaya.go
```diff
@@ -0,0 +1,370 @@
...
+	})
+	g.Go(func() error {
+		if z.Dummy {
+			proofType = z.ProofType
+			resp, _ := z.DummyProofProducer.RequestProof(opts, batchID, meta, z.Tier(), requestAt)
```
davidtaikocha, line 95: As a good practice, better to check the `error` check, in case someone modify `z.DummyProofProducer.RequestProof` later.

packages/taiko-client/prover/proof_producer/pivot_proof_producer.go
YoGhurt111: The logic handling of PivotProofProducer is very similar to the logic handling of ProofProducerPacaya, but there are some slight differences in detail. I haven't figured out how to optimize this code yet, so I'll add a TODO here for now.
This doesn't affect the code merge

packages/taiko-client/prover/proof_producer/pivot_proof_producer.go
YoGhurt111: fixed

packages/taiko-client/prover/proof_producer/dummy_producer.go
YoGhurt111: Sure

packages/taiko-client/prover/proof_producer/proof_producer_pacaya.go
YoGhurt111: No, actually `Dummy=true` in `ProofProducerPacaya` is only for `ProofType = op`.
It means that it always request real proof even if the `Dummy=true` when the `ProofType = sgx`.
Maybe it is a better practice that change the name.

packages/taiko-client/prover/proof_producer/proof_producer_pacaya.go
YoGhurt111: Get it!

davidtaikocha (dismissed):

mask-pp (approved):

smtmfft (approved):

PR #19009, opened 2025-02-28
fix(taiko-client): fix timestamp lowerbound

smtmfft:


packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/pacaya.go
```diff
@@ -150,12 +150,16 @@ func (i *BlocksInserterPacaya) InsertBlocks(
 		}
 
 		// Validate the timestamp.
-		timestampUpperBound := meta.GetProposedAt() - maxAnchorHeightOffset*12
-		if timestampUpperBound < parent.Time {
-			timestampUpperBound = parent.Time
+		timestampUpperBound := meta.GetProposedAt()
```
mask-pp, line 153: LGTM, and maybe this way can be simpler:
`
timestamp = max(timestamp, meta.GetProposedAt()-maxAnchorHeightOffset*12)
timestamp = min(timestamp, meta.GetProposedAt())
`

mask-pp (approved):

davidtaikocha (approved):

PR #19020, opened 2025-03-03
chore(protocol): add back ERC20Vault (without solver) as ERC20VaultOriginal

dantaik:
To mitigate the risk in the solver feature, we should keep the original ERC20Vault code as-is so on Taiko mainnet, we will use the original code.


dantaik:
The diff between the code in this PR (left) and the one used by our mainnet (right):

Screenshot 2025-03-03 at 11 23 30 AM
Screenshot 2025-03-03 at 11 23 36 AMScreenshot 2025-03-03 at 11 23 44 AM

YoGhurt111 (approved):

xiaodino (approved):

PR #19022, opened 2025-03-03
fix(protocol): erc20 vault solver logic

AnshuJalan:


dantaik:
@AnshuJalan May I propose a different fix, which implies smaller change:

-Make `BridgeTransferOp` only "solvable" if its `solverFee` is non-zero otherwise, when the message is invoked, ignore the solver's operation entirely.
- In `onMessageInvocation`, replace ` to.sendEtherAndVerify` with `to.sendEther` to allow it to fail when the `BridgeTransferOp` is solveable and was solved by a solver.

PR #19026, opened 2025-03-04
fix(protocol): fix Ether surplus transfer issue

dantaik:
- Make `BridgeTransferOp` only _solvable_ if its `solverFee` is non-zero; otherwise, when the message is invoked, ignore the solver's operation entirely.
- In `onMessageInvocation`, replace ` to.sendEtherAndVerify` with `to.sendEther` to allow it to fail when the `BridgeTransferOp` is solved by a solver.


packages/protocol/test/shared/tokenvault/ERC20Vault.t.sol
```diff
@@ -244,182 +244,6 @@ contract TestERC20Vault is CommonTest {
         assertEq(David.balance, etherAmount);
     }
 
-    function test_20Vault_receive_erc20_solved() public {
```
dantaik, line 247: these tests are duplicate

AnshuJalan (approved):

mask-pp (approved):

davidtaikocha (approved):

PR #19041, opened 2025-03-05
feat(taiko-client): update preconf API signature verification

davidtaikocha:
ref: https://www.notion.so/taikoxyz/Signatures-in-current-taiko-client-preconfirmation-implementation-1ad9673143d680bc9cb0c15d498b6884

Instead of checking if the `payload` is signed by `block.coinbase`, we change to check if it's signed by `PreconfWhitelist.getOperatorForCurrentEpoch()` or `PreconfWhitelist.getOperatorForNextEpoch()`.

### DIFF
--- a/packages/taiko-client/driver/preconf_blocks/api.go
+++ b/packages/taiko-client/driver/preconf_blocks/api.go
@@ -22,7 +22,7 @@ import (
 )
 
 // ValidateSignature validates the signature of the request body.
-func (b *BuildPreconfBlockRequestBody) ValidateSignature() (bool, error) {
+func (b *BuildPreconfBlockRequestBody) ValidateSignature(addresses []common.Address) (bool, error) {
 	payload, err := rlp.EncodeToBytes(b.ExecutableData)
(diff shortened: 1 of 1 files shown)


YoGhurt111 (approved):

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -33,7 +33,19 @@ func (b *BuildPreconfBlockRequestBody) ValidateSignature() (bool, error) {
 		return false, err
 	}
 
-	return crypto.PubkeyToAddress(*pubKey).Hex() == b.ExecutableData.FeeRecipient.Hex(), nil
+	// NOTE: right now P2PSequencerAddresses() may will return empty list, if
+	// there is no `PreconfWhitelist` contract deployed, if so, we skip the
+	// signature check temporarily.
+	if len(addresses) == 0 {
```
mask-pp, line 39: If so, we also need to verify the signature.

PR #19043, opened 2025-03-05
fix(taiko-client): skip `headL1Origin` check when chain only has the genesis block

davidtaikocha:


YoGhurt111 (approved):

mskrzypkows:
Shouldn't `rpc.L2Engine.SetHeadL1Origin(ctx, meta.L1Origin.BlockID);` be called somewhere while doing `/preconfBlocks`?

mskrzypkows (approved):
Verified, works

PR #19050, opened 2025-03-07
chore(protocol): stop processing transitions once a conflict is identified

dantaik:


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -294,6 +294,8 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
 
                     // Invalidate the conflict transition
                     state.transitions[slot][tid].blockHash = 0;
+                    metasLength = i + 1;
```
dantaik, line 297: We also need to change the length of `ctxs`  as `ctxs` is used by `verifyProof`.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -294,6 +294,11 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
+                    assembly {
+                        mstore(ctxs, metasLength)
```
dantaik, line 299: However, changing the length of `ctxs` may actually make the proof invalid as the proof verifier may expect that the ctx will not be changed by the code here. So breaking out here in general maybe an issue for proof verifiers.

I think we should drop this PR.

PR #19087, opened 2025-03-13
fix(protocol): enforce last block time shift to be 0

dantaik:


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -171,9 +171,17 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
 
             require(info_.anchorBlockHash != 0, ZeroAnchorBlockHash());
 
-            info_.lastBlockId = stats2.numBatches == config.forkHeights.pacaya
-                ? stats2.numBatches + uint64(params.blocks.length) - 1
-                : lastBatch.lastBlockId + uint64(params.blocks.length);
+            uint96 livenessBond;
```
dantaik, line 174: you can ignore the change here, it is done to avoid stack too deep issue.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -778,10 +785,11 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
                 : _params.lastBlockTimestamp;
 
             require(lastBlockTimestamp_ <= block.timestamp, TimestampTooLarge());
+            require(_params.blocks[numBlocks - 1].timeShift == 0, LastBlockTimeShiftNotZero());
```
dantaik, line 788: this is the real change.

YoGhurt111 (approved):

PR #19090, opened 2025-03-13
feat(protocol): enforce first block time shift to be 0

dantaik:


packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -22,10 +22,9 @@ interface ITaikoInbox {
         // transactions in calldata or blobs, the block will contains as many transactions as
         // possible.
         uint16 numTransactions;
-        // For the first block in a batch,  the block timestamp is the batch params' `timestamp`
-        // plus this time shift value;
-        // For all other blocks in the same batch, the block timestamp is its parent block's
-        // timestamp plus this time shift value.
+        // The time difference (in seconds) between the timestamp of this block and
+        // the timestamp of the previous block in the same batch. For the first block in a batch,
+        // there is not previous block in the same batch, so the time shift must be 0.
```
linoscope:
```suggestion
        // there is no previous block in the same batch, so the time shift must be 0.
```

(Actually, this typo is from my PR :sweat_smile: )

dantaik, line 27:
```suggestion
        // there is no previous block in the same batch, so the time shift must be 0.
```

davidtaikocha (approved):

YoGhurt111 (approved):

davidtaikocha (approved):

PR #19093, opened 2025-03-13
chore(protocol): fix an ERC20Vault test

dantaik:


packages/protocol/test/shared/tokenvault/ERC20Vault.t.sol
```diff
@@ -587,14 +587,15 @@ contract TestERC20Vault is CommonTest {
...
 
         uint256 aliceBalanceBefore = BridgedERC20(bridgedTokenAddr).balanceOf(Alice);
-        console.log("Alice amount before = %s", aliceBalanceBefore);
+        // console.log("Alice amount before = %s", aliceBalanceBefore);
 
         // Define the amounts
         uint256 amount = 1e18;
         uint256 solverFee = 2e18;
 
         // Approve only the amount, not the solver fee
```
AnshuJalan, line 596: I think the comment needs to be removed because now the entire amount is being approved.

PR #19100, opened 2025-03-17
refactor(protocol): improve LibPreconfUtils.getBeaconBlockRoot()

dantaik:
Two changes:

- make sure we try to query up to 32 times
- break when the target timestamp is greater than `block.timestamp`


mask-pp (approved):

xiaodino (approved):

AnshuJalan (approved):

davidtaikocha (approved):

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
```diff
@@ -6,18 +6,27 @@ import "./LibPreconfConstants.sol";
...
+    /// @dev Caller should verify the returned value is not 0.
     /// @param timestamp The timestamp for which the beacon block root is to be retrieved.
     /// @return The beacon block root as a bytes32 value.
+
     function getBeaconBlockRoot(uint256 timestamp) internal view returns (bytes32) {
+        if (timestamp < LibPreconfConstants.getGenesisTimestamp(block.chainid)) {
+            return bytes32(0);
+        }
         uint256 targetTimestamp = timestamp + LibPreconfConstants.SECONDS_IN_SLOT;
-        while (true) {
```
dantaik, line 18: @AnshuJalan I think I got myself confused by the loop in this function. According to https://eips.ethereum.org/EIPS/eip-4788, the implementation should be as simply as: 

```solidity
    function getBeaconBlockRoot(uint256 timestamp) internal view returns (bytes32) {
        (bool success, bytes memory result) =
            LibPreconfConstants.getBeaconBlockRootContract().staticcall(abi.encode(timestamp));

        if (success && result.length == 32) {
            return abi.decode(result, (bytes32));
        } else {
            return bytes32(0); // or revert
        }
    }

```

YoGhurt111 (approved):

mask-pp (approved):

YoGhurt111 (approved):

smtmfft (approved):

YoGhurt111 (approved):

davidtaikocha (approved):

YoGhurt111 (approved):

mask-pp (approved):

davidtaikocha (approved):

PR #19106, opened 2025-03-17
refactor(taiko-client): introduce `requestHttpProof` to remove duplicated code in `proofProducer` package

davidtaikocha:
Also added a simple test for `proofBuffer`


mask-pp (approved):

packages/taiko-client/prover/proof_producer/proof_producer_pacaya.go
```diff
@@ -75,6 +70,7 @@ func (z *ProofProducerPacaya) RequestProof(
 	)
 
 	g.Go(func() error {
+		// NOTE: right now we don't use the pivot proof for Pacaya, since its still not ready.
```
davidtaikocha, line 73: @YoGhurt111 Plz correct me if i'm wrong

YoGhurt111 (approved):

packages/taiko-client/prover/proof_producer/proof_producer_pacaya.go
YoGhurt111: Close! We don't need single pivot proof for pacaya but in design we need to tell Raiko to generate the pivot proof.
So when it is not ready in Raiko's end, we need to set `Dummy` true in `PivotProducer` to generate the optimizm proof and set the pivot verifier op impl to accept the optimizm proof.

PR #19113, opened 2025-03-18
feat(protocol): multiple improvements for PreconfWhitelist2 

dantaik:
BEGIN_COMMIT_OVERRIDE
feat(protocol): allow owner to change "operator change delay" (even to zero)
feat(protocol): allow operators to remove themselves from the whitelist
refactor(protocol): optimize getOperatorForCurrentEpoch gas cost
END_COMMIT_OVERRIDE


smartprogrammer93 (dismissed):

YoGhurt111 (approved):

smtmfft (approved):

YoGhurt111 (approved):

linoscope:
Right now, only the operator themselves can remove themself from the lookahead without delay easily. For the whitelist owner to emergency-exit the operator, it requires 3 txs:
- Set the delay to 0.
- Remove the operator.
- Re-set the delay to 2.

What do you think of adding an `emergencyExitOperator` function that the whitelist owner can call so the whitelist owner can also exit preconf operators without delay with a single tx?

The use case I have in mind is that the whitelist owner monitors the preconfs from the outside and can force-exit preconfers who are not preconfing for some period of time.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
```diff
@@ -16,61 +16,62 @@ contract PreconfWhitelist2 is EssentialContract, IPreconfWhitelist {
...
+    }
+
+    //// @notice Allows the caller to remove themselves as an operator immediately.
+    function removeSelf() external {
+        _removeOperator(msg.sender, 0);
     }
 
     /// @notice Consolidates the operator mapping by removing operators whose removal epoch has
     /// passed, swapping removed operators with the last entry, and decrementing the operatorCount.
     function consolidate() external {
```
linoscope, line 65: Can't this function be called by anyone, and as a result, potentially change the return value of `getOperatorForCurrentEpoch()` by changing the ordering of the `operators`?

linoscope: I think we want to swap the removed operator with the next operator in the `operators` array (not the last), but that requires shifting a lot of indexes..

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
```diff
@@ -124,12 +125,43 @@ contract PreconfWhitelist2 is EssentialContract, IPreconfWhitelist {
...
+    function epochStartTimestamp(uint256 _offset) public view returns (uint64) {
         return uint64(
-            LibPreconfUtils.getEpochTimestamp() + offset * LibPreconfConstants.SECONDS_IN_EPOCH
+            LibPreconfUtils.getEpochTimestamp() + _offset * LibPreconfConstants.SECONDS_IN_EPOCH
         );
     }
 
+    function _addOperator(address _operator, uint8 _operatorChangeDelay) internal {
+        require(_operator != address(0), InvalidOperatorAddress());
+        require(operators[_operator].activeSince == 0, OperatorAlreadyExists());
```
AnshuJalan, line 136: A removed operator cannot be re-added since `activeSince` for a removed operator is not reset. Is this intended?

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
AnshuJalan: ^ I have a similar concern

dantaik: > ~I think we want to swap the removed operator with the next operator in the `operators` array (not the last), but that requires shifting a lot of indexes..~ Edit: Well, even then the shifted indexes can change the result of `getOperatorForCurrentEpoch`

Good catch.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
dantaik: This is another good catch, I'll need to add a test for it. We should allow adding a removed operator back.

smartprogrammer93: It can be added after calling consolidate

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
dantaik: Fixed by new commits. Please take another look.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
dantaik: Fixed by new commits. Please take another look.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
```diff
@@ -16,82 +16,82 @@ contract PreconfWhitelist2 is EssentialContract, IPreconfWhitelist {
...
+                    address nextOperator = operatorMapping[j + 1];
+                    operators[nextOperator].index = j;
+                    operatorMapping[j] = nextOperator;
                 }
-                // Remove the operator from the mapping
+                // Remove the last operator as it has been shifted
                 delete operators[operator];
-                delete operatorMapping[lastIndex];
+                delete operatorMapping[_operatorCount - 1];
                 _operatorCount--;
```
AnshuJalan, line 86: Can change this to `--operatorCount`

AnshuJalan (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
linoscope: Maybe I am missing something, but doesn't the issue still exist, even if we preserve the ordering of the operators due to the `uint8(uint256(root) % operatorCount)` changing? For example:

```
Before consolidate():
- operatorCount = 4
- operatorMapping = [A, B, C, D]
- root % 4 = 1 (selects B)

After consolidate() removes A:
- operatorCount = 3
- operatorMapping = [B, C, D]
- root % 3 = 1 (selects C)
```

dantaik:
> Right now, only the operator themselves can remove themself from the lookahead without delay easily. For the whitelist owner to emergency-exit the operator, it requires 3 txs:
> 
> * Set the delay to 0.
> * Remove the operator.
> * Re-set the delay to 2.
> 
> What do you think of adding an `emergencyExitOperator` function that the whitelist owner can call so the whitelist owner can also exit preconf operators without delay with a single tx?
> 
> The use case I have in mind is that the whitelist owner monitors the preconfs from the outside and can force-exit preconfers who are not preconfing for some period of time.

Added `function removeOperator(address _operator, bool _effectiveImmediately) external onlyOwner `

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
linoscope: If we find it too complex to fix this, I think it's fine to keep the removed operators in the array; removals should be rare, cost-wise they only add one additional loop+check, and the whitelist is a temporal solution anyway.

dantaik: Yes, still a bug here...

dantaik: Fixed, but the gas cost is almost linear to `operatorCount`, haven't yet figured out a way to improve it.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
```diff
@@ -124,12 +142,45 @@ contract PreconfWhitelist2 is EssentialContract, IPreconfWhitelist {
...
+
+        uint64 inactiveSince = epochStartTimestamp(_operatorChangeDelay);
+        operators[_operator].inactiveSince = inactiveSince;
+        operators[_operator].activeSince = 0;
+
+        emit OperatorRemoved(_operator, inactiveSince);
+    }
+
+    /// @dev The cost of this function is primarily linear with respect to operatorCount.
     function _getOperatorForEpoch(uint64 _epochTimestamp) internal view returns (address) {
```
dantaik, line 184: open for gas-efficiency improvement ideas.

packages/protocol/test/layer1/preconf/whitelist/PeconfWhitelist2.t.sol
```diff
@@ -194,6 +229,72 @@ contract TestPreconfWhitelist2 is Layer1Test {
...
+
+        whitelistNoDelay.consolidate();
+        candidates = whitelistNoDelay.getOperatorCandidatesForCurrentEpoch();
+        assertEq(candidates.length, 2);
+        assertEq(candidates[0], Bob);
+        assertEq(candidates[1], Carol);
+
+        vm.stopPrank();
+    }
+
```
linoscope, line 297: Suggestion of an additional test:

```suggestion

    function test_whitelistNoDelay_consolidationWillNotChangeCurrentEpochOperator() external {
        _setBeaconBlockRoot(bytes32(uint256(5)));

        vm.startPrank(whitelistOwner);
        whitelistNoDelay.addOperator(Alice);
        whitelistNoDelay.addOperator(Bob);
        whitelistNoDelay.addOperator(Carol);
        whitelistNoDelay.addOperator(David);
        whitelistNoDelay.removeOperator(Alice);

        address operatorBeforeConsolidate = whitelistNoDelay.getOperatorForCurrentEpoch();
        whitelistNoDelay.consolidate();
        address operatorAfterConsolidate = whitelistNoDelay.getOperatorForCurrentEpoch();
        assertEq(operatorBeforeConsolidate, operatorAfterConsolidate);

        vm.stopPrank();
    }
```

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
linoscope: The added cost in `getOperatorForEpoch` makes me wonder if `consolidate()` is really worth it. Alternative being just keeping the removed in the array, as mentioned above

> I think it's fine to keep the removed operators in the array; removals should be rare, cost-wise they only add one additional loop+check, and the whitelist is a temporal solution anyway.

linoscope: Keeping `consolidate()` means we get `O(operatorCount)` in `getOperatorForEpoch` while removing `consolidate()` and keeping out-dated operators in array means `O(removedOperatorCount)` in worst case. So if we have 3 removed operators it becomes worth it in terms of cost complexity.

linoscope: Anyway, think we can revisit if cost becomes an issue and do not need to be a blocker right now.

smartprogrammer93 (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist2.sol
dantaik: I will optimize the gas cost in another PR to make the cost O(1)

linoscope:
LGTM

YoGhurt111 (approved):

mask-pp (approved):

smtmfft (approved):

davidtaikocha (approved):

mask-pp (approved):

PR #19133, opened 2025-03-20
feat(preconf-monitor): preconf monitor + consolidator

cyberhorsey:


cyberhorsey:
done in rust via ejector

PR #19135, opened 2025-03-21
feat(taiko-client): make `prover.dummy` available for producer again

YoGhurt111:


packages/taiko-client/prover/init.go
```diff
@@ -193,7 +195,7 @@ func (p *Prover) initPacayaProofSubmitter(txBuilder *transaction.ProveBlockTxBui
 		Verifier:            pivotVerifierAddress,
 		RaikoHostEndpoint:   p.cfg.RaikoHostEndpoint,
 		JWT:                 p.cfg.RaikoJWT,
-		Dummy:               p.cfg.Dummy,
+		Dummy:               true, // Note: since the pivot proof is not ready, so we use p proof instead
```
davidtaikocha, line 198:
```suggestion
		Dummy:               true, // Note: since the pivot proof is not ready, so we use dummy proof instead
```

davidtaikocha (approved):

mask-pp (approved):

PR #19148, opened 2025-03-24
fix(taiko-client): buffer size & error check

YoGhurt111:


mask-pp (approved):

davidtaikocha (approved):

packages/taiko-client/prover/init.go
```diff
@@ -124,13 +124,13 @@ func (p *Prover) initProofSubmitters(
 					RaikoRequestTimeout: p.cfg.RaikoRequestTimeout,
 					Dummy:               p.cfg.Dummy,
 				}
-				bufferSize = 0
+				bufferSize = 1
```
davidtaikocha, line 127: Shall we add a comment to explain why it should be `1`?

davidtaikocha (changes requested):

packages/taiko-client/prover/init.go
YoGhurt111: Now, we change it to read the config from flag.

davidtaikocha (approved):

PR #19166, opened 2025-03-27
fix(taiko-client): fix a bug in `syncer.checkReorg()`

davidtaikocha:


mask-pp (approved):

YoGhurt111 (approved):

packages/taiko-client/driver/chain_syncer/blob/syncer.go
```diff
@@ -475,12 +475,9 @@ func (s *Syncer) checkReorg(ctx context.Context, blockID *big.Int) (*rpc.ReorgCh
...
 	}
 
-	if reorgCheckResult == nil {
-		// 2. Parent block
-		reorgCheckResult, err = s.rpc.CheckL1Reorg(
-			ctx,
-			new(big.Int).Sub(blockID, common.Big1),
-		)
+	// 2. If the verified block check is passed, we check the parent block.
+	if reorgCheckResult == nil || !reorgCheckResult.IsReorged {
```
mask-pp, line 479: Is the `reorgCheckResult == nil ` case imposiable?

davidtaikocha: Yes, just for safety, as suggested in https://github.com/taikoxyz/taiko-mono/pull/19166#discussion_r2015343862

PR #19172, opened 2025-03-28
fix(taiko-client): check `--prover.proveUnassignedBlocks` flag value before proving expired blocks

davidtaikocha:


packages/taiko-client/prover/event_handler/block_proposed.go
```diff
@@ -340,6 +340,21 @@ func (h *BlockProposedEventHandler) checkExpirationAndSubmitProofOntake(
 		tier = encoding.TierGuardianMinorityID
 	}
 
+	// If the current prover is not the assigned prover, and `--prover.proveUnassignedBlocks` is not set,
+	// we should skip proving this block.
+	if !h.proveUnassignedBlocks &&
```
YoGhurt111, line 345: We should also add this check in `checkExpirationAndSubmitProofPacaya`.

davidtaikocha: Added

YoGhurt111 (approved):

mask-pp (approved):

Issue #19160, opened 2025-03-25
Simplifying Base Fee Calculation and Transitioning to Node-Based Computation


dantaik:
## Background

In Taiko, the L2 base fee is determined by the protocol contract within TaikoAnchor's anchor transactions. The current mathematical approach is a derivative of Vitalik's concept of using an AMM curve for base fee calculations. However, the existing smart contract implementation is intricate and challenging to comprehend.

## Proposed Simplified Mathematics

Below is the Ethereum EIP-1559 formula, where Ethereum's gas target presumes a block time of 12 seconds:

$$
B_{\text{current}} = B_{\text{parent}} + B_{\text{parent}} \times \left( \frac{G_{\text{parent}}}{G_{\text{target}}} - 1 \right) \times \frac{1}{D}
$$

To accommodate blockchains with variable block times, we can generalize the formula as follows:

$$
B_{\text{current}} = B_{\text{parent}} + B_{\text{parent}} \times \left( \frac{G_{\text{parent}}}{G_{\text{target}}} - \frac{T_{\text{parent}}}{12} \right) \times \frac{1}{D}
$$

It is crucial to recognize that this generalized formula incentivizes block builders to create larger blocks rather than splitting them into smaller ones, even when a portion of the base fee is given to the block builder. This can be demonstrated through the following example:

### Example: Comparing One Large Block vs Two Smaller Ones

Consider a scenario where the parent block has a base fee of 100 gwei, a gas target of 10,000,000 gas, and a block time of 12 seconds. 

#### Single Block Scenario
- Gas used: 10,000,000
- Base fee calculation:

$$
  B_{\text{current}} = 100 + 100 \times \left( \frac{10,000,000}{10,000,000} - \frac{0}{12} \right) \times \frac{1}{8} = 100 + 100 \times \left( 1 - 0 \right) \times \frac{1}{8} = 100 + 12.5 = 112.5 \text{ gwei}
$$

Total base fee collected: 112.5 gwei × 10,000,000 gas = 1,125,000,000 gwei

#### Split Block Scenario
Now, let's split this block into two smaller blocks with zero block time between them.

**Block 1:**
- Gas used: 5,000,000
- Base fee calculation:

$$
[...]

PR #19183, opened 2025-03-31
feat(protocol): simplify base fee calculation & refactor TaikoAnchor

dantaik:
The PR include the following changes:

- Replaces the existing base fee calculation with the last equation from GitHub i[ssue #19160](https://github.com/taikoxyz/taiko-mono/issues/19160#issuecomment-2766021846) to reduce complexity and minimize potential bugs. The new approach is significantly simpler.

- Refactors code organization by splitting fork-specific logic into separate base contracts, making it easier to add support for new forks in the future.

I'll follow up with simulations using Taiko mainnet blocks to verify the change's impact. Parameter adjustments may also follow based on simulation results.

## TODOs

- [x] write unit tests (to verify there is no math overflow) - make sure we use the same types in the simulation instead of using float/double. 
- [ ] simulate the new approach with Taiko mainnet data.


xiaodino (approved):

dantaik:
@AnshuJalan I won't merge this PR without your approval :D

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
```diff
@@ -0,0 +1,98 @@
...
+    )
+        external
+        nonZeroBytes32(_anchorStateRoot)
+        nonZeroValue(_anchorBlockId)
+        nonZeroValue(_baseFeeConfig.gasIssuancePerSecond)
+        nonZeroValue(_baseFeeConfig.adjustmentQuotient)
+        onlyGoldenTouch
+        nonReentrant
+    {
+        require(shastaForkHeight != 0 && block.number >= shastaForkHeight, L2_FORK_ERROR());
```
davidtaikocha, line 57: Shall we change it to `require(shastaForkHeight == 0 || block.number >= shastaForkHeight, L2_FORK_ERROR());` to allow Shasta fork from genesis? Since in `ShastaAnchor.constructor `, `shastaForkHeight == 0` is allowed.

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
```diff
@@ -0,0 +1,98 @@
...
+
+    constructor(
+        address _resolver,
+        address _signalService,
+        uint64 _pacayaForkHeight,
+        uint64 _shastaForkHeight
+    )
+        PacayaAnchor(_resolver, _signalService, _pacayaForkHeight)
+    {
+        require(shastaForkHeight == 0 || shastaForkHeight > _pacayaForkHeight, InvalidForkHeight());
```
davidtaikocha, line 26: When `shastaForkHeight` is `0`, shall we also enforce `_pacayaForkHeight` and `_ontakeForkHeight ` to be `0`?

dantaik: Please check the last commit https://github.com/taikoxyz/taiko-mono/pull/19183/commits/6b21053e7f4bb2ec2bf7e564c43f08fb387ed2cb

dantaik: If shastaForkHeight == 0, it means shasta fork is not schedule to be active, it has to be 1 if we want to use shasta right after genesis.

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
dantaik: If shastaForkHeight == 0, it means shasta fork is not schedule to be active, it has to be 1 if we want to use shasta right after genesis.

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
dantaik: If shastaForkHeight == 0, it means shasta fork is not schedule to be active,  but _pacayaForkHeight may be scheduled to be active or is active already.

Due to historical reasons, there is no _ontakeForkHeight on L2.

mask-pp (approved):

PR #19197, opened 2025-04-02
feat(protocol): deploy version 1.12.1 address manager

YoGhurt111:


packages/protocol/contracts/shared/bridge/OnlyOwnerBridge.sol
```diff
@@ -0,0 +1,708 @@
...
+import "../signal/ISignalService.sol";
+import "./IBridge.sol";
+import "./IQuotaManager.sol";
+
+/// @title OnlyOwnerBridge
+/// @notice See the documentation for {IBridge}.
+/// @dev Labeled in AddressResolver as "bridge". Additionally, the code hash for the same address on
+/// L1 and L2 may be different.
+/// @custom:security-contact security@taiko.xyz
+contract OnlyOwnerBridge is EssentialContract, IBridge {
```
davidtaikocha, line 19: Could u comment whats the difference between `OnlyOwnerBridge` and `Bridge`?

YoGhurt111: OK

packages/protocol/script/layer1/UpgradeBridgeL2.s.sol
```diff
@@ -0,0 +1,50 @@
...
+    function run() external broadcast {
+        Multicall3.Call3[] memory calls = new Multicall3.Call3[](3);
+        // bridge upgrades to onlyOwnerBridgeImpl
+        calls[0].target = 0x1670090000000000000000000000000000000001;
+        calls[0].allowFailure = false;
+        calls[0].callData = abi.encodeCall(UUPSUpgradeable.upgradeTo, (onlyOwnerBridgeImpl));
+        // init to store new sharedAddressManager
+        calls[1].target = 0x1670090000000000000000000000000000000001;
+        calls[1].allowFailure = false;
+        calls[1].callData = abi.encodeCall(Bridge.init, (delegateOwner, sharedAddressManager));
```
davidtaikocha, line 34: If I remember correctly, `Bridge.init` can only be called once per proxy.

packages/protocol/script/layer1/UpgradeBridgeL2.s.sol
```diff
@@ -0,0 +1,50 @@
...
+        // bridge upgrades to original bridgeImpl
+        calls[2].target = 0x1670090000000000000000000000000000000001;
+        calls[2].allowFailure = false;
+        calls[2].callData = abi.encodeCall(UUPSUpgradeable.upgradeTo, (bridgeImpl));
+
+        DelegateOwner.Call memory dcall = DelegateOwner.Call({
```
davidtaikocha, line 40: Shall we also upgrade vaults?

packages/protocol/script/layer1/UpgradeBridgeL2.s.sol
YoGhurt111: I upgraded the proxy to `onlyOwnerBridgeImpl` to avoid the limit, then call init().

packages/protocol/script/layer1/UpgradeBridgeL2.s.sol
YoGhurt111: Yes

davidtaikocha: And lets also add the calls to downgrade proxy impls here?

YoGhurt111: I think there is no need to downgrade.

PR #19201, opened 2025-04-03
fix(protocol): change modifier to `onlyOwner`

YoGhurt111:
The reasons for changing the modifier for `deleteInstances` to `onlyOwner` are:
1. Since the contract of `SgxVerifier` don't have resolver, when the `deleteInstances` function is called, it will revert.
2. The address  of `B_SGX_WATCHDOG` on Mainnet is guardian prover contract, since after pacaya we don't support gp at all, so remove the logic.

### DIFF
--- a/packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
+++ b/packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
@@ -115,10 +115,7 @@ contract SgxVerifier is EssentialContract, IVerifier {
 
     /// @notice Deletes SGX instances from the registry.
     /// @param _ids The ids array of SGX instances.
-    function deleteInstances(uint256[] calldata _ids)
-        external
-        onlyOwner
-    {
+    function deleteInstances(uint256[] calldata _ids) external onlyOwner {
         uint256 size = _ids.length;
         for (uint256 i; i < size; ++i) {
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
```diff
@@ -115,10 +115,7 @@ contract SgxVerifier is EssentialContract, IVerifier {
 
     /// @notice Deletes SGX instances from the registry.
     /// @param _ids The ids array of SGX instances.
-    function deleteInstances(uint256[] calldata _ids)
-        external
-        onlyFromOwnerOrNamed(LibStrings.B_SGX_WATCHDOG)
```
dantaik, line 120: should we also delete the definition of B_SGX_WATCHDOG?

YoGhurt111: OK

davidtaikocha (approved):

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
YoGhurt111: Done

dantaik (approved):

PR #19202, opened 2025-04-03
feat(taiko-client): clear buffer & make `proofPollingInterval` configurable

YoGhurt111:


packages/taiko-client/prover/proof_submitter/proof_submitter_pacaya.go
```diff
@@ -297,6 +303,14 @@ func (s *ProofSubmitterPacaya) BatchSubmitProofs(ctx context.Context, batchProof
 		s.txBuilder.BuildProveBatchesPacaya(batchProof),
 		batchProof,
 	); err != nil {
+		proofBuffer.ClearItems(uint64BatchIDs...)
+		// Resend the proof request
+		for _, proofResp := range batchProof.ProofResponses {
```
davidtaikocha, line 308: Shall we move this check to some place after the `ErrUnretryableSubmission` check?

YoGhurt111: No, in the previous case on Hekla, it is not a `ErrUnretryableSubmission`.

davidtaikocha (approved):

packages/taiko-client/prover/proof_submitter/proof_submitter_pacaya.go
YoGhurt111: And I think its better to regenerate the proof for `ErrUnretryableSubmission`.

davidtaikocha: yes so lets keep it as it is.

mask-pp (approved):

PR #19209, opened 2025-04-04
feat(protocol): support hekla pivot setup

smtmfft:


davidtaikocha (approved):

dantaik:
What's 'pivot'?

smtmfft:
> What's 'pivot'?

The decentralized guardian, which is now a sgx proof originated from taiko-client/geth (unlike raiko, whose proof comes from its own evm). we will try tdx pivot in near future. 
So far, by design, prover needs to offer 2 different proofs, i.e., pivot + another(either sgx/zk/etc)

dantaik (approved):

PR #19220, opened 2025-04-04
fix: script not executing due to incorrect shebang

famouswizard:
The script wasn’t running directly on some systems. The shebang was missing the required `!` character (`#/bin/sh` instead of `#!/bin/sh`).


RogerLamTd:
ref #19221

PR #19222, opened 2025-04-05
fix(taiko-client): allow the api remove preconf blocks

mask-pp:


davidtaikocha (approved):

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -259,7 +259,7 @@ func (s *PreconfBlockAPIServer) RemovePreconfBlocks(c echo.Context) error {
 		"currentHead", currentHead.Number.Uint64(),
 	)
 
-	if reqBody.NewLastBlockID < canonicalHeadL1Origin.BlockID.Uint64() {
+	if reqBody.NewLastBlockID <= canonicalHeadL1Origin.BlockID.Uint64() {
```
davidtaikocha, line 262: I think its OK to let the latest new head to be the highest canonical head?

davidtaikocha (changes requested):

packages/taiko-client/driver/preconf_blocks/api.go
mask-pp: Although this has no effect, it is beyond the scope of preconf.

davidtaikocha: for this API to remove preconf blocks, We SHOULD allow preconfer resets to the latest canonical head, no?

mask-pp: Oh I see this way can clean all the preconf blocks.

PR #19225, opened 2025-04-06
feat(taiko-client): improve `ImportChildBlocksFromCache` and increase cache size

davidtaikocha:
After this fix, `ImportChildBlocksFromCache` can always find the correct most extended children.


packages/taiko-client/driver/preconf_blocks/queue.go
```diff
@@ -60,20 +60,43 @@ func (q *payloadQueue) get(id uint64, hash common.Hash) *eth.ExecutionPayload {
...
+func (q *payloadQueue) getChildren(parentID uint64, parentHash common.Hash) []*eth.ExecutionPayload {
 	q.lock.RLock()
 	defer q.lock.RUnlock()
 
-	for _, item := range q.payloads {
-		if item == nil {
-			return nil // no more items
+	longestChildren := []*eth.ExecutionPayload{}
+
+	var searchLongestChildren func(currentpayload *eth.ExecutionPayload, chain []*eth.ExecutionPayload)
```
mask-pp, line 71:
```suggestion
	var searchLongestChildren func(currentPayload *eth.ExecutionPayload, chain []*eth.ExecutionPayload)
```

packages/taiko-client/driver/preconf_blocks/queue.go
```diff
@@ -60,20 +60,43 @@ func (q *payloadQueue) get(id uint64, hash common.Hash) *eth.ExecutionPayload {
...
+			}
+			if item.id == uint64(currentpayload.BlockNumber)+1 && item.payload.ParentHash == currentpayload.BlockHash {
+				children = append(children, item.payload)
+			}
 		}
-		if item.id == id+1 && item.payload.ParentHash == hash {
-			return item.payload
+		if len(children) == 0 {
+			if len(chain) > len(longestChildren) {
+				longestChildren = append([]*eth.ExecutionPayload{}, chain...)
```
mask-pp, line 84:
```suggestion
				longestChildren = slices.Clone(chain)
```

packages/taiko-client/driver/preconf_blocks/queue.go
```diff
@@ -60,20 +60,43 @@ func (q *payloadQueue) get(id uint64, hash common.Hash) *eth.ExecutionPayload {
...
+		for _, child := range children {
+			searchLongestChildren(child, append(chain, child))
 		}
 	}
-	return nil
+
+	searchLongestChildren(&eth.ExecutionPayload{
+		BlockNumber: eth.Uint64Quantity(parentID),
+		BlockHash:   parentHash,
+	}, []*eth.ExecutionPayload{})
```
mask-pp, line 97:
```suggestion
	}, nil)
```

packages/taiko-client/driver/preconf_blocks/queue.go
```diff
@@ -60,20 +60,43 @@ func (q *payloadQueue) get(id uint64, hash common.Hash) *eth.ExecutionPayload {
...
+				children = append(children, item.payload)
+			}
 		}
-		if item.id == id+1 && item.payload.ParentHash == hash {
-			return item.payload
+		if len(children) == 0 {
+			if len(chain) > len(longestChildren) {
+				longestChildren = append([]*eth.ExecutionPayload{}, chain...)
+			}
+			return
```
mask-pp, line 86:
```suggestion
			} else if len(longestChildren) > 0 &&
				len(chain) == len(longestChildren) &&
				longestChildren[len(longestChildren)-1].Timestamp > chain[len(chain)-1].Timestamp {
				longestChildren = slices.Clone(chain)
			}
			return
```

mask-pp: Same length child chain, use the newest one.

packages/taiko-client/driver/preconf_blocks/queue.go
davidtaikocha: Although they are the same, `[]*eth.ExecutionPayload{}` looks more readable to me, no?

PR #19229, opened 2025-04-07
feat(protocol): add a few governance/treasury related contracts

dantaik:
## Summary

- Added **TaikoTreasuryVault**: A contract to hold treasury tokens and Ether, decoupling asset ownership from the DAO.
- Introduced **MinimalOwner**: A lightweight owner contract that owns all contracts and is itself owned by the DAO. This structure simplifies a potential future transition from the Taiko DAO to a shared L2 DAO (e.g., from the EF).
- Implemented **TokenLocker**: A non-upgradable contract that linearly unlocks TAIKO tokens over time, preventing even the DAO from depleting tokens too quickly.


packages/protocol/contracts/layer1/governance/MinimalOwner.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+/// @title MinimalOwner
+/// @notice
+///   A minimal contract that can:
+///   1) Own other contracts (receive ownership transfers),
+///   2) Transfer ownership (for exmaple, Taiko DAO),
```
dantaik, line 8:
```suggestion
///   2) Transfer ownership (for exmaple, Taiko DAO),
```

xiaodino (approved):

packages/protocol/test/layer1/governance/MinimalOwner.t.sol
```diff
@@ -0,0 +1,58 @@
...
+        vm.startPrank(owner);
+        vm.expectRevert(MinimalOwner.SameAddress.selector);
+        minimalOwner.transferOwnership(owner);
+        vm.stopPrank();
+    }
+
+    function test_minimalOwner_ForwardCall() public {
+        vm.startPrank(owner);
+        (bool success,) = target.call(data);
+        require(success);
```
adaki2004, line 46: Trying to wrap my head around this. How can it be success, if target is not a deployed smart contract, and data is trying to call `bytes data = abi.encodeWithSignature("someFunction()");` on it ?

packages/protocol/contracts/layer1/governance/TokenLocker.sol
```diff
@@ -0,0 +1,72 @@
...
+        require(!initialized, AlreadyInitialized());
+        require(amount != 0, AmountIsZero());
+
+        initialized = true;
+        totalLocked = amount;
+
+        require(token.transferFrom(msg.sender, address(this), amount), TransferFailed());
+    }
+
+    function unlock(address recipient, uint256 amount) external onlyOwner {
```
adaki2004, line 52: 2 things:
1. I think we should use also `reentrancyGuard` here. (Less probability of abuse, because taiko token needs to be upgraded to a malicious code, but technically possible if that is done, that `totalUnlocked += amount;` is drained a bit more than allowed in such case.)

2. Should not we re-use the same paradigm as with `TokenUnlock` contracts we already have.

So for example the following function signature:
- `function unlock(address recipient, uint256 amount) external onlyOwner` //  same as like at `TokenUnlock`, but there we have `onlyRecipient` like [here](https://github.com/taikoxyz/taiko-mono/blob/074aeb756c8cd2bcfa6d888151d228d52ae0cce7/packages/protocol/contracts/layer1/team/TokenUnlock.sol#L143). 
- but also allow this signature:` function unlock() external nonReentrant` so that everyone would be able to call it to help get the tokens forward to the owner (=recipient). Like [here](https://github.com/taikoxyz/taiko-mono/blob/074aeb756c8cd2bcfa6d888151d228d52ae0cce7/packages/protocol/contracts/layer1/team/TokenUnlock.sol#L159) also.

dantaik: 1. Done
2. I think we should make this contract not-upgradable so even the DAO cannot spend token to fast, just in case some KOL manages to get his proposal approved to withdraw most of tokens in one go.

3. Unlike the TokenUnlock contract, recipient is not determined when the contract is deployed, the value should be specified by a DAO proposal.

packages/protocol/test/layer1/governance/MinimalOwner.t.sol
dantaik: Added a few checks for the target address.

dantaik:
@bearni95 @xiaodino please also test theses contracts with the DAO contracts.

adaki2004 (approved):

PR #19240, opened 2025-04-08
fix(taiko-client): don't return err when single sp1 proof is null

YoGhurt111:
Since the single sp1 proof from raiko is null, we need to ignore the case.

### DIFF
--- a/packages/taiko-client/prover/proof_producer/common.go
+++ b/packages/taiko-client/prover/proof_producer/common.go
@@ -45,6 +45,7 @@ func (res *RaikoRequestProofBodyResponseV2) Validate() error {
 	if res.Data.Status == ErrZkAnyNotDrawn.Error() {
 		return ErrZkAnyNotDrawn
 	}
+	// Note: Since the single sp1 proof from raiko is null, we need to ignore the case.
 	if ProofTypeZKSP1 != res.ProofType &&
 		(res.Data.Proof == nil || len(res.Data.Proof.Proof) == 0) {
 		return errEmptyProof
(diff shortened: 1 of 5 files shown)


mask-pp (approved):

packages/taiko-client/prover/proof_producer/common.go
```diff
@@ -45,7 +45,8 @@ func (res *RaikoRequestProofBodyResponseV2) Validate() error {
 	if res.Data.Status == ErrZkAnyNotDrawn.Error() {
 		return ErrZkAnyNotDrawn
 	}
-	if res.Data.Proof == nil || len(res.Data.Proof.Proof) == 0 {
+	if ProofTypeZKSP1 != res.ProofType &&
```
davidtaikocha, line 48: Let's add a comment explaining why should check SP1 here, to let others who read this code understand.

packages/taiko-client/prover/proof_producer/compose_proof_producer.go
```diff
@@ -93,8 +93,10 @@ func (s *ComposeProofProducer) RequestProof(
 			); err != nil {
 				return err
 			} else {
-				proof = common.Hex2Bytes(resp.Data.Proof.Proof[2:])
 				proofType = resp.ProofType
+				if ProofTypeZKSP1 != proofType {
```
davidtaikocha, line 97: ditto

packages/taiko-client/prover/proof_producer/compose_proof_producer.go
```diff
@@ -236,6 +238,13 @@ func (s *ComposeProofProducer) requestBatchProof(
 	}
 
 	if err := output.Validate(); err != nil {
+		log.Debug(
+			"Validation result",
```
davidtaikocha, line 242:
```suggestion
			"Proof output validation result",
```

packages/taiko-client/prover/proof_producer/compose_proof_producer.go
YoGhurt111: Done.

davidtaikocha (approved):

PR #19241, opened 2025-04-09
fix(taiko-client): fix bug about set head l1Origin

mask-pp:


cyberhorsey:
this is fixed in p2p-payloadQueue-2 already, and iterated blocks in correct order

mask-pp:
> > this is fixed in p2p-payloadQueue-2 already, and iterated blocks in correct order. current code iterates last block first, then every other block from start forward
> 
> It can't fix the bug, similar to the following picture: image

Oh sorry, the commit works.

PR #19242, opened 2025-04-09
feat(protocol): rename pivot verifier to geth verifier

smtmfft:


RogerLamTd (approved):

packages/protocol/contracts/layer1/devnet/verifiers/DevnetVerifier.sol
```diff
@@ -11,7 +11,7 @@ contract DevnetVerifier is ComposeVerifier {
 
     constructor(
         address _taikoInbox,
-        address _pivotVerifier,
+        address _gethVerifier,
```
dantaik, line 14: shall it be sgxGethverifier?

dantaik (approved):

PR #19247, opened 2025-04-10
feat(protocol): introduce prover market for Proposer-Prover Separation

dantaik:
This PR introduces a minimal implementation of a simple prover market **with the least possible changes to TaikoInbox**. The goal is to establish a basic foundation that we can iterate on. Once it’s in place, we can identify what’s not working and improve it progressively through future iterations.


dantaik:
### Q: Every time a proposer submits a batch, the prover is deducted a liveness bond. Won’t this make the capital requirement for provers substantially large(and slightly unpredictable)?

Yes, you’re right — this is a known trade-off in my current implementation. Because the Alethia protocol supports parallel (out-of-order) proving, liveness bonds are locked for all unverified batches. So if there are n such batches, the prover must have liveness bond posted for each of them, which can lead to higher and somewhat unpredictable capital requirements.

Additionally, the liveness bond is only returned when the batch is verified using the proof submitted by the assigned prover — not just when a proof is verified. Also in the current protocol, there is also a cooldownWindow, which further extends the lock-up period of the bond.

That said, Dani and I discussed this out-of-order proving design again yesterday. One nice property it offers (and I really like) is that the proof transaction for the n-th batch (e.g., tx2) doesn’t depend on the previous one for batch n-1 (tx1). So, if tx1 gets re-orged out of the chain, it won’t cause tx2 to revert, which increases robustness against re-orgs especially when tx1 and tx2 are from different provers.



### Q: In the TODO you suggest that in the case of a vacant prover checking that the bid is not too high(proposers are not being overcharged). Using a moving average will mean you will have to store the last X bids, and add some complexity. Perhaps a simpler/less sophisticated approach is just ensure that the bid is < proposer liveness bond(or below X% of the proposer liveness bond if the liveness bond is too high)?

This part of the code is still quite early and there’s a lot of room for improvement.

I don’t think we should require proposers to post a liveness bond (as shown in the PR). Instead, when a proposer is about to propose a batch, they should be able to query the current proving fee and feel confident that it won’t be front-run and increased by another prover before their proposal transaction is confirmed. I’m not sure the current implementation guarantees this property. Predictability of the proving fee is essential — proposers must be sure that the fee can go down, but not be increased unexpectedly by front-running bids.

One idea I think is workable: the proposer pays the current winning prover (i.e., the assigned prover) the proving fee at the time of proposing the batch. Then:

- If the assigned prover proves the batch within the proving window, their liveness bond is fully returned.
- If not, 50% of the bond is slashed: half goes to the actual prover (who may or may not be the assigned one), and the other half is burned.

With this model, for each unverified batch, only the assigned prover’s liveness bond needs to be locked. Once the batch is verified, we process the bond accordingly. The proving fee is always paid upfront by the proposer.

In this design, provers determine the proving fee, not proposers. Proposers can choose not to propose if the fee is too high. But in the context of preconfirmation, this still poses a challenge — fee predictability over longer periods may be important, especially if preconfers want to opt-out safely. Unfortunately, I don’t have a perfect solution for this.

The way I’ve been reasoning about it is this: if proving fees are consistently too high and blocks become unprofitable, preconfers will eventually opt out. If all provers collude to maintain high fees, some preconfers may start acting as provers themselves, which introduces competition and should bring the fees down over time. So maybe we don’t need long-term fee predictability (e.g., over 1–2 Ethereum epochs), but I’m still open to being wrong about this.


### Q: bid function: Should we check that _exitTimestamp is in the future(and more specifically enough time in the future so that it makes sense?)* similar to how it is done in requestExit

Yeah, the code isn’t ready for serious review yet. I opened the PR to convince myself, Dani, and Ruby that we don’t need to significantly modify the TaikoInbox contract to support proposer-prover separation.


### Q: I’m not sure I understand the purpose of the outbidThreshold in the ProverMarket contract. Can you ellaborate on what it does?

The goal is to ensure that Prover Bob doesn’t have to wait for the current prover Alice’s available bond balance to fall below the liveness bond before he can outbid her — especially when he’s offering the same fee or a slightly higher one.

A naive implementation would treat Alice as disqualified only when her available bond balance drops below the liveness bond. But in that case, proposing a new batch would revert, and Bob would need to initiate a new bidding process to become the prover. That process could take 12 seconds or more, creating a gap where no proposer can submit new batches — a problematic vacancy.

To avoid this, we could consider Alice “replaceable” when her available bond balance falls below a higher threshold — for example, 100 * livenessBond. If Bob submits a valid bid when Alice’s bond is under that threshold, he becomes the new assigned prover immediately, avoiding any downtime. If no one bids, Alice can still continue proving up to 100 more batches.

Of course, Bob can always outbid Alice with a lower fee. But under this design, we can also allow him to match Alice’s fee or slightly exceed it (e.g., 1.05× her fee), as long as her bond is no longer sufficient to sustain the full proving load.

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -0,0 +1,125 @@
...
+
+    ITaikoInbox public immutable inbox;
+    /// @dev If a prover’s available bond balance is below this threshold, they are not eligible
+    /// to participate in the bidding process.
+    uint256 public immutable biddingThreshold;
+    /// @dev If the current prover’s bond balance falls below this value, they can be outbid by
+    /// another prover even if the new bid offers the same proving fee or only a slightly higher one
+    /// (e.g., 1.01× or 1.05× the current fee).
+    uint256 public immutable outbidThreshold;
+    /// @dev TIf the current prover’s bond balance drops below this threshold, they are considered
```
dantaik, line 35:
```suggestion
    /// @dev If the current prover’s bond balance drops below this threshold, they are considered
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -0,0 +1,125 @@
...
+    /// disqualified (evicted), and the active prover is reset to address(0).
+    uint256 public immutable provingThreshold;
+    uint256 public immutable minExitDelay;
+    address internal prover;
+    uint64 internal fee; // proving fee per batch
+    mapping(address account => uint256 exitTimestamp) internal exitTimestamps;
+
```
dantaik, line 42: add a __gap here

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -0,0 +1,125 @@
...
+        inbox = ITaikoInbox(_inbox);
+
+        biddingThreshold = _biddingThreshold;
+        outbidThreshold = _outbidThreshold;
+        provingThreshold = _provingThreshold;
+
+        minExitDelay = _minExitDelay;
+    }
+
+    function bid(uint64 _fee, uint256 _exitTimestamp) external validExitTimestamp(_exitTimestamp) {
```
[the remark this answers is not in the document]
dantaik: yes

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -10,6 +10,7 @@ import "src/shared/libs/LibNetwork.sol";
 import "src/shared/libs/LibStrings.sol";
 import "src/shared/signal/ISignalService.sol";
 import "src/layer1/verifiers/IVerifier.sol";
+import { IProverMarket } from "src/layer1/prover-market/ProverMarket.sol";
```
dantaik, line 13:
```suggestion
import  "src/layer1/prover-market/IProverMarket.sol";
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -10,6 +10,7 @@ import "src/shared/libs/LibNetwork.sol";
 import "src/shared/libs/LibStrings.sol";
 import "src/shared/signal/ISignalService.sol";
 import "src/layer1/verifiers/IVerifier.sol";
+import  "src/layer1/prover-market/IProverMarket.sol";
```
dantaik, line 13:
```suggestion
import "src/layer1/prover-market/IProverMarket.sol";
```

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -250,8 +253,9 @@ interface ITaikoInbox {
 
     /// @notice Emitted when a batch is verified.
     /// @param batchId The ID of the verified batch.
+    /// @param prover The address of the prover.
     /// @param blockHash The hash of the verified batch.
-    event BatchesVerified(uint64 batchId, bytes32 blockHash);
+    event BatchesVerified(uint64 batchId, address indexed prover, bytes32 blockHash);
```
dantaik, line 258:
```suggestion
    event BatchesVerified(uint64 batchId, address prover, bytes32 blockHash);
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -683,7 +708,7 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
 
                 batch = state.batches[_stats2.lastVerifiedBatchId % _config.batchRingBufferSize];
                 batch.verifiedTransitionId = tid;
-                emit BatchesVerified(_stats2.lastVerifiedBatchId, blockHash);
+                emit BatchesVerified(batchId, blockHash);
```
dantaik, line 711:
```suggestion
                emit BatchesVerified(_stats2.lastVerifiedBatchId, blockHash);
```

PR #19250, opened 2025-04-10
feat(protocol): allow perconfers/proposers to mark blocks

dantaik:
When multiple blocks are preonfirmed by one preconfer, they can be marked differently using the new marker parameter.


davidtaikocha (approved):

adaki2004 (approved):

dantaik:
@smtmfft please take a look.

smtmfft (approved):
only for proposer's own use?? or it's a public protocol value that is read/used by the others.

dantaik:
> only for proposer's own use?? or it's a public protocol value that is read/used by the others.

This allows a preconfer to mark its preconf-ed blocks before the blocks are proposed, then when slashing is enabled, these values will be used to slash the preconfer if 1) the marker value is invalid, or 2) the block preconf-ed is different from the blocks proposed.

PR #19253, opened 2025-04-11
feat(protocol): introduce ShastaForkRouter with ITaikoInbox's functions named

dantaik:
This PR adds `ShastaForkRouter` to support the upgrade from _Pacaya_  to _Shasta_.

We route function calls between forks (e.g., _Pacaya_ and _Shasta_) using function selectors, which are derived from the function name and input types. Since selectors must be unique across forks, we’ve renamed all functions in _Shasta_’s `ITaikoInbox` interface to start with a capital letter (e.g., Foo() instead of foo()).

Going forward, we will alternate the naming convention for interface functions between forks to avoid selector collisions:

- One fork uses foo()
- The next uses Foo()
- Then foo() again, and so on…

This approach ensures each fork’s interface has a unique set of selectors while keeping changes manageable.


adaki2004:
> We route function calls between forks (e.g., _Pacaya_ and _Shasta_) using function selectors, which are derived from the function name and input types. Since selectors must be unique across forks, we’ve renamed all functions in _Shasta_’s `ITaikoInbox` interface to start with a capital letter (e.g., Foo() instead of foo()).
> 
> Going forward, we will alternate the naming convention for interface functions between forks to avoid selector collisions:
> 
>     * One fork uses foo()
> 
>     * The next uses Foo()
> 
>     * Then foo() again, and so on…
> 
> 
> This approach ensures each fork’s interface has a unique set of selectors while keeping changes manageable.

Just .02 about it. I think lower level to capital (then back with the new fork) can possibly create confusions. For example like: maintenance issues, tooling issues (e.g.: when dealing with autogenerated bindings, and capital and lower letter does not distinguished), breaks coding guideline and it is a hard to "see or keep in mind" rule (because e.g.: node devs cannot search for a prefix/suffix _v1, _v2, _v3, or smth).

I suppose keeping suffixes, preffixes. Maybe agree on the fact that we dont use the same starter letter for each for, so `_o`, or `o_` is always `Ontake` and so `o` is unique. As well `p` (Pacaya), `s` (Shasta).

adaki2004:
> > > We route function calls between forks (e.g., _Pacaya_ and _Shasta_) using function selectors, which are derived from the function name and input types. Since selectors must be unique across forks, we’ve renamed all functions in _Shasta_’s `ITaikoInbox` interface to start with a capital letter (e.g., Foo() instead of foo()).
> > > Going forward, we will alternate the naming convention for interface functions between forks to avoid selector collisions:
> > > ```
> > > * One fork uses foo()
> > > 
> > > * The next uses Foo()
> > > 
> > > * Then foo() again, and so on…
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This approach ensures each fork’s interface has a unique set of selectors while keeping changes manageable.
> > 
> > 
> > Just .02 about it. I think lower level to capital (then back with the new fork) can possibly create confusions. For example like: maintenance issues, tooling issues (e.g.: when dealing with autogenerated bindings, and capital and lower letter does not distinguished), breaks coding guideline and it is a hard to "see or keep in mind" rule (because e.g.: node devs cannot search for a prefix/suffix _v1, _v2, _v3, or smth).
> > I suggest keeping suffixes, preffixes. Maybe agree on the fact that we dont use the same starter letter for each for, so `_o`, or `o_` is always `Ontake` and so `o` is unique. As well `p` (Pacaya), `s` (Shasta) ?
> 
> I think if we don't like `GetFoo`, we may want to switch between `_getFoo` and `getFoo`, we should not use `o` or `p` or something as the nameing is even more weird: `ogetFoo`, `pgetFoo`...

I did not mean `ogetFoo` or `pgetFoo`, but rather `getFoo_o`, or `getFoo_p` rather.

My argument against this can be seen by a simple golang abigen-binding generator which used to work with `UpperCamelCase()` notation. So by generating some off-chain code,fork is not visible first hand.

See solidity code in green, and corresponding binding from golang abigen.
kép

dantaik:
> > > > We route function calls between forks (e.g., _Pacaya_ and _Shasta_) using function selectors, which are derived from the function name and input types. Since selectors must be unique across forks, we’ve renamed all functions in _Shasta_’s `ITaikoInbox` interface to start with a capital letter (e.g., Foo() instead of foo()).
> > > > Going forward, we will alternate the naming convention for interface functions between forks to avoid selector collisions:
> > > > ```
> > > > * One fork uses foo()
> > > > 
> > > > * The next uses Foo()
> > > > 
> > > > * Then foo() again, and so on…
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > This approach ensures each fork’s interface has a unique set of selectors while keeping changes manageable.
> > > 
> > > 
> > > Just .02 about it. I think lower level to capital (then back with the new fork) can possibly create confusions. For example like: maintenance issues, tooling issues (e.g.: when dealing with autogenerated bindings, and capital and lower letter does not distinguished), breaks coding guideline and it is a hard to "see or keep in mind" rule (because e.g.: node devs cannot search for a prefix/suffix _v1, _v2, _v3, or smth).
> > > I suggest keeping suffixes, preffixes. Maybe agree on the fact that we dont use the same starter letter for each for, so `_o`, or `o_` is always `Ontake` and so `o` is unique. As well `p` (Pacaya), `s` (Shasta) ?
> > 
> > 
> > I think if we don't like `GetFoo`, we may want to switch between `_getFoo` and `getFoo`, we should not use `o` or `p` or something as the nameing is even more weird: `ogetFoo`, `pgetFoo`...
> 
> I did not mean `ogetFoo` or `pgetFoo`, but rather `getFoo_o`, or `getFoo_p` rather.
> 
> My argument against this can be seen by a simple golang abigen-binding generator which used to work with `UpperCamelCase()` notation. So by generating some off-chain code,fork is not visible first hand.
> 
> See solidity code in green, and corresponding binding from golang abigen. kép

The golong point makes sense. Lets use v1, v2...

dantaik:
@adaki2004 could you work on this. We can use the following conventions here: https://github.com/taikoxyz/taiko-mono/pull/19267

packages/protocol/test/shared/tokenvault/ERC20Vault.h.sol
```diff
@@ -16,11 +16,11 @@ contract PrankTaikoInbox {
         batch = _batch;
     }
 
-    function getBatch(uint64) external view returns (ITaikoInbox.Batch memory) {
+    function v4GetBatch(uint64) external view returns (ITaikoInbox.Batch memory) {
         return batch;
     }
 
-    function isOnL1() external pure returns (bool) {
+    function v4v4IsOnL1() external pure returns (bool) {
```
adaki2004, line 23:
```suggestion
    function v4IsOnL1() external pure returns (bool) {
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -364,23 +364,23 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
     /// @notice Verify batches by providing the length of the batches to verify.
     /// @dev This function is necessary to upgrade from this fork to the next one.
     /// @param _length Specifis how many batches to verify. The max number of batches to verify is
-    /// `pacayaConfig().maxBatchesToVerify * _length`.
-    function verifyBatches(uint64 _length)
+    /// `GetConfig().maxBatchesToVerify * _length`.
```
dantaik, line 367:
```suggestion
    /// `v4GetConfig().maxBatchesToVerify * _length`.
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -32,7 +32,7 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
 
     address public immutable inboxWrapper;
     address public immutable verifier;
-    address public immutable bondToken;
+    address private immutable bondToken;
```
dantaik, line 35:
```suggestion
    address internal immutable bondToken;
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -48,8 +48,8 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
             require(msg.sender == fallbackPreconfer, NotFallbackPreconfer());
         }
 
-        // Both TaikoInbox and TaikoWrapper implement the same ABI for proposeBatch.
-        (info_, meta_) = proposeBatchEntrypoint.proposeBatch(_params, _txList);
+        // Both TaikoInbox and TaikoWrapper implement the same ABI for ProposeBatch.
```
dantaik, line 51:
```suggestion
        // Both TaikoInbox and TaikoWrapper implement the same ABI for IProposeBatch.
```

packages/nfts/contracts/trailblazers-season-2/BadgeRecruitment.sol
```diff
@@ -259,7 +259,7 @@ contract BadgeRecruitment is
 
     /// @notice Get the current configuration
     /// @return The current configuration
-    function getConfig() external view returns (Config memory) {
+    function v4GetConfig() external view returns (Config memory) {
```
dantaik, line 262:
```suggestion
    function getConfig() external view returns (Config memory) {
```

davidtaikocha (approved):

packages/protocol/deployments/hekla-contract-logs.md
```diff
@@ -438,6 +438,7 @@
   - upgraded on Oct 29, 2024 at commit `3d12cb2`
 
 ### shared_resolver
+
```
dantaik, line 441:
```suggestion
```

PR #19254, opened 2025-04-11
feat(protocol): outbid logic if Alice running out of deposit

adaki2004:
I think this check:
`if (currentProver != address(0) && currentProverBalance < outbidThreshold)`
is more apropriate than this:
`if (currentProver == address(0) || currentProverBalance < outbidThreshold)`

because if `currentProver` is `address(0)`, we should by default allow bids.


packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -80,9 +82,12 @@ contract ProverMarket is EssentialContract, IProverMarket {
         (address currentProver, uint64 currentFee, uint256 currentProverBalance) =
             _getCurrentProver();
 
-        if (currentProver == address(0) || currentProverBalance < outbidThreshold) {
+        if (currentProver != address(0) && currentProverBalance < outbidThreshold) {
```
dantaik, line 85: I actually believe the old if-statement is correct -- it handles the `currentProver == address(0)` case correctly.

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -80,9 +82,12 @@ contract ProverMarket is EssentialContract, IProverMarket {
...
             // TODO(dani): ensure the new _fee cannot be too large right...
-            // Using a moving average???
+            // Using a moving average??? -> dani: Seems "on-chain" heavy (gas-inefficient). We
+            // already know Alice (current prover) is soon running out of her deposits, so we only
+            // allow in this case outbidding with a higher amount.
+            require(uint256(_fee) < uint256(currentFee) * 105 / 100, InvalidBid());
```
dantaik, line 90:
```suggestion
            require(uint256(_fee) <= uint256(currentFee) * 105 / 100, InvalidBid());
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
adaki2004: But if `currentProver == 0` it is handled (by default, in old code too) by the end of the section in this function, it will allow the bid to go thru and set the corresponding varoables, no ? I might misunderstanding it.

dantaik: Ah, I see, but still, we need to handle (currentProver==0) differently, as currentFee should not be used, it only represent the fee required by the last prover, which can be zero. We need to have something like:

```
else if (currentProver==0) {
  // ...
} else {
```

adaki2004: Oh - in slack convo with Gustavo, i see what you mean.

adaki2004: > Ah, I see, but still, we need to handle (currentProver==0) differently, as currentFee should not be used, it only represent the fee required by the last prover, which can be zero. We need to have something like:
> 
> ```
> else if (currentProver==0) {
>   // ...
> } else {
> ```

Ok. Added a very light, moving average calculation here.

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -122,4 +132,35 @@ contract ProverMarket is EssentialContract, IProverMarket {
...
+            uint256 currentFee = (feeHistory & shiftedMask) >> (i * 64);
+            sum += currentFee;
+        }
+        
+        // add 10% margin to the average
+        return (sum / feeCount) * 110 / 100;
+    }
+
+    function _updateFeeHistory(uint64 _newFee) private {
+        // shift existing fees to the left (discard oldest)
```
dantaik, line 155: why not  `feeHistory = (feeHistory << 64) | _newFee`

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -122,4 +132,35 @@ contract ProverMarket is EssentialContract, IProverMarket {
...
+        }
+        
+        uint256 sum = 0;
+        uint256 mask = uint256(type(uint64).max);
+        
+        for (uint8 i = 0; i < feeCount; i++) {
+            uint256 shiftedMask = mask << (i * 64);
+            uint256 currentFee = (feeHistory & shiftedMask) >> (i * 64);
+            sum += currentFee;
+        }
```
dantaik, line 148:
```solidity
for (uint i; i < 4; ++i) {
  uint fee = (feeHistory >> (64*i)) & type(uint64).max;
  if (fee ==0)  break;
  total += fee;
}
avg = i == 0 ? 0 : total / i;
```

dantaik: no need to keep track of feeCount, feeHistory is *rolling*

dantaik (changes requested):

dantaik:
So why calculate the avg of the last 4 bids instead of maintaining a moving average?

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
adaki2004: Indeed, no need storage level `feeCount`, but locally needed tho. (In case we dont have 4, we dont divide with 4).

adaki2004:
> So why calculate the avg of the last 4 bids instead of maintaining a moving average?

Isn't it a moving average ? If new (accepted) bid is incoming, it is recorded. I agree that this 4 data sample might seem low count if we need very precise data. But in case we just need a rough, "good enough something" to base the max threshold, this can be a gas efficient way of doing so, in my opinion.

PR #19260, opened 2025-04-13
feat(protocol): tract avgProverMarketFee and cap bids with the average

dantaik:


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -205,7 +210,17 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
+
+                // Update avg prover market fee
+                require(stats1.genesisHeight == 0, "loaded already"); // TODO(daniel): remove this
+                stats1 = state.stats1; // load from storage
+                stats1.avgProverMarketFee = uint64(
+                    (
+                        (proverMarketFeeChangeFactor - 1) * stats1.avgProverMarketFee
+                            + uint64(proverFee / (1 gwei))
+                    ) / proverMarketFeeChangeFactor
+                );
```
adaki2004, line 223: I think this is a bit "on-chain heavy". I mean, this code is done when there is a proposal. While the fee only changes when there is an incoming bid. (at least with prover market).
I propose a very similar approach, but inside `ProverMarket`, and just the cycle of 4 fees can give a good, rough starting point to calculate the max threshold.

See this PR: https://github.com/taikoxyz/taiko-mono/pull/19254

dantaik: If the moving average is updated by bids in the prover market, then the bidder can manipulate the average. In this PR, the moving average is only updated when the fee is actually used by a preconfer, avoiding such manipulation.

dantaik: Maybe these two average can work together.

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -84,9 +87,12 @@ contract ProverMarket is EssentialContract, IProverMarket {
         external
         validExitTimestamp(_exitTimestamp)
     {
-        require(_fee % (1 gwei) == 0, FeeNotDivisibleByFeeUnit());
-        require(_fee / (1 gwei) <= type(uint64).max, FeeTooLarge());
-        uint64 fee_ = uint64(_fee / (1 gwei));
+        require(_fee % feeUnit == 0, FeeNotDivisibleByFeeUnit());
+        require(_fee / feeUnit <= type(uint64).max, FeeTooLarge());
```
ggonzalez94, line 91: Wouldn't it be simpler if the bid is already expressed in feeUnit(gwei) as a uint64. I think having different representations(what people use to bid vs how fees are represented internally) leads to more complexity and might even lead to more confusion

ggonzalez94: Alternatively if you represent everything with a `uint96` should be enough to represent all Taiko supply, but I think you already had a reserved `uint64` in the struct that you are reusing, so likely not possible

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -122,10 +128,16 @@ contract ProverMarket is EssentialContract, IProverMarket {
             _getCurrentProver();
         return currentProverBalance < provingThreshold
             ? (address(0), 0)
-            : (currentProver, 1 gwei * currentFee);
+            : (currentProver, feeUnit * currentFee);
+    }
+
+    /// @dev The maximum fee that can be used by provers to bid.
+    /// The current implementation returns a 5 times of the average fee.
```
ggonzalez94, line 135: nit: 
```suggestion
    /// The current implementation returns a 5 times of the moving average.
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
ggonzalez94: I agree that if we allow the average to naively change in the `bid` there's an option for the prover to manipulate that. You put the max bid possible, then submit a bunch of other bids undercutting you by the minimum possible in the same block effectively pushing the moving average higher, and then repeat the process with a higher number. Admittedly, this is a bit elaborated and would cost significant L1 gas fees.
I think if we want to eliminate the calculation in the `proposeBatch` function, one option we have is to only update the moving average once per block(or every X blocks). This way the average still changes slowly and gives enough time for other provers(or Taiko) to come in and bid a reasonable amount. The downside is that we would have to store an extra `uint64`, but significantly reduces the cost in the inbox(which is what should be called more often) since we don't have to do an `SSTORE` anymore.

ggonzalez94 (commented):
I like this approach. Left a few suggestions, the most important one being where the moving average fee is calculated and stored

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: See https://github.com/taikoxyz/taiko-mono/pull/19264

dantaik:
Now I prefer https://github.com/taikoxyz/taiko-mono/pull/19264 over this one.

PR #19262, opened 2025-04-13
fix(taiko-client): fix an occasional `engine.SYNCING` error when receiving P2P preconf blocks

davidtaikocha:
Also added some improvements for Ontake blocks `basefee` calculation in this PR, which is not necessary but a nice to have.

### DIFF
--- a/packages/taiko-client/driver/preconf_blocks/server.go
+++ b/packages/taiko-client/driver/preconf_blocks/server.go
@@ -163,7 +163,10 @@ func (s *PreconfBlockAPIServer) OnUnsafeL2Payload(
 		"peer", from,
 		"blockID", uint64(msg.ExecutionPayload.BlockNumber),
 		"hash", msg.ExecutionPayload.BlockHash.Hex(),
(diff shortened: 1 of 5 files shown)


packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -206,7 +210,7 @@ func (s *PreconfBlockAPIServer) OnUnsafeL2Payload(
 	if err != nil && !errors.Is(err, ethereum.NotFound) {
 		return fmt.Errorf("failed to fetch parent header by hash: %w", err)
 	}
-	if parentInFork == nil && parentInCanonical == nil {
+	if parentInFork == nil && (parentInCanonical == nil || parentInCanonical.Hash() != msg.ExecutionPayload.ParentHash) {
```
davidtaikocha, line 213: Here is the fix, we need to make sure the block hashes are matched.

YoGhurt111 (approved):

cyberhorsey (approved):

PR #19264, opened 2025-04-14
feat(protocol): enable ProverMarket to keep track of a fee moving average

dantaik:
Alternative to https://github.com/taikoxyz/taiko-mono/pull/19260


packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -125,6 +136,25 @@ contract ProverMarket is EssentialContract, IProverMarket {
...
+            uint64 _avgFee = avgFee;
+
+            avgFee = _avgFee == 0
+                ? fee
+                : uint64((_avgFee * (FEE_CHANGE_FACTOR - 1) + fee) / FEE_CHANGE_FACTOR);
+        }
+    }
+
+    function getMaxFee() public view returns (uint256) {
+        uint256 _max = avgFee * MAX_FEE_MULTIPLIER;
```
dantaik, line 154:
```suggestion
        uint256 _max = MAX_FEE_MULTIPLIER * avgFee;
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -125,6 +136,25 @@ contract ProverMarket is EssentialContract, IProverMarket {
...
     }
 
+    function onProverAssigned() external onlyFrom(address(inbox)) {
+        if (assignmentCount > FEE_CHANGE_THRESHOLD) {
+            return;
+        }
+
+        if (++assignmentCount == FEE_CHANGE_THRESHOLD) {
+            uint64 _avgFee = avgFee;
+
```
dantaik, line 146:
```suggestion
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -125,6 +136,24 @@ contract ProverMarket is EssentialContract, IProverMarket {
...
+    function onProverAssigned() external onlyFrom(address(inbox)) {
+        if (assignmentCount > FEE_CHANGE_THRESHOLD) {
+            return;
+        }
+
+        if (++assignmentCount == FEE_CHANGE_THRESHOLD) {
+            uint64 _avgFee = avgFee;
+            avgFee = _avgFee == 0
+                ? fee
+                : uint64((_avgFee * (FEE_CHANGE_FACTOR - 1) + fee) / FEE_CHANGE_FACTOR);
```
dantaik, line 148:
```suggestion
                : uint64(((FEE_CHANGE_FACTOR - 1) * _avgFee + fee) / FEE_CHANGE_FACTOR);
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -40,9 +46,13 @@ contract ProverMarket is EssentialContract, IProverMarket {
     uint64 internal fee; // proving fee per batch
 
     /// @dev Slot 2
+    uint64 internal avgFee; // moving average of fees
```
dantaik, line 49:
```suggestion
    uint64 public avgFee; // moving average of fees
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -85,7 +95,7 @@ contract ProverMarket is EssentialContract, IProverMarket {
         validExitTimestamp(_exitTimestamp)
     {
         require(_fee % (1 gwei) == 0, FeeNotDivisibleByFeeUnit());
-        require(_fee / (1 gwei) <= type(uint64).max, FeeTooLarge());
+        require(_fee <= getMaxFee(), FeeTooLarge());
```
adaki2004, line 98: This check should not be around line 107 ? In this if clause `if (currentProver == address(0) || currentProverBalance < outbidThreshold)` ?

dantaik: Fixed, please take another look.

adaki2004 (changes requested):

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -77,42 +92,43 @@ contract ProverMarket is EssentialContract, IProverMarket {
...
     }
 
-    function bid(
-        uint256 _fee,
-        uint256 _exitTimestamp
-    )
-        external
-        validExitTimestamp(_exitTimestamp)
-    {
+    function bid(uint256 _fee, uint64 _exitTimestamp) external validExitTimestamp(_exitTimestamp) {
```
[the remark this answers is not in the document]
dantaik: This bond is a fixed amount per batch, independent of the bid itsel,  even if the prover bids 0, the potential loss of the bond should still be enough to incentivize timely proofs.

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -176,15 +181,15 @@ contract ProverMarket is EssentialContract, IProverMarket {
         return _max.min(type(uint64).max) * 1 gwei;
     }
 
-    function _getCurrentProver() public view returns (address, uint64, uint256) {
-        address currentProver = prover;
+    function _getCurrentProver() public view returns (address, uint64) {
```
adaki2004, line 184: 2 things.

1. Should not it be `private` if we use the `_` prefix ?

2. I think it technically works but worth to give some commenting (or renaming) here a bit , because it gives a bit of a confusion or inconsistency.

So there are 2 functions with same name, but it works differently:
- it returns fee as uint64, directly from fee, without multiplying by 1 gwei ( i know it is for off-chain use)
- It does not check `bondBalanceOf(prover) `against provingThreshold. 
- it might return a prover and a fee that is already exited based only prover balance. I know it is checked in line 110 : `else if (inbox.bondBalanceOf(prover_) < outbidThreshold)`  but still, the 2 functions not consistent.
- `getCurrentProver() `applies an extra eviction filter (provingThreshold)

I think i know what you mean here, but at least naming not ideal.

How about this:
- wrapping _getCurrentProver (below) into getCurrentProver(), and also please see the comment i put into the code.

```
function _getCurrentProver() internal view returns (address, uint64) {
    address _prover = prover;
    if (
        _prover == address(0) ||
        block.timestamp >= provers[_prover].exitTimestamp ||
        inbox.bondBalanceOf(_prover) < provingThreshold // this is the case where in bid, it would be equal to else if (inbox.bondBalanceOf(prover_) < outbidThreshold) so maybe use a 3rd return value, to signal this special else if case.
    ) {
        return (address(0), 0);
    } else {
        return (_prover, fee);
    }
}
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -77,63 +93,103 @@ contract ProverMarket is EssentialContract, IProverMarket {
...
+    }
+
+    function getMaxFee() public view returns (uint256) {
+        uint256 _max = MAX_FEE_MULTIPLIER * avgFee;
+        return _max.min(type(uint64).max) * 1 gwei;
     }
 
-    function _getCurrentProver() public view returns (address, uint64, uint256) {
-        address currentProver = prover;
+    function _getCurrentProver() public view returns (address, uint64) {
```
dantaik, line 184:
```suggestion
    function _getCurrentProver() internal view returns (address, uint64) {
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
dantaik: Please see the latest commit, but I'm not sure I follow the _outbidThreshold_ part.

`inbox.bondBalanceOf(_prover) < provingThreshold` cannot be part of `_getCurrentProver` as we still need a non-zero current prover then allow out bidding it if

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -77,63 +88,100 @@ contract ProverMarket is EssentialContract, IProverMarket {
...
+            // The current prover has less than outbidThreshold, so the new prover can set any
+            // fee as long as it's not larger than the current fee
+            require(_newFeeInGwei <= currentFeeInGwei, FeeLargerThanCurrent());
         } else {
-            require(fee_ < currentFee * 9 / 10, InvalidBid());
+            // The current prover has more than outbidThreshold, so the new prover can set any
+            // fee as long as it's not larger than 90% of the current fee
+            require(
+                _newFeeInGwei <= currentFeeInGwei * NEW_BID_PERCENTAGE / 100, FeeLargerTooLarge()
+            );
```
adaki2004, line 115: Logically looks good (not only this but the whole file), how about this small change here:

```
if (currentProver != address(0)) {
    uint256 maxFee;

    if (inbox.bondBalanceOf(currentProver) < outbidThreshold) {
        maxFee = currentFeeInGwei;
    } else {
        // Apply 90% rule
        maxFee = currentFeeInGwei * NEW_BID_PERCENTAGE / 100;
    }

    require(_newFeeInGwei <= maxFee, FeeLargerThanCurrent());
}
```

I think it is a little bit more readable, and removes duplicate `require` and keeps the logic centered around the `maxFee` per specific scenario.

Anyways, it does not affect testing. I'll go ahead and try do some testing.

dantaik: feel free to apply the changes in your PR/commits.

adaki2004 (approved):

PR #19271, opened 2025-04-15
feat(taiko-client): preconf handover skip slots

cyberhorsey:
the driver should be aware of the boundary. thus, for a `preconf_handover_skip_slots = 4`:

the current sequencer for epoch N sequences slots 28-31 of epoch N-1, and slots 0-27 of epoch N.
the next sequencer can start in slot 28.

### DIFF
--- a/packages/taiko-client/driver/config.go
+++ b/packages/taiko-client/driver/config.go
@@ -115,6 +115,15 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
 		return nil, err
 	}
 
+	preconfHandoverSkipSlots := c.Uint64(flags.PreconfHandoverSkipSlots.Name)
+	if rpc.L1Beacon != nil && preconfHandoverSkipSlots > rpc.L1Beacon.SlotsPerEpoch {
+		return nil, fmt.Errorf(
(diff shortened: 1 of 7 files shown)


davidtaikocha (approved):

packages/taiko-client/driver/config.go
```diff
@@ -125,5 +126,6 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
 		PreconfBlockServerCORSOrigins: c.String(flags.PreconfBlockServerCORSOrigins.Name),
 		P2PConfigs:                    p2pConfigs,
 		P2PSignerConfigs:              signerConfigs,
+		PreconfHandoverSkipSlots:      c.Uint64(flags.PreconfHandoverSkipSlots.Name),
```
YoGhurt111, line 129: Let‘s check the value `c.Uint64(flags.PreconfHandoverSkipSlots.Name)` not larger than `s.rpc.L1Beacon.SlotsPerEpoch` before we use it.

davidtaikocha: Checks added

YoGhurt111 (approved):

PR #19278, opened 2025-04-16
fix(protocol): fix liveness bond bug

dantaik:


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -233,8 +243,6 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
             batch.reserved4 = 0;
             // SSTORE }}
 
-            _debitBond(params.proposer, config.livenessBondBase);
```
dantaik, line 236: This a bug fix. It indicates we don't have test to cover the case when a prover market is used by the inbox.

dantaik: @

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -211,12 +210,23 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
+                    // proposer pay the prover fee.
+                    _debitBond(info_.proposer, proverFee);
 
-                proverMarket.onProverAssigned(meta_.prover, proverFee, meta_.batchId);
+                    if (proverFee < config.livenessBondBase) {
+                        _debitBond(meta_.prover, config.livenessBondBase - proverFee);
+                    } else {
+                        _creditBond(meta_.prover, proverFee - config.livenessBondBase);
+                    }
+                    proverMarket.onProverAssigned(meta_.prover, proverFee, meta_.batchId);
```
dantaik, line 225: shall we also call onProverAssigned even if the prover and the proposer are the same address?

adaki2004: i dont think so. In that case there is no 'actual' payment and so it does not reflect and effect on the avgFee (+ we save gas).

packages/protocol/deployments/test_inbox_measure_gas_used.txt
```diff
@@ -1,5 +1,5 @@
 See `test_inbox_measure_gas_used` in InboxTest_ProposeAndProve.t.sol
 
-Gas per proposing: 177599
-Gas per proving + verification: 154638
-Total: 332237
\ No newline at end of file
+Gas per proposing: 176962
+Gas per proving + verification: 154329
+Total: 331291
```
adaki2004, line 5: Obviously this is not including going thru the real `TaikoInbox`, will add tests in a separate PR today.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: Maybe we should also add tests by using the real TaikoInbox and but a MockProverMarket to test that TaikoInbox debits/credits bonds from/to proposer/provers correctly.

packages/protocol/deployments/test_inbox_measure_gas_used.txt
dantaik: We are using `ConfigurableInbox`, which is a sub-contract of `TaikoInbox` implementation.

dantaik: No real prover though.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
adaki2004: Already started doing it.

PR #19281, opened 2025-04-16
feat(protocol): Enable prover market via taiko inbox + overflow fix

adaki2004:
Added tests which hook in a real `ProverMarket` into `Inbox` tests, which helped the gas comparison/benchmarks and the same time test `creditBond` / `debitBond` logic too.

The gas comparisons can be found in the following file accordingly:
```
test_inbox_measure_gas_used.txt --> Original
test_inbox_with_provermarket_diff_prover_and_proposer_fee_above_liveness_measure_gas_used.txt
test_inbox_with_provermarket_diff_prover_and_proposer_measure_gas_used.txt
test_inbox_with_provermarket_same_prover_as_proposer_measure_gas_used.txt
```

The issue found is this:
`(currentProver, 1 gwei * uint256(currentFeeInGwei));`

So without this `1 gwei`: `uint32`, `currentFeeInGwei` is `uint64`. The upconvert will happen till `uint64` only, but with that it can easily lead to arithmetic overflow - which i ran into during the tests obviously, so this is a fix for it.


packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -86,6 +86,12 @@ contract ProverMarket is EssentialContract, IProverMarket {
         minExitDelay = _minExitDelay;
     }
 
+    /// @notice Initializes the ProverMarket contract
+    /// @param _addressManager The address of the address manager contract
+    function initialize(address _addressManager) external initializer {
+        __Essential_init(_addressManager);
```
dantaik, line 92: `__Essential_init` accepts the owner address as its parmeter.

We should not use `_addressManager`, now we use `_resolver`, and it needs to be initialized in the constructor.

dantaik: See Bridge.sol as an example.

packages/protocol/test/layer1/based/InboxTestBase.sol
```diff
@@ -10,6 +10,7 @@ abstract contract InboxTestBase is Layer1Test {
 
     ITaikoInbox internal inbox;
     TaikoToken internal bondToken;
+    ProverMarket proverMarket;
```
dantaik, line 13:
```suggestion
    ProverMarket internal proverMarket;
```

dantaik (changes requested):

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -86,6 +86,12 @@ contract ProverMarket is EssentialContract, IProverMarket {
         minExitDelay = _minExitDelay;
     }
 
+    /// @notice Initializes the ProverMarket contract
+    /// @param _addressManager The address of the address manager contract
+    function initialize(address _addressManager) external initializer {
```
dantaik, line 91:
```suggestion
    function init(address _owner) external initializer {
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
dantaik, line 92:
```suggestion
        __Essential_init(_owner);
```

packages/protocol/contracts/layer1/prover-market/ProverMarket.sol
```diff
@@ -138,7 +144,7 @@ contract ProverMarket is EssentialContract, IProverMarket {
         (address currentProver, uint64 currentFeeInGwei) = _getCurrentProverAndFeeInGwei();
         return currentProver == address(0)
             ? (address(0), 0)
-            : (currentProver, 1 gwei * currentFeeInGwei);
+            : (currentProver, 1 gwei * uint256(currentFeeInGwei));
```
dantaik, line 147: there is another occurance of 1 gwei * currentFeeInGwei. How about we define a constant:

uint256 GWEI = uint(1 gwei) and use it in the code.

dantaik (approved):

PR #19318, opened 2025-04-22
feat(taiko-client): add witness mode support to L2 EE interactions

johntaiko:


packages/taiko-client/pkg/rpc/engine.go
```diff
@@ -32,12 +34,24 @@ func NewJWTEngineClient(url, jwtSecret string) (*EngineClient, error) {
...
+// ForkchoiceUpdated updates the forkchoice on the execution client.
+func (c *EngineClient) ForkchoiceUpdated(
+	ctx context.Context,
+	fc *engine.ForkchoiceStateV1,
+	attributes *engine.PayloadAttributes,
+) (*engine.ForkChoiceResponse, error) {
+	if c.witness {
+		return c.forkchoiceUpdatedWithWitness(ctx, fc, attributes)
+	}
+	return c.ForkchoiceUpdated(ctx, fc, attributes)
```
YoGhurt111, line 51:
```suggestion
	return c.forkchoiceUpdated(ctx, fc, attributes)
```

packages/taiko-client/cmd/flags/common.go
```diff
@@ -68,6 +68,13 @@ var (
 		Category: commonCategory,
 		EnvVars:  []string{"JWT_SECRET"},
 	}
+	Witness = &cli.BoolFlag{
```
YoGhurt111, line 71: Shall we move this flag to `driverCategory`?

packages/taiko-client/driver/chain_syncer/blob/blocks_inserter/pacaya.go
```diff
@@ -36,6 +38,8 @@ type BlocksInserterPacaya struct {
 	calldataFetcher    txlistFetcher.TxListFetcher
 	blobFetcher        txlistFetcher.TxListFetcher
 	mutex              sync.Mutex
+
+	witnessCache *lru.Cache[uint64, *hexutil.Bytes] // Cache for witnesses
```
YoGhurt111, line 42: Looks like the `witnessCache` is not used in client?

johntaiko: Can be used after raiko supporting

packages/taiko-client/cmd/flags/common.go
johntaiko: This flag is only work with l2endpoint and jwt
```go
l2AuthClient, err = NewJWTEngineClient(cfg.L2EngineEndpoint, cfg.JwtSecret, cfg.Witness)
```

johntaiko: Needs more opinions, @davidtaikocha @smtmfft

davidtaikocha:
Closed since it's outdated.

PR #19325, opened 2025-04-23
feat(protocol): remove average fee tracking from ProverMarket

dantaik:
From Gustavo:

> I believe we can get rid of the maxFee altogether and simplify the inbox and prover market. Let me try to make the case to why:
> - The goal of setting a cap on the fee was to avoid provers setting unreasonable high fees under certain circumstancesThe only moment a prover can set an arbitrary fee is when there is designated prover(a.k.a they have exited). Since there is a minExitDelay (given it is chosen to be long enough) this should allow competition to happen and the price to converge to its real value.Even if this does not happen, and provers collude or only a few step in and the fee is set too high - we now have the mechanism for proposers to self prove. This in my opinion is enough of a safe guard for the proposer(he can decide to run a backstop prover or have a service he can pay to in the case it is needed). It also serves as an incentive that deter provers for charging too much(there’s an opt out mechanism)
>
> What are the advantages with this:
> - Simpler code. Easier to read, reason about and auditLess room for bugsMore gas efficient(no need to update the moving average on block proposal, even if it was only done on the first FEE_CHANGE_FACTOR times)More “free market”. > - Setting a cap allows the market of provers to converge to a more real value, and avoids situations we haven’t really considered where the cap prevents true price discovery of happeningNo risk of getting into a situation where the EMA is too low, and even 2x the max fee is not desirable for a prover to bid. Notice this is a situation where the protocol would be “stuck” until prices go down or a prover comes in and bids even if at a loss.
>
>
> Risks:
> Provers collude more often than I’m giving credit for and all proposers end up having to run backstop prover or hire off-chain services since the proving fees are too unpredictable.


adaki2004 (approved):
lgtm

ggonzalez94 (commented):
Looks cleaner and I can see the gas cost for proposing is much closer to that before the prover market

PR #19329, opened 2025-04-24
fix(protocol): fix _getOperatorForEpoch in PreconfWhitelist.sol

dantaik:
Related to https://github.com/taikoxyz/taiko-mono/pull/19327.

@cyberhorsey I think your fix won't work -- you make the timestamp (for which the beacon root is queried) bigger, which makes the return value less predictable as the random seed. I think we should make the value smaller so we query a older beacon root as the random seed.

Please test both your PR and mine to verify.


cyberhorsey:
https://github.com/taikoxyz/taiko-mono/pull/19328

I deployed this one, and its working so far.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -209,17 +218,18 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
...
     /// @dev The cost of this function is primarily linear with respect to operatorCount.
     function _getOperatorForEpoch(uint64 _epochTimestamp) internal view returns (address) {
-        if (_epochTimestamp < LibPreconfConstants.SECONDS_IN_EPOCH) {
-            return address(0);
-        }
+        uint256 timeShift = selectorEpochOffset * LibPreconfConstants.SECONDS_IN_EPOCH;
+        // We use the beacon root at or after ` _epochTimestamp - timeShift` as the random number to
+        // select an operator.
+        // selectorEpochOffset must be big enough to ensure a non-zero beacon root is
+        // available and immutable.
```
linoscope, line 225: Why does having big enough `selectorEpochOffset` ensure non-zero beacon root? In other words, isn't something like following enough?

```suggestion
        // selectorEpochOffset must be big enough to ensure that the beacon root is finalized and
        // cannot change due to reorg.
```

PR #19330, opened 2025-04-24
feat(protocol): add preconf slasher

AnshuJalan:
This PR covers `Preconfirmation Structure`, `Preconfirmation Violations`, `Missing EOP` and `Invalid EOP` sections of our LLD doc


packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,162 @@
...
+        uint256 heightOfFirstBlockInBatch =
+            parsedEvidence.batchInfo.lastBlockId - parsedEvidence.batchInfo.blocks.length;
+        if (parsedEvidence.preconfedBlockHeader.number == heightOfFirstBlockInBatch) {
+            // If the preconfirmed block is the first block in the batch, we compare the parent hash
+            // against the verified block hash of the previous batch, since the "batch blockhash" is
+            // basically the hash of the last block.
+            ITaikoInbox.TransitionState memory parentTransition =
+                taikoInbox.v4GetBatchVerifyingTransition(uint64(parsedPayload.batchId - 1));
+            require(
+                parentTransition.blockHash != parsedEvidence.preconfedBlockHeader.parentHash,
```
adaki2004, line 103: hmm, should not be `==` instead of `!=` ?

AnshuJalan: Right

AnshuJalan:
I feel I will also be including Preconf EOP violations in this same contract. I was inclining toward having a separate contract for EOPs, but now that I think about it, the preconfer would then need to opt into another contract just for EOP. This is inconvenient.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,257 @@
...
+    uint256 public slashAmountWei;
+    address public immutable urc;
+
+    constructor(address _resolver) EssentialContract(_resolver) { }
+
+    function init(address _owner, address _urc, uint256 _slashAmountWei) external initializer {
+        __Essential_init(_owner);
+        slashAmountWei = _slashAmountWei;
+        urc = _urc;
+    }
```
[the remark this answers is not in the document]
adaki2004: So if `urc` is immutable (and `slashAmountWei` not) then something like this:
```suggestion
    constructor(address _urc) EssentialContract(address(0)) { 
        urc = _urc;
    }

    function init(address _owner, uint256 _slashAmountWei) external initializer {
        __Essential_init(_owner);
        slashAmountWei = _slashAmountWei;
    }
```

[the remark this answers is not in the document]
AnshuJalan: Ah yes, already noticed this issue and fixed on my machine. Will be pushing along with other stuff in a min

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,275 @@
...
+        require(parsedPayload.eop == true, NotEndOfPreconfirmation());
+
+        ITaikoInbox.Batch memory batch = taikoInbox.v4GetBatch(uint64(parsedPayload.batchId));
+
+        // Slash if another block was proposed after EOP in the same batch
+        if (parsedEvidence.preconfedBlockHeader.number != batch.lastBlockId) {
+            uint256 slashAmountWei = getSlashAmountWei().invalidEOP;
+            emit SlashedInvalidEOP(committer, parsedPayload, slashAmountWei);
+            return slashAmountWei;
+        }
```
adaki2004, line 222: Just trying to wrap my head around (the base checks) - i'm following the doc's assertions and the "transition verified" is checked when `_slashPreconfirmationViolation()` is called, but here, not.

This one i'm talking about:
       ```
 ITaikoInbox.Batch memory batch = taikoInbox.v4GetBatch(uint64(parsedPayload.batchId));
        ITaikoInbox.TransitionState memory transition =
            taikoInbox.v4GetBatchVerifyingTransition(uint64(parsedPayload.batchId));

        // Validate that the batch has been verified
        require(transition.blockHash != bytes32(0), BatchNotVerified());
```

I think we should check it here too (and maybe when calling `_slashMissingEOP()` ), no 🤔  ? (If so, maybe an internal function which reutrns the batch (if verified) can make sense and be reused.

Comment (AnshuJalan) [reply packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol:222]: For EOP violation checks, we do not need to wait for the block proof since our main goal is to check if more blocks or batches were proposed after the EOP was issued. The transition is required when we intend to compare the content of the blocks

Comment (adaki2004) [reply packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol:222]: Thanks Anshu, makes sense!

File: packages/protocol/contracts/layer1/preconf/libs/LibPreconfConstants.sol [hunk 4]
```diff
@@ -16,6 +16,8 @@ library LibPreconfConstants {
     uint256 internal constant TWO_EPOCHS = 2 * SECONDS_IN_EPOCH;
     uint256 internal constant DISPUTE_PERIOD = 2 * SECONDS_IN_EPOCH;
 
+    bytes32 internal constant PRECONF_DOMAIN_SEPARATOR = keccak256("TAIKO_PRECONF");
```
dantaik, line 19:
```suggestion
    bytes32 internal constant PRECONF_DOMAIN_SEPARATOR = keccak256("TAIKO_ALETHIA_PRECONF");
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
...
+        uint256 anchorId;
+        // Blockhash of the L1 anchor block
+        bytes32 anchorHash;
+        // ID of the batch that will be containing this block in TaikoInbox
+        uint256 batchId;
+        // Hash of the header of the preconfirmed block
+        bytes32 blockHash;
+        // `true` if this preconfer is not going to deliver anymore
+        // preconfirmations after this block
+        bool eop; // End-Of-Preconf flag
```
dantaik, line 30: I know we briefly discussed this in our call, but I still prefer using an enum here. It allows us to explicitly mark a block as one of the following:

- 1: Start of a batch
- 2: End of a batch, but not end of preconfirmation
- 3: End of a batch and end of preconfirmation
- 0: All other blocks

By enforcing the correct marker for each block, it becomes easier to identify mistakes. Also, if we decide not to use markers 1 and 2, we don’t need to change the protocol at all. Using an enum gives us this flexibility.

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
...
+    // Byte-encoded and used as `ISlasher.Commitment.payload`.
+    struct CommitmentPayload {
+        // Taiko specific DS
+        bytes32 domainSeparator;
+        // Chainid of Taiko
+        uint256 chainId;
+        // Timestamp of the L1 slot in which the preconfer is the proposer.
+        // In the case of falback preconfer, this will be the timestamp of the last slot of the
+        // epoch.
+        uint256 l1ProposalSlotTimestamp;
```
dantaik, line 19: if slot 3,4,5 are empty and slot 6 is Alice, then `l1ProposalSlotTimestamp` is the timestamp of slot 6?

If so, it seems the timestamp is more associated with preconfer Alice, not her proposals. I suggest to rename it to `preconferSlotTimestamp`

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    // Ethereum block header
+    struct BlockHeader {
+        bytes32 parentHash;
+        bytes32 ommersHash;
+        address coinbase;
+        bytes32 stateRoot;
+        bytes32 transactionsRoot;
+        bytes32 receiptRoot;
```
dantaik, line 76:
```suggestion
        bytes32 receiptsRoot;
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
...
+        bytes bloom; // 256 bytes
```
dantaik, line 77:
```suggestion
        bytes logsBloom; // 256 bytes
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
...
+        ViolationType violationType;
+        // This is the BatchInfo of the batch that contains the block at height X
+        ITaikoInbox.BatchInfo batchInfo;
+        // This is the BatchMetadata of the batch that contains the block at height X
+        ITaikoInbox.BatchMetadata batchMetadata;
+        // This is the BatchMetadata of the next batch that contains the block at height X + 1
+        // This is only used for EOP violations
+        ITaikoInbox.BatchMetadata nextBatchMetadata;
+        // Header of the preconfirmed block at height X
+        BlockHeader preconfedBlockHeader;
```
dantaik, line 45: But this should be TaikoBlockHeader, BlockHeader has a comment "// Ethereum block header"

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
...
+    struct BlockhashProofs {
+        // Blockhash value
+        bytes32 value;
+        // Patricia trie account proof
+        bytes[] accountProof;
+        // Patricia trie storage proof
+        bytes[] storageProof;
+    }
+
+    // Ethereum block header
```
dantaik, line 69:
```suggestion
    // Taiko block header
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    // Evidence to prove preconfirmation violation of a block at a certain height X
+    struct Evidence {
+        // Type of preconfirmation violation
+        ViolationType violationType;
+        // This is the BatchInfo of the batch that contains the block at height X
+        ITaikoInbox.BatchInfo batchInfo;
+        // This is the BatchMetadata of the batch that contains the block at height X
+        ITaikoInbox.BatchMetadata batchMetadata;
+        // This is the BatchMetadata of the next batch that contains the block at height X + 1
```
dantaik, line 41: what if block X and block X+1 are in the same batch?

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
...
+        uint256 batchId;
+        // Hash of the header of the preconfirmed block
+        bytes32 blockHash;
+        // `true` if this preconfer is not going to deliver anymore
+        // preconfirmations after this block
+        bool eop; // End-Of-Preconf flag
+    }
+
+    // Evidence to prove preconfirmation violation of a block at a certain height X
+    struct Evidence {
```
dantaik, line 34: May I suggest we define a more generic proof struct like this:

```
struct Evidence {
ViolationType violationType;
bytes evidenceData;
}
```
Then depends on the violationType, we decode evidenceData into different sub evidence struct.

dantaik: Or maybe we don't need such a generic strut at all, in the slash function, we can just take the first byte and use it to check the violationType, then we do something like:

```
if (_evidence[:1] == uint(ViolationType.xyz)) {
   EvidenceXyz memory evidence = abi.decode(_evidence[1:], (EvidenceXyz));
   
   // route to a library to minimize the code of this contract so this contract?
   LibEvidenceXyz.slash(evidence) 
 } else ...
 
```

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,277 @@
...
+        returns (uint256)
+    {
+        require(msg.sender == urc, SenderIsNotUrc());
+
+        // Parse the commitment payload and evidence
+        CommitmentPayload memory parsedPayload =
+            abi.decode(_commitment.payload, (CommitmentPayload));
+        Evidence memory parsedEvidence = abi.decode(_evidence, (Evidence));
+
+        ITaikoInbox taikoInbox =
```
dantaik, line 44: taikoInbox and l2ChainId can be immutables.

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import { ITaikoInbox } from "src/layer1/based/ITaikoInbox.sol";
```
dantaik, line 4:
```suggestion
import "src/layer1/based/ITaikoInbox.sol";
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -0,0 +1,144 @@
...
+import { ISlasher } from "@eth-fabric/urc/ISlasher.sol";
```
dantaik, line 5:
```suggestion
import "@eth-fabric/urc/ISlasher.sol";
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
dantaik: We didn't enforce this rule, so better to be consistent with the current styling.

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
AnshuJalan: The naming here is referenced from the spec: https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/blocks.py#L38

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
AnshuJalan: Naming referenced from spec: https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/blocks.py#L38

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
AnshuJalan: `nextBatchMetadata` stays empty in that case. I think the suggestion on keeping evidences separate will make this clearer.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,300 @@
...
+        // Validate the pre-images in the evidence
+        require(
+            keccak256(abi.encode(_parsedEvidence.batchMetadata)) == batch.metaHash,
+            InvalidBatchMetadata()
+        );
+        require(
+            keccak256(abi.encode(_parsedEvidence.batchInfo))
+                == _parsedEvidence.batchMetadata.infoHash,
+            InvalidBatchInfo()
+        );
```
adaki2004, line 118: Please confirm, i think the second check 
```
require(
            keccak256(abi.encode(_parsedEvidence.batchInfo))
                == _parsedEvidence.batchMetadata.infoHash,
            InvalidBatchInfo()
        );
```

is redundant, no ?

Because if we hash the `metadata` of `_parsedEvidence` , and compare it with the posted `batch`'s `metaHash` - which already contains the `infoHash`, there is no need to check the `infoHash` from that same `_parsedEvidence` again, no ?

AnshuJalan: The challenger passes the evidence separately, and we still need to confirm if the evidence struct passed as an argument is correct or not.

adaki2004: Right, true!

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
AnshuJalan: https://github.com/taikoxyz/taiko-mono/pull/19365/files

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,300 @@
...
+
+    function _validateBlockHeader(
+        BlockHeader memory _header,
+        bytes32 _expectedHash
+    )
+        internal
+        pure
+    {
+        require(
+            keccak256(LibRLP.encode(abi.encode(_header))) == _expectedHash, InvalidBlockHeader()
```
dantaik, line 297: ChatGPT: The RLP encoding should be done directly on the block header fields, not on the ABI-encoded data

dantaik: Also ChatGPT about how to use the lib correctly:

```

struct MyHeader {
    bytes32 parentHash;
    bytes32 stateRoot;
    uint256 number;
}

function encodeHeader(MyHeader memory header) public pure returns (bytes memory) {
    // Create a dynamic array to hold the components
    LibRLP.List memory list = LibRLP.newList(3); // 3 fields in the struct
    
    // Add each field to the list in order
    LibRLP.addBytes32(list, header.parentHash);
    LibRLP.addBytes32(list, header.stateRoot);
    LibRLP.addUint(list, header.number);
    
    // Encode the list
    return LibRLP.encode(list);
}

```

AnshuJalan: These functions do not exist in the library, GPT probably lacks context of the latest library. I am using their `encode(bytes)` function. However, I think may be that's not completely right to use. I still have to create individual list items

AnshuJalan: Update in latest commit

dantaik:
Screenshot 2025-05-03 at 12 42 24

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,319 @@
...
+        taikoInbox = ITaikoInbox(_taikoInbox);
+        l2ChainId = taikoInbox.v4GetConfig().chainId;
+    }
+
+    function init(address _owner) external initializer {
+        __Essential_init(_owner);
+    }
+
+    /// @inheritdoc ISlasher
+    function slash(
```
dantaik, line 38: may I suggest to move most of the code into a view function, for example, validateSlash(),  so that it can be used to verify the inputs are all valid for slashing.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,273 @@
...
+        EvidenceMissingEOP memory _evidence
+    )
+        internal
+        returns (uint256)
+    {
+        // Validate that the commitment is not an EOP
+        require(_payload.eop == false, EOPIsPresent());
+
+        ITaikoInbox.Batch memory nextBatch = taikoInbox.v4GetBatch(uint64(_payload.batchId + 1));
+        require(
```
dantaik, line 256: I don't think the logic here is correct.

AnshuJalan: Right, pushing a fix. I missed checking in current batch

dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,268 @@
...
+
+    function _validateInvalidEOP(
+        CommitmentPayload memory _payload,
+        bytes calldata _evidenceData
+    )
+        internal
+        view
+        returns (uint256)
+    {
+        EvidenceInvalidEOP memory evidence = abi.decode(_evidenceData[1:], (EvidenceInvalidEOP));
```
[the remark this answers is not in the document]
dantaik: It's a bug :D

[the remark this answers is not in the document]
dantaik: There is a bug. I'll create a small PR to fix it.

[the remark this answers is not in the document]
dantaik: https://github.com/taikoxyz/taiko-mono/pull/19521

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,268 @@
...
+        uint256 blockId = evidence.preconfedBlockHeader.number;
+        if (blockId == batch.lastBlockId) {
+            _verifyBatchData(_payload.batchId + 1, evidence.nextBatchMetadata);
+
+            require(
+                evidence.nextBatchMetadata.proposedAt <= _payload.preconferSlotTimestamp,
+                NextBatchProposedInNextPreconfWindow()
+            );
+        } else {
+            // Check if the block is not the last one in the batch
```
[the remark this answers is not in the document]
dantaik: _verifyBatchData only verifies the integraty of batch metadata and info, not that the batch contains the given block.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -0,0 +1,268 @@
...
```
dantaik, line 204:
```suggestion
```

dantaik (approved):

YoGhurt111 (approved):

PR #19335, opened 2025-04-25
feat(protocol): add lookahead store

AnshuJalan:
This PR covers the `Lookahead Posting` section of the LLD doc


packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,258 @@
...
+                        || slashingCommitment.optedOutAt >= epochTimestamp
+                ),
+            OperatorHasNotOptedIntoPreconfSlasher()
+        );
+
+        return slashingCommitment.committer;
+    }
+
+    function _setLookaheadRoot(uint256 epochTimestamp, bytes32 lookaheadRoot) internal {
+        lookahead[epochTimestamp % lookaheadStoreConfig().lookaheadBufferSize] = lookaheadRoot;
```
adaki2004, line 252: So, currently we allow Bob to post. Then Alice, Carol, etc. to post the same lookahead over and over again, overwriting each.
This looks fine, but depending on the monetization/incentivization/slashing strategy, should not we guard against re-writing the lookahead ? Maybe writing it once, and only allow guardian to overwrite ?
Smth like this:
`require(_getLookaheadRoot(nextEpochTimestamp) == bytes32(0), AlreadyPosted());`

AnshuJalan: Actually yes, I missed this check. Lookahead can only be posted once and further overwritten by the guardian if required using a special function

AnshuJalan: Once the lookahead is posted, it cannot be reposted via the `updateLookahead` function since this check fails:
```solidity
 // Only proceed if lookahead is required
        require(isLookaheadRequired(), LookaheadNotRequired());
```

The guardian may update the lookahead using `overwriteLookahead`

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
```diff
@@ -0,0 +1,112 @@
...
+    struct Config {
+        // The size of the lookahead buffer.
+        uint16 lookaheadBufferSize;
+        // The minimum collateral for a registered operator to post the lookahead.
+        uint80 minCollateralForPosting;
+        // The minimum collateral for a registered operator to preconf.
+        uint80 minCollateralForPreconfing;
+    }
+
+    error LookaheadNotRequired();
```
dantaik, line 55: sort errors

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+    error EmptyLeaves();
+    error IndexOutOfBounds();
+    error LeavesTooLarge();
+    /**
+     * @dev Generates a complete Merkle tree from an array of leaves
+     * @dev The tree size is limited to 256 leaves
+     * @param leaves Array of leaf values
+     * @return bytes32 Root hash of the Merkle tree
+     */
+
```
dantaik, line 23:
```suggestion
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+import "@openzeppelin/contracts/utils/math/Math.sol";
+
+/**
+ * @title MerkleTree
+ * @dev Implementation of a binary Merkle tree with proof generation and verification
+ */
+library LibMerkleTree {
+    error EmptyLeaves();
+    error IndexOutOfBounds();
+    error LeavesTooLarge();
```
dantaik, line 16:
```suggestion
    error LeavesTooLarge();

```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+        uint256 index,
+        bytes32[] calldata proof
+    )
+        internal
+        pure
+        returns (bool)
+    {
+        bytes32 computedHash = leaf;
+
+        for (uint256 i = 0; i < proof.length; i++) {
```
dantaik, line 160:
```suggestion
        for (uint256 i; i < proof.length; ++i) {
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+        uint256 index,
+        bytes32[] memory proof
+    )
+        internal
+        pure
+        returns (bool)
+    {
+        bytes32 computedHash = leaf;
+
+        for (uint256 i = 0; i < proof.length; i++) {
```
dantaik, line 128:
```suggestion
        for (uint256 i; i < proof.length; ++i) {
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+        while (size > 1) {
+            height++;
+            size /= 2;
+        }
+
+        bytes32[] memory nodes = new bytes32[](_nextPowerOfTwo);
+        bytes32[] memory proof = new bytes32[](height); // <-- This is the key fix
+
+        // Fill leaf nodes
+        for (uint256 i = 0; i < leaves.length; i++) {
```
dantaik, line 80:
```suggestion
        for (uint256 i; i < leaves.length; ++i) {
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+    function generateTree(bytes32[] memory leaves) internal pure returns (bytes32) {
+        if (leaves.length == 0) revert EmptyLeaves();
+        if (leaves.length == 1) return leaves[0];
+        if (leaves.length > 256) revert LeavesTooLarge();
+
+        uint256 _nextPowerOfTwo = nextPowerOfTwo(leaves.length);
+        bytes32[] memory nodes = new bytes32[](_nextPowerOfTwo);
+
+        // Fill leaf nodes
+        for (uint256 i = 0; i < leaves.length; i++) {
```
dantaik, line 33:
```suggestion
        for (uint256 i; i < leaves.length; ++i) {
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+            nodes[i] = leaves[i];
+        }
+
+        // Build up the tree
+        uint256 n = _nextPowerOfTwo;
+        while (n > 1) {
+            for (uint256 i = 0; i < n / 2; i++) {
```
dantaik, line 40:
```suggestion
            for (uint256 i; i < n / 2; ++i) {
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+// Implementation based on https://github.com/eth-fabric/urc/blob/main/src/lib/MerkleTree.sol
```
dantaik, line 4: if there are tests, maybe we should also clone the tests here.

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+
+        bytes32[] memory nodes = new bytes32[](_nextPowerOfTwo);
+        bytes32[] memory proof = new bytes32[](height); // <-- This is the key fix
+
+        // Fill leaf nodes
+        for (uint256 i = 0; i < leaves.length; i++) {
+            nodes[i] = leaves[i];
+        }
+        // Fill remaining nodes with zero
+        for (uint256 i = leaves.length; i < _nextPowerOfTwo; i++) {
```
dantaik, line 84:
```suggestion
        for (uint256 i = leaves.length; i < _nextPowerOfTwo; ++i) {
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+        uint256 levelSize = _nextPowerOfTwo;
+        uint256 currentIndex = index;
+
+        // Build proof level by level
+        while (levelSize > 1) {
+            uint256 siblingIndex = currentIndex ^ 1; // Get sibling index
+            proof[proofIndex++] = nodes[siblingIndex];
+
+            // Calculate next level
+            for (uint256 i = 0; i < levelSize / 2; i++) {
```
dantaik, line 98:
```suggestion
            for (uint256 i; i < levelSize / 2; ++i) {
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+    {
+        bytes32 computedHash = leaf;
+
+        for (uint256 i = 0; i < proof.length; i++) {
+            if (index % 2 == 0) {
+                computedHash = _efficientKeccak256(computedHash, proof[i]);
+            } else {
+                computedHash = _efficientKeccak256(proof[i], computedHash);
+            }
+            index = index / 2;
```
dantaik, line 166:
```suggestion
            index = index >> 1;
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
```
dantaik, line 134:
```suggestion
            index = index >> 1;
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+            nodes[i] = leaves[i];
+        }
+
+        // Build up the tree
+        uint256 n = _nextPowerOfTwo;
+        while (n > 1) {
+            for (uint256 i = 0; i < n / 2; i++) {
+                nodes[i] = _efficientKeccak256(nodes[2 * i], nodes[2 * i + 1]);
+            }
+            n = n / 2;
```
dantaik, line 43:
```suggestion
            n = n >> 1;
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
```diff
@@ -0,0 +1,194 @@
...
+        if (leaves.length <= 1) return new bytes32[](0);
+
+        uint256 _nextPowerOfTwo = nextPowerOfTwo(leaves.length);
+
+        // Calculate height of tree (log2 of next power of 2)
+        uint256 height = 0;
+        uint256 size = _nextPowerOfTwo;
+        while (size > 1) {
+            height++;
+            size /= 2;
```
dantaik, line 73:
```suggestion
            size >>= 1;
```

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
dantaik: I think we should use >> and << to replace "/2" and "*2"

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
```diff
@@ -0,0 +1,112 @@
...
+     * @param epochTimestamp The timestamp of the epoch.
+     * @return The lookahead root.
+     */
+    function getLookaheadRoot(uint256 epochTimestamp) external view returns (bytes32);
+
+    /**
+     * @notice Returns the configuration of the lookahead store.
+     * @return The configuration of the lookahead store.
+     */
+    function lookaheadStoreConfig() external pure returns (Config memory);
```
dantaik, line 111:
```suggestion
    function getConfig() external pure returns (Config memory);
```

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,266 @@
...
+
+/// @title LookaheadStore
+/// @custom:security-contact security@taiko.xyz
+contract LookaheadStore is ILookaheadStore, EssentialContract {
+    IRegistry public immutable urc;
+    address public immutable guardian;
+    address public immutable preconfSlasher;
+
+    // The timestamp of the last recorded lookahead slot.
+    uint256 public lastRecordedSlotTimestamp;
```
dantaik, line 19:
```suggestion
    uint64 public lastRecordedSlotTimestamp;
```

lets use smaller type so in the future, we can reuse the slot.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,266 @@
...
+            // Merkleize the lookahead buffer entries and update the lookahead root for next epoch
+            bytes32 root = LibMerkleTree.generateTree(leaves);
+            _setLookaheadRoot(nextEpochTimestamp, root);
+
+            emit LookaheadRootUpdated(nextEpochTimestamp, root);
+        }
+    }
+
+    /// @inheritdoc ILookaheadStore
+    function overwriteLookahead(uint256 _epochTimestamp, bytes32 _lookaheadRoot) external {
```
dantaik, line 118:
```suggestion
    function overwriteLookahead(uint256 _epochTimestamp, bytes32 _lookaheadRoot) external onlyFrom(guardian) {
```

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,266 @@
...
+            for (uint256 i; i < lookaheadPayloads.length; ++i) {
+                LookaheadPayload memory lookaheadPayload = lookaheadPayloads[i];
+
+                _validateSlotTimestamp(
+                    lookaheadPayload,
+                    i > 0 ? lookaheadPayloads[i - 1].slotTimestamp : 0,
+                    nextEpochTimestamp
+                );
+
+                address committer =
```
dantaik, line 82: so we can have more than one comitter for a single `updateLookahead` call?

dantaik:
@AnshuJalan I’m trying to understand the implementation.

If slots 2, 3, and 4 are empty and slot 5 has a validator, when building the tree, do you use slot 5’s validator to fill the leaves for slots 2, 3, and 4 as well?

I’m wondering if, when the lookahead is sparse, it might be simpler to represent it like this:

`[(slot1, Alice), (slot5, Bob)]`

Then, when proposing a block in slot 3, the proposer would include this lookahead in calldata, we verify its hash matches the persisted hash, and iterate through the in-memory lookahead to verify that the sender is Bob.

Lets talk about it in our next meeting.

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
AnshuJalan: Afaik, the compiler automatically resolves divisions and multiplications by 2 to bitwise operations in the EVM.

packages/protocol/contracts/layer1/preconf/libs/LibMerkleTree.sol
AnshuJalan: Tests are present: https://github.com/eth-fabric/urc/blob/main/test/MerkleTree.t.sol
Will blend it with the Taiko helpers

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,266 @@
...
+    function overwriteLookahead(uint256 _epochTimestamp, bytes32 _lookaheadRoot) external {
+        require(msg.sender == guardian, SenderIsNotGuardian());
+        _setLookaheadRoot(_epochTimestamp, _lookaheadRoot);
+    }
+
+    // View functions --------------------------------------------------------------------------
+
+    /// @inheritdoc ILookaheadStore
+    function isLookaheadRequired() public view returns (bool) {
+        uint256 epochTimestamp = LibPreconfUtils.getEpochTimestamp();
```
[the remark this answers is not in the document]
AnshuJalan: Waiting on your PR to be merged to rebase

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
```diff
@@ -0,0 +1,112 @@
...
+        ISlasher.SignedCommitment memory signedCommitment
+    )
+        external;
+
+    /**
+     * @notice Called by the guardian to overwrite the lookahead root for an epoch.
+     * @param epochTimestamp The timestamp of the epoch.
+     * @param lookaheadRoot The lookahead root.
+     */
+    function overwriteLookahead(uint256 epochTimestamp, bytes32 lookaheadRoot) external;
```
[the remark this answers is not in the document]
AnshuJalan: The expectation is that we trust the guardian to release the correct lookahead publicly. It's something we can discuss on Monday. It's a trade-off between making it cheaper for the guardians to overwrite the lookahead vs making the lookahead handling in the node a bit more complex.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: Individual preconfers in the lookahead have their own committer addresses

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,266 @@
...
+            require(
+                _lastRecordedSlotTimestamp
+                    <= nextEpochTimestamp + LibPreconfConstants.SECONDS_IN_EPOCH,
+                InvalidLookaheadEpoch()
+            );
+
+            lastRecordedSlotTimestamp = _lastRecordedSlotTimestamp;
+
+            // Merkleize the lookahead buffer entries and update the lookahead root for next epoch
+            bytes32 root = LibMerkleTree.generateTree(leaves);
```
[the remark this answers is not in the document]
AnshuJalan: Now that I think about it, may be a simple hashing is better. We need to hash the array of `LookaheadLeaf` struct since that contains the full set of information that we need. We can probably rename it to `LookaheadSlot` in that case. Let's discuss this too tomorrow

AnshuJalan:
> If slots 2, 3, and 4 are empty and slot 5 has a validator, when building the tree, do you use slot 5’s validator to fill the leaves for slots 2, 3, and 4 as well?

Yes

> I’m wondering if, when the lookahead is sparse, it might be simpler to represent it like this:
> 
> `[(slot1, Alice), (slot5, Bob)]`
> 
> Then, when proposing a block in slot 3, the proposer would include this lookahead in calldata, we verify its hash matches the persisted hash, and iterate through the in-memory lookahead to verify that the sender is Bob.
> 
> Lets talk about it in our next meeting.

If we are going with the simple hash approach as you suggested, then I feel we are implicitly getting this level of simplicity.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,283 @@
...
+        returns (address committer)
+    {
+        IRegistry.OperatorData memory OperatorData =
+            urc.getOperatorData(_lookaheadPayload.registrationRoot);
+        require(
+            OperatorData.unregisteredAt == 0 || OperatorData.unregisteredAt >= _epochTimestamp,
+            OperatorHasUnregistered()
+        );
+        require(
+            OperatorData.slashedAt == 0 || OperatorData.slashedAt >= _epochTimestamp,
```
adaki2004, line 233: It seems in the Fabric's registry contract, the `OperatorData.slashedAt` is block based not timestamp based. 
https://github.com/eth-fabric/urc/blob/d730db52fa0953bf1425f28c9c5f1ff8c0a1bc1e/src/Registry.sol#L433

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,283 @@
...
+        uint256 _epochTimestamp
+    )
+        internal
+        view
+        returns (address committer)
+    {
+        IRegistry.OperatorData memory OperatorData =
+            urc.getOperatorData(_lookaheadPayload.registrationRoot);
+        require(
+            OperatorData.unregisteredAt == 0 || OperatorData.unregisteredAt >= _epochTimestamp,
```
adaki2004, line 229: I think the `unregisteredAt` in `OperatorData` is block based.
https://github.com/eth-fabric/urc/blob/d730db52fa0953bf1425f28c9c5f1ff8c0a1bc1e/src/Registry.sol#L110

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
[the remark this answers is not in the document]
dantaik: My PR merged.

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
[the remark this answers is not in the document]
dantaik: Maybe we can use the same function to allow the guardian to overwrite.

- For non-guardians, overwriting is not allowed once a lookahead is posted.
- For guardians, overwriting is always allowed.

The function would behave the same, except for the overwrite permission check. In other words, events should still be emitted even when a guardian overwrites.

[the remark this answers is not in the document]
AnshuJalan: Made this change in the revamp

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
```diff
@@ -59,4 +59,17 @@ library LibPreconfUtils {
...
+    /// @return The block height at the given timestamp.
+    function getBlockHeightAtTimestamp(uint256 _timestamp) internal view returns (uint256) {
+        if (_timestamp < block.timestamp) {
+            return
+                block.number - (block.timestamp - _timestamp) / LibPreconfConstants.SECONDS_IN_SLOT;
+        } else {
+            return
+                block.number + (_timestamp - block.timestamp) / LibPreconfConstants.SECONDS_IN_SLOT;
+        }
+    }
```
adaki2004, line 74: I think we should use pure block based (for condition checking) and not converting blockheight to timestamp, because slots are scheduled every 12s, but it is possible that slots does not have any block proposed.

See this scenarios:
(For the sake of simplicity let's say an epoch is 4 slots in length).
Slot(100) -> B(100) proposed -> TS: 1000
Slot(101) -> B(101) proposed -> TS: 1012
Slot(102) -> No block proposed -> TS: 1024
Slot(103) -> B(102) proposed -> TS: 1036
<----EPOCH_BOUNDARY---->
Slot(104) -> B(103) proposed -> TS: 1048

So when someone tries to call within `B(103)` (`slot(104)`), the math will do the following:
if( 1000 < 1048) then:
 **103 -( 1048 - 1000 ) / 12 = 99 NOT the desired block 100**

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
```diff
@@ -0,0 +1,106 @@
...
+        // The preconfer operator's committer address that is fetched from the slashing commitment.
+        address committer;
+        // URC registration root of the operator
+        bytes32 operatorRegistrationRoot;
+        // Index of the Operator's registration merkle tree leaf that contains the validator for the
+        // slot.
+        uint256 validatorLeafIndex;
+    }
+
+    struct LookaheadHash {
```
dantaik, line 33: if we use the following struct, we can use one slot for storing the LookaheadHash:

struct LookaheadHash {
  uint48 epochTimestamp; 
  bytes26 lookaheadHash
}

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
```diff
@@ -0,0 +1,106 @@
...
+        uint256 validatorLeafIndex;
+    }
+
+    struct LookaheadSlot {
+        // Timestamp of the slot.
+        uint256 timestamp;
+        // The preconfer operator's committer address that is fetched from the slashing commitment.
+        address committer;
+        // URC registration root of the operator
+        bytes32 operatorRegistrationRoot;
```
dantaik, line 27:
```suggestion
        bytes32 registrationRoot;
```

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
```diff
@@ -0,0 +1,106 @@
...
+        // Registration root of the operator in the URC
+        bytes32 registrationRoot;
+        // Index of the Operator's registration merkle tree leaf that contains the validator
+        // for the slot
+        uint256 validatorLeafIndex;
+    }
+
+    struct LookaheadSlot {
+        // Timestamp of the slot.
+        uint256 timestamp;
```
dantaik, line 23:
```suggestion
        uint256 slotTimestamp;
```

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
```diff
@@ -0,0 +1,106 @@
...
+    error PosterHasBeenSlashed();
+    error PosterHasInsufficientCollateral();
+    error PosterHasNotOptedIn();
+    error PosterHasUnregistered();
+    error SlasherIsNotGuardian();
+    error SlotTimestampIsNotIncrementing();
+
+    event LookaheadHashUpdated(uint256 epochTimestamp, bytes32 lookaheadHash);
+    event LookaheadPosted(uint256 indexed timestamp, LookaheadSlot[] lookaheadSlot);
+    event LookaheadPostedByGuardian(uint256 indexed timestamp, LookaheadSlot[] lookaheadSlot);
```
dantaik, line 68: how about we reuse `LookaheadHashUpdated` and `LookaheadPosted` by adding a "isGuardian" field to these two events?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik: Anshu, is here the comimtter is the actual preconfer address (the delegated address of validators)?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,296 @@
...
+                    operatorRegistrationRoot: lookaheadPayload.registrationRoot,
+                    validatorLeafIndex: lookaheadPayload.validatorLeafIndex
+                });
+
+                lookaheadSlots[i] = lookaheadSlot;
+            }
+
+            // Validate that the last slot timestamp is within the next epoch
+            require(
+                lookaheadSlots[lookaheadSlots.length - 1].timestamp
```
dantaik, line 150: `lookaheadSlots.length - 1` may underflow

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
dantaik: In CL, some slots will be empty and do not have corresponding blocks. Therefore, there is no way to map block height to slots.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
adaki2004: `_lookaheadPayloads.length` cannot be 0 here.

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
AnshuJalan: > In CL, some slots will be empty and do not have corresponding blocks. Therefore, there is no way to map block height to slots, or the other way around....

I agree with this. Also, @adaki2004 our main goal here is to keep a "fixed reference point" for checking the status of the operator. This must be the same reference point used in the lookahead slashing contract that shall be introduced after EIP7917. As long as we use the same function and keep the same reference point, we are good.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,280 @@
...
+        require(operatorData.slashedAt == 0, PosterHasBeenSlashed());
+        require(
+            operatorData.collateralWei >= getConfig().minCollateralForPosting,
+            PosterHasInsufficientCollateral()
+        );
+
+        // Validate the slashing commitment of the lookahead poster
+        IRegistry.SlasherCommitment memory slashingCommitment =
+            urc.getSlasherCommitment(_registrationRoot, guardian);
+        require(slashingCommitment.optedOutAt < slashingCommitment.optedInAt, PosterHasNotOptedIn());
```
dantaik, line 169: do we need to compare these two values with the current block number?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,280 @@
...
+
+        // Validate the lookahead poster's signed commitment
+        address committer = ECDSA.recover(
+            keccak256(abi.encode(_signedCommitment.commitment)), _signedCommitment.signature
+        );
+        require(committer == slashingCommitment.committer, CommitmentSignerMismatch());
+        require(_signedCommitment.commitment.slasher == guardian, SlasherIsNotGuardian());
```
dantaik, line 176: In URC's documentation, the slasher is a contract. Shall we use `address(this)` here instead of guardian?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,280 @@
...
+        if (_previousSlotTimestamp == 0) {
+            require(_lookaheadPayload.slotTimestamp >= _nextEpochTimestamp, InvalidLookaheadEpoch());
+        } else {
+            require(
+                _lookaheadPayload.slotTimestamp > _previousSlotTimestamp,
+                SlotTimestampIsNotIncrementing()
+            );
+        }
+
+        require(
```
dantaik, line 197: we can warp this require-statement in a unchecked

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,280 @@
...
+            lookaheadHash = _calculateLookaheadHash(lookaheadSlots);
+        }
+
+        _setLookaheadHash(_nextEpochTimestamp, lookaheadHash);
+        emit LookaheadPosted(
+            _isPostedByGuardian, _nextEpochTimestamp, lookaheadHash, lookaheadSlots
+        );
+    }
+
+    function _validateLookaheadPoster(
```
dantaik, line 150: Also, should we make this function public so people can test it without a transaction.

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
```diff
@@ -44,19 +47,35 @@ library LibPreconfUtils {
...
 
-    /// @notice Calculates the timestamp of a future epoch based on the genesis timestamp.
-    /// @param _epochOffset The offset from the current epoch.
-    /// @return The timestamp of the future epoch.
-    function getEpochTimestamp(uint256 _epochOffset) internal view returns (uint256) {
-        return getEpochTimestamp() + _epochOffset * LibPreconfConstants.SECONDS_IN_EPOCH;
+    /// @notice Calculates the block height at a given timestamp.
+    /// @param _timestamp The timestamp for which the block height is to be retrieved.
+    /// @return The block height at the given timestamp.
+    function getBlockHeightAtTimestamp(uint256 _timestamp) internal view returns (uint256) {
```
dantaik, line 72: for a CL slot with missed EL block, this function is likely not going to work ...

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
```diff
@@ -0,0 +1,99 @@
...
+        LookaheadSlot[] lookaheadSlot
+    );
+
+    /**
+     * @notice Allows a registered operator to post the lookahead for the next epoch.
+     * @param _registrationRoot The registration root of the posting-operator in the URC.
+     * @param _data The signed commitment containing the lookahead data, or the lookahead data if
+     * posted by the guardian.
+     */
+    function updateLookahead(bytes32 _registrationRoot, bytes calldata _data) external;
```
dantaik, line 79:
```suggestion
    function updateLookahead(bytes32 _registrationRoot, bytes calldata _data, bytes calldatea _proof) external;
```

dantaik: Do we want to support a future proof parameter...

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
AnshuJalan: Aware of this. I'll add a comment to clarify - this is fine because the main goal is consistency between lookahead store and lookahead slasher post EIP 7917, if we use the same block number in both, we are good

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
adaki2004: > > In CL, some slots will be empty and do not have corresponding blocks. Therefore, there is no way to map block height to slots, or the other way around....
> 
> I agree with this. Also, @adaki2004 our main goal here is to keep a "fixed reference point" for checking the status of the operator. This must be the same reference point used in the lookahead slashing contract that shall be introduced after EIP7917. As long as we use the same function and keep the same reference point, we are good.

I think Daniel and me are saying the exact same thing. 
imo, if fabric registry is blockheight, we should go with blockheight, but there is no way we can tie a slot's timestamp to a block.

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
adaki2004: > Aware of this. I'll add a comment to clarify - this is fine because the main goal is consistency between lookahead store and lookahead slasher post EIP 7917, if we use the same block number in both, we are good

But we query fabric's registry, no ? Or that is aligned both with our `slashedAt` and `unregisteredAt` ?

AnshuJalan: We do not have our own versions of `slashedAt` and `unregistered`. On our end, we just need to decide on the timestamp at which we intend to compare the queried values against.

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
AnshuJalan: In what scenario could we need a proof here?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: Any valid operator can be a lookahead poster, so may be that is not necessary and can be checked within the urc directly

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
adaki2004: Please correct me if i'm wrong but this is where the discrepancy is coming in. Because now we have 2 (possible) different timestamps for "same block". (Since you try to derive the blockheight from timestamp, and it can differ from the actual, registry's registered block height.)

kép

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: The guardian contract will be the one slashing incorrect lookahead posters pre-EIP7917, so we have the commitment signed for the guardian's contract address.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: That wouldn't be necessary since this is the primary validity condition in the URC for an opted in proposer

packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol
adaki2004: Yeah, will ping you offline :)

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik: If an operator opted in then opted out in a later time, then `slashingCommitment.optedOutAt > slashingCommitment.optedInAt`, the above check will fail. Is it the intended behavior? If so, we are basically checking that:

- slashingCommitment.optedOutAt = 0 and 
- slashingCommitment.optedInAt != 0

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik: Ah, I thought the guardian is an EOA...

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,293 @@
...
+        IRegistry.SlasherCommitment memory slashingCommitment =
+            urc.getSlasherCommitment(_registrationRoot, guardian);
+        require(slashingCommitment.optedOutAt < slashingCommitment.optedInAt, PosterHasNotOptedIn());
+
+        // Validate the lookahead poster's signed commitment
+        address committer = ECDSA.recover(
+            keccak256(abi.encode(_signedCommitment.commitment)), _signedCommitment.signature
+        );
+        require(committer == slashingCommitment.committer, CommitmentSignerMismatch());
+        require(_signedCommitment.commitment.slasher == guardian, SlasherIsNotGuardian());
```
dantaik, line 190: Do we need to validate `commitmentType` in _signedCommitment

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,293 @@
...
+
+        IRegistry.OperatorData memory operatorData =
+            urc.getOperatorData(_lookaheadPayload.registrationRoot);
+        require(
+            operatorData.unregisteredAt == 0
+                || operatorData.unregisteredAt >= blockHeightAtEpochTimestamp,
+            OperatorHasUnregistered()
+        );
+        require(
+            operatorData.slashedAt == 0 || operatorData.slashedAt >= blockHeightAtEpochTimestamp,
```
dantaik, line 240: If operatorData.slashedAt> 0, should we disallow this operator to preconf?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,293 @@
...
+        uint256 _epochTimestamp
+    )
+        internal
+        view
+        returns (address committer)
+    {
+        uint256 blockHeightAtEpochTimestamp =
+            LibPreconfUtils.getBlockHeightAtTimestamp(_epochTimestamp);
+
+        IRegistry.OperatorData memory operatorData =
```
dantaik, line 232: Is it possible to use one function to verify the operatorData in _validateOperatorInLookaheadPayload and _validateLookaheadPoster with different slash amount and slasher parameters?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: Yes, that is fine. The operator should only be opted in at the time of posting the lookahead. The opt out delay in the URC takes care of the rest

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: The URC follows a proxy key structure. So, the actual delegated ECDSA address in the `operator` within the URC, and the operator can assign a `committer` (the proxy key) for signing commitments on the protocol they opt in to

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: `CommitmentType` will be left empty (not necessary to check) and not used by us.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: We need a reference point the lookahead-slasher contract to decide on a point in time at which we need to check if "a preconfer was supposed to be in the lookahead, but not added but not added by the poster" thus, resulting in the slashing of the poster. Since, we are posting a lookahead for the next epoch in the current epoch, the starting timestamp of the current epoch is used as the reference point. (well, right now we use the starting block and hope URC changes it to timestamp)

I had a discussion on this idea of reference point with @adaki2004 yesterday

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: We may do that, but it would be a large function and might get messy with the conditionals. `operator` in the lookahead and `poster` have different slashing requirements. The `poster` simply needs to be opted in at the time of posting and have enough collateral. For the `operator` in lookahead we use the "reference point" as explained earlier, and we are required to check for instance the historical collateral value at that reference point.

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
dantaik: In the future when Lin's EIP is in place, maybe we require the poster to provide a validity proof of the lookahead then we no longer need staking/slashing.

AnshuJalan: Even after Lin's EIP, proving the entire lookahead would be very expensive, as the deterministic lookahead would contain a list of validator indices, so the proof would be really large and expensive

dantaik: I'm OK without adding this extra _proof params. 

The question to ask is whether we want this approach to be optimistic post the EIP or there is a way to generate a ZKP for the entire lookahead. My understanding is that if we do not post the whole lookahead, the early inclusion will be very hard to achieve even post the EIP.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik: Let me catch up with @adaki2004

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: I think the URC has a problem, let me get back on this @dantaik

AnshuJalan: If this issue is fixed, my original code should work fine

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik: > the starting timestamp of the current epoch is used as the reference point.


- For the poster, we simply check their current collateral balance.
- For the preconfer, ideally we’d know their collateral at the slot’s timestamp. However, since that timestamp is in the future, we can’t predict the value — the preconfer might get slashed before their slot (e.g., by other slashers). Is this understanding correct, @AnshuJalan?

Would it be acceptable to use the current timestamp (i.e., when the lookahead is posted) as the reference point instead? That way, we can just use `operatorData.collateralWei` and avoid the more expensive `getHistoricalCollateral`.

AnshuJalan: > For the preconfer, ideally we’d know their collateral at the slot’s timestamp. However, since that timestamp is in the future, we can’t predict the value — the preconfer might get slashed before their slot (e.g., by other slashers). Is this understanding correct, @AnshuJalan?

Yes, correct, giving us the need of having a fixed reference point.

> Would it be acceptable to use the current timestamp (i.e., when the lookahead is posted) as the reference point instead? That way, we can just use operatorData.collateralWei and avoid the more expensive getHistoricalCollateral.

That would mean the poster does not have a fixed reference point to build the lookahead against. The poster may end up putting preconfers that are valid at the current slot timestamp, but if the lookahead doesn't get included in the current block, they may get prone to slashing by a preconfer who got active at that slot, but was not included in the lookahead. So, using the epoch's start timestamp simplifies the process a lot.

dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -0,0 +1,264 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import { LibPreconfConstants as LPC } from "src/layer1/preconf/libs/LibPreconfConstants.sol";
```
AnshuJalan, line 4: @dantaik Do we want to follow this abbreviation convention? I don't see it anywhere else in Taiko repo

dantaik: I'm ok without the abbreviation. Feel free not to use it :D

dantaik (approved):

PR #19336, opened 2025-04-25
feat(protocol): connect lookahead to preconf router

AnshuJalan:
This PR adds a new preconf router that utilises the lookahead store to validate the calling preconfer


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    // Internal functions ----------------------------------------------------------------------
+
+    function _validateWhitelistPreconfer() internal view {
+        address preconfer = preconfWhitelist.getOperatorForCurrentEpoch();
+        if (preconfer != address(0)) {
+            require(msg.sender == preconfer, NotPreconfer());
+        } else if (fallbackPreconfer != address(0)) {
+            require(msg.sender == fallbackPreconfer, NotFallbackPreconfer());
```
adaki2004, line 84: Question to here (and this is also present in `PreconfRouter.sol`). I would kinda think that in any ways, a "fallbackPreconfer" is n actor, because it can preconf and propose batches.

Now it seems, that if there is a whitelist `preconfer` in the the `WhitelistPreconf` it wil return a "random" (deterministic) address - even if an inactive one - this PreconfRouter2 is not really assigning preconf/batching role to the `fallbackPreconfer`. (Maybe it is by design !?)

Dont know, but for me, this would be a bit more logical:
```
        if (preconfer != address(0)) {
            require(msg.sender == preconfer, NotPreconfer());
            return;
        } 
        if (fallbackPreconfer != address(0)) {
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter2.sol
```diff
@@ -0,0 +1,34 @@
...
+    error OperatorIsNotOptedIn();
+
+    /// @notice Allows a preconfer to propose a batch of transactions
+    /// @param _lookahead The lookahead leaf and merkle proof pointing to the lookahead entry of
+    /// the preconfer
+    /// @param _params The parameters for the batch
+    /// @param _txList The list of transactions to propose
+    /// @return info_ The info of the batch
+    /// @return meta_ The metadata of the batch
+    function proposeBatch(
```
dantaik, line 27: with https://github.com/taikoxyz/taiko-mono/pull/19337/files, maybe we can simply reuse `IProposeBatch`

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,124 @@
...
+        ILookaheadStore.LookaheadLeaf memory _lookaheadLeaf,
+        bytes32 _lookaheadRoot,
+        bytes32[] memory _proof
+    )
+        internal
+        view
+    {
+        // Validate the lookahead against the retrieved lookahead root
+        require(
+            LibMerkleTree.verifyProof(
```
dantaik, line 98: so if in the current look ahead, slot 4, 5, are empty and slot 6 is not. when blocks are proposed in slot 5, will the _lookaheadLeaf correspond to slot 5 or slot 6? If it is for slot 6, then do we need to also prove slot 5 is empty?

AnshuJalan: It will correspond to slot 6 itself and this will take care of the advanced slots: 
```solidity
require(
            block.timestamp > _lookaheadLeaf.prevTimestamp
                && block.timestamp < _lookaheadLeaf.timestamp,
            InvalidLookaheadTimestamp()
        );
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
AnshuJalan: Your understanding is correct, but I think both the original code and your proposed code lead to the same outcome

adaki2004: > Your understanding is correct, but I think both the original code and your proposed code lead to the same outcome

Thanks for checking. I think the original code and this is slightly different. With original code, only one path can be activated.

Imagine this scenario:
Alice is preconfer for epoch (for sake of simplicity whitelist now).
Bob address is set to be fallback.

Original:
1. When Alice returned for `getOperatorForCurrentEpoch()` (and she will be returned for the whole epoch) but she gets offline / out of service within epoch, she will be returned anyways. (And for whitelist it is possible afaik that she can be preconfer for next epochs too).
2. Bob as a fallback tries to "help" and provide preconf service and `proposeBatches` but he wont succeed as only one path can be activated and that is Alice's.

This one:
If Bob detects the outage he can jump in and start offering his service and successfully `proposeBatches`

wdyt ?

AnshuJalan: Yes, the approach makes sense

adaki2004: Added as PR for the whitelist version, feel free to take over if you think good.
https://github.com/taikoxyz/taiko-mono/pull/19349

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,220 @@
...
+        // Both TaikoInbox and TaikoWrapper implement the same ABI for IProposeBatch.
+        (info_, meta_) = proposeEntrypoint.v4ProposeBatch(_params, _txList, hex"");
+
+        // Verify that the sender had set itself as the proposer
+        require(info_.proposer == msg.sender, ProposerIsNotPreconfer());
+    }
+
+    // Internal functions ----------------------------------------------------------------------
+
+    function _validateWhitelistPreconfer() internal view {
```
adaki2004, line 99: Maybe add a modifier, just we have like now on main. https://github.com/taikoxyz/taiko-mono/blob/b5c1f96ac326523927cd3ad4822fb4d97cf9e191/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol#L17

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
adaki2004: (still outstanding, cause logic is not 100% correct)

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,220 @@
...
+        ILookaheadStore.LookaheadSlot[] memory _previousLookahead,
+        ILookaheadStore.LookaheadSlot[] memory _currentLookahead,
+        ILookaheadStore.LookaheadSlot memory _lookaheadSlot
+    )
+        internal
+        view
+    {
+        // Validate the current lookahead data
+        require(
+            keccak256(abi.encode(_currentLookahead)) == _currentLookaheadHash,
```
adaki2004, line 140: Should not it be ?
```suggestion
            keccak256(abi.encode(_epochTimestamp, _currentLookahead)) == _currentLookaheadHash,
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,220 @@
...
+            //
+            // For case 2 and 3, we leave _previousLookahead as an empty array.
+
+            if (_previousLookahead.length != 0) {
+                // Validate previous lookahead data
+                bytes32 previousLookaheadHash = lookaheadStore.getLookaheadHash(
+                    _epochTimestamp - LibPreconfConstants.SECONDS_IN_EPOCH
+                );
+                require(
+                    keccak256(abi.encode(_previousLookahead)) == previousLookaheadHash,
```
adaki2004, line 183: Also should include timestamp if we use it for hashing (in `LookaheadStore`)
```suggestion
                    keccak256(abi.encode(_epochTimestamp - LibPreconfConstants.SECONDS_IN_EPOCH, _previousLookahead)) == previousLookaheadHash,
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
AnshuJalan: Fixed in the PR you raised.

packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter2.sol
```diff
@@ -0,0 +1,21 @@
...
+pragma solidity ^0.8.24;
+
+import "src/layer1/based/IProposeBatch.sol";
+import "src/layer1/based/ITaikoInbox.sol";
+
+/// @title IPreconfRouter2
+/// @custom:security-contact security@taiko.xyz
+interface IPreconfRouter2 is IProposeBatch {
+    error ForcedInclusionNotSupported();
+    error InvalidCurrentLookahead();
```
dantaik, line 11: what if we move all errors to the implementation to avoid defining IPreconfRouter2?

Actually I feel like most errors are very implementation specific, maybe we should move all of them from interfaces to implementations.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,229 @@
...
+        bytes26 currentLookaheadHash = lookaheadStore.getLookaheadHash(epochTimestamp);
+
+        // Try fetching the lookahead for the next epoch.
+        // This call fails if the lookahead for next epoch is not posted, thus requiring the first
+        // preconfer to post the next epoch's lookahead before proposing a batch in the current
+        // epoch.
+        lookaheadStore.getLookaheadHash(epochTimestamp + LibPreconfConstants.SECONDS_IN_EPOCH);
+
+        if (_isEmptyLookahead(epochTimestamp, currentLookaheadHash)) {
+            // The current lookahead is empty, so we use a whitelisted preconfer
```
dantaik, line 68: if the current lookahead is empty, can we also force the first proposer to post the lookahead?

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,229 @@
...
+    constructor(
+        address _resolver,
+        address _lookaheadStore,
+        address _preconfWhitelist,
+        address _iProposeBatch,
+        address _preconfSlasher,
+        address _urc,
+        address _fallbackPreconfer
+    )
+        EssentialContract(_resolver)
```
dantaik, line 34:
```suggestion
        EssentialContract()
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,229 @@
...
+    IPreconfWhitelist public immutable preconfWhitelist;
+    IProposeBatch public immutable iProposeBatch;
+    IRegistry public immutable urc;
+    address public immutable preconfSlasher;
+    address public immutable fallbackPreconfer;
+
+    uint256[50] private __gap;
+
+    constructor(
+        address _resolver,
```
dantaik, line 26:
```suggestion
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,229 @@
...
+    function _validateProposer(ILookaheadStore.LookaheadSlot memory _lookaheadSlot) internal view {
+        // Sender must be the expected committer (i.e the preconfer) for the current preconfing
+        // period
+        require(msg.sender == _lookaheadSlot.committer, ProposerIsNotPreconfer());
+
+        // Ensure that the associated operator is active and opted into the preconf slasher
+        IRegistry.OperatorData memory operatorData =
+            urc.getOperatorData(_lookaheadSlot.registrationRoot);
+        require(operatorData.slashedAt == 0, OperatorIsSlashed());
+        require(operatorData.unregisteredAt == 0, OperatorIsUnregistered());
```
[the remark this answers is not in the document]
adaki2004: According to this, we only set it 'now' at up-most. So not in the future.
https://github.com/eth-fabric/urc/blob/5f18225c4d027b49084605129b7666a37ea13412/src/Registry.sol#L72
(This is why i also said it here: https://github.com/taikoxyz/taiko-mono/pull/19445#discussion_r2086375203 )

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
linoscope, line 68: We may want to limit the preconf window of the fallback preconfer to fewer slots (instead of full epoch), like 16 slots. Then, after they post their batch with next epoch lookahead, if the next epoch lookahead contains a preconfer, then the first preconfer in next epoch lookahead can start preconfing after slot 17.

linoscope: On second thought, it's probably not worth the complexity.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -0,0 +1,229 @@
...
+                // Validate previous lookahead data
+                uint256 previousEpochTimestamp =
+                    _epochTimestamp - LibPreconfConstants.SECONDS_IN_EPOCH;
+                bytes26 previousLookaheadHash =
+                    lookaheadStore.getLookaheadHash(previousEpochTimestamp);
+                require(
+                    LibPreconfUtils.calculateLookaheadHash(
+                        previousEpochTimestamp, _previousLookahead
+                    ) == previousLookaheadHash,
+                    InvalidPreviousLookahead()
```
linoscope, line 193: Shouldn't we do this check _before_ doing `if (_previousLookahead.length != 0)` in case `_previousLookahead` is wrong?

dantaik (approved):

dantaik (approved):

PR #19343, opened 2025-04-28
feat(protocol): revamp lookahead store 

AnshuJalan:
Instead of merkle-izing lookahead slots, we may simply hash the slots. For a full lookahead of 32 slots, it is about 3.5-4x cheaper (in terms total hashing cost) to simply hash the lookahead versus merkle-izing it.

This PR also removes `lastRecordedTimestamp` from the storage. It seems to introduce unnecessary complexity for the guardian to correctly overwrite the lookahead. This change requires us to modify the logic in the new `PreconfRouter` that indirectly relied on `lastRecordedTimestamp`.


packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -159,6 +107,62 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
+        returns (bytes32, LookaheadSlot[] memory)
+    {
+        if (_lookaheadPayloads.length == 0) {
+            // The poster claims that the lookahead for the next epoch has no preconfers
+            bytes32 emptyLookaheadHash = _calculateEmptyLookaheadHash(_nextEpochTimestamp);
+            _setLookaheadHash(_nextEpochTimestamp, emptyLookaheadHash);
+
+            return (emptyLookaheadHash, new LookaheadSlot[](0));
+        } else {
+            LookaheadSlot[] memory lookaheadSlots = new bytes32[](_lookaheadPayloads.length);
```
adaki2004, line 125: Should not it be ?
```suggestion
            LookaheadSlot[] memory lookaheadSlots = new LookaheadSlot[](_lookaheadPayloads.length);
```

dantaik:
@AnshuJalan I think you can merge this PR to its base branch, I'll continue reviewing it later on

dantaik (approved):

PR #19347, opened 2025-04-28
feat(protocol): allow sharing base fees with more than 2 parties

dantaik:
@davidtaikocha Let me know if this change (removing the use of extraData) works for the node.
I hope this PR will also be useful when the Taiko Stack is used by other L2s, where they might want to share 1% of their base fees with Taiko Labs or the Taiko DAO. In that case, the base fee could be split three ways.


packages/protocol/contracts/layer1/mainnet/MainnetInbox.sol
```diff
@@ -64,4 +61,37 @@ contract MainnetInbox is TaikoInbox {
...
+
+    function _getBaseFeeSharings()
+        internal
+        pure
+        virtual
+        returns (ITaikoInbox.BaseFeeSharing[] memory baseFeeSharings_)
+    {
+        baseFeeSharings_ = new ITaikoInbox.BaseFeeSharing[](1);
+        baseFeeSharings_[0] = ITaikoInbox.BaseFeeSharing({
+            account: 0x0000000000000000000000000000000000000000, // TODO(david): update this address
```
dantaik, line 93: @davidtaikocha

packages/protocol/contracts/layer1/hekla/HeklaInbox.sol
```diff
@@ -79,36 +79,31 @@ contract HeklaInbox is TaikoInbox {
...
+
+    function _getBaseFeeSharings()
+        internal
+        pure
+        virtual
+        returns (ITaikoInbox.BaseFeeSharing[] memory baseFeeSharings_)
+    {
+        baseFeeSharings_ = new ITaikoInbox.BaseFeeSharing[](1);
+        baseFeeSharings_[0] = ITaikoInbox.BaseFeeSharing({
+            account: 0x0000000000000000000000000000000000000000, // TODO(david): update this address
```
dantaik, line 105: @davidtaikocha

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -62,14 +62,18 @@ interface ITaikoInbox is IBondManager, IProveBatches {
         BlockParams[] blocks;
     }
 
+    struct BaseFeeSharing {
+        address account;
+        uint8 percentage;
+    }
     /// @dev This struct holds batch information essential for constructing blocks offchain, but it
     /// does not include data necessary for batch proving.
+
```
dantaik, line 71:
```suggestion
```

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -62,14 +62,17 @@ interface ITaikoInbox is IBondManager, IProveBatches {
         BlockParams[] blocks;
     }
 
+    struct BaseFeeSharing {
+        address account;
+        uint8 percentage;
+    }
```
dantaik, line 68:
```suggestion
    }

```

davidtaikocha:
Like we discussed before, if we remove the usage of `extraData`, new L2 nodes will have issues when doing the P2P sync.

dantaik:
> Like we discussed before, if we remove the usage of `extraData`, new L2 nodes will have issues when doing the P2P sync, as they don't know the real percentage for each block in P2P network.

I think we have two options here:

- put `baseFeeSharings` as part of the signed block data structure, if these values don't match the batch's info, then the signer will be penalized.
- use the node's current config's  `baseFeeSharings` values

davidtaikocha:
> > Like we discussed before, if we remove the usage of `extraData`, new L2 nodes will have issues when doing the P2P sync, as they don't know the real percentage for each block in P2P network.
> 
> I think we have two options here:
> 
> * put `baseFeeSharings` as part of the signed block data structure, if these values don't match the batch's info, then the signer will be penalized.
> * use the node's current config's  `baseFeeSharings` values

For P2P network, I don't mean the preconf P2P network, I mean the current go-ethereum P2P network, when we start a new L2 node from genesis, it needs to sync through this go-ethereum P2P network, and during the syncing process, when this new node receive a L2 block, it has no way to know the percentage unless its in the `extraData`.

PR #19349, opened 2025-04-28
fix(protocol): fix PreconfRouter permission check issue

adaki2004:
As per description/scenario in here: https://github.com/taikoxyz/taiko-mono/pull/19336#discussion_r2063062610


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -42,12 +42,18 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
...
+
+        if (preconfer != address(0) && msg.sender == preconfer) {
+            isValid = true;
+        }
+
+        if (fallbackPreconfer != address(0) && msg.sender == fallbackPreconfer) {
+            isValid = true;
         }
 
+        require(isValid, NotPreconferOrFallback());
```
dantaik, line 55:
```suggestion
        require(
            preconfer != address(0) && msg.sender == preconfer
            || fallbackPreconfer != address(0) && msg.sender == fallbackPreconfer, 
         NotPreconferOrFallback());
```

dantaik:
Good catch.
Simplified the code with https://github.com/taikoxyz/taiko-mono/pull/19349/commits/55fb7ea0fbdc38414f263e30c447162f745b53c8, please double check there is no issue.

dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -14,6 +14,15 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
 
     uint256[50] private __gap;
 
+    modifier onlyFromEpochOperatorOrFallback() {
+        require(
+            msg.sender == fallbackPreconfer
```
dantaik, line 19: no need to check for address(0) as msg.sender will never be address(0)

AnshuJalan (approved):

dantaik (approved):

PR #19354, opened 2025-04-28
refactor(protocol): use one storage slot for LookaheadHash

dantaik:
This is to optimize away 5K gas (non-zero to non-zero SSTORE)


adaki2004:
A truncated `bytes26` hash offers less collision resistance than a full `bytes32` hash. 
Given we write the lookahead 1x in every 6.4 minutes, gaining 5K gas / epoch, i dont think it worth the security downgrade (while `bytes26` should be safe  in itself too, tho still..).

dantaik:
> A truncated `bytes26` hash offers less collision resistance than a full `bytes32` hash. Given we write the lookahead 1x in every 6.4 minutes, gaining 5K gas / epoch, i dont think it worth the security downgrade (while `bytes26` should be safe in itself too, tho still..).


The following are from ChatGPT:


### Question:
If I keccak some data into a bytes32, how likely is a hash collision? What if I only use the least significant 26 bytes?

### Answer:
Hashing data into a bytes32 (256 bits) with Keccak-256 makes collision probability negligible, requiring around 2^{128} samples to reach a 50% chance (by the birthday paradox). If only the least significant 26 bytes (208 bits) are used, the collision probability slightly increases but still remains negligible, requiring about 2^{104} samples for a significant risk. In both cases, collisions are practically impossible in real-world scenarios.

dantaik:
> > > A truncated `bytes26` hash offers less collision resistance than a full `bytes32` hash. Given we write the lookahead 1x in every 6.4 minutes, gaining 5K gas / epoch, i dont think it worth the security downgrade (while `bytes26` should be safe in itself too, tho still..).
> > 
> > 
> > The following are from ChatGPT:
> > ### Question:
> > If I keccak some data into a bytes32, how likely is a hash collision? What if I only use the least significant 26 bytes?
> > ### Answer:
> > Hashing data into a bytes32 (256 bits) with Keccak-256 makes collision probability negligible, requiring around 2^{128} samples to reach a 50% chance (by the birthday paradox). If only the least significant 26 bytes (208 bits) are used, the collision probability slightly increases but still remains negligible, requiring about 2^{104} samples for a significant risk. In both cases, collisions are practically impossible in real-world scenarios.
> 
> yes, exactly that is what i'm saying in my comment :) Question is: is this really worth 156 gas / L1 block ? (or 5K / epoch, possibly even less cause we will have empty lookaheads i guess too)

I don't think we'll regret saving extra money.

AnshuJalan (approved):
Good change! :)
I had this in mind

PR #19358, opened 2025-04-28
feat(taiko-client): introduce `GET /status` for preconfirmation API server

davidtaikocha:


packages/taiko-client/driver/preconf_blocks/queue.go
```diff
@@ -43,6 +44,7 @@ func (q *payloadQueue) put(id uint64, payload *eth.ExecutionPayload) {
 		id:      id,
 		payload: payload,
 	}
+	q.totalCached++
```
YoGhurt111, line 47: Can the `totalCached` be larger than `maxTrackedPayloads` in design?

davidtaikocha: yes it can be, since it's a counter.

YoGhurt111 (approved):

cyberhorsey (approved):

PR #19359, opened 2025-04-29
feat(protocol): measure and store `lastAnchorGasUsed` in TaikoAnchor

dantaik:
This opens the door for supporting ideas similar to https://x.com/Brechtpd/status/1854192593804177410 but using the second transaction, not the first one.

### DIFF
--- a/packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
+++ b/packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
@@ -68,7 +68,9 @@ abstract contract ShastaAnchor is PacayaAnchor {
 
         signalService.receiveSignals(_signalSlots);
 
-        lastAnchorGasUsed = uint32(ANCHOR_GAS_LIMIT - gasleft());
+        // We need to add one SSTORE from non-zero to non-zero (5000), one addition (3), and one
+        // subtraction (3).
+        lastAnchorGasUsed = uint32(ANCHOR_GAS_LIMIT - gasleft() + 5006);
     }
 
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
```diff
@@ -26,6 +26,12 @@ abstract contract ShastaAnchor is PacayaAnchor {
         shastaForkHeight = _shastaForkHeight;
     }
 
+    modifier recordTxGasUsed() {
+        uint256 _gasleft = gasleft();
+        _;
+        lastAnchorGasUsed = uint32(_gasleft - gasleft());
```
davidtaikocha, line 32: I think here we still miss the `21_000` base gas for calling this method, and all the gas for the calldata input? Not sure if this matters.

dantaik: In protocol, now we enforce the anchor tx's gas limit value to be 1000000

davidtaikocha: I see but not sure if we need the accurate `gasUsed` value for the `anchorV4` transaction, if so, we are still missing one `SSTORE` + one `SUB` in the latest impl?

dantaik: modified the code again

davidtaikocha (approved):

YoGhurt111 (approved):

PR #19361, opened 2025-04-29
feat(taiko-client): gossip req/resp + endofsequencer mark + ws endpoint

cyberhorsey:
can request and respond to blocks to fetch missing blocks. Confirmed working on Helder, just needs some cleanup IMO.

https://github.com/taikoxyz/optimism/pull/28

edit: Currently running on Helder and Masaya, will be pushing tweaks as needed.


packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -298,6 +302,231 @@ func (s *PreconfBlockAPIServer) OnUnsafeL2Payload(
 	return nil
 }
 
+// OnUnsafeL2Response implements the p2p.GossipIn interface.
+func (s *PreconfBlockAPIServer) OnUnsafeL2Response(
```
cyberhorsey, line 306: we can probably just reuse OnUnsafeL2Payload here and change some logs/checks....

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -298,6 +302,234 @@ func (s *PreconfBlockAPIServer) OnUnsafeL2Payload(
...
+				ParentHash:    block.ParentHash(),
+				FeeRecipient:  block.Coinbase(),
+				ExtraData:     block.Extra(),
+				PrevRandao:    eth.Bytes32(block.MixDigest()),
+				BlockNumber:   eth.Uint64Quantity(block.NumberU64()),
+				GasLimit:      eth.Uint64Quantity(block.GasLimit()),
+				GasUsed:       eth.Uint64Quantity(block.GasUsed()),
+				Timestamp:     eth.Uint64Quantity(block.Time()),
+				BlockHash:     block.Hash(),
+				Transactions:  []eth.Data{hexutil.Bytes(txs)},
```
cyberhorsey, line 523: @davidtaikocha getting `unexpected EOF` or `invalid header` on zlib decompressing when receiving the block sometimes, if you have time to look at this tonight lmk

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -36,7 +36,8 @@ type ExecutableData struct {
 // preconf blocks creation requests.
 type BuildPreconfBlockRequestBody struct {
 	// @param ExecutableData engine.ExecutableData the data necessary to execute an EL payload.
-	ExecutableData *ExecutableData `json:"executableData"`
+	ExecutableData  *ExecutableData `json:"executableData"`
+	EndOfSequencing *bool           `json:"endOfSequencing"`
```
thedevbirb, line 40: When calling `s.p2pNode.GossipOut().PublishL2Payload(/*...*/)` it seems that this field is missing! At line 160 of this file I see the following:

```go
			if err := s.p2pNode.GossipOut().PublishL2Payload(
				c.Request().Context(),
				&eth.ExecutionPayloadEnvelope{
					ExecutionPayload: &eth.ExecutionPayload{
						BaseFeePerGas: eth.Uint256Quantity(u256),
						ParentHash:    header.ParentHash,
						FeeRecipient:  header.Coinbase,
						ExtraData:     header.Extra,
						PrevRandao:    eth.Bytes32(header.MixDigest),
						BlockNumber:   eth.Uint64Quantity(header.Number.Uint64()),
						GasLimit:      eth.Uint64Quantity(header.GasLimit),
						GasUsed:       eth.Uint64Quantity(header.GasUsed),
						Timestamp:     eth.Uint64Quantity(header.Time),
						BlockHash:     header.Hash(),
						Transactions:  []eth.Data{reqBody.ExecutableData.Transactions},
					},
				},
				s.p2pSigner,
			); err != nil {
				log.Warn("Failed to propagate the preconfirmation block to the P2P network", "error", err)
			}
```

And `ExecutionPayloadEnvelope` is as follows:
```go
type ExecutionPayloadEnvelope struct {
	EndOfSequencing       *bool             `json:"endOfSequencing,omitempty"`
	ParentBeaconBlockRoot *common.Hash      `json:"parentBeaconBlockRoot,omitempty"`
	ExecutionPayload      *ExecutionPayload `json:"executionPayload"`
}
```

Maybe you forgot to add the `EndOfSequencing` field there?

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -307,10 +316,20 @@ func (s *PreconfBlockAPIServer) GetStatus(c echo.Context) error {
...
+		if ok {
+			endOfSequencingMarkerReceived = true
+		}
+	}
+
 	return c.JSON(http.StatusOK, Status{
 		Lookahead:                     s.lookahead,
 		TotalCached:                   s.payloadsCache.getTotalCached(),
 		HighestUnsafeL2PayloadBlockID: s.highestUnsafeL2PayloadBlockID,
+		EndOfSequencingMarkerReceived: endOfSequencingMarkerReceived,
```
thedevbirb, line 332: Is there a subscription-based way to know when the previous sequencer has provided a `EndOfSequencing` mark? So far it seems we have to poll many times this endpoint to check if we can start sequencing earlier.

thedevbirb (commented):
Two questions!

packages/taiko-client/driver/preconf_blocks/api.go
cyberhorsey: I did! will add, ty

packages/taiko-client/driver/preconf_blocks/api.go
cyberhorsey: no there isn't.. I can add for you and we can add it to the spec. I'll emit the end of sequencing mark and the epoch it's for

packages/taiko-client/driver/preconf_blocks/api.go
cyberhorsey: added!

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -307,10 +317,20 @@ func (s *PreconfBlockAPIServer) GetStatus(c echo.Context) error {
 	s.lookaheadMutex.Lock()
 	defer s.lookaheadMutex.Unlock()
 
+	endOfSequencingMarkerReceived := false
+
+	if s.rpc.L1Beacon == nil {
+		_, ok := s.sequencingEndedForEpoch.Get(s.rpc.L1Beacon.CurrentEpoch())
```
davidtaikocha, line 323: Shall we also allow the current preconfer in his epoch to query the marker of the previous epoch?

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -313,6 +397,279 @@ func (s *PreconfBlockAPIServer) OnUnsafeL2Payload(
...
+
+	return nil
+}
+
+// OnUnsafeL2Response implements the p2p.GossipIn interface.
+func (s *PreconfBlockAPIServer) OnUnsafeL2Response(
+	ctx context.Context,
+	from peer.ID,
+	msg *eth.ExecutionPayloadEnvelope,
+) error {
```
davidtaikocha, line 426: Shall we also check and call `s.sequencingEndedForEpoch.Add` in `OnUnsafeL2Response`?

cyberhorsey: as you can only query a parent block (because you only know if you missed a block when another block comes in ,and the parent isnt in the cache or canonical chain), you will never receive an `OnUnsafeL2Response` with an EndOfSequencing marker. If you miss the *final block*, you missed it, and have already begun sequencing (if you are the sequencer) on top of the last block you have received, after the handover_ms_buffer of the sequencer.

cyberhorsey: I implemented the above feature.

packages/taiko-client/driver/preconf_blocks/api.go
cyberhorsey: Not sure we need to do that - he has already started sequencing because of the `handoverSlots`. so the `lookahead.NextOperator` takes over at `slotsPerEpoch - handoverSlots`.

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, gossip_request_response]
```
YoGhurt111, line 5:
```suggestion
    branches: [main]
```
When it is ready to merge, can quickly revert this change

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -307,10 +317,20 @@ func (s *PreconfBlockAPIServer) GetStatus(c echo.Context) error {
 	s.lookaheadMutex.Lock()
 	defer s.lookaheadMutex.Unlock()
 
+	endOfSequencingBlockHash := common.Hash{}
+
+	if s.rpc.L1Beacon == nil {
```
thedevbirb, line 322:
```suggestion
	if s.rpc.L1Beacon != nil {
```

A typo here! Only if the L1 beacon client is NOT nil we should run the code below.

packages/taiko-client/driver/driver.go
```diff
@@ -386,87 +395,150 @@ func (d *Driver) cacheLookaheadLoop() {
...
+		if err != nil {
+			log.Error("Failed to fetch the latest L1 head for lookahead", "error", err)
+
+			return err
+		}
+
+		if latestSeenBlockNumber == seenBlockNumber {
+			// Leave some grace period for the block to arrive.
+			if lastSlot != currentSlot &&
+				uint64(time.Now().UTC().Unix())-d.rpc.L1Beacon.TimestampOfSlot(currentSlot) > 6 {
```
YoGhurt111, line 453: What does the `6` mean here?

packages/taiko-client/driver/driver.go
```diff
@@ -386,87 +395,150 @@ func (d *Driver) cacheLookaheadLoop() {
...
+						"error", err,
 					)
-
-					lastSlot = currentSlot
 				}
-
-				continue
+			} else {
+				log.Info(
+					"end of sequencing already seen",
```
YoGhurt111, line 427:
```suggestion
					"End of sequencing already seen",
```

packages/taiko-client/driver/driver.go
```diff
@@ -386,87 +395,150 @@ func (d *Driver) cacheLookaheadLoop() {
...
+				log.Info("lookahead requesting end of sequencing for epoch", "epoch", epoch)
+				if err := d.p2pNode.GossipOut().PublishL2EndOfSequencingRequest(
+					context.Background(),
+					epoch,
+				); err != nil {
 					log.Warn(
-						"Lookahead possible missed slot detected",
-						"currentSlot", currentSlot,
-						"latestSeenBlockNumber", latestSeenBlockNumber,
+						"failed to publish end of sequencing request",
```
YoGhurt111, line 420:
```suggestion
						"Failed to publish end of sequencing request",
```

packages/taiko-client/driver/driver.go
```diff
@@ -386,87 +395,150 @@ func (d *Driver) cacheLookaheadLoop() {
...
+		if isSequencer && !*wasSequencer {
+			log.Info("lookahead transitioning to sequencing for operator")
 
-			if latestSeenBlockNumber == seenBlockNumber {
-				// Leave some grace period for the block to arrive.
-				if lastSlot != currentSlot &&
-					uint64(time.Now().UTC().Unix())-d.rpc.L1Beacon.TimestampOfSlot(currentSlot) > 6 {
+			hash, seen := d.preconfBlockServer.GetSequencingEndedForEpoch(epoch)
+			if !seen {
+				log.Info("lookahead requesting end of sequencing for epoch", "epoch", epoch)
```
YoGhurt111, line 414:
```suggestion
				log.Info("Lookahead requesting end of sequencing for epoch", "epoch", epoch)
```

packages/taiko-client/driver/driver.go
cyberhorsey: it can be moved to a constant for readability, it's simply the grace period of seconds we allow for a missed slot

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/ontake.go
```diff
@@ -191,3 +191,8 @@ func (i *BlocksInserterOntake) InsertBlocks(
 
 	return nil
 }
+
+// not used
```
davidtaikocha, line 195: So shall we just remove this method?

packages/taiko-client/driver/chain_syncer/chain_syncer.go
```diff
@@ -46,12 +46,13 @@ func New(
 	p2pSync bool,
 	p2pSyncTimeout time.Duration,
 	blobServerEndpoint *url.URL,
+	latestBlockIDSeenInEventCh chan uint64,
```
davidtaikocha, line 49: Is this identical to `headL1Origin`?

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -69,6 +71,29 @@ func (s *PreconfBlockAPIServer) BuildPreconfBlock(c echo.Context) error {
...
+	if err != nil {
+		return s.returnError(c, http.StatusInternalServerError, err)
+	}
+
+	if parent.Number.Uint64() < s.latestBlockIDSeenInEvent {
+		log.Warn("The parent block ID is smaller than the latest block ID seen in event",
+			"parentBlockID", parent.Number.Uint64(),
+			"latestBlockIDSeenInEvent", s.latestBlockIDSeenInEvent,
+		)
+		return s.returnError(c, http.StatusBadRequest,
```
davidtaikocha, line 84: seems https://github.com/taikoxyz/taiko-mono/blob/main/packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go#L293 will also handle this issue?

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go
```diff
@@ -99,6 +102,11 @@ func (i *BlocksInserterPacaya) InsertBlocks(
 		lastPayloadData *engine.ExecutableData
 	)
 
+	// update lastBlockID
+	if i.latestBlockIDSeenInEventCh != nil {
+		i.latestBlockIDSeenInEventCh <- meta.GetLastBlockID()
```
davidtaikocha, line 107: Since its sent / updated before the real unpreconf blocks / `headL1Origin` inserted into the L2 node, will there be an issue?

packages/taiko-client/driver/driver.go
```diff
@@ -385,87 +399,152 @@ func (d *Driver) cacheLookaheadLoop() {
...
 
-			latestSeenBlockNumber, err := d.rpc.L1.BlockNumber(d.ctx)
-			if err != nil {
-				log.Error("Failed to fetch the latest L1 head for lookahead", "error", err)
-				continue
-			}
+		err := d.preconfBlockServer.CheckLookaheadHandover(d.PreconfOperatorAddress, slot)
+		isSequencer := err == nil
+
+		if isSequencer && !*wasSequencer {
```
davidtaikocha, line 413: does `!*wasSequencer` means `wasSequencer` will never be `nil`? otherwise it will panic?

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -511,8 +995,10 @@ func (s *PreconfBlockAPIServer) UpdateLookahead(l *Lookahead) {
...
 }
 
-// checkLookaheadHandover returns nil if feeRecipient is allowed to build at slot globalSlot (absolute L1 slot).
-func (s *PreconfBlockAPIServer) checkLookaheadHandover(feeRecipient common.Address, globalSlot uint64) error {
+// CheckLookaheadHandover returns nil if feeRecipient is allowed to build at slot globalSlot (absolute L1 slot).
+// and checks the  handover window to see if we need to request the end of sequencing
+// block.
+func (s *PreconfBlockAPIServer) CheckLookaheadHandover(feeRecipient common.Address, globalSlot uint64) error {
 	s.lookaheadMutex.Lock()
 	la := s.lookahead
```
davidtaikocha, line 1003: It's just a copy of a pointer? shall we change the `s.lookaheadMutex.Unlock()` to `defer s.lookaheadMutex.Unlock()`?

cyberhorsey: done

packages/taiko-client/driver/driver.go
cyberhorsey: because the arg is passed into the function but we can actually ignore that and treat it the same as `lastSeenBlockNumber` here and not pass it into the closure

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go
cyberhorsey: But every Event *will* actually be processed, or else the chain will halt regardless, so we dont ever want to preconf state behind `meta.GetLastBlockID()` of the latest Event. Right? If there is an error in the event insertion, we still *dont want to preconf state behind meta.GetLastBlockID()`, there is an error in the driver itself that needs to recover or bug fix to process the events, and without this, `HeadL1Origin` doesnt update, or updates later (because it takes some time to process each block), and the sequencer can call `/preconfBlocks` and preconf bad state?

packages/taiko-client/driver/chain_syncer/chain_syncer.go
cyberhorsey: it updates before headL1Origin because we need to cycle through every single block first to check if its preconfirmed which in a large batch can take some time and sequencers will be able to preconf a block that will just get re-orged/wiped out, but gossiped.

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/ontake.go
cyberhorsey: yeah its a hangover from previous code, removed

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -310,6 +401,374 @@ func (s *PreconfBlockAPIServer) OnUnsafeL2Payload(
...
+	defer s.mu.Unlock()
+	// Ignore the message if it is from the current P2P node.
+	if from != "" && s.p2pNode.Host().ID() == from {
+		log.Debug("Ignore the message from the current P2P node", "peer", from)
+
+		return nil
+	}
+
+	// only respond if you are the current sequencer, *not* the active sequencer in the slot.
+	if s.preconfOperatorAddress.Hex() != s.lookahead.CurrOperator.Hex() {
```
davidtaikocha, line 696: If the sidecar just starts at its own slot, not during the last one's handover window, will there be an issue?

cyberhorsey: Yes, we should actually let anyone respond with the current block and remove this check, but just make sure that if many people respond with the same block, that it's OK, and we also don't hit the limit in the optimism lib to ignore messages.

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go
```diff
@@ -154,7 +175,7 @@ func (i *BlocksInserterPacaya) InsertBlocks(
 				parent,
 			)
 			if err != nil {
-				log.Debug("Failed to check if batch is in canonical chain already", "batchID", meta.GetBatchID(), "err", err)
+				log.Warn("Failed to check if batch is in canonical chain already", "batchID", meta.GetBatchID(), "err", err)
```
davidtaikocha, line 178: Shall we also update the wording in this `WARN`, since most of time its not `Failed to`, but just because batch is not preconfed.

davidtaikocha (approved):

YoGhurt111 (approved):

PR #19362, opened 2025-04-30
chore(protocol): refactor LookaheadStore

dantaik:


packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -279,15 +275,4 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
     {
         return lookahead[_epochTimestamp % getConfig().lookaheadBufferSize];
     }
-
-    function _calculateLookaheadHash(
```
AnshuJalan, line 283: @dantaik Afaik, it is cheaper to use the internal function when calling directly from the same contract, because the pointers get re-used

dantaik: reverted

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -146,16 +143,15 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
 
             // Validate that the last slot timestamp is within the next epoch
             require(
-                lookaheadSlots[lookaheadSlots.length - 1].slotTimestamp
-                    <= _nextEpochTimestamp + LibPreconfConstants.SECONDS_IN_EPOCH,
+                prevSlotTimestamp <= _nextEpochTimestamp + LibPreconfConstants.SECONDS_IN_EPOCH,
```
dantaik, line 146: @AnshuJalan shall this be "<"?

AnshuJalan: I think yes, you are right .

AnshuJalan (approved):
Looks good to me, expect that `<=` to be changed to `<` that you pointed out.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik, line 146:
```suggestion
                prevSlotTimestamp < _nextEpochTimestamp + LibPreconfConstants.SECONDS_IN_EPOCH,
```

PR #19365, opened 2025-04-30
feat(protocol): add block type in preconf payload

AnshuJalan:
Discussion reference: https://github.com/taikoxyz/taiko-mono/pull/19330#discussion_r2061126822

Slashing is only available for messing up `BatchEndAndEOP` since that is directly connected to handover of preconfing to next preconfer and must be correct. The preconfing-node <> driver communication will have no concept of Batch otherwise afaiu.


packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -7,6 +7,13 @@ import "@eth-fabric/urc/ISlasher.sol";
 /// @title IPreconfSlasher
 /// @custom:security-contact security@taiko.xyz
 interface IPreconfSlasher is ISlasher {
+    enum BlockType {
+        BatchStart,
+        BatchEnd,
```
linoscope, line 12: I think `BatchStart` is redundant because we know that from the incrementing `batchId`. Also, if we need an explicit `BatchEnd` it will remove flexibility to dynamically choose the end of batch (you aren't sure if a specific preconf is end of batch without knowing the size of the next preconf). @dantaik What use case did you have in mind for these values?

linoscope: Only potentially useful enum item I can think of now is an `Empty` item which explicitly signifies an empty preconf, so preconfer can continue publishing "heart beats" even when mempool is empty. Not 100% sure if it's needed right now (it can be expressed via empty `blockhash`, or the fair exchange overseer can also monitor mempool to know if it's empty).

dantaik: Ok, lets keep these changes on hold. After Anshu's Slasher pr is merged, we can ask ourselves if these changes are really needed.

dantaik:
@AnshuJalan Do you still want to merge this PR or you've changed your mind?

PR #19368, opened 2025-04-30
feat(protocol): gas optimization and fallback fix for `PreconfRouter2`

adaki2004:
2 main changes:
- put `calculateLookaheadHash()` into the `LibPreconfUtils` cause it is stateless and called by 2 contracts too (`PreconfRouter2` and `LookaheadStore`) - so that in `PreconfRouter2` we dont initiate and external call, which saves us lot of gas (approx. 800-1000 depending on the calldata) / each `proposeBatch()`
- When whitelist branch is activated, the logic was still incorrect. (See same mods merged into main already:  https://github.com/taikoxyz/taiko-mono/pull/19349/files )


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -96,14 +105,7 @@ contract PreconfRouter2 is IPreconfRouter2, EssentialContract {
...
 
-    function _validateWhitelistPreconfer() internal view {
-        address preconfer = preconfWhitelist.getOperatorForCurrentEpoch();
-        if (preconfer != address(0)) {
-            require(msg.sender == preconfer, NotPreconfer());
-        } else if (fallbackPreconfer != address(0)) {
-            require(msg.sender == fallbackPreconfer, NotFallbackPreconfer());
-        }
-    }
+    function _validateWhitelistPreconfer() internal view onlyFromPreconferOrFallback { }
```
dantaik, line 108: usually when you have a internal validation function, the modifier uses the function, not the other way around.

Since onlyFromPreconferOrFallback is not used by external/public functions, I still prefer the implementation code be kept in _validateWhitelistPreconfer

dantaik: Only this concern, the rest looks good to me.

adaki2004: Ok, added modifier body to function body.

PR #19369, opened 2025-04-30
refactor(protocol): try to unify validator validation into `_validateOperator`

dantaik:
@adaki2004 I'm not sure I understand Anshu's brief comments in the other PR. Do you think this PR make sense? I'm trying to add a _validateValidator function.

@AnshuJalan let me know if the _validateValidator will work, this will help me understand the differences in the validation logics for poster and preconfers.


packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -251,4 +260,15 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
+    function _calculateLookaheadHash(
+        uint256 _epochTimestamp,
+        LookaheadSlot[] memory _lookaheadSlots
+    )
+        internal
+        pure
+        returns (bytes26)
+    {
+        return bytes26(keccak256(abi.encode(_epochTimestamp, _lookaheadSlots)));
+    }
```
adaki2004: Small comment, i think this one shall go to the `LibPreconfUtils`, cause it is stateless and the `PreconfRouter2` is calling into this contract (via the public function). We can avoid doing that if this is in the library, so saving a lot of gas per each `proposeBatch`. (See my suggestion: https://github.com/taikoxyz/taiko-mono/pull/19368 )

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -167,98 +179,57 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
+        return abi.decode(_signedCommitment.commitment.payload, (LookaheadPayload[]));
     }
 
-    /// @dev Validates if the operator is registered and has not been slashed at the given epoch
-    /// timestamp. We use the epoch timestamp of the epoch in which the lookahead is posted to
-    /// validate the registration and slashing status.
-    function _validateOperatorInLookaheadPayload(
-        LookaheadPayload memory _lookaheadPayload,
-        uint256 _epochTimestamp
+    function _validateOperator(
```
dantaik, line 199: @adaki2004 based on your understanding, is this function reusable?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -167,98 +179,58 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
-            PosterHasInsufficientCollateral()
-        );
+        require(_signedCommitment.commitment.slasher == guardian, SlasherIsNotGuardian());
 
-        // Validate the slashing commitment of the lookahead poster
-        IRegistry.SlasherCommitment memory slasherCommitment =
-            urc.getSlasherCommitment(_registrationRoot, guardian);
-        require(slasherCommitment.optedOutAt < slasherCommitment.optedInAt, PosterHasNotOptedIn());
+        (, IRegistry.SlasherCommitment memory slasherCommitment) = _validateOperator(
+            _registrationRoot, block.timestamp, getConfig().minCollateralForPosting, guardian
```
dantaik, line 187: note that block.timestamp is used here for calling `getHistoricalCollateral` inside _validateOperator.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
adaki2004: According to my understanding, this (original) function basically  during trying to register the lookahead for the `nextEpoch` in the `currentEpoch` does the following:
1. Checks if operator (in the current epoch) is still in the operator contract.
2. not unregistered  at all OR not unregistered BEFORE this epoch started (this one is a little bit troublesome check, cause of the possible `block.height` mismatch -> But not critical too much if we reuse the same block.height across this contract, since we just have the same "reference point". There might be some edge case tho, where the URC says he is unregistered already but we think he did not, so we allow him to update the lookahead, but since i dont see any slashing just yet in this code it is up to us how do we handle, but i dont see it safety critical. Let alone if we change to block.timestamp in Fabri'c registry -> the issue will be solved. )
3. not slashed at all OR slashing happened not happened BEFORE this epoch started (-> same applies here as point nr. 2).

Please @AnshuJalan let me know if i understand it correctly.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -146,16 +159,15 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
 
             // Validate that the last slot timestamp is within the next epoch
             require(
-                lookaheadSlots[lookaheadSlots.length - 1].slotTimestamp
-                    <= _nextEpochTimestamp + LibPreconfConstants.SECONDS_IN_EPOCH,
+                prevSlotTimestamp <= _nextEpochTimestamp + LibPreconfConstants.SECONDS_IN_EPOCH,
```
dantaik, line 162: @AnshuJalan should this be "<" instead?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik: it's about the diference of using `getHistoricalCollateral` vs `operatorData.collateralWei`

dantaik: All timestamp we use for `getHistoricalCollateral` calls are all future timestamps, shall we simply use `operatorData.collateralWei`? @AnshuJalan 

Sorry I'm a bit slow on this -- haven't really reviewed the URC code yet, that's why the lack of URC knowledge.

dantaik:
> Generally speaking, can we say it is basically almost the same (or close to identical) logic wise what we have right now ? Somehwat i see it like that - sure some return values and memory vars changed, but overall the logic seems identical. If so, i think it is just small semantics then.
> 
> Imo, if we mapped out the functionality, then we can do refactoring later if
> 
> * it makes sense
> * makes the code more readable
> 
> but first maybe map out the logic :)
> 
> (also i think fabric's registry should be `block.timestamp` based as @dantaik already created the PR so that there wont be any kind of mismatch possibility)

I think we just need be 100% sure about the logics :D

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -127,14 +132,27 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
 
                 prevSlotTimestamp = lookaheadPayload.slotTimestamp;
 
+                prevSlotTimestamp = lookaheadPayload.slotTimestamp;
```
dantaik, line 135:
```suggestion
```

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -212,21 +206,25 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
+
         operatorData_ = urc.getOperatorData(_registrationRoot);
         require(
-            operatorData_.unregisteredAt == 0 || operatorData_.unregisteredAt >= _timestamp,
+            operatorData_.unregisteredAt == 0 || operatorData_.unregisteredAt >= referenceTimestamp,
             OperatorHasUnregistered()
         );
         require(
-            operatorData_.slashedAt == 0 || operatorData_.slashedAt >= _timestamp,
+            operatorData_.slashedAt == 0 || operatorData_.slashedAt >= referenceTimestamp,
```
adaki2004: i think the second part's  (`operatorData_.slashedAt >= referenceTimestamp` ) equality check **can** evaluate to true, but that `operatorData_.slashedAt > block.timestamp` is not. If Fabric's registry is also using timestamp now (which i now see this one is using your PR) - it will maximum be equal if we are slashing the operator `now`. (We dont "slash in the future" i think.)

So i think this is why Ansu was using last epoch's time.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -212,21 +206,25 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
             IRegistry.SlasherCommitment memory slasherCommitment_
         )
     {
+        uint256 referenceTimestamp = block.timestamp;
+
         operatorData_ = urc.getOperatorData(_registrationRoot);
         require(
-            operatorData_.unregisteredAt == 0 || operatorData_.unregisteredAt >= _timestamp,
+            operatorData_.unregisteredAt == 0 || operatorData_.unregisteredAt >= referenceTimestamp,
```
adaki2004: Same applies here, to `_unregisteredAt`, like here: https://github.com/taikoxyz/taiko-mono/pull/19369/commits/a1e718505e43dd6d734393605cb3e782a8a8e037#r2070128810

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: > All timestamp we use for getHistoricalCollateral calls are all future timestamps, shall we simply use operatorData.collateralWei
They are past timestamps. A poster may not post the lookahead exactly at the beginning of the epoch, it could get delayed, but we need the collateral at the reference point i.e the epoch start

AnshuJalan (approved):
I think this PR is fine in the form it is now. `_validateOperator` should work correctly

packages/protocol/package.json
```diff
@@ -54,7 +54,7 @@
     "p256-verifier": "github:taikoxyz/p256-verifier#v0.1.0",
     "risc0-ethereum": "github:risc0/risc0-ethereum#v2.0.2",
     "solady": "github:Vectorized/solady#v0.0.231",
-    "urc": "github:eth-fabric/urc#main",
+    "urc": "github:dantaik/urc#use_time_instead_if_height",
```
dantaik, line 57:
```suggestion
    "urc": "github:eth-fabric/urc#main",
```

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik: Can we use the timestamp when the lookahead is posted instead? The poster becomes slashable only after publishing an invalid lookahead, so their collateral at the time of posting is what really matters. Otherwise, if they had sufficient collateral at the start of the epoch but were slashed at time x (where x > epoch start), then by the time they post an invalid lookahead at time y (where y > x), their collateral might be too low.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -192,47 +198,37 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
+            IRegistry.OperatorData memory operatorData_,
+            IRegistry.SlasherCommitment memory slasherCommitment_
+        )
     {
-        IRegistry.OperatorData memory operatorData =
-            urc.getOperatorData(_lookaheadPayload.registrationRoot);
+        operatorData_ = urc.getOperatorData(_registrationRoot);
         require(
-            operatorData.unregisteredAt == 0 || operatorData.unregisteredAt >= _epochTimestamp,
+            operatorData_.unregisteredAt == 0 || operatorData_.unregisteredAt >= _timestamp,
```
dantaik, line 216:
```suggestion
            operatorData_.unregisteredAt == 0 || operatorData_.unregisteredAt > _timestamp,
```

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -191,54 +226,57 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
-        uint256 _epochTimestamp
+    /// validate the registration and sloashing status.
+    /// @param _registrationRoot The registration root of the operator
+    /// @param _operationTimestamp The timestamp of the operation (posting or preconfing)
+    /// @param _collateralTimestamp The timestamp for querying the collateral balance
+    /// @param _minCollateral The minimum collateral required for the operation
+    /// @param _slasher The slasher address
+    function _validateOperator(
+        bytes32 _registrationRoot,
+        uint256 _operationTimestamp,
```
AnshuJalan, line 237: We do not need this different reference point `_operationTimestamp`, the previous implementation where the same `_referenceTimestamp` was used to check both collateral value and slashing/registration status was correct.

In this implementation, the `_operationTimestamp` for an operator within lookahead is the slot's timestamp, but the slot is in the future so the `slashedAt` and `unregisteredAt` timestamp will always be less than `_operationTimestamp`. 

One extra point that might make it clearer: So, if I am posting the lookahead for `epoch X + 1` in `epoch X`, I only want to check if the operator I am adding in the lookahead was:
- Was registered when `epoch X` started 
- Was not slashed when `epoch X` started
- Opted in when `epoch X` started

Even if the operator became inactive (i.e one of the above conditions was not fulfilled) after the epoch started, I'll still add the preconfer in the lookahead because my "view of the lookahead" is the beginning of the current epoch, and not when it gets posted.

AnshuJalan (changes requested):

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik: Thanks Anshu. I'm leaving my desktop, could you follow up with a fix to this pull request for me, then feel free to merge it to your branch.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -192,53 +209,44 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
-            OperatorHasBeenSlashed()
+            operatorData_.unregisteredAt == 0 || operatorData_.unregisteredAt > _timestamp,
+            OperatorHasUnregistered()
         );
         require(
-            _lookaheadPayload.validatorLeafIndex < operatorData.numKeys, InvalidValidatorLeafIndex()
+            operatorData_.slashedAt == 0 || operatorData_.slashedAt > _timestamp,
+            OperatorHasBeenSlashed()
         );
+        require(operatorData_.collateralWei >= _minCollateral, OperatorHasInsufficientCollateral());
```
dantaik, line 238:
```suggestion
          uint256 collateralWei = _collateralTimestamp >= block.timestamp
            ? operatorData_.collateralWei
            : urc.getHistoricalCollateral(_registrationRoot, _collateralTimestamp);
```

dantaik: @AnshuJalan

PR #19371, opened 2025-05-01
chore(protocol): rename entrypoint to iProposeBatch

dantaik:


adaki2004 (approved):
LGTM but for consistency then we need this too : https://github.com/taikoxyz/taiko-mono/pull/19372

 (Maybe merging there then - cause targetting diff branches ? )

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -9,7 +9,7 @@ import "../iface/IPreconfWhitelist.sol";
 /// @title PreconfRouter
 /// @custom:security-contact security@taiko.xyz
 contract PreconfRouter is EssentialContract, IProposeBatch {
-    IProposeBatch public immutable proposeEntrypoint;
+    IProposeBatch public immutable  iProposeBatch;
```
dantaik, line 12:
```suggestion
    IProposeBatch public immutable iProposeBatch;
```

PR #19374, opened 2025-05-02
feat(protocol): enable using existing blobs

dantaik:
Using BlobRefRegistry to enable proposing batches using blobs that are created in early transactions.


packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -62,20 +62,24 @@ interface ITaikoInbox is IBondManager, IProveBatches {
         BlockParams[] blocks;
     }
 
+    struct BlobInfo {
+        bytes32[] hashes;
+        bytes32 refHash;
```
adaki2004, line 67: So when the block proposal comes with blob carrying transactions, we can have multiple blob (hashes), that is the `bytes32[] hashes;`.

But my question is: isnt' it possible to reference multiple `refHash` (when was posted previously) too ? Why `refHash` is not an array here, that is my question.

dantaik: Given Ethereum will support more blobs, a single ref pointing to multiple blobs within the same block should be sufficient for now. While using refHash[] is feasible, it would require additional handling on the node/driver side. We can switch to refHash[] later if needed.

PR #19378, opened 2025-05-02
fix(protocol): ensure empty blobHashes for normal proposals in TaikoWrapper

dantaik:
Before this fix, one can propose a batch with an non-exist and non-zero blob hashes


YoGhurt111 (approved):

davidtaikocha (approved):

smtmfft (approved):

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -43,7 +43,8 @@ interface ITaikoInbox {
         uint32 byteOffset;
         // The byte size of the blob.
         uint32 byteSize;
-        // The block number when the blob was created.
+        // The block number when the blob was created. This valeu is only non-zero when
```
adaki2004, line 46:
```suggestion
        // The block number when the blob was created. This value is only non-zero when
```

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -43,7 +43,8 @@ interface ITaikoInbox {
...
+        // `ablobHashes` are non-empty.
```
adaki2004, line 47:
```suggestion
        // `blobHashes` are non-empty.
```

davidtaikocha (approved):

PR #19382, opened 2025-05-03
feat(protocol): add ProverBlacklist to allow provers to blacklist proposers

dantaik:
If a proposer submits batches that are too expensive to zk-prove, the prover can blacklist the proposer, removing any future obligation to prove that proposer’s batches.

**One issue with this blacklist is that when there are not many preconfers, a malicious prover can blacklist all preconfers then use very low bid to effectively disable the prover market entirely. Need to fix it but don't know how!**


dantaik:
One issue with this blacklist is that when there are not many preconfers, a malicious prover can blacklist all preconfers the use very low bid to effectively disable the prover market entirely. Need to fix it.

dantaik:
EDIT: see https://github.com/taikoxyz/taiko-mono/issues/19384

PR #19385, opened 2025-05-05
feat(protocol): replace onchain ProverMarket with offchain markets

dantaik:
## Summary

This PR removes the recently introduced `IProverMarket` interface and implementation and instead allows proposers to obtain prover signatures off-chain for authentication.

## Rationale

ZK-proving EVM code can be extremely costly, especially if a malicious proposer deliberately constructs blocks that are expensive to prove. Given Taiko’s permissionless design, we must mitigate this vector; otherwise, the protocol is vulnerable to spam attacks. Without proper prover protection, all provers could be discouraged from participating.

## Proposed Solution

This PR reflects a shift in design: instead of requiring on-chain coordination through IProverMarket, provers can now inspect batches off-chain and provide signed quotes to proposers. The proposer selects the best offer and includes the corresponding prover’s signature in the block proposal. This ensures both flexibility and cost protection.

## Implementation

One open question in the current implementation is how to compute the `authHash` - i.e., what exactly the prover should sign. Since the `BatchInfo` and `BatchMetadata` are not available prior to the batch being proposed, the prover cannot sign over them directly.

### TODO 

- [x] @adaki2004 We may want to define a pre-proposal data structure that includes all necessary information for the prover to evaluate and quote a batch, and that can be deterministically linked to the eventual proposal. This ensures the signed authHash remains valid and verifiable when the batch is later submitted on-chain.
- [ ] @adaki2004 Add unit tests.
- [x] @adaki2004 Allow any ERC20 token or even Ether to be used as the prover fee payment token. If the token is the same as the bond token, use the bond balance directly to save gas.
- [x] @adaki2004 Support contract address as prover address - https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/SignatureChecker.sol

---

_Tip for reviewers: This PR consists of two main commits 1) Removal of IProverMarket 2) Implementation of prover signature verification. You may want to focus on [the second commit](https://github.com/taikoxyz/taiko-mono/pull/19385/commits/7c38bf2e6c7fce8fd85859d3f7b9f9180ddd620e) only._


packages/protocol/test/layer1/based/InboxTest_ProposeAndProve.t.sol
```diff
@@ -586,218 +586,223 @@ contract InboxTest_ProposeAndProve is InboxTestBase {
         );
 
         console2.log(str);
-        vm.writeFile("./gas-reports/inbox_without_provermarket.txt", str);
+        vm.writeFile("inbox_without_provermarket.txt", str);
```
dantaik, line 589:
```suggestion
        vm.writeFile("./gas-reports/inbox_without_provermarket.txt", str);
```

packages/protocol/contracts/layer1/based/LibProverAuth.sol
```diff
@@ -0,0 +1,39 @@
...
+/// .
+/// @custom:security-contact security@taiko.xyz
+library LibProverAuth {
+    using ECDSA for bytes32;
+
+    struct ProverAuth {
+        uint96 fee;
+        uint64 validUntil;
+        bytes signature;
+    }
```
adaki2004, line 18: Will try to work on a solution (targeting this PR as base) which might have more replay protection.

Just for reference, when we lauched mainnet we had these hooks checking the assignment like below:

```
struct ProverAssignment {
        address feeToken;
        uint64 expiry;
        uint64 maxBlockId;
        uint64 maxProposedIn;
        bytes32 metaHash;
        bytes32 parentMetaHash;
        TaikoData.TierFee[] tierFees;
        bytes signature;
    }
```

The current problem tho we have is, that the input to the `proposeBatch`, is not deterministically deriveable to an end-state:

See what data we have at proposal time:
kép

@dantaik @davidtaikocha can we put something into the `BatchParams`, i mean is it something doable, that we have an "post-state-hash" ?
```
    struct BatchParams {
        ...
        BlobParams blobParams;
        BlockParams[] blocks;
        bytes32 postState
```

dantaik: I think the following can be hashed into the `dataHash` to be signed by provers:

- The final BatchParams or its hash
- The txList calldata or its hash
- The proposer address
- The batch's ID -- if requested by the prover
- The validUntil -- if requested by the prover

I'm not sure about how postState can be used here. Maybe a small draft PR to show your idea?

adaki2004: > I think the following can be hashed into the `dataHash` to be signed by provers:
> 
>     * the chain id
> 
>     * The final BatchParams or its hash
> 
>     * The txList calldata or its hash
> 
>     * The proposer address
> 
>     * The batch's ID -- if requested by the prover
> 
>     * The validUntil -- if requested by the prover
> 
> 
> I'm not sure about how postState can be used here. Maybe a small draft PR to show your idea?

Yes, exactly that is my question, maybe i did not articulated it properly. 

Is my understanding/question correct that: we want to avoid scenario where malicious proposer can make a prover spend a lot of time/resource on proving a batch/block ?

Because if the answer is yes, then the problem is we need to know (somehow, and i dont know, how the node can get that when prover would sign a prover- assigment) the post-state and put it into struct. Otherwise it is not possible.

2 scenarios:
A. `Alice`  as malicious proposer, proposing a lot of `ECDSA` recovery operations
B. `Bob` (trusted) 
wants to assign `ProverPete` the right to prove. 

`ProverPete` by default cannot be sure what exactly he/she signs (A and B) with the above data (since no "pre and post" state derivation in that  ? Cause if blob is used, the `txList` can be empty, and batch params or it's hash also does not exactly tells "what" transactions he said `"YES"` to ?

(vs. in case we have such the TxList information against a signature check, that works, that eliminates the above problem).

adaki2004: OR the prover also does not care at all about it - because he simply just "charges" for the worst case scenario so that he is covered for sure.

adaki2004: or `_calculateTxsHash` is enough to have a deterministic "signed list of TXNs" (even if that uses blobhases) ?

dantaik:
@adaki2004 I further refactored the code and remove the support of Ether as the fee token -- this is to avoid making the proposing batch function *payable*.

dantaik:
@adaki2004 how about testing this feature in a new PR with this one merged first?

packages/protocol/contracts/layer1/based/LibProverAuth.sol
```diff
@@ -13,66 +13,63 @@ library LibProverAuth {
...
+        auth_ = abi.decode(_proverAuth, (ProverAuth));
 
-        // If `validUntil` is used, make sure it's still valid
-        if (auth.validUntil != 0 && auth.validUntil < block.timestamp) {
-            revert InvalidValidUntil();
-        }
+        require(auth_.prover != address(0), InvalidProver());
+        require(auth_.feeToken != address(0), EtherAsFeeTokenNotSupportedYet());
+        require(auth_.validUntil == 0 || auth_.validUntil >= block.timestamp, InvalidValidUntil());
+        require(auth_.batchId == 0 || auth_.batchId == _batchId, InvalidBatchId());
```
adaki2004, line 50: `batchId` can be "frontrun" (or well, determined with `proposeBatch()` function call ). How can Bob (prover) ensure that this batch (maybe far ahead in the future), gets included with 100% probability with given `batchId` ?

adaki2004:
> @adaki2004 how about testing this feature in a new PR with this one merged first?

Sure, i'm ok. If i find any issues during testing, we just fix that on that branch.

packages/protocol/contracts/layer1/based/LibProverAuth.sol
dantaik: This will work with revert protection -- if the transaction will revert, the transaction will not be included at all.

dantaik: https://github.com/ethereum/EIPs/pull/8267/files

adaki2004 (approved):

PR #19387, opened 2025-05-05
feat(protocol): make `LibProverAuth` more robust

adaki2004:
Some notes: 
- Did not add `batchId` to the data structure (to be signed) because `batchId` is dependent when the actual batch is proposed (technically front-runnable).


packages/protocol/contracts/layer1/based/LibProverAuth.sol
```diff
@@ -5,35 +5,69 @@ import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
...
 /// @notice This library is used to validate the prover authentication.
-/// @dev This libarary's validateProverAuth function is made public to reduce TaikoInbox's code size
+/// @dev This library's validateProverAuth function is made public to reduce TaikoInbox's code size
 /// .
 /// @custom:security-contact security@taiko.xyz
 library LibProverAuth {
     using ECDSA for bytes32;
 
     struct ProverAuth {
+        address feeToken; //Slot1
```
dantaik, line 15: these are not going to stored, so no need to label with slot number.

dantaik (approved):

adaki2004:
@dantaik added some things (fix for `meta.prover` + `isValidSignatureNow` support) you listed as todos in the base PR.

Please if you think it is still good, merge it, then i'll add tests too on that base branch.

dantaik (approved):

PR #19397, opened 2025-05-06
feat(protocol): add unit tests for LibProverAuth (+some fixes)

adaki2004:
tests for `LibProverAuth` (+some fixes). More to be added.
Beside the 2 small fixes (Circular dependnecy in `BatchParams` we have `ProverAuth`, and in `ProverAuth` we have `batchParamsHash`).

Some observation / good to remember:
When a prover signs the `proverAuth`, `batchParams.proposer` shall be signed but when a proposer calls `proposeBatch` it needs to see if he/she shall fill that variable, cause in case of a `wrapper`, the on-chain submitted data has to be `address(0)`. See this comment of mine: 
```
// We have to put together a batchParams.proposer = Alice, to match the has BUT before
// sending it shall be erased from the batchParams.proposer cause of auto-fill!
```


adaki2004:
@dantaik donno why the whole commit history is shown (but you can see the only 4 files diffs) - maybe because the base branch `offchain_prover_market` got deleted. Anyways, diff is small.

Will add more edge case tests in the next hours / day.

dantaik (approved):

dantaik:
Maybe also add one test for TaikoInbox to use a non-empty proverAuth?

adaki2004:
> Maybe also add one test for TaikoInbox to use a non-empty proverAuth?

Added more tests, where more batches are prposed and proved.

dantaik:
must be somethign wrong with this branch: shasta_upgrade2, I just merged main into this one.

PR #19402, opened 2025-05-07
feat(protocol): blacklist non-voting addresses for Taiko token

dantaik:
- [ ] @davidtaikocha once this PR is merged, please upgrade the TaikoToken contract ASAP.


xiaodino (approved):

packages/protocol/contracts/layer1/token/TaikoToken.sol
```diff
@@ -47,9 +45,48 @@ contract TaikoToken is TaikoTokenBase {
...
+    /// for voting.
+    function getPastTotalSupply(uint256 timepoint) public view override returns (uint256) {
+        address[] memory nonVotingAccounts = getNonVotingAccounts();
+        uint256 nonVotingSupply;
+        for (uint256 i; i < nonVotingAccounts.length; ++i) {
+            nonVotingSupply += balanceOf(nonVotingAccounts[i]);
+        }
+        nonVotingSupply += balanceOf(address(0));
+        return super.getPastTotalSupply(timepoint) - nonVotingSupply;
+    }
```
adaki2004, line 67: I think there is a logical issue, and a possible attack vector too here.

So `getPastTotalSupply()` gets a total supply in the **past**. But we deduct the current **nonVotingSupply**.

Scenario:
A snapshot time is 1st of Jan. (**timepoint** in the past)
On the 10th of Jan, Bob (whale) deposited a lot of tokens (from L1 to L2) - so bridge has a lot of tokens now (in the present) which will be excluded from the `totalVotingPower` , which will distort the min veto ratio. (see: https://github.com/aragon/taiko-contracts/blob/ec67bfe50edf74d54d1ccf20896675d6b4199b30/src/OptimisticTokenVotingPlugin.sol#L317 )

When the `OptimisticTokenVoting` plugin queries this information:
```
function totalVotingPower(uint256 _timestamp) public view returns (uint256) {
        return votingToken.getPastTotalSupply(_timestamp);
}
```

It will be disorted, because it has a total supply (in the past) deducting some supply (from the present).

adaki2004: I understand it. But the issue - i described - also exists.

dantaik: Let me think about it...

dantaik: Since the TAIKO token has a fixed supply with no minting or burning mechanism, `votingToken.getPastTotalSupply(_timestamp)` always returns the same value, regardless of `_timestamp`. Therefore, we can consider excluding tokens held directly by the treasury or DAO from the voting supply but `not` tokens in ERC20Vault, as users can manipulate vault balances through deposits and withdrawals.

dantaik: One minor issue: with this modification, the return value of `getPastTotalSupply` will increase if some Taiko tokens are later transferred out of non-voting addresses. This could cause a previously approved proposal to fall just below the approval threshold. However, I believe this edge case is acceptable and the impact should be minimal.

adaki2004: Yes, it is not too trivial to (fully) solve. I like this smaller set of addresses approach.

adaki2004 (approved):

packages/protocol/contracts/layer1/token/TaikoToken.sol
dantaik: Lets keep this PR open for a few days so our Security Council members may have a chance to offer some feedback.

packages/protocol/contracts/layer1/token/TaikoToken.sol
```diff
@@ -11,10 +11,12 @@ import "src/shared/token/TaikoTokenBase.sol";
 /// 0x10dea67478c5F8C5E2D90e5E9B26dBe60c54d800 (token.taiko.eth)
 /// @custom:security-contact security@taiko.xyz
 contract TaikoToken is TaikoTokenBase {
-    address private constant _TAIKO_L1 = 0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a;
-    address private constant _ERC20_VAULT = 0x996282cA11E5DEb6B5D122CC3B9A1FcAAD4415Ab;
+    address public constant TAIKO_FOUNDATION_TREASURY = 0x363e846B91AF677Fb82f709b6c35BD1AaFc6B3Da;
+    // TODO(bernet): what's the address?
+    address public constant TAIKO_DAO_CONTROLLER = 0x0000000000000000000000000000000000000000;
```
dantaik, line 16: https://github.com/taikoxyz/taiko-mono/pull/19423

dantaik:
@davidtaikocha once this PR is merged, please upgrade the TaikoToken contract ASAP.

davidtaikocha (approved):

PR #19416, opened 2025-05-08
refactor(protocol): remove resolver dependency from TaikoAnchor

dantaik:
From Anshu:

> Hey  [@Daniel Wang](https://taikoorg.slack.com/team/U0613U7RD8E)
 Is this required in Pacaya? https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/mainnet/resolvers/RollupResolver.sol
Afaik, everything was shifted to immutable variables, so inbox and anchor contract do not resolve any address.
This is also redundant: https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer2/based/anchor/TaikoAnchor.sol#L15

This PR addresses only the second part of the question.

### DIFF
--- a/packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
+++ b/packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
@@ -87,13 +87,7 @@ abstract contract PacayaAnchor is OntakeAnchor {
         _;
     }
 
-    constructor(
-        address _resolver,
-        address _signalService,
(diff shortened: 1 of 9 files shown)


adaki2004 (approved):

adaki2004:
First part (`RollupResolver`) is used by the verifiers (sgx+zk).

YoGhurt111 (approved):

smtmfft (approved):

PR #19427, opened 2025-05-10
refactor(protocol): refactor PreconfSlasher

dantaik:
…gEOP` & `_validateInvalidEOP` (#19421)


packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -207,33 +200,30 @@ contract PreconfSlasher is IPreconfSlasher, EssentialContract {
...
+                evidence.nextBatchMetadata.proposedAt <= _payload.preconferSlotTimestamp,
+                NextBatchProposedInNextPreconfWindow()
+            );
+        } else {
+            // Check if the block is not the last one in the batch
+            uint256 lastBlockInPreviousBatch = batch.lastBlockId - evidence.batchInfo.blocks.length;
+            require(
+                evidence.preconfedBlockHeader.number > lastBlockInPreviousBatch
+                    && evidence.preconfedBlockHeader.number < batch.lastBlockId,
+                BlockNotInBatch()
```
AnshuJalan, line 224: Why is this ranged check required? Shouldn't the original condition to check if it is (or not) the last block in the batch be more concise: `evidence.preconfedBlockHeader.number != batch.lastBlockId`

dantaik: Is it possible that `evidence.preconfedBlockHeader.number == 100` while the batch contains block 110-120?  We need to make sure preconfedBlockHeader must be covered by the batch.

dantaik:
@AnshuJalan I merged my other PR to this one, feel free to use this PR whatever way you feel comfortable. I think I'm done with commenting, further issues may be found in unit tests.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -97,104 +103,68 @@ contract PreconfSlasher is IPreconfSlasher, EssentialContract {
...
         }
 
-        // Ensure that the anchor block has not been reorged out
-        require(
-            _payload.anchorHash == evidence.batchInfo.anchorBlockHash, PossibleReorgOfAnchorBlock()
-        );
+        // The preconfirmed blockhash must not match the hash of the proposed block for a
+        // preconfirmation violation
+        bytes32 l2BlockHash = evidence.blockhashProofs.l2BlockHeader.hash();
+        require(_payload.blockHash != l2BlockHash, PreconfirmationIsValid());
```
adaki2004, line 136:
```suggestion
        require(_payload.blockHash != evidence.blockhashProofs.l2BlockHeader.hash(), PreconfirmationIsValid());
```

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -97,104 +103,68 @@ contract PreconfSlasher is IPreconfSlasher, EssentialContract {
...
+
+        // Validate that the batch has been verified
+        ITaikoInbox.TransitionState memory transition =
```
dantaik, line 139: Is it a good idea to split this function into 2: one for slashing before the batch is verified, the other is for after the batch is verified...

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol
```diff
@@ -80,7 +80,7 @@ interface IPreconfSlasher is ISlasher {
     // The EVM `slot` containing the blockhash is calculated dynamically based on the block number.
     struct BlockhashProofs {
         // Blockhash value
-        bytes32 value;
+        LibBlockHeader.BlockHeader l2BlockHeader;
```
dantaik, line 83: a better name would be "actualBlockHeader"

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
dantaik: l2BlockHash is used again below.

BTW, I think l2BlockHash can be renamed to `actualBlockHash"

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -247,29 +211,51 @@ contract PreconfSlasher is IPreconfSlasher, EssentialContract {
...
+        view
+        returns (ITaikoInbox.Batch memory batch_)
+    {
+        batch_ = taikoInbox.v4GetBatch(uint64(_batchId));
+        require(keccak256(abi.encode(_metadata)) == batch_.metaHash, InvalidBatchMetadata());
+    }
+
+    function _calcBlockHashSlot(uint256 _blockId) internal pure returns (bytes32) {
+        // The mapping is in the first slot
+        return keccak256(abi.encode(_blockId, bytes32(0)));
```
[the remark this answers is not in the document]
dantaik: Good catch 👍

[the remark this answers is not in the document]
dantaik: https://github.com/taikoxyz/taiko-mono/blob/de7ce0d4934efd9323853d43ffc2ef064166547c/packages/protocol/layout/layer2-contracts.txt#L136

Should be slot 251

dantaik, line 259:
```suggestion
        return keccak256(abi.encode(_blockId, bytes32(uint256(251))));
```

AnshuJalan:
@dantaik Please check my last commit. I think we need to check the anchor slot validity using the stateroot from a verified batch, because right now, it is checking it via the stateroot in the evidence, and there is no guarantee that the evidence' state root is correct – we need verified L2 state

dantaik:
> @dantaik Please check my last commit. I think we need to check the anchor slot validity using the stateroot from a verified batch, because right now, it is checking it via the stateroot in the evidence, and there is no guarantee that the evidence' state root is correct – we need verified L2 state

LGTM. How about merging this PR first then I can take another thorough review :D

PR #19430, opened 2025-05-11
feat(taiko-client): replace `latestSeenProposalCh` with `latestBlockIDSeenInEvent`

### LINKED ISSUE #19420: taiko-driver: `/status` endpoint might return stale `highestUnsafeL2PayloadBlockID` info during reorgs
### Describe the bug

**Description**
The `/status` endpoint introduced in #19358 serves a field called `highestUnsafeL2PayloadBlockID` which is updated in two places:

1. When calling `/preconfBlock`:
https://github.com/taikoxyz/taiko-mono/blob/3d8829cf09d75c79308e837fa5813c537d5c9268/packages/taiko-client/driver/preconf_blocks/api.go#L165-L169

2. When a payload is received from the p2p in `OnUnsafeL2Payload`:
https://github.com/taikoxyz/taiko-mono/blob/3d8829cf09d75c79308e837fa5813c537d5c9268/packages/taiko-client/driver/preconf_blocks/server.go#L308-L312

If an L2 reorg is triggered after taiko-driver reacts to a `BatchProposed` event, then the value returned by this endpoint can be too high because some blocks have been removed. A correct value will be returned as soon as we hit again one of the two code paths above. Until that moment, such the endpoint will return stale information.

**Example**
Suppose sequencer A produces blocks 100..150, then proposes a batch causing a reorg with blocks 100..140. Then the `/status` endpoint will return `highestUnsafeL2PayloadBlockID == 150`, while it should be accurate to return the value of the head L1 origin, which is 140.

**Desired behaviour**
Upon detecting an L2 reorg, taiko-driver should return `highestUnsafeL2PayloadBlockID` equal to `head_l1_origin` instead of the previously highest block sent.

### Steps to reproduce

_No response_

davidtaikocha:


packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go
```diff
@@ -98,15 +99,11 @@ func (i *BlocksInserterPacaya) InsertBlocks(
 			len(meta.GetBlobHashes()) != 0,
 			true,
 		)
-		parent          *types.Header
-		lastPayloadData *engine.ExecutableData
+		lastestSeenProposal = &encoding.LastSeenProposal{TaikoProposalMetaData: metadata}
```
cyberhorsey, line 102: typo "lastestSeen" should be "latestSeen"

davidtaikocha: fixed

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go
```diff
@@ -236,6 +242,10 @@ func (i *BlocksInserterPacaya) InsertBlocks(
 		metrics.DriverL2HeadHeightGauge.Set(float64(lastPayloadData.Number))
 	}
 
+	// Mark the last seen proposal as not preconfirmed and send it to the channel.
+	latestSeenProposal.PreconfChainReorged = true
```
YoGhurt111, line 246: I don't get the logic why it is a preconf reorg when handled a normal batch.

davidtaikocha: If the code hits this, which means the proposed blocks are not preconfed, because of the preconf checking above.

PR #19435, opened 2025-05-11
refactor(taiko-client): polish #19361

davidtaikocha:
No logic change, ready to review and merge now @cyberhorsey @YoGhurt111


packages/taiko-client/driver/preconf_blocks/util.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if headL1Origin != nil && uint64(msg.ExecutionPayload.BlockNumber) <= headL1Origin.BlockID.Uint64() {
+		return nil, fmt.Errorf(
+			"preconfirmation block ID (%d) is less than or equal to the current head L1 origin block ID (%d)",
+			msg.ExecutionPayload.BlockNumber,
+			headL1Origin.BlockID,
+		)
+	}
+
+	return headL1Origin, nil
```
YoGhurt111, line 67: If err is `ethereum.NotFound`, here would return nil, nil.
It makes no sense.

davidtaikocha: I think its ok, in our design, `headL1Origin` might be `nil`, but added some more comments for this method.

YoGhurt111 (approved):

cyberhorsey (approved):

PR #19442, opened 2025-05-13
feat(protocol): remove chain/bridge watch dogs from Bridge

dantaik:
The current chain watchdog is the guardian prover. After pacaya, guardian prover will be removed.


packages/protocol/contracts/shared/libs/LibStrings.sol
```diff
@@ -8,11 +8,9 @@ library LibStrings {
     bytes32 internal constant B_SGX_GETH_AUTOMATA = bytes32("sgx_geth_automata");
     bytes32 internal constant B_BOND_TOKEN = bytes32("bond_token");
     bytes32 internal constant B_BRIDGE = bytes32("bridge");
-    bytes32 internal constant B_BRIDGE_WATCHDOG = bytes32("bridge_watchdog");
```
davidtaikocha, line 11: Shall we also remove the log here? https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/script/layer2/PostGenesisQuery.s.sol#L55

dantaik: I think we can keep it there.

davidtaikocha (approved):

adaki2004 (approved):

PR #19445, opened 2025-05-13
refactor(protocol): try to refactor PreconfRouter2

dantaik:


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -120,8 +142,20 @@ contract PreconfRouter2 is IPreconfRouter2, EssentialContract {
         // Ensure that the associated operator is active and opted into the preconf slasher
         IRegistry.OperatorData memory operatorData =
             urc.getOperatorData(_lookaheadSlot.registrationRoot);
-        require(operatorData.slashedAt == 0, OperatorIsSlashed());
-        require(operatorData.unregisteredAt == 0, OperatorIsUnregistered());
+
+        require(
```
dantaik, line 146: I changed these checks to match the logic in LookaheadStore -- https://github.com/taikoxyz/taiko-mono/blob/de7ce0d4934efd9323853d43ffc2ef064166547c/packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol#L231

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -55,43 +67,53 @@ contract PreconfRouter2 is IPreconfRouter2, EssentialContract {
...
 
-        // Try fetching the lookahead for the next epoch.
-        // This call fails if the lookahead for next epoch is not posted, thus requiring the first
-        // preconfer to post the next epoch's lookahead before proposing a batch in the current
-        // epoch.
-        lookaheadStore.getLookaheadHash(epochTimestamp + LibPreconfConstants.SECONDS_IN_EPOCH);
+        {
+            bytes26 currLookaheadHash = lookaheadStore.getLookaheadHash(epochTimestamp);
+            if (currLookaheadHash == 0) {
+                // TODO: post the lookahead for current epoch here
```
dantaik, line 90: @AnshuJalan this is a point for discussion.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -55,43 +67,53 @@ contract PreconfRouter2 is IPreconfRouter2, EssentialContract {
...
+
+        {
+            // Try fetching the lookahead for the next epoch.
+            // This call fails if the lookahead for next epoch is not posted, thus requiring the
+            // first preconfer to post the next epoch's lookahead before proposing a batch in the
+            // current epoch.
+            uint256 nextEpochTimestamp = epochTimestamp + LibPreconfConstants.SECONDS_IN_EPOCH;
+            bytes26 nextLookaheadHash = lookaheadStore.getLookaheadHash(nextEpochTimestamp);
+            if (nextLookaheadHash == 0) {
+                // TODO: post the lookahead for next epoch here
```
dantaik, line 104: Another point for discussion.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
adaki2004: Not exactly matching in my opinion. 

` operatorData.registeredAt <= block.timestamp` is redundant, because it (`registeredAt`) will not be a "future" time.

Also `operatorData.unregisteredAt > block.timestamp` and `operatorData.slashedAt > block.timestamp` will never be true.

So operator may `unregister()` or may be `slashed()` in the same block, but best case it will be `==` but not possible that the `slashedAt` or `unRegisteredAt` are bigger than current time (`block.timestamp`).

I

dantaik: So the logics in LookaheadStore is incorrect?!

dantaik: Are you sure registeredAt and unregisteredAt cannot be future timestamp? I thought they meant some operators have been scheduled to register/unregister in a future timestamp.

adaki2004: > So the logics in LookaheadStore is incorrect?!

No, i think it is correct there. 

Also not saying, it is "not necessarily" incorrect here, but the parts i highlighted are redundant.

Cause if you look it here:
https://github.com/eth-fabric/urc/blob/5f18225c4d027b49084605129b7666a37ea13412/src/Registry.sol#L110
`unregistereAt` is the actual (now) block timestamp. So it is not set in the future.

To answering your quesiton, even if unregistered, it is subject to slashing still, because there is an unregistration delay, which allows slashing. So unregistering, does not mean, it cannot be slashed.

See: https://github.com/eth-fabric/urc/blob/5f18225c4d027b49084605129b7666a37ea13412/src/Registry.sol#L211

So that is why i'm saying, those extra check are redundant:
1. `operatorData.registeredAt <= block.timestamp`, because if `operatorData.registeredAt != 0`, it is sure it will be "past" OR best case equal to block.timestamp anyways. (This is also not set in the future).
2. `operatorData.unregisteredAt > block.timestamp` -> This cannot be true since unregistering happens best case at `block.timestamp`
2. `operatorData.slashedAt > block.timestamp` -> This cannot be true since slashing happens best case at `block.timestamp`

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -55,43 +67,53 @@ contract PreconfRouter2 is IPreconfRouter2, EssentialContract {
...
+            } else {
+                _validateLookahead(nextEpochTimestamp, nextLookahead, nextLookaheadHash);
+            }
+        }
 
-        if (_isEmptyLookahead(epochTimestamp, currentLookaheadHash)) {
+        if (currLookahead.length == 0) {
             // The current lookahead is empty, so we use a whitelisted preconfer
+            require(prevLookahead.length == 0, InvalidPreviousLookahead()); // Do not need it
```
AnshuJalan, line 112: Should be removed, right?

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
AnshuJalan: Agreeing with Dani's comment

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -180,50 +198,33 @@ contract PreconfRouter2 is IPreconfRouter2, EssentialContract {
...
-                    block.timestamp > previousLookaheadSlot.slotTimestamp
-                        && block.timestamp <= _lookaheadSlot.slotTimestamp,
-                    InvalidLookaheadTimestamp()
-                );
+            uint256 prevEpochTimestamp = _epochTimestamp - LibPreconfConstants.SECONDS_IN_EPOCH;
+            bytes26 prevLookaheadHash = lookaheadStore.getLookaheadHash(prevEpochTimestamp);
+
+            if (prevLookaheadHash != 0 && _prevLookahead.length != 0) {
+                _validateLookahead(prevEpochTimestamp, _prevLookahead, prevLookaheadHash);
+                prevSlotTimestamp = _prevLookahead[_prevLookahead.length - 1].slotTimestamp;
```
AnshuJalan, line 206:
```solidity
if (_prevLookahead.length != 0) {
   uint256 prevEpochTimestamp = _epochTimestamp - LibPreconfConstants.SECONDS_IN_EPOCH;
   bytes26 prevLookaheadHash = lookaheadStore.getLookaheadHash(prevEpochTimestamp);
   
   _validateLookahead(prevEpochTimestamp, _prevLookahead, prevLookaheadHash);
    prevSlotTimestamp = _prevLookahead[_prevLookahead.length - 1].slotTimestamp;
}
```

The call `lookaheadStore.getLookaheadHash(prevEpochTimestamp);` fails if the previous lookahead was not posted, so we should only trigger the call if the sender has passed a non-empty lookahead.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
AnshuJalan: I intended to keep the posting of next epoch's lookahead separate from the Preconf Router since we are treating the Lookahead Store to be as independent of an entity as possible.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
AnshuJalan: Actually, it is not safe after EIP7917 either, because the preconfer can post a faulty lookahead + propose a block based on the lookahead in the same transaction, which is something we do not intend. We must give some time (at least slot) to the other actors to find the incorrectness.

AnshuJalan: CC: @linoscope

linoscope: Yeah, posting current epoch lookahead has issues as @AnshuJalan mentions. As long as we have the "require first preconfer in epoch post lookahead of next epoch" thing having no lookahead should be an edge case, so think it is fine to use fallback in that case.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
AnshuJalan: False alarm.

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
AnshuJalan: Me and Lin had a convo and decided it is indeed better to allow posting next epoch's lookahead via the router itself.
PR: https://github.com/taikoxyz/taiko-mono/pull/19511

dantaik:
@AnshuJalan I closed this PR, if there is anything in this PR you think is valuable, please copy to your PR. Thanks.

AnshuJalan:
@dantaik I think this PR should be merged after a couple of minor changes. Some of the refactoring makes it cleaner. Should I reopen it and makes the changes?

dantaik:
> @dantaik I think this PR should be merged after a couple of minor changes. Some of the refactoring makes it cleaner. Should I reopen it and makes the changes?

Feel free :D

dantaik:
replaced by https://github.com/taikoxyz/taiko-mono/pull/19520

PR #19454, opened 2025-05-14
chore(protocol): deploy mainnet ontake L1 with Pacaya height check

YoGhurt111:


packages/protocol/contracts/layer1/based/TaikoData.sol
```diff
@@ -7,7 +7,7 @@ import "src/shared/data/LibSharedData.sol";
 /// @notice This library defines various data structures used in the Taiko protocol.
 /// @custom:security-contact security@taiko.xyz
 library TaikoData {
-    uint256 public constant PACAYA_FORK_HEIGHT = 1_299_888;
```
dantaik, line 10: I notice an issue with TaikoInbox, TaikoInbox should not use PACAYA_FORK_HEIGHT as a constant at all, it should read the value from the config.

dantaik (approved):

davidtaikocha (approved):

PR #19459, opened 2025-05-14
feat(protocol): update old fork `oldFork` for mainnet pacaya deployment script

davidtaikocha:
plz double check the code: https://etherscan.io/address/0xaa64d5a3a26d1e76acaf6e22c199d02d58076a01#code @YoGhurt111

### DIFF
--- a/packages/protocol/script/layer1/mainnet/DeployMainnetPacayaL1.s.sol
+++ b/packages/protocol/script/layer1/mainnet/DeployMainnetPacayaL1.s.sol
@@ -116,7 +116,7 @@ contract DeployMainnetPacayaL1 is DeployCapability {
                     uint8(inclusionWindow), uint64(inclusionFeeInGwei), taikoInbox, address(1)
                 )
             ),
-            data: abi.encodeCall(ForcedInclusionStore.init, (contractOwner)),
+            data: abi.encodeCall(ForcedInclusionStore.init, (address(0))),
             registerTo: rollupResolver
         });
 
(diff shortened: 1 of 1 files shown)


packages/protocol/script/layer1/mainnet/DeployMainnetPacayaL1.s.sol
```diff
@@ -165,6 +164,7 @@ contract DeployMainnetPacayaL1 is DeployCapability {
         Ownable2StepUpgradeable(sharedResolver).transferOwnership(contractOwner);
         Ownable2StepUpgradeable(rollupResolver).transferOwnership(contractOwner);
         Ownable2StepUpgradeable(proofVerifier).transferOwnership(contractOwner);
+        Ownable2StepUpgradeable(store).transferOwnership(contractOwner);
```
dantaik, line 167: this will only transfer the ownership, the the new owner has to accept the ownership.

dantaik: I think the few lines starting at 170 should be commented out, also please also log which contracts' ownership should be accepted by admin.taiko.eth.

YoGhurt111 (approved):

packages/protocol/script/layer1/mainnet/DeployMainnetPacayaL1.s.sol
davidtaikocha: Yes, those are only for reminding.

PR #19466, opened 2025-05-15
fix(taiko-client): check pointer before logging

cyberhorsey:
can be nil during resync leading to panic


davidtaikocha (approved):

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -815,10 +815,16 @@ func (s *PreconfBlockAPIServer) ImportMissingAncientsFromCache(
 	currentPayload *eth.ExecutionPayload,
 	headL1Origin *rawdb.L1Origin,
 ) error {
+	var headL1OriginBlockId uint64
+
```
davidtaikocha, line 819:
```suggestion
```

YoGhurt111 (approved):

PR #19469, opened 2025-05-16
fix(protocol): improve isContract() by checking EIP7702 delegation

dantaik:
https://x.com/apoorveth/status/1923241437593821230?s=46


packages/protocol/contracts/shared/libs/LibAddress.sol
```diff
@@ -76,4 +77,23 @@ library LibAddress {
             result_ = abi.decode(data, (bool));
         }
     }
+
+    function isContract(address _addr) internal view returns (bool) {
+        return Address.isContract(_addr) // code size > 0
+            && delegationOf(_addr) == address(0); // not an EOA with 7702 delegation
+    }
```
adaki2004, line 84: Trying to wrap my head around this, but i dont realy get it. 

So
condition `A`: `codesize` needs to be bigger than 0
condiiton `B`: `_addr` shall **not** be delegate

How `A` AND `B` signals it is a contract ?

Maybe i'm wrong but i think this is the logic (in my head):

`A`: stays same
`B`:  `_addr` **shall be** delegate

And then, `A` OR `B` to signal it is a contract.

dantaik: A: the address has code
B: the address is not an EOA who has delegate its code to a contract.

adaki2004: > B: the address is not an EOA who has delegate its code to a contract.

How this: "the address , to which `_addr `delegating to is `address(0)`"  means it is not EOA ? I think it is the opposite.

Delegation to `address(0)` means NOT delegating, hence it is an EOA.

davidtaikocha (approved):

packages/protocol/contracts/shared/libs/LibAddress.sol
dantaik: Dani, could you help adding a few tests?

adaki2004: Will do (most prob monday cause have a lot of private things on the weekend), but this is my source of info:
https://x.com/optimizoor/status/1923279661905133980?s=46

Delegating to zero address equivalent to no delegation.

adaki2004: Leaving it to clarify: The intention of this is to NOT classify for` isContract() == true`, when an EOA delegates to a contract.

dantaik:
Screenshot 2025-05-19 at 5 31 09 PM

Approved by ChatGPT:D

PR #19470, opened 2025-05-16
feat(protocol): prototype a simplified version of SignalBoost without ToB

dantaik:
https://ethresear.ch/t/signal-boost-l1-interop-plugin-for-rollups/22354

Code are not tested yet, so will not merge anytime soon.

## TODO

- [ ] test it
- [ ] bring Chainlink data feed to L2 using this approach


packages/protocol/contracts/layer1/signal-boost/IStateQuerier.sol
```diff
@@ -0,0 +1,19 @@
...
+import "contracts/shared/signal-boost/IStateQuery.sol";
+
+/// @title IStateQuerier
+/// @notice Contract for querying the state of multiple contracts.
+/// @custom:security-contact security@taiko.xyz
+interface IStateQuerier {
+    event QueryResult(IStateQuery.Query query, IStateQuery.QueryResult result);
+
+    /// @notice Queries the state of multiple contracts.
+    /// @param _queries An arry of queries, each containing a target contract and a payload.
```
dantaik, line 13:
```suggestion
    /// @param _queries An array of queries, each containing a target contract and a payload.
```

packages/protocol/contracts/layer2/signal-boost/IStateConsumer.sol
```diff
@@ -0,0 +1,18 @@
...
+import "contracts/shared/signal-boost/IStateQuery.sol";
+
+/// @title IStateConsumer
+/// @notice Interface for state consumers.
+/// @custom:security-contact security@taiko.xyz
+interface IStateConsumer {
+    /// @notice Consumes the state of a contract.
+    /// @param _request The query request.
+    /// @param _response The query response.
+    function consume(
```
JasonVranek, line 13: One thing I was trying was having the Sequencer batch prove all of the `QueryResult` values on the L2 against the signal. Essentially post `N` Merkle proofs and then store the results to a mapping like this:

```solidity
mapping(bytes32 requestHash => bytes output) public requestOutputs;
//...
// function accepts an array of requests and responses, for each:
bytes32 requestHash = keccak256(abi.encode(_requests[i]));
requestOutputs[requestHash] = _responses[i];
```
If the sequencer fronts these costs, then downstream L2 users only need to query the `requestOutputs` mapping with their known `requestHash` and decode the response to the type they expect - no need for them to learn the Merkle proof which could require additional communication with the sequencer to fetch. This should really be a DevEx / UX improvement. 

Also, implementation-wise, if you go this way you can simplify `hashQueriesToSignal` to just do a hash-chain rather than merkle tree for efficiency. 

[See here for the example code](https://github.com/eth-fabric/signal-boost/blob/d8f74f8e1c30ca8f48581c09ed3378d5db5d41e2/src/SignalProver.sol#L21)

packages/protocol/contracts/layer2/signal-boost/StateVerifier.sol
```diff
@@ -0,0 +1,61 @@
...
+    )
+        external
+        nonReentrant
+    {
+        uint256 n = _queries.length;
+        require(n == _results.length, InvalidParamSizes());
+        require(n == _consumers.length, InvalidParamSizes());
+
+        bytes32 signal =
+            LibStateQuery.hashQueriesToSignal(l1ChainId, block.timestamp, _queries, _results);
```
davidtaikocha, line 52: I still see we are supporting `blockParams.timeShift` in protocol, seems if we want to make this work, we need to disable `blockParams.timeShift`?

PR #19471, opened 2025-05-16
feat(protocol): propose batch with expected last block ID

merklefruit:
Added functionality to `TaikoWrapper` to allow sequencers to specify the expected last block ID in a proposal.

This optional feature aims to significantly increases the reliability of parallel batch proposals without risking to 
cause unintentional L2 reorgs.

### DIFF
--- a/packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
+++ b/packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.24;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "src/layer1/based/IProposeBatch.sol";
+import "src/layer1/based/ITaikoInbox.sol";
 import "./ForcedInclusionStore.sol";
 
 /// @title TaikoWrapper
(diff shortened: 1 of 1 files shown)


dantaik:
Please review https://github.com/taikoxyz/taiko-mono/pull/19472 and let me know if it will work for you?

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
```diff
@@ -100,6 +102,29 @@ contract TaikoWrapper is EssentialContract, IProposeBatch {
...
+    )
+        external
+        onlyFromOptional(preconfRouter)
+        nonReentrant
+        returns (ITaikoInbox.BatchInfo memory, ITaikoInbox.BatchMetadata memory)
+    {
+        (ITaikoInbox.BatchInfo memory info_, ITaikoInbox.BatchMetadata memory meta_) =
+            this.v4ProposeBatch(_params, _txList, bytes(""));
+
+        uint96 lastBlockId = ITaikoInbox(address(inbox)).v4GetBatch(meta_.batchId).lastBlockId;
```
dantaik, line 118: BTW, lastBlockId is also available in the BatchInfo.

merklefruit:
> Please review #19472 and let me know if it will work for you?

Actually @davidtaikocha pointed out that this functionality can be replicated by specifying `parentMetaHash` here: https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/based/TaikoInbox.sol#L749

That's more elegant and requires no changes so I'll close this

PR #19475, opened 2025-05-18
feat(protocol):  enforce tail-only anchoring

dantaik:
Enable node/client to enforce per-block anchoring rules. The supported options are:

1.	Uniform anchoring: All blocks share the same non-zero anchor ID and state root.
2.	Tail-only anchoring: Only the last block in the batch has a non-zero anchor ID and state root.
3.	Head-only anchoring: Only the first block in the batch has a non-zero anchor ID and state root.
4.	Head and tail anchoring: Both the first and last blocks must have non-zero anchor ID and state root; intermediate blocks may optionally have non-zero values.

This PR suggests to enforce option 2 (tail-only anchoring) in node/client.

Previously, the anchor ID must be selected before the first block of a batch is proposed, meaning it has to be known upfront. With this PR, the anchor ID can instead be selected just before building the last (tail) block in the batch, allowing it to reference a much more recent block.


davidtaikocha (approved):

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
```diff
@@ -62,10 +68,22 @@ abstract contract ShastaAnchor is PacayaAnchor {
...
+            // For the final block in the batch, _anchorStateRoot must be non-zero.
+            require(_anchorStateRoot != 0, ZeroAnchorStateRoot());
+            _syncChainData(_anchorBlockId, _anchorStateRoot);
+        }
+
+        if (_signalSlots.length != 0) {
+            signalService.receiveSignals(_signalSlots);
+        }
 
         // We need to add one SSTORE from non-zero to non-zero (5000), one addition (3), and one
```
davidtaikocha, line 88: Not related to this PR, shall we create an `init2` method in  `TaikoAnchor` to add sth like `lastAnchorGasUsed = 1`, to ensure the `lastAnchorGasUsed` below is non-zero from the beginning?

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
```diff
@@ -30,17 +33,22 @@ abstract contract ShastaAnchor is PacayaAnchor {
...
-    /// @param _parentGasUsed The gas used in the parent block.
-    /// @param _baseFeeConfig The base fee configuration.
-    /// @param _signalSlots The signal slots to mark as received.
+    /// @dev The gas limit for this transaction is required to be set to 1,000,000 gas.
+    /// @dev Although this function can be invoked freely due to the public availability of the
+    /// golden touch private key,
+    /// the Taiko node ensures that the first transaction of each block is always this anchor
+    /// transaction. Any subsequent calls will be reverted with L2_PUBLIC_INPUT_HASH_MISMATCH.
+    /// @param _anchorBlockId This is the `anchorBlockId` value in the metadata of this block. This
+    /// value is required to be 0 for all blocks except the last one in the batch.
```
linoscope, line 44: Is my understanding correct that the `anchorBlockId` set in `TaikoInbox.v4ProposeBatch` will be the anchor ID for only the final L2 block in the batch, and not for all L2 blocks in the batch (as it is right now)?

linoscope: I think this makes sense and is needed for the same-slot signals to actually work (I am guessing this is the motivation).

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
```diff
@@ -62,10 +68,22 @@ abstract contract ShastaAnchor is PacayaAnchor {
...
+            require(_signalSlots.length == 0, NonEmptySignalSlots());
+        } else {
+            // For the final block in the batch, _anchorStateRoot must be non-zero.
+            require(_anchorStateRoot != 0, ZeroAnchorStateRoot());
+            _syncChainData(_anchorBlockId, _anchorStateRoot);
+        }
+
+        if (_signalSlots.length != 0) {
+            signalService.receiveSignals(_signalSlots);
+        }
```
linoscope, line 86: nit: We only need to sync signal slots in final batch.

```suggestion
            _syncChainData(_anchorBlockId, _anchorStateRoot);

            if (_signalSlots.length != 0) {
                signalService.receiveSignals(_signalSlots);
            }
        }
```

linoscope (approved):

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
dantaik: > Is my understanding correct that the `anchorBlockId` set in `TaikoInbox.v4ProposeBatch` will be the anchor ID for only the final L2 block in the batch, and not for all L2 blocks in the batch (as it is right now)?

In Pacaya, all blocks in a batch share the same anchor id and anchor hash (approach#1), but with this change, only the last block will have non-zero anchor id and hash (approach#2)


- approach #1: all blocks share the same non-zero anchor id and hash
- approach #2: only the last block in a batch has a non-zero anchor id and hash
- approach #3: only the first block in a batch has a non-zero anchor id and hash
- approach #4: the first & last block must have a non-zeor anchor id and hash, intermediate blocks can have non-zero values as well.

Maybe we should try approach#4 for flexibility...

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
dantaik: No need, as this feature is not expected to be used right after the fork.

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
davidtaikocha: imo current changes are not enough for approach#4 ? If we want to do so, we need to update the rules in Inbox? Personally I prefer approach #2 for a better preconfer experience, but will also lead to a worse bridging delay...

linoscope: Agree that approach 4 would be ideal, it would let the preconfer dynamically increase the anchor id while preconfing and constructing batches, which would lead to shorter bridge delay. But that would require changes in inbox as David mentions, and not sure how much cost it adds.

linoscope: I wasn't thinking of the expiration issue. Yeah, if we are thinking of that, 2 will be better.

Was thinking of providing good UX for users. Suppose a preconfer is assigned to slots `10` to `20`, and suppose they only submit batch at `20`. And let's say the "anchor lag" (they set `latest_slot - anchor_lag` as anchor ID) is 4. Then:
- Option 1: All preconfed blocks will have anchor ID of `10-4=6)`
- Option 2: Final preconfed block will have anchor ID of `20-4=16`. The rest of the blocks will have `9-4=5` assuming previous preconfer submitted batch at `9`.
- Option 3: First block will have anchor ID of `10-4=6`. Final block will have `20-4=16`.
- Option 4: First block will have anchor ID of `10-4=6`, second will have `11-4=7`, third `12-4=8`, ...

So with option 4, the anchor ID will be pretty up-to-date during the preconfing window, allowing for faster L1->L2 messaging (e.g., deposits) UX.

linoscope: But nothing forces the preconfer to keep up-to-date anchor ID with option 4 actually. Well, I guess if there are L2 txs that depend on deposits, they would want to pull in up-to-date anchor ID to include such txs. But probably not a strong incentive.

Also, not sure how much complexity (and added cost in inbox) is required with option 4. It may not be worth it in the end if it is too much.

dantaik: I think we can try tail-only anchoring option first --only the last block in the batch) has a non-zero anchor block id.

davidtaikocha (approved):

packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
```diff
@@ -62,10 +64,20 @@ abstract contract ShastaAnchor is PacayaAnchor {
...
-        _syncChainData(_anchorBlockId, _anchorStateRoot);
         _updateParentHashAndTimestamp(parentId);
 
-        signalService.receiveSignals(_signalSlots);
+        if (_anchorBlockId == 0) {
+            // This block must not be the last block in the batch.
+            require(_anchorStateRoot == 0, NonZeroAnchorStateRoot());
+        } else {
+            // This block must be the last block in the batch.
+            require(_anchorStateRoot != 0, ZeroAnchorStateRoot());
```
linoscope, line 74: It makes sense to require the final block to be non-zero anchor, as we don't want preconfers to not sync with L1 at all, as that would lead to bad bridging UX. But do we need to require that all non-final blocks have zero anchor?

It may be that the _optimal_ strategy for preconfers is to just set the final block as non-zero. But a preconfer who wants to get orderflow from recently deposited (or even same slot) accounts may be willing to set the intermediary anchor to non-zero.

dantaik: I’ve thought about this — such a change would require us to pass a list of anchor block IDs along with their corresponding values, and verify each one on L1. This would significantly increase calldata size (2 bytes32 per block) and gas cost.

linoscope: >  such a change would require us to pass a list of anchor block IDs along with their corresponding values, and verify each one on L1. This would significantly increase calldata size (2 bytes32 per block) and gas cost.

Ah, good point. I think it is fine as long as we make it optional and don't add overhead to the "single anchor ID for batch" case. But it will require some more consideration, so agree to start simple as done in this PR.

linoscope:
@dantaik One last question - Is my understanding correct that the enforcement of the anchor ID being only in the last block will come from the client/prover side node implementation?

PR #19479, opened 2025-05-18
chore(protocol): minor changes to PreconfSlasher

dantaik:
This PR tries to fix a bug where the block in question is the last block in a batch. In such case, we cannot use L2 anchor's storage-based proof verification to verify the block hash as this last block's hash will only be written by the first block in the next batch.

### DIFF
--- a/packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
+++ b/packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
@@ -125,12 +125,9 @@ contract PreconfSlasher is IPreconfSlasher, EssentialContract {
         if (batchInfo.proposer != _committer) {
             // If the beacon block root is not available, it means that the preconfirmed block
             // was reorged out due to an L1 reorg.
-            require(
-                LibPreconfUtils.getBeaconBlockRootAt(_payload.preconferSlotTimestamp) == 0,
-                ReorgedPreconf()
-            );
-
-            return getSlashAmount().reorgedPreconf;
+            if (LibPreconfUtils.getBeaconBlockRootAt(_payload.preconferSlotTimestamp) == 0) {
+                return getSlashAmount().reorgedPreconf;
+            }
         }
 
         LibBlockHeader.BlockHeader memory actual = evidence.parentBlockhashProofs.actualBlockHeader;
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -102,68 +102,77 @@ contract PreconfSlasher is IPreconfSlasher, EssentialContract {
...
+                ReorgedPreconf()
+            );
+
+            return getSlashAmount().reorgedPreconf;
         }
 
+        LibBlockHeader.BlockHeader memory actual = evidence.parentBlockhashProofs.actualBlockHeader;
+        LibBlockHeader.BlockHeader memory verified = evidence.verifiedBlockHeader;
+
+        require(actual.number == preconfed.number, InvalidActualBlockHeader());
```
dantaik, line 136: additional checks here

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasher.sol
```diff
@@ -102,69 +102,81 @@ contract PreconfSlasher is IPreconfSlasher, EssentialContract {
...
-        // at the same height as the preconfirmed block.
-        LibTrieProof.verifyMerkleProof(
-            evidence.verifiedBlockHeader.stateRoot,
-            taikoAnchor,
-            _calcBlockHashSlot(blockId),
-            actualBlockHash,
-            evidence.blockhashProofs.accountProof,
-            evidence.blockhashProofs.storageProof
-        );
+        if (verified.number == blockId) {
```
dantaik, line 166: if  the block in question is the last block in a batch, even if this batch is verified, it block's hash is not written into storage by anchor tx -- anchor only write the parent block hash, not the current block hash into storage.

AnshuJalan:
@dantaik For this issue, will it not be simpler to then just get verified stateroot from the next batch instead of the current batch? It will simplify the logic, and I believe it is fine to use the next batch since the slashing window is 1 day

dantaik:
> @dantaik For this issue, will it not be simpler to then just get verified stateroot from the next batch instead of the current batch? It will simplify the logic, and I believe it is fine to use the next batch since the slashing window is 1 day

Theoretically yes. But as of know we have `  require(transition.blockHash == verified.hash(), InvalidVerifiedBlockHeader());` which means the verified batch is indeed the batch that contains the block in question. If we have to supply another batch's data, will it complicate the code?

I'm ok to this option if you want to prototype it.

dantaik:
@AnshuJalan please take another look at this PR.

AnshuJalan:
Looks good to me

AnshuJalan (approved):

adaki2004 (approved):

PR #19485, opened 2025-05-19
feat(protocol): improve DelegateOwner

dantaik:
- Remove permission check for acceptOwnership
- Remove Admin and use immutables


adaki2004:
@dantaik

Deployed the `DelegateOwner` and `TestDelegateOwned` (this latter one is for "real" tests, during the DAO learning-in phase) contracts on taiko. (Please see deployment logs).

As a test, the **DAO** (on L1) can operate / control the `TestDelegateOwned` contracts (with the help of the `DelegateOwner`). 

For example the **DAO** now can initiate a:
- `setEtherAmount()` and `withdraw()` functions as it wishes.

Funded the contract with 0.001 ETH so the `withdraw() `of that chunk would go thru successfully as a test.

packages/protocol/contracts/layer2/DelegateOwner.sol
```diff
@@ -45,54 +38,29 @@ contract DelegateOwner is EssentialContract, IMessageInvocable {
...
-
-        remoteChainId = _remoteChainId;
-        remoteOwner = _remoteOwner;
-        admin = _admin;
     }
 
     /// @inheritdoc IMessageInvocable
-    function onMessageInvocation(bytes calldata _data) external payable onlyAdminOrRemoteOwner {
+    function onMessageInvocation(bytes calldata _data) external payable {
+        require(msg.sender == l1Bridge, DO_INVALID_SENDER());
```
davidtaikocha, line 59: imo here should be `msg.sender == l2Bridge`?

packages/protocol/contracts/layer2/DelegateOwner.sol
```diff
@@ -13,21 +13,14 @@ import "../shared/bridge/IBridge.sol";
...
 
     /// @notice The next transaction ID.
-    uint64 public nextTxId; // slot 2
+    uint64 public nextTxId; // slot 1
 
-    /// @notice The real owner on L1, supposedly the DAO.
-    address public remoteOwner;
-
-    uint256[48] private __gap;
+    uint256[49] private __gap;
```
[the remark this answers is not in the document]
dantaik: How about we just keep the previously deployed one for Hekla without upgrading it.

packages/protocol/contracts/layer2/DelegateOwner.sol
dantaik: This is a bug!
We should also have l2Bridge immutable @adaki2004

packages/protocol/contracts/layer2/DelegateOwner.sol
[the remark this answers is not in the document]
adaki2004: Should be fixed (and also deployed again, updated the deployment data too).

adaki2004:
@davidtaikocha @dantaik 
Still having issues with contract verification (as last time with the DAO contracts) with regards to the `ERC1967Proxy` contract on Taiko. (Implementation is verified).

Error: Failed to get standard json input

Context:
- cannot resolve file at "...." while the file is clearly there...

As you can see in the screenshot the file is clearly there:
image

I tried with at least 30-40 variations. from the node_modules folder, from the out folder (as in the remappings) all result in that error message. Maybe David could give it a try, idk.

packages/protocol/contracts/layer2/DelegateOwner.sol
```diff
@@ -13,20 +13,16 @@ import "../shared/bridge/IBridge.sol";
...
-    address public admin;
+    // Was remoteChainId + admin before being immutable
+    uint256 public __deprecated;
 
     /// @notice The next transaction ID.
     uint64 public nextTxId; // slot 2
 
-    /// @notice The real owner on L1, supposedly the DAO.
-    address public remoteOwner;
-
```
davidtaikocha, line 29: Shall we also put a placeholder here?

dantaik: It's OK not to.

YoGhurt111 (approved):

davidtaikocha (approved):

PR #19488, opened 2025-05-19
feat(protocol): add optional lastBlockId check in PreconfRouter

merklefruit:
Supersedes #19471, pointing to the right base branch used in current preconf team devnets.

The difference with the earlier approach is to use `PreconfRouter` which is the real entrypoint 
that requires to be called by the operator EOA.

The approach using parentMetaHash ended up not working because it requires knowing the `block.timestamp` of
when the transaction will land in a block, which makes it unreliable.

### DIFF
--- a/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
+++ b/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
@@ -11,6 +11,8 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
     address public immutable proposeBlockEntrypoint;
     address public immutable preconfWhitelist;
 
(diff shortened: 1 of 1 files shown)


dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -47,4 +49,21 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
         // Verify that the sender had set itself as the proposer
         require(meta_.proposer == msg.sender, ProposerIsNotTheSender());
     }
+
+    function proposeBatchWithExpectedLastBlockId(
```
dantaik, line 53:
```suggestion
    function proposeBatch(
```

PR #19498, opened 2025-05-21
chore(protocol): introduce Controller and build DAO proposal-0002

dantaik:
This PR refactors `TaikoDAOController` and `DelegateController` to inherit from a common base contract `(Controller),` enabling both to hold Ether and ERC20 tokens.

It also introduces a standardized way to construct cross-chain actions for Taiko DAO proposals. These actions can now be verified by re-running tests and performing byte-for-byte comparisons with the proposal outputs. This eliminates the need for Security Council members to manually decode binary action data, significantly simplifying the verification process.

The existing `DelegateOwner` contract remains unchanged, as it is still in use by Hekla. For Taiko Alethia mainnet, DelegateController will serve as the owner of all contracts and the custodian of all DAO-held assets.

Once the new contracts are reviewed, I'll handle the TODOs in the tests.


packages/protocol/contracts/shared/goverance/Controller.sol
```diff
@@ -0,0 +1,61 @@
...
+    /// @notice Execute a single action.
+    /// @param _action The action to execute
+    /// @return result_ The raw returned data from the action
+    function _executeAction(Action memory _action) internal returns (bytes memory result_) {
+        bool success;
+        (success, result_) = _action.target.call{ value: _action.value }(_action.data);
+        require(success, ActionFailed());
+        emit ActionExecuted(_action.target, _action.value, _action.data);
+    }
+}
```
adaki2004, line 61: So let me get this straight.

You made a common base class/contract for the `DelegateController` (`DelegateOwner` only remains for Hekla) and `TaikoDAOController` to handle/execute calls ?

I think it is troublesome, from maintenance perspective (diff. type of testing on hekla vs. mainnet) and not only that but feature parity now.

So we have now this:
```
struct Action {
    address target;
    uint256 value;
    bytes data;
}
```

But this is not enough for `DelagateController` becaause in `DelegateOwner` previously it was like that:
  ```
struct Call {
      uint64 txId;
      address target;
      bool isDelegateCall;
      bytes txdata;
  }
```

So `delegateCall` vs. `call` distinguishing is missing, and this code section (i commenting this PR review) is lacking that:
```

(bool success, bytes memory result) = call.isDelegateCall
            ? call.target.delegatecall(call.txdata)
            : call.target.call{ value: msg.value }(call.txdata);
```

(also a good feature having `LibBytes.revertWithExtractedError(result)` ).

Comment (dantaik) [reply packages/protocol/contracts/shared/goverance/Controller.sol:61]: BTW, DelegateController and DelegateOwner are storage compatible. We can upgrade Hekla's DelegateOwner to DelegateController later.

File: packages/protocol/contracts/shared/goverance/Controller.sol [hunk 2]
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    /// @notice Execute a single action.
+    /// @param _action The action to execute
+    /// @return result_ The raw returned data from the action
+    function _executeAction(Action memory _action) internal returns (bytes memory result_) {
+        bool success;
+        (success, result_) = _action.target.call{ value: _action.value }(_action.data);
+        require(success, ActionFailed());
```
adaki2004, line 58: I think would be much easier to debug (if something fails), smth like this:
```
bytes memory usefulRevertMsg = reLibBytes.revertWithExtractedError(result_);
require(success, ActionFailed(usefulRevertMsg));
```

dantaik: Done

adaki2004 (approved):
lgtm (especially if you tested on a testnet/mainnet :) ).

davidtaikocha (approved):

packages/protocol/test/layer1/proposals/BuildProposal.sol
```diff
@@ -0,0 +1,76 @@
...
+            value: 0,
+            data: abi.encodeCall(IBridge.sendMessage, (message))
+        });
+
+        bytes memory callData = abi.encodeCall(TaikoDAOController.execute, (allActions));
+        console2.log("Proposal actions: ================================================");
+        console2.log("Num of L1 actions:", l1Actions.length);
+        console2.log("Num of L2 actions:", l2Actions.length);
+        console2.log("L2 gas limit:", l2GasLimit);
+        console2.log("Target:", L1.DAO_CONTROLLER);
```
adaki2004, line 62: I ran it (locally, dry-run) it seems good (dry run success), but the TaikoDaoController contract implementation is still the 'old one' - not this PR's one.

adaki2004: Beside that LGTM

adaki2004 (approved):

packages/protocol/test/layer1/proposals/BuildProposal.sol
dantaik: We have a DAO proposal to upgrade it first.

packages/protocol/script/layer1/proposals/BuildProposal.sol
```diff
@@ -0,0 +1,87 @@
...
+        returns (Controller.Action memory)
+    {
+        return Controller.Action({
+            target: _target,
+            value: 0,
+            data: abi.encodeCall(UUPSUpgradeable.upgradeTo, (_newImpl))
+        });
+    }
+
+    function tryrunL1Actions(uint64 executionId, uint32 l2GasLimit) internal broadcast {
```
adaki2004, line 38: Seems nothing is calling into it. (Code section in Proposal0002.s.sol is commented out).

adaki2004 (approved):

adaki2004 (approved):

davidtaikocha (approved):

YoGhurt111 (approved):

PR #19512, opened 2025-05-22
chore(protocol): update tcbs

RogerLamTd:


johntaiko (approved):

davidtaikocha (approved):

smtmfft:
better to organize in a new folder like 0525, and then delete those deprecated after we fully test these in devnet/hekla.

RogerLamTd:
@YoGhurt111 can you apply these new TCBs to hekla/devnet to test so we can be sure it doesn't cause existing users any problems? Thanks!

PR #19522, opened 2025-05-23
fix(protocol): allow preconfer from next epoch to propose in advanced in current epoch

AnshuJalan:
This PR does two things:
- It fixes an issue where the first preconfer from the next epoch was not able to propose in advanced in the current epoch.
  - Eg: [x x x x Pa y y y y] [z z z Pb v v v v] 
  - Pb was able to propose at slots **z**, but not in **y**
- While implementing this fix, I noticed we do not need the previous lookahead to be passed.
  - In the above example, we were covering proposals in slots **z** using the previous lookahead's last entry as the reference, but that is not needed. We can just use the epoch's starting timestamp as the reference boundary.
  - And slots **y** are already getting covered by the newly implemented scenario.

### DIFF
--- a/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
+++ b/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
@@ -69,8 +69,8 @@ contract PreconfRouter2 is EssentialContract, IProposeBatch {
     {
         (
             uint256 slotIndex,
-            ILookaheadStore.LookaheadSlot[] memory prevLookahead,
             ILookaheadStore.LookaheadSlot[] memory currLookahead,
+            ILookaheadStore.LookaheadSlot[] memory nextLookahead,
             bytes memory nextLookaheadUpdateData
         ) = abi.decode(
             _lookaheadData,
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -95,15 +95,21 @@ contract PreconfRouter2 is EssentialContract, IProposeBatch {
                 (bytes32 registrationRoot, bytes memory data) =
                     abi.decode(nextLookaheadUpdateData, (bytes32, bytes));
                 lookaheadStore.updateLookahead(registrationRoot, data);
+
+                // Refetch the lookahead hash for the next epoch.
+                nextLookaheadHash = lookaheadStore.getLookaheadHash(nextEpochTimestamp);
```
dantaik, line 100: Maybe we don't need `(bytes32 registrationRoot, bytes memory data)` as `bytes32 registrationRoot` + `ILookaheadStore.LookaheadSlot[] memory nextLookahead` is sufficient enough for the registration?

dantaik:
Also suggested some changes in https://github.com/taikoxyz/taiko-mono/pull/19532 with 2 questions there

dantaik (changes requested):

PR #19525, opened 2025-05-23
feat(taiko-client): check router on taiko wrapper before allowing preconf block to be made

cyberhorsey:
Will block preconfs from being submitted if we disable the router, barring a sequencer issue of failing to detect this and tryign to preconf blocks.


davidtaikocha (approved):

davidtaikocha:
Changed to use `s.rpc.GetPreconfRouterPacaya` (which includes RPC timeout), and approved.

YoGhurt111 (approved):

PR #19527, opened 2025-05-23
feat(protocol): increase randomness delay epochs in PreconfWhitelist

linoscope:
In one devnet run, we missed 34 slots, which triggered a whitelist-lookahead change halfway through the epoch. The cause is our randomness source: we draw it from the first block of the previous epoch, so the delay is only one epoch. That short buffer leaves us exposed to missed slots and reorgs. There doesn't seem to be a strong reason to have such a tight delay in randomness source, and we should extend it to two epochs (or more) to make the lookahead more robust.

TODOs:
- [x] Fix tests.
- [x] Think how this interacts with `operatorChangeDelay`
  - Added comments in https://github.com/taikoxyz/taiko-mono/pull/19527/commits/d800890450f39f04ddbd0b6c3a20ab62e33efa61


dantaik (approved):

adaki2004 (approved):
LGTM (but as a reminder, if we wanna take this into `pacaya`, we need to cherry pick, cause main is for `shasta` already.)

davidtaikocha (approved):

linoscope:
> (but as a reminder, if we wanna take this into `pacaya`, we need to cherry pick, cause main is for `shasta` already.)

Thanks for review & good point. My intent is to make our devnet runs more stable, as I highly doubt that we will have such deep reorgs/missed slots in mainnet. If you think the negative of including any changes at this point outweighs the benefit, I am fine with leaving it out for now. @cyberhorsey What do you think? Since you are most deepest in the testing process, will be happy to follow your call here.

dantaik:
What's the downside to use an even bigger value, for example, 6.

dantaik:
> LGTM (but as a reminder, if we wanna take this into `pacaya`, we need to cherry pick, cause main is for `shasta` already.)

I think pacaya has not used this feature yet. Not too late to change.

dantaik:
So if I read all the comments correctly, we are all in favor of such change, right? If so, I think we can increase the value to 6, in the code we can assert the current epoch is always at least 6 epochs later than Genesis - the contract will not function in the first 6 epochs post Genesis, which I think is fine.

linoscope:
> This can indeed happen in mainnet, or any network. It is always possible for your node to lag behind the chain due to a bug.
> 
> So by increasing the randomness further backwards, we can do _predictive lookahead_, such that the node missing slots due to an error (not the chain itself missing slots) or lagging behind, does not affect the lookahead calculations in the driver of the sequencer.

@cyberhorsey Thanks for the input, yeah, good point. We should definitely incorporate this change then.

> So if I read all the comments correctly, we are all in favor of such change, right? If so, I think we can increase the value to 6, in the code we can assert the current epoch is always at least 6 epochs later than Genesis - the contract will not function in the first 6 epochs post Genesis, which I think is fine.

@dantaik Got it, will set default to `6`. I also want to refactor a bit and move the `RANDOMNESS_DELAY_EPOCH` to the constructor, so it can be set together with the `operatorChangeDelay` (which is highly related). Will do that too.

linoscope:
@dantaik @adaki2004 Thanks for the reviews & approval. I have done a minor refactor since you guys have approved where I move the `RANDOMNESS_DELAY` from a constant to a constructor parameter. https://github.com/taikoxyz/taiko-mono/pull/19527/commits/1ff55235ec8913536c7531635d7191039f44d061

Please take a look and re-approve it if looks good :pray:

@dantaik As for increasing the delay to `6`, I am inclined to keep the delay at 2 for now. Our local workflow spins up two entire preconf stacks with Kurtosis so we can check if it is working, and a delay of 6 would add a lot of idle time. Ideally, we’d tweak `DeployProtocolOnL1` for local development to use a shorter delay, but that modification will take some effort. Doubling the delay from 1 to 2 already cuts the risk substantially, and we can always increase it later if we find the need (or once we’ve updated the local deployment script).

dantaik (approved):

PR #19529, opened 2025-05-24
fix(taiko-client): curr range lookahead fix

cyberhorsey:
current lookahead incorrectly returns current range of the current epoch for regardless of operator, this fixes it.  Not a big issue because preconf sidecars also keep track of this, but should be fixed.

### DIFF
--- a/.github/workflows/taiko-client--docker.yml
+++ b/.github/workflows/taiko-client--docker.yml
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, lookahead_fix_2]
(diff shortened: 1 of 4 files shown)


.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, lookahead_fix_2]
```
davidtaikocha, line 5:
```suggestion
    branches: [main]
```

davidtaikocha (approved):

YoGhurt111 (approved):

PR #19546, opened 2025-05-26
feat(taiko-client): fallback preconf

YoGhurt111:
It seems no fallback preconf logic in our normal proposer, so create this PR to add it.


packages/taiko-client/proposer/proposer.go
```diff
@@ -236,13 +236,15 @@ func (p *Proposer) fetchPoolContent(allowEmptyPoolContent bool) ([]types.Transac
...
-	}
-	if preconfRouter != rpc.ZeroAddress {
-		log.Info("Preconfirmation router is set, skip proposing", "address", preconfRouter, "time", time.Now())
-		return nil
+	if p.rpc.PacayaClients.PreconfRouter != nil {
+		fallbakcPreconferAddress, err := p.rpc.PacayaClients.PreconfRouter.FallbackPreconfer(&bind.CallOpts{Context: ctx})
+		if err != nil {
+			return fmt.Errorf("failed to get fallback preconfer address: %w", err)
+		}
+		if fallbakcPreconferAddress != p.proposerAddress || fallbakcPreconferAddress != p.ProverSetAddress {
```
davidtaikocha, line 244: I also think we shouldn't always propose when the proposer is the fallback proposer, it should only propose when there is no block in X minutes? Otherwise, it will just keep reorging preconf blocks?

YoGhurt111: Yes, I forgot it, let me fix it

packages/taiko-client/proposer/proposer.go
```diff
@@ -236,13 +236,15 @@ func (p *Proposer) fetchPoolContent(allowEmptyPoolContent bool) ([]types.Transac
...
 	// Check if the preconfirmation router is set, if so, skip proposing.
-	preconfRouter, err := p.rpc.GetPreconfRouterPacaya(&bind.CallOpts{Context: ctx})
-	if err != nil {
-		return fmt.Errorf("failed to fetch preconfirmation router: %w", err)
-	}
-	if preconfRouter != rpc.ZeroAddress {
-		log.Info("Preconfirmation router is set, skip proposing", "address", preconfRouter, "time", time.Now())
-		return nil
+	if p.rpc.PacayaClients.PreconfRouter != nil {
+		fallbakcPreconferAddress, err := p.rpc.PacayaClients.PreconfRouter.FallbackPreconfer(&bind.CallOpts{Context: ctx})
```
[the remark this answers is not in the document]
YoGhurt111: Fixed.

packages/taiko-client/cmd/flags/proposer.go
```diff
@@ -95,6 +97,14 @@ var (
 		Category: proposerCategory,
 		EnvVars:  []string{"L1_REVERT_PROTECTION"},
 	}
+	// Preconfirmation related
+	FallbackTimeout = &cli.DurationFlag{
+		Name:     "preconfirmation.fallback.timeout",
```
davidtaikocha, line 102:
```suggestion
		Name:     "preconfirmation.fallback.forcePushTimeout",
```

packages/taiko-client/cmd/flags/proposer.go
```diff
@@ -95,6 +97,14 @@ var (
...
+		Usage:    "Timeout to propose L2 transactions as a fallback preconfer",
+		Category: proposerCategory,
+		Value:    5 * time.Minute,
```
davidtaikocha, line 105:
```suggestion
		Value:    2 * time.Minute,
```

Let's use 2min at first.

packages/taiko-client/proposer/proposer.go
```diff
@@ -235,21 +235,37 @@ func (p *Proposer) fetchPoolContent(allowEmptyPoolContent bool) ([]types.Transac
...
 	}
 
+	// Check if the preconfirmation router is set, if so, skip proposing.
+	if p.rpc.PacayaClients.PreconfRouter != nil {
+		fallbackPreconferAddress, err := p.rpc.PacayaClients.PreconfRouter.FallbackPreconfer(&bind.CallOpts{Context: ctx})
+		if err != nil {
+			return fmt.Errorf("failed to get fallback preconfer address: %w", err)
+		}
+		if fallbackPreconferAddress != p.proposerAddress || fallbackPreconferAddress != p.ProverSetAddress {
+			log.Info("Preconfirmation is running and proposer isn't the fallback preconfer, skip proposing", "time", time.Now())
```
davidtaikocha, line 250:
```suggestion
			log.Info("Preconfirmation is activated and proposer isn't the fallback preconfer, skip proposing", "time", time.Now())
```

packages/taiko-client/proposer/proposer.go
```diff
@@ -235,21 +235,37 @@ func (p *Proposer) fetchPoolContent(allowEmptyPoolContent bool) ([]types.Transac
...
+			return nil
+		}
+		// As a fallback preconfer, need to enable preconfirmation server
+		latest, err := p.rpc.L2.BlockByNumber(ctx, nil)
+		if err != nil {
+			return fmt.Errorf("failed to get latest preconf block: %w", err)
```
davidtaikocha, line 256:
```suggestion
			return fmt.Errorf("failed to get the latest block: %w", err)
```

packages/taiko-client/proposer/proposer.go
```diff
@@ -235,21 +235,40 @@ func (p *Proposer) fetchPoolContent(allowEmptyPoolContent bool) ([]types.Transac
...
-		return nil
-	}
-
 	// Wait until L2 execution engine is synced at first.
 	if err := p.rpc.WaitTillL2ExecutionEngineSynced(ctx); err != nil {
 		return fmt.Errorf("failed to wait until L2 execution engine synced: %w", err)
 	}
 
+	// Check if the preconfirmation router is set, if so, skip proposing.
+	if p.rpc.PacayaClients.PreconfRouter != nil {
```
cyberhorsey: this check is not equivilant to the previous check of checking if it's in the address manager. this checks if we created a contract binding..so it will not be dynamic when we add the preconf router.

Once a preconf router is detected, we should instantiate an instance of it, then enter the fallback timeout loop.

cyberhorsey (changes requested):

packages/taiko-client/proposer/proposer.go
```diff
@@ -235,21 +235,40 @@ func (p *Proposer) fetchPoolContent(allowEmptyPoolContent bool) ([]types.Transac
...
+		fallbackPreconferAddress, err := p.rpc.PacayaClients.PreconfRouter.FallbackPreconfer(&bind.CallOpts{Context: ctx})
+		if err != nil {
+			return fmt.Errorf("failed to get fallback preconfer address: %w", err)
+		}
+		if fallbackPreconferAddress != p.proposerAddress || fallbackPreconferAddress != p.ProverSetAddress {
```
cyberhorsey: should be and, not or

packages/taiko-client/proposer/proposer.go
```diff
@@ -235,21 +235,40 @@ func (p *Proposer) fetchPoolContent(allowEmptyPoolContent bool) ([]types.Transac
...
+			log.Info("Preconfirmation is activated and proposer isn't the fallback preconfer, skip proposing",
+				"time",
+				time.Now(),
+			)
+			return nil
+		}
+		// As a fallback preconfer, need to enable preconfirmation server
```
cyberhorsey: we are not a fallback preconfer, we are basically going to just be a fallback proposer.

cyberhorsey (changes requested):

packages/taiko-client/proposer/proposer.go
YoGhurt111:
```suggestion
		if fallbackPreconferAddress != p.proposerAddress && fallbackPreconferAddress != p.ProverSetAddress {
```

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main, envelope_signature]
+    branches: [main, feature/fallback_preconf]
```
```suggestion
    branches: [main]
```

dantaik (approved):

cyberhorsey:
I will push up some changes to this branch I think it needs + deploy an instance of it on Masaya and test tomorrow, as this needs to make it in for a mainnet release version.

packages/taiko-client/proposer/proposer.go
```diff
@@ -235,21 +236,53 @@ func (p *Proposer) fetchPoolContent(allowEmptyPoolContent bool) ([]types.Transac
...
+		}
+		if fallbackPreconferAddress != p.proposerAddress && fallbackPreconferAddress != p.ProverSetAddress {
+			log.Info("Preconfirmation is activated and proposer isn't the fallback preconfer, skip proposing",
+				"time",
+				time.Now(),
+			)
+			return nil
+		}
+		// As a fallback proposer, need to enable preconfirmation server
+		latest, err := p.rpc.L2.BlockByNumber(ctx, nil)
```
cyberhorsey: this is gonna be a tricky one I think because in case of a reorg I think we have a race condition where when the event comes in, the chain will rewind to the first block after the last proposed block ID, and thus will be "old", and if this triggers while that happens, the fallback preconfer will propose a block, screwing everything up. 

possible solution is something perhaps better like caching the chain head, and when we don't see the head increase for N amount of time, then we propose.

Because this would handle that edge case above. Bc in that case, the chain head would go backwards, then catch up again, and the condition would still not trigger.

cyberhorsey (changes requested):

packages/taiko-client/proposer/proposer.go
cyberhorsey: also the comment here doesn't really make sense

YoGhurt111: Got your point, the core meaning is that using the block timestamp to check is not accurate enough.
Will consider your provided possible solution.

YoGhurt111: > also the comment here doesn't really make sense

It means as a fallback proposer, you should be able to receive the latest prenconf block.

YoGhurt111: Implemented it in https://github.com/taikoxyz/taiko-mono/pull/19546/commits/8e0eaa1a65a0f9324cd3f321a9a14629473c8b1b

cyberhorsey: still has same issue, i updated it.

cyberhorsey:
@YoGhurt111 after my commits this now works, tested on Masaya, and I will fix the tests when I find time later tonight or tomorrow!

packages/taiko-client/proposer/proposer.go
```diff
@@ -164,6 +195,15 @@ func (p *Proposer) eventLoop() {
 				log.Error("Proposing operation error", "error", err)
 				continue
 			}
+		case h := <-l2HeadCh:
+			// only update if the head is greater than the last updated head.
+			// we dont want to track reorgs, just the highest l2head seen.
+			if h.Number.Uint64() > p.l2HeadUpdate.blockID {
+				p.l2HeadUpdate = l2HeadUpdateInfo{
```
YoGhurt111: I have a question here: if the preconf block height is 110, but the event reorged the L2 height to 105, and after this, preconfer no longer works, it seems the fallback proposer also won't work. @cyberhorsey

cyberhorsey: hm, actually, it's the opposite. fallback would propose when it shouldn't.

scenario 1: the block would remain at 110, and let's say the block time was like 10:10 PM and the p2p reorg to block 105 at 10:12 PM, we still at 110, and then preconfer submitting blocks 106, 107, but now 2 mins has passed since 110 and we start a proposal, but a preconfer is actually working (just p2p reorg intentionally).

scenario 2: block 110 time 10:10:00, event reorg onchain reorgs to 105, block time is 10:10:05, no further blocks (preconfer is broken or whatever), we do propose BC we still have 110.

the issue with only keeping track of block times is, or updating on every block, is if they reorg via event to a very old block (is: sequencer sequences 100+ blocks but an event comes in and it's a reorg so all their 100 blocks go away), we would immediately start proposing when we shouldn't, BC the latest block is now old.

I guess it simply makes more sense to track how long it's been since any block comes in on the channel at all, rather than the blocks time itself, or block ID. I can push up something like this

packages/taiko-client/cmd/flags/proposer.go
```diff
@@ -95,12 +97,21 @@ var (
 		Category: proposerCategory,
 		EnvVars:  []string{"L1_REVERT_PROTECTION"},
 	}
+	// Preconfirmation related
+	FallbackTimeout = &cli.DurationFlag{
+		Name:     "preconfirmation.fallback.forcePushTimeout",
+		Usage:    "Timeout to propose L2 transactions as a fallback preconfer",
+		Category: proposerCategory,
+		Value:    2 * time.Minute,
```
YoGhurt111, line 105: @cyberhorsey How about you think the default value? Is it suitable?

cyberhorsey: I think it's OK, I don't think there's any excuse for a sequencer to have two minutes without a block without them being completely stalled.

packages/taiko-client/proposer/proposer.go
YoGhurt111: If the height 106 from event is a new fork and preconfer is broken, how about this case?

cyberhorsey: Same solution, we just use the actual time since receiving any new head update at all. Whether it's backwards or forwards - we don't care about the block time, but the time we received the head.

YoGhurt111: But we don't update the received time when the block num is less than we recorde last time
`if h.Number.Uint64() > p.l2HeadUpdate.blockID`

cyberhorsey: that's what I mean by solution when I said I'll push up a fix!

YoGhurt111: Got it

cyberhorsey: I pushed it up

YoGhurt111:
> @YoGhurt111 after my commits this now works, tested on Masaya, and I will fix the tests when I find time later tonight or tomorrow!

Currently I have no much time to fix the test, some contract related stuff confused me.
You said you tested on Masaya, did you use the ProverSet contract to propose? 
I guess it's no?
I think it will be fail with current ProverSet status like Hekla.

cyberhorsey:
this is tested and works btw, we will remove proverset on mainnet so it's all good

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, feature/fallback_preconf]
```
```suggestion
    branches: [main]
```

cyberhorsey (approved):

PR #19554, opened 2025-05-27
feat(protocol): improve lookahead posting and liveness

AnshuJalan:
Improved lookahead posting:
- Whitelist preconfers who are posting the lookahead via the preconf router are not required to be URC staked operators.
- `nextLookahead` in the `PreconfRouter2` is now being re-used for lookahead posting.
- Removed the `isPostedByProtector` flag from the events since I realised the node does not really need to differentiate between lookahead posted by protector vs an operator.

Improved liveness:
- If the current preconfer is slashed, the whitelist preconfer for the current epoch can take over the preconfing
- If there are no preconfers remaining for the rest of the slots in the current epoch, and the next epoch's lookahead is empty, the whitelist preconfer for the current epoch can fill in for the remaining slots.


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -133,9 +155,23 @@ contract PreconfRouter2 is EssentialContract, IProposeBatch {
...
 
-                    preconfSlotTimestamp = nextLookahead[0].slotTimestamp;
+                    if (nextLookahead.length == 0) {
+                        // This the special case when the next lookahead is empty
+                        // Eg: [x x x Pa y y y] [     empty    ]
+                        //     [  curr epoch  ] [  next epoch  ]
+                        //
+                        // The empty slots y will be taken over by the whitelist preconfer
+                        // for the current epoch.
+                        // The upper boundry of the preconfing period is the last slot of the
```
dantaik, line 165:
```suggestion
                        // The upper boundary of the preconfing period is the last slot of the
```

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -102,21 +112,33 @@ contract PreconfRouter2 is EssentialContract, IProposeBatch {
...
 
-            if (currLookahead.length == 0 || nextLookahead.length == 0) {
+            if (currLookahead.length == 0) {
                 // The current lookahead is empty, so we use a whitelisted preconfer
                 _validateWhitelistPreconfer();
             } else {
-                uint256 preconfSlotTimestamp;
-                uint256 prevSlotTimestamp;
+                uint256 preconfSlotTimestamp; // Upper boundry of the preconfing period
+                uint256 prevSlotTimestamp; // Lower boundry of the preconfing period
```
dantaik, line 140:
```suggestion
                uint256 prevSlotTimestamp; // Lower boundary of the preconfing period
```

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -40,23 +49,29 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
-        if (isPostedByProtector) {
+        bool isLookaheadRequired_ = isLookaheadRequired();
+
+        if (_registrationRoot == bytes32(0)) {
+            // If the registration root is 0, the lookahead is posted by a whitelist preconfer
+            // (via preconf router) or it is posted the protector.
+            if (msg.sender == preconfRouter) {
+                require(isLookaheadRequired_, LookaheadNotRequired());
+            } else {
+                require(msg.sender == protector, NotProtectorOrPreconfRouter());
```
dantaik, line 62: in else {}, should we also check `isLookaheadRequired_`?

dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: No, because the protector should be able to force update the lookahead if a faulty one is already posted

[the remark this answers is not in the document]
AnshuJalan: This is OK, the protector will have additional logic to perform this lookahead update besides the `ISlasher` functions. I agree, the naming is a bit confusing. The protector will be going away once we have EIP-7917, because then we can have an actual permission-less `LookaheadSlasher`

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter2.sol
```diff
@@ -166,7 +202,11 @@ contract PreconfRouter2 is EssentialContract, IProposeBatch {
                     InvalidLookaheadTimestamp()
                 );
 
-                _validateProposer(_lookaheadSlot);
+                if (nextLookahead.length == 0) {
+                    _validateWhitelistPreconfer();
```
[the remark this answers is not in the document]
AnshuJalan: I think you are right to be confused, because there is a problem here. If I want to place the conditional where it currently is, I need to do `if(slotIndex == type(uint256).max && nextLookahead.length == 0)`. I'll refactor it to be included in the `if (slotIndex == type(uint256).max)` conditional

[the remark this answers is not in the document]
AnshuJalan: Fixed in the latest commit. Let me know if it is understandable now :)

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
dantaik: Do you mean inside else {}, `isLookaheadRequired_` can be either true or false, and both values are acceptable? Just want to make sure I understand the logic.

adaki2004 (approved):

PR #19576, opened 2025-06-02
fix(taiko-client): filter genesis block verified for hekla

cyberhorsey:
broken in a long ago refactor, hekla is unsyncable bc it has a specific blockverified event in it's genesis.

### DIFF
--- a/.github/workflows/taiko-client--docker.yml
+++ b/.github/workflows/taiko-client--docker.yml
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


RogerLamTd (approved):

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, fix_filter_genesis_block_verified]
```
```suggestion
    branches: [main]
```

kimo-ice (approved):

PR #19577, opened 2025-06-02
CI: migrate workflows to golangci-lint-action v8

eeemmmmmm:
Bumps golangci-lint-action to v8 for future-proofing against runner updates. Workflows compile the same. Reference: https://github.com/golangci/golangci-lint-action/releases/tag/v8.0.0


RogerLamTd:
dependency upgrades are done by bot now

eeemmmmmm:
Got it, thanks for the answer

PR #19584, opened 2025-06-03
feat(protocol): introduce provability bond to fight prover-killer attack

dantaik:
## Problem

If the protocol mandates that all blocks must be zk-proven, a malicious actor could launch a **prover-killer attack** by proposing batches containing intentionally zk-expensive opcodes and precompiles. By assigning themselves as the prover and refusing to prove, the attacker forfeits their liveness bond—but the real damage is that no one else is willing to take on the proving cost, effectively stalling block verification.

## Proposed Solution

Introduce a **provability bond** required from the proposer for each batch. This bond is separate from the **liveness bond** and is only refunded if the batch is verified with a zk-proof submitted within a new, longer time window called the **Extended Proving Window (EPW)**.

If the batch is only proven after the EPW expires, the bond is forfeited: part of it is burned, and the rest is used to reward the late prover.

We assume all batches are eventually zk-provable, regardless of opcode complexity. The provability bond acts as the proposer’s economic guarantee that their batch is **provable within EPW**. For simplicity, we assume all batches are indeed provable within EPW.

### Example and Capital Implications

To be effective, the provability bond should be large enough to cover worst-case proving costs—for example, $1,000 per batch. According to [our Dune dashboard](https://dune.com/taiko_xyz/taikobcr), there are consistently around 1,000 unverified batches, implying ~$1,000,000 in total provability bonds locked across all proposers. For a Layer 2 protocol, this is a reasonable and non-blocking capital requirement.

## Implementation

The straightforward implementation is to refund the provability bond to the proposer when their batch is verified. However, this would require storing the `proposer` in a new slot in the `Batch` struct, increasing gas costs by over 5,000 per batch.

An alternative design avoids this gas overhead:

- When the **first transition** of a batch is proven:
  - **If within EPW**: 
    - Refund the provability bond to the proposer.
    - Require the prover to post a combined bond equal to the proposer’s provability bond + prover’s liveness bond.
  - **If beyond EPW**:
    - The provability bond is not refunded.
    - The prover only posts their liveness bond.

- If the **verifying transition is the first transition**:
  - **If proven within EPW**: prover receives both their liveness bond and the refunded provability bond.
  - **If proven after EPW**: prover still receives both bonds (provability bond is redirected from proposer to prover).

- If the **verifying transition is not the first**:
  - **If proven within EPW**: prover gets only their liveness bond.
  - **If proven after EPW**: prover receives both liveness and provability bonds.

## Summary

Regardless of whether the verifying transition is the first or not:
- **If proven after EPW** → prover receives: `liveness bond + provability bond`
- **If proven within EPW** → prover receives: `liveness bond` only

This alternative avoids introducing a new storage slot in `Batch`, preserving gas efficiency.

## TODO

I would like to get a first round of review before proceeding with the following TODOs.

- [x] add unit tests
- [x] figure out mainnet configuration values for the new parameters


packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -174,6 +181,8 @@ interface ITaikoInbox is IBondManager, IProveBatches {
         uint32 blockMaxGasLimit;
         /// @notice The amount of Taiko token as a prover liveness bond per batch.
         uint96 livenessBond;
+        /// @notice The amount of Taiko token as a proposer's provability bond per batch.
+        uint96 provabilityBond;
```
linoscope, line 188: What is the rationale for adding this new field instead of just increasing the `livenessBond` (and potentially also the liveness window)?

linoscope: Is it important that we have two stages in the window (the addition of the "extended" window) or is it important that the provabilityBond is accounted to the proposer and not prover (which, in the future, I assume can be different from proposer)?

adaki2004:
Do we know what do we gain (some more security i guess) and what do we lose (gas complexity) if instead we just apply this logic vs. increase the livenesss bond ?

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
dantaik: The provabilityBond is posted by proposers to discourage them from submitting prover-killer batches. In contrast, the livenessBond is posted by provers to ensure they submit proofs in a timely manner.

The Extended Proving Window (EPW) defines the threshold for what counts as “provable” within reasonable cost. If a proof is generated beyond the EPW, it indicates that the batch was expensive to prove and potentially malicious.

dantaik: > The provabilityBond is posted by proposers to discourage them from submitting prover-killer batches. In contrast, the livenessBond is posted by provers to ensure they submit proofs in a timely manner.
> 
> The Extended Proving Window (EPW) defines the threshold for what counts as “provable” within reasonable cost. If a proof is generated beyond the EPW, it indicates that the batch was expensive to prove and potentially malicious.



Half of the liveness bond should be large enough to incentivize open market provers to prove the block within the EPW if the assigned prover is offline. The provability bond, on the other hand, should be large enough to ensure that if no external prover is available, the proposer is economically motivated to prove their own block within the EPW.

dantaik:
> Do we know what do we gain (some more security i guess) and what do we lose (gas complexity) if instead we just apply this logic vs. increase the livenesss bond ?

The purpose of this feature is to discourage proposers from submitting prover-killer batches. Increasing the liveness bond only penalizes provers, not the proposers who introduce the unprovable or excessively costly batches in the first place.

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
linoscope: Are you assuming on-chain prover markets here? Then this might make sense, as we don't want to end up in a situation where a batch is posted on-chain that nobody wants to prove.

If we are assuming off-chain prover markets, I think it's up to the prover not to commit to unprovable blocks (e.g., via the ZK gas metering tool that discussed) hence just increasing the prover bond should be enough.

Also, even in the on-chain scenario, enshrining a ZK-gas accounting mechanism would let us pay provers in proportion to the gas they actually consume. That still incentivizes proofs for “prover-killer” batches while avoiding unprovable commitments.

linoscope: Or maybe you were exploring ways to enable on-chain prover markets without relying on ZK gas metering?

dantaik:
> Do we know what we gain (e.g., security) and what we lose (e.g., gas cost) if we apply this logic instead of simply increasing the liveness bond?

---

**Replying to both Lin and Dani:**

Your questions made me rethink the trade-offs—perhaps increasing the liveness bond is a better solution in some cases, though it comes with its own drawbacks. Let me think out loud.

We can assume provers know the proposers they’re working with—otherwise, they wouldn’t sign `ProverAuth` objects in response to proposer requests. If a proposer is known to be malicious, provers should—and likely will—refuse to work with them. For professional provers with long-term ambitions, avoiding bad proposers should be manageable.

From the protocol's perspective, our main goal is to ensure that even zk-expensive batches eventually get proven by someone in the open market. Raising the liveness bond could help guarantee that.

### The downside?

If a batch is simple and cheap to prove, but the assigned prover is offline due to technical issues, they still risk losing half of their liveness bond. That’s arguably too harsh. We could increase the proving window to reduce that risk, but long-term we want to minimize the proving window to reduce L2-to-L1 bridging latency.

BTW, I’m not trying to revisit the idea of an onchain prover market—I don’t think that’s the right direction to pursue.

---

## **EDIT**

Compared to simply increasing the liveness bond, I think the approach in this PR provides several advantages, with only a small increase in gas cost:

- **Separate proving windows for liveness and provability**  
  We decouple the Proving Window (PW) and Extended Proving Window (EPW), allowing us to tune them independently. For example, we can shrink PW to the minute level over time, while keeping EPW relatively large.

- **Shifts risk to proposers instead of provers**  
  Proposers can’t surprise provers with prover-killer batches. The upfront capital is posted by the proposer, and they bear the worst-case loss. The prover’s maximum loss remains bounded by their liveness bond.

- **Prevents targeted prover attacks**  
  With a higher liveness bond approach, a malicious proposer could pre-obtain multiple `ProverAuth` signatures and submit a series of killer batches, potentially causing large losses for the prover. This makes the prover’s trust in proposers a significant economic liability. By contrast, in the current approach, that risk is internalized by the proposer.

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
dantaik: Please see my response below.

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -105,22 +105,29 @@ interface ITaikoInbox is IBondManager, IProveBatches {
         bytes32 stateRoot;
     }
 
+    enum ProofTiming {
+        OutOfExtendedProvingWindow, // 0
+        InProvingWindow, // 1
+        InExtendedProvingWindow // 2
+
```
dantaik, line 114:
```suggestion
```

adaki2004:
I did not specifiy explicitly my thinking about the "increase liveness" argument, but i thought when you referred to the old off-chain prover market, immediately this version came to my mind below:

See the `LibProposing.sol` from that:
https://github.com/taikoxyz/taiko-mono/blob/0623b66d3180cc6a87e2a372fd57d84568a041dc/packages/protocol/contracts/L1/libs/LibProposing.sol#L199

So here, the "task" of the proposer is not only to propose, but also the ability to submit the first prove. Which means proposer is incentivized to submit on-time. Prover on the other hand not necessarily incentivized to go offline or hold on providing with the proof, because he does not gain anything at all from that - only lose. (business, reputation, etc.)

dantaik:
> I did not specifiy explicitly my thinking about the "increase liveness" argument, but i thought when you referred to the old off-chain prover market, immediately this version came to my mind below:
> 
> See the `LibProposing.sol` from that:
> 
> https://github.com/taikoxyz/taiko-mono/blob/0623b66d3180cc6a87e2a372fd57d84568a041dc/packages/protocol/contracts/L1/libs/LibProposing.sol#L199
> 
> So here, the "task" of the proposer is not only to propose, but also the ability to submit the first prove. Which means proposer is incentivized to submit on-time. Prover on the other hand not necessarily incentivized to go offline or hold on providing with the proof, because he does not gain anything at all from that - only lose. (business, reputation, etc.)

Sorry Dani, what's the point you try to make, I think I'm lost here :D

dantaik:
Let’s hold off on reviewing this PR and revisit it after whitelist-preconf with forced inclusion is live on mainnet. Minimizing breaking protocol changes will make that rollout smoother.

ggonzalez94:
adding my 2 cents here for the record after speaking with @dantaik about this. While I like the solution, I agree it is better to hold off for the moment, given this is not an attack vector until we have full zk coverage and decentralized preconfs. Also until then we might think of other ways of solving this such that it requires less/no protocol changes and/or is more capital efficient. For example:
- If the tooling to estimate proving cost off-chain is good and fast enough then provers can sign the request for a specific batch knowing what they are committing to. This, together with increasing the livenessBond solves the problem.
- If the research effort to price zk gas is successful and that becomes a reliable metric we can use that to cap blocks to prevent them from becoming too expensive to prove

One thing that we should think about sooner is how to price forced inclusions higher or require a bond from the force includer to prevent this attack even with a whitelisted set of preconfers

dantaik:
@ggonzalez94 @adaki2004

After further consideration and discussions with Jeff, Yue, and other engineers, I’ve changed my mind - I now support merging this feature to enable 100% ZK as early as possible.

The protocol change is straightforward, and the additional gas cost is minimal. In return, we gain a robust and transparent defense against prover-killer attacks. Achieving full ZK coverage early will also help us surface any potential issues with protocol-enforced ZK requirements.

dantaik:
Forced inclusion can still introduce killer blocks, which needs to be addressed in a separate PR to work with the provability bond.

linoscope:
> The protocol change is straightforward, and the additional gas cost is minimal. In return, we gain a robust and transparent defense against prover-killer attacks. Achieving full ZK coverage early will also help us surface any potential issues with protocol-enforced ZK requirements.

My thinking that, until provers have ways to protect against prover killer blocks (e.g., ZK gas tools, reputation-based black listing of proposers, etc) we won't have a prover market and proposers will provide proofs themselves. But from what you are saying, I assume you are trying to introduce off-protocol prover market _before_ having prevention mechanisms?

smtmfft:
A problem here I think is how to calculate the zk payload during the proposal in time.
There’s an underlying assumption that both the prover and proposer have a reasonably accurate estimate of the proving workload for each block. Based on that, it becomes possible to negotiate the bond fee more effectively. For example, Risc0’s Boundless includes a dry-run option (which is essentially mandatory) that precomputes the cycle count, and then the network fee can then be derived from the estimated cycles. However, the problem is, this dry-run takes at least ~30s, so it's unacceptable for our proposer. So at least for now (i.e., before a mature zk estimation, or zk itself comes out), should we use maximum prover bond??

dantaik:
> > The protocol change is straightforward, and the additional gas cost is minimal. In return, we gain a robust and transparent defense against prover-killer attacks. Achieving full ZK coverage early will also help us surface any potential issues with protocol-enforced ZK requirements.
> 
> My thinking that, until provers have ways to protect against prover killer blocks (e.g., ZK gas tools, reputation-based black listing of proposers, etc) we won't have a prover market and proposers will provide proofs themselves. But from what you are saying, I assume you are trying to introduce off-protocol prover market _before_ having prevention mechanisms?

My goal is to enable a 100% zk-based system even before tools for estimating zk proving costs become available. For killer batches, the cost of generating a proof is only known after the fact. The protocol should therefore provide provers with mechanisms to recover these potentially significant expenses. Without such mechanisms, provers are forced to absorb the cost entirely.

By allowing provers to delay proof submission and submit out of the extended proving window, the protocol offers greater flexibility and cost control.

The introduction of the ProverAuth object lays the foundation for an off-chain prover marketplace. However, this market cannot function effectively until zk-cost estimation tools are available to inform pricing and economic decisions.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -411,18 +413,38 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
+                    ts.proofTiming = uint8(ProofTiming.InProvingWindow);
+                    ts.prover = meta.prover;
+                } else if (block.timestamp <= proposedAt + config.extendedProvingWindow) {
+                    ts.proofTiming = uint8(ProofTiming.InExtendedProvingWindow);
+                    ts.prover = msg.sender;
+                } else {
+                    ts.proofTiming = uint8(ProofTiming.OutOfExtendedProvingWindow);
+                    ts.prover = msg.sender;
+                }
             }
```
ggonzalez94: This function is already pretty lengthy, why don't we at least refactor this into it's own function?
```
/// @dev Decides which time window we are in and who should be recorded as the prover.
    function _determineProofTiming(
        uint256 _proposedAt,
        Config memory _config,
        address _assignedProver
    )
        internal
        view
        returns (uint8 timing_, address prover_)
    {
        if (block.timestamp <= _proposedAt + _config.provingWindow) {
            return (uint8(ProofTiming.InProvingWindow), _assignedProver);
        } else if (block.timestamp <= _proposedAt + _config.extendedProvingWindow) {
            return (uint8(ProofTiming.InExtendedProvingWindow), msg.sender);
        } else {
            return (uint8(ProofTiming.OutOfExtendedProvingWindow), msg.sender);
        }
    }
```

and then call it from here:
```
        {
              uint256 proposedAt = uint256(meta.proposedAt).max(stats2.lastUnpausedAt);
              (ts.proofTiming, ts.prover) =
                  _determineProofTiming(proposedAt, config, meta.prover);
          }
```

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -106,22 +106,30 @@ interface ITaikoInbox is IBondManager, IProveBatches {
...
     //  @notice Struct representing transition storage
     /// @notice 4 slots used.
     struct TransitionState {
         bytes32 parentHash;
         bytes32 blockHash;
         bytes32 stateRoot;
         address prover;
-        bool inProvingWindow;
+        uint8 proofTiming;
```
ggonzalez94: Is there a reason for using `uint8` instead of the enum? I think the underlying representation of an enum is `uint8`, but it looks nicer and you get some compile check benefits

ggonzalez94 (commented):
Left a couple small comments.
My main hesitation around merging this now(but it is not a super strong opinion, so feel free to go ahead) is that we are preemptively fixing a problem that might take a few months to play out. This might end up being the best solution at the end, but if we don't think this is an issue until we have decentralized preconfs then we can avoid preemptively:
- Complicating the protocol
- Continue researching ways for provers to more accurately price batches. The ideal scenario is all of this is done off-chain and the `livenessBond` is enough

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
dantaik: Changed to use enum

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: Adopted

adaki2004:
Current code size limit violations (this is why Ci/CD fails):
kép
kép
kép
kép

dantaik:
> Current code size limit violations (this is why Ci/CD fails): kép kép kép kép

Once the PR is approved, I'll refactor the code to use libraries.

linoscope:
> My goal is to enable a 100% zk-based system even before tools for estimating zk proving costs become available. For killer batches, the cost of generating a proof is only known after the fact. The protocol should therefore provide provers with mechanisms to recover these potentially significant expenses. Without such mechanisms, provers are forced to absorb the cost entirely.

One concern is that we are just "kicking the can down the road" here. Yes, the prover will have less risk, but now the proposer now bare the risk, and without a ZK gas tool, the proposer is not sure if the block they are proposing is prover-killer or not. In other words, requiring proposer to put up a bond is good for preventing malicious proposers who intentionally propose prover-killer blocks, but if the proposer can unintentionally propose prover-killer blocks the bond will just disincentivize being a proposer.

I agree with @ggonzalez94 with
> My main hesitation around merging this now(but it is not a super strong opinion, so feel free to go ahead) is that we are preemptively fixing a problem that might take a few months to play out. This might end up being the best solution at the end, but if we don't think this is an issue until we have decentralized preconfs then we can avoid preemptively:

We do need to think of dealing with prover-killer blocks, but I am not csure that requiring proposers to put up a bond is the correct solution, even in the short term (maybe it is, but not convinced yet). My intuition right now is that we probably need some form of ZK cost estimation tool to get 100% ZK coverage, even if the tool is super simple like counting the number of costly precompile usage in a block.

packages/protocol/gas-reports/inbox_without_provermarket.txt
```diff
@@ -1,5 +1,5 @@
 See `test_inbox_measure_gas_used` in InboxTest_ProposeAndProve.t.sol
 
-Gas per proposing: 195386
-Gas per proving + verification: 160360
-Total: 355746
\ No newline at end of file
+Gas per proposing: 250469
+Gas per proving + verification: 169406
+Total: 419875
```
dantaik: Before this PR and https://github.com/taikoxyz/taiko-mono/pull/19589/files#diff-aad629df7af06bcd5781c14b615145853ad7ebc078238b3797d8e0c7b4d8368f, **the Total was 329094, now it's 419875, a 27.6% increase**, why? @adaki2004 can you investigate into it?

dantaik:
> One concern is that we are just "kicking the can down the road" here. Yes, the prover will have less risk, but now the proposer now bare the risk, and without a ZK gas tool, the proposer is not sure if the block they are proposing is prover-killer or not. In other words, requiring proposer to put up a bond is good for preventing malicious proposers who intentionally propose prover-killer blocks, but if the proposer can unintentionally propose prover-killer blocks the bond will just disincentivize being a proposer.

I believe we do need both the zk-estimation tool and the provability bond feature in this PR -- they are both essential. Without the former, legitimate proposers could be unfairly slashed if malicious killer transactions are included. Without the latter, a malicious proposer could deliberately attack the entire chain.

dantaik:
> > Current code size limit violations (this is why Ci/CD fails): kép kép kép kép
> 
> Once the PR is approved, I'll refactor the code to use libraries.

This is fixed.

packages/protocol/gas-reports/inbox_without_provermarket.txt
adaki2004: Please see my almost "line-to-line" comparison for the `main` vs. `one_nonzero_anchor_id_per_batch` : https://github.com/taikoxyz/taiko-mono/pull/19644 .

packages/protocol/contracts/layer1/based/libs/LibAuth.sol
```diff
@@ -2,70 +2,55 @@
...
-    }
-
-    error InvalidBatchId();
-    error InvalidSignature();
-    error InvalidValidUntil();
-    error InvalidProver();
-    error EtherAsFeeTokenNotSupportedYet();
-    error SignatureNotEmpty();
-
     function validateProverAuth(
```
ggonzalez94, line 13: nit to be consistent with internal functions being named with an underscore
```suggestion
    function _validateProverAuth(
```

ggonzalez94: actually I see some internals have it and others don't. Do we have a convention around it?

packages/protocol/contracts/layer1/based/libs/LibAuth.sol
```diff
@@ -2,70 +2,55 @@
 pragma solidity ^0.8.24;
 
 import "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
+import { ITaikoInbox as I } from "../ITaikoInbox.sol";
```
ggonzalez94, line 5: A bit of a nit:
I'm not a huge fan of single letters for variable names or imports, maybe even if it is a bit more verbose we use `Inbox`?

packages/protocol/contracts/layer1/based/libs/LibVerify.sol
```diff
@@ -78,9 +79,28 @@ library LibVerification {
...
-                    creditBond(_state, ts.prover, bondToReturn);
+                    {
+                        uint96 bondToReturn;
+                        if (ts.proofTiming == I.ProofTiming.InProvingWindow) {
+                            // all liveness bond is returned to the prover, this is not a reward.
+                            bondToReturn = batch.livenessBond;
+                            if (tid == 1) bondToReturn += batch.provabilityBond;
+                        } else if (ts.proofTiming == I.ProofTiming.InExtendedProvingWindow) {
+                            // prover is rewarded with bondRewardPtcg% of the liveness bond.
+                            bondToReturn = batch.livenessBond * _config.bondRewardPtcg / 100;
```
ggonzalez94, line 90: Since we are going trough this overhaul it might be worth having two separate reward percentages. Since these are hardcoded values that are returned by `_getConfig` it shouldn't be an issue and it gives us more flexibility

packages/protocol/contracts/layer1/based/libs/LibPropose.sol
```diff
@@ -0,0 +1,397 @@
...
+
+            output_.params = abi.decode(_params, (I.BatchParams));
+
+            {
+                if (_input.inboxWrapper == address(0)) {
+                    if (output_.params.proposer == address(0)) {
+                        output_.params.proposer = msg.sender;
+                    } else {
+                        require(output_.params.proposer == msg.sender, I.CustomProposerNotAllowed());
+                    }
```
ggonzalez94, line 60: I may be missing some context here, but why don't we just always set params.proposer = msg.sender? Is it to avoid proposers thinking they can just set a custom address and we overwriting that? Maybe if the doc is explicit we can simplify that
```suggestion
                    output_.params.proposer = msg.sender
```

ggonzalez94: Also more broadly... why is it that if the call happens trough the InboxWrapper they can set a custom proposer but when calling the Inbox directly they can't?

packages/protocol/contracts/layer1/based/libs/LibVerify.sol
```diff
@@ -78,9 +79,28 @@ library LibVerification {
...
+                            if (tid == 1) bondToReturn += batch.provabilityBond;
+                        } else if (ts.byAssignedProver) {
+                            // The assigned prover gets back his liveness bond, and 100% provability
+                            // bond.
+                            // This allows him to user a higher gas price to submit his proof first.
+                            bondToReturn = batch.provabilityBond;
```
ggonzalez94, line 96: Am I understanding correctly that this if statement is reached if this was proven outside the provabilityWindow BUT by the assigned prover? In other words, they were late but they still did it?
If that's the case should they keep their full bond? They still have an incentive to verify the batch and retain a portion of it

ggonzalez94 (commented):
Did another pass and left a few comments. They are mostly small things, I think the whole logic for the bonds looks good and I really like that we are moving the logic to libraries(TaikoInbox was getting way too big).
I do think there's room for splitting the libraries even more(with some internal/smaller functions instead of the single monolithic ones), but that can be addressed in a separate PR.

dantaik:
Closed in favor of https://github.com/taikoxyz/taiko-mono/pull/19664

PR #19586, opened 2025-06-05
feat(protocol): mark forced-inclusion blocks in extraData

dantaik:
The 16th bit in block header's extraData field will be 1 for blocks from forced inclusion store.


linoscope (commented):
Change itself LGTM, but want confirmations from implementors of forced-inclusion support in sidecars that this indeed makes things easier (I am pretty sure it does, but need to confirm)

mikhailUshakoff (approved):
LGTM

adaki2004 (approved):

PR #19587, opened 2025-06-05
feat(protocol): adjust base fee on L2

dantaik:
Ethereum developers are testing a fourfold increase of the gas limit to 150 million as part of the upcoming Fusaka hard fork, scheduled for late 2025.

150 million gas per block (12 seconds) translates into 12.5 million gas per second. Assuming Taiko preconfirm blocks per 2 seconds, our block gas limit should be 25 million.


cyberhorsey:
@dantaik is this PR intended for Shasta? Requires driver and taiko-geth changes and a mandatory node upgrade

dantaik:
> @dantaik is this PR intended for Shasta? Requires driver and taiko-geth changes and a mandatory node upgrade

We need to either cherrypick changes to apply to the Pacaya release branch, or  upgrade Hekla first to Shasta. This depends on how many changes we need to cherrypick, for example, if we also need https://github.com/taikoxyz/taiko-mono/pull/19586, then we probably want to upgrade to Shasta first?

packages/protocol/contracts/layer1/mainnet/MainnetInbox.sol
```diff
@@ -33,15 +33,20 @@ contract MainnetInbox is TaikoInbox {
...
+            // Assuming Taiko preconfirm blocks per 2 seconds, our block gas limit should be 25
+            // million.
+            blockMaxGasLimit: 25_000_000,
             livenessBond: 125e18, // 125 Taiko token per batch
             stateRootSyncInternal: 4,
             maxAnchorHeightOffset: 96,
             baseFeeConfig: LibSharedData.BaseFeeConfig({
                 adjustmentQuotient: 8,
-                gasIssuancePerSecond: 5_000_000,
+                gasIssuancePerSecond: 12_500_000, // 12.5 million gas per second
```
adaki2004, line 47: For `gasIssuancePerSecond`, should not we shoot for a very realistic setup, according to data (Hekla + Mainnet), cause now we need to take `baseFee` into consideration, and actually pay to preconfers/side-car runners  (possibly from revenue) ?

If so, our current mainnet gas issuance per day is like cca: 78.600.000.000, which translates to **841K gas issued / second** on mainnet.

I guess we will be running an incentivization campaign for sure, but overshooting with 15x seems a bit high, no ? Meaning we wont be gaining anything based on traffic/demand, since `baseFee` will always be low.

packages/protocol/contracts/layer1/mainnet/MainnetInbox.sol
```diff
@@ -33,15 +33,22 @@ contract MainnetInbox is TaikoInbox {
...
+                // Assuming Taiko preconfirm blocks per 2 seconds, our block gas limit should be 25
+                // million.
                 adjustmentQuotient: 8,
-                gasIssuancePerSecond: 5_000_000,
+                gasIssuancePerSecond: 12_500_000, // 12.5 million gas per second
+                // TODO(daniel): need to change this magic number to make sure the the min price stay
+                // 0.01 gwei
                 minGasExcess: 1_344_899_430, // 0.01 gwei
-                maxGasIssuancePerBlock: 600_000_000 // two minutes: 5_000_000 * 120
+                maxGasIssuancePerBlock: 250_000_000 // two minutes: 12_500_000 * 20
```
dantaik, line 51: we probably need to change this value (maxGasIssuancePerBlock) down a lot.

PR #19589, opened 2025-06-05
feat(protocol): make L1-anchoring optional per block but at least once per batch

dantaik:
## Previous behavior:
- Only the last (tail) block was allowed to have a non-zero anchor ID.

## Proposed changes:
- Each block may anchor to a different L1 block.
- At least one anchor ID is required per batch
- No two blocks may anchor to the same L1 block, block with higher ID must anchor to a higher L1 block id.
- The first/last blocks in a batch are no longer treated specially.


adaki2004:
> @adaki2004 could you please take over and ready this PR?

Most probably unfortunately only can start on Tuesday. 
(Friday is taken out holiday, and Monday is Whitsun Monday - public holiday, and today i'm in the middle of some revm thing.)

packages/protocol/test/layer1/based/InboxTestBase.sol
```diff
@@ -319,7 +319,6 @@ abstract contract InboxTestBase is Layer1Test {
             console2.log(unicode"│    |── lastBlockTimestamp:", batch.lastBlockTimestamp);
             console2.log(unicode"│    |── lastBlockId:", batch.lastBlockId);
             console2.log(unicode"│    |── livenessBond:", batch.livenessBond);
```
dantaik, line 339:
```suggestion
            console2.log(unicode"│    |── livenessBond:", batch.livenessBond);
            console2.log(unicode"│ |── anchorBlockId:", batch.anchorBlockId);
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -143,13 +143,7 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
                 state.batches[(stats2.numBatches - 1) % config.batchRingBufferSize];
 
-            (uint64 anchorBlockId, uint64 lastBlockTimestamp) = _validateBatchParams(
-                params,
-                config.maxAnchorHeightOffset,
-                config.maxSignalsToReceive,
-                config.maxBlocksPerBatch,
-                lastBatch
-            );
+            uint64 lastAnchorBlockId;
```
adaki2004, line 149: i think it should be

```suggestion
            uint64 lastAnchorBlockId = lastBatch.achorBlockId;
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -253,8 +284,8 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
 
                 // SSTORE #2 {{
                 batch.batchId = stats2.numBatches;
-                batch.lastBlockTimestamp = lastBlockTimestamp;
-                batch.anchorBlockId = anchorBlockId;
+                batch.lastBlockTimestamp = info_.lastBlockTimestamp;
+                batch.anchorBlockId = lastAnchorBlockId;
```
adaki2004: Just had a quick look on this PR (public holiday here so not really working today), but i think this sentence (in PR description) is not satisfied:

`At least one anchor ID is required per batch`

So if you check the added logic some lines above:

```
for (uint256 i; i < nBlocks; ++i) {
     if (params.blocks[i].anchorBlockId != 0) {
         ....
         LOGIC
         ....
      }
```
  }


Since `params.blocks[i].anchorBlockId` is an input of this function, what happens if it completely let is be 0 everywhere (every block) ? I dont see anything forcing that at least 1 `params.blocks[i].anchorBlockId` is NOT 0.

Maybe we should put something like 

`require (lastBatch.anchorBlockId < lastAnchorBlockId)`

If you agree with my findings please put a 👍  to these findings and i'll incorporate these tomorrow.

YoGhurt111:
Any updates about this PR?

adaki2004:
> Any updates about this PR?

Yeah, started addressing it’s issues and fix tests in another PR, will commit them some point today.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -176,15 +172,50 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
+
+                        // anchor block id must be strictly increasing
+                        require(
+                            params.blocks[i].anchorBlockId > lastAnchorBlockId,
+                            AnchorBlockIdSmallerThanParent()
+                        );
+                        lastAnchorBlockId = params.blocks[i].anchorBlockId;
+
+                        info_.anchorBlockIds[i] = lastAnchorBlockId;
+                        info_.anchorBlockHashes[i] = blockhash(params.blocks[i].anchorBlockId);
```
linoscope: This [this doc](https://github.com/OpenZeppelin/minimal-rollup/blob/62a540693ab5e17c7d45794643ceefe284d944a4/documentation/PreemptiveProvableAssertions.md#:~:text=but%20as%20an%20optimisation%2C%20the%20sequencer%20could%20reproduce%20the%20entire%20chain%20of%20L1%20block%20headers%20on%20L2%20(starting%20from%20the%20last%20validated%20one)%20when%20proving%20the%20assertions.%20If%20the%20last%20block%20hash%20is%20validated%20on%20L1%2C%20this%20implicitly%20validates%20the%20entire%20chain.) by @nikeshnazareth makes a good point that we can optimize by only requiring the last `blockhash` to be fetched here, as the validity of the latest blockhash transitively validates the past block hashes too.

> as an optimisation, the sequencer could reproduce the entire chain of L1 block headers on L2 (starting from the last validated one) when proving the assertions. If the last block hash is validated on L1, this implicitly validates the entire chain.

This would save L1 gas cost for the reduced blockhash fetching. However, this requires changes in the L2 execution/prover code, so it can be left as TODO for now.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -176,15 +172,56 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
-                    anchorBlockId: anchorBlockId,
-                    anchorBlockHash: blockhash(anchorBlockId),
+                    anchorBlockIds: new uint64[](nBlocks), // to be initialised later
+                    anchorBlockHashes: new bytes32[](nBlocks), // to be initialised
+                        // later
                     baseFeeConfig: config.baseFeeConfig
                 });
 
-                require(info_.anchorBlockHash != 0, ZeroAnchorBlockHash());
+                require(info_.lastBlockTimestamp != 0, NoBlockIsAnchored());
```
linoscope: Probably just missing something, but how does this assertion check that "no block is anchored" (which I assume is to assert that at least one block in batch has non-zero anchor id) ?

linoscope: Or potentially it's a naming issue of `NoBlockIsAnchored`.

linoscope: Actually, from reading the code more not sure when `info_.lastBlockTimestamp` can be zero.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -176,15 +172,56 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
+
+                for (uint256 i; i < nBlocks; ++i) {
+                    if (params.blocks[i].anchorBlockId != 0) {
+                        if (lastAnchorBlockId == 0) {
+                            require(
+                                params.blocks[i].anchorBlockId + config.maxAnchorHeightOffset
+                                    >= block.number,
+                                AnchorBlockIdTooLarge()
+                            );
```
linoscope: Don't we have to check this for the `lastAnchorBlockId != 0` case too?

linoscope: Actually, why have special treatment for the `if (lastAnchorBlockId == 0)` case at all?

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: `blockhash` opcode only cost 20 gas. I think it's OK keep the current approach.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: This line shall be removed.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: We only need to check the first anchoring block to make sure the anchor id is not too small, all other anchoring blocks to make sure their anchor ids are strictly increasing...

dantaik:
This PR now have a few more changes, please take another look @adaki2004 @YoGhurt111 @linoscope @ggonzalez94

adaki2004:
> This PR now have a few more changes, please take another look @adaki2004 @YoGhurt111 @linoscope @ggonzalez94

I see multiple compile errors now. Have you tried compiling locally ?

```
Error (6275): Source "contracts/layer1/based/libs/LibVerify.sol" not found: File not found. Searched the following locations: "/Users/keszeyd/work/alethia_try/taiko-mono/packages/protocol".
ParserError: Source "contracts/layer1/based/libs/LibVerify.sol" not found: File not found. Searched the following locations: "/Users/keszeyd/work/alethia_try/taiko-mono/packages/protocol".
  --> contracts/layer1/based/TaikoInbox.sol:14:1:
   |
14 | import "./libs/LibVerify.sol";

Error (9589): Identifier is not a library name.
  --> contracts/layer1/based/TaikoInbox.sol:33:11:
   |
33 |     using LibVerify for ITaikoInbox.State;
   |           ^^^^^^^^^

Error: Compilation failed
 ELIFECYCLE  Command failed with exit code 1.
 ELIFECYCLE  Test failed. See above for more details.
 

Error (7920): Identifier not found or not unique.
   --> contracts/layer1/based/TaikoInbox.sol:791:18:
    |
791 |         returns (ProofTiming timing_, address prover_)
```

I guess `LibVerify` is just got renamed to `LibVerification` (in main so one merge from base caused this), but `ProofTiming` struct is not defined.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
linoscope: Reading more now, think my confusion stems from the naming. Will leave comments on the latest ver.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -183,15 +179,54 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
 
-                require(info_.anchorBlockHash != 0, ZeroAnchorBlockHash());
+                for (uint256 i; i < nBlocks; ++i) {
+                    uint64 anchorBlockId = params.blocks[i].anchorBlockId;
+                    if (anchorBlockId != 0) {
+                        if (lastAnchorBlockId == 0) {
+                            // This is the first non zero anchor block id in the batch.
+                            require(
+                                anchorBlockId + config.maxAnchorHeightOffset >= block.number,
+                                AnchorIdTooLarge()
```
linoscope: Shouldn't this be:
```suggestion
                                AnchorIdTooSmall()
```

We are checking that anchor id isn't too far in the past, i.e., too small.

Note that `AnchorIdTooSmall` doesn't exist in current ver so this won't compile.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -682,28 +716,13 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
 
         unchecked {
-            if (_params.anchorBlockId == 0) {
-                anchorBlockId_ = uint64(block.number - 1);
-            } else {
-                require(
-                    _params.anchorBlockId + _maxAnchorHeightOffset >= block.number,
-                    AnchorBlockIdTooSmall()
-                );
-                require(_params.anchorBlockId < block.number, AnchorBlockIdTooLarge());
```
linoscope, line 699: Is it intentional that we removed this?

adaki2004: Basically (AFAIK) the logic (slightly changed) is placed into the code section of the new changes where:

```
for (uint256 i; i < nBlocks; ++i) {
    uint64 anchorBlockId = params.blocks[i].anchorBlockId;
    if (anchorBlockId != 0) {
        ... HERE ...
      }
 }
```

linoscope: But I don't see the `AnchorBlockIdTooLarge` part?

linoscope: Well, there is `AnchorIdTooLarge` but as I mentioned in other thread I think that is supposed to be `AnchorIdTooSmall`

dantaik:
> > This PR now have a few more changes, please take another look @adaki2004 @YoGhurt111 @linoscope @ggonzalez94
> 
> I see multiple compile errors now. Have you tried compiling locally ?
> 
> ```
> Error (6275): Source "contracts/layer1/based/libs/LibVerify.sol" not found: File not found. Searched the following locations: "/Users/keszeyd/work/alethia_try/taiko-mono/packages/protocol".
> ParserError: Source "contracts/layer1/based/libs/LibVerify.sol" not found: File not found. Searched the following locations: "/Users/keszeyd/work/alethia_try/taiko-mono/packages/protocol".
>   --> contracts/layer1/based/TaikoInbox.sol:14:1:
>    |
> 14 | import "./libs/LibVerify.sol";
> 
> Error (9589): Identifier is not a library name.
>   --> contracts/layer1/based/TaikoInbox.sol:33:11:
>    |
> 33 |     using LibVerify for ITaikoInbox.State;
>    |           ^^^^^^^^^
> 
> Error: Compilation failed
>  ELIFECYCLE  Command failed with exit code 1.
>  ELIFECYCLE  Test failed. See above for more details.
>  
> 
> Error (7920): Identifier not found or not unique.
>    --> contracts/layer1/based/TaikoInbox.sol:791:18:
>     |
> 791 |         returns (ProofTiming timing_, address prover_)
> ```
> 
> I guess `LibVerify` is just got renamed to `LibVerification` (in main so one merge from base caused this), but `ProofTiming` struct is not defined.

please run `pnpm clean` first

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -242,9 +242,9 @@ interface ITaikoInbox is IBondManager, IProveBatches {
     /// @param blockHash The hash of the verified batch.
     event BatchesVerified(uint64 batchId, bytes32 blockHash);
 
-    error AnchorBlockIdSmallerThanParent();
-    error AnchorBlockIdTooLarge();
-    error AnchorBlockIdTooSmall();
+    error AnchorIdSmallerOrEqualThanLastBatch();
+    error AnchorIdSmallerThanParent();
+    error AnchorIdTooLarge();
```
dantaik:
```suggestion
    error AnchorIdTooSmall();
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: `AnchorIdTooLarge` changed to `AnchorIdTooSmall`.

Now we don't check `_params.anchorBlockId < block.number` because if `_params.anchorBlockId >= block.number`, then `blockhash` opcode will return zero, then the tx will revert.

linoscope: > AnchorIdTooLarge changed to AnchorIdTooSmall.

Thanks!

> Now we don't check _params.anchorBlockId < block.number because if _params.anchorBlockId >= block.number, then blockhash opcode will return zero, then the tx will revert.

Makes sense.

packages/protocol/gas-reports/inbox_without_provermarket.txt
```diff
@@ -1,5 +1,5 @@
 See `test_inbox_measure_gas_used` in InboxTest_ProposeAndProve.t.sol
 
-Gas per proposing: 168855
-Gas per proving + verification: 160239
-Total: 329094
\ No newline at end of file
+Gas per proposing: 195386
```
dantaik: @adaki2004 also here, why gas cost go up so high...

adaki2004: Please see my best analysis i could do here: https://github.com/taikoxyz/taiko-mono/pull/19644

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -183,15 +179,54 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
                     blobByteOffset: params.blobParams.byteOffset,
                     blobByteSize: params.blobParams.byteSize,
                     gasLimit: config.blockMaxGasLimit,
-                    lastBlockId: lastBatch.lastBlockId + uint64(params.blocks.length),
-                    lastBlockTimestamp: lastBlockTimestamp,
+                    lastBlockId: lastBatch.lastBlockId + uint64(nBlocks),
+                    lastBlockTimestamp: _validateBatchParams(
```
ggonzalez94: nit: I suggest we keep the function call before we construct the struct. It is easier to reason about it IMO than having it here embeded

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -183,15 +179,54 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
-                require(info_.anchorBlockHash != 0, ZeroAnchorBlockHash());
+                for (uint256 i; i < nBlocks; ++i) {
+                    uint64 anchorBlockId = params.blocks[i].anchorBlockId;
+                    if (anchorBlockId != 0) {
+                        if (lastAnchorBlockId == 0) {
+                            // This is the first non zero anchor block id in the batch.
+                            require(
+                                anchorBlockId + config.maxAnchorHeightOffset >= block.number,
+                                AnchorIdTooSmall()
+                            );
```
ggonzalez94: This is a dumb thing. But I think this is just easier to read / reason about:
```suggestion
                            require(
                                anchorBlockId >= block.number - config.maxAnchorHeightOffset,
                                AnchorIdTooSmall()
                            );
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -183,15 +179,54 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
+                    anchorBlockHashes: new bytes32[](nBlocks), // to be initialised
+                        // later
                     baseFeeConfig: config.baseFeeConfig
                 });
 
-                require(info_.anchorBlockHash != 0, ZeroAnchorBlockHash());
+                for (uint256 i; i < nBlocks; ++i) {
+                    uint64 anchorBlockId = params.blocks[i].anchorBlockId;
+                    if (anchorBlockId != 0) {
+                        if (lastAnchorBlockId == 0) {
```
ggonzalez94: Can you simplify some of this conditional logic by setting `lastAnchorBlobckId = lastBatch.anchorBlockId` before the for loop? This way you don't have to differentiate about it being zero the first time and avoid the deep ifs. You do lose the ability to recognize between `AnchorIdSmallerThanParent` and `AnchorIdSmallerOrEqualThanLastBatch`, but I don't know how important that is

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -183,15 +179,54 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
+                                AnchorIdSmallerOrEqualThanLastBatch()
+                            );
+                        } else {
+                            // anchor block id must be strictly increasing
+                            require(anchorBlockId > lastAnchorBlockId, AnchorIdSmallerThanParent());
+                        }
+                        lastAnchorBlockId = anchorBlockId;
+
+                        info_.anchorBlockIds[i] = lastAnchorBlockId;
+                        info_.anchorBlockHashes[i] = blockhash(anchorBlockId);
```
ggonzalez94: nit: this is functionally correct as at this point `lastAnchorBlockId` and `anchorBlockId` have the same value, but for clarity we should use one or the other.
```suggestion
                        info_.anchorBlockIds[i] = anchorBlockId;
                        info_.anchorBlockHashes[i] = blockhash(anchorBlockId);
```

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -183,15 +179,54 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
...
+                    }
+                }
+
+                // Ensure that if msg.sender is not the inboxWrapper, at least one block must have a
+                // non-zero anchor block id. Otherwise, delegate this validation to the inboxWrapper
+                // contract.
+                require(
+                    msg.sender == inboxWrapper || lastAnchorBlockId != 0,
+                    NoAnchorBlockIdWithinThisBatch()
+                );
```
ggonzalez94: sorry if this is obvious, but I'm not familiar with the `InboxWrapper`. Why can't we have the validation for the anchor txs in a single place? Regardless if it comes from a delayed batch(which is what I think the wrapper handles?)

ggonzalez94: I think it is because forced inclusions don't require this validation but regular batches(which might still come trough the wrapper) do?

ggonzalez94 (commented):
Looking good. Just did a first pass and will continue later.
I left a few suggestions, but have a couple overall questions/thoughts:
1. > No two blocks may anchor to the same L1 block, block with higher ID must anchor to a higher L1 block id.

Why do we have this restriction if L2 blocks are faster than L1 blocks? So you may have multiple L2 blocks on the duration of 1 L1 slot. Is the idea that if multiple L2 blocks share the same anchor you just write one and the others are assumed to have the same one?

2. I imagine we can refactor/encapsulate some of the logic into private/internal functions to avoid the propose batch function to keep growing, but I haven't thought about it yet. I'll try to do that and come back with some suggestion

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: When we set up a private devnet, `block.number` may be smaller than `config.maxAnchorHeightOffset`, so we need to remove "-" and use "+"

packages/protocol/contracts/layer1/based/TaikoInbox.sol
dantaik: Yes that's the reason. But I also don't like that the validation logics are split across these two contracts.

ggonzalez94: Why do we need to support both paths? Can't we rely on the `inboxWrapper` immutable variable being set on the inbox? In that case we can keep all the validation logic on the wrapper itself.

Another option, which I like more is making the Inbox itself a force inclusion store, and checking its own storage to see if there are delayed txs. This can be achieved with an abstract contract that the Inbox inherits, here's an [example](https://github.com/OpenZeppelin/minimal-rollup/blob/main/src/protocol/taiko_alethia/TaikoInbox.sol#L78) of how it is implemented in the minimal-rollup.
I know this is a larger refactor, and I'm not suggesting we implement it as part of this PR, but I want to know what you think about this approach

dantaik:
Closed in favor of https://github.com/taikoxyz/taiko-mono/pull/19664

PR #19601, opened 2025-06-10
feat(protocol): fix test issues

adaki2004:


adaki2004:
Seems i'm finding the test issue and it is indeed a test issue (something to do with how we using the `vm.prank()` ) but give me some time to figure it out exactly.

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -217,6 +217,12 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
                     }
                 }
 
+                // To ensure there is at least 1 anchoBlockId within the submitted blocks of the
+                // batch
+                require(
+                    lastBatch.anchorBlockId < lastAnchorBlockId, NoAnchorBlockIdWithinThisBatch()
```
adaki2004, line 223: This line ensures this requirement of:

- `At least one anchor ID is required per batch`

Without it, it was possible to not fill the 
`batchParams.blocks[i0].anchorBlockId` and still succeed as proposal.

dantaik, line 223: I think we should check `lastAnchorBlockId !=0` instead. On line 223, we have already checked against lastBatch.anchorBlockId

YoGhurt111: I agree.
Shall we keep this logic
```
if (_params.anchorBlockId == 0) {
                anchorBlockId_ = uint64(block.number - 1);
            }
```
And if one anchorBlockId in BlockParams is 0, all anchorBlockId in BlockParams should be 0.

dantaik: Also, forced inclusion batch (which contains only 1 block) do not need to anchor -- otherwise, the anchor block id cannot be determined when the forced inclusion request is created.

```
require(lastAnchorBlockId != 0 || params.blocks.length == 1 , ...) 
```

adaki2004: > I think we should check `lastAnchorBlockId !=0` instead. On line 223, we have already checked against lastBatch.anchorBlockId

I think this suggestion is less strict. (might not satisfy your initial requirement of: `No two blocks may anchor to the same L1 block`)

So with this one, i think it is possible to have 
BATCH: 100
and
BATCH: 101

Anchor to the same L1 height If both has 1 L2 blocks inside.

See this line: 

```
require(
   params.blocks[i].anchorBlockId >= lastBatch.anchorBlockId,
    AnchorBlockIdSmallerThanParent()
);
```
Is this ok with you ?

adaki2004: > I agree. Shall we keep this logic
> 
> ```
> if (_params.anchorBlockId == 0) {
>                 anchorBlockId_ = uint64(block.number - 1);
>             }
> ```
> 
> And if one anchorBlockId in BlockParams is 0, all anchorBlockId in BlockParams should be 0.

Currently we allow them to be 0 BUT with the restrictions that at least 1 has to be set, and set bigger than the previous one.

dantaik: The idea is to request the proposer to provide an valid, non-zero value if the proposer would like to sync with L1. If the anchorBlockId is 0. the proposer is not synching L1 for this L2 block, then we should not override 0 with a default value.

PR #19602, opened 2025-06-10
fix(protocol): allow params.proposer == msg.sender

dantaik:
This is necessary when proverAuth is used as `params.proposer` will be hashed


YoGhurt111:
Why is `LibProverAuth` removed?

adaki2004:
> Why is `LibProverAuth` removed?

Because it was a duplicate. (There was one in `contracts/layer1/based` and `contracts/layer1/based/libs`).

adaki2004 (approved):

YoGhurt111 (approved):

PR #19609, opened 2025-06-12
chore(protocol): use type(uint16).max for numTransaction in forced transaction

dantaik:
The previous implementation is bug free but also very confusing. This implementation use a fixed number so the max transactions per forced inclusion batch is always type(uint16).max.


packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
```diff
@@ -119,7 +119,7 @@ contract TaikoWrapper is EssentialContract, IProposeBatch {
         require(p.isForcedInclusion, ITaikoInbox.InvalidForcedInclusion());
 
         // Need to make sure enough transactions in the forced inclusion request are included.
-        require(p.blocks[0].numTransactions >= MIN_TXS_PER_FORCED_INCLUSION, InvalidBlockTxs());
+        require(p.blocks[0].numTransactions != MAX_TXS_PER_FORCED_INCLUSION, InvalidBlockTxs());
```
YoGhurt111: Why not using `<` here?

linoscope: If `<` then proposer can bypass the forced inclusion by setting `0` as `numTransactions`.

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
```diff
@@ -48,7 +48,7 @@ contract TaikoWrapper is EssentialContract, IProposeBatch {
     error InvalidSignalSlots();
     error OldestForcedInclusionDue();
 
-    uint16 public constant MIN_TXS_PER_FORCED_INCLUSION = 512;
```
linoscope, line 51: IIRC `512` was chosen from the concern that a malicious forced inclusion submitter submits a batch with many txs, incurring lots of proving costs to the preconfer. However, thinking of it now, a malicious submitter can just post a forced-inclusion batch with 1 tx that uses crazy amount of proving resource, so guess number of txs doesn't matter much actually.

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
```diff
@@ -119,7 +119,7 @@ contract TaikoWrapper is EssentialContract, IProposeBatch {
         require(p.isForcedInclusion, ITaikoInbox.InvalidForcedInclusion());
 
         // Need to make sure enough transactions in the forced inclusion request are included.
```
linoscope: Think the main confusion point was the unclarity on whether forced-inclusion batch submitters have to sufficiently fill their batch to some amount. Would be good to clarify that that is not the case, like:
```suggestion
        // Preconfer must consume `MAX_TXS_PER_FORCED_INCLUSION` txs in a forced-inclusion request.
        // Note that the forced-inclusion request may contain fewer than `MAX_TXS_PER_FORCED_INCLUSION`
        // txs; the L2 execution will run the provided txs.  This rule is purely to prevent the
        // preconfer from bypassing the forced-inclusion mechanism with, e.g., `numTransactions = 0`.
```

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
dantaik: Using "<=" or "<" is logically incorrect, using ">=" is confusing, thus this change.

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
dantaik: Ensuring blocks are provable is very important if we need 100% ZK.

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
dantaik: Added a similar comment.

YoGhurt111 (approved):

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
linoscope: Yeah, makes sense. So the intuition is that instead of limiting the proving cost via the number of txs (which doesn't work anyway if few txs consume a lot of proving resources), we rely on the general mechanism of preventing proof-killing-blocks that is needed for 100% ZK.

However, one concern is whether we can really rely on the "general mechanism" for this in forced inclusion. For example, if the general mechanism is off-protocol tooling used in prover markets, that won't help here as there is no longer an off-protocol agreement taking place between the force-included batch submitter and the prover.

Several ways to address this:
- Enshrine the concept of "consumed ZK gas" in protocol, and let the protocol limit the prover cost by enabling things like "block is empty if it exceeds specified ZK gas".
  - Pros:
    - Opens doors to on-chain prover markets.
    -

linoscope: Yeah, makes sense. So the intuition is that instead of limiting the proving cost via the number of txs (which doesn't work anyway if few txs consume a lot of proving resources), we rely on the general mechanism of preventing proof-killing-blocks that is needed for 100% ZK.

However, one concern is whether we can really rely on the "general mechanism" in the context of forced inclusion. For example, if the general mechanism is off-protocol tooling used in prover markets, that won't help here as there is no longer an off-protocol agreement between the force-included batch submitter and the prover. Requiring the proposer to post a bond as done in https://github.com/taikoxyz/taiko-mono/pull/19584 won't help either, as the proposer (=preconfer) isn't the one who created the batches.

Without addressing this issue, there will be a DoS vector where a malicious entity continuously posts prover-killer blocks through forced inclusion and imposes costs on preconfers.

Several ways to address this come to mind:
| Option | Description | Pros | Cons |
|--------|-------------|------|------|
| **(A) Impose a separate gas limit for forced-inclusion batches** | Enable forced-inclusion anchor tx specifies its own (lower) gas ceiling. | • Simpler than option (B)<br> | • Gas is only a rough proxy for proving cost, so the limit will be conservative.<br>• Requires protocol changes for per-batch gas caps through anchor tx. |
| **(B) Track *ZK-gas* in-protocol and mark the block empty if it exceeds a cap** | Define a “ZK-gas” metric; forced-inclusion batches use a hard-coded, conservative cap. | • Precise—caps real proving cost, not gas usage.<br>• Also enables on-chain prover markets. | • Depends on the completion of the ZK-gas effort. |

I may be overlooking simpler solutions. We should also gauge the real-world severity of this attack. It looks serious, but need more thought. For example, the attacker’s ability to keep the attack going depends on the L1/L2 fees they must pay for every proof-killing block.

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
linoscope: Looks good!

packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
linoscope: Anyway, this issue is independent of this PR and a more general issue around forced inclusion itself. Created a separate issue to discuss this: https://github.com/taikoxyz/taiko-mono/issues/19618

linoscope (approved):
LGTM. Let's discuss the prover-killer mitigations in separate issue: https://github.com/taikoxyz/taiko-mono/issues/19618

adaki2004 (approved):

PR #19610, opened 2025-06-12
feat(protocol): tune shasta base fee parameters

dantaik:
Following the proposal in https://chainbound.notion.site/Taiko-Base-Fee-Analysis-2095abfafc1980199212ca4bdf792703

- change block gas limit to 32 million
- change adjustmentQuotient from 8 to 40 (a 5x)
- change gasIssuancePerSecond from 5 million to 1 million (a 1/5)
- maintain adjustmentQuotient* gasIssuancePerSecond constant
- change minGasExcess 1,344,899,430  to 1,440,000,000

Related to this PR, I also propose to cap the base fee to 2.5 gwei for now https://github.com/taikoxyz/taiko-mono/pull/19611


YoGhurt111 (approved):

johntaiko (approved):

adaki2004:
Side notes of mine:

After yesterday's meeting with Jeff and Brecht i'd (ideally) let the baseFee calculation as is.

On Etheruem it is there to throttle the traffic -> avoid too quick state growth ->  HW requirements issues -> decentralization, etc..

`gasIssuancePerSecond:` 1M gas/sec -> 12M gas/ 12 seconds, aka we below etheruem’s gas target (!).

I agree if we need a quick relief to sympthoms, might be a way, but like scaling down ethereum and then AFAIK 
`adjustmentQuotient from 8 to 40 (a 5x)` -> would mean that the `baseFee` changes 5x slower ! (not faster).

YoGhurt111:
We can upgrade this configuration when we are in Pacaya, is it nice to do?

dantaik:
> Side notes of mine:
> 
> After yesterday's meeting with Jeff and Brecht i'd (ideally) let the baseFee calculation as is.
> 
> On Etheruem it is there to throttle the traffic -> avoid too quick state growth -> HW requirements issues -> decentralization, etc..
> 
> `gasIssuancePerSecond:` 1M gas/sec -> 12M gas/ 12 seconds, aka we below etheruem’s gas target (!).
> 
> I agree if we need a quick relief to sympthoms, might be a way, but like scaling down ethereum and then AFAIK `adjustmentQuotient from 8 to 40 (a 5x)` -> would mean that the `baseFee` changes 5x slower ! (not faster).

It's slower from block to block, but given we have 2 second block time after preconf, the compounded change every 12 seconds is not as small.

PR #19611, opened 2025-06-12
feat(protocol): restrict base fee in a range [0.005 - 2.5 gwei]

dantaik:
Related to https://github.com/taikoxyz/taiko-mono/pull/19610


adaki2004 (approved):
Generally I'm OK with that, if we wanna go this way. (But then it obviously requires changes in `gasIssuancePerSecond` (aka the gasTarget of ours) and downplaying that variable is hmm, not sure if the right direction.) 
Still advocating for an off-chain min priority fee increase.

YoGhurt111 (approved):

PR #19615, opened 2025-06-12
Proposer bond forced inclusion

dantaik:


packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol
```diff
@@ -83,61 +74,74 @@ contract TaikoWrapper is EssentialContract, IProposeBatch {
...
+            IForcedInclusionStore.ForcedInclusion memory inclusion =
+                forcedInclusionStore.consumeOldestForcedInclusion(msg.sender);
+
+            params = _buildInclusionParams(inclusion, inclusionProverAuth);
+
+            (bool success, bytes memory returnData) = address(inbox).call(
+                abi.encodeCall(ITaikoInbox.v4ProposeBatch, (abi.encode(params), "", ""))
+            );
+
+            if (!success) {
```
dantaik: The user can deposit bond tokens into the inbox just before this transaction, making it succeed instead of fail. This prevents the preconfer from reliably predicting the transaction outcome.

PR #19625, opened 2025-06-16
chore(repo): add bridged token address form

RogerLamTd:


dantaik:
Why do we need this? Bridged tokens are deployed automatically by the first user cross-chain transfer.

RogerLamTd:
> Why do we need this? Bridged tokens are deployed automatically by the first user cross-chain transfer.

@jdcjonathan referred a case where  coingecko rejected a ticket for adding a dex listing due to the token address not being listed on docs, we can vet the contracts and tokens manually and add then to docs this way

dantaik (approved):

YoGhurt111 (approved):

PR #19626, opened 2025-06-16
feat(protocol): fix anchor reusal across-batches

adaki2004:
@dantaik in [this](https://github.com/taikoxyz/taiko-mono/pull/19601#discussion_r2142220392) comment you asked why this line is needed:

`require( lastBatch.anchorBlockId < lastAnchorBlockId, NoAnchorBlockIdWithinThisBatch());`

We merged the deletion of it but i still think it is incorrect and does not guard against this original requirement: `No two blocks may anchor to the same L1 block`.

We are protected from same anchor intra-batch but we are not protected from across-batch same L1 anchor.

So the edge-case is this:

```
BATCH: 100
and
BATCH: 101

Anchor to the same L1 height If both has 1 L2 blocks inside.
```

The test case is added (and if you remove the fix in `TaikoInbox.sol` you will see, the test goes thru successfully).


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -228,6 +228,11 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
                     NoAnchorBlockIdWithinThisBatch()
                 );
 
+                require(
```
dantaik: why not change line 208 to 

```solidity
anchorBlockId > lastBatch.anchorBlockId,
```

adaki2004: good catch, that result in the exact same thing! committed.

PR #19627, opened 2025-06-16
fix(protocol): re-use operator if consolidate was not called

cyberhorsey:
we can end up with multiple operators in the mapping `mapping(uint256 index => address operator) public operatorMapping;` and a single operator in the `mapping(address operator => OperatorInfo info) public operators;` if you add an operator, remove that operator, then re-add them without calling consolidate. 

@YoGhurt111 will need this PR cherry-picked for the taiko-alethia-devnet-v2.3.1 branch! will do after merge


YoGhurt111 (approved):

YoGhurt111:
CMIIW, I think current branch status is below:
client: `taiko-alethia-client-v1.5.1`
protocol: `taiko-alethia-protocol-v2.3.1`

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -172,18 +172,31 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
...
-        operators[_operator] = OperatorInfo({
-            activeSince: activeSince,
-            inactiveSince: 0, // no removal scheduled.
-            index: _operatorCount
-        });
-        operatorMapping[_operatorCount] = _operator;
-        unchecked {
-            operatorCount = _operatorCount + 1;
+        if (info.inactiveSince != 0) {
+            info.activeSince = activeSince;
```
dantaik:
```
info.activeSince = activeSince;
info.inactiveSince = 0;
```
these two lines are shared by ` if` and `else`, it can be moved out of the if-ese-statement.

dantaik (approved):

YoGhurt111 (approved):

YoGhurt111 (approved):

smtmfft (approved):

YoGhurt111 (approved):

PR #19629, opened 2025-06-17
feat(protocol): fix reusing old index for mapping

adaki2004:
(Should be cherry-picked to current preconf branch for hekla + mainnet candidates).

The bug:

1. Remove operator: Sets `activeSince = 0` and `inactiveSince = timestamp`, but keeps them in the mapping array
2. Re-add same operator: The check `require(info.activeSince == 0, OperatorAlreadyExists())` passes (because `activeSince` was reset to `0`)
2.1 New position created: Instead of reusing the existing position, a new slot is created at the end of the array

Bug: Operator appears at both their old position (marked inactive) AND new position (active)

( `consolidate()` did not fix duplicates just removed operators whose removal period has expired. Issue is that with `delay = 0`, the `inactiveSince= currentEpochTimestamp` , so the condition `inactiveSince <= currentEpochTimestamp` might not trigger removal immediately, depending on the exact timing.).


adaki2004:
Haven't seen @cyberhorsey PR, but i saw it just got merged so this one is obsolete, solving the same issue. Closing it.

PR #19648, opened 2025-06-20
fix(protocol): use `v4GetBaseFee` to verify base fee

YoGhurt111:


packages/protocol/contracts/layer2/based/anchor/ShastaAnchor.sol
```diff
@@ -124,4 +124,24 @@ abstract contract ShastaAnchor is PacayaAnchor {
             basefee_ = MIN_BASE_FEE;
         }
     }
+
+    /// @dev Verifies that the base fee per gas is correct and updates the gas excess.
+    /// @param _parentGasUsed The gas used by the parent block.
+    /// @param _baseFeeConfig The configuration parameters for calculating the base fee.
+    function _v4VerifyBaseFeeAndUpdateGasExcess(
```
dantaik, line 131: why not just use _verifyBaseFeeAndUpdateGasExcess?

YoGhurt111: Because we use `getBasefeeV2` in `_verifyBaseFeeAndUpdateGasExcess`, it doesn't contain the content of this PR https://github.com/taikoxyz/taiko-mono/pull/19611

adaki2004: Gavin i also think it is more readable to use simply the `_verifyBaseFeeAndUpdateGasExcess` and instead of creating a completely new wrapper function, just call the `v4GetBaseFee` in `_verifyBaseFeeAndUpdateGasExcess()` (instead of the `getBasefeeV2()` ). Wdyt ?

YoGhurt111: > Gavin i also think it is more readable to use simply the `_verifyBaseFeeAndUpdateGasExcess` and instead of creating a completely new wrapper function, just call the `v4GetBaseFee` in `_verifyBaseFeeAndUpdateGasExcess()` (instead of the `getBasefeeV2()` ). Wdyt ?

Because we also need `getBasefeeV2()` for Pacaya blocks in the meantime.
In the anchor contract, we need to include the code of the two forks before and after(Pacaya and Shasta), so that we can have a stable transition from Pacaya to Shasta.

adaki2004: > Because we also need `getBasefeeV2()` for Pacaya blocks in the meantime. In the anchor contract, we need to include the code of the two forks before and after(Pacaya and Shasta), so that we can have a stable transition from Pacaya to Shasta.


That is right, but the only addition (compared to Pacaya) is this check:

```
if (basefee_ > MAX_BASE_FEE) {
    basefee_ = MAX_BASE_FEE;
}
if (basefee_ < MIN_BASE_FEE) {
    basefee_ = MIN_BASE_FEE;
}
```


And we are (with hekla + mainnet) at the lowest level which is like 0.01 gwei, but in the code we set the MIN_BASE_FEE to 0.005 gwei.

image

So i think it is safe to do it like that, because what could happen ? Base fee goes lower to 0.005 gwei. But that is the "worst case" scenario. I dont have very strong opinion on that, but to me it would make sense not to introduce a new function, but your call guys!

YoGhurt111: We adjusted the base fee config in shasta fork, So it is possible to go lower to 0.005 in shasta fork?

adaki2004: > We adjusted the base fee config in shasta fork, So it is possible to go lower to 0.005 in shasta fork?

AFAIK it is still cannot g below 0.01 gwei, in which case this `if{}.. else {}` is harmless anyways (so 0 effect). I'm just giving my 0.02 that this new wrapper function is not needed but the call is yours.

YoGhurt111: It is indeed possible to move the base fee check to the Pacaya fork.
But we are not sure whether we want to enable this check in Pacaya. If we need to enable this check, we need to announce first. Because the minimum value will not be reached, the maximum value may be reached in theory, although it is difficult to reach the maximum value under the current tx situation.

YoGhurt111:
No need

PR #19664, opened 2025-06-26
feat(protocol): implement Shasta fork

### LINKED ISSUE #19706: Optimize calldata cost by packing function and event parameters
## Context

In Pacaya, many calldata and event fields contain unnecessary zero bytes. To reduce gas costs, we propose using packed byte representations for struct data in the `propose` and `prove` functions, as well as in related events.

## Idea

- Use packed `bytes` encoding (e.g., `abi.encodePacked`) for struct parameters in Inbox functions and emitted events.
- Emit compact data by encoding struct fields before emitting events.
- Avoid non-deterministic or aggressive compression rules. Use consistent, lossless packing so AI tools can:
  - Autogenerate encode/decode functions (Codec Libraries)
  - Autogenerate complete unit test coverage

## Development Guidelines

- **Protocol Baseline**  
  The core protocol implementation must use `abi.encode()` and `abi.decode()` for struct serialization and deserialization:
  - Ensures ABI compatibility
  - Provides a gas-cost reference baseline

- **Optimized Variant**  
  Optimized packing/decoding logic should be implemented separately and remain opt-in.

## Optional Enhancements

- Use custom Solidity annotations (e.g., `@custom:encode packed`) to tag structs and events for AI-assisted code generation and testing.
- Maintain a reusable AI prompt template that can regenerate encoders, decoders, and tests when a struct or event definition changes.

### LINKED ISSUE #19823: Enable a transition to cover multiple batches
## Context

In Pacaya, the protocol verifies a chain of L2 blocks using a linked list of transitions. Each batch corresponds to a transition, where the transition represents a link from:

```
<parent batch's last block hash> → <current batch's last block hash>
```

This means proving `x` consecutive batches requires creating `x` transitions.

## Idea

If a prover proves multiple consecutive batches (e.g., `x` batches), we can reduce overhead by creating **a single transition** that spans them all:

```
<first batch's last block hash> → <last batch's last block hash>
```

Note: batches proved in one transaction are consecutive and their transitions are linked

## Benefits

- Fewer `SLOAD` and `SSTORE` operations during verification
- Reduced onchain state size and gas costs

### LINKED ISSUE #19824: Remove all inbox wrapper contracts
In the current Shasta design, the `Inbox` contract is wrapped by `TaikoWrapper`, which is itself wrapped by `PreconfRouter`. This layering introduces ambiguity in how `msg.sender` is interpreted at each level. As a result, permission logic and forced inclusion handling become harder to reason about.

The idea is to remove the wrappers and allows the `Inbox` to directly call into third-party contracts to perform permission checks and handle forced inclusion. This ensures that `msg.sender` within the `Inbox` contract reliably reflects the actual proposer.

To further discourage wrapping the `Inbox`, we may also remove return values from the `propose` and `prove` functions. This reduces the incentive or need for other contracts to wrap it for inspection or extension.

Overall, this change improves clarity, maintainability, and encourages a cleaner contract architecture.

## Hooks

We could also consider adding hooks to the Inbox contract, allowing proposers to call their own contracts for additional validation. However, this may not be strictly necessary.

### LINKED ISSUE #19829: Remove anchor transaction and TaikoAnchor entirely
## Context

With the plan to [deprecate `taiko-geth`](https://github.com/taikoxyz/taiko-mono/issues/19830) and adopt `taiko-reth` for Unzen, the current anchor transaction and `TaikoAnchor` contract design becomes obsolete.

## Idea

We should remove both the anchor transaction mechanism and the `TaikoAnchor` contract. Instead, the necessary protocol data (e.g., block hashes) should be stored and accessed via `taiko-reth`'s **system call**.

The goal is to:

- Store all historical L2 block hashes at the execution layer level
- Expose these hashes to a predefined contract that implements the `IBlockHashProvider` interface

Open question: should `IBlockHashProvider` be implemented as a **precompile**, or a **regular system contract** that is only modifiable via the system call? @davidtaikocha

dantaik:
## New Contract Architecture

- Created `layer1/based2/` directory with completely new inbox implementation  
- Introduced `AbstractInbox.sol` as the main contract with modular function pointer pattern via `LibBinding`  
- Split into focused interfaces: `IInbox`, `IPropose`, `IProve`, `IBondManager2` with versioned functions (suffix `4`)  
- Removed tier-based proof system and contestation mechanisms in favor of a single-proof model  
- ℹ️ **Missing: In theory, many calldata parameters used for proposing batches can also be moved into blobs. While this PR doesn’t introduce such fundamental changes, they are being explored in [PR 19839](https://github.com/taikoxyz/taiko-mono/issues/19838)](https://github.com/taikoxyz/taiako-mono/pull/19839.)**

## Data Structure Overhaul

- Switched from individual blocks to batch-based processing with new `Batch`, `BatchContext`, and `BatchMetadata` structs  
- Optimized storage with `TransitionState` packing `batchId` and partial `parentHash` into a single slot  
- Implemented ring buffer pattern for batch/transition storage (configurable sizes)  
- Reduced storage operations: proposing reduced from 4 `SLOAD` + 3 `SSTORE` to 1 `SLOAD` + 2 `SSTORE`  

## Library Reorganization

- `libs/`: Core logic libraries (`LibPropose`, `LibProve`, `LibVerify`, `LibValidate`, `LibProver`, `LibForks`)  
- `codec/`: 7 encoding/decoding libraries for efficient data serialization  
- `state/`: Storage management (`LibState`, `LibBonds`) with optimized read/write operations  
- Function pointer pattern allows swapping implementations without contract upgrades  

## Gas Optimization Techniques

- Bit-packing in `HeaderExtraInfo` (sharing percentage, forced inclusion flag, gas issuance) into 128 bits  
- Skipped calldata emission for direct EOA calls using `_isOuterMostTransaction()`  
- Reused transition slots, saving ~3,900 gas per batch  
- Stored only the summary hash on-chain instead of the full `Summary` struct  
- Hierarchical batch metadata hashing for efficient verification  
- Some consequtive batches can end up with one single storage write when proved together.

## Technical Implementation Changes

- New entry points: `propose4(bytes)`, `prove4(bytes, bytes)` for batch operations  
- Simplified verifier interface: `verifyProof(bytes32 aggregatedProvingHash, bytes proof)`  
- Asynchronous block proposals (proofs not required at proposal time)  
- Batch verification supports processing multiple batches in a single transaction  
- Event changes: emit minimal data and skip redundant fields to reduce gas usage


packages/protocol/contracts/layer1/based2/TaikoInbox2.sol
```diff
@@ -0,0 +1,293 @@
...
+    // Define errors locally
+    error ContractPaused();
+
+    // External functions ------------------------------------------------------------------------
+
+    constructor(
+        address _inboxWrapper,
+        address _verifier,
+        address _signalService
+    )
```
adaki2004: `bondToken` ?

packages/protocol/contracts/layer1/based2/ITaikoInbox2.sol
```diff
@@ -0,0 +1,270 @@
...
+        /// @notice Historical heights of the forks.
+        ForkHeights forkHeights;
+        /// @notice The token used for bonding.
+        address bondToken;
+    }
+
+    /// @notice Struct holding the state variables for the {Taiko} contract.
+    struct State {
+        // Ring buffer for proposed batches and a some recent verified batches.
+        mapping(uint256 batchId_mod_batchRingBufferSize => bytes32 metaHash) batches;
```
adaki2004: Previously we stored the batches, now the hash of the batch from on-chain perspective, it would be a great saving, but questions:
- how and who will store these batches (data) - which are required for proving ? Prover needs to store and reconstruct metadata off-chain, BUT some part of the metadata is decided on-chain (like `proposer, coinbase, createdIn`, etc.) so basically it would make the off-chain proving much more complicated.

I'm not saying it is impossible to figure out but it definitely creates an extra workload for the off-chain stuff. On-chain cost wise i think it is a good reduction but it would require heavy and significant redesign of the node (off-chain data storage,  reconstruct logic, possible to reconstruct on-the fly, etc.)

(if the off-chain storage is -for whatever reason- unavailable it would make proving impossible.)

adaki2004: Also this way it is incompatiible with the underlying data storage structure. These mappings are just pointers indeed but the real incompatibility isn't the mapping declaration itself, but the data stored at those locations.

After upgrade, reading data, `TaikoInbox2` trying to read a `bytes32` from a location that possibly contains a multi-slot `Batch` struct.

dantaik: @smtmfft could you share your thoughts on this approach from a proving perspective? Even if it makes proving more difficult, I believe we should still aim to minimize onchain costs to maximize profitability.

@adaki2004 please pay attention to the BatchProveMetadata,  BatchProposeMetadata, TransitionMeta structs to understand how proving and verification works with the new design.

dantaik: As long as the layout is compatible, the data types don’t matter. Remember, we use ForkRouter to route different batches to different fork implementations. The only additional work required is in handling the verification between the last batch of the previous fork and the first batch of the new fork - this needs some extra code.

packages/protocol/.claude/settings.local.json
```diff
@@ -0,0 +1,17 @@
+{
```
linoscope: Is this file meant to be committed to repo?

packages/protocol/contracts/layer1/based2/ITaikoInbox2.sol
```diff
@@ -0,0 +1,279 @@
...
+
+    //  @notice Struct representing transition storage
+    /// @notice 2 slots used for each transition.
+    struct TransitionState {
+        uint256 batchIdAndPartialParentHash;
+        bytes32 metaHash;
+    }
+
+    struct Summary {
+        uint48 numBatches;
```
linoscope: What is the intuition of this `numBatches`? Is it the total number of batches ever proposed? Would be good to comment since it's not clear from just the name.

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
```diff
@@ -0,0 +1,381 @@
...
+            require(anchorBlockHashes_[k] != 0, ZeroAnchorBlockHash());
+
+            anchorFound = true;
+            lastAnchorBlockId_ = anchorBlockId;
+            k++;
+        }
+
+        // Ensure that if msg.sender is not the inboxWrapper, at least one block must
+        // have a non-zero anchor block id. Otherwise, delegate this validation to the
+        // inboxWrapper contract.
```
linoscope: Where in the "inboxWrapper" is this check done? Couldn't find in `TaikoWrapper` or `PreconfRouter` but maybe there are more "wrappers". Also, wonder if such check should even be done in the wrapper.

packages/protocol/.claude/settings.local.json
dantaik: Will remove.

dantaik:
@AnshuJalan @linoscope @adaki2004 @cyberhorsey @smtmfft @YoGhurt111

I think I'm done with this PR in terms of features. Tests are not to be done and I hope to add them in separate small PRs later. I hope you guys can find time to give it a design review.

This PR assumes L2 base fee is implemented in node/driver using EIP-4396 (see https://github.com/taikoxyz/taiko-mono/issues/19725) and each batch can change the gas issuance per second value and the value will be passed into L2 block header's extraData.

❤️🍻

packages/protocol/contracts/layer1/based2/ITaikoInbox2.sol
dantaik: renamed to nextBatchId

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
dantaik: fixed comment.

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
```diff
@@ -0,0 +1,405 @@
...
+        uint256 _numBlocks,
+        uint48 _parentLastBlockId
+    )
+        internal
+        pure
+        returns (uint48)
+    {
+        unchecked {
+            uint256 firstBlockId = _parentLastBlockId + 1;
+            uint256 lastBlockId = uint48(firstBlockId + _numBlocks);
```
davidtaikocha:
```suggestion
            uint256 lastBlockId = uint48(_parentLastBlockId + _numBlocks);
```

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,315 @@
...
+        bytes32 summaryHash; // slot 4
+        bytes32 __deprecatedStats1; // slot 5
+        bytes32 __deprecatedStats2; // slot 6
+        mapping(address account => uint256 bond) bondBalance;
+        uint256[43] __gap;
+    }
+
+    /// @notice Emitted when a batch is synced.
+    /// @param summary The Summary data structure.
+    event SummaryUpdated(Summary summary);
```
davidtaikocha: Looks like this event is not in use?

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,350 @@
...
+    )
+        external
+        initializer
+    {
+        _init(_owner, _genesisBlockHash, _gasIssuancePerSecond);
+    }
+
+    /// @inheritdoc IPropose
+    function propose4(
+        bytes calldata _packedSummary,
```
davidtaikocha: Seems after the Shasta fork, a proposer has to assemble the latest `Summary` off-chain, but how does a new proposer know some fields like `gasIssuanceUpdatedAt` under the current ABI? Since `Summary` is not saved on-chain anymore, and `event SummaryUpdated(Summary summary);` is not in use. Maybe we should at least emit the `SummaryUpdated` event each time after `saveSummaryHash`? (and also save `summaryUpdatedAt` on-chain? which will make the proposer to fetch the event efficiently)

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,350 @@
...
+        LibState.Access memory access = _getReadWrite();
+
+        // Prove batches and get aggregated hash
+        bytes32 aggregatedBatchHash;
+        (summary, aggregatedBatchHash) = LibProve.prove(config, access, summary, inputs);
+
+        // Verify the proof
+        IVerifier2(config.verifier).verifyProof(aggregatedBatchHash, _proof);
+
+        _saveSummaryHash(keccak256(abi.encode(summary)));
```
davidtaikocha: Seems this is unnecessary here and we should also always avoid updating `summary` in `prove4`, right? Otherwise, if two provers want to prove their batches in the same L1 block, all with correct proofs, the second one will still revert.

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
davidtaikocha: Another side-effect which may be worth mentioning is that this design will enforce all proposers / provers to enable `revert-protection` by default. And separated proposing / proving transactions can't be landed in one same L1 block anymore. Not sure if we want this.

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,350 @@
...
+        I.TransitionMeta[] memory transitionMetas = LibCodec.unpackTransitionMetas(_packedTrans);
+
+        I.Config memory config = _getConfig();
+        LibState.Access memory access = _getReadWrite();
+
+        // Propose batches
+        summary = LibPropose.propose(config, access, summary, batches, evidence);
+
+        // Verify batches
+        summary = LibVerify.verify(config, access, summary, transitionMetas);
```
davidtaikocha, line 79: From the perspective of proposing profitability, I guess under this design, probly nobody will be willing to provide any `transitionMetas`... which will squeeze their off-chain block-building time window, and give them extra `LibCodec.unpackTransitionMetas` + `LibVerify.verify` gas cost on-chain...

davidtaikocha: Before post-whitelist preconf, this might be a little bit tricky for proposers because of the `_config.cooldownWindow` and they don't know when their transactions will be landed on L1, even for `12` secs difference, which may still lead to the revert of their `propose4` transactions. Since if they want to save money, they want to provide the exact transitions which will be verified...right?

YoGhurt111:
Looks like we should develop a offchain storage to save the `Summary` specific fields, and proposers and provers should know the correct `Summary`?

davidtaikocha:
> Looks like we should develop a offchain storage to save the `Summary` specific fields, and proposers and provers should know the correct `Summary`?

Maybe not, if we can utilize the `SummaryUpdated ` event.

packages/protocol/contracts/layer1/based2/IInbox.sol
dantaik: fixed

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: True, this new protocol is designed assuming preconfirmation is in place and at an given time, a known preconfer is the only one who can propose.

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: Very true. Fixed.

dantaik:
> Looks like we should develop a offchain storage to save the `Summary` specific fields, and proposers and provers should know the correct `Summary`?

Summary is emitted in SummaryUpdated event.

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: This protocol is designed with preconfirmation as a precondition, if the preconfer is uncertain when his tx will land on L1, revert protection must be enabled to avoid tx revert.

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: I'm assuming preconf is already enabled when this protocol is used.

davidtaikocha:
> > Looks like we should develop a offchain storage to save the `Summary` specific fields, and proposers and provers should know the correct `Summary`?
> 
> Summary is emitted in SummaryUpdated event.

Another point about the `SummaryUpdated` event ergonomics is, shall we also save `summaryUpdatedAt` in the storage? Otherwise, at least all newly started provers / proposers have to scan back from the current L1 chain tip for the event, which is inefficient.

YoGhurt111:
> > Looks like we should develop a offchain storage to save the `Summary` specific fields, and proposers and provers should know the correct `Summary`?
> 
> Summary is emitted in SummaryUpdated event.

If so, when we p2p sync a new node, we need to wait a new SummaryUpdated event coming, then we can start to propose or prove.

YoGhurt111:
> > Looks like we should develop a offchain storage to save the `Summary` specific fields, and proposers and provers should know the correct `Summary`?
> 
> Maybe not, if we can utilize the `SummaryUpdated ` event.

If so, when we p2p sync a new node, we need to wait a new SummaryUpdated event coming, then we can start to propose or prove.

davidtaikocha:
> > > Looks like we should develop a offchain storage to save the `Summary` specific fields, and proposers and provers should know the correct `Summary`?
> > 
> > 
> > Maybe not, if we can utilize the `SummaryUpdated ` event.
> 
> If so, when we p2p sync a new node, we need to wait a new SummaryUpdated event coming, then we can start to propose or prove.

No need to wait anyway, we can do this: https://github.com/taikoxyz/taiko-mono/pull/19664#issuecomment-3071340539

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,347 @@
...
+        I.Config memory config = _getConfig();
+        LibState.Access memory access = _getReadWrite();
+
+        // Propose batches
+        summary = LibPropose.propose(access, config, summary, batches, evidence);
+
+        // Verify batches
+        summary = LibVerify.verify(access, config, summary, transitionMetas);
+
+        bytes memory packedSummary = LibCodec.packSummary(summary);
```
davidtaikocha: Shall we also re-expose those `LibCodec.packXXX` methods in `AbstractInbox`? Seems not a big cost in smart contracts, but will be a time saver for all client implementers who need to call `propose4` and `prove4`, and will also make their software decoupled with the detailed codec impl.

packages/protocol/contracts/layer1/based2/libs/LibPropose.sol
```diff
@@ -0,0 +1,136 @@
...
+        I.Batch[] memory _batches,
+        I.BatchProposeMetadataEvidence memory _evidence
+    )
+        internal
+        returns (I.Summary memory)
+    {
+        unchecked {
+            require(_batches.length != 0, EmptyBatchArray());
+
+            // Make sure the lask verified batch is not overwritten by a new batch.
```
davidtaikocha:
```suggestion
            // Make sure the last verified batch is not overwritten by a new batch.
```

packages/protocol/contracts/layer1/based2/libs/LibPropose.sol
```diff
@@ -0,0 +1,136 @@
...
+            // Assuming batchRingBufferSize = 100, right after genesis, we can propose up to 99
+            // batches, the following requirement-statement will pass as:
+            //  1 (nextBatchId) + 99 (_batches.length) <=
+            //  0 (lastVerifiedBatchId) + 100 (batchRingBufferSize)
+            require(
+                _summary.nextBatchId + _batches.length
+                    <= _summary.lastVerifiedBatchId + _config.batchRingBufferSize,
```
davidtaikocha: If we've decided to [remove the support of `forcedInclusion`](https://github.com/taikoxyz/taiko-mono/pull/19734), seems then I can't find a use case where any proposer will need to propose multiple batches in one transaction, a simple batch is enough to utilize all available blobs? And the proposer will need to pay more `livenessBond` for that. Correct me if I'm wrong...

packages/protocol/contracts/layer1/based2/libs/LibData.sol
```diff
@@ -0,0 +1,153 @@
...
+    /// @notice Encodes configuration and batch information into the lower 128 bits
+    /// @dev Bit-level encoding for efficient storage:
+    ///      - Bits 0-7: Base fee sharing percentage (0-100)
+    ///      - Bit 8: Forced inclusion flag (0 or 1)
+    ///      - Bits 9-40: Gas issuance per second (32 bits)
+    ///      - Bits 41-127: Reserved for future use
+    /// @param _baseFeeSharingPctg Base fee sharing percentage (0-100)
+    /// @param _batch Batch information containing forced inclusion flag and gas issuance
+    /// @return Encoded data as bytes32 with information packed in lower 128 bits
+    function _encodeExtraData(
```
davidtaikocha: Another consensus change that we can discuss here for the next hardfork is, shall we change to decode `parent.extraData` (instead of `currentBlock.extraData`) to calculate the current block's `basefee`? Which will make the base fee deterministic for block building and make our network compatible with Ethereum's basefee mechanism.

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+    /// @dev Contains block-specific parameters and anchor information
+    struct Block {
+        /// @notice Maximum number of transactions in this block
+        /// @dev If insufficient transactions in calldata/blobs, block contains as many as possible
+        uint16 numTransactions;
+        /// @notice Time difference in seconds between this block and its parent within the batch
+        /// @dev For the first block in a batch, this should be 0 (no parent in same batch)
+        uint8 timeShift;
+        /// @notice Optional anchor block ID for L1-L2 synchronization
+        uint48 anchorBlockId;
```
YoGhurt111, line 20: Seems it's not used, and already defined externally in https://github.com/taikoxyz/taiko-mono/blob/e9dcfcaa9ffeefed31af006f734e0dd0a37462c2/packages/protocol/contracts/layer1/based2/IInbox.sol#L64

packages/protocol/contracts/layer1/based2/libs/LibVerify.sol
```diff
@@ -0,0 +1,148 @@
...
+        I.Summary memory _summary,
+        I.TransitionMeta[] memory _trans
+    )
+        internal
+        returns (I.Summary memory)
+    {
+        unchecked {
+            uint48 batchId = _summary.lastVerifiedBatchId + 1;
+
+            if (!LibForks.isBlocksInCurrentFork(_config, batchId, batchId)) {
```
YoGhurt111: We should use blockId to determine fork, not batchId.

packages/protocol/contracts/layer1/based2/libs/LibVerify.sol
```diff
@@ -0,0 +1,148 @@
...
+            uint256 nTransitions = _trans.length;
+
+            for (; batchId < stopBatchId; ++batchId) {
+                (bytes32 tranMetaHash, bool isFirstTransition) =
+                    _access.loadTransitionMetaHash(_config, _summary.lastVerifiedBlockHash, batchId);
+
+                if (tranMetaHash == 0) break;
+
+                require(i < nTransitions, TransitionNotProvided());
+                require(tranMetaHash == keccak256(abi.encode(_trans[i])), TransitionMetaMismatch());
```
YoGhurt111: I think the proposer prefers to verify only its own batch, but it seems that the check here will cause it to revert if the next batch to be verified is not its own batch. What is the design idea here?

YoGhurt111: I understand that the core idea is to save on-chain storage costs.
But there seems to be a lack of incentives for verification

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+        uint48 lastBlockId;
+        /// @notice Bond amount for liveness guarantee in Gwei
+        uint48 livenessBond; // Gwei
+        /// @notice Bond amount for provability guarantee in Gwei
+        uint48 provabilityBond; // Gwei
+    }
+
+    /// @notice Complete metadata for a batch combining all phases
+    /// @dev Aggregates build, propose, and prove metadata for batch processing
+    struct BatchMetadata {
```
YoGhurt111, line 195: There seems to be no proposed metadata information in either event  `SummaryUpdated` or event `Proposed`,it will cause the block to be unable to be produced.
eg. How can we calculate the L2 block timestamp without `lastBlockTimestamp` and `timeShift`?

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,347 @@
...
+
+        return summary;
+    }
+
+    /// @inheritdoc IProve
+    /// @dev In prevous versions, proving a block may also trigger block verification, in this
+    /// upgrade, this is no longer the case as we would like to ensure more certainty for provers
+    /// and let proposers to manage the uncertaity of verification cost.
+    function prove4(
+        bytes calldata _packedBatchProveInputs,
```
YoGhurt111: Can we add `BatchMetadata` in event `SummaryUpdated` or event `Proposed`?  I think it would be great helpful to generate `_packedBatchProveInputs`.

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+/// @notice Interface for the Taiko Alethia protocol inbox
+/// @dev Defines all data structures and function signatures for the simplified
+///      based rollup protocol without tier-based proof system
+/// @custom:security-contact security@taiko.xyz
+interface IInbox {
+    /// @notice Represents a block within a batch
+    /// @dev Contains block-specific parameters and anchor information
+    struct Block {
+        /// @notice Maximum number of transactions in this block
+        /// @dev If insufficient transactions in calldata/blobs, block contains as many as possible
```
ggonzalez94: I'm not sure I fully understand the comment.
- since the proposer is in control of the block they can know the exact number of txs. Why is this the max and not just the actual number of tx the block has?
- I wasn't able to understand what the insufficient txs in calldata/blobs mean in this context

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+        uint16 numTransactions;
+        /// @notice Time difference in seconds between this block and its parent within the batch
+        /// @dev For the first block in a batch, this should be 0 (no parent in same batch)
+        uint8 timeShift;
+        /// @notice Optional anchor block ID for L1-L2 synchronization
+        uint48 anchorBlockId;
+        /// @notice Number of cross-chain signals in this block
+        uint8 numSignals;
+        /// @notice Whether this block references an anchor block
+        bool hasAnchor;
```
ggonzalez94: can't this be deduced by the fact that anchorBlockId is different than zero?

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    /// @notice Contains blob data for a batch
+    /// @dev Supports both direct blob hashes and blob indices for different scenarios
```
ggonzalez94, line 31: maybe I should know this, but why do we need both? what are the different scenarios?

ggonzalez94: I imagine forced inclusions?

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+        /// @notice Gas issuance rate per second for this batch
+        uint32 gasIssuancePerSecond;
+        /// @notice Whether this is a forced inclusion batch
+        bool isForcedInclusion;
+        /// @notice Prover authorization data
+        bytes proverAuth; // length <= type(uint16).max
+        /// @notice Signal slots for cross-chain messages
+        bytes32[] signalSlots; // length <= type(uint8).max
+        /// @notice Anchor block IDs for L1-L2 synchronization
+        uint48[] anchorBlockIds; // length <= type(uint16).max
```
ggonzalez94: why we need this array if we already have the anchors in the `Block` struct?

packages/protocol/contracts/layer1/based2/Inbox.sol
```diff
@@ -0,0 +1,197 @@
...
+pragma solidity ^0.8.24;
+
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "src/shared/signal/ISignalService.sol";
+import "src/shared/signal/LibSignals.sol";
+import "./AbstractInbox.sol";
+import "./libs/LibBonds.sol";
+import "./IBondManager2.sol";
+
+/// @title TaikoInbox
```
ggonzalez94: nit to match the name:

```suggestion
/// @title Inbox
```

packages/protocol/contracts/layer1/based2/Inbox.sol
```diff
@@ -0,0 +1,197 @@
...
+/// original Taiko-Based Contestable Rollup (BCR) but with the tier-based proof system and
+/// contestation mechanisms removed.
+///
+/// Key assumptions of this protocol:
+/// - Block proposals and proofs are asynchronous. Proofs are not available at proposal time,
+///   unlike Taiko Gwyneth, which assumes synchronous composability.
+/// - Proofs are presumed error-free and thoroughly validated, with subproofs/multiproofs management
+/// delegated to IVerifier contracts.
+///
+/// @dev Registered in the address resolver as "taiko".
```
ggonzalez94: nit: all of this is already documented in `AbstractInbox`. Maybe we just pick one place and don't repeat. 
I would suggest we just put:
`/// @inheritdoc AbstractInbox` here

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,347 @@
...
+
+    // -------------------------------------------------------------------------
+    // External Functions
+    // -------------------------------------------------------------------------
+
+    /// @notice Initializes the contract with owner and genesis block hash
+    /// @param _owner The owner address
+    /// @param _genesisBlockHash The genesis block hash
+    /// @param _gasIssuancePerSecond The initial gas issuance per second
+    function init4(
```
ggonzalez94, line 42: Just out of curiosity: do we still need to suffix the functions with `4`? I imagine we did this when we had multiple versions of the same function in the same contract

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,347 @@
...
+            LibCodec.unpackTransitionMetas(_packedTransitionMetas);
+
+        I.Config memory config = _getConfig();
+        LibState.Access memory access = _getReadWrite();
+
+        // Propose batches
+        summary = LibPropose.propose(access, config, summary, batches, evidence);
+
+        // Verify batches
+        summary = LibVerify.verify(access, config, summary, transitionMetas);
```
ggonzalez94: I know we are now doing this as well, but I don't think I fully appreciate the idea that verification is forced on proposing for a few reasons:
- it makes the cost of proposing unpredictable(depending on whether you need to verify or not and the cost of that run your proposing costs will change)
- it makes the calls to verify perhaps more frequent than necessary(propose is called at most once per epoch, and usually more often) while zk proofs will usually land less frequently
- it makes preconfers potentially less profitable, something we want to avoid especially at the early stages of preconfs

Doesn't it make more sense to move the call to verify to:
A) `prove`: which is called less frequently and can be subsidized if necessary
B) as a standalone function. Since there's no incentive to call it we can make a requirement in `prove`for example that you cannot keep pushing validity proofs if you don't call verify and finalize the chain. The same can be done with propose if we prefer to pass the burden to proposers.
This makes it cleaner, and batching can still be achieved with a multicall or 7702 address so that they don't have to actually be two txs

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+    /// @notice Evidence structure for batch proposal metadata validation
+    /// @dev Contains hashes and metadata needed to verify batch proposals
+    struct BatchProposeMetadataEvidence {
+        /// @notice Left hash for merkle proof verification
+        bytes32 leftHash;
+        /// @notice Hash of the prove metadata
+        bytes32 proveMetaHash;
+        /// @notice Proposal metadata to be validated
+        BatchProposeMetadata proposeMeta;
+    }
```
ggonzalez94: I'm still trying to wrap my head around this struct, but if I understand correctly we use it to validate the parent block of the batch. If that's the case maybe we can have more meaningful name and comments. Let me try to propose something(and feel free to only take the parts that make sense to you):


```suggestion
   /// @notice Reference to a parent batch for chain continuity validation
    /// @dev When proposing new batches, we need to prove knowledge of the parent batch
    ///      while also accessing specific parent data for validation. This structure
    ///      optimizes gas usage by providing:
    ///      1. Pre-computed hashes to verify the parent batch (avoiding 200+ bytes of data)
    ///      2. The minimal parent data actually needed for validation (18 bytes)
    ///      
    ///      Parent Batch Hash Structure:
    ///                      parentBatchHash
    ///                      /              \
    ///            parentContentHash    parentRightHash
    ///            (passed as-is)       /              \
    ///                          proposeMetaHash   parentProveHash
    ///                          (computed from     (passed as-is)
    ///                           proposeMeta)
    ///
    struct ParentBatchReference {
        /// @notice Pre-computed hash of (parentBatchId + parentBuildMetaHash)
        /// @dev Contains the parent's content data (transactions, blobs, etc.) which
        ///      we don't need for validation, so we just pass its hash
        bytes32 parentContentHash;
        /// @notice Pre-computed hash of parent's prove metadata
        /// @dev Contains the parent's timing and bond data which we also don't need,
        ///      so we just pass its hash
        bytes32 parentProveHash;
        /// @notice The parent's actual proposal metadata
        /// @dev This is the only parent data we actually need for validation
        BatchProposeMetadata proposeMeta;
    }
```

ggonzalez94 (commented):
Lots more to review, but here are a few questions and suggestions. Sorry if a lot of this has already been discussed in the past, you can just provide a quick answer or link to the right place

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+    /// @dev Contains all data needed for batch validation and processing
+    struct Batch {
+        /// @notice Address that proposed this batch
+        address proposer;
+        /// @notice Coinbase address for block rewards (can be zero)
+        address coinbase;
+        /// @notice Timestamp of the last block in this batch
+        uint48 lastBlockTimestamp;
+        /// @notice Gas issuance rate per second for this batch
+        uint32 gasIssuancePerSecond;
```
ggonzalez94, line 63: why does a proposer has the ability to pick the `gasIssuancePerSecond`? isn't this a protocol defined rule?

ggonzalez94: oh, this is part of the proposed new mechanism to allow proposers to have influence over the gas limit and gas target right? I'm still hesitant of how that would play out, but that's a separate conversation

packages/protocol/contracts/layer1/based2/libs/LibVerify.sol
```diff
@@ -0,0 +1,148 @@
...
+                emit I.Verified(batchId << 48 | _trans[i].lastBlockId, _trans[i].blockHash);
+                _summary.lastVerifiedBlockHash = _trans[i++].blockHash;
+            }
+
+            if (lastSyncedBatchId != 0) {
+                _summary.lastSyncedAt = uint48(block.timestamp);
+                _summary.lastSyncedBlockId = _trans[lastSyncedBatchId].lastBlockId;
+                _access.syncChainData(
+                    _config, _summary.lastSyncedBlockId, _trans[lastSyncedBatchId].stateRoot
+                );
```
ggonzalez94: I couldn't find a place where we are updating `_summary.lastVerifiedBatchId`, I think this is a bug and we should add something like:

```sol
if (batchId > _summary.lastVerifiedBatchId + 1) {
  _summary.lastVerifiedBatchId = uint48(batchId - 1);
}
```

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
ggonzalez94: I actually just saw the comment on the `prove` function that we explicitely want to pass this uncertainty to proposers, curious to understand why this is a good idea

packages/protocol/contracts/layer1/based2/libs/LibPropose.sol
ggonzalez94: Even with `forcedInclusions` isn't there a way to just include them in the same batch(i.e. appending the forced included blocks at the end)? It does seem that only supporting one batch at a time would be simpler, easier to reason about and more gas efficient

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
```diff
@@ -0,0 +1,391 @@
...
+    // -------------------------------------------------------------------------
+    // Private Functions
+    // -------------------------------------------------------------------------
+
+    /// @notice Validates the proposer
+    /// @dev Handles both direct proposing and inbox wrapper scenarios
+    /// @param _config Protocol configuration
+    /// @param _batch The batch being validated
+    function _validateProposer(I.Config memory _config, I.Batch memory _batch) internal view {
+        if (_config.inboxWrapper == address(0)) {
```
ggonzalez94: Do we need to keep supporting the path with and without and `inboxWrapper`? If this version of the protocol has the assumption that it always works with preconfs I think we should remove some of this conditional logic, and potentially unify a lot of the peripheral contracts(if you agree I can experiment with a PR that does this)

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
```diff
@@ -0,0 +1,391 @@
...
+        }
+    }
+
+    /// @notice Validates the gas issuance per second for a batch
+    /// @dev Ensures that the gas issuance per second is within a 1% range of the last recorded
+    ///      value
+    /// @param _config The protocol configuration
+    /// @param _summary The current protocol summary
+    /// @param _batch The batch being validated, which includes the gas issuance per second
+    function _validateGasIssuance(
```
ggonzalez94, line 112: Like I said I'm not convinced on the idea of proposers having the option to influence gas issuance on the protocol(not sure the incentives are correct for them to reflect a true cost of the L1, and I think the answer varies widely depending on how often you expect you will propose, and it is a relation that is hard to reason about). But regardless of this, can this logic be moved to the node to the derivation pipeline? I don't see a strong reason why this should happen on-chain

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
```diff
@@ -0,0 +1,391 @@
...
+            require(firstBlockTimestamp_ >= _parentLastBlockTimestamp, TimestampSmallerThanParent());
+        }
+    }
+
+    /// @notice Validates cross-chain signals in the batch
+    /// @dev Verifies that all referenced signals have been properly sent
+    /// @param _access Read/write access functions
+    /// @param _config Protocol configuration
+    /// @param _batch The batch containing signal references
+    function _validateSignals(
```
ggonzalez94, line 187: this is for same slot message passing mechanism and not for regular signals right? I haven't really analyzed the full gas implications for what may not be used very frequently(we might be paying the cost of this validation on every batch while this is used very infrequently). Maybe @nikeshnazareth provable assertion mechanism can be more efficient here

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
```diff
@@ -0,0 +1,391 @@
...
+            hasAnchorBlock = true;
+            lastAnchorBlockId_ = anchorBlockId;
+            anchorIndex++;
+        }
+
+        // Ensure that if msg.sender is not the inboxWrapper, at least one block must
+        // have a non-zero anchor block id.
+        if (_config.inboxWrapper != address(0)) {
+            require(hasAnchorBlock, NoAnchorBlockIdWithinThisBatch());
+        }
```
ggonzalez94: same here: I think we can operate under the assumption that we already have preconfs and there's always a wrapper

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
```diff
@@ -0,0 +1,391 @@
...
+            // Validate and decode blocks
+            require(_numBlocks != 0, BlockNotFound());
+            uint48 firstBlockId = _parentLastBlockId + 1;
+            uint48 lastBlockId = uint48(_parentLastBlockId + _numBlocks);
+            require(
+                LibForks.isBlocksInCurrentFork(_conf, firstBlockId, lastBlockId),
+                BlocksNotInCurrentFork()
+            );
+            return lastBlockId;
+        }
```
ggonzalez94: I wonder if some of this logic(and more broadly some of the validations in the propose function) can be moved off-chain to the node.
cc @YoGhurt111 since you may know better

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+        Blobs blobs;
+    }
+
+    /// @notice Output structure containing validated batch information
+    /// @dev This struct aggregates all validation results for efficient batch processing
+    struct BatchContext {
+        /// @notice Address authorized to prove this batch
+        address prover;
+        /// @notice Hash of all transactions in the batch
+        bytes32 txsHash;
```
ggonzalez94: I think this is actually the hash of all blob hashes and not a hash of txs right?

```suggestion
        /// @notice Hash of all blob hashes in the batch
        bytes32 blobsHash;
```

ggonzalez94 (commented):
Spent some more time on this today and left a few more comments and suggestions. Nothing major, except for one potential bug

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
YoGhurt111: We currently have validations for these parameters, and we believe that our partners will not violate these rules during the whitelist stage. However, if we open up free proposals during the post-whitelist stage, it is difficult to guarantee what data they will propose to the chain.
Therefore, it is best to add a check to the contract to ensure that the contract will not crash.

YoGhurt111: > I wonder if some of this logic(and more broadly some of the validations in the propose function) can be moved off-chain to the node. cc @YoGhurt111 since you may know better

We currently have validations for these parameters, and we believe that our partners will not violate these rules during the whitelist stage. However, if we open up free proposals during the post-whitelist stage, it is difficult to guarantee what data they will propose to the chain.
Therefore, it is best to add a check to the contract to ensure that the contract will not crash.

ggonzalez94: But we could move some of these checks to the prover to enforce them? If things can be enforced by the prover vs smart contract check, I think we should prefer the former

ggonzalez94:
I'm playing with simplifying the preconf router, wrapper and forced inclusions into a single contract. Am I right to assume that the new design only supports blobs for batches and not calldata? I don't think we have the `txList` as part of the batch anymore

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: Will create a separate wrapper contract to expose the pack/unpack functions.

dantaik: see https://github.com/taikoxyz/taiko-mono/pull/19773

packages/protocol/contracts/layer1/based2/libs/LibPropose.sol
dantaik: Let me add a TODO -- if we can move forced inclusion support to the core protocol, then I'll change propose-batches to propose-a-single-batch

packages/protocol/contracts/layer1/based2/libs/LibData.sol
dantaik: Do you mean we should offer encode/decoder for extraData as well? I think it's a good idea.

dantaik: See https://github.com/taikoxyz/taiko-mono/pull/19773

packages/protocol/contracts/layer1/based2/IInbox.sol
dantaik: Right, but fixed in https://github.com/taikoxyz/taiko-mono/pull/19773

packages/protocol/contracts/layer1/based2/libs/LibVerify.sol
dantaik: fixed in https://github.com/taikoxyz/taiko-mono/pull/19773

packages/protocol/contracts/layer1/based2/libs/LibVerify.sol
dantaik: The proposer needs to verify up to N batches, regardless whole proposed or proved those batches. Verifying other people's batches shall not revert.

packages/protocol/contracts/layer1/based2/IInbox.sol
dantaik: This needs to be discussed in more detail. The current idea is that you can extract the propose4 call’s parameters (i.e., the calldata), and—together with information emitted in the Proposed event (such as msg.sender, timestamp, and block number, which are accessible from the event context even if not explicitly defined)—you should be able to reconstruct all the necessary information.

However, I recall that extracting the parameters from propose4 requires tracing. Is it still true? We can certain emit the  propose4 call’s parameters, but it's just more expensive.

dantaik: We have 2 options:

- emit the calldata in an event, this is more expensive
- ensure propose4 can only be called as the outer most transaction.

This issue remains to be fixed.

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: The current idea is to build BatchMetadata offchain without emitting it directly to save gas.

dantaik:
> I'm playing with simplifying the preconf router, wrapper and forced inclusions into a single contract. Am I right to assume that the new design only supports blobs for batches and not calldata? I don't think we have the `txList` as part of the batch anymore

Yes

packages/protocol/contracts/layer1/based2/IInbox.sol
davidtaikocha: IMO we can just extract the `propose4` call’s parameters, and it doesn't require tracing, just need to impl a calldata decoder in client, we know the `txHash` for each event, and then fetching and decoding the calldata in the transaction body, its not an issue.

packages/protocol/contracts/layer1/based2/ITaikoInbox2.sol
dantaik: please see https://github.com/taikoxyz/taiko-mono/pull/19775

packages/protocol/contracts/layer1/based2/libs/LibPropose.sol
dantaik: please see https://github.com/taikoxyz/taiko-mono/pull/19775

packages/protocol/contracts/layer1/based2/IInbox.sol
```diff
@@ -0,0 +1,410 @@
...
+        /// @notice Hash of the transition metadata
+        bytes32 metaHash;
+    }
+
+    /// @notice Summary of the current protocol state
+    /// @dev Contains key metrics and identifiers for protocol operation
+    struct Summary {
+        /// @notice ID to be assigned to the next batch
+        uint48 nextBatchId;
+        /// @notice ID of the last block synced from L1
```
linoscope: If I understand correctly:

```suggestion
        /// @notice ID of the last L2 block synced to L1
```

packages/protocol/contracts/layer1/based2/IInbox.sol
linoscope: +1 that the comment needs clarification.

packages/protocol/contracts/layer1/based2/IInbox.sol
linoscope: I think we can just remove this field, as afaiu we removed calldata support (for now) and for blobs we already have `blobHashes`. At least I had to remove it to make the code compile.

linoscope: If you agree on removing this, we also need to modify `_calculateTxsHash` to `_calculateBlobHashes` and stop returning the `txsHash`.a

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,424 @@
...
+        returns (I.Summary memory)
+    {
+        LibBinding.Bindings memory bindings = _getBindings();
+
+        (
+            I.Summary memory summary,
+            I.Batch[] memory batches,
+            I.ProposeBatchEvidence memory evidence,
+            I.TransitionMeta[] memory transitionMetas
+        ) = bindings.decodeProposeBatchesInputs(_inputs);
```
ggonzalez94, line 62: why do we use the bindings here since we are already calling it from the inbox? it seems like unnecessary indirection. Or is it to be flexible to switch to a library implementation in the future?

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,424 @@
...
+        I.Config memory config = _getConfig();
+
+        // Propose batches
+        summary = LibPropose.propose(bindings, config, summary, batches, evidence);
+
+        // Verify batches
+        summary = LibVerify.verify(bindings, config, summary, transitionMetas);
+
+        bytes memory summaryEncoded = bindings.encodeSummary(summary);
```
ggonzalez94: same: why do we use the bindings here since we are already calling it from the inbox? it seems like unnecessary indirection. Or is it to be flexible to switch to a library implementation in the future?

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -0,0 +1,424 @@
...
+    function propose4(bytes calldata _inputs)
+        external
+        override(I, IPropose)
+        nonReentrant
+        returns (I.Summary memory)
+    {
+        LibBinding.Bindings memory bindings = _getBindings();
+
+        (
+            I.Summary memory summary,
```
ggonzalez94: I see we removed the call to `_validateSummary` when we merged the new library binding. Is this a mistake or are we doing the validation somewhere else?

packages/protocol/contracts/layer1/based2/IInbox.sol
dantaik: Comment improved.

packages/protocol/contracts/layer1/based2/IInbox.sol
dantaik: Yes

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: Multiple batches can be proved by different provers. If verification is triggered during the prove function, then the first call to prove would update the Summary, causing subsequent prove calls for other batches to revert. This is a side effect of the current design, where **the Summary is not fully stored on-chain and must be provided as an input parameter**.

A more critical issue is the timing and cost uncertainty of verification. ZK Proofs are typically submitted a few minutes (in practice) after the prover quotes a price to the proposer. If the prover is responsible for covering the verification cost—and that cost is uncertain on chain—they risk incurring a significant loss.

Proposers (who are also preconfers) are less affected by this issue, as they control L1 sequencing directly or indirectly. They can schedule all prove calls to occur after their corresponding propose calls, mitigating the risk of unexpected verification costs. This Shasta fork is specifically designed with preconfirmation in mind, making this sequencing strategy practical.

packages/protocol/contracts/layer1/based2/IInbox.sol
dantaik: @ggonzalez94 as mentioned early, you probably want to create separate refactor PRs to offer some naming alternatives. I'm terrible at naming.

dantaik: Feel free to do it anytime you want :D

packages/protocol/contracts/layer1/based2/IInbox.sol
dantaik: Yes. The current issue with Taiko’s base fee mechanism is that developers arbitrarily forced block proposers to sell a fixed amount of gas per second, without giving them the ability to adjust their supply. Under the new design, block proposers can now collectively adjust gas supply based on profitability.

Proposers will compete in an open market. For example, Alice may choose to increase the amount of gas she sells per second to maximize her profit, potentially forcing out less efficient proposers like Bob, who is barely breaking even. In response, Bob might reduce his gas supply. If the competitive pressure drives Bob into losses, he may exit the market entirely, allowing Alice to build even more blocks. This dynamic benefits users by encouraging efficiency and lowering costs through competition.

packages/protocol/contracts/layer1/based2/libs/LibVerify.sol
dantaik: Fixed

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
dantaik: they are indeed same slot messages. If proposers dont' benefit from these messages, then they will not include any signals here to avoid any extra cost.

packages/protocol/contracts/layer1/based2/libs/LibValidate.sol
dantaik: I hope another L2 can also be launched based on this protocol with inboxWrapper set to address(0) to tempoarary disable force inclusion.

dantaik:
@claude please review the code changes by ignoring commit history or reviewer comments.

dantaik:
@claude please review the code changes by ignoring commit history or reviewer comments.

ggonzalez94 (approved):

davidtaikocha (approved):

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: Removed this function from binding.

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: Also removed. These are results of refactoring.

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: We may not need the 4 postfix for all functions, but some—like depositBond—do require it to avoid conflicts or ensure clarity. For consistency and to prevent future ambiguity, it’s better to apply the 4 postfix to all public and external functions.

packages/protocol/contracts/layer1/based2/IInbox.sol
dantaik: Removed

dantaik:
@claude please review

YoGhurt111 (approved):

PR #19668, opened 2025-06-26
feat(protocol): apply a minimal basefee of 0.1 gwei

dantaik:
Following up on today’s preconf implementers call, I’ve created this small PR to ensure some mainnet profitability. I don’t believe it requires any changes to the node, driver, or preconf-sidecars (no fork needed), but please confirm @cyberhorsey @YoGhurt111 .

### DIFF
--- a/packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
+++ b/packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
@@ -24,7 +24,7 @@ abstract contract PacayaAnchor is OntakeAnchor, ITaiko {
 
     /// @notice Golden touch address is the only address that can do the anchor transaction.
     address public constant GOLDEN_TOUCH_ADDRESS = 0x0000777735367b36bC9B61C50022d9D0700dB4Ec;
-    uint256 public constant BASEFEE_MIN_VALUE = 100000000000; // 0.1 gwei
+    uint256 public constant BASEFEE_MIN_VALUE = 100_000_000_000; // 0.1 gwei
 
     ISignalService public immutable signalService;
     uint64 public immutable pacayaForkHeight;
--- a/packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
+++ b/packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
@@ -24,6 +24,7 @@ abstract contract PacayaAnchor is OntakeAnchor, ITaiko {
 
     /// @notice Golden touch address is the only address that can do the anchor transaction.
     address public constant GOLDEN_TOUCH_ADDRESS = 0x0000777735367b36bC9B61C50022d9D0700dB4Ec;
+    uint256 public constant BASEFEE_MIN_VALUE = 100000000000; // 0.1 gwei
 
     ISignalService public immutable signalService;
     uint64 public immutable pacayaForkHeight;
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
```diff
@@ -24,6 +24,7 @@ abstract contract PacayaAnchor is OntakeAnchor, ITaiko {
 
     /// @notice Golden touch address is the only address that can do the anchor transaction.
     address public constant GOLDEN_TOUCH_ADDRESS = 0x0000777735367b36bC9B61C50022d9D0700dB4Ec;
+    uint256 public constant BASEFEE_MIN_VALUE = 100000000000; // 0.1 gwei
```
linoscope: Isn't this `100 gwei`? Or am I missing something here.
image

packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
```diff
@@ -24,6 +24,7 @@ abstract contract PacayaAnchor is OntakeAnchor, ITaiko {
 
     /// @notice Golden touch address is the only address that can do the anchor transaction.
     address public constant GOLDEN_TOUCH_ADDRESS = 0x0000777735367b36bC9B61C50022d9D0700dB4Ec;
+    uint256 public constant BASEFEE_MIN_VALUE = 100_000_000_000; // 0.1 gwei
```
linoscope: Isn't this 100 gwei?
image

linoscope:
```suggestion
    uint256 public constant BASEFEE_MIN_VALUE = 100_000_000; // 0.1 gwei
```

thedevbirb:
Just to provide more context, the `minGasExcess` of `${CHAIN}Inbox.sol` e.g. https://github.com/taikoxyz/taiko-mono/blob/edf2041a593ccd754f01c45e6d948411b2faa1c4/packages/protocol/contracts/layer1/hekla/HeklaInbox.sol#L98-L98
can achieve a similar result by setting it to `1_440_000_000`, according to the article https://www.notion.so/chainbound/Taiko-Base-Fee-Analysis-2095abfafc1980199212ca4bdf792703?source=copy_link#20f5abfafc1980638a63ffce637de220.

The difference with a higher `minGasExcess` is that base fee would also grow higher past the the target. For example, with the proposed min gas excess it would look like this[^1]:

image

While with the current value for `minGasExcess` the growth would be less, see:

image

If we only want to change the minimum base fee here, without touching the growth of base fee itself, then hardcoding `100_000_000` wei (0.1 gwei) as min base fee is good for me.

[^1]: **NOTE** that I'm using a 40M gas limit otherwise the points would not fit in the picture :). In practice, with 241M gas as in Taiko mainnet (see https://taikoscan.io/block/1226997) base fee already grows a lot between full blocks, as shown in the article.

packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
thedevbirb: Correct, since one gwei is 10**9 wei, then that should be `100_000_000` wei instead.

cyberhorsey:
@YoGhurt111 when this PR is merged we need to cherry pick it onto the v2.3.1 protocol branch and then upgrade hekla and masaya

adaki2004 (approved):

YoGhurt111:
No need to change the driver code

YoGhurt111 (approved):

PR #19681, opened 2025-07-01
fix(taiko-client): update unsafe l2 to use the latest chain head 

cyberhorsey:
if geth didnt actually update the reorg, it can lead to a situatiuon where the reorg was falsely detected and the chain wasnt actually updated, because the ending hash was the same, however an intermediary block in the middle of the batch was detected as a L2 chain reorg because of a missing block.

A sequencer can propose the same block multiple times, and then propose a batch where the ending hash is the samse but because of our cached origin IDs, it detects as a reorg.


cyberhorsey:
i think i will fix this in a better way actually.

PR #19686, opened 2025-07-03
fix(taiko-client): make the check of preconf blocks synchronously

YoGhurt111:
@cyberhorsey Can we just make the check of preconf blocks synchronously and simply first? The block would be processed one by one, although it will take a long time, in this case the interval between batches is long enough that we have enough time to process the batch slowly. WDYT?


cyberhorsey:
Would seriously delay the preconfing and if a reorg hit we would have been preconfing on bad state for a long time and cause another reorg. Not sure about this. Also not sure if this is the cause of the issues....we added CPU and memory monitoring to the grafana, so let's monitor.

PR #19698, opened 2025-07-07
feat(protocol): add peer IP to preconfwhitelist

### LINKED ISSUE #19687: Update whitelist to add peer IDs, driver poll
Rather than hard-coded p2p static peers, whitelist should be able to store the p2p peer IDa for each sequencer, as well as update them. Driver should grab this on boitup and use this as the static peers, as well as pill for changes or subscribe to events for changes and update peers.

cyberhorsey:
closes #19687

and now every driver will auto-connect to sequencers, ensuring they are part of their mesh. If we update or add operators, we dont have to publish new simple-taiko-node images.

We should add a subsequent PR that listens for events and updates the operators potentially.

@YoGhurt111 I believe we can do something like this:

```
api := p2p.NewP2PAPIBackend(d.p2pNode, log.Root(), metrics.P2PNodeMetrics)

// whatever multiAddr we get from the event, either added or changeMultiaddr event
addr := "/ip4/34.30.143.124/tcp/4001/p2p/16Uiu2HAmGz1wUxLnEpcJMZ2Pd5FePrXtGCSKCqX34n3jfmsdiXgu"
if err := api.ConnectPeer(d.ctx, addr); err != nil {
    return fmt.Errorf("failed to connect peer %q: %w", addr, err)
}
```

whenever an event comes in.


packages/taiko-client/driver/driver.go
```diff
@@ -128,6 +129,16 @@ func (d *Driver) InitFromConfig(ctx context.Context, cfg *Config) (err error) {
...
 
+		// fetch and append the peerIps frm the contracts, add them as staticIps
+		peerIps, err := d.rpc.GetPreconfWhitelistPeerIps(nil)
+		if err != nil {
+			return err
+		}
+
+		staticPeers := make([]core.Multiaddr, 0, len(peerIps))
+		// convert string to multiaddr
+		cfg.P2PConfigs.StaticPeers = append(cfg.P2PConfigs.StaticPeers, staticPeers...)
```
merklefruit: `staticPeers` should be populated here

cyberhorsey: should be removed actually, also i added deduplification

merklefruit: lgtm

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -14,6 +14,7 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
         uint64 activeSince; // Epoch when the operator becomes active.
         uint64 inactiveSince; // Epoch when the operator is no longer active.
         uint8 index; // Index in operatorMapping.
+        string peerIp; // peer ip to connect to for the taiko driver
```
merklefruit: Since these need to be conformant to the libp2p `multiAddr`s formed of transport socket address and peer ID, could we rename this field `string multiAddr` for better consistency?

packages/taiko-client/driver/driver.go
```diff
@@ -128,6 +129,14 @@ func (d *Driver) InitFromConfig(ctx context.Context, cfg *Config) (err error) {
 			return err
 		}
 
+		// fetch and append the peerIps frm the contracts, add them as staticIps.
+		// allow this call to fail.
+		peerIps, err := d.rpc.GetPreconfWhitelistPeerIps(nil)
```
merklefruit: Just a thought: does it make sense to establish connection with sequencers as _static peers_ if we may need to update them (e.g. change or add some) with the node being live?

One alternative may be to treat them as regular peers instead, and create a custom routine similar to [`extraHost.monitorStaticPeers`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/p2p/host.go#L120) to make sure they are always dialed. 

The main difference I can see is that this would not require any diff to the op-lib.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
cyberhorsey: sure!

packages/taiko-client/driver/driver.go
cyberhorsey: sure - this makes sense. I was going to do a small diff to the op-lib.

With this method, adding them as regular peers, is ther an exposed method from the op-lib that lets us add them? I'll check but assume so since you say it requires no diff, would be much more useful

cyberhorsey: yeah
```
api := p2p.NewP2PAPIBackend(d.p2pNode, log.Root(), metrics.P2PNodeMetrics)

// connect to a specific multiaddr:
addr := "/ip4/34.30.143.124/tcp/4001/p2p/16Uiu2HAmGz1wUxLnEpcJMZ2Pd5FePrXtGCSKCqX34n3jfmsdiXgu"
if err := api.ConnectPeer(d.ctx, addr); err != nil {
    return fmt.Errorf("failed to connect peer %q: %w", addr, err)
}
```

looks good. @YoGhurt111 will be attempting to add this!

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -60,4 +66,7 @@ interface IPreconfWhitelist {
     ///      of randomness.
     /// @return The address of the operator.
     function getOperatorForNextEpoch() external view returns (address);
+
+    /// @notice Updates the peer IP address for a specific operator.
```
thedevbirb:
```suggestion
    /// @notice Updates the multiaddr for a specific operator.
```

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -6,14 +6,20 @@ pragma solidity ^0.8.24;
...
     /// @param activeSince The timestamp when the operator became active.
-    event OperatorAdded(address indexed operator, uint256 activeSince);
+    event OperatorAdded(address indexed operator, string multiAddr, uint256 activeSince);
 
     /// @notice Emitted when an operator is removed from the whitelist.
     /// @param operator The address of the operator that was removed.
     /// @param inactiveSince The timestamp when the operator became inactive.
     event OperatorRemoved(address indexed operator, uint256 inactiveSince);
 
+    /// @notice Emitted when an operator updates their peer IP address.
```
thedevbirb:
```suggestion
    /// @notice Emitted when an operator updates their multiaddr.
```

packages/taiko-client/driver/driver.go
```diff
@@ -250,6 +285,25 @@ func (d *Driver) eventLoop() {
...
+			}
+		case e := <-operatorChangedMultiAddrCh:
+			log.Info(
+				"New OperatorChangedMultiAddr event",
+				"operator", e.Operator,
+				"multiAddr", e.MultiAddr,
+			)
+			if err := d.connectPeer(e.MultiAddr); err != nil {
+				log.Error("Failed to handle new OperatorChangedMultiAddr event", "error", err)
+			}
```
thedevbirb: Do we need to disconnect from the previous peer multiaddr or this is a non-issue?

packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -1044,6 +1046,47 @@ func (c *Client) GetPreconfRouterPacaya(opts *bind.CallOpts) (common.Address, er
...
+			continue // Skip if the address is empty.
+		}
+		operator, err := c.PacayaClients.PreconfWhitelist.Operators(opts, address)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get operator %s: %w", address.Hex(), err)
+		}
+		multiaddr, err := multiaddr.NewMultiaddr(operator.MultiAddr)
+		if err != nil {
+			continue
+		}
```
thedevbirb: Maybe logging that a certain operator has an invalid multiaddr can be helpful?

thedevbirb (commented):
Minor things

packages/taiko-client/driver/driver.go
YoGhurt111: It can reuse the original connection if the multiaddr is same.
Does this answer your question?

packages/taiko-client/pkg/rpc/methods.go
YoGhurt111: Sounds good and done! https://github.com/taikoxyz/taiko-mono/pull/19698/commits/f7c18b6581108eb1580dce279a166163d276cd1c

packages/taiko-client/driver/driver.go
```diff
@@ -149,6 +151,24 @@ func (d *Driver) InitFromConfig(ctx context.Context, cfg *Config) (err error) {
...
+						defer d.peerConnectionWG.Done()
+						if err := backoff.Retry(func() error {
+							return api.ConnectPeer(d.ctx, peerAddr)
+						}, backoff.WithContext(backoff.NewConstantBackOff(d.RetryInterval), d.ctx)); err != nil {
+							log.Error("Attempt to connect to peer failed", "error", err)
+						}
+					}(multiAddr.String())
+				}
+				d.peerConnectionWG.Wait()
+			}
```
merklefruit: Connecting these peers at startup is great, however we'd also want to have a routine running in the background to make sure that they are connected at all times. Here is what the op library does for static peers: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/p2p/host.go#L120

we could do something similar here, like this pseudo-code:

```go
// in the driver, add the list of sequencer multiaddrs:
struct Driver {
	--snip--
	sequencerMultiAddrs []core.Multiaddr
}

// when configuring the node:
sequencerMultiAddrs, err := d.rpc.GetPreconfWhitelistMultiAddrs(nil)
if err == nil && len(sequencersMultiAddrs) > 0 {
	// call ConnectPeer on all of them, and save them in d.sequencerMultiAddrs
	d.initSequencerPeerings(sequencerMultiAddrs)

	// every minute, check and retry to connect if disconnected
	go d.monitorSequencerPeerings() 
}
```

then, when we get an OperatorAdded/OperatorMultiAddrChanged event, 
we can just add its multiAddr to the list of `d.sequencerMultiAddrs` so the background task
will automatically try to connect it at the next timer tick.

packages/taiko-client/driver/driver.go
thedevbirb: In this PR it seems that we try to connect to those peers at init time, by reading the whitelist, but ideally we should have some recurrent job to check whether we're still connected to those. So we probably need to save them in a map and try to connect with them periodically. If an operator changes its addr, we should probably kick out the old addr out of this map.

Does it make sense? cc @merklefruit

thedevbirb: See also https://github.com/taikoxyz/taiko-mono/pull/19698#discussion_r2192424822

packages/taiko-client/driver/driver.go
YoGhurt111: Got your point, will refactor it.

YoGhurt111: https://github.com/taikoxyz/taiko-mono/pull/19698/commits/f885744e87a7d46ff0dc4d42f9fd7a0f2411376e

packages/taiko-client/driver/driver.go
YoGhurt111: https://github.com/taikoxyz/taiko-mono/pull/19698/commits/f885744e87a7d46ff0dc4d42f9fd7a0f2411376e

packages/taiko-client/driver/driver.go
merklefruit: that looks great! awesome work

packages/taiko-client/driver/driver.go
thedevbirb: Perfect!

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, whitelist_peer_ips]
```
YoGhurt111, line 5:
```suggestion
    branches: [main]
```

YoGhurt111 (approved):

packages/protocol/test/layer1/preconf/whitelist/PreconfWhitelist.t.sol
```diff
@@ -116,9 +118,9 @@ contract TestPreconfWhitelist is Layer1Test {
         assertEq(whitelist.getOperatorForNextEpoch(), address(0));
 
         vm.prank(whitelistOwner);
-        whitelist.addOperator(Alice);
+        whitelist.addOperator(Alice, "");
```
ggonzalez94, line 121: I don't think it is inherently bad to use an empty address for the tests, but they won't reflect real costs on gas benches because of the string size

packages/protocol/gas-reports/layer1-contracts.txt
```diff
@@ -84,9 +84,9 @@ InboxTest_ProposeAndProve:test_inbox_reprove_by_transition_with_same_parent_hash
 InboxTest_ProposeAndProve:test_inbox_reprove_the_same_batch_with_same_transition_will_do_nothing() (gas: 1273783)
 InboxTest_ProposeAndProve:test_inbox_ring_buffer_will_be_reused() (gas: 15700908)
 InboxTest_StopBatch:test_inbox_num_batches_verified() (gas: 8678196)
-PreconfRouterTest:test_preconfRouter_proposeBatch() (gas: 379185)
-PreconfRouterTest:test_preconfRouter_proposeBatch_notOperator() (gas: 328307)
-PreconfRouterTest:test_preconfRouter_proposeBatch_proposerNotSender() (gas: 377815)
+PreconfRouterTest:test_preconfRouter_proposeBatch() (gas: 453061)
+PreconfRouterTest:test_preconfRouter_proposeBatch_notOperator() (gas: 402183)
+PreconfRouterTest:test_preconfRouter_proposeBatch_proposerNotSender() (gas: 451691)
```
ggonzalez94: Cost of proposing increased significantly. I'm still trying to understand why, as I imagine we don't read the `multiAddr` of the operator

ggonzalez94: Ok, I found the reason why. When proposing we call `getOperatorPerEpoch` on the `PreconfWhiteList` contract, which in turn calls the `isOperatorActive` function.
This function loads the `OperatorInfo` struct into memory, which now holds the entire string for `multiAddr`.

If we are keeping this as part of the struct in storage, at least we should change the `isOperatorActive` function to use storage instead of memory so that we avoid doing multile `SLOAD` unnecesarily

packages/protocol/test/layer1/preconf/whitelist/PreconfWhitelist.t.sol
ggonzalez94: Yeah the gas costs really change. Empty string is 1 slot vs 4 slots for an actual address(if I understand the format correctly). I think it is worth using something with the right length and format at least

packages/protocol/gas-reports/layer1-contracts.txt
ggonzalez94: A lot of it is actually because the test adds operators(which won't be the case on regular batch proposals), but the point above still holds

ggonzalez94: https://github.com/taikoxyz/taiko-mono/pull/19698/commits/ae1ecb9ca834bf02bca3888dbe18f3a32d3bd092

ggonzalez94 (commented):
Did a first pass trough the smart contracts and left a few comments. I’m missing a lot of context on this, but it is a bit odd that we are storing ip addresses on-chain. Isn’t there a mechanism where this can be propagated via the p2p network off-chain?

And in the case that the answer is not / too complicated for now - maybe we can at least consider not storing the entire multiAddr onchain and just emitting it as part of the event? If I’m not mistaken it occupies 4 slots on the contract, so definitely expensive.

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -6,14 +6,20 @@ pragma solidity ^0.8.24;
...
 
     /// @notice Emitted when an operator is removed from the whitelist.
     /// @param operator The address of the operator that was removed.
     /// @param inactiveSince The timestamp when the operator became inactive.
     event OperatorRemoved(address indexed operator, uint256 inactiveSince);
 
+    /// @notice Emitted when an operator updates their multiaddr.
+    /// @param operator The address of the operator that was removed.
+    /// @param multiAddr The new multiaddr.
+    event OperatorChangedMultiAddr(address indexed operator, string multiAddr);
```
ggonzalez94: nit: Perhaps we can make this a more generic event in case we add fields to the operator in the future that we want to emit in the event.

```suggestion
    /// @notice Emitted when an operator is updated.
    /// @param operator The address of the operator.
    /// @param multiAddr The new multiaddr.
    event OperatorUpdated(address indexed operator, string multiAddr);
```

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -30,6 +31,11 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
 
     uint256[46] private __gap;
 
+    modifier onlyOwnerOrOperator(address _operator) {
```
ggonzalez94, line 34: nit: we only use this modifier in one place. Instead of having an indirection we can just have the require directly on the function and throw a custom error

ggonzalez94 (commented):
left a couple more nits

dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -52,8 +58,24 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
     }
 
     /// @inheritdoc IPreconfWhitelist
-    function addOperator(address _operator) external onlyOwner {
-        _addOperator(_operator, operatorChangeDelay);
+    function addOperator(address _operator, string calldata multiAddr) external onlyOwner {
```
dantaik, line 61: If our preconfirmation partner reaches out requesting a change to the IP address, we won’t have a clear way to justify the request, right?

@cyberhorsey — could you share the rationale behind adding IP addresses on-chain?

cyberhorsey: I don't know ehSr you mean by the first paragraph about justify the request.

The rationale is that all nodes should be directly peered to any sequencers but in a gossip mesh that isn't guaranteed, and also, if a sequencer is added *after* a node is started, they won't be int he driver's p2p "--p2p.static" peers list, and we must publish a new simple Taiko node image with that peer in the list, and get node operators to update.

With this, nothing has to happen when an operator is added or they change their sequencer box setup. Operators are added seamlessly to the network and all nodes will have all the sequencers as peers, no image or flag updates.

.github/workflows/taiko-client--docker.yml
cyberhorsey: keep this in as we build the images to test on Masaya until PR is either merged or denied

cyberhorsey:
Closed, will have to think of an alternative solution involving several updates to the optimism library itself.

PR #19701, opened 2025-07-07
fix(taiko-client): update L1 origin of parent if the block is building on a orphaned parent

cyberhorsey:


RogerLamTd (approved):

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -917,6 +920,38 @@ func (s *PreconfBlockAPIServer) TryImportingPayload(
...
+				Withdrawals:  make([]*types.Withdrawal, 0),
+				Version:      engine.PayloadV2,
+				TxListHash:   &txListHash,
+			}
+		)
+
+		payloadID := args.Id()
+		// update L1 Origin if the parent block is in the fork chain, we are building
+		// on an orphaned block.
+		_, err = s.rpc.L2Engine.UpdateL1Origin(ctx, &rawdb.L1Origin{
```
YoGhurt111, line 964: Do we need to update the parent block's L1Origin either?

cyberhorsey: ACtually, I think this Pr needs work still. Ignore it for now and will re-request review

cyberhorsey: @YoGhurt111 i updated it and the comparison check as well

YoGhurt111 (approved):

PR #19715, opened 2025-07-09
feat(protocol): add overseer for preconfs

AnshuJalan:
This PR introduces the Overseer component responsible for:
- Slashing malicious commitments via the URC
  - When designated as the slasher, the Overseer can slash commitments deemed malicious.
  - This capability currently applies only to lookahead commitments until EIP-7917 is activated.
- Blacklisting the validators of preconf operators for subjective faults:
  - Validators can be blacklisted if they violate protocol expectations (e.g., failing to adhere to fair exchange commitments).
  - Operator of blacklisted validators is excluded from insertion into the lookahead.


packages/protocol/contracts/layer1/preconf/iface/IProtector.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.27;
+
+import { ISlasher } from "@eth-fabric/urc/ISlasher.sol";
+
+/// @title IProtector
+/// @custom:security-contact security@taiko.xyz
+interface IProtector is ISlasher {
```
linoscope: Should we rename `Protector` to `Overseer` so we can unify with our fair exchange solution?

linoscope: Entity-wise I see them as same entity. Or are there any implementation-wise difference between them?

AnshuJalan: May be we can call this the `Overseer` and also add the blacklist functionality here itself?

packages/protocol/contracts/layer1/preconf/impl/Protector.sol
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        numSigners = uint128(_signers.length);
+        signingThreshold = _signingThreshold;
+    }
+
+    function _verifySignatures(bytes32 digest, bytes[] memory _signatures) internal view {
+        address lastSigner;
+        address currentSigner;
+
```
mikhailUshakoff:
```suggestion
require(signingThreshold <= _signatures.length, InvalidSigningThreshold());
```
Should we check that there are enough signatures in the `_signatures` array?

packages/protocol/contracts/layer1/preconf/impl/Protector.sol
```diff
@@ -0,0 +1,147 @@
...
+        for (uint256 i; i < _signatures.length; ++i) {
+            // Recover the signer from the signature
+            currentSigner = ECDSA.recover(digest, _signatures[i]);
+            require(signers[currentSigner], NotAnExistingSigner());
+
+            // To prevent replay of signatures
+            require(lastSigner < currentSigner, SignersMustBeSortedInAscendingOrder());
```
linoscope: Why is the order important of `_signatures` important?

mikhailUshakoff: In my opinion, the goal is to prevent signing twice from the same address.

packages/protocol/contracts/layer1/preconf/iface/IProtector.sol
```diff
@@ -0,0 +1,38 @@
...
+
+    /// @notice Updates the signing threshold required for multi-signature operations.
+    /// @param _signingThreshold The new threshold value.
+    /// @param _signatures Array of signatures from existing signers authorizing the update.
+    function updateSigningThreshold(
+        uint64 _signingThreshold,
+        bytes[] memory _signatures
+    )
+        external;
+}
```
linoscope: We don't have interface for `slash`.

AnshuJalan: It's inherited from ISlasher

packages/protocol/contracts/layer1/preconf/impl/Protector.sol
AnshuJalan: Ahh, strange pretty sure I added it. Might have gotten removed by mistake. Thanks for pointing out!

packages/protocol/contracts/layer1/preconf/impl/Protector.sol
linoscope: Ah, I see. But don't we also need a check that the length of `_signatures` is same as `numSigners`?

linoscope: Ah, @mikhailUshakoff you already had similar comment above.

packages/protocol/contracts/layer1/preconf/iface/IProtector.sol
linoscope: I see. But still think some documentation of `function slash` would be good. Right now, there is no explaination of what `IProtector` is doing as a whole, expect that it's some form of multisig. Maybe add docs to the `IProtector` interface itself?

linoscope: Basically explain that it is a slasher contracts that let's the multisig conduct slashing, and is expected to be used temporarily for lookahead enforcement until EIP lands.

packages/protocol/contracts/layer1/preconf/iface/IProtector.sol
linoscope: > May be we can call this the Overseer and also add the blacklist functionality here itself?

Sounds good. In this case the `IOverseer` should explain that it is for slashing/blackliting objective faults.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,261 @@
...
+import "src/shared/common/EssentialContract.sol";
+
+/// @title Overseer
+/// @notice The Overseer is responsible for the following:
+/// - Slashing malicious commitments via the URC when assigned as the slasher
+///   (currently only for lookahead commitments until EIP-7917 is live)
+/// - Blacklisting preconf operators based on subjective faults. For instance, non-adherence to
+///   fair exchange.
+///   (blacklisted operators are not inserted in the lookahead)
+/// @custom:security-contact security@taiko.xyz
```
linoscope:
```suggestion
/// @title Overseer
/// @notice The Overseer is a multisig that is responsible for the following:
/// - Slashing malicious commitments via the URC when assigned as the slasher.
///   Currently only for slashing invalid lookahead commitments until EIP-7917 is live.
/// - Blacklisting preconf operators based on subjective faults. For instance, non-adherence to
///   fair exchange. Blacklisted operators are not inserted in the lookahead.
/// @custom:security-contact security@taiko.xyz
```

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,261 @@
...
+    constructor(address _urc) {
+        urc = _urc;
+    }
+
+    function init(uint64 _signingThreshold, address[] memory _signers) external initializer {
+        __Essential_init(address(0));
+        __OverseerInit(_signingThreshold, _signers);
+    }
+
+    /// @dev Based on URC's ISlasher interface
```
linoscope: Thinking we should ditch URC slashing for overseer, and just rely on blacklisting. Should simplify the code a lot, not only in contracts but also the overseer node code as it doesn't have to handle different handling for subjective blacklisting and subjective slashing. Since lookahead can be overwritten immediately by overseer, don't think high crypoeconomic collateral slashing is needed. Also, the URC integration part will go away anyway with EIP-7917, which I expect to be before the post-whitelist launch.

What do you think?

linoscope: Talked offline, since posting invalid lookahead can potentially gain MEV that outweights the missed opportunity of blacklisting (especially if the blacklisted operator has few validators), so better to slash. Won't matter anyway if we have lookahead EIP before mainnet, but in the narrow chance that the post-whitelist launch is before EIP, then better to slash.

linoscope (approved):
LGTM

AnshuJalan:
The latest change allows blacklisting consensus layer validator keys (BLS) instead of the operator's URC registration root. 

P.S: This required me to use the latest version of Solady, which caused some code to break due to a conflicting `LibBytes` library. All issues have been fixed.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,289 @@
...
+import "src/shared/common/EssentialContract.sol";
+import "@solady/src/utils/MerkleProofLib.sol";
+import "@solady/src/utils/MerkleTreeLib.sol";
+
+/// @title Overseer
+/// @notice The Overseer is responsible for blacklisting validators of preconf operators based on
+/// subjective faults.
+/// For instance, non-adherence to fair exchange.
+/// @dev Operators of blacklisted validators are not inserted in the lookahead. This is done to
+/// prevent the lookahead from being polluted by invalidators.
```
linoscope:
```suggestion
/// prevent the lookahead from being polluted by invalid validators.
```

?

linoscope: Though I like the sound of "invalidators" :slightly_smiling_face:

packages/protocol/contracts/layer1/preconf/libs/LibPreconfConstants.sol
```diff
@@ -23,6 +23,18 @@ library LibPreconfConstants {
...
 
     bytes32 internal constant PRECONF_DOMAIN_SEPARATOR = keccak256("TAIKO_ALETHIA_PRECONF");
+    bytes32 internal constant ADD_OVERSEER_SIGNER_DOMAIN_SEPARATOR =
+        keccak256("TAIKO_ALETHIA_ADD_OVERSEER_SIGNER");
+    bytes32 internal constant REMOVE_OVERSEER_SIGNER_DOMAIN_SEPARATOR =
+        keccak256("TAIKO_ALETHIA_REMOVE_OVERSEER_SIGNER");
+    bytes32 internal constant UPDATE_OVERSEER_SIGNING_THRESHOLD_DOMAIN_SEPARATOR =
+        keccak256("TAIKO_ALETHIA_UPDATE_OVERSEER_SIGNING_THRESHOLD");
+    bytes32 internal constant OVERSEER_SLASH_DOMAIN_SEPARATOR =
+        keccak256("TAIKO_ALETHIA_OVERSEER_SLASH");
```
linoscope:
```suggestion
```

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,289 @@
...
+    // -----------------------------------------------------------------------------------
+
+    /// @inheritdoc IOverseer
+    function blacklistValidators(
+        BLS.G1Point[] calldata _validatorPubKeys,
+        bytes[] memory _signatures
+    )
+        external
+    {
+        // Merkleize the validators to a single validator's merkle root
```
linoscope, line 42: How much overhead do we have with Merkalizing the validator pub keys? I expect blacklisting to be done on a per-validator basis rather in bulk for most cases, so if the overhead is large we can consider blacklisting per validator.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,289 @@
...
+
+        // The set of validators must not be already blacklisted
+        require(
+            blacklistTimestamps.blacklistedAt <= blacklistTimestamps.unBlacklistedAt,
+            ValidatorsAlreadyBlacklisted()
+        );
+
+        // If the validators were unblacklisted, the overseer must wait for a delay before
+        // blacklisting
+        // them again in order to not mess up the lookahead.
```
linoscope: Do we mess up the lookahead in this case? And not fall back to fallback preconfer, as AFAIU we do when just blacklisting a validator who hasn't been unblacklisting.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
AnshuJalan: I think this was done by cursor, haha. But may be invalid validators is more semantically correct

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
AnshuJalan: It is not that much since it is just keccak, and not the more expensive sha256. If an operator has made a subjective fault they can keep repeating it if we only blacklist one of its validators and not all of them, right?

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
AnshuJalan: A blacklist -> unblacklist -> blacklist (or vice versa) must not happen cross-epoch (so a delay of at least one epoch is needed), because this can be used to incorrectly slash a lookahead poster for inserting a validator who was blacklisted or not inserting an unblacklisted validator.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
linoscope: Yup, let's replace with invalid validators.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
linoscope: >  If an operator has made a subjective fault they can keep repeating it if we only blacklist one of its validators and not all of them, right?

Actually, this raises an interesting point - If an operator misbehaves we probably want to blacklist both the validator and the operator, as the operator is the one who is doing the fault. Can we blacklist the validator, and transitivitly blacklist the operator that the validator is assigned to? Guess the issue there would be when a single validator is assigned to multiple operators?

AnshuJalan: > Guess the issue there would be when a single validator is assigned to multiple operators

Correct. Also, blacklisting the operator is not very useful and it is cheap to recreate a new operator in the URC

linoscope: Hmm, rethinking, I think it makes sense to blacklist the whole validator set that is assigned to the faulty operator key, which I think you were indicating from the start.

packages/protocol/contracts/layer1/preconf/iface/IOverseer.sol
```diff
@@ -0,0 +1,105 @@
...
+/// @custom:security-contact security@taiko.xyz
+interface IOverseer {
+    struct BlacklistTimestamps {
+        uint128 blacklistedAt;
+        uint128 unBlacklistedAt;
+    }
+
+    struct Config {
+        uint256 blacklistDelay;
+        uint256 unblacklistDelay;
```
linoscope: Can you add comments that explain these fields?

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
linoscope: What I am worried about is a validator being unblacklisted, and with 1 day of `blacklistDelay`, they are immune to black listing for 1 day and can violate fair exchange freely in that duration. Wonder if there are other ways to prevent wrong lookahead poster slashing by observing the `blacklistedAt` and `unblacklistedAt` of the overseer contract.

AnshuJalan: We can keep a history of blacklist/unblacklist timestamps and check changes around the boundary of an epoch, but that would add some complications

AnshuJalan: Btw, the delay would be just one epoch

dantaik:
@claude please review

packages/protocol/contracts/layer1/preconf/iface/IOverseer.sol
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import "@solady/src/utils/ext/ithaca/BLS.sol";
+
+/// @title IOverseer
+/// @custom:security-contact security@taiko.xyz
+interface IOverseer {
+    struct BlacklistTimestamps {
+        uint128 blacklistedAt;
```
dantaik: I think using uint48 is good enough

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,289 @@
...
+
+    /// @inheritdoc IOverseer
+    function updateSigningThreshold(
+        uint64 _signingThreshold,
+        bytes[] memory _signatures
+    )
+        external
+    {
+        // The new threshold must not exceed the number of signers
+        require(_signingThreshold <= numSigners, InvalidSigningThreshold());
```
dantaik: _signingThreshold shall also !=0

dantaik: Lets use if-revert -- I recently learned that require(..., CustomError) will only convert CustomError as "CustomError" string.

packages/protocol/contracts/layer1/preconf/iface/IOverseer.sol
```diff
@@ -0,0 +1,108 @@
...
+    /// @return The current configuration of the overseer
+    function getConfig() external view returns (Config memory);
+
+    /// @notice Returns the proof of inclusion of a validator in the blacklist
+    /// @param _validatorPubKeys consensus layer public keys of the validators
+    /// @param _validatorIndex index of the validator in the validator set
+    /// @return The proof of inclusion of the validator in the blacklist
+    function getValidatorBlacklistInclusionProof(
+        BLS.G1Point[] calldata _validatorPubKeys,
+        uint256 _validatorIndex
```
dantaik, line 63: _validatorPubKeys indicate we are handling multiple validators, but _validatorIndex indiciates otherwise, is _validatorIndex the index of the first validator ?

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,289 @@
...
+                    nonce++,
+                    _validatorsRoot
+                )
+            );
+            _verifySignatures(digest, _signatures);
+        }
+
+        blacklist[_validatorsRoot].unBlacklistedAt = uint128(block.timestamp);
+
+        emit Unblacklisted(_validatorsRoot, block.timestamp);
```
dantaik: maybe we shall emit _validatorPubKeys in the Blacklisted event.

packages/protocol/contracts/layer1/preconf/libs/LibPreconfConstants.sol
```diff
@@ -23,6 +23,16 @@ library LibPreconfConstants {
     uint256 internal constant RANDOMNESS_DELAY_EPOCHS = 2;
 
     bytes32 internal constant PRECONF_DOMAIN_SEPARATOR = keccak256("TAIKO_ALETHIA_PRECONF");
+    bytes32 internal constant ADD_OVERSEER_SIGNER_DOMAIN_SEPARATOR =
+        keccak256("TAIKO_ALETHIA_ADD_OVERSEER_SIGNER");
+    bytes32 internal constant REMOVE_OVERSEER_SIGNER_DOMAIN_SEPARATOR =
+        keccak256("TAIKO_ALETHIA_REMOVE_OVERSEER_SIGNER");
```
dantaik: if a constant is only used in one contract, maybe we should move it to the contract.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,289 @@
...
+import "@solady/src/utils/MerkleTreeLib.sol";
+
+/// @title Overseer
+/// @notice The Overseer is responsible for blacklisting validators of preconf operators based on
+/// subjective faults.
+/// For instance, non-adherence to fair exchange.
+/// @dev Operators of blacklisted validators are not inserted in the lookahead. This is done to
+/// prevent the lookahead from being polluted by invalid validators.
+/// @custom:security-contact security@taiko.xyz
+contract Overseer is IOverseer, EssentialContract {
```
dantaik: Shall we remove singer management code and use a gnosis safe as the authornized transactor? Or at least we move singer management to a new SimpleMultisig contract?

packages/protocol/contracts/layer1/preconf/libs/LibBlockHeader.sol
```diff
@@ -31,7 +31,7 @@ library LibBlockHeader {
     }
 
     function encodeRLP(BlockHeader memory _blockHeader) internal pure returns (bytes memory) {
-        LibRLP.List memory list = LibRLP.l();
```
dantaik, line 34: shall we not use `LibRLP.l()`, according to claude:

-----


Based on my analysis of the LibRLP library, here's the difference between LibRLP.l() and LibRLP.p():

  LibRLP.l() - List creation function

  - Creates a new empty RLP list
  - This is the starting point for building an RLP-encoded list
  - Returns an empty List struct that can be used to append items

  LibRLP.p() - Push/append function

  - Appends an item to an existing RLP list
  - Takes a list and a value as parameters
  - Returns the list for function chaining
  - Can append various types: uint256, address, bool, bytes, or even another List

  Example Usage:

  // Create an empty list
  LibRLP.List memory list = LibRLP.l();

  // Append items to the list
  list = LibRLP.p(list, someAddress);
  list = LibRLP.p(list, someUint256);
  list = LibRLP.p(list, someBytes);

  // Or chain them:
  list = LibRLP.p(LibRLP.p(LibRLP.l(), value1), value2);

  In the code you saw earlier:
  LibRLP.List memory list = LibRLP.l();  // Create empty list
  list = LibRLP.p(list, abi.encodePacked(_blockHeader.parentHash));  // Append parentHash
  list = LibRLP.p(list, abi.encodePacked(_blockHeader.ommersHash));  // Append ommersHash
  // ... continue appending other fields

  This pattern is used to build RLP-encoded data structures, particularly for Ethereum block headers and transaction encoding.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
AnshuJalan: I think this was the case for Solidity `0.8.26`, but starting `0.8.27`, only the signature is used. Here's an excerpt from their release: 
```
The release introduced a new overload — require(bool, <error>), which reverts with the signature of the error rather than Error(string) used by the other variant.
```

AnshuJalan: The repository is using 0.8.30 for compilation

packages/protocol/contracts/layer1/preconf/libs/LibBlockHeader.sol
AnshuJalan: Seems like Claude is not aware of the newer Solady version. The newer version that I have upgraded to, no longer has ` LibRLP.l()`.  `LibRLP.p()` itself returns an empty list.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
AnshuJalan: I think moving the signer management to a new contract is a cleaner option

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
AnshuJalan: The URC allows for up to 256 validators for an operator. Given that our plan is to blacklist the operator's entire validator set, emitting the keys here would be fairly expensive as it would require `128 * 256 bytes` of memory expansion, in the worst case.

AnshuJalan: Alternatiely, we may read the validator set from the transaction calldata. The node is already doing that for the URC, so we will have some reusable code.

packages/protocol/contracts/layer1/preconf/iface/IOverseer.sol
AnshuJalan: Hmm, the naming scheme is ambiguous. I was trying to refer to an index within the first argument i.e the pub keys list.

I have switched to using "PubKeys" instead of referring to the "validator" in the latest commit.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
AnshuJalan: I don't think this situation would turn up, because in the signer removal function we are checking `require(numSigners > signingThreshold, CannotRemoveSignerWhenThresholdIsReached());`

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
dantaik: Reading from calldata is tricky, if the calldata is for an inner transaction, we need tracing feature to be turned on before we can read it.

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
dantaik: good to know. thanks

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
dantaik: But in our case, we can control which EOA calls it, so the tx won't be an inner tx. No need to worry then.

dantaik:
@claude please review

dantaik (approved):

KorbinianK (approved):

bearni95 (approved):

PR #19716, opened 2025-07-09
feat(protocol): introduces a PID controller to adjust base fee

dantaik:
## Summary

This PR introduces a new contract `PIDBaseFeeController`, which implements a Proportional-Integral-Derivative (PID) controller to dynamically adjust the base fee in Taiko's based rollup protocol.

The goal is to replace the current static or curve-based base fee logic, which has proven ineffective — often resulting in a flat base fee that fails to reflect actual demand or usage patterns. This new design is both **simpler** and **more adaptive**, and better fits the proposer-driven architecture of a based rollup.

### 1. Proposer-Controlled Gas Target
- Proposers can slightly adjust the block gas target per block (or per batch, per epoch)
  - `+1%`, `-1%`, or `no change`.
- This allows the gas target to evolve gradually and reflect consensus among active proposers.
- No need for the protocol to predefine an accurate gas target.

### 2. PID-Based Base Fee Adjustment
The base fee is updated using a PID controller: baseFee += kP * error + kI * ∑error + kD * Δerror

where `error = actualGasUsed - gasTarget`

The implementation:
- Uses scaled coefficients (×1000) for precision.
- Adds **anti-windup** logic to avoid runaway integral accumulation during saturation.
- Introduces **integral decay** to reduce long-term drift.
- Optionally dampens sharp changes with a derivative term.

## Benefits

- **Self-correcting**: Initial base fee and gas target do not need to be accurate.
- **Adaptive**: Reacts to actual block usage, converging toward fair pricing.
- **Stable**: Avoids overreaction and handles bursty traffic well.
- **Simple**: Easy to audit and reason about.
- **Efficient**: Optimized for per-block updates, with minimal gas usage.

## Next Steps

- [ ] Simulate and tune PID coefficients


adaki2004:
Is the approach considering:
`A`: baseFee as "the" revenue
OR
`B`: baseFee as an additional revenue to the priority fee (where the latter is the main source of income).

If `A`, what happens in case + / - 1% is simply not enough, because L1 has huge spikes ? 

If `B` then also somehow it seems to me that the option to increase/decrease the gas target (= their revenue) why wouldn't they do it with just the priority fee ?

btw, I like the idea of moving to a direction where it is more profitable for the DAO, for sure.

PR #19719, opened 2025-07-10
feat(taiko-client): add domain envelope type, update l1Origin with signature and forced inclusion marker

### LINKED ISSUE #19688: Forced Inclusion Store client fix
Regarding what we discuss in the call earlier, here is what I understood as the fix for the invalid forced-inclusions causing a reorg.

Cause: 
If someone sends an invalid transaction as forced inclusion through L1, the sequencer processing it will produce an empty block (with only the anchor tx). Before Shasta, we won't have access to the "isForcedInclusion" bitFlag in the header's extradata, so for now how sequencers try to find the forced inclusions is based only on scanning blocks looking for transaction hashes that match the ones in the ForcedInclusionStore queue. Now if a block only has the anchor tx, it's impossible for other sequencers to know if it was forced-included or not.

If one of such "empty forced blocks" was produced by sequencer A, but they didn't manage to propose it, then sequencer B will trigger a L2 reorg when proposing it.
The cause of the reorg is not clear, but it appears to have something to do with a mismatching txListHash between the preconfirmed and proposed block.

Solution:
The easiest short term solution will be to include an extra field in the PayloadEnvelope of gossiped preconfirmed blocks to let others know that a block is forced. If this info is able to be cached in Geth, then we can also recover the forced blocks when syncing unsafe blocks by querying Geth, removing the chance for this ambiguity. Medium term this is solved with the new bitfalg in Shasta anyway.

In practice the solution will involve:
 Update op-lib envelope type to include isForcedInclusion
 Update driver /preconfBlocks API body to allow sequencers to specify isForcedInclusion  
 Update driver and geth L1Origin to include isForcedInclusion too

### LINKED ISSUE #19720: Allow any p2p node to respond to a message request

cyberhorsey:
closes #19688 
closes #19720

With this PR, we use the L1Origin to store the ForcedInclusion marker as well as the signature of the original signer of the envelope. This way, any node with the block being requested can respond with it, not only the original sequencer.

This makes a missed message a much more salvagable situation.

Also, by using a domain `*preconf.Envelope`, if we want to add anyting additional to gossiping that would also store in L1Origin, we set ourselves up for much easier future implementations.


packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go
```diff
@@ -273,23 +273,23 @@ func (i *BlocksInserterPacaya) InsertBlocks(
 // InsertPreconfBlocksFromExecutionPayloads inserts preconfirmation blocks from the given execution payloads.
 func (i *BlocksInserterPacaya) InsertPreconfBlocksFromExecutionPayloads(
```
YoGhurt111, line 274:
```suggestion
func (i *BlocksInserterPacaya) InsertPreconfBlocksFromEnvelopes(
```

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go
```diff
@@ -314,15 +314,15 @@ func (i *BlocksInserterPacaya) InsertPreconfBlocksFromExecutionPayloads(
 // the given execution payload.
 func (i *BlocksInserterPacaya) insertPreconfBlockFromExecutionPayload(
```
YoGhurt111, line 315:
```suggestion
func (i *BlocksInserterPacaya) insertPreconfBlockFromEnvelope(
```

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go
```diff
@@ -461,9 +468,9 @@ func (i *BlocksInserterPacaya) IsBasedOnCanonicalChain(
 
 	log.Debug(
 		"Check if block is based on canonical chain",
-		"blockID", uint64(executableData.BlockNumber),
-		"blockHash", executableData.BlockHash,
-		"parentHash", executableData.ParentHash,
+		"blockID", uint64(envelope.Payload.BlockNumber),
+		"blockHash", envelope.Payload.BlockHash,
+		"parentHash", envelope.Payload.ParentHash,
```
YoGhurt111, line 476:
```suggestion
		"parentHash", envelope.Payload.ParentHash,
		"isForcedInclusion", envelope.IsForcedInclusion,
```

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -44,7 +45,7 @@ var (
 
 // preconfBlockChainSyncer is an interface for preconfirmation block chain syncer.
 type preconfBlockChainSyncer interface {
-	InsertPreconfBlocksFromExecutionPayloads(context.Context, []*eth.ExecutionPayload, bool) ([]*types.Header, error)
+	InsertPreconfBlocksFromExecutionPayloads(context.Context, []*preconf.Envelope, bool) ([]*types.Header, error)
```
YoGhurt111, line 48:
```suggestion
	InsertPreconfBlocksFromEnvelopes(context.Context, []*preconf.Envelope, bool) ([]*types.Header, error)
```

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -537,7 +557,7 @@ func (s *PreconfBlockAPIServer) OnUnsafeL2EndOfSequencingRequest(
 // find all the missing ancients and import them, an error will be returned.
 func (s *PreconfBlockAPIServer) ImportMissingAncientsFromCache(
 	ctx context.Context,
-	currentPayload *eth.ExecutionPayload,
+	currentPayload *preconf.Envelope,
```
YoGhurt111, line 601:
```suggestion
	currentEnvelope *preconf.Envelope,
```

YoGhurt111:
LGTM, maybe need to rename all 'payload' to 'envelope', but it's ok to test it first, focus on the important part!

YoGhurt111 (approved):

cyberhorsey:
> LGTM, maybe need to rename all 'payload' to 'envelope', but it's ok to test it first, focus on the important part!

Yes, feel free if you have time to do stuff like this. I finally got tests passing but I got 2 hours sleep before preconf meeting and now I will have a nap hahahah. But yes, would be appreciated. I will deploy this to Masaya later this evening by pushing up a build to this branch, leave it running for awhile, before we OK it. But I think it looks good.

dantaik (approved):

packages/taiko-client/driver/preconf_blocks/queue.go
```diff
@@ -8,17 +8,18 @@ import (
...
 	"github.com/taikoxyz/taiko-mono/packages/taiko-client/internal/metrics"
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/pkg/preconf"
 )
 
 // maxTrackedPayloads is the maximum number of prepared payloads the execution
 // engine tracks before evicting old ones.
 const maxTrackedPayloads = 768 // equal to `maxBlocksPerBatch`
 
-// payloadQueueItem represents an id->payload tuple to store until it's retrieved
+// payloadQueueItem represents an id->envlope tuple to store until it's retrieved
```
RogerLamTd:
```suggestion
// payloadQueueItem represents an id->envelope tuple to store until it's retrieved
```

packages/taiko-client/driver/preconf_blocks/queue.go
```diff
@@ -37,52 +38,53 @@ func newPayloadQueue() *payloadQueue {
 }
 
 // put inserts a new payload into the queue at the given id.
```
RogerLamTd:
```suggestion
// put inserts a new payload item into the queue at the given id.
```

RogerLamTd (approved):

RogerLamTd (approved):

PR #19721, opened 2025-07-10
test(taiko-client): remove the dependency of `debug_setHead` for better node implementations diversity support

davidtaikocha:
After this PR, the L2 chain head will be reset to height `1` after each test case (not reset to genesis anymore), to make the tests compatible with other L2 Taiko node implementations which don't support `debug_setHead` (e.g. `reth`, `nethermind`). Then they can use `engine_forkchoiceUpdate` to reset the chain head to a non-genesis height. 

Also including:
- Removed `testify.Suite` dependency in `anchor_tx_constructor_test.go` and `fixed_k_signer_test.go` to solve an import-cycle issue.
- Some changes to pass the tests when initial L2 chain head is `1`.
- Changed some `errors.Is(err, ethereum.NotFound)` to `err.Error() == ethereum.NotFound.Error()` to make the code compatible with other non-golang L2 node implementations.

### DIFF
--- a/packages/taiko-client/internal/docker/start.sh
+++ b/packages/taiko-client/internal/docker/start.sh
@@ -5,8 +5,7 @@ source scripts/common.sh
 if [ "$L2_NODE" == "l2_geth" ];then
     DOCKER_SERVICE_LIST=("l1_node" "l2_geth")
   else
-    echo "unsupported L2_NODE: $L2_NODE"
-    exit 1
+    DOCKER_SERVICE_LIST=("l1_node")
 fi
 
 # start docker compose services
--- a/packages/taiko-client/internal/testutils/helper.go
+++ b/packages/taiko-client/internal/testutils/helper.go
@@ -10,7 +10,6 @@ import (
 	"github.com/cenkalti/backoff/v4"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/params"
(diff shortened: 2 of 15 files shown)


packages/taiko-client/internal/docker/start.sh
```diff
@@ -5,8 +5,7 @@ source scripts/common.sh
 if [ "$L2_NODE" == "l2_geth" ];then
     DOCKER_SERVICE_LIST=("l1_node" "l2_geth")
   else
-    echo "unsupported L2_NODE: $L2_NODE"
-    exit 1
+    DOCKER_SERVICE_LIST=("l1_node")
```
YoGhurt111, line 8: If we set the L2_NODE(not l2_geth), it will not exit and keep testing with only `l1_node`, it will fail, why not exiting here?

packages/taiko-client/internal/testutils/helper.go
```diff
@@ -318,3 +322,105 @@ func SendDynamicFeeTx(
 	}
 	return tx, nil
 }
+
+func (s *ClientTestSuite) resetToBaseBlock(key *ecdsa.PrivateKey) {
+	// Propose an empty block at genesis.
+	t := s.TxMgr("proposer", key)
+	inbox := common.HexToAddress(os.Getenv("PROVER_SET"))
```
YoGhurt111, line 329:
```suggestion
	inbox := common.HexToAddress(os.Getenv("TAIKO_INBOX"))
```

YoGhurt111:
LGTM, just left one question

packages/taiko-client/internal/docker/start.sh
davidtaikocha: Since rn in docker-compose, we only have `l2_geth`. So its just a change for other impls to be able to test by this script locally.

YoGhurt111: Got your point, they can set the env to their local port

YoGhurt111 (approved):

cyberhorsey (approved):

PR #19722, opened 2025-07-10
feat(protocol): allow operators to register a new address(sequencer)

ggonzalez94:
## The problem

Some of our preconf parterns use `web3signer` which is not compatible with op lib signer. This means that for the driver you have to enter your PK when initializing it. This does not allow to keep the PK that holds funds on a secure vault or somewhere else. The solution is to use two different addresses, one for proposing(holds ETH and TAIKO) and other for the driver software(can have 0 funds).
The protocol only needs to be aware of the proposer address(for access control), while the preconf software needs to know both(this is only off-chain).
@cyberhorsey can elaborate if I missed something important.

## The solution

Long term I believe we should avoid storing this on-chain, which means making the driver able to use `web3signer` or similar. But for the whitelist launch I believe it is an acceptable trade-off to store the sequencer address on-chain in a reverse mapping that only the proposer can update.
This PR adds sequencer address support to the PreconfWhitelist contract, allowing operators to register both a sequencer address and a driver address.

**Design Decisions**
1. Storage Optimization: Changed epoch fields to uint32 to fit all operator data in one storage slot, reducing gas costs. This should still be enough for another ~4 300 yrs even with something as extreme as 1s slots

** Gas Impact **
  - Minimal impact on the critical path (proposals) - no changes to getOperatorForCurrentEpoch(). This is not clear from the current snapshots, because the tests for proposing also add operators.
  - Slight increase in operator management operations due to additional storage.

## TODO's

- [x] Decide if on the mappings and functions we should use `proposer` instead of `operator` for the names for more clarity
- [x] <s>Provide a method to allow operators to update their sequencer address without needing to un-register</s>
   - Since this is a very infrequent operation, and to avoid the node having to listen and adjust to a new event + add a new storage variable for update delays and tracking old and new sequencers, operators can unregister and request to be added with a new sequencer address instead
- [x] Review gas benches more carefully
- [ ] <s>Create a test that actually measures the cost of proposing without the overhead of adding operators to have a good baseline to how much things change, not only on this PR but going forward</s>
    - Implemented on this [PR](https://github.com/taikoxyz/taiko-mono/pull/19738)

## IMPORTANT
Since we are changing the storage layout of the `OperatorInfo` struct we SHOULD remove all operators and consolidate before doing the upgrade, then add them again. This will avoid the code trying to read old storage slots with the new layout. If we follow this order everything should work fine @cyberhorsey


ggonzalez94:
@claude review this PR and provide concrete feedback and suggestions

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -5,14 +5,16 @@ pragma solidity ^0.8.24;
 /// @custom:security-contact security@taiko.xyz
 interface IPreconfWhitelist {
     /// @notice Emitted when a new operator is added to the whitelist.
-    /// @param operator The address of the operator that was added.
+    /// @param operator The sequencer address of the operator that was added.
+    /// @param driver The driver address of the operator that was added.
     /// @param activeSince The timestamp when the operator became active.
-    event OperatorAdded(address indexed operator, uint256 activeSince);
+    event OperatorAdded(address indexed operator, address indexed driver, uint256 activeSince);
```
cyberhorsey: I think if we keep the event name OperatorAdded and the term operator to describe a whitelisted entity but we can all the two parameters "sequencer" and "proposer" for clarity. Only "proposer" needs to be known onchain. Only "sequencer" needs to be known to the driver.

We currently have a mismatch in param named as the "addOperator" function calls them sequencer and driver, here we call them operator and driver. I believe sequencer and proposer are the correct entity names that describe their purpose.

To think of it in functional terms: an "operator" has a "sequencer" and a "proposer".

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -5,14 +5,16 @@ pragma solidity ^0.8.24;
...
 
     /// @notice Emitted when an operator is removed from the whitelist.
-    /// @param operator The address of the operator that was removed.
+    /// @param operator The sequencer address of the operator that was removed.
+    /// @param driver The driver address of the operator that was removed.
     /// @param inactiveSince The timestamp when the operator became inactive.
-    event OperatorRemoved(address indexed operator, uint256 inactiveSince);
+    event OperatorRemoved(address indexed operator, address indexed driver, uint256 inactiveSince);
```
cyberhorsey: ditto for operatoradded

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -11,16 +11,22 @@ import "src/shared/common/EssentialContract.sol";
 /// @custom:security-contact security@taiko.xyz
 contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
     struct OperatorInfo {
-        uint64 activeSince; // Epoch when the operator becomes active.
-        uint64 inactiveSince; // Epoch when the operator is no longer active.
+        uint32 activeSince; // Epoch when the operator becomes active.
+        uint32 inactiveSince; // Epoch when the operator is no longer active.
         uint8 index; // Index in operatorMapping.
+        address driverAddress; // Driver address for this operator (for off-chain use).
```
cyberhorsey: sequencerAddress

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -11,16 +11,22 @@ import "src/shared/common/EssentialContract.sol";
...
     event Consolidated(uint8 previousCount, uint8 newCount, bool havingPerfectOperators);
     event OperatorChangeDelaySet(uint8 delay);
 
+    /// @dev The operator address is the sequencer address.
     mapping(address operator => OperatorInfo info) public operators;
     mapping(uint256 index => address operator) public operatorMapping;
+    /// @dev Reverse mapping from driver to operator(sequencer). This is only used by off-chain actors.
+    /// @dev Note: Entries are never deleted. Off-chain actors should verify the sequencer is still active
+    /// by checking that operators[sequencer].activeSince != 0 before trusting the mapping.
+    mapping(address driver => address operator) public driverToSequencer;
```
cyberhorsey: address sequencer > address proposer public sequencerToProposer;

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -210,25 +220,27 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
...
             operators[_operator].activeSince = 0;
 
             havingPerfectOperators = false;
-            emit OperatorRemoved(_operator, inactiveSince);
+            emit OperatorRemoved(_operator, driver, inactiveSince);
         }
     }
 
     /// @dev The cost of this function is primarily linear with respect to operatorCount.
-    function _getOperatorForEpoch(uint64 _epochTimestamp) internal view returns (address) {
```
cyberhorsey, line 231: can't see in the diff, will pull down this PR and ensure this function is still returning what is expected and leave a further comment! (on mobile)

cyberhorsey:
From implementing in the client side, to make this feature optional (ie: from what I can tell only Nethermind has this request), the driver would pull down the sequencer address, and the protocol would be aware of the proposer address. I haven't given it too muchr bought yet but they should either remain being able to be set to the same address to avoid complexity, or the driver should just pull down both and allow either address to sequence during the operator's epoch. I'm going to take a poke around the driver code today and see what paths this affects.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -11,16 +11,22 @@ import "src/shared/common/EssentialContract.sol";
...
     }
 
     event Consolidated(uint8 previousCount, uint8 newCount, bool havingPerfectOperators);
     event OperatorChangeDelaySet(uint8 delay);
 
+    /// @dev The operator address is the sequencer address.
     mapping(address operator => OperatorInfo info) public operators;
     mapping(uint256 index => address operator) public operatorMapping;
+    /// @dev Reverse mapping from driver to operator(sequencer). This is only used by off-chain actors.
+    /// @dev Note: Entries are never deleted. Off-chain actors should verify the sequencer is still active
```
linoscope: Why "never deleted"? Shouldn't we delete the entries at `function consolidate()`, when we do `delete operators[operator];`?

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -164,14 +170,15 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
...
-    function epochStartTimestamp(uint256 _offset) public view returns (uint64) {
-        return uint64(
+    function epochStartTimestamp(uint256 _offset) public view returns (uint32) {
+        return uint32(
             LibPreconfUtils.getEpochTimestamp() + _offset * LibPreconfConstants.SECONDS_IN_EPOCH
         );
     }
 
-    function _addOperator(address _operator, uint8 _operatorChangeDelay) internal {
+    function _addOperator(address _operator, address _driver, uint8 _operatorChangeDelay) internal {
```
linoscope: We may want to make the `driver` (or `sequencer` after renaming) optional by keeping the `_addOperator` unchanged and adding a separate `addDriver` function. That way, if you don't need separate key for P2P you can just ignore the driver part. What do you guys think @ggonzalez94 @cyberhorsey

linoscope: On second thought, probably simpler for driver, etc, to always have the `sequencer` key, and have it be the same as the `proposer` key if the separation is not needed.

ggonzalez94: agree, I think it is cleaner. This means this is unchanged right?

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
ggonzalez94: It is mostly a gas optimization thing, and the client software can handle that off-chain. Even if you get some refund from setting the storage from non zero -> zero the cost of doing the cold SLOAD is bigger. Do you think this can be an issue even if well documented? I imagine clients still need to check if the operator is active before accepting the blocks, so it seems that we are not really losing anything here.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
linoscope: Yes

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
linoscope: I see, makes sense. Then would suggest to mention that we don't delete due to gas optimization in comment, it's a natural thing to wonder when reading the code.

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
ggonzalez94: Updated to use proposer and sequencer

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
ggonzalez94: included that in the comment

packages/protocol/gas-reports/layer1-contracts.txt
```diff
@@ -136,9 +136,9 @@ PreconfRouter2Test:test_whitelistPreconferProposesInSlotItselfWhenPreconferHasUn
 PreconfRouter2Test:test_whitelistPreconferProposesWhenCurrentLookaheadIsEmpty_case1(uint256) (runs: 256, μ: 475903, ~: 475940)
 PreconfRouter2Test:test_whitelistPreconferProposesWhenCurrentLookaheadIsEmpty_case2(uint256) (runs: 256, μ: 452037, ~: 452074)
 PreconfRouter2Test:test_whitelistPreconferProposesWhenNextLookaheadIsEmpty(uint256) (runs: 256, μ: 479017, ~: 479068)
-PreconfRouterTest:test_preconfRouter_proposeBatch() (gas: 379185)
-PreconfRouterTest:test_preconfRouter_proposeBatch_notOperator() (gas: 328307)
-PreconfRouterTest:test_preconfRouter_proposeBatch_proposerNotSender() (gas: 377815)
+PreconfRouterTest:test_preconfRouter_proposeBatch() (gas: 448324)
```
ggonzalez94: gas diffs for proposing look big because the test also adds multiple operators(which now has become more expensive) The actual cost for proposing should be unchanged

dantaik:
I’m still trying to fully understand the challenge here.

Without support from our node/driver for something like Web3Signer, we seem to require two separate keys:

- A proposer key that holds Ether and is used to send L1 transactions.
- An operator key that signs preconf-ed blocks but holds no Ether.

This PR appears to establish a mapping between the two addresses.

However, doesn’t the proposer key still face the same issue—that it cannot be managed by a secure vault?

ggonzalez94:
> I’m still trying to fully understand the challenge here.
> 
> Without support from our node/driver for something like Web3Signer, we seem to require two separate keys:
> 
> * A proposer key that holds Ether and is used to send L1 transactions.
> * An operator key that signs preconf-ed blocks but holds no Ether.
> 
> This PR appears to establish a mapping between the two addresses.
> 
> However, doesn’t the proposer key still face the same issue—that it cannot be managed by a secure vault?

Yes, that's the correct way to think about it. 
For the question: The proposer software does have the possibility to do remote signing, but our driver software uses a library that does not easily support it. @cyberhorsey can provide more details.

linoscope:
@dantaik What @ggonzalez94 mentions below is correct:

> The proposer software does have the possibility to do remote signing, but our driver software uses a library that does not easily support it

Since L1 proposals are just Ethereum transactions, we can naturally use web3Signer (we still had to make upstream changes to support blob txs though). However, since the P2P messages are NOT standard Ethereum transactions, they are not supported out of the box. We had multiple solutions but agreed to go with the "additional sigining key for P2P" approach in this PR. Now, the L1 proposing key can be kept in a secure vault, while the P2P signing key can be passed as env variable to the driver and kept in-memory.

For reference, below message summarizes the possible options we considered and their pros/cons:
image

cyberhorsey:
> @dantaik What @ggonzalez94 mentions below is correct:
> 
> > The proposer software does have the possibility to do remote signing, but our driver software uses a library that does not easily support it
> 
> Since L1 proposals are just Ethereum transactions, we can naturally use web3Signer (we still had to make upstream changes to support blob txs though). However, since the P2P messages are NOT standard Ethereum transactions, they are not supported out of the box. We had multiple solutions but agreed to go with the "additional sigining key for P2P" approach in this PR. Now, the L1 proposing key can be kept in a secure vault, while the P2P signing key can be passed as env variable to the driver and kept in-memory.
> 
> For reference, below message summarizes the possible options we considered and their pros/cons: image

Res, regarding this, it's simply optimisms p2p lib RemoteSigner that would require an eligible proxy, or this addition.

cyberhorsey:
> I’m still trying to fully understand the challenge here.
> 
> Without support from our node/driver for something like Web3Signer, we seem to require two separate keys:
> 
> * A proposer key that holds Ether and is used to send L1 transactions.
> * An operator key that signs preconf-ed blocks but holds no Ether.
> 
> This PR appears to establish a mapping between the two addresses.
> 
> However, doesn’t the proposer key still face the same issue—that it cannot be managed by a secure vault?

No, it can be. Because the actual sequencer itself (either NM, Gattaca or CB) will implement the call to web3signer themselves to support it when proposing the L1 transaction, which is a drop in replacement.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -11,21 +11,40 @@ import "src/shared/common/EssentialContract.sol";
...
+    /// @dev An operator consists of a proposer address(the key to this mapping) and a sequencer
+    /// address.
+    ///     The proposer address is their main identifier and is used on-chain to identify the
+    /// operator and decide if they are allowed to propose.
+    ///     The sequencer address is used off-chain to to identify the address that is emitting
+    /// preconfirmations.
+    ///     NOTE: These two addresses may be the same, it is up to the operator to decide.
+    mapping(address proposer => OperatorInfo info) public operators;
+    mapping(uint256 index => address proposer) public operatorMapping;
+    /// @dev Reverse mapping from sequencer to proposer. This is only used by off-chain actors.
```
cyberhorsey: I think theres confusion, actually only the sequencer is used off-chain, not the proposer. The sequencer is pulled down by each individual preconfing sidecar + the driver and cached in a lookahead to say `this address is able to sequence blocks`. The driver does not care about who can propose, that is done entirely in the sequencer and only the contracts will check the proposer.

A driver or preconfing sidecar currently calls `getOperatorForCurrentEpoch` and `getOperatorForNextEpoch`. Now, this would pull down the proposer. What we actually need a mapping of proposer to sequencer, not a mapping of sequencer to proposer.

Also, gas efficiency for adding/removing operators should not be considered (it will very very rarely be called)  - it would make much more sense for this to be just be deleted and trusted, and require less onchain calls to verify (and speed up preconfing calls).

This, when an operator is registere, we should have
`mapping(address proposer => address sequencer) public proposerToSequencer;`

and a driver and preconfing sidecar implementation is to call
```
proposer, err := c.PacayaClients.PreconfWhitelist.GetOperatorForCurrentEpoch(opts)
if err != nil {
  return common.Addres{}, err

sequencer := c.PacayaClients.PreconfWhitelist.ProposerToSequencer(opts, proposer)
```
and use that sequencer address as the drop-in replacement for the current implementation. THis leaves the call to the onchain contracts alone (completely identical to now - no changes needed to router/wrapper/inbox), and only adds one small call for us to do.

cyberhorsey (changes requested):

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -63,27 +82,61 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
...
     function removeSelf() external {
         _removeOperator(msg.sender, 0);
     }
 
+    /// @notice Updates an operator's configuration, specifically their sequencer address.
+    /// @param _proposer The proposer address of the operator to update.
+    /// @param _newSequencer The new sequencer address for the operator.
+    /// @dev Can be called by either the contract owner or the operator themselves.
+    // TODO: We may want to add a delay for this operation.
+    function updateOperator(address _proposer, address _newSequencer) external {
```
cyberhorsey: We want a delay here, otherwise all preconfing sidecars and driver need to watch for this event and update the sequencer address inplace (since we care about the sequencer, not the proposer), and add tons of unnecessary offchain complexity. It should be subject to the same delay adding an operator is, which would require no changes on anyone's end offchain.

cyberhorsey (changes requested):

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
ggonzalez94: I thought about this a bit more: Do we actually need a delay registered on-chain? If sidecars and drivers are caching the sequencer address they can just use the old one until their cache expires. 
Since this is a contract only used by whitelisted actors, I believe documenting it properly(and maybe emitting the event with the expected delay) should be enough. Implementing the delay on the contract itself increases complexity slightly, and will increase storage. The `OperatorInfo` struct is used when proposing, so we will most likely need to add this delay in a new mapping(or change the `isOperatorActive` function to use storage instead of memory). Operators only need to ensure they don't use the new address right after updating.

If you think this is likely to introduce undesired bugs(because sidecars are implemented differently for example), I'll go ahead and update this to handle the delay on-chain. I just wanna make sure it is needed before going this route.
cc @linoscope in case you have opinions on this

ggonzalez94:
After discussing with @cyberhorsey and @linoscope we decided the reverse mapping was not necessary, so the contract was actually simplified. Off-chain actors can just look at the `sequencerAddress` from the `OperatorInfo` struct for the assigned preconfer and know what the expected sequencer address is.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -62,28 +75,51 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
...
-    /// @notice Removes an operator by address who will become inactive in two epochs.
-    /// @param _operator The address of the operator to remove.
-    /// @param _effectiveImmediately True if the removal should be effective immediately.
-    function removeOperator(address _operator, bool _effectiveImmediately) external onlyOwner {
-        _removeOperator(_operator, _effectiveImmediately ? 0 : operatorChangeDelay);
+    /// @notice Removes an operator by address.
+    /// @param _proposer The proposer address of the operator to remove.
+    /// @param _effectiveImmediately True if the removal should be effective immediately, otherwise
+    /// it will be effective in two epochs.
+    function removeOperator(address _proposer, bool _effectiveImmediately) external onlyOwner {
```
ggonzalez94, line 79: @cyberhorsey I just realized we have 2 `removeOperator` functions that do exactly the same lol. if you set `_effectiveImmediately` this is the same as the one below. Before I merge them I just wanna double check that no off-chain code is relying on the one with a single parameter

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
cyberhorsey: Flow: Sequencer updates WL with new sequencer address, sidecar runs with new sequencer address, all lookahead are cached with old one. Sequencer can not sequence. Driver rejects. Every other driver is running on the p2p network with the old one - every p2p block is rejected from new sequencer. We need to follow the same flow as adding and removing operators. This is what all drivers and sidecars have tons of logic in place to accomodate.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
cyberhorsey: I believe the intent here is if a sequencer is compromised or something or has a critical big we need to eject it and remove it immediately, but I am personally have not used it in add/removal scripts, and instead call regular removal and then consolidate after the epochs roll over.

ggonzalez94: I mean we only need one function for this. What I'm suggesting is getting rid of the other function that does not have this parameter. With this one we can do both immediate ejection or 2 epochs

ggonzalez94: removed the redundant function, let me know if for some reason you think it is still useful

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
ggonzalez94: we ended up removing the `updateOperator` function to avoid the node having to listen and adjust to a new event + add a new storage variable for update delays and tracking old and new sequencers. Operators can unregister and request to be added with a new sequencer address instead

cyberhorsey: thumbs up

cyberhorsey (approved):

cyberhorsey:
Approved and noted, will remove all operators before re-adding, and will test on Masaya first ofc.

Preconfing sidecars need no change at all with this - so it's ready to go when approved and merged, as the client impl is in.

YoGhurt111 (approved):

dantaik (approved):

PR #19717, opened 2025-07-09
feat(repo): add basic flows for claude code to work effectively on the codebase 

ggonzalez94:
## Summary
I've heard very positive things about claude code, and had good experience using it for the past week. I think we should use it more, both locally and within GH. This PR is a first step in that direction.

- The workflow file allows us to mention `@claude` on GH issues and PRs for claude to provide feedback, fix issues or implement small features. You can read more about it [here](https://docs.anthropic.com/en/docs/claude-code/github-actions).
- `CLAUDE.md` allows the gh actions and local claude code CLI to work more effectively with the codebase

## More details
`CLAUDE.md` was generated by claude itself, and I've refined the smart contract section(since that's the one I'm most familiar with). For the client, prover, and frontend side would be great if someone starts using claude code and can improve the sections so that they actually reflect the codebase.

For anyone wanting to start using claude code or dive more into it [this](https://youtu.be/Lue8K2jqfKk?si=Gq17UmJRK3o-wwmB) a really good video.

PR #19723, opened 2025-07-10
feat(repo): Add required permissions for claude to be able to create PRs

ggonzalez94:
I missed this on my previous #19717 PR

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -18,7 +18,7 @@ jobs:
       (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
       (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
       (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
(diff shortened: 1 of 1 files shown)


.github/workflows/claude.yml
```diff
@@ -20,9 +20,9 @@ jobs:
       (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
     runs-on: ubuntu-latest
```
RogerLamTd:
```suggestion
    runs-on: [arc-runner-set]
```

RogerLamTd: @ggonzalez94 our self hosted runner

ggonzalez94: thanks. I imagine our runner does not have stricter network policies that prevent it from accessing  Anthropic's API right?

RogerLamTd: > thanks. I imagine our runner does not have stricter network policies that prevent it from accessing Anthropic's API right?

not that i'm aware of, maybe @kimo-ice would know?

RogerLamTd (approved):

kimo-ice (approved):

PR #19734, opened 2025-07-14
feat(protocol): remove force inclusions support

dantaik:
Depending on the timing of Shasta’s mainnet launch, if permissionless preconfirmation is already live on mainnet, we should remove ForcedInclusion from the core protocol to further simplify it.


ggonzalez94:
@dantaik should we close this after our discussion today?

PR #19644, opened 2025-06-19
[DO NOT MERGE] gas analysis results

adaki2004:
Command to run (and banchmark with base):

`FOUNDRY_TRACE=1 forge test --match-test test_inbox_measure_gas_used -vv`

Notes:
- Please read the whole analysis carefully! 
- I had to coment out emit events both in `one_nonzero_anchor_id_per_batch` and `main` branches due to stack too deep error
- I put the logging in the `main` branch on the same locations as the `one_nonzero_anchor_id_per_batch` (if you want i can attach that file too, but it is locally in my second, benchmark repo)

**v4ProposeBatch analysis:**  

**one_nonzero_anchor_id_per_batch** branch

**1st 10** batches are proposed with this modifier: `WhenMultipleBatchesAreProposedWithDefaultParameters` 
**11th** batch is proposed with 10 blocks with `anchorBlockId` set in 1 block

**1st batch** 
=== v4ProposeBatch Gas Analysis Start ===
  Initial gas: 77364543
  Gas after validation block: 77355947 consumed: 8596
  Creating BatchInfo for 1 blocks
  Gas after BatchInfo creation: 77351283 consumed: 13260
  Starting anchor block processing loop
  Block 0 processed anchor block 2
  Gas after anchor loop: 77348164 consumed: 16379
  Gas after validation+txsHash: 77346779 consumed: 17764
  Gas after metadata+fork check: 77342919 consumed: 21624
  Gas after bond operations: 77303963 consumed: 60580
  Gas after storage updates: 77233407 consumed: 131136
  Gas after verification: 77215145 consumed: 149398
  === **Total gas consumed: 150402** ===

**2nd batch**
  === v4ProposeBatch Gas Analysis Start ===
  Initial gas: 76456586
  Gas after validation block: 76450490 consumed: 6096
  Creating BatchInfo for 1 blocks
  Gas after BatchInfo creation: 76445826 consumed: 10760
  Starting anchor block processing loop
  Block 0 processed anchor block 3
  Gas after anchor loop: 76442707 consumed: 13879
  Gas after validation+txsHash: 76441322 consumed: 15264
  Gas after metadata+fork check: 76437462 consumed: 19124
  Gas after bond operations: 76424406 consumed: 32180
  Gas after storage updates: 76353850 consumed: 102736
  Gas after verification: 76342888 consumed: 113698
  === **Total gas consumed: 114702** === 
  ..
**10th batch is the same**

 **This is where we measure the gas!!**
**11th batch:** 
[...]

PR #19738, opened 2025-07-14
test(protocol): improve gas benches for proposing

ggonzalez94:
## Context

Current gas benches are pretty messy and it is hard to find a single number to track if changes to proposing are increasing or decreasing with each PR. Just checking at snapshots is not ideal, since they reflect the cost of the test function - which is not a direct reflection of the cost of the function(some tests have a lot of setup in the test function itself), making this metric not really meaningful.
The best option for measuring gas IMO is using the granular [gas snapshot cheatcodes](https://getfoundry.sh/forge/gas-tracking/gas-section-snapshots).
- We already have this for proposing and proving when proposing via the inbox [directly](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/test/layer1/based/InboxTest_ProposeAndProve.t.sol#L557). **But the test was not using [isolated mode](https://getfoundry.sh/forge/gas-tracking/gas-section-snapshots#description), so the results were inaccurate by over 50%**
- We do not have good benches for proposing via the new `PreconfRouter`, which makes it more challenging to review PRs that touch this.

## Solution

This PR is a first step to meaningfully improve this:

- Use isolation mode on tests that use `vm.startSnapshotGas`
- Refactor the PreconfRouter tests to also do granular snapshots and to use the actual inbox instead of the mockInbox
- Merge the results of both benches into a single json to make it easier to compare

## TODOs
- [ ] double check some of the configuration variables in `PreconfRouterTestBase`
- [ ] merge some of the common setup logic between `PreconfRouterTestBase` and `InboxTestBase`


packages/protocol/gas-reports/inbox_without_provermarket.txt
```diff
@@ -1,5 +1,5 @@
 See `test_inbox_measure_gas_used` in InboxTest_ProposeAndProve.t.sol
 
-Gas per proposing: 168855
-Gas per proving + verification: 160239
-Total: 329094
\ No newline at end of file
+Gas per proposing: 243378
```
ggonzalez94: Without using isolation on the tests when doing snapshots was giving very misleading results

packages/protocol/test/layer1/based/InboxTest_ProposeAndProve.t.sol
```diff
@@ -554,10 +555,10 @@ contract InboxTest_ProposeAndProve is InboxTestBase {
         ITaikoInbox.BatchParams memory batchParams;
         batchParams.blocks = new ITaikoInbox.BlockParams[](10);
 
-        vm.startSnapshotGas("proposeBatch");
+        vm.startSnapshotGas("ProposeAndProve", "proposeBatchNoRouter");
```
adaki2004, line 558: Yes, so here i tried to dump every info (i had): https://github.com/taikoxyz/taiko-mono/pull/19644#issue-3159669204 but i think the main statement was overshadowed by a huge number of dry measurements facts/debug prints 😄 , which is: 

`I think also gas measurements not averaging out just using the last measurement, which might distort!!`

I think here is stil the case. So with the modifiers below we already proposed and proved:
`WhenMultipleBatchesAreProposedWithDefaultParameters`
`WhenMultipleBatchesAreProvedWithCorrectTransitions`

Wdyt ? Is there a way to averaging out ? Maybe snapshot in the modifiers OR move their logic here - or separate functions ?

ggonzalez94: @claude can you explain this + the linked comment and recommend what you think would be the most accurate to measure gas here?

ggonzalez94: @adaki2004 I think I now understand your concern(I hadn't when I reviewed the original discussion). As a general principle I believe the gas benches should reflect as close as possible the actual costs that proposers and provers will face when calling this on mainnet. I imagine in most cases the storage slots will be cold(for reading, writing should be already non zero because of ring buffer) when they access it, so the benches should reflect that. If you agree I think we should:
1. move all the preparation logic to setUp, base clase or whatever that does not warm the storage slots
2. separate proposing and proving benches into separa tests
3. Not sure really how important the "average" is here. If the gas bench that gets recorded reflects closely what will happen in prod I think we are fine.

Thoughts? If I got a big part of this wrong happy to chat it over on Discord or jump into a call

adaki2004: Yes, i agree with you, we shall be close to mainnet but i think 1-2 realistic (close enough) measurements shall be enough, maybe no need to rewrite the whole test suite (imho) - because also if new design by Daniel is approved, then we have to rewrite (if big chunk changes, donno yet).

I forgot to mention all hands but i’ll be on holiday for 10 (working) days from tomorrow on but can do async for sure if needed.

ggonzalez94: Actually after playing with this for a bit, it might be more tricky than I was expecting. It will probably require deeper changes to the test structure(i.e. not use the modifiers so they don't warm up the storage slots). The difference is not insignificant, but I think the utility of the tests even as they are still stand so that you can actually have a sense for how much proposing and proving costs increase/decrease with each change.

dantaik: @ggonzalez94 maybe we can target the `shasha_reimplemented` branch for improving gas measurement? We need to write new tests for Shasta after all.

ggonzalez94: yes, I was planning to do that as well. But would be good to have a base to compare don't you think. If we like this path(or some version of this), we can use these gas benches to understand how much better the shasta changes are. WDYT?

packages/protocol/test/layer1/based/InboxTestBase.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity ^0.8.24;
 
 import "../Layer1Test.sol";
 import "test/layer1/based/helpers/Verifier_ToggleStub.sol";
+import { InboxTestConfigUtils } from "test/layer1/based/helpers/InboxTestConfigUtils.sol";
```
dantaik, line 6:
```suggestion
import "test/layer1/based/helpers/InboxTestConfigUtils.sol";
```

dantaik (approved):

dantaik:
I think for shasta, we need to rewrite most of the tests.

dantaik:
@ggonzalez94 shall we close this one?

PR #19754, opened 2025-07-19
feat(taiko-client): remove some unnecessary checks in `IsBasedOnCanonicalChain`

davidtaikocha:
`IsBasedOnCanonicalChain` is only used by `insertPreconfBlockFromEnvelope`, which can only insert a block when its parent is in the canonical chain, if there is a gap, it will fail.


cyberhorsey:
@davidtaikocha can you explain why this is not necessary and can be removed? I need to take a look at the whole file not just the diff but don't have time right now, and we are already bundling a release so any code additions / removals we need to be very careful about

davidtaikocha:
Closed this since to pass the tests, this PR should be part of https://github.com/taikoxyz/taiko-mono/pull/19755 , and will explain there.

PR #19755, opened 2025-07-19
feat(taiko-client): improve envelope caching

davidtaikocha:
- introduce `tryPutEnvelopeIntoCache` method to reduce duplicated code
- changes in https://github.com/taikoxyz/taiko-mono/pull/19754
- change to always try to search for cached ancients and always cache the inserted block, which will remove the dependency of the current tricky undocumented geth behavior. (if a block is in a fork, and not too far away from the current height, we can still query it by `eth_getBlockByHash`)


packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -1039,7 +1030,7 @@ func (s *PreconfBlockAPIServer) TryImportingPayload(
 		if msg.Signature != nil {
 			sig = *msg.Signature
 		}
-		// update L1 Origin if the parent block is in the fork chain, we are building
+		// Update L1 Origin if the parent block is in the fork chain, we are building
 		// on an orphaned block.
```
davidtaikocha, line 1081: And after this PR is merged, looks like then we are also safe to remove this orphaned blocks check too?

cyberhorsey: can you clarify why? I agree with the refactor of the try put envelope into cache, but haven't fully read the rest of the diff and we have to be very careful about what we refactor / change rn.

davidtaikocha:
For https://github.com/taikoxyz/taiko-mono/pull/19754#issuecomment-3092702846 , since `IsBasedOnCanonicalChain` is only used by `insertPreconfBlockFromEnvelope` for the check that the given `envelope` can be inserted into the current canonical chain, and for the inner `createExecutionPayloadsAndSetHead`, the given `envelope`'s parent (not ancestors) MUST be part of the current canonical chain, so we can / should remove those code.

packages/taiko-client/driver/preconf_blocks/server.go
davidtaikocha: Since after these changes, if the given envelope is an orphaned block, it will just be cached if we can't find all its missing parent blocks till the canonical from cache, so we won't try to insert it anymore. Correct me if I'm wrong.

cyberhorsey: but we should be inserting it, because it's an intentional reorg via p2p in this situation, so we would cache it and then never actually build the chain? is that not the case here? someone is reorging via p2p and we need to set the canonical to the incoming block with the orphan as it's parent which now makes it canonical.

with this change, we insert into the cache forever until a batchproposed event comes in, or am I misreading the diff?

and IMO we should be doing separate PRs for refactors (try put envelope into cache function) and feature/fixes (the orphaned block change) to be able to parse diffs cleanly and separate non functionality changes from functionality changes

davidtaikocha: Yes, but since after this change, we will insert each envelope into the cache: https://github.com/davidtaikocha/taiko-mono/blob/tryPutPayloadIntoCache/packages/taiko-client/driver/preconf_blocks/server.go#L1119, so if an envelope of an orphan block comes in, we can still import all its parents in the cache (unless the reorg is longer than the cache size or the driver restarted which leads to the missed cache, then we need to wait until a `batchProposed` event comes in). 

For sure, the current impl works, but like I mentioned in the PR description, the reason why I suggest these changes is that the orphan block check relies on a tricky undocumented geth behavior (if a block is in a fork, and its height is not too far away from the current tip, we can still be able to query it by `eth_getBlockByHash` in geth RPC), this is not the spec and geth only atm, which might also will be removed / changed in the later geth release (and it won't be mentioned as a break change). So seems the suggested changes will be safer for us...

davidtaikocha: > we should be doing separate PRs for refactors

Opened a PR only for the refactors here: https://github.com/taikoxyz/taiko-mono/pull/19769 , will update this PR when its merged.

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -1084,43 +1090,32 @@ func (s *PreconfBlockAPIServer) TryImportingPayload(
...
+		log.Info(
+			"Updated L1 Origin for the parent block in the fork chain",
 			"peer", from,
 			"blockID", uint64(msg.ExecutionPayload.BlockNumber),
 			"hash", msg.ExecutionPayload.BlockHash.Hex(),
 			"signature", common.Bytes2Hex(sig[:]),
 		)
 	}
 
-	if parentInFork == nil && (parentInCanonical == nil || parentInCanonical.Hash() != msg.ExecutionPayload.ParentHash) {
```
YoGhurt111, line 1095: @davidtaikocha Is this relying on tricky geth behavior?

davidtaikocha: Yes, this is to remove the reliance on the tricky geth hehavior.

YoGhurt111 (approved):

dantaik:
@claude please review

dantaik:
@claude please review

dantaik:
@claude please review

dantaik:
@claude please review

PR #19766, opened 2025-07-22
feat(protocol): deployment script for preconfirmation related contracts for mainnet

YoGhurt111:
Plz review this PR carefully.


dantaik (approved):

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
```diff
@@ -0,0 +1,55 @@
...
+        address fallbackPreconfProposer = 0x68d30f47F19c07bCCEf4Ac7FAE2Dc12FCa3e0dC9;
+        address proofVerifier = 0xB16931e78d0cE3c9298bbEEf3b5e2276D34b8da1;
+        address taikoToken = 0x10dea67478c5F8C5E2D90e5E9B26dBe60c54d800;
+        address signalService = 0x9e0a24964e5397B566c1ed39258e21aB5E35C77C;
+        address oldFork = 0x904Da4C5bD76f932fE09fF32Ae5D7E3d2A5D2264;
+
+        address whitelist = deployProxy({
+            name: "preconf_whitelist",
+            impl: address(new PreconfWhitelist()),
+            data: abi.encodeCall(PreconfWhitelist.init, (address(0), 2, 2)),
```
davidtaikocha: Shall we also call `transferOwnership` for `PreconfWhitelist` and `PreconfRouter` in this script?

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
```diff
@@ -0,0 +1,55 @@
...
+    function run() external broadcast {
+        address taikoWrapper = 0x9F9D2fC7abe74C79f86F0D1212107692430eef72;
+        address rollupResolver = 0x5A982Fb1818c22744f5d7D36D0C4c9f61937b33a;
+        address taikoInbox = 0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a;
+        address store = 0x05d88855361808fA1d7fc28084Ef3fCa191c4e03;
+        address fallbackPreconfProposer = 0x68d30f47F19c07bCCEf4Ac7FAE2Dc12FCa3e0dC9;
+        address proofVerifier = 0xB16931e78d0cE3c9298bbEEf3b5e2276D34b8da1;
+        address taikoToken = 0x10dea67478c5F8C5E2D90e5E9B26dBe60c54d800;
+        address signalService = 0x9e0a24964e5397B566c1ed39258e21aB5E35C77C;
+        address oldFork = 0x904Da4C5bD76f932fE09fF32Ae5D7E3d2A5D2264;
```
ggonzalez94, line 33: This is not a request for this PR:
- But perhaps we should consider defining common addresses that are used across several deployment scripts in a common constants file. Less room for errors

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
ggonzalez94: Any reason why we are setting owner as the zero address both here and on the preconf router?

ggonzalez94 (commented):
LGTM, left just one question and one suggestion for the future

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
YoGhurt111: Fixed in this commit https://github.com/taikoxyz/taiko-mono/pull/19766/commits/34c1812928d94c36c9048be9f4d25efaee799ca8
init contractOwner to `admin.taiko.eth` directly when deploy proxy.

YoGhurt111: Fixed in this commit https://github.com/taikoxyz/taiko-mono/pull/19766/commits/34c1812928d94c36c9048be9f4d25efaee799ca8
init contractOwner to `admin.taiko.eth` directly when deploy proxy.

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
YoGhurt111: We don't have the sol file about the common addresses now, but have a deployment log file here https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/deployments/mainnet-contract-logs-L1.md

YoGhurt111:
I'm not sure if the `_operatorChangeDelay`(2) and `_randomnessDelay`(2) for WL are what we want.
Need a confirmation @dantaik @ggonzalez94 @cyberhorsey

YoGhurt111:
And another question, the `maxBatchesToVerify` should be `8`, same as the value on Hekla? @dantaik @ggonzalez94 @cyberhorsey

YoGhurt111:
Thanks for the detailed review from @davidtaikocha and @ggonzalez94!!!

ggonzalez94:
> I'm not sure if the `_operatorChangeDelay`(2) and `_randomnessDelay`(2) for WL are what we want. Need a confirmation @dantaik @ggonzalez94 @cyberhorsey

I believe 2 epochs are what we want here



> And another question, the `maxBatchesToVerify` should be `8`, same as the value on Hekla? @dantaik @ggonzalez94 @cyberhorsey

Created  #19770 in case we need to update to the Hekla config since I see this was on the items in our deployment checklist

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
ggonzalez94: this is nice, but I think we would still benefit from having such a constants file and import into the scripts don't you think? Anyways it is a nice to have for the future to avoid potential typos on addresses.

YoGhurt111: Yes, it would be better!

dantaik (approved):

ggonzalez94 (dismissed):

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
```diff
@@ -0,0 +1,58 @@
...
+            // Not sure if the _operatorChangeDelay and _randomnessDelay are what we want
+            data: abi.encodeCall(PreconfWhitelist.init, (contractOwner, 2, 2)),
+            registerTo: rollupResolver
+        });
+
+        address router = deployProxy({
+            name: "preconf_router",
+            impl: address(new PreconfRouter(taikoWrapper, whitelist, fallbackPreconfProposer)),
+            data: abi.encodeCall(PreconfRouter.init, (contractOwner)),
+            registerTo: rollupResolver
```
cyberhorsey, line 48: this registeres it to the rollup resolver, immediately enabling preconfs. This would destroy the network currently if ran.

 We need to *not* register the preconf_router, but deploy it, and register it to the resolver *only when we are ready to enable preconfs*.

IMO It should be three different scripts:

1) deploy whitelist and router and wrapper - to be run immediately, to provide the addresses, so nodes can be set up

2) upgrade the PacayForkRouter/inbox - to be done before mainnet launch

3) enable the preconf router - to be done after all nodes and sequencers are set up, added to WL, deployed, synced, and ready we are ready to enable preconfs.

@dantaik @ggonzalez94 please revoke approvals and please do not run this script @YoGhurt111.

Preconf router should be registeredTo `address(0)` on deploy.

cyberhorsey (changes requested):

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
```diff
@@ -0,0 +1,58 @@
...
+        });
+        address wrapper = address(new TaikoWrapper(taikoInbox, store, router));
+        // Need to call `upgradeTo`, to address: 0x9F9D2fC7abe74C79f86F0D1212107692430eef72
+        console2.log("taikoWrapper: ", wrapper);
+        address newFork =
+            address(new MainnetInbox(taikoWrapper, proofVerifier, taikoToken, signalService));
```
cyberhorsey, line 55: TaikoInbox.sol
```
  if (inboxWrapper == address(0)) {
                    if (params.proposer == address(0)) {
                        params.proposer = msg.sender;
                    } else {
                        require(params.proposer == msg.sender, CustomProposerNotAllowed());
                    }

                    // blob hashes are only accepted if the caller is trusted.
                    require(params.blobParams.blobHashes.length == 0, InvalidBlobParams());
                    require(params.blobParams.createdIn == 0, InvalidBlobCreatedIn());
                    require(params.isForcedInclusion == false, InvalidForcedInclusion());
                } else {
                    require(params.proposer != address(0), CustomProposerMissing());
                    require(msg.sender == inboxWrapper, NotInboxWrapper());
                }
```

TaikoWrapper.sol
```
 /// @inheritdoc IProposeBatch
    function v4ProposeBatch(
        bytes calldata _params,
        bytes calldata _txList,
        bytes calldata
    )
        external
        onlyFromOptional(preconfRouter)
        nonReentrant
        returns (ITaikoInbox.BatchInfo memory, ITaikoInbox.BatchMetadata memory)
 ```

PreconfRouter.sol

```
 /// @inheritdoc IProposeBatch
    function v4ProposeBatch(
        bytes calldata _params,
        bytes calldata _txList,
        bytes calldata /* _additionalData */
    )
        external
        returns (ITaikoInbox.BatchInfo memory info_, ITaikoInbox.BatchMetadata memory meta_)
    {
        // Sender must be the selected operator for the epoch
        address preconfer = preconfWhitelist.getOperatorForCurrentEpoch();
        if (preconfer != address(0)) {
            require(msg.sender == preconfer, NotPreconfer());
        } else if (fallbackPreconfer != address(0)) {
        }
```


if wrapper is enabled on the mainnet inbox, only the wrapper may call it. If preconf router is enabled on the wrapper via address manager, only preconf router may call it. 

I hope everyone sees the issue here and am shocked it got a double approval.

**Review by cyberhorsey — CHANGES_REQUESTED** [review]

File: packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol (hunk 3 above)
Comment (ggonzalez94) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: Since you requested changes it won't be able to get merged, so no worries.
This is why I was asking on slack about the resolvers and address managers, it was(and in many ways still is) very confusing - and I went trough this PR even before trying to understand that, so thanks for catching this.

1. agree on PreconfWhitelist and PreconfRouter deployment. Can the router be registered immediately? I didn't find a reason why not. Also I imagine when you say deploy the TaikoWrapper I imagine you mean a new implementation and then propose an upgrade?
2. I didn't see any changes on the PacayaForkRouter, should we still upgrade it?
3. ok, agree

Comment (cyberhorsey) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: 1) no the router can not be registered immediately. thats what enables preconfs. we can register the whitelist. re: taikoWrapper, depends if its a proxy - it looks like in this script we arent using a proxy for it. im unsure of mainnet status
2) we are upgrading the pacayaforkrouter because we have a new fork (the mainnetinbox with the new taikoWrapper). if the current taikoWrapper is a proxy, its fine to not upgrade the pacayaforkrouter. If the current taikoWrapper is *not* a proxy, we need to upgrade the pacayaforkrouter. I'm unsure of what it's state is on mainnet as I havent been involved in any mainnet deploys recently.

Comment (ggonzalez94) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: 1. sorry, I meant register the whitelist** registering the router is the whole issue that breaks things.
Re: TaikoWrapper - it is a [proxy](https://etherscan.io/address/0x9F9D2fC7abe74C79f86F0D1212107692430eef72#code). The script does not propose an upgrade, but Gavin's comment specifies it should be upgraded.
2. Wrapper is a proxy ^^ so we should be fine if that's the case. But I don't even know what the fork router does TBH, so let's wait for @dantaik to weigh in.

Comment (cyberhorsey) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: yep let's scrap this PR and break the scripts down into deploying the preconf router and WL and registering the WL only, then a wrapper upgrade, then the "enable preconfs" script IMO

Comment (cyberhorsey) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: also the fork router routes calls to old fork or new fork, like when we deploy an upgrade that takes place at a specific block height

File: packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol (hunk 4 above)
Comment (YoGhurt111) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:55]: No, the newFork(TaikoInbox with wrapper) will be enabled when the `upgradeTo` txs sent by multi-sig are signed.

File: packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol (hunk 3 above)
Comment (YoGhurt111) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: We can register the `preconf_router` and the `preconf_whitelist` and set `admin.taiko-eth` as contract owner immediately.
Because we don't use the `AddressResolver` to get the proxy that we will use in the contract, we actually use the proxy passed in the creation ,like the router in `new TaikoWrapper(taikoInbox, store, router)`.
So as long as we don't update the proxy, we won't enable preconf.
cc @ggonzalez94 @cyberhorsey

Comment (YoGhurt111) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: We split it with 2 steps:
1. Deploy all impl and proxy that won't be used immediately.
2. Send the upgrade txs with the multi-sig and execute the txs when you need to enable preconf.

Comment (cyberhorsey) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: Ah, we use the actual preconfRouter setting hardcoded address, not the resolver.

Still, we should separate the scripts into things we plan to do in phases IMO. The updates to the router (which need to be MTI sig signed) should probably go in a separate script, and the deployment of the preconf router and whitelist (which should be done immediately) in another. 

The resolver / hardcodedAddress combination is quite a weird design choice IMO :D

File: packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol (hunk 4 above)
Comment (cyberhorsey) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:55]: Ah, so this is simply the deploys, and nothing is actually upgraded? If so, we can go back to it being one script. Just make sure we don't send the upgrade to transactions yet.

Comment (YoGhurt111) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:55]: Yes, nothing is actually upgraded. Since I will deploy them with my own EOA and my EOA don't have the access to upgrade, haha

File: packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol (hunk 3 above)
Comment (YoGhurt111) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: Yes, it's hardcoded.
Actually all of the script work is step 1.
The resolver design is outdated, but users will query the addresses by the resolver in previous design, so we should keep it up to date.

Comment (YoGhurt111) [reply packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol:48]: Btw, we use the hardwallet and safe UI to propose the multi-sig tx.
So we don't put the deployment and the upgrade in one script.

File: packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol [hunk 5]
```diff
@@ -0,0 +1,58 @@
...
+        address signalService = 0x9e0a24964e5397B566c1ed39258e21aB5E35C77C;
+        address oldFork = 0x904Da4C5bD76f932fE09fF32Ae5D7E3d2A5D2264;
+        // admin.taiko.eth
+        address contractOwner = 0x9CBeE534B5D8a6280e01a14844Ee8aF350399C7F;
+
+        address whitelist = deployProxy({
+            name: "preconf_whitelist",
+            impl: address(new PreconfWhitelist()),
+            // Not sure if the _operatorChangeDelay and _randomnessDelay are what we want
+            data: abi.encodeCall(PreconfWhitelist.init, (contractOwner, 2, 2)),
```
ggonzalez94: let's remove this comment and add a constant just for clarity. One of my commits had that if you wanna use it

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
```diff
@@ -0,0 +1,58 @@
...
+        });
+
+        address router = deployProxy({
+            name: "preconf_router",
+            impl: address(new PreconfRouter(taikoWrapper, whitelist, fallbackPreconfProposer)),
+            data: abi.encodeCall(PreconfRouter.init, (contractOwner)),
+            registerTo: rollupResolver
+        });
+        address wrapper = address(new TaikoWrapper(taikoInbox, store, router));
+        // Need to call `upgradeTo`, to address: 0x9F9D2fC7abe74C79f86F0D1212107692430eef72
```
ggonzalez94: I would add a comment just for clarity that this should be the final step to enable preconfs and should be done after adding the operators and waiting for at least 2 epochs

ggonzalez94 (approved):
Thanks, LGTM. Left a couple last very minor suggestions

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
dantaik: All upgrades needs to be done in gnosis safe wallet with their transaction builder.

dantaik:
> maxBatchesToVerify

Yes

dantaik:
> _randomnessDelay

I think these are right values.

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
dantaik: > 1. sorry, I meant register the whitelist** registering the router is the whole issue that breaks things.
>    Re: TaikoWrapper - it is a [proxy](https://etherscan.io/address/0x9F9D2fC7abe74C79f86F0D1212107692430eef72#code). The script does not propose an upgrade, but Gavin's comment specifies it should be upgraded.
> 2. Wrapper is a proxy ^^ so we should be fine if that's the case. But I don't even know what the fork router does TBH, so let's wait for @dantaik to weigh in.

The fork router routs function calls to two different implementations based on function selector, the old/current fork and the new fork. Since it is function selector-based, new fork's functions must be different from the old fork. Using ForkRouter will avoid one single implementation from implementing both the current fork and the new fork in one contract.

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/19766/commits/141779f91541b09f3204741b596d9509b26bf9e0

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/19766/commits/141779f91541b09f3204741b596d9509b26bf9e0

PR #19770, opened 2025-07-22
chore(protocol): update max batches to verify to match hekla config

ggonzalez94:
This was on our deployment list, and we updated the HeklaInbox with this config - but not the MainnetInbox


ggonzalez94:
closing this since it was already added to the original branch

PR #19774, opened 2025-07-23
feat(protocol): propose a single batch & emit inputs in Proposed event

dantaik:
@ggonzalez94 if you try to merge forced inclusion to the core protocol, this branch is a better base -- I removed custom pack/unpack functions to make changing structs much easier, also this branch enables proposing a single batch, not multiple ones.


ggonzalez94:
@dantaik I only saw this after creating #19780 , I'll work a bit more on some ideas I have there and then try to merge both. Would be great if you still take a look at that one and let me know what you think

PR #19419, opened 2025-05-09
feat(protocol): remove `RollupResolver`

adaki2004:
Remove `RollupResolver`

PR #19776, opened 2025-07-23
feat(protocol): remove the awkward ITaiko interface

dantaik:


packages/protocol/contracts/shared/tokenvault/ERC20Vault.sol
```diff
@@ -442,10 +442,9 @@ contract ERC20Vault is BaseVault {
     /// @param _op Parameters for the solve operation
     function solve(SolverOp memory _op) external payable nonReentrant whenNotPaused {
         if (_op.l2BatchMetaHash != 0) {
-            // Verify that the required L2 batch containing the intent transaction has been proposed
-            address taiko = resolve(LibNames.B_TAIKO, false);
-            if (!ITaiko(taiko).isInbox4()) revert VAULT_NOT_ON_L1();
+            if (block.chainid != L1_CHAIN_ID) revert VAULT_NOT_ON_L1();
 
+            address taiko = resolve(LibNames.B_TAIKO, false);
```
ggonzalez94: This will make testing and deploying to chains other than mainnet very difficult(because of chainId).
What if we do this instead and take advantage of the resolver:

```suggestion
           address taiko = resolve(LibNames.B_TAIKO, true);
            if (taiko == address(0)) revert VAULT_NOT_ON_L1();
```

This should work on L1 where the inbox is registered and fail on L2 where it isn't

ggonzalez94 (commented):
Pushed a commit to remove `isInbox4` from the new `AbstractInbox` and left a suggestion to simplify how to know if we are on L1 or not

packages/protocol/contracts/shared/tokenvault/ERC20Vault.sol
ggonzalez94: Btw I see we [removed the RollupResolver](https://github.com/taikoxyz/taiko-mono/pull/19419) from the codebase. Is the idea that things will get moved to a different resolver or are we just counting on using it(since it's been deployed), but not having it on the codebase anymore?

dantaik: Removed `if (block.chainid != L1_CHAIN_ID)` check.

dantaik: > Btw I see we [removed the RollupResolver](https://github.com/taikoxyz/taiko-mono/pull/19419) from the codebase. Is the idea that things will get moved to a different resolver or are we just counting on using it(since it's been deployed), but not having it on the codebase anymore?

The resolver can resolve an address using a (chainId, name) pair, making it suitable for contracts that need to reference cross-chain addresses. For local address resolution, we now rely on immutables instead.

ggonzalez94 (approved):
LGTM

PR #19780, opened 2025-07-23
feat(protocol): a proposed design for forced inclusions

ggonzalez94:
Still draft, creating this to get early feedback and iterate.

## Proposed Design
Simplify the preconf access control and forced inclusion logic.

This PR creates a unified `PreconfManager` contract that combines preconf access control and forced inclusion handling for the Shasta fork. It replaces the old `TaikoWrapper` + `PreconfRouter` pattern with a single, efficient gateway.

Instead of expecting separate parameters in the propose function, the proposer is expected to prepend the forced inclusion batch to their request, allowing this function to send it directly to the inbox, where the batches are unpacked and the validation can occur more naturally.

### Other options
- we could embed the entire forced inclusion logic(and even the preconf logic) on the inbox itself, but I don't think this would result in a cleaner design.
- we could  move to a single batch design and change the forced inclusion logic so that the batch includes the forced included txs, but does not need to be a separate batch. If we decide to go in that direction, this PR can still be a stepping stone for that.

## Implementation Details

1. **PreconfManager** – Single gateway contract that:
   - Validates preconfer authorization (whitelist or fallback)
   - Ensures forced inclusions are processed when due
   - Verifies the returned blob hash matches the expected forced inclusion

2. **Extended Return Values** – Modified interfaces to return forced inclusion blob hash:
   ```solidity
   function propose4(...) returns (Summary memory, bytes32 forcedInclusionBlobHash)
   ```

## TODOs

- [ ] decide if we like this design
- [ ] implement `_nextBatchId` that actually gets the info from the inbox
- [ ] review and write proper tests for the `PreconfManager` and `ForcedInclusionStore`
- [ ] under this design forced inclusions can only be processed when due, we can be more flexible and allow forced inclusions to be processed before if the proposer decides to


ggonzalez94:
Closing in favor of #19796

PR #19783, opened 2025-07-24
feat(taiko-client): remove an unnecessary call for beacon sync

davidtaikocha:
We can simply trigger a beacon sync by a FCU call.


YoGhurt111 (approved):

cyberhorsey:
why is this call unnecessary? reth doesn't support it and we need it removed?

davidtaikocha:
> why is this call unnecessary? reth doesn't support it and we need it removed?

Its unnecessary if you take a look at both geth and reth FCU source code, but we can still call it. It's also ok, the problem is our request body is incompatible with the spec, we left `executableData.transactions` as `nil`, it seems its ok to geth, but can't pass reth's check. So removing it will bring our calls closer to the spec. 

Another choice is we update this method to fill the `executableData.transactions` before sending the request.

But yeah, better @cyberhorsey @YoGhurt111 can double-check the `taiko-geth` source code.

dantaik:
@claude please review.

davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/pull/19860

PR #19786, opened 2025-07-24
feat(protocol): add a new role to remove operators in the PreconfWhitelist contract 

ggonzalez94:
### The Problem
Right now only operators can self eject or the owner of the contract has to do it. Since the owner is a multisig, this action is likely going to be slow. If an operator goes offline during this period, we want to eject them as soon as possible. @cyberhorsey has a software that monitors operators and can decide if they should be ejected.

### The Solution
Add a new role to the `PreconfWhitelist` contract that allows a new address(an EOA in this case) to remove operators.

- Add a new role and allow the owner to set it. This could have been an immutable variable, but: this gives us some flexibility to change who is assigned to the role(key compromise, new setup, etc.), and these are not gas sensitive operations(they should rarely happen)
- Add a script to deploy the new implementation. **IMPORTANT: Check the instructions on the script**
- Fix typo on the test file name


packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -24,23 +24,7 @@ interface IPreconfWhitelist {
...
-    /// lookahead,
-    ///      set _operatorChangeDelay >= _randomnessDelayEpochs. If you want to allow lookahead
-    ///      to be affected by operator changes (e.g., for emergency evictions), set
-    /// _operatorChangeDelay = 0.
-    function init(
-        address _owner,
-        uint8 _operatorChangeDelay,
-        uint8 _randomnessDelayEpochs
-    )
-        external;
```
ggonzalez94, line 43: I don't know why I remove this, but I guess it is kinda awkward that you have an initialized as part of the interface

ggonzalez94:
@cursor please review this PR

ggonzalez94:
@claude please review this PR

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
cyberhorsey: does this break anything by removing? no other contract depends on IPreconfWhitelist.init ,right?

cyberhorsey (dismissed):

cyberhorsey:
LGTM but should we make this PR off main an then cherry pick it, or do we just do it striaght to the protocol branch like this? im not sure how we make protocol changes to the branch tbh.

Wlil start the new service for this next week.

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
dantaik: We should remove init from interface :D

dantaik:
> LGTM but should we make this PR off main an then cherry pick it, or do we just do it striaght to the protocol branch like this? im not sure how we make protocol changes to the branch tbh.
> 
> Wlil start the new service for this next week.

We should apply to main then cherry pick.

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -127,6 +141,18 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist {
         emit Consolidated(_previousCount, _operatorCount, _havingPerfectOperators);
     }
 
+    /// @notice Sets the ejecter address.
+    /// @param _ejecter The new ejecter address.
+    function setEjecter(address _ejecter) external onlyOwner {
```
dantaik: Can we actually using a mapping and set multiple ejectors?

dantaik (dismissed):

ggonzalez94:
> We should apply to main then cherry pick.

I will cherry pick the commits then and open a new PR to main. As a general rule isn't it safer/easier to merge to the deployment branch and then merge that branch back to main

 > Can we actually using a mapping and set multiple ejectors?
 
 Making these changes now as well

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
ggonzalez94: updated

YoGhurt111 (approved):

PR #19788, opened 2025-07-25
feat(protocol): add new ejecter role to main branch

ggonzalez94:
Ref #19786

It also has a couple commits we had not merged into main from the deployment scripts


ggonzalez94:
@claude can you do a quick review/sanity check of this PR?

dantaik (approved):

packages/protocol/contracts/layer1/mainnet/MainnetInbox.sol
```diff
@@ -30,11 +30,12 @@ contract MainnetInbox is TaikoInbox {
...
-            batchRingBufferSize: batchRingBufferSize_,
-            maxBatchesToVerify: 16,
-            blockMaxGasLimit: 32_000_000,
-            livenessBond: 25e18, // 25 Taiko token per batch
+            maxUnverifiedBatches: 324_000, // DO NOT CHANGE!!!
+            batchRingBufferSize: 360_000, // DO NOT CHANGE!!!
+            maxBatchesToVerify: 8,
+            blockMaxGasLimit: 240_000_000,
+            livenessBondBase: 25e18, // 25 Taiko token per batch
+            livenessBondPerBlock: 0, // deprecated
```
YoGhurt111: Since the code in main branch is already shasta's code, so we don't need to cherry pick these config to main branch, just changing the value is enough.

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
```diff
@@ -0,0 +1,60 @@
...
+
+import "@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol";
+import "test/shared/DeployCapability.sol";
+import "src/layer1/preconf/impl/PreconfWhitelist.sol";
+import "src/layer1/preconf/impl/PreconfRouter.sol";
+import "src/layer1/forced-inclusion/TaikoWrapper.sol";
+import "src/layer1/mainnet/MainnetInbox.sol";
+import "src/layer1/fork-router/PacayaForkRouter.sol";
+
+contract DeployMainnetPreconf is DeployCapability {
```
YoGhurt111: I think we don't need the script anymore after we launch the preconf on mainnet, so I think it's ok to leave it in the `taiko-alethia-protocol-v2.3.1` branch.

packages/protocol/contracts/layer1/mainnet/MainnetInbox.sol
ggonzalez94: not sure I follow

packages/protocol/script/layer1/mainnet/DeployMainnetPreconf.s.sol
ggonzalez94: but we are supposed to merge this to main and then cherry pick to `taiko-alethia-protocol-v2.3.1` right? How do you suggest we handle this script? Feel free to just update the branch directly if it is easier

packages/protocol/script/layer1/based/DeployProtocolOnL1.s.sol
```diff
@@ -493,7 +493,7 @@ contract DeployProtocolOnL1 is DeployCapability {
             name: "preconf_whitelist",
             impl: address(new PreconfWhitelist()),
             data: abi.encodeCall(PreconfWhitelist.init, (owner, 2, 2)),
-            registerTo: address(0)
+            registerTo: rollupResolver
```
YoGhurt111:
```suggestion
            registerTo: address(0)
```

YoGhurt111 (approved):

PR #19798, opened 2025-07-27
(chore): broken image links in the bridging documentation by correcting relative paths

kilavvy:
##  Changes
- **File:** `packages/docs-site/src/content/docs/taiko-alethia-protocol/protocol-architecture/bridging.md`
- **Fixes:** 2 image path corrections
  - `bridging-source-chain.webp` - removed `~` prefix
  - `bridging-dest-chain.webp` - removed `~` prefix

##  Why
The `~` prefix was causing images to not display properly in the documentation. This change ensures images render correctly by using absolute paths from the root.

##  Testing
- Images now load correctly in documentation
- No breaking changes to existing functionality


RogerLamTd:
the images are displaying correctly already

PR #19803, opened 2025-07-28
feat(protocol): unify Proposed and SummaryUpdated

dantaik:
Also: if `propose4` is called as the outermost transaction, we can safely skip emitting its `inputs` calldata in the `SummaryUpdated` event to save gas.

### DIFF
--- a/packages/protocol/contracts/layer1/based2/AbstractInbox.sol
+++ b/packages/protocol/contracts/layer1/based2/AbstractInbox.sol
@@ -78,6 +78,8 @@ abstract contract AbstractInbox is EssentialContract, IInbox, IPropose, IProve {
         _saveSummaryHash(keccak256(abi.encode(summary)));
 
         // Skip calldata emission for gas optimization when directly called by EOA
+        // TODO: can we call PreconfRouter from within this contract, instead of using the
+        // PreconfRouter as the msg.sender?
         emit Proposed(
             lastProposedBatchId,
             _isOuterMostTransaction() ? bytes("") : _inputs,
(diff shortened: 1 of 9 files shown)


dantaik:
@claude please review.

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -68,13 +68,22 @@ abstract contract AbstractInbox is EssentialContract, IInbox, IPropose, IProve {
...
         // Verify batches
         summary = LibVerify.verify(bindings, config, summary, transitionMetas);
 
         _saveSummaryHash(keccak256(abi.encode(summary)));
-        emit SummaryUpdated(bindings.encodeSummary(summary));
+
+        // Skip calldata emission for gas optimization when directly called by EOA
+        emit Proposed(
+            lastProposedBatchId,
+            _isOuterMostTransaction() ? bytes("") : _inputs,
```
dantaik, line 85: I now realized there is a wrapper: PreconfRouter. So this trick doesn't work unless we redesign the router such that router works as a permission checker, not an inbox wrapper. @ggonzalez94

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
```diff
@@ -68,13 +68,24 @@ abstract contract AbstractInbox is EssentialContract, IInbox, IPropose, IProve {
...
+        uint48 lastProposedBatchId = summary.nextBatchId - 1;
 
         // Verify batches
         summary = LibVerify.verify(bindings, config, summary, transitionMetas);
 
         _saveSummaryHash(keccak256(abi.encode(summary)));
-        emit SummaryUpdated(bindings.encodeSummary(summary));
+
+        // Skip calldata emission for gas optimization when directly called by EOA
+        // TODO: can we call PreconfRouter from within this contract, instead of using the
```
dantaik:
```suggestion
        // TODO: can we call IPreconfWhitelist or LookAhead from within this contract, instead of using the
```

packages/protocol/contracts/layer1/based2/AbstractInbox.sol
dantaik: See https://github.com/taikoxyz/taiko-mono/pull/19812

PR #19808, opened 2025-07-28
fix(protocol): fix a bug in forced inclusion proposer

dantaik:
In forced inclusion, `params.proposer` must always be set to address(0). Then, within the inbox contract, it is reassigned to `msg.sender`.

@Claude @Cursor please review.
@gavin please cherrypick this PR to the preconf release.


packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -92,19 +92,18 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
             BatchParams memory params = abi.decode(_params, (BatchParams));
 
             {
-                if (inboxWrapper == address(0)) {
-                    if (params.proposer == address(0)) {
-                        params.proposer = msg.sender;
-                    } else {
-                        require(params.proposer == msg.sender, CustomProposerNotAllowed());
-                    }
+                if (params.proposer == address(0)) {
```
AnshuJalan, line 95: @dantaik I think the issue is fixable on the taiko-wrapper level. Forced inclusion batches are going via taiko-wrapper, so the `msg.sender` here is the taiko wrapper, and we cannot debit bond from the wrapper. 

May be enforcing `params.proposer` in taiko-wrapper's `_validateForcedInclusionParams(..)` can do the trick, with no changes in the inbox

dantaik: Ah I see.

PR #19810, opened 2025-07-28
fix(protocol): validate proposer in TaikoWrapper for forced inclussions

ggonzalez94:
## The issue

> I think when forced including blocks, the proposer who adds a forced inclusions can assign anyone to be the params.proposer and making them pay the liveness bond. This is because the check that enforces the value of params.proposer is in the  preconf router, but the forced inclusions go directly through the wrapper.

## The solution

This PR fixes the issue by making sure the proposer in both regular and forced included batches is the same. It relies on the upstream validation of the `PreconfRouter` for who the proposer should be.

TODO's:

- [ ] If we rely on the validation of the router, we should not allow this contract to function alone(change the `onlyFromOptional` modifier to `onlyFrom`)
- [ ] Add tests


AnshuJalan (approved):
LGTM

dantaik (approved):

PR #19813, opened 2025-07-28
chore(protocol): make sure the wrapper is not used alone

ggonzalez94:
Follow up to #19810

The `TaikoWrapper` should never be used without the `PreconfRouter`, otherwise we still have the vulnerability that any proposer can be set since the validation is skipped in the inbox(when using a wrapper) and is only done on the router.
We also discussed that it does not make sense to have the wrapper without preconfs, since you do not need forced inclusions if you do not have preconfs.
This PR makes sure the wrapper is never used without a router for additional security.

I also wanted to add some tests, but the current setup is using mocks heavily, so the change would be pretty large. Since we are changing the protocol design in shasta, let's just work on the tests there.


ggonzalez94:
@claude please review this PR

dantaik (approved):

smtmfft (approved):

PR #19814, opened 2025-07-28
fix(bridge-ui): fix issue where TokenInput quota validation always returning false

MozirDmitriy:
Replace reactive statement that was overwriting quota validation result

The `$: exceeds = false;` reactive block was resetting the exceeds variable 
after it was set by checkIfAmountExceedsQuota(), causing quota validation 
to always return false. Changed to regular variable declaration to fix this.


KorbinianK (approved):

dantaik:
@claude please review

dantaik (approved):

PR #19833, opened 2025-07-29
chore(protocol): remove return-values from propose4()

dantaik:
## Summary
  - Removed Summary return value from propose4() function in TaikoInbox
  - Affects AbstractInbox.sol, IInbox.sol, and IPropose.sol

## Rationale
  The propose4() function was returning a Summary struct that duplicates data already emitted in the Proposed event. Since:
  - No external contracts or off-chain code currently use this return value
  - Events are the standard pattern for protocol state changes
  - Discourage another contracts from calling propose4() and promote EOA to call it.


dantaik:
@claude please review with concise feedback

PR #19847, opened 2025-07-30
feat(protocol): prototyping a very different Shasta protocol

### LINKED ISSUE #19838: Move Calldata Proposal Parameters into Blobs
Currently, many batch and block parameters are submitted as calldata for validation within the protocol. This approach contributes significantly to the L1 proposal cost and requires on-chain Solidity logic for parameter validation.

This PR moves these parameters into blobs instead. The driver will take responsibility for validating them. In principle, any data verifiable in Solidity can also be verified by the prover, based on the parent L1 block’s state and header.

By offloading parameter validation from Solidity to the prover and reducing calldata size, this change significantly lowers L1 proposal costs and simplifies on-chain logic.

dantaik:
I’ll draft a document to walk through this PR. I’m genuinely excited about the direction, even though it will make driver and node implementation more challenging.

Our goal is get most code implemented by **Aug 20th**.
It will be great if someone can claim and work on the following breakdowns:

## High priorities
- [ ] Implement L2 contracts @dantaik WIP
- [x] Implement SyncedBlockManager @dantaik  WIP
- [x] Implement InboxStateManager using ring buffer @dantaik WIP
- [ ] Integrate `PreconfWhitelist` with the new design @ggonzalez94 
    -  https://github.com/taikoxyz/taiko-mono/pull/19879
- [ ] Implement forced inclusions @ggonzalez94 
    - https://github.com/taikoxyz/taiko-mono/pull/19881 WIP
- [ ] Implement BondManager. On L1, there is a nonzero exit-to-withdrawal delay; on L2, such delay can be configured to 0. But L2 and L1 share the same code. @ggonzalez94 
- [ ] Implement multi-proposal claim - same as in based2/libs/LibVerify.sol and based2/libs/LibProve.sol
- [ ] New Anchor to support IBlockHashProvider

## Low priorities
- [ ] Change some new contracts to use EssentialContract for upgradability
- [ ] Write docs for proving service - wait for L2 contracts code-readiness.
- [ ] Ensure worst case zk-provabability (gas limit per block, max number of blocks in a proposal, max proposals per L1 second/block)
- [ ] Implement same-slot messaging

## Do not work on the following:
- [ ] Implement LibCodec (wait for structs and functions to finalize)


dantaik:
@claude please review

packages/protocol/contracts/layer1/based3/ShastaInbox.sol
```diff
@@ -0,0 +1,150 @@
...
+        store.initialize();
+    }
+
+    // -------------------------------------------------------------------------
+    // External Transactional Functions
+    // -------------------------------------------------------------------------
+
+    /// @notice Proposes a new proposal of L2 blocks
+    /// @param _blobIndex Index of the blob in the current transaction
+    function propose(uint48 _blobIndex) external {
```
Brechtpd: Now that revert protection isn't required anymore with preconfs, you can actually do nothing here, the proposal can be found back by traversing the transaction trees of previous blocks in the prover. This is how Optimism proposes blocks and this is I think the most efficient way you can get things. Proposing is just data.

If there are some L1 things that need to be checked (for example for L1CALL support) then of course this does need to happen here for revert protection to work. But otherwise this can do nothing.

packages/protocol/contracts/layer1/based3/offchain-specs/BlockPreparer.py
```diff
@@ -0,0 +1,137 @@
...
+        # Timestamp calculations with constraints
+        timestamp = max(blk.timestamp, parent_block_header.timestamp + self.MIN_L2_BLOCK_TIME)
+        timestamp = min(timestamp, reference_block_header.timestamp + self.L1_BLOCK_TIME)
+        timestamp = min(timestamp, reference_block_header.timestamp + self.MAX_BLOCK_TIMESTAMP_OFFSET)
+        
+        # Update block timestamp
+        blk.timestamp = timestamp
+        
+        # Gas calculations based on block time
+        block_time = timestamp - parent_block_header.timestamp
```
davidtaikocha: Seems under the current `timestamp calculations`, a proposer can manage to create some `block_time == 0 && gas_limit == 0` blocks at the end of each proposal? Which will at least break the classic `EIP-4396` calculation imo.

packages/protocol/contracts/layer1/based3/ShastaInbox.sol
dantaik: We still have to process bonds on L1 unfortunately.

dantaik:
@claude please review.
@ggonzalez94 @Brechtpd @davidtaikocha @linoscope @smtmfft please take a look. 

I also want briefly go over the code and my ideas in our next Monday's team meeting.

packages/protocol/contracts/layer1/based3/ShastaInbox.sol
Brechtpd: I think we can be a bit creative here! No reason these transfers cannot happen on L2 I think:
- For each block with system call: transfer bond from prover to protocol on L2 (signature is checked etc...)
- Block is eventually proven somehow, if it doesn't have to be returned to the original owner some dispute info is stored somewhere (a lot depends on how blocks are proven here so leaving it open  how that is done).
- On L2: original prover can withdraw its bond back after N blocks or S seconds, unless someone withdraws it first by showing it was disputed

So if L1 reads are not cheap (which is currently the case for Alethia), in the normal case the bond is returned cheaply, if it disputed it may be a bit more expensive but that's fine because normally doesn't happen. If a prover wants to withdraw faster than the fixed time that would also be possible similar to the dispute mechanism.

Needing the taiko tokens on L2 could have some "bootstrap problem" because the tokens have to get there somehow but I think alethia is live for long enough that shouldn't be a problem anymore.

Btw the main reason I like not storing any information onchain for proposing is that you have simple and very efficient proof aggregation by just running over ranges of L1 blocks. Initially gwyneth worked like this but I dropped it because for sync composability with L1 it can't really work like this.

packages/protocol/contracts/layer1/based3/ShastaInbox.sol
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    // -------------------------------------------------------------------------
+    // External Transactional Functions
+    // -------------------------------------------------------------------------
+
+    /// @inheritdoc IShastaInbox
+    function propose(BlobLocator[] memory _blobLocators) external {
+        for (uint48 i; i < _blobLocators.length; ++i) {
+            _propose(_validateBlockLocator(_blobLocators[i]));
```
davidtaikocha: Looks like `_validateBlockLocator` doesn't support `forcedInclusion` anymore (it's only able to verify blobs in the current transaction), so seems one `BlobLocator` is enough to locate all available blobs for each transaction?

packages/protocol/contracts/layer1/based3/ShastaInbox.sol
```diff
@@ -0,0 +1,197 @@
...
+
+    /// @notice Proposes a new proposal of L2 blocks
+    /// @param _content The content of the proposal
+    function _propose(BlobSegment memory _content) private {
+        uint48 proposalId = store.incrementAndGetProposalId();
+
+        // Create a new proposal.
+        // Note that the contentHash is not checked here to empty proposal data.
+        uint48 proposedBlockTimestamp = uint48(block.timestamp - 128);
+        uint48 proposedBlockNumber = uint48(block.number - 1);
```
davidtaikocha: Shall we set these two values to:
```suggestion
        uint48 proposedBlockTimestamp = uint48(block.timestamp);
        uint48 proposedBlockNumber = uint48(block.number);
```

And change the `referenceBlockHash` below from:
```solidity
referenceBlockHash: blockhash(proposedBlockNumber),
```
to:
```solidity
referenceBlockHash: blockhash(proposedBlockNumber - 1),
```

packages/protocol/contracts/layer1/based3/DESIGN.md
```diff
@@ -0,0 +1,154 @@
...
+When blob data is invalid, the system gracefully degrades to a default specification with empty blocks and zero values, ensuring the protocol remains operational even under adverse conditions. This resilience is crucial for maintaining liveness in a decentralized environment where data availability cannot be guaranteed.
+
+### Protocol System Calls
+
+The rollup client executes system calls during block production that do not consume gas, enabling protocol-level operations without impacting user transactions. The block head call executes before the first transaction in each block, validating and setting block parameters. This call:
+
+- Ensures timestamps fall within acceptable ranges based on parent and reference blocks
+- Calculates prevRandao by mixing the block number with parent randomness
+- Determines fee recipients with fallback to the DAO treasury for zero addresses
+- Updates anchor block information when valid new anchors are provided
```
davidtaikocha: IMO the system call is designed to only update the storage ( `Updates anchor block information when valid new anchors are provided` ) but not validations, all validations should be done by `driver` software and pass the final fields for the block to the node.

packages/protocol/contracts/layer1/based3/DESIGN.md
```diff
@@ -0,0 +1,154 @@
...
+
+The proposal object emitted on L1 contains essential metadata for block construction and verification. The reference block hash enables L1 state verification through Merkle proofs but does not directly affect L2 block world states, providing flexibility in proof generation while maintaining security. The proposal content, encoded within EIP-4844 blobs, specifies the actual execution parameters including transaction lists, block timestamps, fee recipients, and gas issuance updates.
+
+When blob data is invalid, the system gracefully degrades to a default specification with empty blocks and zero values, ensuring the protocol remains operational even under adverse conditions. This resilience is crucial for maintaining liveness in a decentralized environment where data availability cannot be guaranteed.
+
+### Protocol System Calls
```
davidtaikocha: And we can later upgrade the `TaikoAnchor` contract (or a new L2 contract) in `based3/`, but ensure only a special `msg.sender` is able to call it, like `0xfffffffffffffffffffffffffffffffffffffffe` for [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935).

packages/protocol/contracts/layer1/based3/DESIGN.md
```diff
@@ -0,0 +1,154 @@
...
+
+- Ensures timestamps fall within acceptable ranges based on parent and reference blocks
+- Calculates prevRandao by mixing the block number with parent randomness
+- Determines fee recipients with fallback to the DAO treasury for zero addresses
+- Updates anchor block information when valid new anchors are provided
+
+The timestamp validation ensures blocks cannot be timestamped in the future relative to L1, maintains monotonic time progression, and prevents excessive backdating beyond 128 seconds. The anchor block mechanism allows L2 to reference recent L1 state while preventing reference to blocks that are too old or potentially reorganized.
+
+The block tail call executes after the last transaction in each block, updating protocol state based on execution results. It calculates gas excess using EIP-1559 principles where excess increases with issuance and decreases with usage. On the final block of each proposal, it may update the gas issuance rate within strict bounds of 99% to 101% of the current rate, preventing dramatic economic changes while allowing gradual adaptation to network conditions.
```
davidtaikocha: > On the final block of each proposal, it may update the gas issuance rate within strict bounds of 99% to 101% of the current rate, preventing dramatic economic changes while allowing gradual adaptation to network conditions.

We may have to use `block.extraData` to `mark the final block of each proposal + the new gas_issuance_per_second` instead of always setting it to empty, or it might won't work when p2p syncing / preconfirmation.

packages/protocol/contracts/layer1/based3/DESIGN.md
```diff
@@ -0,0 +1,154 @@
...
+
+Gas management operates at multiple levels. Each block has a fixed limit of 100 million gas, providing predictable capacity for users and applications. The protocol tracks gas excess using EIP-1559 principles, where excess increases with issuance and decreases with usage, creating a self-balancing system. The target utilization is approximately 50% of the gas limit, with base fees adjusting to maintain this equilibrium over time.
+
+Base fee calculations follow standard EIP-1559 formulas but adapted for the protocol's variable block times. Gas excess updates occur in the block tail call after measuring actual gas usage, ensuring fees reflect real network demand. The base fee derives from gas excess, creating price signals that guide users toward efficient resource utilization.
```
davidtaikocha: Shall we skip all empty blocks (`gasUsed=0`) caused by invalid blob encoding here? But seems this will cause the base fee calculation may depend not only on the block's parent...

packages/protocol/contracts/layer1/based3/ShastaInbox.sol
```diff
@@ -0,0 +1,197 @@
...
+            store.setClaimRecordHash(proposal.id, claim.parentClaimHash, claimRecordHash);
+            emit Proved(proposal.id, proposal, claimRecord);
+        }
+
+        bytes32 claimsHash = keccak256(abi.encode(_claims));
+        verifyProof(claimsHash, _proof);
+    }
+
+    /// @inheritdoc IShastaInbox
+    function finalize(ClaimRecord[] memory _claimRecords) external {
```
davidtaikocha: Seems it won't be called in `propose()` anymore, then any motivation for anyone to call this method? Looks like can't find it in `DESIGN.md`.

packages/protocol/contracts/layer1/based3/offchain-specs/BlockCalls.py
```diff
@@ -0,0 +1,199 @@
...
+            proto_state.gas_issuance_per_second,
+            timestamp - parent_timestamp,
+            gas_used,
+        )
+
+        # the following code runs only once per proposal at the very end
+        if proto_state.block_index == batch_size - 1:
+            proto_state.gas_issuance_per_second = (
+                self._calculate_gas_issuance_per_second(
+                    proto_state.gas_issuance_per_second,
```
davidtaikocha:
```suggestion
                    current_gas_issuance_per_second,
```

packages/protocol/contracts/layer1/shasta/offchain-specs/BlockCalls.py
```diff
@@ -0,0 +1,287 @@
...
+        # provable: -> content -> proposal_hash
+        block_args: BlockArgs,
+        # provable: -> parent_block_number -> parent_block_hash
+        block_number: int,
+        # provable: -> parent_block_hash
+        parent_prev_randao: str,
+        # provable: -> parent_block_hash
+        parent_timestamp: int,
+        # provable: -> up to 128 ancester block headers -> anchor_block_hash
+        anchor_block_hash: str,
```
linoscope, line 42: How do we expect the validity of the `anchor_block_hash` to be checked? I expected that, for example, we:
- We call `blockhash(proposedBlockNumber)` and store the result in `struct Proposal`.
- At proof generation time, we pass the `blockhashes` of required L1 blocks as private input, and check that they form a chain up to `blockhash(proposedBlockNumber)` .

But looks like we never call `blockhash` in inbox contract, so maybe you had something else in mind?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,362 @@
...
+        inboxStateManager.setProposalHash(proposalId, proposalHash);
+
+        emit Proposed(proposal);
+        return (_coreState);
+    }
+
+    function _prove(Proposal memory _proposal, Claim memory _claim) private {
+        bytes32 proposalHash = keccak256(abi.encode(_proposal));
+        if (proposalHash != _claim.proposalHash) revert ProposalHashMismatch();
+        if (proposalHash != inboxStateManager.getProposalHash(_proposal.id)) {
```
linoscope, line 191: Would it be possible to free the storage slot for `_proposal.id` to get gas refunds here?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,362 @@
...
+        uint48 timestamp = uint48(block.timestamp);
+        uint48 referenceBlockNumber = uint48(block.number);
+
+        Proposal memory proposal = Proposal({
+            id: proposalId,
+            proposer: msg.sender,
+            provabilityBond: provabilityBond,
+            livenessBond: livenessBond,
+            timestamp: timestamp,
+            proposedBlockNumber: referenceBlockNumber,
```
linoscope: Nit: Just for consistency in naming:

```suggestion
        uint48 proposedBlockNumber = uint48(block.number);

        Proposal memory proposal = Proposal({
            id: proposalId,
            proposer: msg.sender,
            provabilityBond: provabilityBond,
            livenessBond: livenessBond,
            timestamp: timestamp,
            proposedBlockNumber: proposedBlockNumber,
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,362 @@
...
+    /// @return coreState_ The updated core state
+    /// @return syncedBlock_ The synced block information
+    function _finalize(
+        CoreState memory _coreState,
+        ClaimRecord[] memory _claimRecords
+    )
+        private
+        returns (CoreState memory coreState_, ISyncedBlockManager.SyncedBlock memory syncedBlock_)
+    {
+        if (keccak256(abi.encode(_coreState)) != inboxStateManager.getCoreStateHash()) {
```
linoscope: I think this will always reverts as it validates that `_coreState` matches the stored state, but `_coreState` has already been modified by `_propose` (which increments `nextProposalId`) without being stored to state.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    /// @dev Finalizes proposals by verifying claim records and updating state.
+    /// @param _coreState The current core state.
+    /// @param _claimRecords The claim records to finalize.
+    /// @return coreState_ The updated core state
+    /// @return syncedBlock_ The synced block information
+    function _finalize(
+        CoreState memory _coreState,
+        ClaimRecord[] memory _claimRecords
```
linoscope, line 220: Can't the propose always set `_claimRecords` to be empty to cut gas costs?

packages/protocol/contracts/layer2/shasta/iface/IAnchor.sol
```diff
@@ -0,0 +1,47 @@
...
+pragma solidity ^0.8.24;
+
+/// @title IAnchor
+/// @notice Interface for the Anchor contract that manages L2 state synchronization with L1
+/// @dev This contract stores critical state information for L2 block production and gas management
+/// @custom:security-contact security@taiko.xyz
+interface IAnchor {
+    /// @notice State structure containing L2 synchronization data
+    /// @dev Packed struct to optimize storage usage
+    struct State {
```
davidtaikocha, line 11: If i remember correctly, our post-whitelist preconfirmation slashing design still relies on the `_blockhashes` map in the old `TaikoAnchor` contract, shall we also keep this map in this new contract (or we can create another system contract)? (another question will be how to do the data migration)

linoscope: Yes, that mapping would be needed for preconf slashing to work.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
ggonzalez94: I believe we probably don't need this check since we are already passing the calculated `coreState` to this function

packages/protocol/contracts/layer1/shasta/offchain-specs/BlockCalls.py
```diff
@@ -0,0 +1,287 @@
...
+        #  provable: -> anchor_block_hash
+        expected_anchor_bond_credits_hash: str,
+        # provable: -> expected_anchor_bond_credits_hash
+        bond_credit_ops: List[Tuple[Address, int]],
+    ) -> Tuple[int, str, Address, int, int]:
+        """
+        System call invoked before the the first transaction in every block.
+        This call does not consume gas.
+
+        Returns:
```
matthiasgoergens, line 52: This mentions six values to be returned, but the type (and code) only mention five?

I would suggest using perhaps a named tuple or even a dataclass to group the returned value in, so that the type can self-document the return in a machine checkable way.

Here it seems that `designated_prover` is mentioned in the comment, but not actually returned.  (I don't know which way it's intended.  But I suspect that the comment is probably wrong, and not the code?)

packages/protocol/contracts/layer1/shasta/offchain-specs/BlockCalls.py
```diff
@@ -0,0 +1,287 @@
...
+        """
+        Calculate gas issuance per second
+        """
+        if new_gas_issuance_per_second == 0:
+            return current_gas_issuance_per_second
+
+        min = (
+            current_gas_issuance_per_second
+            * self.GAS_ISSUANCE_PER_SECOND_MIN_OFFSET
+            / 100
```
matthiasgoergens, line 174: I would suggest folding these kinds of 'units' (looks it's meant to be a percentage?) into the constants themselves, instead of littering them throughout the code.

It's just a lot less error-prone in the longer run.

packages/protocol/contracts/layer1/shasta/offchain-specs/BlockCalls.py
```diff
@@ -0,0 +1,287 @@
...
+        gas_issuance = current_gas_issuance_per_second * block_time
+        return max(current_gas_excess + gas_issuance - gas_used, 0)
+
+    def _calculate_gas_issuance_per_second(
+        self, current_gas_issuance_per_second: int, new_gas_issuance_per_second: int
+    ) -> int:
+        """
+        Calculate gas issuance per second
+        """
+        if new_gas_issuance_per_second == 0:
```
matthiasgoergens, line 168: This seems like a weird special case.  I would suggest using `None` or similar to mark this case, instead of declaring one value to be special.  Less likely to bite you in the butt later.

packages/protocol/contracts/layer1/shasta/offchain-specs/BlockCalls.py
```diff
@@ -0,0 +1,287 @@
...
+    ) -> Address:
+        """
+        Calculate the designated prover for the proposal
+        """
+        # Recover prover address from signature
+        try:
+            # Hash the proposal data using keccak256
+            # TODO: Implement proper ABI encoding for the proposal
+            # For now, create a simple hash of the proposal data
+            proposal_str = f"{proposal.id}{str(proposal.proposer)}{str(proposal.prover)}{proposal.provability_bond}{proposal.liveness_bond}{proposal.reference_block_timestamp}{proposal.reference_block_number}"
```
matthiasgoergens, line 223: This might be simpler, and you have to worry less about making your projection injective?:

```suggestion
            proposal_str = str((proposal.id, str(proposal.proposer), str(proposal.prover), proposal.provability_bond,proposal.liveness_bond,proposal.reference_block_timestamp, proposal.reference_block_number))"
```

(Modulo formatting.  You probably want to break this into multiple lines.)

packages/protocol/contracts/layer1/shasta/offchain-specs/BlockCalls.py
```diff
@@ -0,0 +1,287 @@
...
+            proposal_data = proposal_str.encode("utf-8")
+            message_hash = keccak(proposal_data)
+
+            # Recover the address from the signature
+            account = Account.recover_message(
+                message_hash, signature=proposal_content.prover_signature
+            )
+            prover = cast(Address, account)
+        except Exception:
```
matthiasgoergens, line 232: If you don't care about which exception you get, might as well just make it like this?:
```suggestion
        except:
```

If your linter doesn't like this, perhaps it's a sign that you should listen to?

packages/protocol/contracts/layer1/shasta/offchain-specs/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+To verify syntax without running the code:
+
+```bash
+python3 -m py_compile *.py
+```
+
+### Type Checking (Optional)
+
+If using type hints, run mypy for static type checking:
```
matthiasgoergens, line 33: You are using type hints, so no need to write a conditional here (and in the section title).

packages/protocol/contracts/layer1/shasta/offchain-specs/requirements.txt
```diff
@@ -0,0 +1,3 @@
+eth-typing>=3.0.0
+mypy>=1.0.0
+black
```
matthiasgoergens, line 3: It's usually a good idea to end every line with a newline character.

packages/protocol/contracts/layer1/shasta/offchain-specs/Types.py
```diff
@@ -0,0 +1,71 @@
...
+    content: BlobSegment
+
+
+## --- Offchain  ---
+@dataclass
+class Transaction:
+    """Represents a transaction in the system."""
+
+    to: Address
+    value: int
```
matthiasgoergens, line 36: Is any int fine, or is it restricted to a specific number of bits?

packages/protocol/contracts/layer1/shasta/offchain-specs/Types.py
```diff
@@ -0,0 +1,71 @@
...
+from eth_typing import Address
+
+
+## --- Onchain ---
+@dataclass
+class BlobSegment:
+    """Represents a blob segment from IInbox."""
+
+    blobHashes: List[str]  # bytes32[] in Solidity
+    offset: int  # uint32 in Solidity
```
matthiasgoergens, line 12: Here and elsewhere you could use a more specific type to reflect what solidity is doing?  (But not sure, if it's worth it.)

matthiasgoergens (commented):
Just a few comments and questions.  Feel free to ignore.

packages/protocol/contracts/layer1/based3/ShastaInbox.sol
dantaik: Thank you, I had similar ideas and have them implemented in the current code.

YoGhurt111 (approved):

PR #19855, opened 2025-07-31
feat(protocol): support multiple blobs for new shasta design

ggonzalez94:
This is a small PR to support multiple blobs when proposing in the new shasta design.

@dantaik I saw you mentioned in your `Design.md` dock that if no blobs are provided or `blockhash(blobIndex)` is invalid the node will replace this with an empty block? Does this mean this will be enforced in the derivation pipeline and we can get rid of the validations I added?
Even if that's the case not sure if it's worth given how cheap those two ifs will be.


dantaik:
@ggonzalez94 please review the my PR again as I handled it differently.

PR #19857, opened 2025-08-01
feat(protocol): the script of upgrading prover set for preconf on Hekla

YoGhurt111:


dantaik (approved):

packages/protocol/script/layer1/hekla/UpgradeHeklaProverSet.s.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import "@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol";
+import "test/shared/DeployCapability.sol";
+import "contracts/layer1/provers/ProverSet.sol";
```
YoGhurt111:
```suggestion
import "src/layer1/provers/ProverSet.sol";
```

PR #19858, opened 2025-08-01
feat(protocol): prepare the calldata for mainnet preconf

YoGhurt111:


dantaik (approved):

packages/protocol/script/layer1/mainnet/DeployWLAndWrapper.s.sol
```diff
@@ -0,0 +1,61 @@
...
+    function run() external broadcast {
+        address taikoInbox = 0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a;
+        address store = 0x05d88855361808fA1d7fc28084Ef3fCa191c4e03;
+        address router = 0xD5AA0e20e8A6e9b04F080Cf8797410fafAa9688a;
+        address rollupResolver = 0x5A982Fb1818c22744f5d7D36D0C4c9f61937b33a;
+        address taikoToken = 0x10dea67478c5F8C5E2D90e5E9B26dBe60c54d800;
+        address sequencer = 0x5F62d006C10C009ff50C878Cd6157aC861C99990;
+        address ejecter = address(0); // TODO: need the ejecter address
+
+        address whitelist = address(new PreconfWhitelist());
```
ggonzalez94: I believe we already have a script that deploys the whitelist contract that we haven't run yet. Can't we just use that one?
https://github.com/taikoxyz/taiko-mono/blob/taiko-alethia-protocol-v2.3.1/packages/protocol/script/layer1/mainnet/DeployWhitelistWithEjecter.sol

ggonzalez94: this one already outputs the calldata, so I'm fine if we remove the old script and use this one instead if you prefer. But let's not keep both since it is confusing

packages/protocol/script/layer1/mainnet/DeployWLAndWrapper.s.sol
```diff
@@ -0,0 +1,61 @@
...
+            "Add operator calldata: ",
+            abi.encodeCall(PreconfWhitelist.addOperator, (sequencer, sequencer))
+        );
+
+        address wrapper = address(new TaikoWrapper(taikoInbox, store, router));
+        console2.log(
+            "Upgrading wrapper calldata: ", abi.encodeCall(UUPSUpgradeable.upgradeTo, (wrapper))
+        );
+
+        address proverSetImpl = new ProverSet(rollupResolver, taikoInbox, taikoToken, router);
```
ggonzalez94: we haven't modified the prover set AFAIK. Do we need to deploy it again?

packages/protocol/script/layer1/mainnet/DeployWLAndWrapper.s.sol
```diff
@@ -0,0 +1,61 @@
...
+    modifier broadcast() {
+        require(privateKey != 0, "invalid private key");
+        vm.startBroadcast(privateKey);
+        _;
+        vm.stopBroadcast();
+    }
+
+    function run() external broadcast {
+        address taikoInbox = 0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a;
+        address store = 0x05d88855361808fA1d7fc28084Ef3fCa191c4e03;
```
ggonzalez94: nit on naming for clarity: 
```suggestion
        address forcedInclusionStore = 0x05d88855361808fA1d7fc28084Ef3fCa191c4e03;
```

ggonzalez94 (commented):
Left a few comments and questions

packages/protocol/script/layer1/mainnet/DeployWLAndWrapper.s.sol
YoGhurt111: We need to set the entrypoint  in ProverSet as PreconfRouter, previously this value was TaikoWrapper.

packages/protocol/script/layer1/mainnet/DeployWLAndWrapper.s.sol
YoGhurt111: We can use the [DeployWhitelistWithEjecter.sol](https://github.com/taikoxyz/taiko-mono/blob/taiko-alethia-protocol-v2.3.1/packages/protocol/script/layer1/mainnet/DeployWhitelistWithEjecter.sol).
Just to put all the required calldata together for easy review.

packages/protocol/script/layer1/mainnet/DeployWLAndWrapper.s.sol
YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/19858/commits/717f2db880b155e92ba863c8fd5817c5c8e06599

packages/protocol/script/layer1/mainnet/DeployWLAndWrapper.s.sol
YoGhurt111: And this value only can be passed in constructor, so we need to deploy a new impl to update the value.

packages/protocol/script/layer1/mainnet/DeployWLAndWrapper.s.sol
ggonzalez94: shall we remove the other script then?

ggonzalez94 (approved):

packages/protocol/script/layer1/mainnet/DeployWLAndWrapper.s.sol
YoGhurt111: Removed.

PR #19881, opened 2025-08-05
feat(protocol): incorporate forced inclusions in new shasta design

ggonzalez94:
Borrows a lot of structure from the old `ForcedInclusionStore` but simplified for the new shasta

### Design principles
- It is ok if including a forced inclusion is more expensive(we store more data)
- Proposing should continue to be as cheap as possible
- Inclusion delay is now measured in seconds instead of batches, since we don't store the `nextBatchId` in the inbox anymore(only the hash of `CoreState` is stored)

### Implementation
- Created a new `IForcedInclusionStore` and `ForcedInclusionStore` in the new directory since the interface and data structure are slightly incompatible with the old one
- Validates due inclusions in the Inbox


ggonzalez94:
@claude please review

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -119,10 +127,26 @@ contract Inbox is IInbox {
...
+            (coreState, forcedInclusionProposal) =
+                _processForcedInclusion(coreState, forcedInclusionFrame);
+        } else {
+            // Ensure no forced inclusion is due when none is provided
+            _ensureNoForcedInclusionDue();
+        }
+
+        // Build proposals array
+        Proposal[] memory proposals =
+            _buildProposalsArray(proposal, forcedInclusionProposal, hasForcedInclusion);
```
dantaik, line 149:
```suggestion
            _buildProposalsArray(proposal, forcedInclusionProposal, true);
```

dantaik (approved):

PR #19882, opened 2025-08-06
feat(protocol): design and implement Shasta upgrade

### LINKED ISSUE #19838: Move Calldata Proposal Parameters into Blobs

dantaik:
This branch contains the new Shasta protocol design and implementation. Our goal is to complete the majority of the code by **August 20th**.

### 🔺 High Priorities

- [x] Implement `SyncedBlockManager` – @dantaik 
- [x] Implement `InboxStateManager` using ring buffer – @dantaik 
- [x] Refactor new contracts to use `EssentialContract` for upgradability @dantaik - https://github.com/taikoxyz/taiko-mono/pull/19890
- [x] Integrate `PreconfWhitelist` with the new design – @ggonzalez94  → https://github.com/taikoxyz/taiko-mono/pull/19879
- [x] Implement forced inclusions – @ggonzalez94 (WIP)  → https://github.com/taikoxyz/taiko-mono/pull/19881
- [x] Add new `Anchor` to support `IBlockHashProvider` @dantaik 
- [x] Implement L2 contracts & write documentation regarding block building pipeline  – @dantaik (WIP) → https://github.com/taikoxyz/taiko-mono/pull/19908
- [x] Implement `BondManager` – @ggonzalez94  → L1 has a nonzero exit-to-withdrawal delay; L2 can configure this delay to 0, but shares the same code
    - @dantaik I worked on two options, but I think the time based one is still preferable
        - #19904: uses finalization checks to ensure the proposer has no unfinalized proposals before withdrawing
        - #19905: uses a time window before allowing the proposer to withdraw
- [x] Allow proposer to specify a deadline after which the proposal should fail. With preconfirmation, landing a proposal too late will ruin the next preconfer's handling of their own proposals.
- [x] Writing L1 contract tests (using Claude Code, suite1) @dantaik  →  https://github.com/taikoxyz/taiko-mono/pull/19912
- [x] implement fees for (and provability bond in) forced inclusions #19945 @ggonzalez94 
- [ ] SignalService shall interact with ISyncedBlockManager (remove LibSignals.sol and its constants) and maybe disable multihop?
- [x] implement Nikesh's optimisation suggestions - https://github.com/taikoxyz/taiko-mono/issues/19982 @dantaik   → https://github.com/taikoxyz/taiko-mono/pull/19985

---

### 🔻 Low Priorities
- [x] Ensure worst-case zk-provability:
  - gas limit per block  
  - max number of blocks per proposal  
  - max proposals per L1 block/second
- [x] Use `BlobSlice` for ForcedInclusions and remove the size param @dantaik 
- [x] Try very simple claim aggregation @dantaik 
- [x] Implement `LibCodec` – wait until relevant structs and functions are finalized -- only after testing are done for the new inbox.
- [x] Remove same-slot messaging from SignalService and ERC20Vault @dantaik (we can restore to previous release with minor changes) → https://github.com/taikoxyz/taiko-mono/pull/19959
- [x] Writing L1 contract tests (using Claude Code, suite2) @ggonzalez94 
- [x] Optimize ForcedInclusionStore to 1) use a ring buffer, and 2) offer consume() function that consumes a due inclusion -- if none is due, then revert. We also need to change inbox to call consume() directly without querying if the oldest inclusion is due to avoid one SLOAD. @ggonzalez94


packages/protocol/contracts/layer2/based/docs/Derivation.md
```diff
@@ -0,0 +1,360 @@
...
+2. **Version Extraction**: Extract the version number from bytes `[offset, offset+32)` (only version `0x1` is valid for Shasta)
+3. **Size Extraction**: Extract the data size from bytes `[offset+32, offset+64)`
+4. **Decompression**: Apply ZLIB decompression to the data slice
+5. **Decoding**: RLP decode the decompressed data into a `ProposalManifest` struct
+
+#### Default Manifest Conditions
+
+A default manifest is returned when any of the following validation criteria fail:
+
+- **Blob validation**: `blobHashes.length` is zero or exceeds `PROPOSAL_MAX_BLOBS`
```
ggonzalez94: Why do we need `PROPOSAL_MAX_BLOBS`? Shouldn't we allow the proposal to use the full blob capacity if needed? and constrain it by gas limit or amount of blocks per proposal?

packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    /// @notice The maximum number of blocks allowed in a proposal. If we assume block time is as
+    /// small as one second, 384 blocks will cover an Ethereum epoch.
+    uint256 internal constant PROPOSAL_MAX_BLOCKS = 384;
+
+    /// @notice The maximum number of transactions allowed in a block in the manifest.
+    uint256 internal constant BLOCK_MAX_RAW_TRANSACTIONS = 4096 * 2;
+
+    /// @notice The maximum number of transactions allowed in the actual L2 block.
+    uint256 internal constant BLOCK_MAX_TRANSACTIONS = 4096;
```
ggonzalez94: what is the difference between these two parameters? why can we have twice as much txs in the manifest than the actual block?

packages/protocol/contracts/layer2/based/docs/Derivation.md
```diff
@@ -0,0 +1,360 @@
...
+default.blocks = new Block[](1);
+```
+
+A default manifest contains a single empty block, effectively serving as a fallback mechanism for invalid proposals.
+
+### Metadata Validation and Computation
+
+With the extracted manifest, metadata computation proceeds using both the manifest data and the parent block's metadata (`parent.metadata`). The following sections detail the validation rules for each metadata component:
+
+#### `timestamp` Validation
```
ggonzalez94: another alternative is to always ensure blocks are separated by `BLOCK_TIME` and insert empty blocks if there are none similar to what [OP does](https://specs.optimism.io/protocol/derivation.html#overview).

Pros: 
- consistent block time
- easier to enforce fair exchange issues on preconfs?

Cons
- more expensive to produce proofs?

packages/protocol/contracts/layer2/based/docs/Derivation.md
```diff
@@ -0,0 +1,360 @@
...
+
+Timestamp validation is performed collectively across all blocks and may result in block count reduction:
+
+1. **Upper bound enforcement**: If `metadata.timestamp > proposal.originTimestamp`, set `metadata.timestamp = proposal.originTimestamp`
+2. **Lower bound calculation**: `lowerBound = max(parent.metadata.timestamp + 1, proposal.originTimestamp - TIMESTAMP_MAX_OFFSET)`
+3. **Lower bound enforcement**: If `metadata.timestamp < lowerBound`, set `metadata.timestamp = lowerBound`
+4. **Block pruning**: If `metadata.timestamp > proposal.originTimestamp` after adjustments, discard this and all subsequent blocks
```
ggonzalez94: isn't this impossible to hit since step one already sets metadata.timestamp = proposal.originTimestamp?

ggonzalez94 (commented):
Started reviewing the `Derivation` doc, left a few initial comments/questions

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,669 @@
...
+
+        // Handle forced inclusion if required
+        if (IForcedInclusionStore(config.forcedInclusionStore).isOldestForcedInclusionDue()) {
+            IForcedInclusionStore.ForcedInclusion memory forcedInclusion = IForcedInclusionStore(
+                config.forcedInclusionStore
+            ).consumeOldestForcedInclusion(msg.sender);
+
+            (coreState, proposal) = _propose(config, coreState, forcedInclusion.blobSlice, true);
+            emit Proposed(proposal, coreState);
+        }
```
davidtaikocha: Not necessary, but to me it seems to put `forced inclusion` after the `regular proposal` will be more flexible for (whitelisted) pre-confirmers to decide when to propose?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,725 @@
...
+    {
+        bytes32 storedNextProposalHash = _proposalRecord(_config, _proposals[0].id + 1).proposalHash;
+
+        if (storedNextProposalHash == bytes32(0)) {
+            // Next slot is empty, only one proposal expected
+            require(_proposals.length == 1, IncorrectProposalCount());
+        } else {
+            // Next slot is occupied, need to prove it contains a smaller proposal id
+            require(_proposals.length == 2, IncorrectProposalCount());
+            require(_proposals[1].id < _proposals[0].id, NextProposalIdSmallerThanLastProposalId());
```
davidtaikocha, line 544: A little bit confused here, if `_proposals[1].id < _proposals[0].id`, how can `_checkProposalHash(_config, _proposals[1]);` promise that the `_proposals` are the last ones proposed? Shall it be:

```solidity
require(_proposals[1].id > _proposals[0].id, NextProposalIdSmallerThanLastProposalId());
_verifyLastProposal(_config, _proposals[1]);
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,725 @@
...
+            _checkProposalHash(_config, _proposals[1]);
```
davidtaikocha: Little bit confused here, how can this promise that `_proposals[0]` is the latest proposed one? (since we are using its `coreStateHash`)

packages/protocol/docs/Derivation.md
```diff
@@ -0,0 +1,436 @@
...
+| -------------------- | ------- | ---------------------------------- |
+| `number`             | uint256 | `number`                           |
+| `timestamp`          | uint256 | `timestamp`                        |
+| `difficulty`         | uint256 | `difficulty`                       |
+| `gasLimit`           | uint256 | `gasLimit`                         |
+| `id`                 | uint48  | `withdrawalsRoot` (bytes 0-5)      |
+| `numBlocks`          | uint16  | `withdrawalsRoot` (bytes 6-7)      |
+| `index`              | uint16  | `withdrawalsRoot` (bytes 8-9)      |
+| `isForcedInclusion`  | bool    | `withdrawalsRoot` (byte 10, bit 0) |
+| `isLowBondProposal`  | bool    | `withdrawalsRoot` (byte 10, bit 1) |
```
davidtaikocha: If we've decided to use `anchor` / `updateState` transactions, imo we only need to store `basefeeSharingPctg` in `block.extradata`, then the P2P syncing will work.

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -0,0 +1,343 @@
...
+        returns (bool isLowBondProposal_, address designatedProver_)
+    {
+        // Fork validation
+        require(block.number >= shastaForkHeight, L2_FORK_ERROR());
+
+        // Prevent duplicate calls within same block
+        _trackParentBlockHash(block.number - 1);
+
+        // Handle prover designation on first block
+        if (_blockIndex == 0) {
```
smtmfft, line 166: Should there be code to check whether _blockIndex is unique and in ascending order within a single proposal?

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -0,0 +1,343 @@
...
+    /// @param _anchorBlockHash L1 block hash at _anchorBlockNumber.
+    /// @param _anchorStateRoot L1 state root at _anchorBlockNumber.
+    /// @return isLowBondProposal_ True if proposer has insufficient bonds.
+    /// @return designatedProver_ Address of the designated prover.
+    function updateState(
+        // Proposal level fields - define the overall batch
+        uint48 _proposalId,
+        address _proposer,
+        bytes calldata _proverAuth,
+        bytes32 _bondInstructionsHash,
```
smtmfft, line 145: why we need expected hash here? if _bondInstructions is verifiable, since this contract is on-chain, so there is no way to forge a fake one? or it's for easy check in L1??

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -0,0 +1,343 @@
...
+        (designatedProver_, provingFeeGwei) =
+            _validateProverAuth(_proposalId, _proposer, _proverAuth);
+
+        // Check bond sufficiency
+        isLowBondProposal_ = !bondManager.hasSufficientBond(_proposer, provingFeeGwei);
+
+        // Handle low bond proposals
+        if (isLowBondProposal_) {
+            // Use previous designated prover
+            designatedProver_ = _state.designatedProver;
```
smtmfft, line 241: no need to check if _state.designatedProver has enough bound?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,810 @@
...
+    // mapping(uint256 batchId_mod_batchRingBufferSize => Batch batch) batches;
+    // mapping(uint256 batchId => mapping(bytes32 parentHash => uint24 transitionId)) transitionIds;
+    // mapping(uint256 batchId_mod_batchRingBufferSize => mapping(uint24 transitionId =>
+    //         TransitionState ts)) transitions;
+    // bytes32 __reserve1;
+    // Stats1 stats1;
+    uint256[5] private __slotsUsedByPacaya;
+
+    // Reserved slot for future migration compatibility
+    uint256 private __reservedSlot;
```
ggonzalez94: I don't think I understand the purpose of this slot. Usually you put variables for future migrations at the end, not in the middle 🤔

ggonzalez94: I think this is actually `Stats2` in Pacaya? Maybe we can express this more cleanly

ggonzalez94: proposed a cleaner version in #19947

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,810 @@
...
+        require(amount > 0, NoBondToWithdraw());
+
+        // Clear balance before transfer (checks-effects-interactions)
+        bondBalance[msg.sender] = 0;
+
+        // Transfer the bond
+        IERC20(getConfig().bondToken).safeTransfer(msg.sender, amount);
+
+        emit BondWithdrawn(msg.sender, amount);
+    }
```
ggonzalez94: what if we allow withdrawals to be executed by any address and sent to the address that owns the balance? This would allow people to withdraw from any address that has enough gas.


```suggestion
   /// @notice Withdraws bond balance to `_address`.
    /// @dev This function allows deposits that were made on Pacaya to be withdrawn.
    /// @param _address The address to withdraw the bond to
    function withdrawBond(address _address) external nonReentrant {
        uint256 amount = bondBalance[_address];
        require(amount > 0, NoBondToWithdraw());

        // Clear balance before transfer (checks-effects-interactions)
        bondBalance[_address] = 0;

        // Transfer the bond
        IERC20(getConfig().bondToken).safeTransfer(_address, amount);

        emit BondWithdrawn(_address, amount);
    }
```

packages/protocol/contracts/layer1/shasta/impl/InboxOptimized.sol
```diff
@@ -0,0 +1,293 @@
...
+
+import "./Inbox.sol";
+import "../libs/LibCodec.sol";
+
+/// @title InboxOptimized
+/// @notice Combines slot reuse and claim aggregation optimizations for the Inbox contract
+/// @dev This contract merges the optimizations from InboxWithSlotReuse and
+/// InboxWithClaimAggregation
+/// to provide both storage optimization through slot reuse and gas optimization through claim
+/// aggregation
```
ggonzalez94: These two contracts don't exist in the codebase anymore, so it doesn't make sense to reference them
```suggestion

```

packages/protocol/contracts/layer1/shasta/impl/ForcedInclusionStore.sol
```diff
@@ -0,0 +1,147 @@
...
+    function isOldestForcedInclusionDue() external view returns (bool) {
+        uint256 deadline = getOldestForcedInclusionDeadline();
+        return deadline != type(uint256).max && block.timestamp >= deadline;
+    }
+
+    /// @notice Gets the deadline for the oldest forced inclusion.
+    /// @return The deadline for the oldest forced inclusion or `type(uint256).max` if there is no
+    /// forced inclusion in the queue
+    function getOldestForcedInclusionDeadline() public view returns (uint256) {
+        if (head == tail) return type(uint256).max;
```
nikeshnazareth: Is this check redundant with the timestep check below? Since it seems cheaper, perhaps `consumeOldestForcedInclusion` should use this check instead.

packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    /// @notice The maximum anchor block number offset from the proposal origin block number.
+    uint256 internal constant ANCHOR_MAX_OFFSET = 128;
+
+    /// @notice The maximum number timestamp offset from the proposal origin timestamp.
+    uint256 internal constant TIMESTAMP_MAX_OFFSET = 12 * 32;
+
+    /// @notice The maximum block gas limit change per block in permyriad (1/10,000).
+    /// @dev E.g., 10 = 0.1 basis points = 0.001% = 10/100,000
```
nikeshnazareth: This comment seems to change the value in both steps. Should `10` imply `10 / 10,000`, which would be `0.1%`.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,810 @@
...
+        return _getCapacity(config);
+    }
+
+    /// @notice Gets the configuration for this Inbox contract
+    /// @dev This function must be overridden by subcontracts to provide their specific
+    /// configuration
+    /// @return _ The configuration struct
+    function getConfig() public view virtual returns (Config memory);
+
+    /// @notice Decodes data into CoreState, BlobReference array, and ClaimRecord array
```
nikeshnazareth:
```suggestion
    /// @notice Decodes data into deadline, CoreState, Proposal array, BlobReference, and ClaimRecord array
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,815 @@
...
+    {
+        bytes32 storedNextProposalHash = _proposalRecord(_config, _proposals[0].id + 1).proposalHash;
+
+        if (storedNextProposalHash == bytes32(0)) {
+            // Next slot is empty, only one proposal expected
+            require(_proposals.length == 1, IncorrectProposalCount());
+        } else {
+            // Next slot is occupied, need to prove it contains a smaller proposal id
+            require(_proposals.length == 2, IncorrectProposalCount());
+            require(_proposals[1].id < _proposals[0].id, NextProposalIdSmallerThanLastProposalId());
```
nikeshnazareth: I believe this error is described backwards (since it's describing the success condition)

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,815 @@
...
+            _checkProposalHash(_config, _proposals[1]);
```
nikeshnazareth: As per David's comment, I believe this check should be:

```suggestion
            _hashProposal(_proposals[1]) == storedNextProposalHash;
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -136,10 +136,13 @@ abstract contract Inbox is EssentialContract, IInbox {
         coreState = _finalize(config, coreState, claimRecords);
 
         // Verify capacity for new proposals
-        _verifyCapacity(config, coreState);
+        uint256 availableCapacity = _getAvailableCapacity(config, coreState);
+        require(availableCapacity > 0, ExceedsUnfinalizedProposalCapacity());
 
-        // Process forced inclusion if required
-        coreState = _processForcedInclusion(config, coreState);
+        if (availableCapacity > 1) {
```
nikeshnazareth: I agree the capacity check should account for the forced inclusion as well, so this is a good fix. But now it creates the possibility that the forced inclusion is postponed indefinitely if the buffer is full and proposals are created at the rate they are proven. I'd recommend insisting that the forced inclusion is the next proposal once it's due, so if there is a forced inclusion, the chain would only progress where there are two slots available.

packages/protocol/contracts/layer1/shasta/impl/ForcedInclusionStore.sol
```diff
@@ -0,0 +1,147 @@
...
+        unchecked {
+            delete queue[head++]; // delete element at head AND THEN increment head
+            _feeRecipient.sendEtherAndVerify(inclusion_.feeInGwei * 1 gwei);
+        }
+    }
+
+    /// @inheritdoc IForcedInclusionStore
+    function isOldestForcedInclusionDue() external view returns (bool) {
+        uint256 deadline = getOldestForcedInclusionDeadline();
+        return deadline != type(uint256).max && block.timestamp >= deadline;
```
nikeshnazareth: I think the advantage of using `type(uint256).max` for no deadline (instead of `0`) is that it functions like an unreachable deadline, so the first clause here is redundant

```suggestion
        return block.timestamp >= deadline;
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,856 @@
...
+        uint48 _proposalId,
+        bytes32 _parentClaimHash,
+        bytes32 _claimRecordHash
+    )
+        internal
+        virtual
+    {
+        ProposalRecord storage proposalRecord = _proposalRecord(_config, _proposalId);
+        bytes32 compositeKey = _composeClaimKey(_proposalId, _parentClaimHash);
+        proposalRecord.claimHashLookup[compositeKey].claimRecordHash = _claimRecordHash;
```
nikeshnazareth: I think there is a bug here: this just overwrites the claim record hash, whether or not there was a previous value. Consider the situation:
- a prover submits a proof within the window. The corresponding claim record is saved.
- the proof window expires before the proposal is finalized. 
- Another prover submits the proof and overwrites the claim, allowing them to receive the liveness bond. This is the claim that gets finalized.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,841 @@
...
+            // Check if there are more proposals to finalize
+            if (currentProposalId >= _coreState.nextProposalId) break;
+
+            // Try to finalize the current proposal
+            (bool finalized, uint48 nextProposalId) = _finalizeProposal(
+                _config,
+                _coreState,
+                currentProposalId,
+                i < _claimRecords.length ? _claimRecords[i] : lastFinalizedRecord,
+                i < _claimRecords.length
```
AnshuJalan: @dantaik If this condition is false, then we revert in `_finalizeProposal`: `require(_hasClaimRecord, ClaimRecordNotProvided());`

This means the proposer cannot finalize fewer than `_config.maxFinalizationCount` claims. Is this correct? We may not always have `_config.maxFinalizationCount` to verify.

nikeshnazareth: I think when we reach the end of the proposals that can be finalized, it will return (before the `require`) on:

```solidity
if (storedHash == 0) return (false, _proposalId);
```

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -0,0 +1,145 @@
...
+    /// @notice Represents a proposal for L2 blocks.
+    struct Proposal {
+        /// @notice Unique identifier for the proposal.
+        uint48 id;
+        /// @notice Address of the proposer.
+        address proposer;
+        /// @notice The L1 block timestamp when the proposal was accepted.
+        uint48 originTimestamp;
+        /// @notice The L1 block number when the proposal was accepted.
+        uint48 originBlockNumber;
```
linoscope, line 34: We should add a `originBlockHash` so derivation can derive the blockhashes of the anchor blocks (mentioned this offline but leaving this comment so we don't forget)

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,841 @@
...
+
+        if (storedNextProposalHash == bytes32(0)) {
+            // Next slot in the ring buffer is empty, only one proposal expected
+            require(_parentProposals.length == 1, IncorrectProposalCount());
+        } else {
+            // Next slot in the ring buffer is occupied, need to prove it contains a smaller
+            // proposal id
+            require(_parentProposals.length == 2, IncorrectProposalCount());
+            require(
+                _parentProposals[1].id < _parentProposals[0].id,
```
linoscope: From my understanding,  `_parentProposals[1]` is expected to be the proposal stored in the "next" proposal slot (side note: if so, the name is confusing). Below, I assume this is correct.

Now, won't we have to enforce that `_parentProposals[1]` is actually what's stored in the "next slot"? For example, what if we have `proposalHashes` with
```
[7][8][9][4][5][6]
```
where each `[]` contains the `id` of the proposal it stores.

We are _expecting_ to have `[9]` in  `_parentProposal[0]` and `[4]` in `_parentProposal[1]`. However, what prevents a malicious actor from providing `[8]` as `_parentProposal[0]` and `[7]` as `_parentProposal[1]`, hence wrongly overwriting `[9]`?

linoscope: If this is actually a problem, as a fix, I think we can do something like:
```
        } else {
            // Next slot in the ring buffer is occupied, need to prove it contains a smaller
            // proposal id
            require(_parentProposals.length == 2, IncorrectProposalCount());
            
            // New! Verify parentProposals[1] is actually what's in nextBufferSlot
            bytes32 providedHash = _hashProposal(_parentProposals[1]);
            require(providedHash == storedNextProposalHash, IncorrectNextBufferSlotProposal());
            
            require(
                _parentProposals[1].id < _parentProposals[0].id,
                NextProposalIdSmallerThanLastProposalId()
            );
        }
```

linoscope: > From my understanding, _parentProposals[1] is expected to be the proposal stored in the "next" proposal slot

This assumption may be wrong. Tbh I am having trouble wrapping my head around this function.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,841 @@
...
+    function _processForcedInclusion(
+        Config memory _config,
+        CoreState memory _coreState
+    )
+        private
+        returns (CoreState memory)
+    {
+        IForcedInclusionStore store = IForcedInclusionStore(_config.forcedInclusionStore);
+
+        if (!store.isOldestForcedInclusionDue()) {
```
linoscope: Shouldn't we enable consuming forced inclusion blocks _before_ they are due? For example, preconfers would want to consume forced inclusion blocks early, so forced inclusion doesn't interfere with their preconfed state.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,841 @@
...
+        require(_deadline == 0 || block.timestamp <= _deadline, DeadlineExceeded());
+        require(_parentProposals.length > 0, EmptyProposals());
+        require(_hashCoreState(_coreState) == _parentProposals[0].coreStateHash, InvalidState());
+    }
+
+    /// @dev Processes forced inclusion if required
+    /// @param _config The configuration parameters.
+    /// @param _coreState The core state.
+    /// @return  The updated core state or the same core state if no forced inclusion is due.
+    function _processForcedInclusion(
```
linoscope: Also, @cyberhorsey mentioned before that we need to let non-preconfers call `_processForcedInclusion` when forced inclusion blocks are due in order to be categorized as "stage 2". You may already heard this but flagging this here just in case.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
linoscope: > I'd recommend insisting that the forced inclusion is the next proposal once it's due, so if there is a forced inclusion, the chain would only progress where there are two slots available.

Good point & this makes sense to me. So if `availableCapacity == 1` AND forced inclusion is due, the proposer is forced to only consume the forced block and NOT propose his own block.

nikeshnazareth:
I have a couple of suggestions to make the `InboxOptimized1` mechanism a little cleaner.

The less intrusive suggestion is just to use the most-significant bits of the parent claim hash in the slot reuse marker, and leave it in the most-significant position, so there is less bit shifting and the overall effect is more obvious:
- create a `TOP_208_BITS_MASK`
- encoding would be `slotReuseMarker_ = (_parentClaimHash & TOP_208_BITS_MASK) | _proposalId`
- checking for a partial match would be `(_partialParentClaimHash ^ _parentClaimHash) & TOP_208_BITS_MASK == 0`

I think the cleaner suggestion though is to observe:
- the slot reuse marker is only used for the default slot. Therefore we only need one per proposal `bufferSlot` (not one per record in the mapping)
- we could restore the `ProposalRecord` mechanism with a struct containing
  - `proposalHash`
  - `claimRecordHash`
  - `proposalId` (48 bits)
  - `partialParentClaimHash` (208 bits)
- any additional `claimRecordHash` values would still be in their own `_additionalClaimRecordHash` mapping
- in this way, the `proposalId` and `partialParentClaimHash` still function like a `slotReuseMarker` but you can reference them directly. Moreover, we don't need the `ExtendedClaimRecord` or the additional safety checks to make sure that the `_parentClaimHash` is not `_DEFAULT_SLOT_HASH`

packages/protocol/contracts/layer1/shasta/impl/InboxOptimized1.sol
```diff
@@ -0,0 +1,243 @@
...
+            bondInstructions: currentInstructions
+        });
+
+        uint256 finalRecordCount = 1;
+        uint256 currentRecordIndex;
+        uint48 currentGroupStartId = _proposals[0].id;
+
+        // Process remaining proposals
+        for (uint256 i = 1; i < _proposals.length; ++i) {
+            require(_claims[i].parentClaimHash != _DEFAULT_SLOT_HASH, InvalidParentClaimHash());
```
nikeshnazareth: I believe we need this check for `_claims[0]` as well

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
nikeshnazareth: I think this is actually worse in the `InboxOptimized1` case because a record might have a large span, and we can overwrite it with a different record with a smaller span, effectively undoing all the proving work. This can also occur within the proving window.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
ggonzalez94: this was following the previous design we had when forced inclusions came before the regular proposal. I don't have a strong opinion for this and I see a case for both approaches(and ultimately think it is not a big deal for 99% of the cases).
But the case for having it before is that you can prevent the preconfer from changing the state and making the forced inclusion fail if those txs are contingent on state.

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
ggonzalez94: I had the same question, but after moving bonds to L2 we are now doing bonds on a "best effort" basis, which means there might be edge cases where the assigned prover didn't have a large enough bond.

packages/protocol/docs/Derivation.md
```diff
@@ -0,0 +1,438 @@
...
+```
+
+### Proposer and `isLowBondProposal` Validation
+
+To maintain the integrity of the proposal process, the `proposer` address must pass specific validation checks, which include:
+
+- Confirming that the proposer holds a sufficient balance in the L2 BondManager contract.
+- Ensuring the proposer is not waiting for exiting.
+
+Should any of these validation checks fail (as determined by the `updateState` function returning `isLowBondProposal = true`), the proposal's block array is replaced with a single block containing no transactions. Note that this does not use the default manifest; instead, it modifies only the blocks array while preserving other manifest data.
```
ggonzalez94: Why do we make this distinction for low bond proposals and not use the default manifest? should we allow the proposer who's at fault to still set some of these values?

packages/protocol/docs/Derivation.md
```diff
@@ -0,0 +1,438 @@
...
+- **Future reference**: `manifest.blocks[i].anchorBlockNumber >= proposal.originBlockNumber`
+- **Excessive lag**: `manifest.blocks[i].anchorBlockNumber < proposal.originBlockNumber - ANCHOR_MAX_OFFSET`
+
+**Forced inclusion protection**: For non-forced proposals (`proposal.isForcedInclusion == false`), if no blocks have valid anchor numbers greater than its parent's, the entire manifest is replaced with the default manifest, penalizing proposals that fail to provide proper L1 anchoring.
+
+#### `anchorBlockHash` and `anchorStateRoot` Validation
+
+The anchor hash and state root must maintain consistency with the anchor block number (enforced by the Taiko node/driver):
+
+- If `anchorBlockNumber == parent.metadata.anchorBlockNumber`: Both `anchorBlockHash` and `anchorStateRoot` must be zero
```
ggonzalez94, line 215: dumb question: wouldn't it be more consistent to set it to the parent's anchorBlockHash and anchorStateRoot instead of setting it to zero?

packages/protocol/contracts/layer2/based/docs/Derivation.md
dantaik: The goal is to cap all parameters, ensuring that even the worst-case prover attack remains within defined limits.

packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
dantaik: `BLOCK_MAX_RAW_TRANSACTIONS` – Maximum number of transactions allowed in proposal's blob data. This cap ensures the cost of a worst-case prover attack is bounded.
`BLOCK_MAX_TRANSACTIONS` – Maximum number of transactions that can be included in an actual L2 block.

packages/protocol/contracts/layer2/based/docs/Derivation.md
dantaik: I don’t think enforcing chain liveness should be part of the core protocol. It makes more sense to enforce slashing conditions at the preconfirmation level, which is also more flexible.

packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
dantaik: Improve the comments in https://github.com/taikoxyz/taiko-mono/pull/19973

packages/protocol/contracts/layer2/based/docs/Derivation.md
dantaik: item 4 removed in https://github.com/taikoxyz/taiko-mono/pull/19973

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: @cyberhorsey @YoGhurt111 what do you think?

The current behavior matches the Pacaya fork, forced-inclusion first, followed by a normal proposal.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: `_proposals[1]` will serve as a proof to show either the next slot is 0, or the next slot is used by proposal with id = `nextProposalId - ringBufferSize`

packages/protocol/docs/Derivation.md
dantaik: Is it beneficial to store more information in block header?

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
dantaik: Yes, this is part of what the node will have to verify.

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
dantaik: The expected bond instruction hash does not need to be passed into the anchor contract. However, the node/driver must fetch it from L1 and compare it with the value calculated in the anchor. Regardless of the approach, the node/driver must verify the expected bond instruction hash.

packages/protocol/contracts/layer1/shasta/impl/ForcedInclusionStore.sol
dantaik: @ggonzalez94 I also think the implementation can be improved.  In the Inbox, I propose to call the consume function (or a variation of it) directly without a pre-check, if there is no due inclusion, simply revert and the Inbox shall catch this revert to avoid proposing.

packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
dantaik: Changed to the following in https://github.com/taikoxyz/taiko-mono/pull/19973

```
/// @notice The maximum block gas limit change per block, in millionths (1/1,000,000).
/// @dev For example, 10 = 10 / 1,000,000 = 0.001%

File: packages/protocol/contracts/layer1/shasta/impl/Inbox.sol (hunk 19 above)
Comment (dantaik) [reply packages/protocol/contracts/layer1/shasta/impl/Inbox.sol]: fixed in https://github.com/taikoxyz/taiko-mono/pull/19973

File: packages/protocol/contracts/layer1/shasta/impl/Inbox.sol (hunk 18 above)
Comment (dantaik) [reply packages/protocol/contracts/layer1/shasta/impl/Inbox.sol]: _proposals[1].id < _proposals[0].id is indeed correct:

- _proposals[0]: the latest proposal
- _proposals[1]: its ID should be  `_proposals[0].id - ringBufferSize+1`

File: packages/protocol/contracts/layer1/shasta/impl/Inbox.sol (hunk 20 above)
Comment (dantaik) [reply packages/protocol/contracts/layer1/shasta/impl/Inbox.sol]: Agreed. Made the change in https://github.com/taikoxyz/taiko-mono/pull/19973

File: packages/protocol/contracts/layer1/shasta/impl/ForcedInclusionStore.sol (hunk 15 above)
Comment (ggonzalez94) [reply packages/protocol/contracts/layer1/shasta/impl/ForcedInclusionStore.sol]: yep, this is leftover from previous implementation. I'll work on an improved version of this

File: packages/protocol/contracts/layer1/shasta/impl/Inbox.sol (hunk 18 above)
Comment (ggonzalez94) [reply packages/protocol/contracts/layer1/shasta/impl/Inbox.sol]: yes, but I think the naming we've chosen has been confusing to most people(I already updated it to `parentProposals` to at least reflect that they are not the current L2 blocks being proposed). But I'll probably work on a simpler struct that encompasses all the proposal parameters in a clearer manner

File: packages/protocol/contracts/layer1/shasta/impl/Inbox.sol (hunk 22 above)
Comment (dantaik) [reply packages/protocol/contracts/layer1/shasta/impl/Inbox.sol]: I’m assuming that if the prover is bug-free, it should be impossible to prove two claims for the same proposal that both result in records with the same _parentClaimHash.

Or @nikeshnazareth, did I miss something in your comments? Please share any additional thoughts or details so I can make sure we resolve this correctly.

File: packages/protocol/docs/Derivation.md (hunk 8 above)
Comment (davidtaikocha) [reply packages/protocol/docs/Derivation.md]: Don't think it's necessary, since p2p syncing doesn't need this, and if a client wants to know these, they can just decode them from the anchor's calldata.

File: packages/protocol/contracts/layer1/shasta/impl/Inbox.sol [hunk 31]
```diff
@@ -0,0 +1,899 @@
...
+        Config memory _config,
+        CoreState memory _coreState,
+        LibBlobs.BlobSlice memory _blobSlice,
+        bool _isForcedInclusion
+    )
+        private
+        returns (CoreState memory)
+    {
+        unchecked {
+            uint256 parentBlockNumber = block.number - 1;
```
ggonzalez94, line 718: nit: I imagine the reason why we use `block.number - 1` for the origin number of this proposal instead of the current block is to be able to get it's hash(which is not possible for the current block).
Maybe we can add a small comment explaining this
```suggestion
            // use previous block as the origin for the proposal to be able to call `blockhash`
            uint256 parentBlockNumber = block.number - 1;
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: @linoscope your example above is good, but in the code we check ` _parentProposals[1].id < _parentProposals[0].id,`, therefore, if you provide [7][8] as the inputs, then 8<7 will fail.

Or maybe this was a bug but got fixed recently. Please see https://github.com/taikoxyz/taiko-mono/pull/19973/files

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: I'll leave forced inclusion related logics to @ggonzalez94

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: Need to figure out who this works with preconf

dantaik:
> I have a couple of suggestions to make the `InboxOptimized1` mechanism a little cleaner.
> 
> The less intrusive suggestion is just to use the most-significant bits of the parent claim hash in the slot reuse marker, and leave it in the most-significant position, so there is less bit shifting and the overall effect is more obvious:
> 
> * create a `TOP_208_BITS_MASK`
> * encoding would be `slotReuseMarker_ = (_parentClaimHash & TOP_208_BITS_MASK) | _proposalId`
> * checking for a partial match would be `(_partialParentClaimHash ^ _parentClaimHash) & TOP_208_BITS_MASK == 0`
> 
> I think the cleaner suggestion though is to observe:
> 
> * the slot reuse marker is only used for the default slot. Therefore we only need one per proposal `bufferSlot` (not one per record in the mapping)
> * we could restore the `ProposalRecord` mechanism with a struct containing
>   
>   * `proposalHash`
>   * `claimRecordHash`
>   * `proposalId` (48 bits)
>   * `partialParentClaimHash` (208 bits)
> * any additional `claimRecordHash` values would still be in their own `_additionalClaimRecordHash` mapping
> * in this way, the `proposalId` and `partialParentClaimHash` still function like a `slotReuseMarker` but you can reference them directly. Moreover, we don't need the `ExtendedClaimRecord` or the additional safety checks to make sure that the `_parentClaimHash` is not `_DEFAULT_SLOT_HASH`

Created this issue to follow up later https://github.com/taikoxyz/taiko-mono/issues/19982

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,899 @@
...
+        uint256 slotReuseMarker;
+    }
+
+    // ---------------------------------------------------------------
+    // State Variables
+    // ---------------------------------------------------------------
+
+    /// @dev Ring buffer for storing proposal hashes.
+    /// This variable reuse the `batches slot in pacaya fork.
+    mapping(uint256 bufferSlot => bytes32 proposalHash) internal _proposalHashes;
```
ggonzalez94: I think reusing this slot will cause a "deadlock" scenario in the `verifyChainHead` function. I imagine the reason we want to reuse this slot is so that we don't need to "warm up" the buffer and it is already full to start with.

But: When reading from the reused storage slot, it will read the first 32 bytes of whatever Batch struct data was stored there from Pacaya. Since Batch structs contain actual data (metaHash, batchId, etc.), these bytes are almost certainly non-zero.
The Consequence: The contract will incorrectly think the ring buffer has already wrapped around, requiring 2 proposals to be provided when only 1 should be needed. Proposers won't be able to provide a correct `Proposal` since the slot is filled with "garbage" from the point of view of this contract

packages/protocol/docs/Derivation.md
dantaik: Lets see which approach makes it easy to implement the prover. Either way should be OK. @smtmfft

packages/protocol/docs/Derivation.md
dantaik: Good questions, I need to refresh my memory about block derivation :D

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
linoscope: > Or maybe this was a bug but got fixed recently.

Oh, indeed, it was fixed by this change:
image

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
nikeshnazareth: I was referring to the error `NextProposalIdSmallerThanLastProposalId`. That's the condition we want.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
nikeshnazareth: As I understand it, the prover should be stateless, so it just validates that the transition is correct and ignores which other proofs have been submitted before.

In the standard `Inbox` case, the prover will accept a valid claim that has no bond instruction and it will also accept a valid claim (the same transition) that has a bond instruction sending funds from Alice to Bob. So if Alice proves on time, and then Bob submits the same proof after the proof window (but before finalization), the claim will now say that Alice owes Bob the liveness bond.


In the `InboxOptimized1` case, Alice can send a chain of proofs that start from some `_parentClaimHash` and have a span of 5. Bob can then send a proof with the same `_parentClaimHash` and a span of 1. This will overwrite the work that Alice has put in to prove the transitions 2-5.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
YoGhurt111: I also think putting forced-inclusion first would be better, so the hacker can't attack the txs in forced-inclusion.

packages/protocol/docs/Derivation.md
dantaik: Lets keep the following:

```
| Metadata Component   | Type    | Header Field                              |
| -------------------- | ------- | ----------------------------------------- |
| `number`             | uint256 | `number`                                  |
| `timestamp`          | uint256 | `timestamp`                               |
| `difficulty`         | uint256 | `difficulty`                              |
| `gasLimit`           | uint256 | `gasLimit`                                |
| `basefeeSharingPctg` | uint8   | First byte in `extraData`                 |
| `isLowBondProposal`  | bool    | Lowest bit in the 2nd byte in `extraData` |

```

I'm not quite sure about `isLowBondProposal` but I want to leave it there just in case we do need to use this to slash the preconfers. See https://github.com/taikoxyz/taiko-mono/pull/19995

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
dantaik: Since the inbox uses a ring buffer, there is an upper bound on how many proposals can remain unverified. The design assumes that for any preconfirmer, the minimum bond balance on L2 must at least cover the case where all proposals are made by a single proposer or assigned to a single prover.

This means the buffer size cannot simply be “the bigger the better.” A larger buffer increases the required minimum bond balance for proposers/provers. Currently, the buffer size for MainnetShastaInbox is set to 2400.

Please see https://github.com/taikoxyz/taiko-mono/pull/19995/files for how 2400 is calculated, but I'm more comfortable to use 3600:

```solidity
    /// @dev Ring buffer size for storing proposal hashes.
    /// Assumptions:
    /// - D = 2: Proposals may continue without finalization for up to 2 days.
    /// - P = 4: On average, 1 proposal is submitted every 4 Ethereum slots (≈48s).
    ///
    /// Calculation:
    ///   _RING_BUFFER_SIZE = (86400 * D) / 12 / P
    ///                     = (86400 * 2) / 12 / 4
    ///                     = 3600
    uint64 private constant _RING_BUFFER_SIZE = 3600;
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: renamed `NextProposalIdSmallerThanLastProposalId` to `InvalidLastProposalProof `: see https://github.com/taikoxyz/taiko-mono/pull/19995/files

packages/protocol/contracts/layer1/shasta/impl/ForcedInclusionStore.sol
dantaik: @ggonzalez94

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: I think this is indeed a bug. I'll try to fix it :D Really appreciate this feedback @nikeshnazareth

dantaik: @nikeshnazareth do you think this [PR](https://github.com/taikoxyz/taiko-mono/pull/19996/files) will fix the issue above for both `Inbox` and `InboxOptimized1`?

packages/protocol/docs/Derivation.md
dantaik: I I think it make sense to also enforce using the default manifest if `isLowBondProposal` is true. Made changes here: https://github.com/taikoxyz/taiko-mono/pull/19995/files

packages/protocol/docs/Derivation.md
dantaik: Using the parent’s `anchorBlockHash` and `anchorStateRoot` requires re-verification via Merkle proof. Setting them to 0 indicates no proof is needed, since they were already verified in the parent.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: I think we can now ignore this issue as in the new code we are no longer aiming to use existing slots.

packages/protocol/docs/Derivation.md
```diff
@@ -0,0 +1,432 @@
...
+
+### Transaction Execution
+
+TODO:
+
+- `BLOCK_MAX_TRANSACTIONS` is applied
+
+## Base Fee Calculation
+
+The calculation of block base fee shall follow [EIP-4396](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4396.md#specification).
```
davidtaikocha, line 428: Since EIP-4396 uses `parent_block_time = self.parent(block).timestamp - self.parent(self.parent(block)).timestamp`, so maybe we still need to define what's the initial base fee (`0.025gwei`?) for the first two Shasta blocks here. And what's the `BLOCK_TIME_TARGET` (`2s`?).

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -0,0 +1,202 @@
...
+        /// @notice Whether the proposal is from a forced inclusion.
+        bool isForcedInclusion;
+        /// @notice The percentage of base fee paid to coinbase.
+        uint8 basefeeSharingPctg;
+        /// @notice Blobs that contains the proposal's manifest data.
+        LibBlobs.BlobSlice blobSlice;
+    }
+
+    /// @notice Represents a proposal for L2 blocks.
+    struct Proposal {
```
YoGhurt111, line 46: Should we keep the original `SuggestedFeeRecipient` here?
To make the L2 coinbase flexiable?
In L1, proposer(msg.sender) can be contract and We don't enforce that the L2 coinbase must be the same contract address if we have the field.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -0,0 +1,887 @@
...
+                originBlockHash: blockhash(parentBlockNumber),
+                isForcedInclusion: _isForcedInclusion,
+                basefeeSharingPctg: _config.basefeeSharingPctg,
+                blobSlice: _blobSlice
+            });
+
+            Proposal memory proposal = Proposal({
+                id: _coreState.nextProposalId++,
+                proposer: msg.sender,
+                timestamp: uint48(block.timestamp),
```
YoGhurt111, line 730: 2 questions here:
1. How to differentiate timestamps between L2 blocks within a `Proposal`？
2. If one propose includes both `ForceInclusion` and normal `Proposal`, do they both have the timestamp set to `timestamp: uint48(block.timestamp)`?

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
YoGhurt111: Should we also add `lastBlockID` into struct `Proposal`?
Without this field, it's really hard to work with preconf in L2.
For example, my L2 head is `5`(with preconf), when a latest Proposal has been seen, I can know the `lastBlockID` is `2` in parent Proposal, so I will check the blocks from `3`.

packages/protocol/contracts/layer1/shasta/impl/MainnetShastaInbox.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    //     coreStateHash = keccak256(abi.encode(_coreState));
+    //     emit CoreStateSet(_coreState);
+    // }
+
+    /// @notice Gets the configuration for this Inbox contract
+    /// @return _ The configuration struct with shasta-specific settings
+    // TODO: figure out these values
+    function getConfig() public pure override returns (Config memory) {
+        return Config({
```
YoGhurt111, line 49: Was `blockMaxGasLimit` originally an on-chain config, and now it's being changed to off-chain consensus?

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -0,0 +1,343 @@
...
+    /// @param _anchorStateRoot L1 state root at _anchorBlockNumber.
+    /// @return isLowBondProposal_ True if proposer has insufficient bonds.
+    /// @return designatedProver_ Address of the designated prover.
+    function updateState(
+        // Proposal level fields - define the overall batch
+        uint48 _proposalId,
+        address _proposer,
+        bytes calldata _proverAuth,
+        bytes32 _bondInstructionsHash,
+        LibBonds.BondInstruction[] calldata _bondInstructions,
```
linoscope, line 146: How are the values of `_bondInstructions` derived? Is it from the `emit BondInstructed` event of the proposal being processed?

linoscope: One thing to consider compatibility with preconfs: If anchor execution depends on finalizations of earlier blocks at proposal time, you can’t know at preconf time which bindInstructions must execute. That uncertainty makes it hard to preconfirm the exact L2 state.

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -0,0 +1,343 @@
...
+    /// @param _proverAuth Encoded ProverAuth for prover designation (empty after block 0).
+    /// @param _bondInstructionsHash Expected cumulative hash after processing instructions.
+    /// @param _bondInstructions Bond credit instructions to process for this block.
+    /// @param _blockIndex Current block index within the proposal (0-based).
+    /// @param _anchorBlockNumber L1 block number to anchor (0 to skip anchoring).
+    /// @param _anchorBlockHash L1 block hash at _anchorBlockNumber.
+    /// @param _anchorStateRoot L1 state root at _anchorBlockNumber.
+    /// @return isLowBondProposal_ True if proposer has insufficient bonds.
+    /// @return designatedProver_ Address of the designated prover.
+    function updateState(
```
linoscope, line 140: One thing that may simplify preconfs is adding a rawTxListHash to the anchor tx that commits to the unfiltered “raw tx list” used at preconf/proposal. This lets slashing focus on divergences between the committed list and what was proposed, avoiding false positives when the L2 state root shifts due to parent reorgs the preconfer didn’t cause.

Will need to redesign and reconsider preconf slashing with @AnshuJalan to confirm.

packages/protocol/contracts/layer2/based/BondManager.sol
```diff
@@ -0,0 +1,223 @@
...
+    // State Variables
+    // ---------------------------------------------------------------
+
+    /// @notice The address of the inbox contract that is allowed to call debitBond and creditBond
+    address public immutable authorized;
+
+    /// @notice ERC20 token used as bond.
+    IERC20 public immutable bondToken;
+
+    /// @notice Minimum bond required on L1
```
linoscope: I think it can be in L2 too?

```suggestion
    /// @notice Minimum bond required
```

packages/protocol/docs/Derivation.md
ggonzalez94: created an issue to track this #20044

ggonzalez94:
@claude please review and tell me if there's any critical issue to fix or pay attention to before merging

ggonzalez94:
On claude's review:

1. Not an issue since Bond instructions are included in the TransitionRecord hash (so you cannot submit two transitions with different instructions that hash to the same value)

2. I think some of this was already addressed and there's an open issue about this(I agree that simplifying the transition aggregation logic is likely a good idea)

3. done

And the rest are either done, or missunderstandings from claude

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
ggonzalez94: @linoscope you are opening an issue for this right?

bearni95 (approved):

ggonzalez94 (approved):
LGTM

RogerLamTd (approved):

ggonzalez94 (approved):

YoGhurt111 (approved):

PR #19883, opened 2025-08-06
feat(protocol): enable claim record aggregation

dantaik:
Notice I also change the code to proposed the forced inclusion proposal before the regular proposal. I don't think it should matter.


dantaik:
@claude please review

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -150,11 +185,14 @@ contract Inbox is IInbox {
     // Private Functions
     // -------------------------------------------------------------------------
 
-    /// @dev Proposes a new proposal of L2 blocks.
-    /// @param _coreState The core state of the inbox.
-    /// @param _frame The frame of the proposal.
-    /// @return coreState_ The updated core state.
-    /// @return proposal_ The created proposal.
+    /// @dev Proposes a new L2 block and updates the inbox state
+    /// @notice Creates a proposal with associated bonds and stores its hash on-chain.
```
ggonzalez94, line 211: nit: we souldn't use `notice` for internal functions and the notice shouldn't come after the `dev`

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -182,39 +220,47 @@ contract Inbox is IInbox {
...
-    /// @param _claim The claim containing the proof details.
-    function _prove(Proposal memory _proposal, Claim memory _claim) private {
+    /// @dev Proves a single proposal by validating the claim and creating the claim record
+    /// @notice This function validates the proposal hash and determines the bond decision
+    /// based on proof timing. The resulting claim record may be aggregated with others
+    /// for gas efficiency
+    /// @param _proposal The proposal to prove
+    /// @param _claim The claim containing the state transition details
+    /// @return claimRecord_ The claim record that may be aggregated with others
+    function _prove(
```
ggonzalez94: nit: `_prove` is actually confusing since you expect this function to call the verifier. Maybe another name?
```suggestion
    function _buildClaimRecord(
```

ggonzalez94 (commented):
Still going trough this, but the gains are clearly significant. I wish there was a simpler way to achieve this as the code is getting slightly more complex than I'd like, but i haven't been able to think of anything concrete yet

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: Done

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: Done

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -209,39 +245,47 @@ contract Inbox is IInbox {
...
             provabilityBond: _proposal.provabilityBond,
-            proofTiming: proofTiming
+            bondDecision: bondDecision,
+            nextProposalId: _proposal.id + 1
         });
 
-        bytes32 claimRecordHash = keccak256(abi.encode(claimRecord));
-        inboxStateManager.setClaimRecordHash(_proposal.id, _claim.parentClaimHash, claimRecordHash);
-        emit Proved(_proposal, claimRecord);
+        emit Proved(_proposal, claimRecord_);
```
davidtaikocha, line 279: Shall we emit the aggregated records instead of emitting the "raw records" here? Seems then this will be more convenient for proposers to build the records params for finalization.

dantaik: I'm not sure if emitting the aggregated record is the right approach, I consider claim record aggregation is an internal optimisation and shall not be visible to block builders and provers.

dantaik: See https://github.com/taikoxyz/taiko-mono/pull/19895/files

PR #19884, opened 2025-08-06
feat(protocol): fix BlobLocator issues while keeping an optimized most common use case

ggonzalez94:
(AI generated description)

# The Problem

The current BlobLocator to Frame conversion has some limitations/issues:

- **Lost information**: The size field from BlobLocator is completely discarded during conversion, making it impossible for L2 to know exactly how much data to read
- **Inflexible offset handling**: Only a single offset is stored in Frame despite potentially having multiple blobs, forcing an assumption that only the first blob can have an offset while subsequent blobs must be fully utilized
- **No partial blob support**: Cannot specify that only a portion of the last blob should be used, limiting efficient data packing across blobs

# The Solution

Introduced a hybrid dual-mode design that optimizes for the common case while providing full flexibility for blob sharing:

## Full Blobs Mode (default, optimized)

- Uses only 10 bytes of calldata (blobStartIndex + numBlobs)
- Assumes consecutive, fully-utilized blobs
- Frame contains only blobHashes[] with empty offsets[] and sizes[] arrays

## Blob Sharing Mode (flexible)

- Allows custom offset and size per blob via BlobDescriptor[]
- Each blob can be partially utilized with specific byte ranges
- Required for forced inclusions that need precise offset/size specifications

The design automatically detects the mode based on whether the blobs array is populated, providing maximum efficiency for the common case while maintaining the flexibility needed for advanced use cases.


ggonzalez94:
closing this since `size` will be part of the blob itself and we will assume contiguous frames(i.e. only the first blob can have an offset)

PR #19891, opened 2025-08-07
feat(protocol): multiple changes applied

dantaik:
Changes applied:

 • Added LibBlobs.sol library for centralized blob handling with BlobLocator and BlobFrame structs
  • Created BlockHashProvider.sol and IBlockHashManager.sol for L2 block hash management
  • Extracted bond operations to LibBondOperation.sol with aggregateBondOperation() helper function
  • Renamed bond fields to include "Gwei" suffix for clarity (e.g., bondWei → bondGwei)
  • Simplified forced inclusion by removing separate frame parameter and using LibBlobs.BlobFrame
  • Changed Proposed event to emit single proposal instead of arrays
  • Added anchorBlockNumber and anchorBlockHash to Anchor's State struct
  • Moved blob validation logic to centralized LibBlobs.validateBlobLocator() function
  • Renamed Frame to BlobFrame for better naming consistency
  • Removed multi-proposal handling logic in favor of single proposal processing


dantaik:
@claude  please review

PR #19893, opened 2025-08-07
feat(protocol): implement new Anchor contract

dantaik:
Will work on a Notion based design doc for easier collaboration - https://www.notion.so/taikoxyz/Deriving-blocks-in-Shasta-2489673143d680c79dadcac0256292f5


dantaik:
@claude please review again

packages/protocol/contracts/layer2/shasta/impl/Anchor.sol
```diff
@@ -3,96 +3,224 @@ pragma solidity ^0.8.24;
...
-                bondManager.creditBond(_bondOperations[i].receiver, _bondOperations[i].credit);
+        // Process bond operations and update hash
+        _newState.bondOperationsHash = _processBondOperations(_newState, _bondOperations);
+
+        // Apply gas issuance rate adjustment with bounds
+        _newState.gasIssuancePerSecond = _adjustGasIssuanceRate(_newState.gasIssuancePerSecond);
+
+        // Atomically update state
+        _state = _newState;
+        emit StateUpdated(_state);
```
davidtaikocha: Since system-call is not a real onchain transaction, its receipt is also not queryable through RPC, so seems its meaningless to emit any event in a system-call method?

packages/protocol/contracts/layer2/shasta/impl/Anchor.sol
```diff
@@ -3,96 +3,223 @@ pragma solidity ^0.8.24;
...
+    /// @dev Validates anchor block data and saves to synced block manager
+    function _validateAndSaveAnchorBlock(State memory _newState) private {
+        // Validate block data integrity
+        if (_newState.anchorBlockHash == 0) revert InvalidAnchorBlockHash();
+        if (_newState.anchorStateRoot == 0) revert InvalidAnchorStateRoot();
+
+        // Ensure monotonic block progression
+        if (_newState.anchorBlockNumber <= _state.anchorBlockNumber) {
+            revert InvalidAnchorBlockNumber();
+        }
```
davidtaikocha: IMO these checks should all be done by `driver` && `prover`, and system call should only do simple slot-writing (and will never revert if the caller is the `systemTransactor`), not sure if its ok for the contracts to double-check them, but seems the node also doesn't know how to deal with the revert of a system-call?

dantaik: Will do later.

packages/protocol/contracts/layer2/shasta/impl/Anchor.sol
```diff
@@ -3,96 +3,224 @@ pragma solidity ^0.8.24;
...
-    /// @dev Only the anchor transactor address can call this function
-    ///      This ensures state updates come from the L2 system itself
+    /// @notice Atomically updates the anchor state with L1 synchronization data
+    /// @param _newState New state containing L1 block info and configuration
+    /// @param _bondOperations Bond credit operations to execute
+    /// @dev Security: Only callable by the keyless anchor transactor address
+    /// TODO(daniel): remove all validations as node cannot afford reverting.
     function setState(
         State memory _newState,
         LibBondOperation.BondOperation[] memory _bondOperations
```
davidtaikocha: If we still want to keep the ethereum `block` structure as it is (which i think is important since otherwise it will break the hash calculation for all the ethereum tools), seems it's not easy to put `LibBondOperation.BondOperation[]` as a parameter of a system call (maybe still possible if we limit the length),  all parameters that used for a system call method should either be derived from the current chain state or in `block.extradata` (or some unused fields in block for L2), if not, the P2P sync can't work, they can't know how to recover the block without reading the L1 again.

packages/protocol/contracts/layer2/shasta/impl/Anchor.sol
dantaik: Done

packages/protocol/contracts/layer2/shasta/impl/Anchor.sol
dantaik: Damn, this is an issue then.....

dantaik: this means L2 bond processing may not be possible...

davidtaikocha: Rn i think maybe we can use `block.withdrawals`... let me also think about it...
```go
type Withdrawal struct {
	Index     uint64         `json:"index"`          // monotonically increasing identifier issued by consensus layer
	Validator uint64         `json:"validatorIndex"` // index of validator associated with withdrawal
	Address   common.Address `json:"address"`        // target address for withdrawn ether
	Amount    uint64         `json:"amount"`         // value of withdrawal in Gwei
}
```

davidtaikocha: > Rn i think maybe we can use block.withdrawals... let me also think about it...

After double-checking, I think we can reuse it.

PR #19905, opened 2025-08-08
feat(protocol): implement BondManager with time based withdrawal window

ggonzalez94:
Alternate approach to #19904

**Pros**:
- No extra `SSTORE` in the propose function
- No circular dependency between the BondManager and the Inbox

**Cons**:
-  In theory proposal can remain unfinalized indefinitely, but in practice after the `extendedProvingWindow` the incentives are very strong for a prover to come in. A safe value for this is `extendedProvingWindow` + buffer, where buffer is large enough even for very difficult to prove batches


packages/protocol/contracts/layer1/shasta/impl/BondManagerL1.sol
```diff
@@ -0,0 +1,101 @@
...
+    function withdraw(address to, uint96 amount) external override(BondManager, IBondManager) {
+        Bond storage bond_ = bond[msg.sender];
+
+        if (
+            bond_.withdrawalRequestedAt == 0
+                || block.timestamp < bond_.withdrawalRequestedAt + withdrawalDelay
+        ) {
+            // Active proposer or withdrawal delay not passed yet, can only withdraw excess above
+            // minBond
+            require(bond_.balance - amount >= minBond, MustMaintainMinBond());
```
dantaik:
```suggestion
            require(bond_.balance >= minBond + amount, MustMaintainMinBond());
```

dantaik:
@ggonzalez94 is it possible to merge BondManagerL1, BondManagerL2 into a single BondManager, such that if it is deployed to L1, we config minBond and withdrawalWindow to a non zero value, if on L2, we confirg them to 0s.

dantaik:
I'm done reviewing this PR, left some comments.

dantaik:
@ggonzalez94 do you think across Inbox, Anchor, and BondManager, we should always use gwei as the unit?

PR #19913, opened 2025-08-11
chore(taiko-client): lower request sync margin

cyberhorsey:
nobody should be requesting such old blocks, and they are often far behind the tip. I screwed up the tip - we should check *the onchain tip*, not the latest seen proposal tip. The network is being spammed with these requests.

### DIFF
--- a/.github/workflows/taiko-client--docker.yml
+++ b/.github/workflows/taiko-client--docker.yml
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


cyberhorsey:
@claude i think you might be confused, doesnt lowering the requestMargin *reduce* the amount of requests that can be sent? you said it *increases* them.

RogerLamTd (approved):

RogerLamTd (approved):

YoGhurt111 (approved):

PR #19914, opened 2025-08-11
Update GitHub Actions 

Daulox92:
Updates actions/checkout@v4 to actions/checkout@v5 
Reference: https://github.com/actions/checkout/releases/tag/v5.0.0


RogerLamTd:
We have self hosted runners, will update as necessary.

PR #19915, opened 2025-08-12
feat(protocol): change min base fee to 0.025 gwei

dantaik:
@YoGhurt111 @cyberhorsey @davidtaikocha do we need a hark fork for this?

### DIFF
--- a/packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
+++ b/packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
@@ -23,7 +23,7 @@ abstract contract PacayaAnchor is OntakeAnchor {
 
     /// @notice Golden touch address is the only address that can do the anchor transaction.
     address public constant GOLDEN_TOUCH_ADDRESS = 0x0000777735367b36bC9B61C50022d9D0700dB4Ec;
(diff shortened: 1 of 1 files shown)


davidtaikocha (approved):
Don't think we need a hardfork, should be ok to proxy upgrade the `TaikoAnchor`.

packages/protocol/contracts/layer2/based/anchor/PacayaAnchor.sol
```diff
@@ -23,7 +23,7 @@ abstract contract PacayaAnchor is OntakeAnchor {
 
     /// @notice Golden touch address is the only address that can do the anchor transaction.
     address public constant GOLDEN_TOUCH_ADDRESS = 0x0000777735367b36bC9B61C50022d9D0700dB4Ec;
-    uint256 public constant BASEFEE_MIN_VALUE = 100_000_000; // 0.1 gwei
+    uint256 public constant BASEFEE_MIN_VALUE = 25_000_000; //  0.25 gwei
```
davidtaikocha:
```suggestion
    uint256 public constant BASEFEE_MIN_VALUE = 25_000_000; //  0.025 gwei
```

YoGhurt111:
No need, just need to coordinate this change with the wallet service provider and update the mapping value on node side

ggonzalez94 (approved):

PR #19917, opened 2025-08-12
feat(protocol): move bond process entirely to L2

dantaik:
I like the idea of moving all bond-related operations to L2 (from @ggonzalez94 ). However, I haven’t yet figured out how to incentivize provers to prove proposals submitted by an address with little or no bond balance on L2.

- Moves all bond operations from L1 to L2 - Bonds are now handled exclusively on Layer 2 instead of Layer 1
- Removes bond-related code from L1 contracts - Cleans up L1 by removing bond functionality
- Introduces virtual claim records aggregation - Adds _aggregateClaimRecords as a virtual function
- Open design question - Still determining how to incentivize provers when proposers have low/no bond balance on L2
- Simplified Anchor's setState function


packages/protocol/contracts/layer2/based/docs/Derivation.md
```diff
@@ -130,6 +130,18 @@ struct ProposalManifest {
...
+### Proposer Validation
+
+To ensure the integrity of the proposal process, the `proposer` address must undergo validation checks. These checks include:
+- Verifying that the proposer has an adequate balance in the L2 BondManager contract.
+- Ensuring there are no pending withdrawal requests from this proposer on L2.
+
+If any of these validation checks fail, the proposal's manifest extraction will be bypassed, and a default manifest will be utilized instead.
+
+TODO: 
+
```
dantaik: Need to work on this.

dantaik:
## Challenge and Potential Solution

An undesirable side effect introduced by this PR is that a **proposer can submit proposals without maintaining sufficient bonds on L2**. We refer to these as *low-bond proposals*.

To mitigate this, one part of the solution is: if the proposer's L2 bond balance falls below a **protocol-defined threshold**, the system will derive **only a single empty block** from the proposal. This makes such proposals cheap to prove, reducing the burden on provers and disincentivizing spamming.

However, this leads to a critical question: **Who is responsible for proving low-bond proposals?**  
If no designated prover is specified, there's a risk that no one proves the block — and the chain stalls due to lack of finalization.

A potential solution is as follows:

- Every L2 block records its **designated prover**, either in the block header or the world state.
- The designated prover is **never `address(0)`**.
- When a low-bond proposal results in an empty block, that block **inherits its parent’s designated prover**.

### Implications:

- If a legitimate prover is the designated prover of the latest L2 block, and a series of low-bond proposals are submitted, that prover becomes responsible for proving all of them — **without earning proving fees**.
- If they fail to do so, they are **penalized via liveness bonds**, which can then be used to **incentivize other provers** to step in.

## Open Questions & Feedback Request

- This mechanism places significant burden on a designated prover in certain edge cases.
- How can we distribute or rotate this responsibility more fairly?
- Are there alternative approaches to handle proving of low-bond proposals without risking chain finality?

Tagging @ggonzalez94, @linoscope, and @davidtaikocha — would love your feedback or suggestions on better mechanisms.

ggonzalez94:
> A potential solution is as follows:
> 
> Every L2 block records its designated prover, either in the block header or the world state.
> The designated prover is never address(0).
> When a low-bond proposal results in an empty block, that block inherits its parent’s designated prover.
> Implications:
> If a legitimate prover is the designated prover of the latest L2 block, and a series of low-bond proposals are submitted, that prover becomes responsible for proving all of them — without earning proving fees.
> If they fail to do so, they are penalized via liveness bonds, which can then be used to incentivize other provers to step in.
> Open Questions & Feedback Request
> This mechanism places significant burden on a designated prover in certain edge cases.
> How can we distribute or rotate this responsibility more fairly?
> Are there alternative approaches to handle proving of low-bond proposals without risking chain finality?

I see a couple issues with this approach:
1) like @dantaik mentioned it is unfair with the parent's designated prover in the case of multiple low-bond proposals
2) perhaps more importantly we now have to deal with the case where that prover does not have a sufficient bond. You end up in a bit of a chicken and egg problem

A different proposed solution is #19923 where we use a common pool that can be funded by Taiko and/or as a tax to regular proposers when it is below a certain threshold

dantaik:
@claude please review again

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -167,33 +149,24 @@ abstract contract Inbox is EssentialContract, IInbox {
...
         (Proposal[] memory proposals, Claim[] memory claims) = _data.decodeProveData();
 
         require(proposals.length == claims.length, InconsistentParams());
         require(proposals.length != 0, EmptyProposals());
 
-        uint48[] memory proposalIds = new uint48[](proposals.length);
-        ClaimRecord[] memory claimRecords = new ClaimRecord[](proposals.length);
+        ClaimRecord[] memory claimRecords = _buildClaimRecords(config, proposals, claims);
 
         for (uint256 i; i < proposals.length; ++i) {
```
ggonzalez94: I think there's a bug here when we use the prove function with claim aggregation. Please check #19933

PR #19923, opened 2025-08-12
feat(protocol): a proposed solution to incentivize proving "low bond proposals"

ggonzalez94:
## Problem
  With bonds now on L2, proposers can submit L1 proposals without sufficient L2 bonds. These"low-bond proposals" will be derived as single empty blocks with no economic incentive for provers, potentially stalling the chain.

## Solution
 Implement a <s>self-sustaining </s>proving incentive  pool on L2:
 
  - <s>Proving Tax: Legitimate proposals contribute a small tax when the pool balance is low</s>
  - Proving Rewards: Provers of low-bond proposals receive rewards from this pool
  - Automatic Detection: The anchor transaction detects low-bond proposals during block execution
  
 <s> The "tax" on legitimate proposals should be extremely rare, since proposers are heavily disincentivized to propose without enough bonds(they pay L1 tx fees, and they will get slashed for not honoring their preconfs). We can further reduce the attack surface by allowing only a single proposal per L1 block.</s>
 
 There's no more tax. The pool is now handled on the BondManager directly and relies on deposits(i.e. Taiko Labs or the DAO fund it)

On the anchor's `updateState` function we check the proposer balance at the moment of that proposal(which should be the same value that nodes see at derivation time), and record if this is a low bond proposal. Then when the proposal is proven and we know the actual prover we can pay them a reward from the pool.
  
  
  ## TODO
If we like this design we should still do the following:

- [x] fix broken tests due to constructor parameter changes on the `ShastaAnchor`
- [x] @dantaik now that all the bond logic is on L2 should we just use `wei` as unit? We no longer need the optimization and is less error prone(no conversions)


packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -497,30 +499,33 @@ abstract contract Inbox is EssentialContract, IInbox {
         view
         returns (LibBondInstruction.BondInstruction[] memory bondInstructions_)
```
ggonzalez94: if we go with this approach we can further simplify and always return a single `BondInstruction` instead of an array since no longer have the case where this has to return an empty array

dantaik:
>  @dantaik now that all the bond logic is on L2 should we just use wei as unit? We no longer need the optimization and is less error prone(no conversions)

Yes.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: This is for claim record aggregation, so one record can cover multiple proposals and `bondInstructions_` can contains as many potential bond instructions as needed on L2 for all these blocks.

dantaik:
I’m not a fan of the tax idea—it’s another factor that could discourage validators from opting into preconfirmation. Instead, we could set up a pool funded by DAO treasury tokens. If the pool balance drops below a certain threshold, we replenish it.

For low-bond proposals, we can assign the last known designated prover and reward them with a fixed fee of 100 TAIKO tokens for each proof.

This approach isn’t perfect, but it’s acceptable as long as low-bond proposals are kept rare by ensuring such proposers incur losses. We can add an additional slashing condition for low-bond proposals, on top of the existing ones, so that these proposers are not only temporarily banned from preconfirming blocks but also lose some Ether or TAIKO tokens for each low-bond proposal.

However, if a proposer can submit an unlimited number of proposals, slashing per low-bond proposal may not be effective. In that case, we might need to throttle the number of proposals per L1 block or per epoch—for example, allowing at most two proposals per L1 block, which I find reasonable.

ggonzalez94:
> I’m not a fan of the tax idea—it’s another factor that could discourage validators from opting into preconfirmation. Instead, we could set up a pool funded by DAO treasury tokens. If the pool balance drops below a certain threshold, we replenish it.

The "tax" was just to ensure the protocol was self sustaining in case Taiko or the DAO stopped funding it. As long as the pool is above the threshold it should never kick in. But I can remove it if you prefer.

> For low-bond proposals, we can assign the last known designated prover and reward them with a fixed fee of 100 TAIKO tokens for each proof.

Why do you think it is better to assign someone explicitly? wouldn't it be better to make it permisionless and just reward whoever submitted the actual proof? or is the idea to avoid race conditions on L1 to submit the proof?

 > but it’s acceptable as long as low-bond proposals are kept rare by ensuring such proposers incur losses.
 
 I think there's already a strong disincentive with L1 fees and preconf slashing(when enabled), but we can further limit it. The only thing we have to make sure is that the reward is clearly lower than the cost of producing a low bond proposal

ggonzalez94:
Had to resolve a bunch of conflicts when merging, but should be ready now:
- removed the concept of taxes entirely. Proposers never have to pay for the low proposal bonds
- switched from gwei->wei for simplicity since now everything is handled on L2
- moved the proving pool reward to the BondManager and simplified the logic

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -46,28 +46,29 @@ abstract contract ShastaAnchor is PacayaAnchor {
...
+    uint256 public immutable lowBondProvingReward;
+
     // ---------------------------------------------------------------
     // State variables
     // ---------------------------------------------------------------
 
-    /// @notice Current state of the anchor proposal being processed.
-    /// @dev Two slots used to store the state:
-    State private _state;
+    State public _state;
```
ggonzalez94: a couple open questions:
1. if we go with this approach do we still ned to store `designatedProver` and `isLowBondProposal` on the state? As far as i can tell we were doing that do designate the previous prover
2. what's the benefit of having the state variables on a struct? isn't it easier to just have them flat on the contract?

ggonzalez94:
@claude please review

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -401,34 +403,33 @@ abstract contract Inbox is EssentialContract, IInbox {
...
-                    }
-                }
-                return bondInstructions;
+                // Very late proof - transfer provability bond
+                bondInstructions_[0] = LibBonds.BondInstruction({
+                    proposalId: _proposal.id,
+                    bondType: LibBonds.BondType.PROVABILITY,
+                    payer: _proposal.proposer,
+                    receiver: _claim.actualProver
+                });
```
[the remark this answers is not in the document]
ggonzalez94: not true

PR #19924, opened 2025-08-12
feat(protocol): add preconf router settings for handoverWindow

ggonzalez94:
from our Slack:
> can we put the handoverSlots onchain into the l1 config
> we can't update the handover from 4 to 6 because every driver has to be aware of the handover slots or they will reject blocks
> and it's a flag currently
> if it's onchain I. can add it to the driver to pull it down and we can change it at will
> l1 blob transactions have a 4 slot expected inclusion window - this is the cause of a large reorg


packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
```diff
@@ -9,4 +9,14 @@ interface IPreconfRouter is IProposeBatch {
     error ForcedInclusionNotSupported();
     error NotPreconferOrFallback();
     error ProposerIsNotPreconfer();
+
+    /// @notice Configuration struct for preconf-related settings
+    struct Config {
+        /// @notice The number of slots for hand over
+        uint256 handOverSlots;
```
cyberhorsey, line 16: should we add a  buffer at the end of the struct in case we add new configs like the other structs

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -71,6 +71,12 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
         );
     }
 
+    /// @inheritdoc IPreconfRouter
+    function getConfig() external pure returns (IPreconfRouter.Config memory) {
+        // TODO(jeff): determine the correct number of hand over slots before we deploy
+        return IPreconfRouter.Config({ handOverSlots: 4 });
```
cyberhorsey: 4 is correct, and then we add a driver pr, get all nodes to update with an ASAP mandatory update, wait a day or two, then update it to 6 or 8. do we want an update config method or do it via contract upgrade?

cyberhorsey (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
ggonzalez94: I think it requires the same amount of signatures, unless we assign it to a specific role(like we did with the ejecter). So maybe contract upgrade is fine?

packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
ggonzalez94: added

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
cyberhorsey: sure

packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
cyberhorsey: ty

dantaik:
If we change the config, we will need a hard fork, right? @YoGhurt111 @davidtaikocha ?

Maybe we have to hard code a new constant value in the code, without changing the conf object or any interface.

cyberhorsey:
> If we change the config, we will need a hard fork, right? @YoGhurt111 @davidtaikocha ?
> 
> Maybe we have to hard code a new constant value in the code, without changing the conf object or any interface.

client will need to know what block number it updated at and it would be annoying. You know solidity better than me so I trust you to implement such a thing where we can just add new values (such as "targetBlockTime" for example) into the "config" without having to make drastic changes and hard fork.

davidtaikocha:
> If we change the config, we will need a hard fork, right?

In theory, we may not need a hardfork, but still need to find a way to let all drivers / preconf-sidecars use the new value at the same time...

YoGhurt111:
> If we change the config, we will need a hard fork, right? @YoGhurt111 @davidtaikocha ?
> 
> Maybe we have to hard code a new constant value in the code, without changing the conf object or any interface.

I think probably not, this value is only meaningful to the sequencer. We can have a meeting call to complete the work at the same time to avoid hardfork.

YoGhurt111:
If we ask all drivers to use a new version(read the config on chain), isn't this the same as requiring users to use a fixed flag value?
Does this value change frequently?

cyberhorsey:
> If we ask all drivers to use a new version(read the config on chain), isn't this the same as requiring users to use a fixed flag value? Does this value change frequently?

it's the same the first time, and then not after. because we can update it onchain and the drivers will react to it.

cyberhorsey:
actually on second thought, I think only sequencer nodes care about the handoverSkipSlots. Regular nodes dont even need to know about it... but this would still be useful for when we have more than 2 sequencers, to be able to update it without contacting each sequencer node and updating exactly at the same time.

YoGhurt111 (approved):

dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -71,6 +71,11 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
         );
     }
 
+    /// @inheritdoc IPreconfRouter
+    function getConfig() external pure returns (IPreconfRouter.Config memory) {
+        return IPreconfRouter.Config({ handOverSlots: 4 });
```
YoGhurt111:
```suggestion
        return IPreconfRouter.Config({ handOverSlots: 6 });
```
@cyberhorsey If I remembered correctly, we decided it as `6`?

cyberhorsey: yes but it needs to be deployed as 4(the current one) then changed

cyberhorsey:
@ggonzalez94 @YoGhurt111 I think we need an update function with an event so we actually know when to change the handover slots, otherwise we need to poll it every slot, no?

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
YoGhurt111: I think no need, we fixed `4` in the client and when we can get the config on chain, we can read `6`

YoGhurt111:
> @ggonzalez94 @YoGhurt111 I think we need an update function with an event so we actually know when to change the handover slots, otherwise we need to poll it every slot, no?

I think 1 call per 4s is ok?

cyberhorsey:
> > @ggonzalez94 @YoGhurt111 I think we need an update function with an event so we actually know when to change the handover slots, otherwise we need to poll it every slot, no?
> 
> I think 1 call per 4s is ok?

works for me to poll it then

cyberhorsey:
@ggonzalez94 @YoGhurt111 we can merge this, deploy to hekla, rollout new version by building gavins PR, then upgrade it on Hekla, if it works well we can add it in the next client release (and upgrading the contract for mainnet can be done whenever beforehand since it doest break anything to deploy it early)

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
cyberhorsey: OK - we just need to make sure sequencers update their client first before we upgrade this contract then, rather than deploying the contract first then a second upgrade.

packages/protocol/contracts/layer1/preconf/iface/IPreconfRouter.sol
```diff
@@ -9,4 +9,14 @@ interface IPreconfRouter is IProposeBatch {
     error ForcedInclusionNotSupported();
     error NotPreconferOrFallback();
     error ProposerIsNotPreconfer();
+
+    /// @notice Configuration struct for preconf-related settings
+    struct Config {
```
linoscope, line 14: Don't we also want a "updateConfig" function that let's multisig that controls whitelist to upgrade the config? Especially considering contract upgrades will be passed to the DAO.

cyberhorsey: I also mentioned this but it got declined in favor of doing a contract upgrade

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
cyberhorsey: this should actually be 8, the agreed upon value, not 6

dantaik:
@ggonzalez94 lets merge this PR?

cyberhorsey:
@dantaik we will be handling this with a timestamp based hard fork since you are handing upgrade authority over to the DAO and it will be delayed to get this feature in

cyberhorsey:
reopened, saw slack message

dantaik:
Ok, I'll move on with the planned ownership transfer without upgrading this contract first. Then we'll ask the Security Council's help to get this contract upgraded as an emergency proposal.

PR #19927, opened 2025-08-13
feat(protocol):  add getNamedSettings to Inbox

dantaik:
This function can be used by off-chain software to synchronise configurations (such as handover slots)


ggonzalez94:
I'd like to understand the hardfork discussion better, because this seems more complicated than just the usual config settings we have

PR #19931, opened 2025-08-13
feat(protocol): lookahead store and eip-7917 based slasher for shasta

AnshuJalan:
The PR:
* Merges the `Overseer`(blacklist #19715 ) and `PreconfRouter2` into the `LookaheadStore`.
* Adds a lookahead slasher based on [EIP-7917](https://eips.ethereum.org/EIPS/eip-7917).

---

### Lookahead Slashing Overview

The idea behind lookahead slashing is as follows:

* We refer to the lookahead maintained by the preconfirmation protocol as the **"preconf lookahead"**, and the lookahead within the beacon state as the **"beacon lookahead."**
* There are two slashable faults related to lookahead posting:

#### Fault 1: Mismatched Validator at Slot X

* The validator inserted at slot X in the preconf lookahead does **not** match the validator present at slot X in the beacon lookahead.
* To clarify: we don't necessarily insert a validator key into the preconf lookahead. Instead, we make a claim that the beacon validator for slot X exists at a specific leaf index in an opted-in operator's registration tree in the URC.
* To prove this fault, we must show that the validator (`preconfLookaheadValPubKey`) at the leaf index of the opted-in operator **does not match** the validator (`beaconLookaheadValPubKey`) in the beacon lookahead.

#### Fault 2: Missing Validator in Preconf Lookahead

* A validator present in the beacon lookahead at slot X **belongs to an opted-in operator**, but was **not inserted** into the preconf lookahead (i.e., slot X in the preconf lookahead was empty).
* To prove this fault, we provide evidence that the validator (`beaconLookaheadValPubKey`) in the beacon lookahead belongs to a valid opted-in operator.

---

Backlog (To be addressed after first review and testing):
- [ ] Improve beacon merkleization using assembly
- [ ] Prevent redundant `MCOPY` in commented locations


AnshuJalan:
@claude please review

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -15,6 +16,7 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
     address public immutable protector;
```
AnshuJalan: Change this to `lookaheadSlasher`

packages/protocol/contracts/layer1/preconf/iface/IOverseer.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+/// @title IOverseer
+/// @custom:security-contact security@taiko.xyz
+interface IOverseer {
+    struct BlacklistTimestamps {
+        uint48 blacklistedAt;
+        uint48 unBlacklistedAt;
+        uint160 _reserved;
```
dantaik: since we don't have a deployment history, maybe we can remove this field?

packages/protocol/contracts/layer1/preconf/libs/LibEIP4788.sol
```diff
@@ -17,82 +19,95 @@ library LibEIP4788 {
...
     {
-        // Validator's BLS public key is verified against the hash tree root within Validator chunks
-        bytes32 pubKeyHashTreeRoot = sha256(abi.encodePacked(validatorBLSPubKey, bytes16(0)));
-        require(pubKeyHashTreeRoot == inclusionProof.validator[0], InvalidValidatorBLSPubKey());
+        BLS.Fp memory compressedValidatorPubKeyFp = BLSUtils.compress(_validatorPubKey);
 
-        // Validator is verified against the validator list in the beacon state
-        bytes32 validatorHashTreeRoot = LibMerkleUtils.merkleize(inclusionProof.validator);
+        // Shifts the 16-byte 0-padding to the end
+        bytes32 x = compressedValidatorPubKeyFp.a << 128 | compressedValidatorPubKeyFp.b >> 128;
```
dantaik, line 57: According to Cursor:
Screenshot 2025-08-27 at 2 32 59 PM

"Since reconstructing y from x involves complex modular arithmetic (cube, square root), a simpler approach would be to use the original uncompressed coordinates if available"

packages/protocol/contracts/layer1/preconf/libs/LibEIP4788.sol
```diff
@@ -17,82 +19,95 @@ library LibEIP4788 {
...
+        bytes32 y = compressedValidatorPubKeyFp.b << 128;
+
+        // Verify: Validator chunks contains the validator public key
+        bytes32 pubKeyHashTreeRoot = sha256(abi.encodePacked(x, y));
```
dantaik, line 61: According to Cursor:

Screenshot 2025-08-27 at 2 51 59 PM

dantaik:
@Claude, could you please review new and modified libraries in this PR that implement the SSZ (Simple Serialize) Merkleization specification, as defined in the Ethereum consensus layer (e.g., the Ethereum 2.0 specification)?

dantaik:
@Claude, could you please review new and modified libraries in this PR that implement the SSZ (Simple Serialize) Merkleization specification, as defined in the Ethereum consensus layer (e.g., the Ethereum 2.0 specification)?

packages/protocol/contracts/layer1/preconf/libs/LibBeaconMerkleUtils.sol
```diff
@@ -2,16 +2,19 @@
...
+library LibBeaconMerkleUtils {
     uint256 internal constant CHUNKS_LENGTH = 8;
     uint256 internal constant TMP_LENGTH = 4;
 
     function hash(bytes32 a, bytes32 b) internal pure returns (bytes32) {
+        // TODO: reuse word with assembly
         return sha256(abi.encodePacked(a, b));
     }
 
+    // TODO: use calldata for chunks
```
dantaik, line 17: Screenshot 2025-08-27 at 3 03 09 PM

Cursor suggested the following:

```solidity
function merkleize(bytes32[] memory chunks) internal pure returns (bytes32) {
    if (chunks.length == 0) return bytes32(0);
    if (chunks.length == 1) return chunks[0];
    
    // SSZ merkleization: handle arbitrary number of chunks
    bytes32[] memory nextLevel = new bytes32[]((chunks.length + 1) / 2);
    
    for (uint256 i = 0; i < chunks.length; i += 2) {
        if (i + 1 < chunks.length) {
            nextLevel[i / 2] = hash(chunks[i], chunks[i + 1]);
        } else {
            // Single chunk at the end
            nextLevel[i / 2] = chunks[i];
        }
    }
    
    return merkleize(nextLevel);
}


function verifyProof(
    bytes32[] memory proof,
    bytes32 root,
    bytes32 leaf, // This should be the hashed leaf node
    uint256 index
) internal pure returns (bool) {
    bytes32 computedHash = leaf;

    for (uint256 i = 0; i < proof.length; i++) {
        if (index % 2 == 0) {
            computedHash = hash(computedHash, proof[i]); // Use the standard hash function
        } else {
            computedHash = hash(proof[i], computedHash);
        }
        index = index / 2;
    }
    return computedHash == root;
}
```

dantaik: Cursor/DeepSeek seems to approve this modified version:


https://gist.github.com/dantaik/4db7766ab2122806edfc227b681af151

packages/protocol/contracts/layer1/preconf/libs/LibBlockHeader.sol
```diff
@@ -31,7 +31,7 @@ library LibBlockHeader {
     }
 
     function encodeRLP(BlockHeader memory _blockHeader) internal pure returns (bytes memory) {
-        LibRLP.List memory list = LibRLP.l();
+        LibRLP.List memory list = LibRLP.p();
```
dantaik, line 31: should we remove the following fields from the Header definitation as they are not available for Taiko:

- uint64 blobGasUsed;
- uint64 excessBlobGas;
- bytes32 parentBeaconBlockRoot;

These fields are not used in the encodeRLP function.

packages/protocol/contracts/shared/governance/SimpleMultisig.sol
```diff
@@ -0,0 +1,166 @@
...
+
+            lastSigner = currentSigner;
+        }
+    }
+
+    // Virtual functions
+    // -----------------------------------------------------------------------------------
+
+    function _getAddSignerDomainSeparator() internal pure virtual returns (bytes32) {
+        return keccak256("ADD_SIGNER");
```
dantaik:
```suggestion
        return keccak256("TAIKO_SIMPLE_MULTISIG_ADD_SIGNER");
```

packages/protocol/contracts/shared/governance/SimpleMultisig.sol
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    function _getRemoveSignerDomainSeparator() internal pure virtual returns (bytes32) {
+        return keccak256("REMOVE_SIGNER");
```
dantaik:
```suggestion
        return keccak256("TAIKO_SIMPLE_MULTISIG_REMOVE_SIGNER");
```

packages/protocol/contracts/shared/governance/SimpleMultisig.sol
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    function _getUpdateSigningThresholdDomainSeparator() internal pure virtual returns (bytes32) {
+        return keccak256("UPDATE_SIGNING_THRESHOLD");
```
dantaik:
```suggestion
        return keccak256("TAIKO_SIMPLE_MULTISIG_UPDATE_SIGNING_THRESHOLD");
```

packages/protocol/contracts/shared/governance/SimpleMultisig.sol
```diff
@@ -0,0 +1,166 @@
...
+
+    /// @param _ds The domain separator
+    /// @param _data The data to be hashed and signed
+    /// @param _signatures The signatures to be verified
+    function _verifySignatures(bytes32 _ds, bytes32 _data, bytes[] memory _signatures) internal {
+        require(_signatures.length >= signingThreshold, InsufficientSignatures());
+
+        bytes32 digest;
+        unchecked {
+            digest = keccak256(abi.encode(_ds, nonce++, _data));
```
dantaik: Screenshot 2025-08-27 at 3 40 15 PM

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,140 @@
...
+
+    function _getUnblacklistDomainSeparator() internal pure virtual returns (bytes32) {
+        return keccak256("TAIKO_ALETHIA_UNBLACKLIST_OVERSEER");
+    }
+
+    // Overrides
+    // -----------------------------------------------------------------------------------
+
+    function _getAddSignerDomainSeparator() internal pure override returns (bytes32) {
+        return keccak256("TAIKO_ALETHIA_ADD_OVERSEER_SIGNER");
```
dantaik:
```suggestion
        return keccak256("TAIKO_ALETHIA_OVERSEER_ADD_SIGNER");
```

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    function _getRemoveSignerDomainSeparator() internal pure override returns (bytes32) {
+        return keccak256("TAIKO_ALETHIA_REMOVE_OVERSEER_SIGNER");
```
dantaik:
```suggestion
        return keccak256("TAIKO_ALETHIA_OVERSEER_REMOVE_SIGNER");
```

packages/protocol/contracts/layer1/preconf/impl/Overseer.sol
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    function _getUpdateSigningThresholdDomainSeparator() internal pure override returns (bytes32) {
+        return keccak256("TAIKO_ALETHIA_UPDATE_OVERSEER_SIGNING_THRESHOLD");
```
dantaik:
```suggestion
        return keccak256("TAIKO_ALETHIA_OVERSEER_UPDATE_THRESHOLD");
```

dantaik (dismissed):

packages/protocol/contracts/layer1/preconf/libs/LibEIP4788.sol
AnshuJalan: The file LibBLS12381 is no longer being used. My logic is based on Solady's BLS implementation. I will remove the unused file.

packages/protocol/contracts/layer1/preconf/libs/LibEIP4788.sol
AnshuJalan: I don't think claude has context of how beacon state is organised. It is not well-documented on the internet.

packages/protocol/contracts/layer1/preconf/libs/LibBeaconMerkleUtils.sol
AnshuJalan: These functions are from the POC implementation and were tested on the mainnet Beacon chain data. It's strange AI models outright think it is incorrect

packages/protocol/contracts/layer1/preconf/libs/LibBlockHeader.sol
AnshuJalan: Agreed.

packages/protocol/contracts/shared/governance/SimpleMultisig.sol
AnshuJalan: This is not a publicly usable nonce. It is isolated to the signers of the multisig, so this problem cannot happen

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -13,8 +14,10 @@ import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 contract LookaheadStore is ILookaheadStore, EssentialContract {
     IRegistry public immutable urc;
     address public immutable protector;
+    address public immutable lookaheadSlasher;
     address public immutable preconfSlasher;
     address public immutable preconfRouter;
```
ggonzalez94: should we update this to inbox given this will now interact with shasta?(and we don't have a router anymore for shasta)

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -13,8 +14,10 @@ import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 contract LookaheadStore is ILookaheadStore, EssentialContract {
     IRegistry public immutable urc;
     address public immutable protector;
```
ggonzalez94: what's the protector?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -55,7 +62,7 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
 
         if (_registrationRoot == bytes32(0)) {
             // If the registration root is 0, the lookahead is posted by a whitelist preconfer
-            // (via preconf router) or it is posted the protector.
+            // (via preconf router) or it is posted the lookahead slasher.
             if (msg.sender == preconfRouter) {
                 require(isLookaheadRequired_, LookaheadNotRequired());
```
ggonzalez94: should we revisit this part? since we don't have a preconfRouter anymore and the inbox does not have any logic to post the lookahead. Not sure what purpose this served on a previous design?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -13,8 +14,10 @@ import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 contract LookaheadStore is ILookaheadStore, EssentialContract {
```
ggonzalez94: Should this contract implement [IProposerChecker](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/shasta/iface/IProposerChecker.sol)? I'm failing to understand what function from this contract the inbox would use to validate if `msg.sender` is allowed to propose. If yes, then we probably need to modify that interface to accept the actual lookahead so that it is passed to this contract and validated against the stored hash

packages/protocol/contracts/layer1/preconf/impl/LookaheadSlasher.sol
```diff
@@ -0,0 +1,337 @@
...
+import "src/layer1/preconf/libs/LibPreconfUtils.sol";
+import "src/layer1/preconf/libs/LibPreconfConstants.sol";
+import "src/shared/common/EssentialContract.sol";
+import "@eth-fabric/urc/lib/MerkleTree.sol";
+
+/// @title LookaheadSlasher
+/// @dev The lookahead contained within the beacon state is referred to
+/// as the "beacon lookahead"
+/// whereas, the lookahead maintained by the preconfing protocol is referred to
+/// as the "preconf lookahead"
```
ggonzalez94: nit: redundant since it is already in the interface?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -259,11 +271,24 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
         // For the poster, we consider the latest collateral value.
         // For the operators within lookahead, we consider the collateral value at the start of
         // the current epoch.
-        uint256 collateralWei = _slasher == protector
+        uint256 collateralWei = _slasher == lookaheadSlasher
             ? operatorData_.collateralWei
             : urc.getHistoricalCollateral(_registrationRoot, _timestamp);
         require(collateralWei >= _minCollateral, OperatorHasInsufficientCollateral());
 
+        IOverseer.BlacklistTimestamps memory blacklistTimestamps =
```
ggonzalez94: this `_validateOperator` function is called both for validating lookahead operators as well as validating lookahead posters right?
Shouldn't these two validations be a bit different? i.e. I don't think it makes sense to check if the poster is blacklisted for example, since that's a mechanism used for non objective faults, and posting the wrong lookahead is always objective, right?

packages/protocol/contracts/layer1/preconf/iface/IOverseer.sol
```diff
@@ -0,0 +1,66 @@
...
+
+/// @title IOverseer
+/// @custom:security-contact security@taiko.xyz
+interface IOverseer {
+    struct BlacklistTimestamps {
+        uint48 blacklistedAt;
+        uint48 unBlacklistedAt;
+    }
+
+    /// @dev These delays prevent the lookahead from being messed up mid-epoch
```
ggonzalez94: Wrong comment?
These are set to 1 day in the implementation and I don't think they explicitly prevent the lookahead from changing mid epoch right? They are more about preventing rapid blacklist/unblacklist cycling

ggonzalez94 (commented):
Did a first pass, there's still context I'm working on catching up to be able to review better. I think some of my other broader points are :
- As discussed in the document we should try to merge the LookaheadStore with the blacklisting functionality and just assign that to a role. This would allow us to get rid of the `SimpleMultisig` contract?
- I feel the `_validateOperator` function in the LookAheadStore(especially the comments) is a bit convoluted, since it tries to explain the validations for two different cases(the lookahead poster and regular validators).
- Update this for shasta contracts and terminology

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -24,54 +29,206 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
+
+        LookaheadData memory data = abi.decode(_lookaheadData, (LookaheadData));
 
-        bool isLookaheadRequired_ = isLookaheadRequired();
+        require(
+            data.slotIndex == type(uint256).max || data.slotIndex < data.currLookahead.length,
+            InvalidSlotIndex()
+        );
+
+        uint256 epochTimestamp = LibPreconfUtils.getEpochTimestamp();
```
ggonzalez94: nit: epochStartTimestamp timestamp for clarity? I had to go to the implementation of `getEpochTimestamp` to understand this was the start of the epoch

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -24,54 +29,206 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
+    }
+
+    /// @inheritdoc IBlacklist
+    function removeOverseers(address[] calldata _overseers) external override onlyOwnerOrOverseer {
+        for (uint256 i = 0; i < _overseers.length; ++i) {
+            address overseer = _overseers[i];
+            require(overseers[overseer], OverseerDoesNotExist());
+            overseers[overseer] = false;
+        }
+        emit OverseersRemoved(_overseers);
```
ggonzalez94, line 329: doesn't it make more sense to have these two functions in the `Blacklist` abstract contract instead of here?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -24,54 +29,206 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
+    modifier onlyOwnerOrOverseer() {
+        require(msg.sender == owner() || overseers[msg.sender], NotOwnerOrOverseer());
+        _;
+    }
+
     /// @inheritdoc ILookaheadStore
-    function updateLookahead(
-        bytes32 _registrationRoot,
-        bytes calldata _data
+    function checkProposer(
```
ggonzalez94, line 65: I'm having a really hard time wrapping my head and readin trough this function. it has too large and too many nested conditions. I hope there's a way we can simplify it using internal functions and/or libraries to make it clearer what the intent is

ggonzalez94: also we should consider changing the name, since it does so much more than checking who the proposer is.
Maybe something like `authorizeProposer`?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -24,54 +29,206 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
...
         }
 
-        return _updateLookahead(LibPreconfUtils.getEpochTimestamp(1), lookaheadSlots);
+        if (validateWhitelistPreconfer) {
+            _validateWhitelistPreconfer(_proposer);
+        } else {
+            _validateOptedInPreconfer(_proposer, _lookaheadSlot);
+        }
+
+        return uint64(endOfSubmissionWindowTimestamp);
```
ggonzalez94: Is there a reason why we wanna return the `endOfSubmissionWindowTimestamp` from this function? If the inbox is calling this should the answer be is this proposer allowed to post yes/no?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -287,6 +472,64 @@ contract LookaheadStore is ILookaheadStore, EssentialContract {
         view
         returns (LookaheadHash storage)
     {
-        return lookahead[_epochTimestamp % getConfig().lookaheadBufferSize];
+        return lookahead[_epochTimestamp % getLookaheadStoreConfig().lookaheadBufferSize];
+    }
+
+    function _validateWhitelistPreconfer(address _proposer) internal view {
```
ggonzalez94, line 614: I think it is worth asking: why do we need to keep both whitelisted + fallback preconfer? If for whatever reason we wanna have a single whitelisted preconfer(which is the same as the fallabck) we could eject all the other whitelisted preconfers and just keep one right?

packages/protocol/contracts/layer1/preconf/impl/Blacklist.sol
```diff
@@ -0,0 +1,107 @@
...
+    function addOverseers(address[] calldata _overseers) external virtual;
+
+    /// @inheritdoc IBlacklist
+    function removeOverseers(address[] calldata _overseers) external virtual;
+
+    // Views
+    // -----------------------------------------------------------------------------------
+
+    function getBlacklistConfig() public pure returns (BlacklistConfig memory) {
+        return BlacklistConfig({ blacklistDelay: 1 days, unblacklistDelay: 1 days });
```
ggonzalez94, line 92: maybe this is something I forgot from our previous conversation:
But my understanding is that blacklisting does not affect already posted Lookaheads, so I'm not sure it will be an issue if the overseer blacklists/unblacklists a validator before this delay.
I understand it might be desirable that they don't constantly toggle this value, but since they are designated and can be removed by the DAO perhaps we don't need to enforce this at the contract level?

ggonzalez94 (commented):
another round of reviews. Now moving to the slashing logic finally.
Feature wise I think this is looking good. My biggest feedback I believe is that the `checkProposer` needs to be refactored to be more readable.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
ggonzalez94: Here's a quick, AI generated attempt to make this cleaner: https://github.com/taikoxyz/taiko-mono/pull/20158

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
ggonzalez94: also should it be `onlyOwner` only? I don't think we want the overseers being able to add others and remove others/themselves?

packages/protocol/contracts/layer1/preconf/impl/LookaheadSlasher.sol
```diff
@@ -0,0 +1,331 @@
...
+        onlyFrom(urc)
+        returns (uint256)
+    {
+        // Todo: move to calldata
+        ILookaheadStore.LookaheadSlot[] memory lookaheadSlots =
+            abi.decode(_commitment.payload, (ILookaheadStore.LookaheadSlot[]));
+
+        // (EvidenceLookahead, EvidenceBeaconValidator, EvidenceInvalidOperator ||
+        // EvidenceMissingOperator)
+        (bytes calldata evidenceX, bytes calldata evidenceY, bytes calldata evidenceZ) =
```
ggonzalez94: I imagine these names are remnant for when we were using similar terminology in the `InboxWrapper`, but we should definitely change it

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: Makes sense, will change

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: I added this based on Daniel's suggestion in your former PR. If we only want the owner to add the overseer then yes, this should be in the Blacklist contract

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: The function reverts if the proposer is not allowed to propose.
`endOfSubmissionWindowTimestamp` that is returned is also the one stored in the metadata and later used by the anchor transaction for preconf slashing

AnshuJalan: So, this logic is somewhat connected with the WIP preconf slasher logic (in notion doc)

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: I believe @dantaik suggested the extra fallback in the original preconf router implementation. I cannot remember what the exact rationale was, but yes, it seems a bit redundant

packages/protocol/contracts/layer1/preconf/impl/Blacklist.sol
AnshuJalan: This delay can actually be just `2 epochs`. The problem is not with affecting the current lookahead, but with affecting the lookahead slashing logic. If we have a blacklist -> unblacklist -> blacklist again, we will have no way figure out when did the first blacklisting happen (unless we start storing historical data, which seems a bit too much), so we cannot slash a lookahead poster for failing to include a non-blacklisted preconfer.

AnshuJalan: But just 2 epochs are not very safe since building and posting evidence for a faulty lookahead may take some time.

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
ggonzalez94: I actually have a [PR](https://github.com/taikoxyz/taiko-mono/pull/20201) to remove this. We can adapt this later

packages/protocol/contracts/layer1/preconf/impl/Blacklist.sol
ggonzalez94: thanks for the clarification. If we can incorporate this in the comments or natspec would be great. I'm sure I'll forget at a later point and will ask you again otherwise 😄

.vscode/settings.json
```diff
@@ -4,7 +4,7 @@
     "editor.formatOnSave": false
   },
   "[solidity]": {
-    "editor.defaultFormatter": "NomicFoundation.hardhat-solidity"
+    "editor.defaultFormatter": "JuanBlanco.solidity"
```
ggonzalez94: not sure if most IDEs are picking this automatically, but if yes this might cause too many diffs on every PR

ggonzalez94 (approved):
LGTM! Left a few final comments, but I think we should be able to merge this soon. I still see the tests are commented out. @AnshuJalan will we re-enable them in a separate PR?

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
ggonzalez94: you mean [this comment](https://github.com/taikoxyz/taiko-mono/pull/20080#discussion_r2324554164)? I think the spirit of the comment was the overseer to be able to change their own address, not add or remove others. Even then I think this should not happen often at all, so I would be fine if for this PR we stick with `onlyOwner` unless @dantaik thinks otherwise

.vscode/settings.json
AnshuJalan: I think this was a mis-commit

packages/protocol/contracts/layer1/preconf/impl/Blacklist.sol
AnshuJalan: Added!

davidtaikocha (approved):

ggonzalez94 (approved):

PR #19933, opened 2025-08-13
fix(protocol): fix issue with prove function when aggregating claims

ggonzalez94:
## Issue
The prove() function iterates over proposals.length but accesses claimRecords[i], causing an array out-of-bounds error when claim aggregation isenabled. With aggregation, multiple continuous proposals are combined into fewer claim records (e.g., 5 proposals → 2 aggregated records), breaking the 1:1 assumption.

## Solution
- Changed loop to iterate over claimRecords.length instead of proposals.length

### DIFF
--- a/packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
+++ b/packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
@@ -95,10 +95,9 @@ interface IInbox {
     /// @param proposal The proposal that was proposed.
     event Proposed(Proposal proposal, CoreState coreState);
 
-    /// @notice Emitted when a proof is submitted for a proposal.
-    /// @param proposal The proposal that was proven.
+    /// @notice Emitted when a proof is submitted
     /// @param claimRecord The claim record containing the proof details.
-    event Proved(Proposal proposal, ClaimRecord claimRecord);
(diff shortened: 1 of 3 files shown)


ggonzalez94:
@claude review again

dantaik (approved):

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -42,6 +42,8 @@ interface IInbox {
 
     /// @notice Represents a claim about the state transition of a proposal.
     struct Claim {
+        /// @notice The proposal's ID.
+        uint48 proposalId;
```
dantaik, line 46: Actually we can fix the issue without adding this field. Please see - https://github.com/taikoxyz/taiko-mono/pull/19932/files#diff-4a1d9a13b9e3582b44595ea8983a7e7aac27cc64b4a7567c5d835e9abb54d3c8

dantaik:
Closed in favor of https://github.com/taikoxyz/taiko-mono/pull/19934, sorry about it @ggonzalez94

dantaik:
> Closed in favor of #19934, sorry about it @ggonzalez94

After a second thought, I think I prefer yours. so I'll go head to merge it.

PR #19934, opened 2025-08-14
fix(protocol): fix finalisation issues

dantaik:


packages/protocol/contracts/layer1/shasta/libs/LibBlobs.sol
```diff
@@ -20,9 +20,9 @@ library LibBlobs {
     /// in this transaction.
     struct BlobReference {
         /// @notice The starting index of the blob.
-        uint48 blobStartIndex;
+        uint8 blobStartIndex;
         /// @notice The number of blobs.
-        uint32 numBlobs;
+        uint8 numBlobs;
```
ggonzalez94, line 25: @dantaik maybe I'm being paranoic here, and the latest numbers are less ambitious than [this](https://x.com/VitalikButerin/status/1895863075527700959) but I think the optimizations we get from using uint8 here(it is just calladata compression once/if we incorporate a custom encoding algorithm) are not worth the risk of this silently starting to fail in the future because blobs have exceeded 256.

What about if we use `uint16`?

ggonzalez94: https://github.com/taikoxyz/taiko-mono/pull/19937

PR #19935, opened 2025-08-14
refactor(protocol): remove coreStateHash

dantaik:
With this change, propose() will have 1 less SSTORE but 1 more SLOAD.


dantaik:
@claude please review

PR #19963, opened 2025-08-17
chore(protocol): optimize pacaya fork to reduce heavy gas usage

ggonzalez94:
## The Problem
The gas cost for proposing grows with the number of blocks, but that growth is **very aggressive**.
Before enabling preconfs when each batch only had 1-2 blocks our cost for proposing was ~200k gas. Now with hundreds of blocks per batch **it can get as bad as 5M gas and an average of ~1.5M gas since we enabled preconfs**. That’s an almost 10x increase.
This is reducing dramatically after Shasta, but we need to do something sooner. Otherwise if gas costs on L1 spike it will become impossible to propose.

## Major contributors to gas cost
- [x] copy return values and memory expansion for the`BatchInfo` struct
- [ ] `BatchProposed` event. Changing this will require a hard fork, so this change is reverted
- [x] abi encoding and decoding of the BatchParams
- [ ] calldata - harder to reduce since it would require more changes in the node
- [ ] `_validateBatchParams`

## Results
The following are the results for a [batch proposal](https://etherscan.io/tx/0x2c65e26d179f301c8d5367e0e80c9f281bc208bdf77fb05677aace4f8b7bf3ee/advanced) with 432 blocks that consumed almost 4M gas

- Initial gas consumption: 3,989,014
   - After removing BatchInfo from the return data: 2,455,434 (`-38.4%`)
   - Not decoding the batch on the `TaikoWrapper`: 2,026,588 (`-17.5%`)
 **Total reduction: `-49.2%`**

If we are willing to remove the `BatchProposed` event we can get another ~900k gas saving, but this will require changes in the node and likely a fork.

## Changes required in the nodes to adopt to this

The current proposal requires no changes to the node, assuming no one was using the return values off-chain or the `proposeBatchWithExpectedLastBlockId`(which I checked was never called and was intended to be used with the `ProverSet`)

## TODO
- [x] fix pipeline since now we have a test that forks from mainnet


dantaik:
These changes may need a hardfork, does it? @davidtaikocha @YoGhurt111

davidtaikocha:
> These changes may need a hardfork, does it? @davidtaikocha @YoGhurt111

`returns` is not part of the function selector, but `BatchInfo` is used in `event BatchProposed`, so these changes are not backwards compatible.

YoGhurt111:
> These changes may need a hardfork, does it? @davidtaikocha @YoGhurt111

We can try to get data from multiple channels on the client to avoid hardfork.
Currently it is not backwards compatible. Need further development.

packages/protocol/contracts/layer1/based/ITaikoInbox.sol
```diff
@@ -65,7 +65,6 @@ interface ITaikoInbox {
     struct BatchInfo {
         bytes32 txsHash;
         // Data to build L2 blocks
-        BlockParams[] blocks;
```
davidtaikocha: > BatchProposed event. It emits all the blocks data that we already have in calldata

IMO another reason that we chose to emit all `BlockParams[] blocks;` data, is because if `proposeBatch` is not being called directly, sometimes it's hard to get its calldata (may require the `trace_log`).

ggonzalez94: @davidtaikocha @dantaik I added this back to the event to avoid a fork. please check it, the current proposal should require zero changes to the node and still achieves almost 50% savings for bigger batches.

> is because if proposeBatch is not being called directly, sometimes it's hard to get its calldata (may require the trace_log)

If we are willing to make this change later, we could retrieve the calldata directly from the external call to the `PreconfRouter` instead

YoGhurt111:
LGTM after fixing the bug.

YoGhurt111 (approved):

packages/protocol/contracts/layer1/based/IProposeBatch.sol
```diff
@@ -12,12 +12,11 @@ interface IProposeBatch {
     /// @param _params ABI-encoded parameters.
     /// @param _txList The transaction list in calldata. If the txList is empty, blob will be used
     /// for data availability.
-    /// @return info_ The info of the proposed batch.
-    /// @return meta_ The mmetadata of the proposed batch.
+    /// @return meta_ The metadata of the proposed batch.
```
dantaik, line 15: shall we add a new function instead?

packages/protocol/contracts/layer1/based/IProposeBatch.sol
```diff
@@ -12,12 +12,11 @@ interface IProposeBatch {
...
     function proposeBatch(
         bytes calldata _params,
         bytes calldata _txList
     )
         external
-        returns (ITaikoInbox.BatchInfo memory info_, ITaikoInbox.BatchMetadata memory meta_);
+        returns (ITaikoInbox.BatchMetadata memory meta_);
```
dantaik, line 21: Shall we define a new struct to be returned by this function (replacing BatchMetadata) to only include data needed by the caller?

dantaik (changes requested):

dantaik (approved):

packages/protocol/contracts/layer1/based/TaikoInbox.sol
```diff
@@ -125,6 +123,7 @@ abstract contract TaikoInbox is EssentialContract, ITaikoInbox, IProposeBatch, I
                 // this is a normal batch, blobs are created and used in the current batches.
                 // firstBlobIndex can be non-zero.
                 require(params.blobParams.numBlobs != 0, BlobNotSpecified());
+                require(params.blobParams.blobHashes.length == 0, InvalidBlobParams());
```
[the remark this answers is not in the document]
YoGhurt111: @ggonzalez94 We can remove this line, bc we already checked in https://github.com/taikoxyz/taiko-mono/blob/585887c4bd836c53a2f9c1446ee80fc9e4927558/packages/protocol/contracts/layer1/based/TaikoInbox.sol#L122

YoGhurt111:
```suggestion
```

PR #19966, opened 2025-08-18
feat(protocol): add lastVerifiedTimestamp

dantaik:
@davidtaikocha will this new field in CoreState help driver to find info faster?


davidtaikocha:
> @davidtaikocha will this new field in CoreState help driver to find info faster?

IMO this will help `proposer` to assemble the params faster.

YoGhurt111 (approved):

PR #19967, opened 2025-08-18
feat(taiko-client): fallback to sgx proof

YoGhurt111:


davidtaikocha (approved):

packages/taiko-client/prover/proof_submitter/proof_submitter.go
```diff
@@ -197,6 +204,9 @@ func (s *ProofSubmitterPacaya) RequestProof(ctx context.Context, meta metadata.T
 					meta,
 					startAt,
 				); err != nil {
+					if time.Since(startAt) > maxProofRequestTimeout {
```
smtmfft, line 209: should we reset the startAt if this one is a fallback from zk?

smtmfft (approved):

packages/taiko-client/prover/proof_submitter/proof_submitter.go
YoGhurt111: I originally thought it wouldn't matter not to reset startAt, since the log at L185 already indicates this is fallback behavior?
But it would be better to reset the startAt, will do it later

cyberhorsey (approved):

dantaik (approved):

packages/taiko-client/prover/proof_submitter/proof_submitter.go
cyberhorsey: is it OK to merge?

YoGhurt111: Will optimize later and test on Hekla

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main,feature/fallback_sgx_proof]
```
```suggestion
    branches: [main]
```

packages/taiko-client/prover/proof_submitter/proof_submitter.go
YoGhurt111: Can be merged.

johntaiko (approved):

PR #19971, opened 2025-08-18
feat(protocol): more fields needed? (WIP)

dantaik:
@YoGhurt111 @cyberhorsey @davidtaikocha 

Which of the new fields in this PR are needed to make building blocks easier?


packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -81,6 +81,14 @@ interface IInbox {
         uint48 nextProposalId;
         /// @notice The ID of the last finalized proposal.
         uint48 lastFinalizedProposalId;
+        /// @notice Block in which the last proposal was proposed.
+        uint48 lastProposedIn;
+        /// @notice Timestamp at which the last proposal was proposed.
+        uint48 lastProposedAt;
```
davidtaikocha, line 87: IMO if these two fields are stored in `CoreState`, then won't help a lot. For all the software wants to know `lastProposedIn` and `lastProposedAt`, they still have to scan the L1 chain from the latest tip and try to find the first `Proposed` event to decode the latest `CoreState`, then they will know these two fields already, since they have the event log.

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -81,6 +81,14 @@ interface IInbox {
...
+        /// @notice Block in which the last fanilization occurred.
+        uint48 lastFinalizedIn;
+        /// @notice Timestamp at which the last fanilization occurred.
+        uint48 lastFinalizedAt;
```
davidtaikocha, line 91: And these two will be helpful to make assembling the params of `propose()` easier.

PR #19969, opened 2025-08-18
feat(taiko-client): Add duration metrics to PreconfBlockApiServer

thedevbirb:
Rationale: to help understanding whether the `/preconfBlock` endpoint is sometimes blocked by p2p methods, and identify potential bottlenecks.

PR #19975, opened 2025-08-19
fix(taiko-client): elapsed time in histograms

thedevbirb:
After playing around with #19969 a bit more, I realized histogram buckets are expressed in seconds, so the data scraped from Prometheus was off. I'm sorry for that. This PR fixes it, and better formats elapsed time in milliseconds.

This is how metrics currently look like on Hekla, while not being the active sequencer.

image

The 0.99-0.999 quantiles can be on the higher side in the >1s range because the default bucket used by Prometheus library are:

```go
// DefBuckets are the default Histogram buckets. The default buckets are
// tailored to broadly measure the response time (in seconds) of a network
// service. Most likely, however, you will be required to define buckets
// customized to your use case.
var DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
```

So buckets above 1s are a bit coarse. However I have observed some high values myself but looking at the logs:

```
DEBUG[08-19|11:30:16.557] OnUnsafeL2Payload completed              elapsed=1659ms
DEBUG[08-19|11:30:57.298] OnUnsafeL2Payload completed              elapsed=1884ms
DEBUG[08-19|11:32:53.034] OnUnsafeL2Payload completed              elapsed=1949ms
DEBUG[08-19|11:33:53.458] OnUnsafeL2Payload completed              elapsed=2390ms
DEBUG[08-19|11:34:42.101] OnUnsafeL2Payload completed              elapsed=1259ms
```

@YoGhurt111 @cyberhorsey if you think more granular buckets are needed, please let me know!


YoGhurt111 (approved):

thedevbirb:
Decided to add a bit more buckets in the latest commit, so we have more accurate measurements in the [0.5s, 2.5s] range.

cyberhorsey:
will build off this PR to add more granular metrics, thank you very much!

cyberhorsey:
@thedevbirb small changes requested!

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -74,9 +74,9 @@ func (s *PreconfBlockAPIServer) BuildPreconfBlock(c echo.Context) error {
 
 	start := time.Now()
 	defer func() {
-		elapsed := float64(time.Since(start).Milliseconds())
-		metrics.DriverPreconfBuildPreconfBlockDuration.Observe(elapsed)
-		log.Debug("BuildPreconfBlock completed", "elapsed", elapsed)
+		elapsed_ms := time.Since(start).Milliseconds()
```
cyberhorsey: golang variable names should be camelCase not snake_case, you rustacean

cyberhorsey (changes requested):

packages/taiko-client/driver/preconf_blocks/api.go
thedevbirb: You're right, really surprised go preferred this dark route and not inherit holy C's snake casing

cyberhorsey: it's flat out camel worship tbh, snake is an evil animal

dantaik (approved):

cyberhorsey (approved):

PR #19984, opened 2025-08-20
fix(protocol): fix a bug with reusing pacaya slots for proposal

ggonzalez94:
This is one possible solution to the issue described [here](https://github.com/taikoxyz/taiko-mono/pull/19882#discussion_r2288498864). The other option is just not to reuse the storage slots from pacaya for proposals.
It also removes a couple unnecessary checks for proposal length(if the proposer sends a larger array they just waste calldata, but it is not an issue).

This solution initializes the ring buffer with empty markers in the `init` function. Since those storage slots are already non-zero(reused from pacaya) each `SSTORE` will "only" cost 5k gas. This means the deployment will be more expensive, but it is a one time operation, and all proposals from the first one will be cheaper.
**Ring buffer is decreased to a more reasonable number of 360 - more than one and a half day of unfinalized proposals**

### DIFF
--- a/packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
+++ b/packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
@@ -41,6 +41,14 @@ abstract contract Inbox is EssentialContract, IInbox {
     }
 
     // ---------------------------------------------------------------
+    // Constants
+    // ---------------------------------------------------------------
+    /// @dev Empty slot marker to distinguish uninitialized slots from actual proposal hashes.
+    /// keccak256("EMPTY_PROPOSAL_SLOT_V1")
+    bytes32 private constant _EMPTY_SLOT_MARKER =
+        0x8159ea2f8547d3aac786e3dd5558567ed0f292248b867bfd642489b7ec86aea9;
+
+    // ---------------------------------------------------------------
     // State Variables
     // ---------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -95,7 +104,15 @@ abstract contract Inbox is EssentialContract, IInbox {
...
         proposal.derivationHash = _hashDerivation(derivation);
-        _setProposalHash(getConfig(), 0, _hashProposal(proposal));
+
+        // Set the genesis proposal at slot 0
+        _setProposalHash(config, 0, _hashProposal(proposal));
+
+        // Initialize remaining ring buffer slots with empty marker to distinguish them
+        // from data from previous fork's storage. This is a one-time cost during deployment.
+        for (uint256 i = 1; i < config.ringBufferSize; ++i) {
+            _proposalHashes[i] = _EMPTY_SLOT_MARKER;
```
dantaik, line 114: won't this transaction be super expensive, assuming the ring buffer size is 2.4K. Maybe we can avoid using the ringbuffer slots used by Pacaya, use a new slot instead.

ggonzalez94: it shouldn't for a reasonable buffer size(anything < 1000 slots), but to keep it safe and allow us to have a larger buffer if we decide to we decided not to reuse the slot. Will open a PR for this tomorrow

PR #19987, opened 2025-08-21
feat(taiko-client): read handover slots on chain

YoGhurt111:
ref: https://github.com/taikoxyz/taiko-mono/pull/19924
@cyberhorsey When you come back, let's decide when we will launch this feature, it need fork on client, means that we need to make sure all driver upgraded before we upgrade the contract.


packages/taiko-client/driver/driver.go
```diff
@@ -423,6 +425,13 @@ func (d *Driver) cacheLookaheadLoop() {
 			return err
 		}
 
+		routerConfig, err := d.rpc.GetPreconfRouterConfig(&bind.CallOpts{Context: d.ctx})
+		if err != nil {
+			log.Warn("Failed to fetch preconf router config, use defaultHandoverSkipSlots instead", "error", err)
```
cyberhorsey: fetching it every lookahead loop is fine but a lot of RPC calls - does it make sense to add an update function with an event and just subscribe to that for changes? I guess either one works tbh.

YoGhurt111: I think 1 call per 4s is ok?

cyberhorsey: sounds good to me, when I'm back I'll review properly and make sure that changing the handover slots mid epoch won't affect the lookahead ranges that are already calculated, if they will, we need to zero out the current ranges when the previous handover and the onchain handover differ and an update has been detected

cyberhorsey:
i pushed up a flag to determien if someone is a sequencer (or expects to be done), otherwise `cacheLookaheadLoop` doesnt need to be ran at all (it only determines *your* lookahead slots), and will reduce RPC calls on all non-sequencer nodes.

After this, all sequencer nodes will need to be updated with `true` with the `preconfirmation.isSequencer` flag.

cyberhorsey:
yep lets deploy the contract upgrade on Hekla + roll this out + upgrade it on Hekla, then we can get a release with this included for mainnet!

davidtaikocha (approved):

linoscope:
Created PR on top of this branch to only read handover at epoch transitions: https://github.com/taikoxyz/taiko-mono/pull/20053

cyberhorsey (approved):

RogerLamTd (approved):

PR #19994, opened 2025-08-22
fix(geth-rpc-gateway): remove premature WriteHeader from CORS middleware

sashass1315:
The CORS middleware was calling w.WriteHeader(http.StatusOK) before 
next.ServeHTTP(w, r), which caused the HTTP status to be fixed at 200 
and prevented the main handler from setting the correct response status.

This caused:
- "superfluous WriteHeader" errors when trying to set real status codes
- All HTTP responses returning 200 instead of actual proxy server status codes
- Violation of HTTP specification and middleware principles

Changes made:
- Removed premature w.WriteHeader(http.StatusOK) from CORS middleware
- Added proper preflight OPTIONS request handling with early return
- Improved CORS headers: fixed Allow-Methods and added Max-Age
- Maintained all existing proxy functionality


davidtaikocha:
Closed since `geth-rpc-gateway` is no longer used.

PR #19995, opened 2025-08-22
docs(protocol): improve Shasta design doc

dantaik:


YoGhurt111 (approved):

davidtaikocha (approved):

packages/protocol/docs/Derivation.md
```diff
@@ -352,19 +352,14 @@ The validated metadata serves three critical functions in block construction:
 
 Metadata encoding into L2 block header fields facilitates efficient peer validation and statistical analysis. The `withdrawalsRoot` field (32 bytes) is repurposed to store multiple metadata values through byte packing:
```
ggonzalez94: Is the `withdrawalsRoot` field still used and repurposed? From the table below I would imagine not?

ggonzalez94 (approved):
Left a small comment

packages/protocol/docs/Derivation.md
dantaik: No long used, it will be the same default value which is non zero I believe in Ethereum.

dantaik: Updated the doc.

PR #19996, opened 2025-08-22
fix(protocol): fix Shasta bug in reproving the same transition with different bond instructions

dantaik:
Try to fix a bug found by Nikesh

> As I understand it, the prover should be stateless, so it just validates that the transition is correct and ignores which other proofs have been submitted before.
> 
> In the standard `Inbox` case, the prover will accept a valid claim that has no bond instruction and it will also accept a valid claim (the same transition) that has a bond instruction sending funds from Alice to Bob. So if Alice proves on time, and then Bob submits the same proof after the proof window (but before finalization), the claim will now say that Alice owes Bob the liveness bond.
> 
> In the `InboxOptimized1` case, Alice can send a chain of proofs that start from some `_parentClaimHash` and have a span of 5. Bob can then send a proof with the same `_parentClaimHash` and a span of 1. This will overwrite the work that Alice has put in to prove the transitions 2-5.


ggonzalez94 (approved):
LGTM

PR #19999, opened 2025-08-25
fix(protocol): fix issue that forced inclusions were forced to be processed even if not due and make them more flexible

ggonzalez94:
This PR:
- fixes an issue in our design where the last forced inclusion was automatically processed even if it wasn't due and the proposer didn't have any say on it and it 
- addresses this [comment](https://github.com/taikoxyz/taiko-mono/pull/19882#discussion_r2283366907) to add the flexibility for a proposer to process multiple forced inclusions before they are due if they want to(they might be incentivized by the fee)
----
## Design
- adds a new parameter to the `ProposeInput` struct to decide how many forced inclusions the proposer wants to process
- simplifies the logic avoiding a low level call. If as a proposer you don't specify any forced inclusion(which is the most common case), there's only one SLOAD to verify none was due

## Not addressed on this PR:
These are some topics related to forced inclusions that are not addressed on this PR, but require discussion and are better addressed as separate issues:
1. Provability bond for forced inclusions - my opinion for this is we probably don't need it since we allow only 1 block per forced inclusion, so the "damage" is contained
2. Add a new delay to forced inclusions, so that if no one processes them proposing becomes permisionless(might be required by L2Beat for stage2, but requires some discussion and design)

---
## TODO
- [x] adjust the tests to the new input struct
    - I don't fully understand the test suite and it has mocks, so for now I removed some of these tests until we clean it up 
- [x] adjust some of the optimized inboxes to encode and decode this new field(if needed)


packages/protocol/contracts/layer1/shasta/impl/ForcedInclusionStore.sol
```diff
@@ -94,45 +94,62 @@ contract ForcedInclusionStore2 is EssentialContract, IForcedInclusionStore {
...
-        lastProcessedAt = uint64(block.timestamp);
+        inclusions_ = new ForcedInclusion[](toProcess);
+        uint256 totalFees;
 
         unchecked {
-            delete queue[head++]; // delete element at head AND THEN increment head
-            _feeRecipient.sendEtherAndVerify(inclusion_.feeInGwei * 1 gwei);
+            for (uint256 i; i < toProcess; ++i) {
+                ForcedInclusion storage inclusion = queue[head + i];
+                inclusions_[i] = inclusion;
```
dantaik: how about `inclusions_[i] = queue[head + i];` and we use `inclusions_[i] ` as it is already in memory

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -149,19 +146,21 @@ abstract contract Inbox is EssentialContract, IInbox {
...
-        if (!forcedInclusionProcessed || availableCapacity > 1) {
-            // Propose the normal proposal after the potential forced inclusion to match the
-            // behavior in Shasta fork.
-            LibBlobs.BlobSlice memory blobSlice =
-                LibBlobs.validateBlobReference(input.blobReference, _getBlobHash);
-            _propose(config, coreState, blobSlice, false);
+        coreState = _processForcedInclusions(config, coreState, input.numForcedInclusions);
+
+        // Verify that no forced inclusion that is due remains in the queue.
+        if (IForcedInclusionStore(config.forcedInclusionStore).isOldestForcedInclusionDue()) {
```
dantaik: use require-statement for consistency?

dantaik: There is a potential issue here, an attacker can create many forced inclusion with the same timestamp, so they will due all together, even if the current proposer have consumed many inclusions, the next one is still due. 

Therefore, I think we can change the if-statement to

```
require(input.numForcedInclusions  > 1 || 
!IForcedInclusionStore(config.forcedInclusionStore).isOldestForcedInclusionDue(), 
UnprocessedForcedInclusionIsDue());
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -149,19 +146,21 @@ abstract contract Inbox is EssentialContract, IInbox {
...
+            revert UnprocessedForcedInclusionIsDue();
         }
+
+        // Propose the normal proposal after the potential forced inclusion to match the
+        // behavior in Pacaya.
```
dantaik: imagine there is only one slot in the ring buffer, and there is a forced inclusion due, then we should still allow the next forced inclusion to be processed. Therefore, we need to change 

```
 require(availableCapacity > input.numForcedInclusions, ExceedsUnfinalizedProposalCapacity());
 ```

to 
```
 require(availableCapacity >= input.numForcedInclusions, ExceedsUnfinalizedProposalCapacity());
```

and wrap the regular proposal in a if-statement:


`if (availableCapacity > input.numForcedInclusions) {...}`

File: packages/protocol/contracts/layer1/shasta/impl/Inbox.sol [hunk 4]
```diff
@@ -563,34 +562,29 @@ abstract contract Inbox is EssentialContract, IInbox {
...
+    /// @notice Consumes up to _numForcedInclusions from the queue and proposes them sequentially
     /// @param _config Configuration containing forced inclusion store address
-    /// @param _coreState Current core state to update if inclusion processed
-    /// @return coreState_ Updated state if forced inclusion processed, unchanged otherwise
-    /// @return forcedInclusionProcessed_ True if a forced inclusion was successfully processed
-    function _processForcedInclusion(
+    /// @param _coreState Current core state to update with each inclusion processed
+    /// @param _numForcedInclusions Maximum number of forced inclusions to process
+    /// @return Updated core state after processing all consumed forced inclusions
+    function _processForcedInclusions(
```
dantaik, line 584: we probably want to return the actual number of forced inclusion processed, then use this number to check if there are still room for the regular proposal.

dantaik:
Left a few comments

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -112,17 +112,14 @@ abstract contract Inbox is EssentialContract, IInbox {
...
-    ///      5. Updates core state and emits Proposed event
-    /// @dev Forced inclusion processing:
-    ///      - Processes exactly one (oldest) forced inclusion per call
-    ///      - Only processed when due, cannot be processed early
-    ///      - Takes priority over regular proposals
+    ///      2. Finalizes eligible proposals up to `config.maxFinalizationCount` to free ring buffer
+    ///         space.
+    ///      3. Process `input.numForcedInclusions` forced inclusions. The proposer is forced to
+    ///         process at least one if it is due.
+    ///      5. Updates core state and emits `Proposed` event
```
linoscope: Nit:
```suggestion
    ///      4. Updates core state and emits `Proposed` event
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
linoscope: Think it should be:
```
require(input.numForcedInclusions  >= 1 || 
!IForcedInclusionStore(config.forcedInclusionStore).isOldestForcedInclusionDue(), 
UnprocessedForcedInclusionIsDue());
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -149,19 +146,21 @@ abstract contract Inbox is EssentialContract, IInbox {
...
+        // Verify that no forced inclusion that is due remains in the queue.
+        if (IForcedInclusionStore(config.forcedInclusionStore).isOldestForcedInclusionDue()) {
+            revert UnprocessedForcedInclusionIsDue();
         }
+
+        // Propose the normal proposal after the potential forced inclusion to match the
+        // behavior in Pacaya.
+        LibBlobs.BlobSlice memory blobSlice =
+            LibBlobs.validateBlobReference(input.blobReference, _getBlobHash);
+        _propose(config, coreState, blobSlice, false);
```
linoscope: Think we should NOT allow proposers to propose if there are overdue force-inclusion blocks in the queue. Otherwise, if 100 such blocks are queued, it would take 100 proposals to clear them. With preconfs, the gap between proposals can be large, so processing 100 proposals could take an extremely long time. And with more blocks being force-included, the delay can keep increasing.

We can do this by either replacing `isOldestForcedInclusionDue` with `getOverdueForcedInclusionCount` and ensuring that `input.numForcedInclusions >= getOverdueForcedInclusionCount` for `_propose` to be called, or just adding this kind of if statement here:
```
        if (availableCapacity - input.numForcedInclusions > 0 
            && !IForcedInclusionStore(config.forcedInclusionStore).isOldestForcedInclusionDue()) {
            LibBlobs.BlobSlice memory blobSlice =
                LibBlobs.validateBlobReference(input.blobReference, _getBlobHash);
            _propose(config, coreState, blobSlice, false);
        }
```

ggonzalez94:
@dantaik @linoscope thanks for the review. I applied most suggestions - a few decisions that maybe are worth highlighting so we can discuss:

> There is a potential issue here, an attacker can create many forced inclusion with the same timestamp, so they will due all together, even if the current proposer have consumed many inclusions, the next one is still due.

We are already limiting to storing one forced inclusion per tx(so the cost of this attack is arguably large). But even then I agree we should defend against it. Instead of just forcing the proposer to process the latest one due, what I did is have a new config parameter that for now it is set to 1, but I think it should be higher. In a true censorship scenario, you want to enforce forced txs to be processed by more than one at a time(otherwise a user that wants to exit to the L1 for example might have to wait for a very long period of time if the queue is too large, which the preconfer can artificially inflate by posting himself a bunch of forced inclusions).
This seems a healthy middleground between you have to process all forced inclusions that are due, and you have to process only one.

dantaik (approved):

PR #20019, opened 2025-08-27
fix(bridge-ui): correct ENS address resolution error message

IrvinGa38:
Replaced the misleading error message `'No ENS name'` with `'No ENS address'` when ENS address resolution fails. 
Corrected the indefinite article from `'a'` to `'an'` in the usage description for the Raiko API key path.


dantaik (approved):

YoGhurt111 (approved):

KorbinianK:
@dantaik It's not bridge UI. We should not have PRSs across packages as it breaks releases. It's also not a misleading error. 

I'd close it tbh.

PR #20023, opened 2025-08-27
feat(protocol): move forced inclusion code into a library to reduce inbox code size

### LINKED ISSUE #20017: Merge ForcedInclusionStore with the Inbox
Right now the ForcedInclusionStore is implemented as a separate contract that the inbox calls. Conceptually this is nice, since you can think of them separately. But it introduces a couple awkward points like the store having a function to unconditionally consume inclusions that can be only called by the inbox and the fact that there's a circular dependency.
A possible better aproach for this is to make the `ForcedInclusionStore` an abstract contract with all the storage and logic and make the inbox inherits from it. This way we can still reason about them separately, but avoid the awkward points.

Only concern is contract size for the inbox growing and getting closer to the limit

ggonzalez94:
Fixes #20017

I also had to do some changes to the test structure to avoid inheritance issues. Now it uses a deployment factory which is cleaner


ggonzalez94:
@cursor please review

ggonzalez94:
@claude please review

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -22,7 +23,7 @@ import { LibBonds } from "src/shared/based/libs/LibBonds.sol";
 ///      - Bond instruction processing for economic security
 ///      - Finalization of proven proposals
 /// @custom:security-contact security@taiko.xyz
-abstract contract Inbox is EssentialContract, IInbox {
+abstract contract Inbox is ShastaForcedInclusionStore, IInbox {
```
dantaik: will this break storage layout compatibility with the current deployment?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -74,7 +75,12 @@ abstract contract Inbox is EssentialContract, IInbox {
     // ---------------------------------------------------------------
 
     /// @notice Initializes the Inbox contract
-    constructor() EssentialContract() { }
+    constructor(
+        uint64 _inclusionDelay,
```
```suggestion
        uint64 _forceInclusionInclusionDelay,
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -74,7 +75,12 @@ abstract contract Inbox is EssentialContract, IInbox {
...
+        uint64 _feeInGwei
```
```suggestion
        uint64 _forceInclusionFeeInGwei
```

packages/protocol/contracts/layer1/shasta/impl/ShastaForcedInclusionStore.sol
```diff
@@ -7,23 +7,22 @@ import { LibAddress } from "src/shared/libs/LibAddress.sol";
...
+/// allow users to pay a fee to ensure their transactions are included in a block. The contract
+/// maintains a FIFO queue of inclusion requests.
 /// @dev Inclusion delay is measured in seconds, since we don't have an easy way to get batch number
 /// in the Shasta design.
 /// @dev We only allow one forced inclusion per L1 transaction to avoid spamming the proposer.
 /// @dev Forced inclusions are limited to 1 blob only, and one L2 block only(this and other protocol
 /// constrains are enforced by the node and verified by the prover)
 /// @custom:security-contact security@taiko.xyz
-contract ForcedInclusionStore2 is EssentialContract, IForcedInclusionStore {
+abstract contract ShastaForcedInclusionStore is EssentialContract, IForcedInclusionStore {
```
dantaik: To minimise the Inbox code size, we can use a library to capture the logics then in Inbox.sol, we use the library's functions.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
ggonzalez94: good catch, ty. Here's a proposed [fix](https://github.com/taikoxyz/taiko-mono/pull/20028). If you are onboard with this approach I can polish it as suggested in the PR description

packages/protocol/contracts/layer1/shasta/impl/ShastaForcedInclusionStore.sol
ggonzalez94: created #20030 to track this

dantaik:
@cursor please review

dantaik:
@claude please review

dantaik (approved):

PR #20028, opened 2025-08-28
fix(protocol): fix layout by introducing a new version of the EssentialContract

ggonzalez94:
fixes the storage layout issue when unifying the forced inclusion store and the inbox


packages/protocol/contracts/layer1/shasta/base/ShastaEssentialContract.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import "src/shared/common/BaseEssentialContract.sol";
+
+/// @title ShastaEssentialContract
```
dantaik, line 6: Maybe we can define a ForcedInclusionState object and create a library to handle it. Then in Inbox.sol, we use the library without inheriting from ShastaForcedInclusionStore... this will avoid all the these mess...

ggonzalez94: I'll try this tomorrow and we can decide which version we like the most! In general I prefer inheritance vs libraries that touch external storage(less indirection), but in this case it might be worth anyways because of the inbox size

ggonzalez94:
closing in favor of #20036

PR #20029, opened 2025-08-28
chore(repo): use updated action, fix token usage

RogerLamTd:
updated to use the new release, also noted that the `CLAUDE_CODE_MAX_OUTPUT_TOKENS` isn't actually used in the action so removed it


ggonzalez94 (approved):

RogerLamTd:
@ggonzalez94 it looks like the v1 action doesn't have built in commenting and will require additional github app configuration, we can revisit the upgrade later

PR #20033, opened 2025-08-29
feat(protocol): implement reinitializer for shasta inbox

### LINKED ISSUE #20009: Implement reinitializer in new Inbox contracts
The new inbox contracts have an initializer function that does some basic setup. This is needed and we wanna maintain it, but on mainnet this initializer **won't be called** since we'll be upgrading the current inbox and the proxy is already deployed.
```sol
function init(address _owner, bytes32 _genesisBlockHash) external initializer {
        __Essential_init(_owner);

        Transition memory transition;
        transition.endBlockMiniHeader.hash = _genesisBlockHash;

        CoreState memory coreState;
        coreState.nextProposalId = 1;
        coreState.lastFinalizedTransitionHash = _hashTransition(transition);

        Proposal memory proposal;
        proposal.coreStateHash = _hashCoreState(coreState);

        Derivation memory derivation;
        proposal.derivationHash = _hashDerivation(derivation);

        _setProposalHash(getConfig(), 0, _hashProposal(proposal));
        emit Proposed(
            encodeProposedEventData(
                ProposedEventPayload({
                    proposal: proposal,
                    derivation: derivation,
                    coreState: coreState
                })
            )
        );
    }
```

What we should do is have a [reinitializer](https://docs.openzeppelin.com/contracts/5.x/api/proxy#Initializable-reinitializer-uint64-) with the logic, but we wanna keep the initializer as well for testnets and local development where the proxy does not exsits. We should move the init logic into a private function and call it from both

ggonzalez94:
Fixes #20009


packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -81,30 +81,16 @@ abstract contract Inbox is EssentialContract, IInbox {
...
-                    derivation: derivation,
-                    coreState: coreState
-                })
-            )
-        );
+    /// @notice Reinitializes the Inbox contract for upgrades
+    /// @dev Used when upgrading from Pacaya to Shasta on mainnet where proxy already exists
+    /// @param _owner The owner of this contract
+    /// @param _genesisBlockHash The hash of the genesis block
+    function initV2(address _owner, bytes32 _genesisBlockHash) external reinitializer(2) {
```
dantaik, line 92: this method should only be callable by the owner, right?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -81,30 +81,16 @@ abstract contract Inbox is EssentialContract, IInbox {
...
+        __Essential_init(_owner);
```
dantaik: maybe we don't want to call `__Essential_init`?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -81,30 +81,16 @@ abstract contract Inbox is EssentialContract, IInbox {
...
+        _initializeInbox(_genesisBlockHash);
```
dantaik, line 99: shall we have a dedicated state variable to indicate if `_initializeInbox` has ever been called, if so, we shall not call `_initializeInbox` but shall increase the initialisation counter.

ggonzalez94: actually a better approach might be to get rid of init entirely and just have initV2. I think it will [work](https://docs.openzeppelin.com/contracts/5.x/api/proxy#Initializable-reinitializer-uint64-) for both new deployments and existing ones. We'll just have to slightly change the way we deploy to use `upgradeAndCall` instead of passing the data to the proxy directly.
I'll read this more carefully during the weekend

dantaik: I think it will work.

PR #20038, opened 2025-08-29
feat(protocol): delete onlyStandaloneTx

dantaik:
I feel like this onlyStandaloneTx guard is no longer necessry. @ggonzalez94 @YoGhurt111


ggonzalez94 (approved):
I remember seeing this was a recommendation of one of the previous audits, but given forced inclusions have a fee, and they will be limited in terms of gas they can use then I agree we can remove it

PR #20042, opened 2025-08-30
chore(protocol): a way to fix Inbox's init function

dantaik:


dantaik:
@claude please review

ggonzalez94 (approved):
I think this will work. I only made a few tweaks:
- only call the essential init when the contract has no owner(to avoid transferring ownership and other calls on upgrades)
- changed the name to `initV2` and used `2` for the reinitializer to better reflect this is the second version

PR #20046, opened 2025-09-01
feat(protocol): remove multi-hop support and catching from SignalService

dantaik:
The question I have is: how to actually perform the upgrade, will it cause some signals not verifiable?

EDIT

@ggonzalez94 maybe we should not make this change, instead, we may have to keep all existing logics, then add extra code to use Checkpoint manager. Once we launch Shasta, then we "clean up" Signal service implementation to remove unused code.


dantaik:
@claude please review

dantaik:
@ggonzalez94 maybe we should not make this change, instead, we may have to keep all existing logics, then add extra code to use Checkpoint manager. Once we launch Shasta, then we "clean up" Signal service implementation to remove unused code.

PR #20048, opened 2025-09-01
feat(repo): introduce an event indexer for `Shasta` protocol

davidtaikocha:
Ref: https://rindexer.xyz/

Some rust code is auto-generated by [rindexer](https://rindexer.xyz/) cli. Basically, this indexer will keep indexing the `Proved`, `Proposed`, `BondInstructed` events with the awareness of chain reorg, and will expose a GraphQL endpoint for other services to query.

Todo:
- [x] decode the raw event data in `handler`


davidtaikocha:
@claude plz review

packages/shasta-indexer/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    let args = args.iter();
+    if args.len() == 0 {
+        enable_graphql = true;
+        enable_indexer = true;
+    }
+
+    for arg in args {
+        match arg.as_str() {
+            "--graphql" => enable_graphql = true,
```
cyberhorsey: should we use a config struct that allows .env to be parsed as well, like the ejector service?

cyberhorsey: https://github.com/taikoxyz/taiko-mono/blob/main/packages/ejector/crates/config/src/lib.rs

davidtaikocha: yep, rn this `main.rs` is auto-generated by `rindexer` cli, and looks like it only uses a `yaml` file as the configuration. When we need more flags to configure later on, we can change to use it.

PR #20049, opened 2025-09-01
fix(bridge-ui): type mismatch in Promise.race result

brawncode:
corrected the type returned by `Promise.race`. previously it was typed as `Record<number, RelayerBlockInfo>`, but since `Promise.race` resolves with a single promise, the correct type is `RelayerBlockInfo`.

removed unnecessary indexing by `srcChainId`, returning the resolved block info directly instead of trying to access it as an object. this ensures the result matches the actual structure returned by the promise.


KorbinianK (approved):

bearni95 (approved):

myronrotter (approved):

KorbinianK:
Thanks for the contribution, but this change is incorrect. `getBlockInfo() `actually returns `Promise<Record<number, RelayerBlockInfo>> `- a dictionary indexed by chain ID, not a single `RelayerBlockInfo`.

The original code correctly:

Receives a `Record<number, RelayerBlockInfo>` from the promise
Extracts the specific chain's info using `[srcChainId]`

Your change would cause a type error because you're trying to assign a Record to a `RelayerBlockInfo`. 

Closing this PR.

PR #20050, opened 2025-09-01
feat(protocol): add `lookaheadSlotTimestamp` into `Proposal`

dantaik:
Per Lin's request, I'm adding one more field called `lookaheadSlotTimestamp` into struct `Proposal` and the Anchor transaction.

@linoscope @AnshuJalan please review the ShastaAnchor.sol file to make sure if it is OK to only save the latest `lookaheadSlotTimestamp` value in state, or you expect the whole history to be persisted in state?


packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -719,6 +729,7 @@ abstract contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
                 id: _coreState.nextProposalId++,
                 proposer: msg.sender,
                 timestamp: uint48(block.timestamp),
+                lookaheadSlotTimestamp: _lookaheadSlotTimestamp,
                 coreStateHash: _hashCoreState(_coreState),
```
AnshuJalan: Minor nit: `proposer`s position should change

```solidity
timestamp: uint48(block.timestamp),
lookaheadSlotTimestamp: _lookaheadSlotTimestamp,
proposer: msg.sender,
coreStateHash: _hashCoreState(_coreState)
```

AnshuJalan (approved):

PR #20062, opened 2025-09-02
fix(protocol): add cooldown window for transitions

### LINKED ISSUE #20060: Reconsider compatibility between preconfs and `bondInstructions`
## Problem
Currently `ShastaAnchor` takes ([code](https://github.com/taikoxyz/taiko-mono/blob/4996b30a0491142c546ce752213abba3c021804f/packages/protocol/contracts/layer2/based/ShastaAnchor.sol#L129)) takes bond instructions (`BondInstruction[]`) as a parameter, which represent the bond credit instructions to process for this L2 block.

While not specified concretely, it seems like the bond instructions will depend on the L2 blocks finalized at L2 block proposal time. If this is the case, this means you can’t know which bondInstructions you must pass to anchor at preconf time . That uncertainty makes it impossible to preconfirm the exact L2 state.

## Task

- Confirm that this is an issue.
- Think of mitigations

dantaik:
Fixes https://github.com/taikoxyz/taiko-mono/issues/20060

Adds the new `effectiveAt` field to the transition record, removing the assumption that the preconfer must be the L1 validator. This change enables us to launch Shasta on mainnet with whitelist preconfirmation.

Note that `cooldownWindow` can be 0


dantaik:
@claude please review

ggonzalez94:
@dantaik should I still review given we determined #20060 is not an issue? I guess this might still be relevant to avoid race conditions between proposer and prover

dantaik:
> @dantaik should I still review given we determined #20060 is not an issue? I guess this might still be relevant to avoid race conditions between proposer and prover

I think this change still have a value: it allows us to launch Shasta before permissionless preconf is live -- preconfers in the whitelist are not L1 validators -- they cannot control if a `prove` function is included before his `propose` function. If this happens, the new transition just proved may enable more proposals to be finalized by the next proposal.

davidtaikocha (approved):

ggonzalez94:
> > @dantaik should I still review given we determined #20060 is not an issue? I guess this might still be relevant to avoid race conditions between proposer and prover
> 
> I think this change still have a value: it allows us to launch Shasta before permissionless preconf is live -- preconfers in the whitelist are not L1 validators -- they cannot control if a `prove` function is included before his `propose` function. If this happens, the new transition just proved may enable more proposals to be finalized by the next proposal.

Ok, I'll check the lookahead and slashing contracts since I already went trough those docs and then come back to this!

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -828,6 +829,11 @@ abstract contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
         bytes32 transitionRecordHash = _hashTransitionRecord(_transitionRecord);
         require(transitionRecordHash == storedHash, TransitionRecordHashMismatchWithStorage());
 
+        // Check if cooldown period has passed
```
ggonzalez94: I think this will not prevent the issue with the race condition(proposer builds their input and a prover manages to get their tx included before), because of the check in line 823.

Imagine you call `_finalize` with 2 transitions, but there are three consecutive ones in storage(for the third one the cooldown window has not passed).
```
 Scenario Setup

  - 3 proposals have stored transition records
  - Proposal 1: Ready to finalize (past cooldown)
  - Proposal 2: Ready to finalize (past cooldown)
  - Proposal 3: Still cooling down (before cooldown)
  - User provides only 2 transition records for
  proposals 1 and 2

  Execution Flow

  1. Iteration 0 (Proposal 1)
    - _hasTransitionRecord = true (record provided
  at index 0)
    - Line 823: Stored hash exists ✓
    - Line 826: Record was provided ✓
    - Line 833: Past cooldown ✓
    - Result: Successfully finalized
  2. Iteration 1 (Proposal 2)
    - _hasTransitionRecord = true (record provided
  at index 1)
    - Line 823: Stored hash exists ✓
    - Line 826: Record was provided ✓
    - Line 833: Past cooldown ✓
    - Result: Successfully finalized
  3. Iteration 2 (Proposal 3)
    - _hasTransitionRecord = false (no record at
  index 2)
    - Line 823: Stored hash exists ✓
    - Line 826: require(_hasTransitionRecord) ❌
    - Result: Reverts with
  TransitionRecordNotProvided()

```

.gitignore
```diff
@@ -127,10 +127,4 @@ packages/taiko-client/opnode_peerstore_db
 packages/taiko-client/opnode_p2p_priv.txt
 
 #Claude
-.claude/settings.local.json
-
-# Ejector
```
ggonzalez94: Is there a reason we are removing these? I think they were added in a client PR. 
cc @cyberhorsey

ggonzalez94 (commented):
I think there's an issue that prevents this PR from effectively solving the race condition.

Moreover, I think we should ALLOW the proposer to finalize transitions before their cooldown, but ONLY FORCE them if there are transitions ready to be finalized that have passed the cooldown. I'll create a small PR that fixes the issue I explained and also tries to implement this logic.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: I see the point now.

dantaik:
@claude please review

ggonzalez94:
@dantaik I think we both overlooked something important that will cause this to not work properly. In `_finalize` when `_hasTransitionRecord` is false the `_transitionRecord` being passed is from the last iteration. So we are not actually checking if the next one is still on its cooling window, which makes the check ineffective.
Cursor explained it well [here](https://github.com/taikoxyz/taiko-mono/pull/20082#discussion_r2323756918).

Since we are only storing the hash of the transition, and by definition we don't want the proposer to send the transition for this record(the whole purpose of this is to solve the race condition), this might be a bit tricky to solve. Let me think about this, and we can hold on merging this for now

ggonzalez94:
I thought about this a bit more, and unfortunately I don't think there's an easy answer/solution to the race problem because of what I described above ^^

I see two possible approaches:
1. We save `effectiveAt` in storage. This way it can be read when finalizing and we can check if a transitionRecord has passed its cooldown period or not easily.
Pros:
- easy to implement
- solves the race problem entirely

Cons:
- Proving becomes more costly. We need an additional `SSTORE` when proving + one additional `SLOAD` when finalizing(proposing)

2. We do nothing and close this PR. Instead the way the proposer is protected from this race condition(the amount of TransitionRecords may change between when he builds this tx and when it is executed) is **using revert protection**.


I don't think neither of these solutions is ideal, but I would lean more towards solution number 2, as it keeps the protocol cost lower, and I imagine it won't be a problem in most of the cases.

dantaik:
@ggonzalez94 I proposed this approach to fixed the issue, https://github.com/taikoxyz/taiko-mono/pull/20091

> I thought about this a bit more, and unfortunately I don't think there's an easy answer/solution to the race problem because of what I described above ^^
> 
> I see two possible approaches:
> 
> 1. We save `effectiveAt` in storage. This way it can be read when finalizing and we can check if a transitionRecord has passed its cooldown period or not easily.
>    Pros:
> 
> * easy to implement
> * solves the race problem entirely
> 
> Cons:
> 
> * Proving becomes more costly. We need an additional `SSTORE` when proving + one additional `SLOAD` when finalizing(proposing)
> 
> 2. We do nothing and close this PR. Instead the way the proposer is protected from this race condition(the amount of TransitionRecords may change between when he builds this tx and when it is executed) is **using revert protection**.
> 
> I don't think neither of these solutions is ideal, but I would lean more towards solution number 2, as it keeps the protocol cost lower, and I imagine it won't be a problem in most of the cases.

@ggonzalez94 I proposed this approach to fixed the issue, https://github.com/taikoxyz/taiko-mono/pull/20091

dantaik:
replaced by https://github.com/taikoxyz/taiko-mono/pull/20091

PR #20063, opened 2025-09-02
feat(protocol): add ANCHOR_MIN_OFFSET to block derivation process

### LINKED ISSUE #20059: Introduce a `MIN_ANCHOR_OFFSET` in post-shasta derivation
### Problem

* Fair preconfirmers risk having their preconfirmations invalidated and slashed if the L1 anchor block undergoes a reorg.
* When the anchor block reorgs, the arguments passed into the anchor transaction within a preconfirmed block are no longer valid, causing the system to deem the preconfirmation slashable.
* Currently, the preconfirmation commitment includes an `anchorHash` field, which can be used to revert slashing if it does not match the eventual anchor block hash. However, this approach is not robust and introduces additional complexity in both the sidecar and the overseer.
* While a preconfirmer can protect themselves by anchoring to an older block, they remain exposed if the previous preconfirmer used a very recent anchor (e.g., only 1 slot old). In such cases, the current preconfirmer may be forced to anchor with less than a 2-slot delay, leaving them vulnerable to reorgs.

### Suggestion

* Introduce a `MIN_ANCHOR_OFFSET` parameter in the derivation process. `MIN_ANCHOR_OFFSET` represents the minimum required offset of the anchor block from the `L1Origin`.
* If `MIN_ANCHOR_OFFSET` is sufficiently large (theoretically 2 is enough, though we recommend a higher value), it would ensure that preconfirmers use an anchor block old enough to mitigate reorg risk.
* This recommendation is made considering that Ethereum has not experienced a reorg greater than 1 block at the execution layer post-Merge.

dantaik:
Addressing concerns from @AnshuJalan https://github.com/taikoxyz/taiko-mono/issues/20059


dantaik:
@claude please review

davidtaikocha (approved):

YoGhurt111 (approved):

PR #20064, opened 2025-09-02
chore(protocol): simplify transition logic by merging InboxOptimized1 into Inbox

ggonzalez94:
This PR aims to substantially simplify how transitions are handled in the protocol, both in storage and their aggregation logic.

1. Merge `InboxOptimized1` into `Inbox`. This avoids some of the indirection of thinking how transitions are stored and aggregated in each use case. This removes the need for some of the functions to be virtual and allows to use a single storage variable(ring buffer with branching capacity  - - check `TransitionRecordSlot`).
2. Since the inbox got more convoluted move the aggregation logic to a new library `LibTransitionAggregation`. This made the Inbox even simpler than the original version and allows us to make this library function external if we ever wanted to in order to reduce code size(probably not a good tradeoff at the moment).
3. Remove all unnecessary hashing functions `_hashTransition`, `_hashTransitionRecord`, etc. and use in place hashing instead


dantaik:
A few thoughts:

- The layered approach (core contract + sub-contracts) helps keep tests modular and makes code reviews easier. Core functions can be tested in a minimal setup, while sub-contracts add their own logic and get separate test suites. This separation reduces unnecessary setup and makes coverage easier to manage.

- I’m not convinced merging sub-contracts back into the parent would reduce code size. I do like the idea of introducing two extra libraries to simplify sub-contracts. The trade-off is that switching some functions from internal to public may shrink bytecode but adds extra gas per call, so we’d need to evaluate whether the savings are worth it.

ggonzalez94:
> The layered approach (core contract + sub-contracts) helps keep tests modular and makes code reviews easier. Core functions can be tested in a minimal setup, while sub-contracts add their own logic and get separate test suites. This separation reduces unnecessary setup and makes coverage easier to manage.

Let me try to convince you why I still think this approach is more readable :)
1. It reduced over 200LOC more than it is adding, so there's less to maintain
2. Allows us to have one single storage for transitions. No more need to keep two separate mental models as to where transitions are stored. I think we should do this regardless if we merge the two contracts or not
3. While the simpler inbox contract was a bit easier to reason about, I think this design embraces complexity better. Instead of having a separate implementation that treats transitions differently, we now have a single source of truth. The reality is that we will always use the aggregation version, so some of the code of the inbox would always go unused
4. Both functions require the same inputs and outputs as far as I recall, so testing should remain same as before. If we need to test internal functions we can always expose those in the concrete implementation for tests and test them independently

I guess my TL;DR: is I think we had code in the original inbox that would always go unused(our deployment would always use the optimized version). We are better off embracing that complexity, and managing it appropriately than treating it as an extension(since only one version will get deployed). 
Maybe we can have a sync later today to discuss the best design?

 > I do like the idea of introducing two extra libraries to simplify sub-contracts. The trade-off is that switching some functions from internal to public may shrink bytecode but adds extra gas per call, so we’d need to evaluate whether the savings are worth it.
 
 Agree the tradeoff is not worth it. I was just highlighting it gives us the flexibility if we ever need that in the future.

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -86,6 +86,20 @@ interface IInbox {
         bytes32 checkpointHash;
     }
 
+    /// @notice Optimized storage for transition records with efficient ring buffer and branching
+    /// @dev Stores transition records with optimized access patterns for the common case
+    struct TransitionRecordSlot {
```
dantaik: It seems this is an implementation specific struct, shall we move it to Inbox.sol?

dantaik (approved):

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -486,28 +428,52 @@ abstract contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+            // Different proposal ID, so we can use the primary slot
+            slot.primaryRecordHash = transitionRecordHash;
+            slot.proposalId = _proposalId;
+            slot.partialParentHash = bytes26(_transition.parentTransitionHash);
+        } else if (
+            _isPartialParentTransitionHashMatch(
+                slot.partialParentHash, _transition.parentTransitionHash
+            )
+        ) {
+            // TODO: verify I'm not missing anything here
```
ggonzalez94, line 484: @dantaik can you double check me on this? before we were just overwriting with the same value, which should be equivalent to just returning

ggonzalez94:
we decided to split this PR into two:
1. libraries for aggregate transitions logic and bond calculation. (we are sure we want this and easier to merge)
2. merge the two inbox implementations and the storage (this still needs some discussion to see if and how we do it)


Closing this one.

PR #20066, opened 2025-09-02
fix(taiko-client): revert pr-20035

cyberhorsey:


RogerLamTd (approved):

ggonzalez94 (approved):
Approving so we can restore testnet stability and we can figure out how to remove the undocumented geth dependencies later and test again

PR #20070, opened 2025-09-03
chore(protocol): improve Inbox initializer

dantaik:


packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -93,7 +93,7 @@ abstract contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     /// upgrade happens. On upgrades this is usually done calling `upgradeToAndCall`
     /// @param _owner The owner of this contract
     /// @param _genesisBlockHash The hash of the genesis block
-    function initV2(address _owner, bytes32 _genesisBlockHash) external reinitializer(2) {
+    function init(address _owner, bytes32 _genesisBlockHash) external reinitializer(3) {
```
ggonzalez94: two quick questions:
1. why reinitializer(3)? since it is the second time we are running it(the first one being the original initalizer)
2. I believe it is a standard practice to name the second initializer something like initV2 to highlight that the contract has been initialized before. Although I don't have a strong opinion on this and I guess the reinitializer modifier itself marks that already

dantaik: 1. I think for the current mainnet, we have called `initializer` and `reinitializer(2)` (I may be wrong about the reinitializer though), so calling `reinitializer(3)` works for both a new deployment and a upgrade for an existing deployment.

dantaik: 2 I don't have a strong option either as long as all naming are consistent.  Maybe we can apply this rule? if `reinitializer(x)` is used, call the function `initVx`?

davidtaikocha (approved):

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
ggonzalez94: you're right, I checked the events and the proxy has been initialized twice. Good memory 😄 
image

I updated it to use initV3, I like the convention

ggonzalez94 (approved):

ggonzalez94 (approved):

PR #20074, opened 2025-09-03
fix(taiko-client): fix preconf router config

cyberhorsey:
also we should pass `--taikoWrapper` to the driver, otherwise a section of API code doesnt work, and we also cant fetch the preconf router config.

Only sequencer nodes need the wrapper.

### DIFF
--- a/.github/workflows/taiko-client--docker.yml
+++ b/.github/workflows/taiko-client--docker.yml
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main, fix_preconf_router_config]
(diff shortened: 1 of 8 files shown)


.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, fix_preconf_router_config]
```
```suggestion
    branches: [main]
```

RogerLamTd (approved):

davidtaikocha (approved):

YoGhurt111 (approved):

PR #20076, opened 2025-09-04
feat(taiko-client): granular metrics for rpc calls

cyberhorsey:
introduces granular metrics for every one of our RPC calls, to be able to find delays/latency issues


cyberhorsey:
@claude review again, g-unit

packages/taiko-client/pkg/rpc/engine.go
```diff
@@ -163,7 +193,11 @@ func (c *EngineClient) SetL1OriginSignature(
 func (c *EngineClient) SetHeadL1Origin(ctx context.Context, blockID *big.Int) (*big.Int, error) {
 	var res *big.Int
 
-	if err := c.CallContext(ctx, &res, "taikoAuth_setHeadL1Origin", blockID); err != nil {
+	methodName := GetRPCMethodName(c.metrics.GetClientType(), "SetHeadL1Origin")
+	err := c.metrics.TrackRequest(ctx, methodName, func() error {
+		return c.CallContext(ctx, &res, methodName, blockID)
```
davidtaikocha, line 198: I see we modified lots of RPC calls in client, since basically all the RPCs will eventually go into this `c.CallContext` method (in `ethclient.go`'s `EthClient`), will it also be possible that we only extend the `*rpc.Client` in `EthClient` (ethclient.go) with a monkey-patched `CallContext` method with the metrics feature (and we can also record the rpc method name in `CallContext`)? Which will make this feature more "black-boxed" not only for the current changes but also for future development, however, I’m not entirely sure it will fully meet the feature requirements... (if its only for tracking duration and latency, i guess probly it can?)

cyberhorsey: thats just for the engine.go calls really, the rest of the them dont use `CallContext`, so we could refactor it just for those but leave the others elsewhere.

davidtaikocha: emm not really, internally all RPCs go into that `CallContext `.

cyberhorsey: we cant really monkey-patch in go, not like you can in python or something. Are you proposing to change the code in `taiko-geth` in the `ethclient.go` package itself? because we shouldnt be modifying dependency files otherwise.

it makes more sense to make a custom `RoundTripper` to pass into the `rpc.Client` and instrument that I think?

davidtaikocha: Or we can just change the `EthClient` in client's `ethclient.go`?

YoGhurt111 (approved):

packages/taiko-client/pkg/rpc/engine.go
cyberhorsey: I still don't know what you mean, you mean in taiko-geth itself? How else do we change the actual CallContext method? I feel like an idiot not understanding what you're saying here lmao BC monkey patching isn't a thing here

PR #20082, opened 2025-09-04
fix(protocol): prevent race condition and allow finalization before cooldown

ggonzalez94:
See context [here](https://github.com/taikoxyz/taiko-mono/pull/20062#pullrequestreview-3187525175)


dantaik (commented):
Shall we rename it to something else? 

- `exemptedUntil` - this transition is exempted from finalization until a certain timestamp
- `finalizationExemptedUntil`
- `finalizationDeadline`
- `finalizationTimestamp`

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -822,18 +824,22 @@ abstract contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+        // If transition record is provided, allow finalization regardless of cooldown
+        // If not provided, and cooldown has passed, revert
+        if (!_hasTransitionRecord) {
+            // Check if cooldown period has passed for forcing
+            if (block.timestamp < _transitionRecord.effectiveAt) {
+                // Cooldown not passed, don't force finalization
+                return (false, _proposalId);
+            }
+            // Cooldown passed, force finalization
+            require(_hasTransitionRecord, TransitionRecordNotProvided());
```
```suggestion
            revert(TransitionRecordNotProvided());
```

PR #20094, opened 2025-09-07
feat(protocol): the deployment script for devnet

YoGhurt111:


packages/protocol/contracts/layer1/shasta/impl/DevnetShastaInbox.sol
```diff
@@ -0,0 +1,82 @@
...
+        InboxOptimized3(
+            _taikoToken,
+            _checkpointManager,
+            _proofVerifier,
+            _proposerChecker,
+            2 hours, // provingWindow
+            4 hours, // extendedProvingWindow
+            16, // maxFinalizationCount
+            _RING_BUFFER_SIZE, // ringBufferSize
+            0, // basefeeSharingPctg
```
davidtaikocha: Let's change it to a non-zero value (like `75` in the current Pacaya inbox) to test this feature in devnet?

johntaiko (approved):

packages/protocol/contracts/layer1/shasta/impl/DevnetShastaInbox.sol
YoGhurt111:
```suggestion
            75, // basefeeSharingPctg
```

YoGhurt111: Done

smtmfft (approved):

PR #20114, opened 2025-09-09
fix(prover-set): add forced inclusion parameter validation for security

Snezhkko:
ProverSet was the only contract in the system that didn't validate batch parameters before 
processing. TaikoWrapper and TaikoInbox both have these same validations, so ProverSet 
should be consistent. The unused ForcedInclusionParamsNotAllowed() error was declared 
but never used - now it serves its intended purpose.

This ensures ProverSet follows the same security patterns as other parts of the system 
and prevents potential misuse of forced inclusion parameters through the prover set.


dantaik:
this contract is no longer used in production and is pending deletion. I prefer to keep it unchanged.

PR #20000, opened 2025-08-25
feat(taiko-client): shasta upgrade in client

YoGhurt111:

PR #20117, opened 2025-09-10
chore(taiko-client): remove unused blockMaxGasLimit from TxListDecompressor

sashass1315:
The TxListDecompressor never used blockMaxGasLimit; it only validates calldata size, decompresses, and RLP-decodes. Gas limit constraints are enforced upstream (engine/proposer) and at the protocol/ZKP level, not at decompression time. Removing this dead field and parameter reduces confusion and improves clarity without changing behavior.


davidtaikocha:
Those changes will be included in https://github.com/taikoxyz/taiko-mono/pull/20000

PR #20125, opened 2025-09-10
chore(protocol): abstract  bond calulcation logic in libraries

ggonzalez94:
Using the libraries for this complex code makes things more readable and avoids repetition in the case of bond calculation. 

See [this](https://github.com/taikoxyz/taiko-mono/pull/20064#issuecomment-3276831845) for full context

Update: Abstracting transition aggregation into a library increases gas cost for large arrays. So this PR only adds a library for bond calculation, and keeps the aggregation logic in the contract itself


packages/protocol/contracts/layer1/shasta/impl/InboxOptimized1.sol
```diff
@@ -53,84 +54,30 @@ contract InboxOptimized1 is Inbox {
...
-            checkpointHash: hashCheckpoint(_input.transitions[0].checkpoint)
-        });
-
-        uint48 currentGroupStartId = _input.proposals[0].id;
-        Transition memory firstTransitionInGroup = _input.transitions[0];
-
-        // Process remaining proposals
-        for (uint256 i = 1; i < _input.proposals.length; ++i) {
+        // Validate all transitions first
+        for (uint256 i = 0; i < _input.proposals.length; ++i) {
```
dantaik:
```suggestion
        for (uint256 i; i < _input.proposals.length; ++i) {
```

packages/protocol/contracts/layer1/shasta/impl/InboxOptimized1.sol
```diff
@@ -53,84 +54,30 @@ contract InboxOptimized1 is Inbox {
...
-        _setTransitionRecordHashAndDeadline(
-            currentGroupStartId, firstTransitionInGroup, currentRecord
+        // Aggregate transitions using library
+        LibTransitionAggregation.AggregatedRecord[] memory aggregatedRecords =
+        LibTransitionAggregation.aggregateTransitions(
+            _input.proposals, _input.transitions, _provingWindow, _extendedProvingWindow
         );
+
+        // Save all aggregated records to storage
+        for (uint256 i = 0; i < aggregatedRecords.length; ++i) {
```
dantaik:
```suggestion
        for (uint256 i; i < aggregatedRecords.length; ++i) {
```

dantaik: One way to fix it is to use function pointer, but it will add some small overhead, not sure if will offset the hash optimisation effort.

dantaik: test_prove_fiveConsecutiveProposals suggest more gas used. maybe in the previous impl, we reuse memory struct so no memory expension, but in the new library-based impl, we return a list of memory structs so there are memory expension?

ggonzalez94: @dantaik I ended up reverting the `LibAggregation` and keeping only `LibBondsL1`. `LibAggreation` is cleaner, but the fact that it cannot modify the mapping directly makes it less efficient when multiple records need to be changed. In this case it might be worth to keep the implementation in the contract itself

dantaik (approved):

davidtaikocha (approved):

YoGhurt111 (approved):

PR #20126, opened 2025-09-11
fix(repo): fix protocol checkout in CI

RogerLamTd:
forks have no access to the token, so will always fail. other tests don't use it or use github.token, should keep consistent where possible


RogerLamTd:
change was made previously to allow contract forge fmt to trigger CI, will reconsider

PR #20128, opened 2025-09-11
fix(eventindexer): incorrect error wrap context in transition proved handler

sashass1315:
"i.saveBlockProvenEvent", which does not exist and was a copy/paste artifact


cyberhorsey:
same changes exist in other PR

PR #20137, opened 2025-09-13
fix CORS middleware forcing 200 OK status

anim001k:
Fixed a critical bug in the CORS middleware that was forcing all responses to return `200 OK`, masking actual HTTP status codes from upstream services.

**Problem:**
- CORS middleware called `w.WriteHeader(http.StatusOK)` before `next.ServeHTTP()`
- This caused all proxy responses (including errors) to return 200 OK
- Preflight OPTIONS requests weren't handled properly

**Solution:**
- Removed forced status code setting in CORS middleware
- Added proper OPTIONS preflight handling with `204 No Content`
- Improved CORS headers configuration for better security


davidtaikocha:
Closed due to `geth-rpc-gateway` has been deprecated.

PR #20147, opened 2025-09-15
docs(docs-site): fix incorrect Hardhat initialization command

brawncode:
noticed the docs referenced `npx hardhat init`, but Hardhat doesn’t actually support that command.
updated it to the correct usage: `npx hardhat`.

**this aligns with both the official documentation and the CLI behavior.**

### DIFF
--- a/packages/docs-site/src/content/docs/guides/app-developers/deploy-a-contract.mdx
+++ b/packages/docs-site/src/content/docs/guides/app-developers/deploy-a-contract.mdx
@@ -81,7 +81,7 @@ Hardhat is an Ethereum development environment for deploying smart contracts, ru
     ```
     - Initialize Hardhat
     ```bash
-    npx hardhat init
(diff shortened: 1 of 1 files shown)


packages/docs-site/src/content/docs/guides/app-developers/deploy-a-contract.mdx
```diff
@@ -81,7 +81,7 @@ Hardhat is an Ethereum development environment for deploying smart contracts, ru
     ```
     - Initialize Hardhat
     ```bash
-    npx hardhat init
+    npx hardhat
```
myronrotter, line 84: `npx hardhat init` is correct for hardhat v2. For hardhat v3 it is `npx hardhat --init`. Just `npx hardhat` is invalid.

myronrotter (changes requested):

PR #20154, opened 2025-09-15
docs(protocol): fix Derivation.md base fee reference to Taiko L2 EIP-1559

Galoretka:
- Replace incorrect EIP-4396 references with Taiko’s EIP-1559-based L2 base fee mechanism.
- Clarify that base fee is computed via TaikoAnchor.getBasefeeV2 using protocol config, parent gas used, and current block timestamp, with details in eip1559_on_l2.md.
- This change is necessary because the previous EIP-4396 reference was inaccurate; the codebase and docs implement and document an EIP-1559-derived mechanism for L2 base fee, not EIP-4396.


dantaik:
But we are actually using 4396, not 1559.

PR #20165, opened 2025-09-16
fix(protocol): remove prover info from the transition struct to allow out of order proving

ggonzalez94:
@davidtaikocha detected an issue that would prevent out of order proving in the procotol. The Transition struct has a reference to its `parentTransitionHash`. But the problem is that in order to have that value you need the `designatedProver` and the `actualProver`. Knowing the designated prover should be possible, but there's no way to know with certainty who the actual prover is from a parent transition if the proof has not been submitted to the L1.

This PR moves both prover addresses to a new `TransitionMetadata` struct, which should not be a problem since I believe we only use it for bond decisions. Now the transition hash is deterministic, and does not depend on the previous prover.


packages/protocol/contracts/layer1/shasta/libs/LibProveInputDecoder.sol
```diff
@@ -62,6 +69,15 @@ library LibProveInputDecoder {
         for (uint256 i; i < transitionsLength; ++i) {
             (input_.transitions[i], ptr) = _decodeTransition(ptr);
         }
+
+        // 3. Decode Metadata array
+        uint24 metadataLength;
+        (metadataLength, ptr) = P.unpackUint24(ptr);
```
dantaik: Maybe we don't need to encode/decode `metadataLength`, and simply reuse `transitionsLength`. 

In the encoder function, we check these two arrays have the same size.

dantaik (approved):

davidtaikocha (approved):

davidtaikocha (approved):

PR #20166, opened 2025-09-17
feat(protocol): inline _getAvailableCapacity

dantaik:


davidtaikocha (approved):

YoGhurt111 (approved):

ggonzalez94 (changes requested):
I don't think is necessary or worth doing this. If you check the `shasta-prove.json` results the cost is actually slightly worst. This is because the compiler already inlines the private function. So you don't really get any benefits AFAIK

PR #20169, opened 2025-09-17
feat(protocol): merge CheckpointManage to Inbox and improve gas

dantaik:
## Summary

CheckpointManager functionality was merged directly into Inbox and Anchor contracts through a new shared checkpoint ring buffer.

## Highlights

- Introduces LibCheckpointStore ring-buffer storage and updates checkpoint interfaces, replacing the dedicated contract, with thorough unit test coverage.
- Inbox now implements ICheckpointStore, enforces maxCheckpointHistory, and persists checkpoints internally. The config struct propagates the history limit field across the system.
- ShastaAnchor and TaikoAnchor now track checkpoints with the new library and require the history cap during construction.


packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -129,11 +130,18 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     mapping(bytes32 compositeKey => TransitionRecordHashAndDeadline hashAndDeadline) internal
         _transitionRecordHashAndDeadline;
 
+    uint256[21] private __gap1;
```
ggonzalez94: curious why we are splitting the gap into multiple pieces? I imagine the reasoning is so that if we add variables in the future they are close to each other? Because in practice they could be added after all the other variables, without breaking the storage layout.
Also, is there a reason for the specific numbers on these gaps?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -129,11 +130,18 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
-    ///  Two slots used
-    LibForcedInclusion.Storage internal _forcedInclusionStorage;
+    /// @dev 50 slots used
+    LibForcedInclusion.Storage private _forcedInclusionStorage;
+    uint256[8] private __gap2;
 
-    uint256[39] private __gap;
+    /// @dev Storage for checkpoint management
+    /// Uses multiple slots for ring buffer storage
+    /// @dev 50 slots used
```
ggonzalez94: This is not correct right?
This occupies two consecutive slots:
```
struct Storage {
        /// @notice Ring buffer as a stack for storing checkpoints
        /// @dev Maps slot indices (0 to maxHistorySize-1) to checkpoint data
        mapping(uint48 slot => ICheckpointStore.Checkpoint checkpoint) checkpoints;
        /// @notice The latest checkpoint number
        uint48 latestCheckpointNumber;
        /// @notice The current top of the stack (ring buffer index)
        uint48 stackTop;
        /// @notice The current number of items in the stack
        uint48 stackSize;
    }

```

Shall we just remove this comment?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -338,6 +350,53 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+        pure
+        virtual
+        returns (bytes memory)
+    {
+        return abi.encode(_input);
+    }
+
+    // ---------------------------------------------------------------
+    // Public Pure Functions (Decoding)
+    // ---------------------------------------------------------------
```
ggonzalez94: redundant? since we already have a section for External Pure Functions (Encoding/Decoding) above

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
```diff
@@ -73,7 +73,7 @@ library LibForcedInclusion {
         address _feeRecipient,
         uint256 _count
     )
-        internal
+        public
```
ggonzalez94, line 77: why are we making this public?

packages/protocol/contracts/layer1/shasta/libs/LibProvedEventEncoder.sol
```diff
@@ -63,7 +67,7 @@ library LibProvedEventEncoder {
     /// @param _data The bytes to decode
     /// @return payload_ The decoded ProvedEventPayload
     function decode(bytes memory _data)
-        internal
+        public
```
ggonzalez94, line 20: is this public now so that it is deployed as a separate contract?

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -16,7 +17,7 @@ import { LibBonds } from "src/shared/based/libs/LibBonds.sol";
 ///      - Cumulative bond instruction processing with integrity verification
 ///      - State tracking for multi-block proposals
 /// @custom:security-contact security@taiko.xyz
-abstract contract ShastaAnchor is PacayaAnchor {
+abstract contract ShastaAnchor is PacayaAnchor, ICheckpointStore {
```
ggonzalez94, line 21: Should we update the natspec above to reflect that the anchor now stores the checkpoints as well? Since it is part of the diff it does not allow me to leave a suggestion

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
```diff
@@ -0,0 +1,138 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import { ICheckpointStore } from "../iface/ICheckpointStore.sol";
+
+/// @title LibCheckpointStore
+/// @notice Library for managing synced L2 checkpoints using a ring buffer
```
ggonzalez94: nit:
```suggestion
/// @notice Library for managing synced L1 or L2 checkpoints using a ring buffer
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -129,11 +130,18 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     mapping(bytes32 compositeKey => TransitionRecordHashAndDeadline hashAndDeadline) internal
         _transitionRecordHashAndDeadline;
 
+    uint256[21] private __gap1;
+
     /// @dev Storage for forced inclusion requests
-    ///  Two slots used
-    LibForcedInclusion.Storage internal _forcedInclusionStorage;
+    /// @dev 50 slots used
```
ggonzalez94: Only uses two slots. We should remove the comment

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -72,7 +73,14 @@ abstract contract ShastaAnchor is PacayaAnchor {
         blockIdToEndOfSubmissionWindowTimeStamp;
 
     /// @notice Storage gap for upgrade safety.
-    uint256[46] private __gap;
+    uint256[43] private __gap1;
+
+    /// @dev Storage for checkpoint management
+    /// 50 slots used
```
ggonzalez94: same comment as the inbox. Only uses 2 slots

ggonzalez94 (commented):
Left a few comments. Will finalize the review in the next few hours.

Are the sizes in the description updated? I'm a bit worried that we might be leaving too little room to add things to the inbox if we need in the future

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
```diff
@@ -0,0 +1,138 @@
...
+    // Structs
+    // ---------------------------------------------------------------
+
+    /// @notice Storage for the checkpoint ring buffer
+    struct Storage {
+        /// @notice Ring buffer as a stack for storing checkpoints
+        /// @dev Maps slot indices (0 to maxHistorySize-1) to checkpoint data
+        mapping(uint48 slot => ICheckpointStore.Checkpoint checkpoint) checkpoints;
+        /// @notice The latest checkpoint number
+        uint48 latestCheckpointNumber;
```
ggonzalez94: nit: `latestCheckpointBlockNumber` might be a more appropiate name

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
```diff
@@ -0,0 +1,138 @@
...
+    /// @param $ The storage struct
+    /// @return _ The latest checkpoint number
+    function getLatestCheckpointNumber(Storage storage $) public view returns (uint48) {
+        return $.latestCheckpointNumber;
+    }
+
+    /// @notice Gets the number of checkpoints
+    /// @param $ The storage struct
+    /// @return _ The number of checkpoints
+    function getNumberOfCheckpoints(Storage storage $) public view returns (uint48) {
```
ggonzalez94, line 127: a bit of a side question: but do we really need to keep track and return the number of checkpoints?
Because the result of this query will always be `_maxCheckpointHistory` once the ring buffer is full

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
```diff
@@ -0,0 +1,138 @@
...
+        uint48 _offset,
+        uint48 _maxCheckpointHistory
+    )
+        public
+        view
+        returns (ICheckpointStore.Checkpoint memory)
+    {
+        (uint48 stackTop, uint48 stackSize) = ($.stackTop, $.stackSize);
+
+        require(stackSize != 0, NoCheckpoints());
```
ggonzalez94: isn't this check redundant because if `stackSize` is zero, then the check below will always revert

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
ggonzalez94: we are also not using the `stackSize` for `getCheckpoint`, so we are already returning zero in those cases.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: You are right. After a second thought, I think maybe removing these gaps will be better.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: fixed

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: fixed

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
dantaik: Making it public will reduce the code size of Inbox smaller as the code will not be compiled in.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: Done

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
dantaik: Created an issue for this feedback: https://github.com/taikoxyz/taiko-mono/issues/20178

packages/protocol/contracts/layer1/shasta/libs/LibProvedEventEncoder.sol
dantaik: Yes

dantaik:
@claude please review

dantaik:
> Left a few comments. Will finalize the review in the next few hours.
> 
> Are the sizes in the description updated? I'm a bit worried that we might be leaving too little room to add things to the inbox if we need in the future

This PR alone won't make the size too small. There is another one coming up soon.

packages/protocol/contracts/layer1/shasta/iface/IProofVerifier.sol
```diff
@@ -9,5 +9,5 @@ interface IProofVerifier {
     /// @dev This function must revert if the proof is invalid
     /// @param _transitionsHash The hash of the transitions to verify
     /// @param _proof The proof data for the transitions
-    function verifyProof(bytes32 _transitionsHash, bytes calldata _proof) external view;
+    function verifyProof(bytes32 _transitionsHash, bytes calldata _proof) external;
```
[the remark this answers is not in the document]
dantaik: My understanding is that the STX proof verifier will mutate its own state, @smtmfft, please confirm.

ggonzalez94 (dismissed):

ggonzalez94 (approved):

dantaik:
@claude please review

dantaik:
@claude review

davidtaikocha (approved):

YoGhurt111 (approved):

dantaik:
@claude which conversions are not resolved?

PR #20186, opened 2025-09-18
feat(protocol): propose function can only be called once per Ethereum block

### LINKED ISSUE #20180: Limit proposing speed
## Problem
If the ring buffer for proposed blocks is small and we don’t throttle the proposing rate, a malicious preconfer could spam many proposals to exhaust the buffer.

## Potential Solution
Restrict the propose function so it can only be called once per L1 block.

dantaik:
## Summary
Fix Inbox vulnerability where malicious preconfers could deplete Inbox's ring buffer by proposing may proposals. Now a preconfer can only call the `propose` function once per L1 block.


dantaik:
@claude please review

davidtaikocha (approved):

YoGhurt111 (approved):

PR #20188, opened 2025-09-18
Prover: Add HTTP timeouts, response size limit, and safer logging

maradini77:
### Description
- **What**:  
  - Added `http.Client` timeouts and transport timeouts.  
  - Limited response body to 8 MiB with an early failure on overflow.  
  - Trimmed logs to show length and a short preview; avoid full body.  
  - Closed response body on non-200 with small discard for connection reuse.

- **Why**:  
  - Prevents hanging requests and potential memory DoS.  
  - Reduces risk of sensitive data leakage in logs.


davidtaikocha:
Closed due to rn don't think we need features like this for prover.

PR #20200, opened 2025-09-19
feat(protocol): shasta verifiers

ggonzalez94:
## Summary
This PR introduces non-upgradeable verifiers for the Shasta protocol, removing the upgradeability pattern from all compose verifiers(this makes gas usage lower), and if we need to upgrade them we can just deploy a new one and upgrade the inbox.

 ## Key Changes

  - **Removed upgradeability**: Compose verifiers now inherit directly from `IProofVerifier` instead of `EssentialContract`, eliminating UUPS patterns and storage gaps
- New `LibPublicInput` library that that uses the `transitionsHash` that is part of the new verifier interface instead of the old context
- Removed the access control logic from `verify` function. Since it is view only it shouldn't matter who calls it

## Future Improvements
- For now I prefixed all verifiers with `Shasta...` to avoid name conflicts with the old contracts(even if they are in different paths). Once we feel confortable removing the old verifiers we can rename these
- There might be room for making the compose verifiers a library that the inbox uses, or other ways of merging the contracts such that we don't spend gas across contract calls. But I'm a bit worried about contract size already exploding, and also wanted to keep this PR focused on the verifier functionality


packages/protocol/contracts/layer1/shasta/verifiers/LibPublicInput.sol
```diff
@@ -0,0 +1,30 @@
...
+/// @custom:security-contact security@taiko.xyz
+library LibPublicInput {
+    /// @notice Hashes the public input for the proof verification.
+    /// @param _transitionsHash The transitions hash to verify.
+    /// @param _verifierContract The contract address which as current verifier.
+    /// @param _newInstance The new instance address. For SGX it is the new signer address, for ZK
+    /// this variable is not used and must have value address(0).
+    /// @param _chainId The chain id.
+    /// @return The public input hash.
+    function hashPublicInputs(
```
dantaik, line 16: shall we inline this function somewhere...

packages/protocol/contracts/layer1/shasta/verifiers/ShastaRisc0Verifier.sol
```diff
@@ -0,0 +1,79 @@
...
+
+import "@risc0/contracts/IRiscZeroVerifier.sol";
+import "src/shared/common/EssentialContract.sol";
+import "src/shared/libs/LibNames.sol";
+import "./LibPublicInput.sol";
+import "../iface/IProofVerifier.sol";
+
+/// @title ShastaRisc0Verifier
+/// @custom:security-contact security@taiko.xyz
+contract ShastaRisc0Verifier is EssentialContract, IProofVerifier {
```
dantaik: I think we no longer need upgradability for verifiers. https://github.com/taikoxyz/taiko-mono/issues/20167

ggonzalez94: I would prefer that, but I see all of the verifiers actually hold state variables. @smtmfft are these still needed or we can get rid of them and just re-deploy the verifiers and don't depend on state?

packages/protocol/contracts/layer1/shasta/verifiers/LibPublicInput.sol
ggonzalez94: It is internal and used by the verifiers, which means it will be inlined by those. Or you meant something else?

packages/protocol/contracts/layer1/shasta/verifiers/ShastaRisc0Verifier.sol
smtmfft: I think we can get rid of them. The state in verifier are authorized image ids, we can update it during the verifier update. However, one downside is other provers need to upgrade accordingly, as there is no backward compatibility.

packages/protocol/contracts/layer1/shasta/verifiers/LibPublicInput.sol
```diff
@@ -0,0 +1,30 @@
...
+        address _verifierContract,
+        address _newInstance,
+        uint64 _chainId
+    )
+        internal
+        pure
+        returns (bytes32)
+    {
+        return keccak256(
+            abi.encode("VERIFY_PROOF", _chainId, _verifierContract, _transitionsHash, _newInstance)
```
YoGhurt111, line 27: Shall we add `prover` parameter here to prevent proof from being stolen?

ggonzalez94: what do you mean with being "stolen"? I think it does not matter if someone calls the verifier directly, the important thing is who calls the inbox. This is where we capture the `msg.sender` and decide what to do with the bonds.
Or you meant something else?

packages/protocol/contracts/layer1/shasta/verifiers/ShastaRisc0Verifier.sol
ggonzalez94: [This](https://github.com/taikoxyz/taiko-mono/pull/20225) is one possible solution. @smtmfft @dantaik WDYT?

packages/protocol/contracts/layer1/shasta/verifiers/LibPublicInput.sol
YoGhurt111: If the designated prover submits a proof when the proving window is about to expire, could a malicious prover observe this transaction in the mempool and front-run it by increasing the tips? Since the proof doesn't contain information about who generated it, wouldn't this proof be stolen? Could this scenario occur?

ggonzalez94: I don't think this is possible since it goes trough the inbox, if you call this contract directly the submission does not get recorded. They will just waste gas

YoGhurt111: What I mainly want to express is that the generated proof doesn't have a strong binding relationship with the prover who generated it. After the proving window, proverA can steal the proof generated by proverB, because the actualProver doesn't participate in the proof generation and verification process.
image
image
Or I can submit a PR how to resolve it.

YoGhurt111: Submit a PR to generally captures what I'm trying to express.
https://github.com/taikoxyz/taiko-mono/pull/20242

ggonzalez94: I haven't thought about this much, but:
If the designatedProver misses their proving window, aren't we ok with them being penalized? They have enough time to submit their proofs if they don't want to risk someone else landing this on-chain before them. So I don't think it is much of a concern. Plus, if we do this we would need to add access control restrictions such that the proof can only be submitted by the prover.

dantaik: There is a serious issue in the current code base, https://github.com/taikoxyz/taiko-mono/issues/20247. I'll work on a fix.

smtmfft: I see that #20247 already added a prover-related address field to each transition (either designated or actual). However, to avoid proof stealing, I think we still need the prover address in the final proof, unless we accept the risk that someone else might steal it after the proving window—which, in theory, should be rare.

YoGhurt111 (approved):

davidtaikocha (approved):

PR #20201, opened 2025-09-19
chore(protocol): simplify and make the whitelist more efficient

ggonzalez94:
I've been running gas benches on a fork of the devnet, and the first thing I found is that the `PreconfWhitelist` check is actually quite expensive, and the main reason is how we are getting the randomness(multiple loops and calling precompiles). I believe this is overkill and we can simplify a lot.

This PR also:
- removes the fallback preconfer(it is the same as just having one whitelisted actor)
- removes some now unused variables and requires


packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -24,6 +24,7 @@ interface IPreconfWhitelist {
     error OperatorAlreadyExists();
     error OperatorAlreadyRemoved();
     error OperatorNotAvailableYet();
+    error CannotRemoveLastOperator();
```
dantaik: sort errors

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -48,11 +44,8 @@ contract PreconfRouter is EssentialContract, IProposeBatch {
     {
         // Sender must be the selected operator for the epoch
         address preconfer = preconfWhitelist.getOperatorForCurrentEpoch();
-        if (preconfer != address(0)) {
-            require(msg.sender == preconfer, NotPreconfer());
-        } else if (fallbackPreconfer != address(0)) {
-            require(msg.sender == fallbackPreconfer, NotFallbackPreconfer());
-        }
+        require(preconfer != address(0), ProposerIsNotPreconfer());
```
dantaik, line 47: we don't need this check as the next check will guarantee preconfer won't be 0 as msg.sender won't be 0.

dantaik:
@cyberhorsey is the current fallback preconfer our preconfer or we have two addresses? Will our software needs to be modified to use one single address?

dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
ggonzalez94: We're not going to use the PreconfRouter anymore, since it is not part of Shasta, so we'll probably remove it from the codebase later

davidtaikocha (approved):

PR #20207, opened 2025-09-20
chore(protocol): add more tests and benches

ggonzalez94:
Work on the test suite, specifically around propose and finalize


ggonzalez94:
closing this one since it became very outdated. Will open a new one when I'm back

PR #20217, opened 2025-09-22
Proposal multi source

dantaik:


packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -42,17 +42,23 @@ interface IInbox {
     /// @notice Contains derivation data for a proposal that is not needed during proving.
     /// @dev This data is hashed and stored in the Proposal struct to reduce calldata size.
 
+    /// @notice Represents a source of derivation data within a Derivation
```
ggonzalez94, line 43: I think the natspec above ^^ is for the `Derivation` struct and it just got misplaced

packages/protocol/contracts/layer1/shasta/impl/InboxOptimized2.sol
```diff
@@ -61,4 +82,156 @@ contract InboxOptimized2 is InboxOptimized1 {
...
+    {
+        return LibHashing.hashBlobHashesArray(_blobHashes);
+    }
+
+    /// @inheritdoc Inbox
+    /// @notice Optimized checkpoint hashing using LibHashing
+    /// @dev Uses LibHashing for efficient checkpoint hashing
+    /// @param _checkpoint The checkpoint data to hash
+    /// @return bytes32 The keccak256 hash of the checkpoint struct
+    function _hashCheckpoint(ICheckpointStore.Checkpoint memory _checkpoint)
```
ggonzalez94, line 153: @dantaik now that we are merging the optimized implementations maybe is a good moment tos ask: I haven't seen enough evidence that custom hashing  reduces gas costs for proposing or proving significantly. See [this PR](https://github.com/taikoxyz/taiko-mono/pull/20207) for more details.
Where have you seen the most gains from this?

packages/protocol/contracts/layer1/shasta/impl/DevnetShastaInbox.sol
```diff
@@ -1,19 +1,15 @@
...
 /// @title DevnetShastaInbox
 /// @dev This contract extends the base Inbox contract for devnet deployment
 /// with optimized reentrancy lock implementation.
-/// @dev DEPLOYMENT: CRITICAL - Must use FOUNDRY_PROFILE=layer1o for deployment.
-///      Contract size (26,293 bytes) exceeds 24KB limit without optimization.
-///      Example: FOUNDRY_PROFILE=layer1o forge build
-/// contracts/layer1/shasta/impl/DevnetShastaInbox.sol
 /// @custom:security-contact security@taiko.xyz
-contract DevnetShastaInbox is InboxOptimized4 {
+contract DevnetShastaInbox is InboxOptimized1 {
```
ggonzalez94: shouldn't this use `InboxOptimized2` instead?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -195,58 +192,94 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+                );
+                sourceCount += forcedInclusions.length;
+            }
 
-        if (input.numForcedInclusions > 0) {
-            // Process forced inclusion if required
-            uint256 numForcedInclusionsProcessed;
-            (coreState, numForcedInclusionsProcessed) = _processForcedInclusions(
-                coreState, input.numForcedInclusions, endOfSubmissionWindowTimestamp
+            require(sourceCount > 0, EmptyDerivationSources());
```
ggonzalez94: not needed? given we initialize `sourceCount` to 1 this check will always be true

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -821,63 +863,6 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
-    /// @dev Creates and stores a new proposal
-    /// @notice Increments nextProposalId and emits Proposed event
-    /// @param _coreState Current state whose hash is stored in the proposal
-    /// @param _blobSlice Blob data slice containing L2 transactions
-    /// @param _isForcedInclusion True if this is a forced inclusion proposal
-    /// @param _endOfSubmissionWindowTimestamp The timestamp of the last slot where the current
-    /// preconfer
-    /// can propose.
-    /// @return Updated core state with incremented nextProposalId
-    function _propose(
```
ggonzalez94: Curious why we decided to inline all this logic directly in the propose function vs keeping it in a separate one? Since this was private already I don't think there are many gas benefits from doing so

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -634,17 +637,76 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
         return keccak256(abi.encode(_proposalId, _parentTransitionHash));
     }
 
-    /// @dev Hashes a Transition struct.
-    /// @param _transition The transition to hash.
-    /// @return _ The hash of the transition.
-    function _hashTransition(Transition memory _transition)
+    // ---------------------------------------------------------------
+    // Encoder Functions (alphabetically sorted)
```
ggonzalez94: nit:
```suggestion
    // Encoder Functions
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -195,58 +192,94 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+                timestamp: uint48(block.timestamp),
+                endOfSubmissionWindowTimestamp: endOfSubmissionWindowTimestamp,
+                proposer: msg.sender,
+                coreStateHash: _hashCoreState(coreState),
+                derivationHash: _hashDerivation(derivation)
+            });
+
+            _setProposalHash(proposal.id, _hashProposal(proposal));
+
+            // Create payload and emit event with stack-optimized approach
```
ggonzalez94: what is the stack optimized approach? putting it on a separate function?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -634,17 +637,76 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
         return keccak256(abi.encode(_proposalId, _parentTransitionHash));
     }
 
-    /// @dev Hashes a Transition struct.
-    /// @param _transition The transition to hash.
-    /// @return _ The hash of the transition.
-    function _hashTransition(Transition memory _transition)
+    // ---------------------------------------------------------------
```
ggonzalez94, line 615: nit: maybe we can group all sorting functions in the same group vs having separate groups

ggonzalez94 (commented):
I really like the fact that forced inclusions and the regular proposal can be part of the same one. Left a few comments

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
ggonzalez94: This [PR](https://github.com/taikoxyz/taiko-mono/pull/20228/files) brings it back and I think the gas increase is negligible

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: remove this deprecated comment

dantaik:
all comments are addressed in the other PR

PR #20219, opened 2025-09-22
refactor(protocol): remove unreachable zero-proof path; require non-empty proofs in SignalService

Galoretka:
- Eliminates the vestigial _receivedSignals mapping and the empty-proof branch in _verifySignalReceived, which could never succeed because nothing sets _receivedSignals to true.
- Drops the unused SS_SIGNAL_NOT_RECEIVED error.
- Updates ISignalService comments to state that a non-empty proof is required, removing outdated references to “marked as received by TaikoL2.”
- This aligns the implementation with current protocol behavior and tests, where L2 no longer marks signals as received and PacayaAnchor explicitly disallows passing signal slots. Prevents misleading API/documentation and unreachable code paths.


packages/protocol/contracts/shared/signal/SignalService.sol
```diff
@@ -20,8 +20,6 @@ contract SignalService is EssentialResolverContract, ISignalService {
     /// @dev Slot 2.
     mapping(address addr => bool authorized) public isAuthorized;
 
-    mapping(bytes32 signalSlot => bool received) internal _receivedSignals;
```
dantaik, line 23: we should keep it to maintain the same storage layout, but we can change it to something like:

```solidity
mapping(bytes32 signalSlot => bool received) private __deprecated_receivedSignals;
```

dantaik (changes requested):

packages/protocol/contracts/shared/signal/SignalService.sol
Galoretka: > we should keep it to maintain the same storage layout, but we can change it to something like:
> 
> ```solidity
> mapping(bytes32 signalSlot => bool received) private __deprecated_receivedSignals;
> ```

made corrections

dantaik (approved):

dantaik (approved):

davidtaikocha (approved):

dantaik:
There is a test failure, maybe the test should be deleted.

```
Error (9582): Member "SS_SIGNAL_NOT_RECEIVED" not found or not visible after argument-dependent lookup in type(contract SignalService).
   --> test/shared/signal/SignalService.t.sol:139:25:
    |
139 |         vm.expectRevert(SignalService.SS_SIGNAL_NOT_RECEIVED.selector);
```

Galoretka:
> There is a test failure, maybe the test should be deleted.
> 
> ```
> Error (9582): Member "SS_SIGNAL_NOT_RECEIVED" not found or not visible after argument-dependent lookup in type(contract SignalService).
>    --> test/shared/signal/SignalService.t.sol:139:25:
>     |
> 139 |         vm.expectRevert(SignalService.SS_SIGNAL_NOT_RECEIVED.selector);
> ```

Yes boss!

dantaik (approved):

PR #20224, opened 2025-09-22
feat(protocol): a radical proposal for pacaya -> shasta transition 

ggonzalez94:
Please read the `pacaya-shasta-transition.md` document carefully.

This PR proposes a radical approach to the Pacaya -> Shasta transition where we get rid of the fork router entirely, and the new inbox is deployed to a new address and does not need to have any backwards compatibility(in storage or otherwise) with Pacaya.
Given this is a completely new protocol design and storage layout, I couldn't think on too many benefits of reusing the fork router.

---

> [!NOTE]
> Introduce an activatable Shasta Inbox with initializer-based activation, remove Pacaya compatibility (including bond withdrawal), update deployment/tests accordingly, and add a transition plan doc.
> 
> - **Shasta Inbox (`contracts/layer1/shasta/impl/Inbox.sol`)**:
>   - Add `init(address owner, address shastaInitializer)` and one-time `activate(bytes32 genesisBlockHash)`; store `_shastaInitializer` and clear it after activation.
>   - Rename/internalize activation logic to `_activateInbox`.
>   - Remove Pacaya compatibility state, `bondBalance`, and `withdrawBond`; drop `BondWithdrawn` event.
> - **Deployment (`script/layer1/based/DeployProtocolOnL1.s.sol`)**:
>   - Initialize proxy via `Inbox.init(owner, sender)` instead of `initV3(owner, L2_GENESIS_HASH)`.
> - **Tests (suite2 deployers)**:
>   - Use new `Inbox.init(owner, activator)` and explicitly call `activate(genesisHash)`.
> - **Docs**:
>   - Add `fork-router/pacaya-shasta-transition.md` detailing the Pacaya→Shasta transition plan and rationale for not using a fork router.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8ccaaa464500fe7f667c51c478e3aff32e10c152. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -155,23 +129,23 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
-        if (owner == address(0)) {
-            __Essential_init(_owner);
-        }
-        _initializeInbox(_genesisBlockHash);
+    /// @notice Activates the inbox so that it can start accepting proposals.
+    /// @dev Only the `shastaInitializer` or the owner can call this function.
+    /// @param _genesisBlockHash The hash of the genesis block
+    function activate(bytes32 _genesisBlockHash) external {
+        require(msg.sender == _shastaInitializer || msg.sender == owner(), ACCESS_DENIED());
+        _activateInbox(_genesisBlockHash);
```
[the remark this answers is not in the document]
ggonzalez94: I'll fix this soon. One option is to only allow calls from the `_shastaInitializer` address and as part of this function call set that address to zero, ensuring it cannot be called again.

dantaik:
I think the most challenging part is how to initialize Shasta’s initial state. Since some batches in Pacaya are not yet finalized, we cannot bootstrap Shasta directly from on-chain data. This means the initial state must be provided by a trusted party.

One caveat is that the trusted party (i.e., us) could accidentally set an incorrect genesis block hash. To mitigate this risk, we should add/enable a re-init function that allows Shasta’s state to be reset if necessary.

For ownership:
- Shasta Inbox should initially be controlled by the admin.taiko.eth multisig.
- Once the transition is complete and all Pacaya blocks are verified/finalized, ownership should then be transferred to delegate.taiko.eth.

@ggonzalez94 — I’ll leave the detailed design to you and the team, as I don’t see a better approach right now.

packages/protocol/contracts/layer1/fork-router/pacaya-shasta-transition.md
```diff
@@ -0,0 +1,48 @@
+## Goal
+
+The goal of this document is to outline a plan for the Pacaya->Shasta transition in the most secure and gas efficient way possible. ** Since the Shasta Inbox contract will still be deployed with a whitelisted set of preconfers, we can make use of those security assumptions to simplify the transition and avoid having to do too much on-chain. **
+
+## Terminilogy
+
+- `FORK_BLOCK`: The L1 block where the Pacaya -> Shasta transition should happen
```
dantaik, line 7: This is the first L1 block in which the Shasta `propose` function can be called without reverting, right?

packages/protocol/contracts/layer1/fork-router/pacaya-shasta-transition.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Steps
+
+1. Deploy all the Shasta contracts, and initialize the inbox with `_genesisBlockHash` set to zero. This sets the inbox to an empty state. ** No proposals should happen to the inbox at this point **
+   NOTE: We can enforce this onchain by reverting if a proposal happens before `FORK_BLOCK`, but this will increase the cost of every subsequent call to the inbox. Given we have a whitelisted set of proposers another option is to rely on that trust assumption, and not add the on-chain check.
+2. Upgrade the Pacaya Inbox to stop accepting proposals after `FORK_BLOCK`.
+3. At `FORK_BLOCK` proposers stop submitting proposals to the Pacaya Inbox. This ensures we have a reliable state. After this moment proposers should start preconfing with the new shasta block structure.
+4. The `SHASTA_INITIALIZER` calls the `activate` function on the shasta inbox with the `_genesisBlockHash` as of the latest Pacaya batch.
```
davidtaikocha, line 45: And from the proving perspective, does this also mean that "A whitelist address optimistically proved / finalized that genesis Shasta block (which also means proofs and cooldown windows of a bunch of final Pacaya blocks might will be skipped before Shasta fork)". And ppl can start using that `_genesisBlockHash` to generate valid cross-chain merkle proof.

packages/protocol/contracts/layer1/fork-router/pacaya-shasta-transition.md
```diff
@@ -0,0 +1,48 @@
...
+   NOTE: This can be computed off-chain, and while it has not been proven yet, the `SHASTA_INITIALIZER` can ensure it will eventually be proven.
```
YoGhurt111, line 46: Is the `SHASTA_INITIALIZER` owned by DAO or by Multi-sig?

packages/protocol/contracts/layer1/fork-router/pacaya-shasta-transition.md
ggonzalez94: correct

packages/protocol/contracts/layer1/fork-router/pacaya-shasta-transition.md
ggonzalez94: yes, feel free to edit the md if it isn't clear

packages/protocol/contracts/layer1/fork-router/pacaya-shasta-transition.md
ggonzalez94: It should be an address that we control, likely an EOA because it needs to call `activate` in a short period of time(we have around 1 epoch window between proposals)

YoGhurt111: I'm just a bit concerned about whether this approach would be accepted by the community. In theory, if malicious intent arises, this could directly modify the L2 asset status.

ggonzalez94: agreed, but given that the storage layout is completely different I don't see a way to do this on-chain. So some trust is unfortunately needed. The good news is that it can be easily verified if the state was set to the right value, and it would become very evident if it didn't

davidtaikocha (approved):

PR #20225, opened 2025-09-22
feat(protocol): make verifiers non upgradable

ggonzalez94:
A possible solution to [this comment](https://github.com/taikoxyz/taiko-mono/pull/20200#discussion_r2365002316).

Verifiers still hold state, but are non upgradable. This allows:
- Less proxy overhead
- Proving system can be upgraded with backwards compatibility(two programs can be set as trusted until all provers migrate)
- If the verifier needs to be changed, a new one can be deployed and the inbox can point to that one

### DIFF
--- a/packages/protocol/contracts/layer1/shasta/verifiers/ShastaRisc0Verifier.sol
+++ b/packages/protocol/contracts/layer1/shasta/verifiers/ShastaRisc0Verifier.sol
@@ -2,14 +2,14 @@
 pragma solidity ^0.8.24;
 
 import "@risc0/contracts/IRiscZeroVerifier.sol";
-import "src/shared/common/EssentialContract.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
 import "src/shared/libs/LibNames.sol";
 import "./LibPublicInput.sol";
 import "../iface/IProofVerifier.sol";
 
 /// @title ShastaRisc0Verifier
 /// @custom:security-contact security@taiko.xyz
-contract ShastaRisc0Verifier is EssentialContract, IProofVerifier {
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/layer1/shasta/verifiers/ShastaRisc0Verifier.sol
```diff
@@ -2,14 +2,14 @@
...
-import "src/shared/common/EssentialContract.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
 import "src/shared/libs/LibNames.sol";
 import "./LibPublicInput.sol";
 import "../iface/IProofVerifier.sol";
 
 /// @title ShastaRisc0Verifier
 /// @custom:security-contact security@taiko.xyz
-contract ShastaRisc0Verifier is EssentialContract, IProofVerifier {
+contract ShastaRisc0Verifier is IProofVerifier, Ownable {
```
dantaik: Lets use `Ownable2StepUpgradeable`. Circle's USDC contacts use Ownable and I transferred the ownership to a wrong address a few days ago, but luckily I found out a way to fix it. I think Ownable2StepUpgradeable is much safer.

ggonzalez94: I thought the whole idea was to make them non upgradable. So I guess you mean `Ownable2Step`(non upgradable). The one thing to keep in mind is that the DAO will have to accept ownership, so we will have to do a separate step. Is that ok?

dantaik: Sorry, yes Ownable2Step instead. Not Ownable2StepUpgradable :D

PR #20230, opened 2025-09-23
ci(protocol): improve claude workflow

dantaik:
Address this feedback:

> Dear Taiko,
https://github.com/taikoxyz/taiko-mono/blob/main/.github%2Fworkflows%2Frepo--claude.yml is vulnerable to Remote Code ExecutionW
While Claude Code will only run when a collaborator comments, it has been provided the instruction You MUST run `pnpm install` first.. While this wouldn’t seem like an issue, since an authorized user told it to do so, there is no verification that the PR is safe.
Between the time-of-comment, and time-of-execution, the PR could have changed. In the above step, https://github.com/taikoxyz/taiko-mono/blob/main/.github%2Fworkflows%2Frepo--claude.yml#L25, when the PR is checked-out, it may have had malicious content pushed after the comment. So, pnpm install could run malicious lifecycle scripts, leading to RCE.

### DIFF
--- a/.github/workflows/repo--claude.yml
+++ b/.github/workflows/repo--claude.yml
@@ -21,8 +21,8 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          # account for forked repos case
-          ref: ${{ steps.comment-branch.outputs.head_ref }}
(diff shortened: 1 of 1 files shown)


dantaik:
@claude please review

davidtaikocha (approved):

ggonzalez94 (approved):

PR #20238, opened 2025-09-23
feat(protocol): introduce ICodec to simplify Inbox testing

dantaik:
## Summary

  Introduces a new codec interface (ICodec) to standardize encoding/decoding operations for Taiko's Shasta inbox system, replacing the previous InboxHelper approach with a more modular codec pattern.


dantaik:
@claude review

ggonzalez94 (commented):
Will the Codec contracts be deployed as standalones? Shouldn't the inbox still reference which one is using? (vs requiring the off-chain software to track it).
I think I'm not 100% sure I understand the benefit of this refactor. If you can ellaborate a bit that would be great

dantaik:
I think the major benefit of having an interface for codec is for the test to always assume a pair of (Inbox, Codec) so tests for Inbox, InboxOptimizedXxx do not need to have if-else to decide which codec to use.

dantaik:
It's a great idea to have a reference from inbox to its codec. I've made the change in a new commit.

davidtaikocha (approved):

davidtaikocha:
> It's a great idea to have a reference from inbox to its codec. I've made the change in a new commit.

Looks like rn we only save an address of a codec, but each version of inbox is still using its own encoding / decoding methods, so we need to  manually align the impl of the saved codec for each inbox with care?

packages/protocol/script/layer1/based/DeployProtocolOnL1.s.sol
```diff
@@ -314,17 +314,17 @@ contract DeployProtocolOnL1 is DeployCapability {
                 )
             );
         }
-        address helper = address(new InboxHelper());
+        address simpleCodec = address(new SimpleCodec());
         address tempFork =
-            address(new DevnetShastaInbox(proofVerifier, whitelist, bondToken, helper));
+            address(new ShastaDevnetInbox(simpleCodec, proofVerifier, whitelist, bondToken));
```
davidtaikocha: I see `ShastaDevnetInbox` is from `InboxOptimized2`, and `SimpleCodec` is not aligned with the encoding / decoding methods in `InboxOptimized2`?

YoGhurt111:
Still using `abi.encode` in `Inbox.sol`, not `ICodec(_codec).encodexxx`

packages/protocol/script/layer1/based/DeployProtocolOnL1.s.sol
dantaik: Good catch, fixed.

dantaik:
> Still using `abi.encode` in `Inbox.sol`, not `ICodec(_codec).encodexxx`

That's the expected behavior. In another PR, we'll use a library, but _codec is only for external use.

davidtaikocha (approved):

YoGhurt111:
> > Still using `abi.encode` in `Inbox.sol`, not `ICodec(_codec).encodexxx`
> 
> That's the expected behavior. In another PR, we'll use a library, but _codec is only for external use.

In this way, we need to manually ensure that the impl of `ICodec` corresponds to the current contract, whether it is `Inbox` or `InboxOptimized2`.

YoGhurt111 (approved):

davidtaikocha (approved):

dantaik:
> > > Still using `abi.encode` in `Inbox.sol`, not `ICodec(_codec).encodexxx`
> > 
> > 
> > That's the expected behavior. In another PR, we'll use a library, but _codec is only for external use.
> 
> In this way, we need to manually ensure that the impl of `ICodec` corresponds to the current contract, whether it is `Inbox` or `InboxOptimized2`.

Yes. Calling codec functions from inbox is too expensive.

dantaik:
@ggonzalez94 @claude  please take another look.

YoGhurt111 (approved):

packages/protocol/contracts/layer1/automata-attestation/AutomataDcapV3Attestation.sol
```diff
@@ -1,8 +1,8 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ^0.8.24;
 
-import "solady/src/utils/Base64.sol";
-import "solady/src/utils/LibString.sol";
+import "solady/utils/Base64.sol";
```
ggonzalez94: any particular reason why we are refactoring all these unrelated imports? or did the AI do it?

packages/protocol/contracts/layer1/preconf/impl/LookaheadSlasher.sol
```diff
@@ -24,9 +24,7 @@ contract LookaheadSlasher is ILookaheadSlasher, EssentialContract {
         address _lookaheadStore,
         address _preconfSlasher,
         uint256 _slashAmount
-    )
-        EssentialContract()
-    {
+    ) {
```
ggonzalez94, line 27: shouldn't we call the parent's contract? I don't think in this case it has super critical logic, but still

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -154,7 +156,7 @@ interface IInbox {
         Transition[] transitions;
         /// @notice Array of metadata for prover information.
         /// @dev Must have same length as transitions array.
-        TransitionMetadata[] metadata;
+        TransitionMetadata[] metadatas;
```
ggonzalez94: nit: we should keep metadata. It is already plural:

> "Metadata" is already a plural form (the singular is "metadatum"), and in English, it's treated as a mass noun or collective noun

packages/protocol/test/layer1/shasta/inbox/suite2/propose/AbstractPropose.t.sol
```diff
@@ -42,20 +42,19 @@ abstract contract AbstractProposeTest is InboxTestSetup {
         vm.startPrank(currentProposer);
         // Act: Submit the proposal
         vm.startSnapshotGas(
-            "shasta-propose",
-            string.concat("propose_single_empty_ring_buffer_", _getInboxContractName())
+            "shasta-propose", string.concat("propose_single_empty_ring_buffer_", inboxContractName)
```
ggonzalez94: I don't know when we did this change. But the `vm.startSnapshotGas`cheatcode should be before the call to propose to capture that, otherwise we are measuring the test setup

ggonzalez94 (commented):
Left a few comments, but otherwise looks good.

packages/protocol/contracts/layer1/automata-attestation/AutomataDcapV3Attestation.sol
dantaik: I revert the changes as I found another way to fix the issue.

packages/protocol/contracts/layer1/preconf/impl/LookaheadSlasher.sol
dantaik: According to the compiler, we shall not, if the constructor doesn't take any parameters -- the call is automatic.

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
dantaik: Reverted to "data"

packages/protocol/test/layer1/shasta/inbox/suite2/propose/AbstractPropose.t.sol
dantaik: Fixed

dantaik:
@ggonzalez94 ready for anther review

ggonzalez94 (approved):

PR #20242, opened 2025-09-23
feat(protocol): add prover into public inputs

YoGhurt111:
refer to: https://github.com/taikoxyz/taiko-mono/pull/20200#discussion_r2372678994
Add prover into public inputs, so the proof only can be used by the prover who generated it.


dantaik:
There is a bigger issue in the base class: designatedProver is not validated.

dantaik:
https://github.com/taikoxyz/taiko-mono/issues/20247 please wait until this issue is addressed

dantaik:
@YoGhurt111 @smtmfft please see https://github.com/taikoxyz/taiko-mono/pull/20250, this PR should fix the issue and this PR should be closed.

PR #20246, opened 2025-09-23
docs(repo): add codex agent guides

ggonzalez94:
## Summary
- add root-level guide for coding agents working across the monorepo
- document protocol package conventions, runbooks, and checklists for Solidity work

## Testing
- not run (docs only)

### DIFF
--- a/Agents.md
+++ b/Agents.md
@@ -1,0 +1,58 @@
+# agents.md — Root (Monorepo Guide for Coding Agents)
+
+> **Canonical brief for this repository.** Agents should also read `/packages/protocol/agents.md` when editing smart contracts. Keep responses terse, deterministic, and actionable. Always include exact file paths and runnable commands.
+
+## Project Snapshot
+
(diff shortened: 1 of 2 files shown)


dantaik (approved):

dantaik:
I think we should use AGENTS.md  instead of Agents.md, and maybe we can also simply link to CLAUDE.md from AGENTS.md without duplicating the content.

```
See [Claude](./CLAUDE.md) for details.
```

ggonzalez94:
> I think we should use AGENTS.md instead of Agents.md, and maybe we can also simply link to CLAUDE.md from AGENTS.md without duplicating the content.
> 
> ```
> See [Claude](./CLAUDE.md) for details.
> ```

this would definitely be preferable, although I'm not sure how good the models are at following instructions when they are presented this way. Have you seen posts or tweets about this? Hopefully claude just adopts Agents.md since it is an open standard

davidtaikocha (approved):

PR #20250, opened 2025-09-24
fix(protocol): fix TransitionMetadata bug and refactor hashing libs

### LINKED ISSUE #20247: bug: TransitionMetadata is not validated or proved
The prove function does not validate or prove the TransitionMetadata input, which means msg.sender can inject arbitrary and potentially malicious values.

dantaik:
## Summary

- **Codec renaming**: Renamed OptimizedCodec.sol → CodecOptimized.sol and SimpleCodec.sol → CodecSimple.sol

- **Hashing refactor**: Replaced direct keccak256(abi.encode()) calls with new library methods un LibHashSimple

- **Transition metadata handling**: Updated method signatures to handle transition metadata arrays and improved metadata processing. This fixes https://github.com/taikoxyz/taiko-mono/issues/20247

@YoGhurt111 @davidtaikocha @smtmfft @ggonzalez94 Each transition is associated with an actualProver address. Consequently, a prover entity may control multiple actualProver addresses and aggregate the proofs of different transitions -- each with a potentially different address -- into a consolidated proof.


YoGhurt111 (approved):

smtmfft (approved):
LGTM

PR #20252, opened 2025-09-24
fix(protocol): compute stored lookahead hash from provided slots

VolodymyrBg:
- The stored lookahead hash was computed from a newly allocated local array that was never populated, while the event emitted the original _lookaheadSlots. This caused the stored hash to mismatch the event payload and the expected hash used by LookaheadSlasher for verification, leading to LookaheadHashMismatch().
- This change computes the hash using the original _lookaheadSlots, aligning the stored value with the emitted data and the interface’s documented behavior (same slots used for commitment payload and storage hash). This also avoids unnecessary memory copying and reduces gas.


ggonzalez94 (commented):
Thanks @VolodymyrBg great catch! Can you also remove the unused local array so we don't waste gas on memory expansion? Thanks!

VolodymyrBg:
> Thanks @VolodymyrBg great catch! Can you also remove the unused local array so we don't waste gas on memory expansion? Thanks!

done

dantaik (approved):

ggonzalez94 (approved):

PR #20253, opened 2025-09-24
refactor(protocol): improve lookahead store readability and add views for assisting the node

AnshuJalan:


packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
```diff
@@ -115,6 +127,19 @@ interface ILookaheadStore {
         pure
         returns (bytes26);
 
+    /// @notice Returns the proposer context for the given lookahead input and epoch.
+    /// @dev This is useful for offchain nodes in determining the next proposer/preconfer.
+    /// @param _data The lookahead data for current and next epoch.
+    /// @param _epochTimestamp The timestamp of the current epoch.
```
ggonzalez94: does this work only for the current epoch or this and next epoch? the comment might be a bit missleading

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -122,114 +117,129 @@ contract LookaheadStore is ILookaheadStore, Blacklist, EssentialContract {
...
 
-        if (result.lookaheadHash == 0) {
-            result = _updateLookaheadForNextEpoch(_nextEpochTimestamp, _data);
+        if (nextLookaheadHash == 0) {
+            _updateLookaheadForNextEpoch(_nextEpochTimestamp, _context, _data);
         } else {
-            // Since the lookahead was not updated in the same transaction, we need to validate
-            // the slot supplied as evidence
-            result.isLookaheadValidationRequired = true;
+            _validateLookahead(_nextEpochTimestamp, _data.nextLookahead, nextLookaheadHash);
```
ggonzalez94, line 138: unrelated to this PR: but can you remind me why we need to validate the next epoch's lookahead if it is already stored? Can't we just not pass this parameter since we are not updating it in this case?

dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -122,114 +117,129 @@ contract LookaheadStore is ILookaheadStore, Blacklist, EssentialContract {
...
+                operatorData.unregisteredAt != type(uint48).max || operatorData.slashedAt != 0
+                    || !isOptedIn || isOperatorBlacklisted(context_.lookaheadSlot.registrationRoot)
+            ) {
+                // If the operator is slashed, unregistered, not opted in, or blacklisted
+                // we use the fallback preconfer (whitelist)
+                context_.isFallback = true;
+                context_.proposer = IPreconfWhitelist(preconfWhitelist).getOperatorForCurrentEpoch();
+            } else {
+                // Use the opted in preconfer
+                context_.proposer = context_.lookaheadSlot.committer;
```
ggonzalez94, line 204: can you remind me why do we need all these checks? I thought we did all the checks when the lookahead is posted and from then we assume it is correct and just use it

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
ggonzalez94: I guess this is necessary when the preconfer is for the next epoch right? I'll create an issue to see if there are ways we can optimize it to avoid running this check and passing it every time

ggonzalez94 (commented):
This looks better. I left a few comments

packages/protocol/contracts/layer1/preconf/iface/ILookaheadStore.sol
AnshuJalan: In theory, it works for all epochs, and we realised we do need this to call for both the current and next epoch. Will update the comment

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: I think this is already optimised in `_handleNextEpochLookahead(..)`. We do not validate the lookahead if it is updated in the current epoch

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
AnshuJalan: The preconfer could still unregister, opt out, or be slashed (by taiko, or any other preconfer) after the lookahead is posted. In that case, we do not allow them to propose

ggonzalez94: I see.... I think this is fine for now. but I'll create an issue so we can discuss this. I believe the gas overhead of this might be too high and we may want to just accept the lookahead for the next 2 epochs as immutable(but of course we need to think about the consequences)

ggonzalez94 (approved):
LGTM

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -80,20 +68,28 @@ contract LookaheadStore is ILookaheadStore, IProposerChecker, Blacklist, Essenti
...
+            InvalidLookaheadTimestamp()
+        );
+
+        // Validate the current lookahead evidence
+        _validateCurrentEpochLookahead(epochTimestamp, data.currLookahead);
 
-        // Step 4: Validate the actual proposer
-        _validateProposer(_proposer, context);
+        // Validate the next lookahead evidence and update the store if required
+        _handleNextEpochLookahead(nextEpochTimestamp, context, data);
```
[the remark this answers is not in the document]
ggonzalez94: it is validated afterwards, so this should be fine. Tagging you @AnshuJalan just in case, but merging this PR

PR #20261, opened 2025-09-25
fix(protocol): decouple bond processing from anchor blocks

dantaik:
## Summary
- Optimizes bond instructions processing in the anchor system by introducing a processing delay
- Removes unnecessary BondInstructed event emissions to reduce gas costs  
- Updates documentation to reflect the new delayed processing approach

## Changes Made
- Added `BOND_PROCESSING_DELAY` constant (set to 6) in `LibManifest.sol`
- Removed `BondInstructed` event and emission from `Inbox.sol` 
- Updated `ShastaAnchor.sol` to process bond instructions from ancestor proposals
- Updated `Derivation.md` documentation to explain the new delayed processing mechanism
- Updated gas snapshots reflecting the optimizations

## Test Plan
- [x] All existing tests pass
- [x] Gas snapshots updated and verified
- [x] Documentation updated to reflect changes

🤖 Generated with [Claude Code](https://claude.ai/code)


dantaik:
@smtmfft please take a look at this PR to address your concerns.
I do have a question: how will you prove the value of  the anchor transaction's `_bondInstructionsHash` parameter is correct?

davidtaikocha (approved):
changes LGTM, still need @smtmfft 's feedback on whether this works well with the proving system.

smtmfft:
> @smtmfft please take a look at this PR to address your concerns. I do have a question: how will you prove the value of the anchor transaction's `_bondInstructionsHash` parameter is correct?

no need because it's part of the transition record, a bad one will not hit the record. from prover's point of view, we simply check if the anchor's input links to the proposal input event.

smtmfft (approved):

PR #20265, opened 2025-09-25
feat(taiko-client): update prover `l1Current` cursor initilization for Shasta protocol

davidtaikocha:


packages/taiko-client/prover/init.go
```diff
@@ -275,6 +275,12 @@ func (p *Prover) initL1Current(startingBatchID *big.Int) error {
 		return err
 	}
 
+	// Try to initialize L1Current cursor for Shasta protocol first.
+	if err := p.initL1CurrentShasta(startingBatchID); err == nil {
```
YoGhurt111, line 279: Shall we add new flag for the "startingBatchID"?
The "startingBatchID" is not continuous in current Pacaya => Shasta transition design.

davidtaikocha: Don't think its necessary to introduce one more flag, after the chain forked into Shasta, the batch ID will be continuous, and we will finalize the first Shasta block as soon as we reach the last Pacaya batch.

YoGhurt111 (approved):

johntaiko (approved):

PR #20268, opened 2025-09-25
fix CORS middleware: avoid premature 200 and handle OPTIONS

anim001k:
- Description:
  - This PR fixes a CORS middleware issue that prematurely sent 200 OK before passing control to the next handler, potentially masking upstream errors.
  - Adds explicit handling for CORS preflight (OPTIONS) with 204 No Content.
  - Sets Access-Control-Allow-Methods to a safe, fixed list ("GET, POST, OPTIONS") instead of mirroring the request method.

- Why:
  - Prevents incorrect 200 responses when upstream returns errors.
  - Ensures standards-compliant CORS behavior for preflight requests.

- Impact:
  - More accurate HTTP status codes and headers in proxied responses.
  - No breaking changes expected for consumers.

### DIFF
--- a/packages/geth-rpc-gateway/main.go
+++ b/packages/geth-rpc-gateway/main.go
@@ -94,10 +94,16 @@ func enableCORS(next http.Handler) http.Handler {
 			w.Header().Set("Vary", "Origin") // Ensure caching based on origin
 		}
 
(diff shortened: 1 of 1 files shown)


davidtaikocha:
Closed due to we've deprecated geth-rpc-gateway.

PR #20270, opened 2025-09-25
feat(protocol): integrate signal service with shasta

### LINKED ISSUE #20010: Implement ISyncBlockManager and consume from Signal Service
The new inbox expects an `ISyncBlockManager` contract where it will store the block hash and state root of synced blocks. This should be consumed by the signal service instead of the current approach where state roots are stored as signals.

A few decisions we still need to think trough: 
- Implement the ISyncBlockManager in a separate contract vs just use the SignalService. One solution I like is doing an abstract [CommitmentStore](https://github.com/OpenZeppelin/minimal-rollup/blob/main/src/protocol/CommitmentStore.sol) and inheriting in the SS like we did on the minimal-rollup.
- Should we store every finalized block? for gas efficiency we might want to store only one every X finalized blocks for example, where X is a configurable parameter.
- Should the ISyncBlockManager implementation use a ring buffer to optimize storage? This is very likely yes, but we need to decide how big the ring buffer should be. If it is too small, then merkle proofs that were created become stale too quickly

### LINKED ISSUE #20011: Simplify SignalService
The current version of the SignalService is overly complex. We can use shasta as an opportunity to simplify it, **but we need to make sure we can upgrade the contract to the new version and have a way to still verify pending messages or a plan to deal with it**.

A non-exhaustive list of things that can be simplified:
1. remove hop signals
2. remove storing state roots as signals

ggonzalez94:
Fixes #20011 , Fixes #20010 
# Signal Service Integration for Shasta

## Summary

This PR integrates SignalService with Shasta while maintaining ABI compatibility for sending and verifying signals. Key improvements include:

- **Removed all hop logic** and retained only the message structure, significantly simplifying the implementation
- **Moved checkpoint tracking to SignalService**, providing:
  - Reduced gas costs for signal verification (the most frequent operation) as SignalService no longer needs external contract calls
  - Smaller contract size for the Inbox contract
  - More streamlined architecture with clear separation of concerns
- **Replaced ring buffer with simple mapping** for checkpoint storage to enable verification by blockId, which wasn't compatible with the ring buffer design. While storage writes are slightly more expensive, this is offset by the removal of other storage variables and overall contract simplification

## Future Work

- Create a v2 of the `verifySignal` functions using a simpler data structure that reduces calldata and provides more streamlined logic. Applications can migrate at their own pace (#20271)
- Add comprehensive test coverage for the new SignalService implementation(#20272)
- Rename the files back to `ISignalService` and `SignalService` removing the Shasta suffix once we remove the old implementation from the codebase

## Design Decisions / Open Questions

- **Removed all hop logic**: This feature wasn't being utilized and added unnecessary complexity
- **State root verification only**: Currently, we only support verifying against a state root. Should we also enable the Inbox and Anchor to submit storage roots of the origin SignalService to reduce verification costs? This would require changes to both the node and prover implementations

---

> [!NOTE]
> Introduce `SignalServiceShasta` and switch Shasta `Inbox`/`Anchor` to external checkpoint persistence, removing in-contract ring buffer and `LibCheckpointStore` with updated configs, deploy scripts, and tests.
> 
> - **Protocol (Shasta)**:
>   - **Signal Service**: Add `ISignalServiceShasta` and implement `SignalServiceShasta` with `saveCheckpoint(Checkpoint)` and `getCheckpoint(blockNumber)`; maintains signal proof verification APIs.
>   - **Checkpoint Storage Migration**:
>     - `Inbox` and `ShastaAnchor` now persist checkpoints via `signalService.saveCheckpoint`, removing `LibCheckpointStore` ring buffer and related storage/APIs.
>     - Update `ICheckpointStore` to support save-by-caller and lookup by `blockNumber`; remove `getLatestCheckpointBlockNumber`/`getNumberOfCheckpoints`.
>   - **Config/Constructors**:
>     - Extend `IInbox.Config` with `signalService` (address); remove `maxCheckpointHistory`.
>     - Update `Inbox`, `ShastaDevnetInbox`, `ShastaMainnetInbox`, `TestInbox*` constructors and `getConfig()` to include `signalService`.
>     - Remove `ICheckpointStore` inheritance and checkpoint fields from `Inbox`/`ShastaAnchor`.
> - **Scripts/Deployment**:
>   - Resolve and pass `"signal_service"` to Shasta inbox deployments/upgrades (`DeployProtocolOnL1.s.sol`, `UpgradeShastaL1.s.sol`).
> - **Tests**:
>   - Replace mock checkpoint store with proxied `SignalServiceShasta`; adjust deployers, helpers, and assertions for new interfaces/args.
> - **Aux**:
>   - Remove `shared/shasta/libs/LibCheckpointStore.sol` and its tests.
>   - Update storage layout, gas reports, and snapshots accordingly.
>   - Minor config/doc cleanups (release-please, README).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 521deef33896cfd7b8ef5774e2022af023fb53a3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
```diff
@@ -0,0 +1,248 @@
...
+    
+    // ---------------------------------------------------------------
+    // New Storage variables
+    // ---------------------------------------------------------------
+
+    /// @notice Storage for checkpoints persisted via the SignalService.
+    /// @dev 2 slots used
+    LibCheckpointStore.Storage private _checkpointStorage;
+
+    uint256[44] private __gap;
```
davidtaikocha: iiuc `LibCheckpointStore.Storage private _checkpointStorage;` will only use `1` slot here?

dantaik:
> Replaced ring buffer with simple mapping for checkpoint storage to enable verification by blockId, which wasn't compatible with the ring buffer design. While storage writes are slightly more expensive, this is offset by the removal of other storage variables and overall contract simplification

Without a ring buffer, how much more gas is used for saving checkpoints? My understanding is, with a ring buffer, we may reuse 3 slots - one for book keeping some pointers, one for block hash, one for state root (3*5000 gas), using maps we need to use 2 new slots (2 * 20000 gas).

packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
ggonzalez94: fixed

ggonzalez94:
> > Replaced ring buffer with simple mapping for checkpoint storage to enable verification by blockId, which wasn't compatible with the ring buffer design. While storage writes are slightly more expensive, this is offset by the removal of other storage variables and overall contract simplification
> 
> Without a ring buffer, how much more gas is used for saving checkpoints? My understanding is, with a ring buffer, we may reuse 3 slots - one for book keeping some pointers, one for block hash, one for state root (3*5000 gas), using maps we need to use 2 new slots (2 * 20000 gas).

Almost - - Before this PR we were doing 4 warm slot access. 3 slots for the `CheckPoint` struct + 1 for the other data in the storage.
After this PR we are doing 2 cold slots (I removed the blockNumber from the `Checkpoint` struct since it is now the key of the mapping)

It is still more expensive, but:
- we need the `blockId` to verify the signals against a root(it is part of the `HopProof` struct). I cannot think of a good way we can get this work with the ring buffer
- `finalize` happens less often, so the gas increase might be acceptable. Verification of signals is more frequent(and usually incurred by users directly), so making that cheaper is important
- the code is cleaner and simpler

One optimization option we have is to store the hash of the `CheckPoint` instead of the individual `stateRoot` and `blockHash`. The problem is that any software that is constructing proofs would need to understand this and adapt

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -89,6 +89,8 @@ interface IInbox {
         /// transition to
         /// finalize the corresponding proposal.
         bytes32 parentTransitionHash;
+        /// @notice The block number associated with the checkpoint.
+        uint48 checkpointBlockNumber;
```
dantaik: what's the benefit of moving out of `Checkpoint` and put it here, avoid 1 SSTORE in saving Checkpoint? If so, I would suggest to keep Checkpoint as is, then in SignalService, in the mapping, introduce and use  `CheckpointRecord` to have only the two SSTORE.

packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
```diff
@@ -0,0 +1,248 @@
...
+import "../../libs/LibNames.sol";
+import "../iface/ICheckpointStore.sol";
+import "../libs/LibCheckpointStore.sol";
+import "../iface/ISignalServiceShasta.sol";
+
+/// @title SignalService
+/// @notice See the documentation in {ISignalService} for more details.
+/// @dev Labeled in address resolver as "signal_service".
+/// @custom:security-contact security@taiko.xyz
+contract SignalService is EssentialContract, ISignalService, ICheckpointStore {
```
dantaik:
```suggestion
contract ShastaSignalService is EssentialContract, ISignalService, ICheckpointStore {
```

packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
```diff
@@ -0,0 +1,248 @@
...
+pragma solidity ^0.8.24;
+
+import "../../common/EssentialContract.sol";
+import "../../libs/LibTrieProof.sol";
+import "../../libs/LibNames.sol";
+import "../iface/ICheckpointStore.sol";
+import "../libs/LibCheckpointStore.sol";
+import "../iface/ISignalServiceShasta.sol";
+
+/// @title SignalService
```
dantaik:
```suggestion
/// @title ShastaSignalService
```

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
```diff
@@ -4,28 +4,17 @@ pragma solidity ^0.8.24;
...
-        /// @dev Maps slot indices (0 to maxHistorySize-1) to checkpoint data
-        mapping(uint48 slot => ICheckpointStore.Checkpoint checkpoint) checkpoints;
-        /// @notice The latest checkpoint number
-        uint48 latestCheckpointBlockNumber;
-        /// @notice The current top of the stack (ring buffer index)
-        uint48 stackTop;
-        /// @notice The current number of items in the stack
-        uint48 stackSize;
+        /// @notice Maps block number to checkpoint data
+        mapping(uint48 blockNumber => ICheckpointStore.Checkpoint checkpoint) checkpoints;
```
dantaik:
```suggestion
        mapping(uint48 blockNumber =>CheckpointRecord checkpoint) checkpoints;
```

dantaik (changes requested):

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
dantaik: See https://github.com/taikoxyz/taiko-mono/pull/20329

dantaik:
@ggonzalez94 let me know once we resolved the merging conflicts so I'll take another look.

ggonzalez94:
> @ggonzalez94 let me know once we resolved the merging conflicts so I'll take another look.

@dantaik resolved all the conflicts with main and applied your suggestion to keep Checkpoint with the 3 fields and use a new struct for storage(`CheckpointRecord`)

ggonzalez94:
@cursor please review the code again

ggonzalez94:
cursor review

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
```diff
@@ -35,104 +32,49 @@ library LibCheckpointStore {
...
-            // Update stack size (capped at maxHistorySize)
-            if (stackSize < _maxCheckpointHistory) {
-                stackSize += 1;
-            }
-        }
-
-        ($.latestCheckpointBlockNumber, $.stackTop, $.stackSize) =
-            (_checkpoint.blockNumber, stackTop, stackSize);
+        $.checkpoints[_checkpoint.blockNumber] =
+            CheckpointRecord({ blockHash: _checkpoint.blockHash, stateRoot: _checkpoint.stateRoot });
```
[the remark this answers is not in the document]
ggonzalez94: this is now intended to be used internally, so it should be fine. As for the increasing blocks I believe they are better validated in the derivation pipeline

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -945,9 +925,7 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
             bytes32 checkpointHash = _hashCheckpoint(_checkpoint);
             require(checkpointHash == _expectedCheckpointHash, CheckpointMismatch());
 
-            LibCheckpointStore.saveCheckpoint(
-                _checkpointStorage, _checkpoint, _maxCheckpointHistory
-            );
+            signalService.saveCheckpoint(_checkpoint);
```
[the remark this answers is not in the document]
dantaik: should we add saveCheckpoint to ISignalService?

packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
```diff
@@ -0,0 +1,242 @@
...
+        returns (bytes32)
+    {
+        return keccak256(abi.encodePacked("SIGNAL", _chainId, _app, _signal));
+    }
+
+    /// @inheritdoc ICheckpointStore
+    function saveCheckpoint(Checkpoint calldata _checkpoint) external override {
+        if (msg.sender != _authorizedSyncer) revert SS_UNAUTHORIZED();
+
+        LibCheckpointStore.saveCheckpoint(_checkpointStorage, _checkpoint);
```
dantaik: Should we emit an `CheckpointSaved` event

packages/protocol/contracts/layer1/shasta/impl/InboxOptimized2.sol
```diff
@@ -114,6 +114,7 @@ contract InboxOptimized2 is InboxOptimized1 {
     }
 
     /// @inheritdoc Inbox
+    /// @dev Uses LibHashing for efficient checkpoint hashing
```
dantaik: remove this line? Seems incorrect/unnecessary

packages/protocol/contracts/shared/shasta/iface/ISignalServiceShasta.sol
```diff
@@ -0,0 +1,96 @@
...
+    /// @notice Verifies if a signal has been received on the target chain.
+    /// @param _chainId The identifier for the source chain from which the
+    /// signal originated.
+    /// @param _app The address that initiated the signal.
+    /// @param _signal The signal (message) to send.
+    /// @param _proof Merkle proof that the signal was persisted on the
+    /// source chain. If this proof is empty, then we check if this signal has been marked as
+    /// received by TaikoL2.
+    /// @return numCacheOps_ The number of newly cached items.
+    function proveSignalReceived(
```
dantaik, line 60: if we do not support caching, then proveSignalReceived can be removed (verifySignalReceived) offers the same functionality without caching.

packages/protocol/contracts/shared/shasta/iface/ISignalServiceShasta.sol
```diff
@@ -0,0 +1,96 @@
...
+/// passing system. It defines methods for sending and verifying signals with
+/// merkle proofs. The trust assumption is that the target chain has secure
+/// access to the merkle root (such as Taiko injects it in the anchor
+/// transaction). With this, verifying a signal is reduced to simply verifying
+/// a merkle proof.
+/// @custom:security-contact security@taiko.xyz
+interface ISignalServiceShasta is ICheckpointStore {
+    /// @dev Proof struct for signal verification
+    /// Maintains the same structure as the original `ISignalService.HopProof` for compatibility
+    struct Proof {
```
dantaik, line 17: Can we keep the original data struct so that we can still use the current bridge relayer and UI, maybe we can further simplify the code later after the Shasta fork. Defer bridge UI/backend change to a later time will make the upgrade happen faster with few dependencies.

packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
```diff
@@ -0,0 +1,242 @@
...
+    // ---------------------------------------------------------------
+
+    /// @dev Deprecated slots used by the old SignalService
+    // - `topBlockId`
+    // - `authorized`
+    uint256[2] private _slotsUsedByPacaya;
+
+    /// @dev Cache for received signals.
+    /// @dev Once written, subsequent verifications can skip the merkle proof validation.
+    mapping(bytes32 signalSlot => bool received) internal _receivedSignals;
```
dantaik, line 48: do we want to remove caching?

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
```diff
@@ -35,104 +32,49 @@ library LibCheckpointStore {
...
-        return $.latestCheckpointBlockNumber;
-    }
-
-    /// @notice Gets the number of checkpoints
-    /// @param $ The storage struct
-    /// @return _ The number of checkpoints
-    function getNumberOfCheckpoints(Storage storage $) public view returns (uint48) {
-        return $.stackSize;
+        CheckpointRecord storage record = $.checkpoints[_blockNumber];
+        if (record.blockHash == bytes32(0)) revert CheckpointNotFound();
```
dantaik: we should use a local var to cache `record.blockHash` as it is used below again.

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
```diff
@@ -4,28 +4,25 @@ pragma solidity ^0.8.24;
 import { ICheckpointStore } from "../iface/ICheckpointStore.sol";
 
 /// @title LibCheckpointStore
-/// @notice Library for managing synced L1 or L2 checkpoints using a ring buffer
-/// @dev This library implements a ring buffer to store the most recent checkpoints.
-/// When the buffer is full, new blocks overwrite the oldest entries. The library
-/// ensures blocks are saved in strictly increasing order by block number.
+/// @notice Library for managing synced L1 or L2 checkpoints
 /// @custom:security-contact security@taiko.xyz
 library LibCheckpointStore {
```
dantaik: This lib is now very small, shall we merge its code to the Signal Service?

dantaik (changes requested):

packages/protocol/contracts/shared/shasta/iface/ISignalServiceShasta.sol
ggonzalez94: I believe caching in the original implementation was to cache the SS storage root. Calling this function still allows you to set the `_receivedSignals` for any signal to true, which makes further verifications much cheaper(you don't need to pass the proof anymore). Please correct me if I'm wrong on this

packages/protocol/contracts/shared/shasta/iface/ISignalServiceShasta.sol
ggonzalez94: you mean the Proof struct? it has the exact same fields as the HopProof struct, just moved to this contract and with the ones we don't use marked as deprecated for clarity. If some client is still using the old struct it should still work, or did I miss something?

packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
ggonzalez94: see this [comment](https://github.com/taikoxyz/taiko-mono/pull/20270#discussion_r2415469270). If we don't want to support this, then yes. But I do believe there may be some use cases for this + some applications may be already relying on the fact that if a signal is verified using `proveSignalReceived` then the next requests can use an empty proof

packages/protocol/contracts/shared/shasta/iface/ISignalServiceShasta.sol
ggonzalez94: also if we remove the function entirely we're breaking compatibility for existing clients right?

packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
ggonzalez94: it is already being emitted from the `LibCheckpointStore`. I'll follow your suggestion below and move the logic to the SS itself

packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol
ggonzalez94: done! AI removed the tests for the LibCheckpointStore. I'll work on adding some for the SS tomorrow

dantaik (approved):

davidtaikocha (approved):

PR #20273, opened 2025-09-26
fix(protocol): improve forced inclusion processing

### LINKED ISSUE #20013: Enable permissionless proposing for forced inclusions after a period of time to achieve stage2
We've heard from the L2Beat team that in order to be classified as stage2 we need to enable permisionless proposal of forced inclusions after some period of time(it is not enough that the proposer is forced to include them, since they could all decide to stop proposing).

I haven't check the written requirements, and feel this might be unnecessary especially when we have decentralized preconfs. This is a requirement that was thought with centralized sequencers in mind IMO.

**If** we have to implement this, we could do it with an additional delay and on the `IProposeChecker` we make proposing permisionless if there's a forced inclusions that has passed that additional delay

dantaik:
## Summary

Processing forced inclusions differently to take care of #20013 

- Reworks `propose` to pull the oldest forced inclusion’s effective timestamp, only enforce `minForcedInclusionCount` when that inclusion is actually overdue
- Skip the proposer checker -- making non-preconfers able to propose -- once the
  the oldest inclusion has more than two delay windows. Note that `endOfSubmissionWindowTimestamp` will be zero in such cases.
- Replaces the external view helper with getOldestForcedInclusionEffectiveTimestamp, exposing the timestamp instead of a boolean so callers can apply their own due checks.


davidtaikocha (dismissed):

dantaik:
@claude please review

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -978,21 +937,98 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+                // input.numForcedInclusions
+                require(
+                    _input.numForcedInclusions >= _minForcedInclusionCount,
+                    UnprocessedForcedInclusionIsDue()
+                );
+            }
+            // If the time elapsed since the oldest forced inclusion's effective timestamp
+            // exceeds twice the forced inclusion delay, any address is allowed to propose.
+            // Otherwise, the permission of msg.sender must be verified.
+            if (block.timestamp <= oldestForcedInclusionTimestamp + _forcedInclusionDelay * 2) {
```
ggonzalez94, line 989: I believe the way this check is constructed is a bug:
Imagine the following scenarion. The forced inclusion queue is empty and there haven't been any forced inclusions for a while. `getOldestInclusionEffectiveTimestamp` then returns `lastProcessedAt`. Once enough time has passed(> 2x _forcedInclusionDelay) anyone would be able to propose. Even if there are no forced inclusions to be processed

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -978,21 +937,98 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+            uint256 oldestForcedInclusionTimestamp =
+                LibForcedInclusion.getOldestInclusionEffectiveTimestamp(_forcedInclusionStorage);
+
+            if (block.timestamp > oldestForcedInclusionTimestamp + _forcedInclusionDelay) {
+                // Ensure the proposer is willing to process at least _minForcedInclusionCount
+                // forced inclusions if the oldest forced inclusion is due.
+                // The actual number of forced inclusions may be less, so we validate against
+                // input.numForcedInclusions
+                require(
+                    _input.numForcedInclusions >= _minForcedInclusionCount,
```
ggonzalez94, line 982: in this case won't we always force him to process at least `_minForcedInclusionCount` if there are **ANY** due?

Let's say : 
- the queue has 10 forced inclusions
- only one is due
- `_minForcedInclusionCount` is set to 8

Then we are forcing the proposer to process at least 8 forced inclusions, even if only one is due

ggonzalez94 (commented):
@dantaik I think this has some underlying issues that I pointed in the comments. I believe an easier and more efficient solution is #20288

PR #20287, opened 2025-09-26
Fix potential token withdrawal vulnerability in TokenUnlock contract

viktorking7:
## Description

**Issue:** The `amountWithdrawable()` function had a logic flaw that could allow withdrawal of tokens that weren't properly vested through the `vest()` function.

**Root Cause:** The original logic `balance.max(locked) - locked` could potentially allow withdrawal of excess tokens if the contract balance exceeded the `amountVested` amount (e.g., from direct token transfers).

**Solution:** 
- Added explicit balance validation against `amountVested` 
- Replaced unclear logic with explicit conditional: `balance > locked ? balance - locked : 0`

**Impact:** Prevents unauthorized token extraction while maintaining all existing functionality.

### DIFF
--- a/packages/protocol/contracts/layer1/team/TokenUnlock.sol
+++ b/packages/protocol/contracts/layer1/team/TokenUnlock.sol
@@ -198,7 +198,14 @@ contract TokenUnlock is EssentialContract {
         uint256 balance = IERC20(taikoToken).balanceOf(address(this));
         uint256 locked = _getAmountLocked();
 
-        return balance.max(locked) - locked;
+        // Ensure we don't allow withdrawal of tokens that weren't properly vested
+        // by limiting balance to the amount that was actually vested
+        uint256 maxWithdrawable = amountVested;
+        if (balance > maxWithdrawable) {
+            balance = maxWithdrawable;
+        }
+
+        return balance > locked ? balance - locked : 0;
     }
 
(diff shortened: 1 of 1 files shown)


ggonzalez94:
Thanks for reporting @viktorking7. I reviewed the issue and I believe it is not an actual problem:

 ## Reproduction Steps

  1. User vests 1000 tokens via vest() → amountVested = 1000
  2. During vesting period, 600 tokens remain locked → locked = 600
  3. Someone accidentally sends 500 tokens directly to the contract → balance = 1500
  4. Formula calculates: max(1500, 600) - 600 = 900 withdrawable
  5. Recipient can withdraw 900 tokens instead of just 400

## Why This is a Non-Issue

  Each TokenUnlock contract is deployed per recipient (isolated instances), meaning:
  - No shared funds between users - impossible to drain other users' tokens
  - Extra tokens must come from external sources (someone's mistake)
  - Recipient withdrawing accidentally sent tokens is not theft

  Verdict: Not a vulnerability. No security risk exists, no protocol funds are threatened, and the "fix" adds unnecessary complexity for an edge case that doesn't compromise the system's integrity.
  
  Please feel free to reopen if I missed anything. Also next time you believe there's a vulnerability please report it to `security@taiko.xyz` instead of opening a public PR. Thanks!

PR #20288, opened 2025-09-26
feat(protocol): enable permisionless proposals after a new forced inclusion deadline

### LINKED ISSUE #20013: Enable permissionless proposing for forced inclusions after a period of time to achieve stage2

ggonzalez94:
Fixes #20013

A different approach to #20273 with the following advantages:
- it does not have some of the underlying issues that the other apporach has(check the comments of that PR)
- no gas increases(it follows the same flow as before, we just make the proposer check conditional on the forced inclusion timestamp)
- less code changes
 

---

> [!NOTE]
> Allow anyone to propose when the oldest forced inclusion exceeds delay*multiplier, refactor forced inclusion handling into unified sources, extend config, and add comprehensive tests/snapshots.
> 
> - **Inbox core (`impl/Inbox.sol`)**:
>   - Make proposer check conditional: if oldest forced inclusion timestamp + `forcedInclusionDelay * permissionlessInclusionMultiplier` < now, proposing is permissionless (submission window = 0); otherwise enforce `ProposerChecker`.
>   - Refactor proposal source building into `_buildDerivationSources`, returning combined forced-inclusion + normal sources and validating due/attempted counts.
>   - Include `permissionlessInclusionMultiplier` in constructor init and `getConfig`.
> - **Forced inclusion library (`libs/LibForcedInclusion.sol`)**:
>   - Change `consumeForcedInclusions` to return `DerivationSource[]` (with extra slot for normal source), remaining queue size, and oldest processed timestamp; accumulate fees and update `lastProcessedAt`.
>   - Keep `isOldestForcedInclusionDue` logic using `lastProcessedAt`.
> - **Interface/config (`iface/IInbox.sol`)**:
>   - Add `permissionlessInclusionMultiplier` to `Config` and expose via `getConfig`.
> - **Deployments**:
>   - Set `permissionlessInclusionMultiplier: 5` in `ShastaDevnetInbox` and `ShastaMainnetInbox` constructors.
> - **Tests (`AbstractPropose.t.sol` et al.)**:
>   - Add tests for single/multiple forced inclusions, reverting when due FI not processed, permissionless proposing when FI too old, and unauthorized cases.
> - **Snapshots**:
>   - Update `shasta-propose.json` and `shasta-prove.json` with new entries and minor value adjustments.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e42ebf1cc689d973407cbda11cff14ac430a8167. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -800,6 +779,60 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+        // Add forced inclusions as derivation sources
+        uint256 sourceCount = 1;
+        IForcedInclusionStore.ForcedInclusion[] memory forcedInclusions;
+
+        if (_input.numForcedInclusions > 0) {
+            forcedInclusions = LibForcedInclusion.consumeForcedInclusions(
+                _forcedInclusionStorage, msg.sender, _input.numForcedInclusions
+            );
+            sourceCount += forcedInclusions.length;
+            oldestForcedInclusionTimestamp = forcedInclusions[0].blobSlice.timestamp;
```
dantaik: I think we need to use the last processed timestamp from the inclusion store here:

oldestForcedInclusionTimestamp = oldestForcedInclusionTimestamp.max(lastProcesstedAt)

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -800,6 +779,60 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+
+        // Verify that at least `minForcedInclusionCount` forced inclusions were attempted to be processed
+        // or none in the queue is due.
+        require(
+            (_input.numForcedInclusions >= _minForcedInclusionCount)
+                || !LibForcedInclusion.isOldestForcedInclusionDue(
+                    _forcedInclusionStorage, _forcedInclusionDelay
+                ),
+            UnprocessedForcedInclusionIsDue()
+        );
```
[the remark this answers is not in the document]
dantaik: I think this is a real bug. 

Also, maybe we can use bulk copy, with assembly, when we copy all returned inclusions objects from the lib to the array with one extra object.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
ggonzalez94: fixed

ggonzalez94:
> I think this is a real bug.
> 
> Also, maybe we can use bulk copy, with assembly, when we copy all returned inclusions objects from the lib to the array with one extra object.

@dantaik can you elaborate which is the bug? from the code selected it is not clear. Maybe it is from a previous version of the PR?
Also where should we do a bulk copy? on the `consumeForcedInclusions` function of the lib or somewhere else? Anyways most of the times I expect there will be zero forced inclusions or very few(as it is the case for most rollups), so not sure how much we should optimize for the case where there are multiple

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
```diff
@@ -40,6 +40,9 @@ interface IInbox {
         uint64 forcedInclusionFeeInGwei;
         /// @notice The maximum number of checkpoints to store in ring buffer
         uint16 maxCheckpointHistory;
+        /// @notice The multiplier to determine when a forced inclusion is too old so that proposing
+        /// becomes permissionless
+        uint8 permissionlessInclusionMultiplier;
```
dantaik, line 45: how about we introduce a 'permissionlessProposalDelay`, to avoid (_forcedInclusionDelay * _permissionlessInclusionMultiplier) in `propose`

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -800,6 +781,63 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
     }
 
+    /// @dev Creates derivation sources array from forced inclusions and regular proposal
+    /// @notice Processes forced inclusions and creates a unified sources array for derivation
+    /// @param _input The ProposeInput containing numForcedInclusions and blobReference
+    /// @return sources Array of derivation sources with forced inclusions first, then regular
+    /// proposal
+    /// @return oldestForcedInclusionTimestamp The timestamp of the oldest forced inclusion that was
+    /// processed. type(uint48).max if there are no forced inclusions.
+    function _createDerivationSources(ProposeInput memory _input)
```
dantaik: "buildDerivationSources" seems to be a better name.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -800,6 +781,63 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+        // We need to capture the last processed timestamp before consuming the forced inclusions
+        uint48 lastProcessedAt = _forcedInclusionStorage.lastProcessedAt;
+
+        if (_input.numForcedInclusions > 0) {
+            forcedInclusions = LibForcedInclusion.consumeForcedInclusions(
+                _forcedInclusionStorage, msg.sender, _input.numForcedInclusions
+            );
+            sourceCount += forcedInclusions.length;
+            oldestForcedInclusionTimestamp =
+                uint48(forcedInclusions[0].blobSlice.timestamp.max(lastProcessedAt));
```
dantaik: `forcedInclusions[0].blobSlice.timestamp` is sload once here then again in line 818, maybe we can avoid it?

dantaik: In general I think the forced inclusion storage sload should be optimized to make sure it is only load once. Now in the propose function, up to 3 SLOAD is needed:

1. `_forcedInclusionStorage.lastProcessedAt`
2. `LibForcedInclusion.consumeForcedInclusions`
3.  `LibForcedInclusion.isOldestForcedInclusionDue`

This may need us to move some code from this function to LibForcedInclusion so only 1 sload and 1 store is needed.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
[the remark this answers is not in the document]
dantaik: Reviewed the code again, false alarm.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -800,6 +781,63 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+                ),
+            UnprocessedForcedInclusionIsDue()
+        );
+
+        // Create sources array and populate it
+        sources = new DerivationSource[](sourceCount);
+
+        uint256 index;
+        // Add forced inclusion sources first
+        for (; index < forcedInclusions.length; ++index) {
```
dantaik: To avoid memory copying, we can make `consumeForcedInclusions` always return array with an extra empty slot. Then assign the normal source to that last element.

dantaik (changes requested):

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -800,6 +781,63 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+            );
+            sourceCount += forcedInclusions.length;
+            oldestForcedInclusionTimestamp =
+                uint48(forcedInclusions[0].blobSlice.timestamp.max(lastProcessedAt));
+        } else {
+            oldestForcedInclusionTimestamp = type(uint48).max;
+        }
+
+        // Verify that at least `minForcedInclusionCount` forced inclusions were attempted to be
+        // processed
```
dantaik, line 774: one line

packages/protocol/contracts/layer1/shasta/iface/IInbox.sol
ggonzalez94: I thought about it. And while I don't have a strong opinion about this the reason I stuck with the multiplier is to make sure that the permisionless delay is always greater than the forcedInclusionDelay. Otherwise if we set these parameters mistakenly they would not make sense.
But if you feel strongly about using permissionlessProposalDelay I'm fine with it as well

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
ggonzalez94: > forcedInclusions[0].blobSlice.timestamp

This is coming from memory, so I don't think it is a problem

> In general I think the forced inclusion storage sload should be optimized to make sure it is only load once. Now in the propose function, up to 3 SLOAD is needed:

I've played around with this a bit and the reason I stuck with this is the following:
1. I tried moving `_forcedInclusionStorage.lastProcessedAt` to the library and the gas cost stayed the same(actually it increased slightly). The compiler is smart enough to optimize this, and we keep it readable.
2. The reason for having `consumeForcedInclusions ` and `isOldestForcedInclusionDue ` serves to purposes:
    1. In the most common case(no forced inclusions) you only do an `SLOAD` in `isOldestForcedInclusionDue ` (`consumeForcedInclusions` never gets called since the proposer passes zero for forced inclusions)
    2. In the case where the proposer processes forced inclusions  `consumeForcedInclusions ` is called AND `isOldestForcedInclusionDue` can be avoided by passing a large number to `numForcedInclusions`. Now I've pushed an optimization where if there are zero forced inclusions remaining after processing we also skip this check. This way if the proposer clears queue, they don't need to pay the cost of the additional `SLOAD`
    
   There may be more efficient ways to handle this, but this seems to be a good balance and work well IMO.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
ggonzalez94: Returning DerivationSources with the length you suggested should be the most optimal(and cleaner).

I tried it on this [PR](https://github.com/taikoxyz/taiko-mono/pull/20310), but it is actually slightly more expensive. The compiler seems to be optimizing the current approach better. We might still prefer this for readability.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -215,51 +218,28 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+            // endOfSubmissionWindowTimestamp = 0).
+            // Otherwise, only the current preconfer can propose.
+            uint48 endOfSubmissionWindowTimestamp;
+            if (
+                block.timestamp
+                    <= oldestForcedInclusionTimestamp
+                        + _forcedInclusionDelay * _permissionlessInclusionMultiplier
+            ) {
+                endOfSubmissionWindowTimestamp =
+                    _proposerChecker.checkProposer(msg.sender, _lookahead);
```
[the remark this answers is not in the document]
ggonzalez94: not true. This is implicitly converted to uint256 when comparing it to `block.timestamp`. We might still want to choose a different approach to avoid future issues since we are running on an unchecked block

dantaik (approved):

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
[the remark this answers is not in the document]
ggonzalez94: This actually ended up being an issue. Resolved in the latest commit

PR #20297, opened 2025-09-28
fix(protocol): fix recall context check to validate source vault address

viktorking7:
- Summary: Corrects checkRecallMessageContext to validate that ctx.from equals this vault’s resolved address on the source chain. Aligns with checkProcessMessageContext. Adds brief @dev docs.
- Motivation: Previously compared ctx.from with msg.sender (bridge on dest chain), causing unintended reverts and blocking recalls.
- Changes:
  - Replace comparison with resolve(ctx.srcChainId, name(), false) in BaseVault.sol
  - Add @dev comments to both context-check helpers
- Impact: Affects recall flows in ERC20/721/1155 vaults; restores intended behavior.

### DIFF
--- a/packages/protocol/contracts/shared/tokenvault/BaseVault.sol
+++ b/packages/protocol/contracts/shared/tokenvault/BaseVault.sol
@@ -72,7 +72,7 @@ abstract contract BaseVault is
         returns (IBridge.Context memory ctx_)
     {
         ctx_ = IBridge(msg.sender).context();
-        address selfOnSourceChain = resolve(ctx_.srcChainId, name(), false);
+        address selfOnSourceChain = resolve(ctx_.srcChainId, name(), LibNames.B_BRIDGE);
         if (ctx_.from != selfOnSourceChain) revert VAULT_PERMISSION_DENIED();
     }
 
--- a/packages/protocol/contracts/shared/tokenvault/BaseVault.sol
+++ b/packages/protocol/contracts/shared/tokenvault/BaseVault.sol
@@ -50,6 +50,8 @@ abstract contract BaseVault is
     /// @return The name of the vault.
     function name() public pure virtual returns (bytes32);
 
+    /// @dev Ensures the message was sent by this vault on the source chain via Bridge.
+    /// Resolves this vault's address on the source chain and compares with ctx.from.
     function checkProcessMessageContext()
         internal
         view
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/shared/tokenvault/BaseVault.sol
```diff
@@ -61,14 +63,17 @@ abstract contract BaseVault is
...
+    /// Resolves this vault's address on the source chain and compares with ctx.from.
     function checkRecallMessageContext()
         internal
         view
         onlyFromNamed(LibNames.B_BRIDGE)
         returns (IBridge.Context memory ctx_)
     {
         ctx_ = IBridge(msg.sender).context();
-        if (ctx_.from != msg.sender) revert VAULT_PERMISSION_DENIED();
+        address selfOnSourceChain = resolve(ctx_.srcChainId, name(), false);
```
dantaik: Seems like a bug: but when a message is recalled, the `ctx.from` should be the bridge contract on the src chain.  Not the vault. See the call of _storeContext.

Image

If after your modification, all tests pass, then that means we need to add test coverage for recalling a message with a Vault.

dantaik:
```suggestion
        address selfOnSourceChain = resolve(ctx_.srcChainId, name(), LibNames.B_BRIDGE);
```

dantaik: Could you add a test please?

packages/protocol/contracts/shared/tokenvault/BaseVault.sol
```diff
@@ -61,14 +63,17 @@ abstract contract BaseVault is
...
+    /// Resolves this vault's address on the source chain and compares with ctx.from.
     function checkRecallMessageContext()
         internal
         view
         onlyFromNamed(LibNames.B_BRIDGE)
         returns (IBridge.Context memory ctx_)
     {
         ctx_ = IBridge(msg.sender).context();
-        if (ctx_.from != msg.sender) revert VAULT_PERMISSION_DENIED();
+        address selfOnSourceChain = resolve(ctx_.srcChainId, name(), LibNames.B_BRIDGE);
```
dantaik, line 75: The original code is not buggy but redundant, please help review this PR: https://github.com/taikoxyz/taiko-mono/pull/20324

dantaik (changes requested):

PR #20304, opened 2025-09-29
feat(protocol): add preconf slasher for shasta

AnshuJalan:
The two-component (L1 and L2) preconf slashing solution that slashes a preconfer for Liveness and Safety faults:
- The slashing conditions are verified on L2.
- The fault is propagated to L1 via the native message passing system.
- The preconfer's collateral in the URC is slashed.

---

> [!NOTE]
> Adds an L1/L2 preconf slashing system with lookahead store/slasher, whitelist/blacklist, Anchor updates, shared interfaces/libs, and tests.
> 
> - **Preconf Slashing (L1)**:
>   - Implement `impl/PreconfSlasherL1.sol` and `impl/UnifiedSlasher.sol` with corresponding interfaces (`iface/IPreconfSlasherL1.sol`, `shared/preconf/IPreconfSlasher.sol`).
>   - Add access controls: `impl/PreconfWhitelist.sol`, `impl/Blacklist.sol` with interfaces (`iface/IPreconfWhitelist.sol`, `iface/IBlacklist.sol`).
>   - Add lookahead components: `impl/LookaheadStore.sol`, `impl/LookaheadSlasher.sol` with interfaces (`iface/ILookaheadStore.sol`, `iface/ILookaheadSlasher.sol`).
>   - Introduce constants in `libs/LibPreconfConstants.sol`.
> - **Preconf Slashing (L2)**:
>   - Add `preconf/PreconfSlasherL2.sol` with interface and layout (`IPreconfSlasherL2.sol`, `PreconfSlasherL2_Layout.sol`).
>   - Update core anchor: `core/Anchor.sol`, `core/Anchor_Layout.sol` for integration.
> - **Shared/Infra**:
>   - Add `shared/libs/LibNames.sol` and `script/gen-layouts.sh`.
> - **Tests**:
>   - New tests for lookahead store and whitelist (`LookaheadStore_PostedByProtectorOrWhitelistedPreconfer.t.sol`, `PreconfWhitelist.t.sol`) and Anchor on L2 (`Anchor.t.sol`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5682eafc4d39895080b6fbc0cced04491ffb1e96. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -152,7 +160,8 @@ abstract contract ShastaAnchor is PacayaAnchor, ICheckpointStore {
         uint48 _anchorBlockNumber,
         bytes32 _anchorBlockHash,
         bytes32 _anchorStateRoot,
-        uint48 _endOfSubmissionWindowTimestamp
+        uint48 _submissionWindowEnd,
+        bytes32 _rawTxListHash
```
linoscope: Let's also update the `DERIVATION.md` file to add these params. So the derivation has to calculate the `rawTxListHash` before doing any filtering, then pass it into the anchor tx.

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -201,10 +210,15 @@ abstract contract ShastaAnchor is PacayaAnchor, ICheckpointStore {
...
-        newState_.endOfSubmissionWindowTimestamp = _endOfSubmissionWindowTimestamp;
+        newState_.submissionWindowEnd = _submissionWindowEnd;
         _state = newState_;
 
-        blockIdToEndOfSubmissionWindowTimeStamp[block.number] = _endOfSubmissionWindowTimestamp;
+        blockIdToPreconfMeta[block.number] = PreconfMeta({
+            anchorBlockNumber: _anchorBlockNumber,
+            submissionWindowEnd: _submissionWindowEnd,
+            rawTxListHash: _rawTxListHash,
+            parentRawTxListHash: blockIdToPreconfMeta[block.number - 1].rawTxListHash
```
linoscope: If `blockIdToPreconfMeta[block.number].parentRawTxListHash` is always `blockIdToPreconfMeta[block.number - 1].rawTxListHash`, do we need  to store `parentRawTxListHash` in `PreconfMeta`? Is it to simplify things in slashing?

linoscope (commented):
Haven't deep-dived into the slashing part yet. Will do once some fixes discussed with @AnshuJalan offline lands.

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
AnshuJalan: > Is it to simplify things in slashing?

Yes. This allows for getting both current and parent hashes in a single call.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
```diff
@@ -0,0 +1,87 @@
...
+    {
+        // Verify calling context
+        require(_challenger == address(this), ChallengerIsNotSelf());
+        require(msg.sender == urc, CallerIsNotURC());
+
+        IPreconfSlasherL2.Fault fault = abi.decode(_evidence, (IPreconfSlasherL2.Fault));
+        IPreconfSlasherL2.Preconfirmation memory preconfirmation =
+            abi.decode(_commitment.payload, (IPreconfSlasherL2.Preconfirmation));
+
+        // Liveness faults are not slashed if the proposer's submission slot was potentially reorged
```
linoscope: There is misalignment in the definition of "liveness fault" here.

Liveness fault means the preconfer may have attempted their best to not invalidate preconf, but they had to because of the L1 missed slot.

For example:
- Suppose preconfer missed his submission (`preconf.submissionWindowEnd` < `_preconfMeta.submissionWindowEnd`). If that preconfer missed his slot in L1, that is a liveness fault. If that preconfer did NOT miss his slot in L1, that is a safety fault (because he should have had the chance to submit his preconfs, but did not).
- Suppose preconfer missed his EOP submission. If that preconfer missed his slot in L1, that is a liveness fault. If that preconfer did NOT miss his slot in L1, that is a safety fault.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
```diff
@@ -0,0 +1,87 @@
...
+            require(
+                LibPreconfUtils.getBeaconBlockRootAt(preconfirmation.submissionWindowEnd)
+                    != bytes32(0),
+                MissedSlot()
+            );
+        }
+
+        IPreconfSlasher.SlashAmount memory amount = getSlashAmount();
+        return
+            (fault == IPreconfSlasherL2.Fault.Liveness) ? amount.livenessFault : amount.safetyFault;
```
linoscope: To continue from the [previous comment](https://github.com/taikoxyz/taiko-mono/pull/20304/files#r2395780707), here, the amount should always be `amount.safetyFault`, as we know the preconfer's L1 slot is not missed at this point.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
AnshuJalan: Fixed in the latest commit

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
AnshuJalan: In the latest commit, I change the cached liveness fault to a safety fault if a slot is not missed, so we may keep this line

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
```diff
@@ -0,0 +1,90 @@
...
+
+        // If the fault computed on the L2 side is a liveness fault,
+        if (fault == IPreconfSlasherL2.Fault.Liveness) {
+            // but the preconfer has not missed its L1 proposal slot, then we actually have
+            // a safety fault.
+            if (
+                LibPreconfUtils.getBeaconBlockRootAt(preconfirmation.submissionWindowEnd)
+                    != bytes32(0)
+            ) {
+                fault = IPreconfSlasherL2.Fault.Safety;
```
linoscope: Seems confusing that the thing we call `Liveness` fault in L2 contract can actually be a safety fault. Maybe call it something different on the L2 side? For example, `PotentialLivenessFault`?

AnshuJalan: Makes sense

AnshuJalan:
Claude review:
- 1, 3, 8: Invalid
- 4, 6, 7: Not a priority
- 2: Validation already done in the URC
- 5: Not required

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasherL1.sol
```diff
@@ -0,0 +1,25 @@
...
+pragma solidity ^0.8.24;
+
+import "@eth-fabric/urc/ISlasher.sol";
+import "src/shared/bridge/IBridge.sol";
+
+/// @title IPreconfSlasherL1
+/// @dev This contract inherits from ISlasher which contains the definition for the `slash` function
+/// that is called by the URC.
+/// @custom:security-contact security@taiko.xyz
+interface IPreconfSlasher is ISlasher, IMessageInvocable {
```
MatusKysel: I don't think we should use this name as we already have one here https://github.com/taikoxyz/taiko-mono/blob/preconf_slasher_shasta/packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasher.sol#L10 and it's confusing, maybe `IPreconfSlasherL1` should be the name of this one ?

ggonzalez94:
@codex please review this PR

packages/protocol/contracts/layer2/preconf/IPreconfSlasherL2.sol
```diff
@@ -0,0 +1,73 @@
...
+    // The slashing reason forwarded to the L1 preconfirmation slasher
+    enum Fault {
+        // Every "liveness fault" computed on L2 is a "potential" liveness fault.
+        // It is the L1 contract that further confirms it.
+        PotentialLiveness,
+        Liveness,
+        Safety
+    }
+
+    // Note: `Commitment` and `SignedCommitment` are extracted here from URC's ISlasher,
```
ggonzalez94: nit: move this to the a ///@dev comment on the Commitment struct?

packages/protocol/contracts/layer2/preconf/IPreconfSlasherL2.sol
```diff
@@ -0,0 +1,73 @@
...
+
+    error InvalidEOPFlag();
+    error NotALivenessFault();
+    error NotASafetyFault();
+    error ParentRawTxListHashMismatch();
+    error ParentSubmissionWindowEndMismatch();
+    error UnexpectedExtraProposalsInPreviousWindow();
+
+    /// @notice Validates if a preconfirmation is slashable and forwards the fault to the
+    /// L1 preconfirmation slasher.
```
ggonzalez94, line 41: nit for clarity:
```suggestion
    /// @notice Validates if a preconfirmation is slashable and forwards the fault to the
    /// L1 preconfirmation slasher.
    /// @dev Safety faults can be immediately validated on L2, while liveness faults are only validated on L1 after the message is sent, since the L2 has no way of knowing if the preconfer missed their slot or not.
```

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,181 @@
...
+        );
+        require(
+            preconfirmation.parentSubmissionWindowEnd == preconfMeta.parentSubmissionWindowEnd,
+            ParentSubmissionWindowEndMismatch()
+        );
+
+        if (_fault == Fault.Liveness) {
+            _validateLivenessFault(preconfirmation, preconfMeta);
+        } else {
+            _validateSafetyFault(preconfirmation, preconfMeta);
```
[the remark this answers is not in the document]
ggonzalez94: @AnshuJalan  I believe this is spot on. We may want to remove the `PotentialLiveness` issue altogether from the enum?

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,181 @@
...
+        external
+    {
+        Preconfirmation memory preconfirmation =
+            abi.decode(_signedCommitment.commitment.payload, (Preconfirmation));
+
+        // Pull the preconf metadata for the block proposed at the preconfirmation height
+        ShastaAnchor.PreconfMeta memory preconfMeta =
+            ShastaAnchor(taikoAnchor).getPreconfMeta(preconfirmation.blockNumber);
+
+        // The parent preconfirmation must not have been messed up during submission
```
ggonzalez94: maybe this is more clear?
```suggestion
        // Make sure the parent preconfirmation was not violated before we slash
```

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,181 @@
...
+            ShastaAnchor.PreconfMeta memory nextPreconfMeta =
+                ShastaAnchor(taikoAnchor).getPreconfMeta(_preconfirmation.blockNumber + 1);
+            require(
+                nextPreconfMeta.submissionWindowEnd > _preconfirmation.submissionWindowEnd,
+                NotALivenessFault()
+            );
+        }
+    }
+
+    function _validateSafetyFault(
```
ggonzalez94: having natspec here would help:
```suggestion
    /// @dev Validates that the claimed fault qualifies as a safety fault. Reverts if not.
    /// @dev Checks that timing is correct (same window) but content was violated
    /// (wrong tx list, wrong anchor, or incorrect EOP usage).
    /// @param _preconfirmation The preconfirmation commitment being challenged.
    /// @param _preconfMeta The actual metadata of the block that was submitted.
    function _validateSafetyFault(
```

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,181 @@
...
+            // Either an EOP-only preconfirmation is violated or an extra block is submitted
+            // in the same window
+            ShastaAnchor.PreconfMeta memory nextPreconfMeta =
+                ShastaAnchor(taikoAnchor).getPreconfMeta(_preconfirmation.blockNumber + 1);
+            require(
+                _preconfirmation.rawTxListHash == bytes32(0)
+                    || _preconfirmation.submissionWindowEnd == nextPreconfMeta.submissionWindowEnd,
+                NotASafetyFault()
+            );
+        }
```
ggonzalez94: I believe we can simplify a bit by returning early and adding some comments. Please check if my understanding is correct
```suggestion
         // Content mismatch is a clear safety fault
        if (
            _preconfirmation.rawTxListHash != _preconfMeta.rawTxListHash
                || uint48(_preconfirmation.anchorBlockNumber) != _preconfMeta.anchorBlockNumber
        ) {
            return;
        }

        // Content matches - only EOP misuse can be a safety fault here
        require(_preconfirmation.eop, InvalidEOPFlag());

        ShastaAnchor.PreconfMeta memory nextPreconfMeta =
            ShastaAnchor(taikoAnchor).getPreconfMeta(_preconfirmation.blockNumber + 1);

        // Two EOP-related safety faults:
        // 1. EOP-only submission (rawTxListHash == 0): Submitting an empty block is considered
        //    censorship since the preconfer could have included transactions
        // 2. Invalid EOP (next block in same window): Preconfer incorrectly marked EOP but
        //    continued submitting blocks in the same window
        require(
            _preconfirmation.rawTxListHash == bytes32(0)
                || _preconfirmation.submissionWindowEnd == nextPreconfMeta.submissionWindowEnd,
            NotASafetyFault()
        );
```

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,181 @@
...
+            preconfirmation.parentRawTxListHash == preconfMeta.parentRawTxListHash,
+            ParentRawTxListHashMismatch()
+        );
+        require(
+            preconfirmation.parentSubmissionWindowEnd == preconfMeta.parentSubmissionWindowEnd,
+            ParentSubmissionWindowEndMismatch()
+        );
+
+        if (_fault == Fault.Liveness) {
+            _validateLivenessFault(preconfirmation, preconfMeta);
```
ggonzalez94: maybe I'm missing something here. But wasn't it the case that liveness faults cannot be validated on L2 and should be directly forwarded to L1 for validation?

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,181 @@
...
+        SignedCommitment memory _signedCommitment
+    )
+        internal
+    {
+        bytes memory callData = abi.encodeWithSelector(
+            IMessageInvocable.onMessageInvocation.selector,
+            abi.encode(_fault, _registrationRoot, _signedCommitment)
+        );
+
+        IBridge.Message memory message = IBridge.Message({
```
ggonzalez94, line 209: since we are not sending any tokens I wonder if we should use the SignalService instead of the bridge? it might be more lightweight and cheaper later to verify on L1?

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
```diff
@@ -0,0 +1,91 @@
...
+        address, /* committer */
+        bytes calldata _evidence,
+        address _challenger
+    )
+        external
+        view
+        returns (uint256)
+    {
+        // Verify calling context
+        require(_challenger == address(this), ChallengerIsNotSelf());
```
ggonzalez94, line 39: I'm not too familar with the URC. What is the challenger? why do we need this check?

packages/protocol/contracts/layer2/preconf/IPreconfSlasherL2.sol
```diff
@@ -0,0 +1,73 @@
...
+        // Keccak256 hash of the raw list of transactions included in the block
+        bytes32 rawTxListHash;
+        // The expected submission window of the parent block
+        uint256 parentSubmissionWindowEnd;
+        // The timestamp of the preconfer's slot in the lookahead
+        uint256 submissionWindowEnd;
+    }
+
+    // The slashing reason forwarded to the L1 preconfirmation slasher
+    enum Fault {
```
ggonzalez94: I see we are using the `PotentialLiveness` fault as a way to signal to the L1 slasher that they should validate if it is a safety fault or not. Since the actor sending the slashing command should know by looking off-chain what kind of fault it is I believe it is better to just keep two types of faults and run the necessary validations?

packages/protocol/contracts/layer1/preconf/iface/IPreconfSlasherL1.sol
ggonzalez94: yeah... this is confusing. Is the PreconfSlasher an old one we cna remove and only keep PreconfSlasherL1 and L2?

ggonzalez94 (commented):
Looking good... left a few comments

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -252,6 +260,14 @@ abstract contract ShastaAnchor is PacayaAnchor, ICheckpointStore {
         return LibCheckpointStore.getNumberOfCheckpoints(_checkpointStorage);
     }
 
+    /// @notice Returns preconfirmation metadata for a given L2 block id.
+    /// @param _blockId The L2 block id.
+    /// @return meta_ The preconfirmation metadata containing submissionWindowEnd and rawTxListHash.
+    function getPreconfMeta(uint256 _blockId) external view returns (PreconfMeta memory meta_) {
+        meta_ = blockIdToPreconfMeta[_blockId];
+        require(meta_.anchorBlockNumber != 0, InvalidBlockId());
```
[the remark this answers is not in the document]
ggonzalez94: this also seems correct?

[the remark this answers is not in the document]
AnshuJalan: > The commit keeps allowing callers of updateState

This is not possible, since this calls to this functions are zk-proven

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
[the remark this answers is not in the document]
AnshuJalan: Right, we did not have `PotentialLiveness` fault in the first run. I think we should have a chat on what the core reason behind introducing this is

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
AnshuJalan: Using the bridge allows us to use the full set of existing relayer tooling (that is tried and tested). Since slashings are rare, the overhead of using the Bridge is fine imo compared to the overhead of developing a custom relayer.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
AnshuJalan: Challenger is the entity who initiated the slashing in the URC

AnshuJalan:
@codex do a re review after changes in last commit

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -70,13 +96,13 @@ abstract contract ShastaAnchor is PacayaAnchor, ICheckpointStore {
...
 
-    mapping(uint256 blockId => uint256 endOfSubmissionWindowTimestamp) public
-        blockIdToEndOfSubmissionWindowTimeStamp;
-
     /// @dev Storage for checkpoint management
     /// @dev 2 slots used
     LibCheckpointStore.Storage internal _checkpointStorage;
 
+    /// @notice Maps L2 block id to preconfirmation metadata.
+    mapping(uint256 blockId => PreconfMeta meta) public blockIdToPreconfMeta;
```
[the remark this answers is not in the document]
AnshuJalan: Not an issue as this is not deployed yet.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
```diff
@@ -0,0 +1,93 @@
...
+                || fault == IPreconfSlasherL2.Fault.MissingEOP
+        ) {
+            // If the preconfer has missed its L1 slot, these faults are classified under liveness
+            // faults, and incur a smaller penalty.
+            if (
+                LibPreconfUtils.getBeaconBlockRootAt(preconfirmation.submissionWindowEnd)
+                    != bytes32(0)
+            ) {
+                return slashAmount.livenessFault;
+            }
```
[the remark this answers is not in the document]
AnshuJalan: Incorrect understanding of `getBeaconBlockRootAt `

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,225 @@
...
+            preconfirmation.parentRawTxListHash == preconfMeta.parentRawTxListHash,
+            ParentRawTxListHashMismatch()
+        );
+
+        // Protection against a scenario where a previous preconfer submitted an extra block
+        // with the same `rawTxListHash` as the one preconfirmed by the current preconfer.
+        require(
+            preconfirmation.parentSubmissionWindowEnd == preconfMeta.parentSubmissionWindowEnd,
+            ParentSubmissionWindowEndMismatch()
+        );
```
linoscope: I _think_ these two checks can be moved within `_validateRawTxListHashOrAnchorBlockMismatchFault`.

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,225 @@
...
+
+        // Protection against a scenario where a previous preconfer submitted an extra block
+        // that it never preconfirmed.
+        require(
+            preconfirmation.parentRawTxListHash == preconfMeta.parentRawTxListHash,
+            ParentRawTxListHashMismatch()
+        );
+
+        // Protection against a scenario where a previous preconfer submitted an extra block
+        // with the same `rawTxListHash` as the one preconfirmed by the current preconfer.
```
linoscope: Think we can remove this comment, it's pretty subtle, and actually, it is already covered in the previous comment:
```
        // Protection against a scenario where a previous preconfer submitted an extra block
        // that it never preconfirmed.
```

The `parentSubmissionWindowEnd` is to protect against the "scenario where a previous preconfer submitted an extra block" case in a robust way.

AnshuJalan: Tbh, this comment is more for my own understanding when revisiting the code in the future. I almost forgot why we added this check, and went back to our DMs to understand the scenario

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
AnshuJalan: I believe, yes. You are right. These are no longer required to be top-level checks with the fault separation

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,225 @@
...
+            _validateInvalidEOPFault(preconfirmation, preconfMeta);
+        }
+
+        _invokePreconfSlasherL1(_fault, _registrationRoot, _signedCommitment);
+    }
+
+    // Internal functions
+    // ---------------------------------------------------------------
+
+    /// @dev Validates that a preconfirmation was never submitted to the inbox.
```
ggonzalez94, line 63: Is this comment accurate? I don't think on this contract we are validating this or **can** validate this

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,225 @@
...
+    {
+        // EOP-only preconfirmations are not expected to be submitted
+        require(
+            !_isEOPOnlyPreconfirmation(_preconfirmation),
+            EOPOnlyPreconfirmationDoesNotRequireSubmission()
+        );
+
+        // The block at preconfirmation height must be submitted in a future submission window
+        require(
+            _preconfMeta.submissionWindowEnd > _preconfirmation.submissionWindowEnd,
```
ggonzalez94, line 79: question: If the preconfer did not submit any preconfed block is there a preconfirmation object we can run checks against?

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,225 @@
...
+            UnexpectedExtraProposalsInPreviousWindow()
+        );
+
+        // Confirm that the block with missing EOP is the last block in it's
+        // submission window
+        ShastaAnchor.PreconfMeta memory nextPreconfMeta =
+            ShastaAnchor(taikoAnchor).getPreconfMeta(_preconfirmation.blockNumber + 1);
+        require(
+            nextPreconfMeta.submissionWindowEnd > _preconfirmation.submissionWindowEnd,
+            NotALivenessFault()
```
ggonzalez94: Shall we update this error now that we have more specific Faults in the enum?

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,225 @@
...
+        view
+    {
+        // Protection against a scenario where a previous preconfer submitted an extra block
+        // that it never preconfirmed.
+        require(
+            _preconfirmation.parentRawTxListHash == _preconfMeta.parentRawTxListHash,
+            ParentRawTxListHashMismatch()
+        );
+
+        // Protection against a scenario where a previous preconfer submitted an extra block
```
ggonzalez94, line 129: Is this comment accurate? it seems we are only validating below that the preconfirmation and the submission have the same parent submission window end, not the content of the preconfirmation

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,225 @@
...
+        // potential `MissedSubmissionFault`
+        require(
+            _preconfirmation.submissionWindowEnd == _preconfMeta.submissionWindowEnd,
+            SubmissionWindowMismatch()
+        );
+
+        // EOP flag must be present
+        require(_preconfirmation.eop, InvalidEOPFlag());
+
+        // If it's an EOP-only preconfirmation, we already have enough evidence for a violation
```
ggonzalez94, line 168: why should issuing an EOP only preconf pass this validation and be considered a potential fault?

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        return slashAmount.safetyFault;
+    }
+
+    /// @dev Invoked by the L2 preconf slasher
+    function onMessageInvocation(bytes calldata _data) external payable {
```
ggonzalez94, line 70: shouldn't this be part of the interface given it is invoked externally by the L2 slasher(indirectly via the bridge)

ggonzalez94 (commented):
This is looking great. I left a few last comments and questions

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /// @dev EOP-only preconfirmations are issued when the preconfer does not intend to
+    /// preconf even a single block in its assigned window.
+    function _isEOPOnlyPreconfirmation(Preconfirmation memory _preconfirmation)
+        internal
+        view
+        returns (bool)
+    {
+        return (_preconfirmation.eop && _preconfirmation.rawTxListHash == bytes32(0));
```
ggonzalez94, line 189: this holds true because even if there are no user txs a preconfer that has already submitted a few blocks of preconfirmations and now they wanna signal their window is over will include at least the anchor right? If that's the case maybe we should explain this in the natspec

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,225 @@
...
+    )
+        internal
+        view
+    {
+        // EOP flag must be missing
+        require(!_preconfirmation.eop, InvalidEOPFlag());
+
+        // Protection against a scenario where the previous preconfer submitted an extra
+        // block that it never preconfirmed.
+        require(
```
ggonzalez94, line 98: do we need this check here? isn't the current preconfer in control of issuing their EOP at the end of their window regardless of what the previous preconfer did?

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
```diff
@@ -0,0 +1,93 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import { IPreconfSlasherL1 } from "src/layer1/preconf/iface/IPreconfSlasherL1.sol";
+import { LibPreconfUtils } from "src/layer1/preconf/libs/LibPreconfUtils.sol";
+import { IBridge } from "src/shared/bridge/IBridge.sol";
+import { EssentialContract } from "src/shared/common/EssentialContract.sol";
+import { IPreconfSlasherL2 } from "src/layer2/preconf/IPreconfSlasherL2.sol";
```
ggonzalez94: We've generally avoided L1<>L2 references in the contracts for organization purposes. Would it make sense to move the slashing fault enum to a shared `IPreconfFault` file or similar?

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
```diff
@@ -0,0 +1,93 @@
...
+        SlashAmount memory slashAmount = getSlashAmount();
+        if (
+            fault == IPreconfSlasherL2.Fault.MissedSubmission
+                || fault == IPreconfSlasherL2.Fault.MissingEOP
+        ) {
+            // If the preconfer has missed its L1 slot, these faults are classified under liveness
+            // faults, and incur a smaller penalty.
+            if (
+                LibPreconfUtils.getBeaconBlockRootAt(preconfirmation.submissionWindowEnd)
+                    != bytes32(0)
```
ggonzalez94: I'm not too familar with this library, but just from intuition isn't this logic inverted? To see if they missed their proposing slot shouldn't we be checking that the beacon block is zero?

```suggestion
                LibPreconfUtils.getBeaconBlockRootAt(preconfirmation.submissionWindowEnd)
                    == bytes32(0)
```

ggonzalez94 (commented):
A few more questions. I believe the check for liveness fault on the `PreconfSlasherL1` is inverted

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
ggonzalez94: oh I think it is defined in `IMessageInvokable` right? then we should set the comment to `@inheritdoc`

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
AnshuJalan: Yes, this is correct. On L1, we are validating whether the L1 proposal, as in whether the Ethereum block itself made it to the chain or not.

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
AnshuJalan: Yes, but - we need to wait for a block to be proposed at the same height as the preconfirmation

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
AnshuJalan: Imagine this scenario:

Preconfed: [1] [2] [3] [4] [5]
Submitted: [1] [2] [3] [4] [4] [5]

The preconfer in the first window copied 4, and submitted it. This causes a duplicate 4 to be in the chain and the slasher compares preconf_5 with submitted_4. This causes preconfirmation slashing.  

In this case, we second preconfer has two options:
- Do not post 4 again and be safe - if the node malfunctions for whatever reason and ends up submitting it anyway, the preconfer gets slashed.
- We check the parent's window - 5's parent is 4, and the expected window in the same as 5, but if the previous preconfer submits it, we have a different window

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
AnshuJalan: Because the preconfer must not submit a block in a window after issuing an eop-only preconf

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
AnshuJalan: So far, my understanding has been the preconfer doesn't submit anything at all if an eop-only preconf is issued.
CC: @linoscope

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
AnshuJalan: If preconfer X preconfs 4 and 5 - 4 does not have EOP
and, preconfer X - 1 submits an extra block i.e 4, then 4 can be slashed for missing EOP as it is the last block in the previous window. So, we need this check for the safety of X.

packages/protocol/contracts/layer1/preconf/impl/PreconfSlasherL1.sol
AnshuJalan: Right, this is inverted. Will fix

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -0,0 +1,427 @@
...
+    /// @notice Current state of the anchor proposal being processed.
+    /// @dev 3 slots used to store the state:
+    State private _state;
+
+    /// @dev Storage for checkpoint management
+    /// @dev 2 slots used
+    LibCheckpointStore.Storage internal _checkpointStorage;
+
+    /// @notice Maps L2 block id to preconfirmation metadata.
+    mapping(uint256 blockId => PreconfMeta meta) public blockIdToPreconfMeta;
```
ggonzalez94: nit: this will be audited and deployed together with shasta, but not used during whitelisted preconfs. Should we add a short natspec explaining this is needed for future slashing conditions?

packages/protocol/contracts/layer2/based/ShastaAnchor.sol
```diff
@@ -0,0 +1,427 @@
...
+        require(block.number >= shastaForkHeight, L2_FORK_ERROR());
+
+        previousState_ = _state;
+        newState_ = previousState_;
+
+        // Prevent duplicate calls within same block
+        _trackParentBlockHash(block.number - 1);
+
+        // Handle prover designation on first block
+        if (_params.blockIndex == 0) {
```
ggonzalez94: you'll have to deal with one more merge conflict when merging with main, we removed the block index from the params to simplify the derivation pipeline, but it should be fairly straight forward

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
ggonzalez94: yes, but what I meant is that this is validated on L1 not on this contract. The fault type means the preconfer missed their slot, but this function alone can't validate that right?

packages/protocol/contracts/shared/preconf/IPreconfSlasher.sol
```diff
@@ -0,0 +1,44 @@
...
+        bytes32 rawTxListHash;
+        // The expected submission window of the parent block
+        uint256 parentSubmissionWindowEnd;
+        // The timestamp of the preconfer's slot in the lookahead
+        uint256 submissionWindowEnd;
+    }
+
+    // The slashing reason forwarded to the L1 preconfirmation slasher
+    enum Fault {
+        // The preconfer did not submit a preconfed block
```
ggonzalez94: nit:
```suggestion
        // The preconfer did not submit a preconfed block to the L1 inbox
```

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
ggonzalez94: but then the preconfirmation object is not signed by preconfer X and he wouldn't be slashed? (X-1 would). Maybe I'm missunderstanding something here, if it is easier we can clarify in our call

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
ggonzalez94: oh. that's because EOP only preconfs are not expected to be submitted right?

ggonzalez94 (approved):
LGTM. Left a few minor nits, but this should be ready to be merged.

There's a few merge conflicts that need to be solved(e.g. we removed the blockIndex from the anchor contract and the contract layouts are stored differently now in their own files). Let me know if you have questions on this

dantaik:
@AnshuJalan could you please resolve the merge conflict with the main branch?

ggonzalez94:
@codex please review again. Focus on any issue that has not been detected on previous reviews or that has come up from the latest changes in main

packages/protocol/contracts/layer2/preconf/PreconfSlasherL2.sol
```diff
@@ -0,0 +1,222 @@
...
+        );
+
+        IBridge.Message memory message = IBridge.Message({
+            id: 0,
+            fee: 0,
+            gasLimit: 0,
+            from: address(0),
+            srcChainId: 0,
+            srcOwner: msg.sender,
+            destChainId: uint64(LibNetwork.ETHEREUM_MAINNET),
```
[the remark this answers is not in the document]
ggonzalez94: fixed: https://github.com/taikoxyz/taiko-mono/pull/20304/commits/74c3b73196aa7395d95e5556edf26a48ff8f6aad

packages/protocol/contracts/layer1/preconf/impl/UnifiedSlasher.sol
```diff
@@ -0,0 +1,94 @@
...
+    address public immutable preconfSlasherL1;
+    address public immutable lookaheadSlasher;
+
+    error CallerIsNotURC();
+    error UnsupportedCommitmentType();
+
+    constructor(address _preconfSlasherL1, address _lookaheadSlasher) EssentialContract() {
+        preconfSlasherL1 = _preconfSlasherL1;
+        lookaheadSlasher = _lookaheadSlasher;
+    }
```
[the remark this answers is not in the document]
ggonzalez94: fixed: https://github.com/taikoxyz/taiko-mono/pull/20304/commits/54af470e5441759e13732d9335ef707c2216b911

packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -643,4 +643,33 @@ contract LookaheadStore is ILookaheadStore, IProposerChecker, Blacklist, Essenti
             commitmentType: 0, payload: abi.encode(_lookahead), slasher: lookaheadSlasher
         });
     }
+
```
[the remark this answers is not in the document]
ggonzalez94: @AnshuJalan I think this makes sense right? I'm changing it to use `LibPreconfConstants.LOOKAHEAD_COMMITMENT_TYPE`. Let me know if you disagree

ggonzalez94 (approved):

PR #20308, opened 2025-09-29
fix(protocol): pass correct L1 controller in L2 upgrade impl constructors

sashass1315:
Problem: New implementations in tests were constructed with the proxy address as the immutable daoController/owner address. After upgrading, runtime checks in onMessageInvocation compare ctx.from to this immutable and would fail because ctx.from is the real L1 controller, not the L2 proxy.

Fix: In DelegateController.t.sol, pass daoController; in DelegateOwner.t.sol, pass eBridge when constructing the new implementation addresses used for UUPS upgrades. This aligns with production deploy scripts and preserves correctness of cross-chain sender validation.


dantaik (approved):

dantaik:
DelegateOwner was used in our Hekla testnet, but since Hekla deprecated now, so I'd suggest to remove DelegateOwner and its tests

davidtaikocha (approved):

dantaik (approved):

dantaik:
@sashass1315 could you resolve merge conflicts with the main branch? Thank you.

PR #20309, opened 2025-09-29
perf: optimize gas price calculation (19x faster)

futreall:
Replace expensive string conversions (BigInt→string→number→BigInt) with native BigInt comparison, improving performance from 138ns to 5.5ns per operation


davidtaikocha:
Closed due to `[‎packages/taikoon-ui](https://github.com/taikoxyz/taiko-mono/pull/20309/files#diff-cb5bdd6a1001cead4625f301b35e72b9db0ed7309d5bb66ad4938061c02c264d)` is no longer used.

PR #20310, opened 2025-09-29
chore(protocol): try to optimize and simplify `consumeForcedInclusions`

ggonzalez94:
context: https://github.com/taikoxyz/taiko-mono/pull/20288#discussion_r2384430944

For some reason it is actually slightly more expensive to do it this way, but the code is actually more readable.

---

> [!NOTE]
> Refactors forced inclusion consumption to return derivation sources (with extra slot) and updates Inbox to use it; refreshes gas reports and snapshots.
> 
> - **Contracts**:
>   - **`LibForcedInclusion.consumeForcedInclusions`**:
>     - Changes signature to `internal` and returns `IInbox.DerivationSource[]` sized `toProcess + 1` (last slot reserved for normal source).
>     - Populates sources with `isForcedInclusion=true`; aggregates fees and updates queue pointers as before.
>     - Adds import of `IInbox` and minor loop/storage tweaks.
>   - **`Inbox._buildDerivationSources`**:
>     - Consumes forced inclusions into pre-marked `DerivationSource[]` and places normal source in the last slot.
>     - Handles zero forced-inclusion case by allocating a single-slot array.
>     - Computes `oldestForcedInclusionTimestamp` from first returned source.
> - **Meta**:
>   - Updates gas report numbers in `packages/protocol/gas-reports/layer1-contracts.txt`.
>   - Updates snapshot values in `packages/protocol/snapshots/shasta-propose.json`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b75f043b113d83cf29a6cb0e1a4ca1cfc410e1f7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -792,22 +792,21 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
         uint256 remainingForcedInclusions = type(uint256).max;
         // We need to capture the last processed timestamp before consuming the forced inclusions
         uint48 lastProcessedAt = _forcedInclusionStorage.lastProcessedAt;
 
         if (_input.numForcedInclusions > 0) {
-            (forcedInclusions, remainingForcedInclusions) = LibForcedInclusion
-                .consumeForcedInclusions(
+            // Get derivation sources with forced inclusions marked and an extra slot for normal
+            // source
+            (sources, remainingForcedInclusions) = LibForcedInclusion.consumeForcedInclusions(
```
dantaik, line 802: We should further combine consumeForcedInclusions and isOldestForcedInclusionDue? If we do so, we simply call to consumeForcedInclusions even if _input.numForcedInclusions is 0.

dantaik (changes requested):

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
```diff
@@ -67,38 +68,34 @@ library LibForcedInclusion {
...
         Storage storage $,
         address _feeRecipient,
         uint256 _count
     )
         public
-        returns (
-            IForcedInclusionStore.ForcedInclusion[] memory inclusions_,
-            uint256 availableAfter_
-        )
+        returns (IInbox.DerivationSource[] memory sources_, uint256 availableAfter_)
```
dantaik, line 82: make this function internal

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -792,22 +792,21 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
                 _forcedInclusionStorage, msg.sender, _input.numForcedInclusions
             );
-            sourceCount += forcedInclusions.length;
             oldestForcedInclusionTimestamp =
-                uint48(forcedInclusions[0].blobSlice.timestamp.max(lastProcessedAt));
+                uint48(sources[0].blobSlice.timestamp.max(lastProcessedAt));
```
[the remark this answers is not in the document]
dantaik: Is this a real bug?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
ggonzalez94: long claude response, but the idea is that we should optimize for the case when `numForcedInclusions` is zero(which is by far the most common case)

```
Recommendation: Keep contracts separate ✅
The current implementation is already optimized for the common case (zero forced inclusions).

Gas Comparison
Current (zero forced inclusions):

✅ Skips consumeForcedInclusions entirely
✅ Small array allocation: new DerivationSource[](1)
✅ Only calls view function isOldestForcedInclusionDue

Proposed (always call consumeForcedInclusions):

❌ 2 SLOADs (head + tail from storage)
❌ Updates lastProcessedAt even with zero inclusions (line 102)
❌ Unnecessary SSTORE on every proposal


Key Issues with Combining

State corruption: lastProcessedAt would update on every proposal regardless of actual consumption, breaking deadline calculations
Gas inefficiency: Common case pays for unnecessary storage operations
Reduced clarity: Current separation makes intent explicit


Potential Micro-Optimization
The existing short-circuit logic already optimizes well:
solidityrequire(
    (_input.numForcedInclusions >= _minForcedInclusionCount)
    || remainingForcedInclusions == 0  // ← Avoids view call in common case
    || !LibForcedInclusion.isOldestForcedInclusionDue(...)
);

Verdict
Current architecture prioritizes the happy path. Combining would trade correctness and gas efficiency for minimal code simplification—not worth it.
```

ggonzalez94: I've also ellaborated on this [here](https://github.com/taikoxyz/taiko-mono/pull/20288#discussion_r2388871161)

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
ggonzalez94: done. It saved a lot of gas

ggonzalez94:
merging this and we can finalize the discussion on the main PR

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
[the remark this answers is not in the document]
ggonzalez94: It was indeed an issue, as the one above. Fixed it by moving that logic inside the library [here](https://github.com/taikoxyz/taiko-mono/pull/20288/commits/d9ef5bbdb8e333efd4d9ab4c77e262f4920fe41d)

PR #20313, opened 2025-09-30
feat(taiko-client): introduce `ShastaInboxAddress` and `PacayaInboxAddress` flags

davidtaikocha:
ref: https://github.com/taikoxyz/taiko-mono/pull/20224


packages/protocol/script/layer1/based/DeployProtocolOnL1.s.sol
```diff
@@ -79,18 +79,18 @@ contract DeployProtocolOnL1 is DeployCapability {
...
         // ---------------------------------------------------------------
         // Deploy rollup contracts
-        (address taikoInboxAddr, address proofVerifier, address whitelist) =
-            deployRollupContracts(sharedResolver, contractOwner);
+        (address pacayaInboxAddr, address shastaInboxAddr, address proofVerifier, address whitelist)
+        = deployRollupContracts(sharedResolver, contractOwner);
 
         // Deploy verifiers
-        OpVerifier opImpl = new OpVerifier(taikoInboxAddr, proofVerifier);
+        OpVerifier opImpl = new OpVerifier(pacayaInboxAddr, proofVerifier);
```
YoGhurt111, line 86: If we separate 2 differnt inbox address, the constructor of most of contracts also need to be changed, otherwise these will be unavailable after Shasta

YoGhurt111: Or do we have any brief way to handle this case?

YoGhurt111:
The code of client looks good to me.

packages/protocol/script/layer1/based/DeployProtocolOnL1.s.sol
davidtaikocha: Guess we need two sets of the verifiers, but let's make the final decision after the Shasta prover implementation, no need to make any decision rn.

YoGhurt111 (approved):

RogerLamTd (approved):

smtmfft (approved):

PR #20315, opened 2025-09-30
feat(protocol): add minCheckpointDelay to rate-limit checkpoint saving

dantaik:
## Summary

Add checkpoint timestamp tracking and minimum delay between checkpoints to prevent excessive checkpoint storage operations.

## Changes

- **IInbox.CoreState**: Add `lastCheckpointTimestamp` field to track when last checkpoint was saved
- **IInbox.Config**: Add `minCheckpointDelay` parameter for configurable delay (in seconds)
- **Inbox.sol**: Only save checkpoints if `block.timestamp >= coreState.lastCheckpointTimestamp + _minCheckpointDelay`
  - When checkpoint is saved, update `lastCheckpointTimestamp` to current `block.timestamp`
- **LibHashOptimized.sol**: Include `lastCheckpointTimestamp` in CoreState hashing
- **Test files**: Update all CoreState and Config constructors to include new fields


dantaik:
@claude please review

davidtaikocha (approved):

dantaik:
@claude please review

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -25,7 +25,15 @@ import { LibMath } from "src/shared/libs/LibMath.sol";
 ///      - Proof verification with transition record management
 ///      - Ring buffer storage for efficient state management
 ///      - Bond instruction processing for economic security
-///      - Finalization of proven proposals
+///      - Finalization of proven proposals with checkpoint rate limiting
+///
+/// @dev Checkpoint Rate Limiting:
+///      The minCheckpointDelay parameter controls how frequently checkpoints are saved to storage.
+///      - Finalization can occur at any frequency without extra cost
```
```suggestion
```

dantaik:
@claude  please review

dantaik:
@claude please review

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -25,7 +25,33 @@ import { LibMath } from "src/shared/libs/LibMath.sol";
...
+///      - Checkpoint storage is rate-limited to save gas and prevent excessive storage operations
+///      - Nodes syncing checkpoint data will only sync at the rate limited by minCheckpointDelay
+///      - Set minCheckpointDelay to 0 to disable rate limiting and save every checkpoint
```
ggonzalez94: maybe we can move this natspec to the finalize function to avoid the main one from getting too long? The natspec of the inbox gives you the TL;DR of the contract, and then for the specifics you go to the function

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -25,7 +25,33 @@ import { LibMath } from "src/shared/libs/LibMath.sol";
...
+///      - Timestamp/block number arithmetic with practical overflow impossibility
+///      See inline comments for specific safety justifications on each unchecked block.
+///
+/// @dev IMPORTANT - Type Conversions in Unchecked Blocks:
+///      Due to aggressive use of unchecked blocks throughout this contract and its subcontracts,
+///      developers MUST explicitly cast values to their proper types before performing mathematical
+///      operations when mixing different numeric types. Without explicit casts, Solidity may
+/// perform
+///      implicit conversions that could lead to unexpected results within unchecked blocks.
+///      Example: uint256(uint48Value) + anotherUint48 instead of uint48Value + anotherUint48
```
ggonzalez94: this is great warning and explanation on this for developers. I think eventually we may want to move this to an engineering or contributing guideline, since this will apply not only to the inbox, but also to the SS, libs, etc.

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -836,72 +873,88 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
-    /// @dev Performs up to `maxFinalizationCount` finalization iterations.
-    /// The caller is forced to finalize transition records that have passed their finalization
-    /// grace period, but can decide to finalize ones that haven't.
-    /// @param _input Input containing transition records and end block header
-    /// @return _ Core state with updated finalization counters
+    /// @dev Finalizes proven proposals and updates the checkpoint.
+    /// Executes up to `maxFinalizationCount` iterations to finalize proposals.
+    /// Rate limiting is applied to prevent excessive checkpoint storage operations,
+    /// ensuring smooth finalization even in high-throughput systems where checkpoint
+    /// storage could become a bottleneck.
```
ggonzalez94: nit for clarity:
```suggestion
    /// Rate limiting is applied to prevent excessive checkpoint syncing. This helps reduce `SSTORE` operations, but causes L2 checkpoints to be made available less frequently on the L1.
```

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -836,72 +873,88 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
-            require(checkpointHash == lastFinalizedRecord.checkpointHash, CheckpointMismatch());
-            LibCheckpointStore.saveCheckpoint(
-                _checkpointStorage, _input.checkpoint, _maxCheckpointHistory
-            );
-        }
+            // Update checkpoint if any proposals were finalized and minimum delay has passed.
+            // Rate limiting: Checkpoints are only saved if at least minCheckpointDelay seconds
+            // have elapsed since the last checkpoint. This prevents excessive checkpoint storage
+            // operations while still allowing finalization to proceed.
+            // Note: minCheckpointDelay can be zero to disable rate limiting.
```
ggonzalez94: maybe we don't need this comments since it is already explained in the natspec of both the inbox and the function?

ggonzalez94 (approved):
Left a few small comments, otherwise looks good to me.

I have one question and one potential optimization we can consider in the future:

- @dantaik can you remind me what is the benefit of finalizing if we are not going to sync the checkpoint? What I mean is can we use the `minCheckpointDelay` parameter to avoid finalization all together when it hasn't passed? or are there things the protocol still benefits from finalizing even if no checkpoint is saved(the only one I could think of is clearing the ring buffer, but in that case we can use that as a check to see if we should finalize)

- One potential optimization to this that we may consider is allowing the proposer to sync a checkpoint if they want to **even when minCheckpointDelay has not passed**. This could be useful if they want to make checkpoints available to the L1 for them or users to withdraw.
I think it is a good design principle in general to have parameters that don't force things to happen before a set time, but allow the proposer(or other actors) to do it if they decide it is beneficial. This could be set as an input parameter or checking the transitions array length

dantaik:
> Left a few small comments, otherwise looks good to me.
> 
> I have one question and one potential optimization we can consider in the future:
> 
> * @dantaik can you remind me what is the benefit of finalizing if we are not going to sync the checkpoint? What I mean is can we use the `minCheckpointDelay` parameter to avoid finalization all together when it hasn't passed? or are there things the protocol still benefits from finalizing even if no checkpoint is saved(the only one I could think of is clearing the ring buffer, but in that case we can use that as a check to see if we should finalize)

There’s little benefit in finalizing without checkpoint synchronization. However, finalization and synchronization have different costs:

- Finalization requires reading one or more transitions.
- Synchronization requires writing checkpoints.

If we only finalize when checkpoint sync is needed, the cost becomes higher and falls unfairly on whichever proposer happens to do it. Allowing proposals to be finalized without requiring checkpoint synchronization spreads the cost more evenly across proposers.

Does this reasoning make sense to you? I’m not entirely certain about the argument, so I’m open to suggestions if simplifying the code would be better.

> * One potential optimization to this that we may consider is allowing the proposer to sync a checkpoint if they want to **even when minCheckpointDelay has not passed**. This could be useful if they want to make checkpoints available to the L1 for them or users to withdraw.
>   I think it is a good design principle in general to have parameters that don't force things to happen before a set time, but allow the proposer(or other actors) to do it if they decide it is beneficial. This could be set as an input parameter or checking the transitions array length

I agree, let me create another small PR on top of it.

PR #20317, opened 2025-09-30
Add HTTP client timeout in proof producer to prevent hanging requests

viktorking7:
### PR Description
- Introduces a 30s `http.Client` timeout in `requestHTTPProofResponse` to avoid indefinitely hanging requests during network stalls.
- Keeps context cancellation semantics intact; the shorter of context deadline vs client timeout will apply.

#### Why
- Without a timeout, network issues can cause goroutines to hang, degrading service reliability and throughput.

#### What changed
- `packages/taiko-client/prover/proof_producer/common.go`:
  - Initialize `http.Client{ Timeout: 30 * time.Second }`


davidtaikocha:
Closed since we have timeout setting in `ctx`.

PR #20319, opened 2025-09-30
refactor(protocol): consolidate forced inclusion checks to minimize storage ops

dantaik:
## Description

Consolidates and optimizes forced inclusion logic in the Inbox contract to minimize storage operations and improve gas efficiency.

### Phase 1: Consolidate Storage Operations
- **LibForcedInclusion.consumeForcedInclusions**: Added `_forcedInclusionDelay` parameter and `isRemainingForcedInclusionDue_` return value
- **Storage Pattern**: Load once at start, process all inclusions, write once at end
- **Inbox._buildDerivationSources**: Eliminated redundant `isOldestForcedInclusionDue` call by using return value from `consumeForcedInclusions`

### Phase 2: Simplify Return Interface
- **Removed unused fields** from `ConsumptionResult` struct: `remainingForcedInclusionCount`, `oldestTimestamp`, `hasRemainingDue`
- **Streamlined return**: Only return `sources` array and `allowsPermissionless` flag
- **Internal tracking**: Use local variables for intermediate calculations

### Phase 3: Refactor for Clarity
- **Extracted helper functions**:
  - `_processInclusions`: Iterates queue and accumulates fees
  - `_consumeAndUpdateStorage`: Processes inclusions and updates storage
  - `_validateForcedInclusionRequirements`: Validates forced inclusion requirements
  - `_isOldestInclusionDue`: Checks if oldest remaining inclusion is due
- **Moved validation**: Validation logic moved from Inbox.sol into `LibForcedInclusion.consumeForcedInclusions`
- **Inlined derivation**: Inlined `_buildDerivationSources` logic directly into `Inbox.propose`

### Phase 4: Gas Optimizations
- **Early returns**: Converted `_validateForcedInclusionRequirements` to use early returns instead of boolean logic
- **Inlined calculation**: Inlined `_isPermissionlessAllowed` calculation directly in `consumeForcedInclusions`
- **Type optimization**: Changed `forcedInclusionDelay` from `uint64` to `uint16` to save storage/calldata

## Benefits

- **Reduces storage reads**: Minimized SLOAD operations in common path
- **Improves code clarity**: Smaller focused functions, better separation of concerns
- **Gas savings**: Early returns, inlined calculations, optimized types
- **Maintains correctness**: All 651 tests pass

## Gas Impact
The storage access pattern is minimal and optimal:
✅ Single initial SLOAD for head/tail/lastProcessedAt
✅ One SLOAD per forced inclusion processed (unavoidable)
✅ Single conditional SLOAD for validation (only when needed, different entry)
✅ Single SSTORE at the end (only when processing occurs)
No further optimization is possible without changing the semantics of the forced inclusion validation logic.


dantaik:
@claude please review

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
```diff
@@ -65,97 +70,230 @@ library LibForcedInclusion {
...
     {
-        (uint48 head, uint48 tail, uint48 lastProcessedAt) = ($.head, $.tail, $.lastProcessedAt);
+        // Validate forced inclusion requirements: must satisfy one of:
+        // 1. Requested minimum required count - return early
+        if (_numForcedInclusionsRequested >= _minForcedInclusionCount) return;
 
-        // Early exit for empty queue (most common case)
-        if (head == tail) return false;
+        // 2. Emptied the queue (remaining is 0 AND requested > 0) - return early
+        if (_remainingForcedInclusionCount == 0 && _numForcedInclusionsRequested > 0) return;
```
ggonzalez94: If the queue is empty I don't think _numForcedInclusionsRequested has to be greater than zero right?
It means there are no forced inclusions to process, so we can safely skip this check

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
```diff
@@ -65,97 +70,230 @@ library LibForcedInclusion {
...
 
-        uint256 timestamp = $.queue[head].blobSlice.timestamp;
+        // 3. No remaining inclusions are due - only check if we reach here
+        if (_remainingForcedInclusionCount == 0) return;
```
ggonzalez94: this is covered by the check above right?

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
```diff
@@ -65,97 +70,230 @@ library LibForcedInclusion {
...
-    /// @return sources_ Array of derivation sources with forced inclusions marked and an extra
-    /// empty
-    /// slot at the end for the normal source. The array size is toProcess + 1, where the last slot
-    /// is uninitialized for the caller to populate.
-    /// @return availableAfter_ Number of forced inclusions remaining in the queue after consuming
-    /// @return oldestForcedInclusionTimestamp_ The timestamp of the oldest forced inclusion that
-    /// was
-    /// processed. block.timestamp if no forced inclusions were consumed.
+    /// @dev See `IInbox.isOldestForcedInclusionDue`
+    function isOldestForcedInclusionDue(
```
ggonzalez94, line 65: shouldn't we reuse the private `isOldestForcedInclusionDue` logic here and call that function?

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
```diff
@@ -65,97 +70,230 @@ library LibForcedInclusion {
...
+        unchecked {
+            for (uint256 i; i < count; ++i) {
+                IForcedInclusionStore.ForcedInclusion storage inclusion = $.queue[head + i];
+                sources[i] = IInbox.DerivationSource(true, inclusion.blobSlice);
+                totalFees += inclusion.feeInGwei;
+            }
+        }
+    }
+
+    /// @dev Validates forced inclusion requirements
```
ggonzalez94: nit: let's improve the description of this function. Something like this?
```suggestion
    /// @dev Ensures no overdue forced inclusions remain unprocessed
```

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
```diff
@@ -65,97 +70,230 @@ library LibForcedInclusion {
...
+    /// @param $ Storage reference
+    /// @param _numForcedInclusionsRequested Number requested
+    /// @param _minForcedInclusionCount Minimum required count
+    /// @param _remainingForcedInclusionCount Number remaining in queue
+    /// @param _head Current queue head position
+    /// @param _lastProcessedAt Timestamp of last processing
+    /// @param _forcedInclusionDelay Delay in seconds before inclusion is due
+    function _validateForcedInclusionRequirements(
```
ggonzalez94: actually maybe we should inline this function instead of making it a separate function? We end up passing a bunch of parameters that we have already calculated and/or gotten from storage, so it is not very well encapsulated or reusable

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
```diff
@@ -65,97 +70,230 @@ library LibForcedInclusion {
...
+
+            // Update queue position and last processed time
+            head_ = _head + uint48(_toProcess);
+            lastProcessedAt_ = uint48(block.timestamp);
+
+            // Write to storage once
+            ($.head, $.lastProcessedAt) = (head_, lastProcessedAt_);
+        } else {
+            // No inclusions processed
+            oldestTimestamp_ = type(uint48).max;
```
ggonzalez94: setting this to max causes an overflow. This is a bug I found on my PR and fixed before we merge it(I originally thought it was fine, but since we are running it inside an unchecked block when adding this with other fields it silently overflowed).
What I did is return `block.timestamp` instead. This way the proposer checks still hold true

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -201,24 +201,30 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
...
+                _forcedInclusionDelay,
+                _permissionlessInclusionMultiplier
+            );
+
+            // Add normal proposal source in last slot
+            result.sources[result.sources.length - 1] =
+                DerivationSource(false, LibBlobs.validateBlobReference(input.blobReference));
+
+            // If forced inclusion is old enough, allow anyone to propose
+            // (endOfSubmissionWindowTimestamp = 0)
```
ggonzalez94: make it clearer that we set this value to zero, not that this is the reason why we allow anyone to propose
```suggestion
            // and set endOfSubmissionWindowTimestamp = 0
```

ggonzalez94 (commented):
Left a few comments. We need to update it with main after merging the other PR, but gas consumption does look smaller

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
dantaik: Oh, then bug is not with `oldestTimestamp_ = type(uint48).max;`, it is because we should convert it to uint256 before adding something to it.

dantaik: Current the addition happens here:

```
uint256(_forcedInclusionDelay) * _permissionlessInclusionMultiplier + oldestTimestamp;
```

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
dantaik: Refactored the code for better readability

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
dantaik: the code and comment is now improved.

dantaik:
@ggonzalez94 @davidtaikocha please take another look

ggonzalez94 (commented):
I won't have enough time before I leave to do a proper review, but functionally this seems to be correct from what I could see. I do have a few comments that may be worth thinking about for maintainability(TL;DR I think I liked your first approach with some of the fixes applied better):
1. Having the forced inclusion logic spread between the `LibForcedInclusion` and the `Inbox` seems awkward. I believe I'm still in favor of keeping most in the library, but if not move it to the inbox and remove the library.
2. I see we removed the `_propose` internal function? Not sure if it was causing any issue, but seemed more readable?
3. I see we are still returning `type(uint48).max` when there are no forced inclusions. I think we are now handling it correctly by casting it to `uint256` before multiplying it, but I think this might be error prone in the sense that we are running inside `unchecked` blocks, and the next person touching the contract might inadvertently use that value for some calculations. Maybe returning `block.timestamp` instead is safer?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -807,6 +864,52 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
         emit Proposed(_encodeProposedEventData(payload));
     }
 
+    /// @dev Finalizes proven proposals and updates checkpoint
+    /// @dev Performs up to `maxFinalizationCount` finalization iterations.
+    /// The caller is forced to finalize transition records that have passed their finalization
+    /// grace period, but can decide to finalize ones that haven't.
+    /// @param _input Input containing transition records and end block header
+    /// @return _ Core state with updated finalization counters
+    function _finalize(ProposeInput memory _input) private returns (CoreState memory) {
```
ggonzalez94, line 873: was finalized modified or just moved from its position?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
```diff
@@ -957,7 +1014,11 @@ contract Inbox is IInbox, IForcedInclusionStore, ICheckpointStore, EssentialCont
         _coreState.lastFinalizedTransitionHash = _transitionRecord.transitionHash;
 
         // Process bond instructions
-        _processBondInstructions(_coreState, _transitionRecord.bondInstructions);
```
ggonzalez94, line 960: any reason to inline the `_processBondInstructions` function?

packages/protocol/contracts/layer1/shasta/libs/LibForcedInclusion.sol
```diff
@@ -137,19 +71,38 @@ library LibForcedInclusion {
...
 
-        uint256 timestamp = $.queue[head].blobSlice.timestamp;
+    /// @dev Checks if the oldest remaining forced inclusion is due (internal variant with
+    /// parameters)
+    /// @param $ Storage reference
+    /// @param head Current queue head position
+    /// @param lastProcessedAt Timestamp of last processing
+    /// @param forcedInclusionDelay Delay in seconds before inclusion is due
+    /// @return True if the oldest remaining inclusion is due for processing
+    function isOldestForcedInclusionDue(
```
ggonzalez94, line 89: nit: `_isOldestForcedInclusionDue` 

prefix the function name with underscore

ggonzalez94 (commented):
Left a couple more comments.

I see we are inlining some functions. Not sure if this was intentional to reduce gas cost or accidentally by AI, but maybe the small gas saving(if any) is not worth it?

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: Moved only.

dantaik:
> I won't have enough time before I leave to do a proper review, but functionally this seems to be correct from what I could see. I do have a few comments that may be worth thinking about for maintainability(TL;DR I think I liked your first approach with some of the fixes applied better):
> 
> 1. Having the forced inclusion logic spread between the `LibForcedInclusion` and the `Inbox` seems awkward. I believe I'm still in favor of keeping most in the library, but if not move it to the inbox and remove the library.

I think most of the code still represents the core behavior of the inbox, and since the inbox’s code size is still within the limit, it makes sense to keep everything in a single file.

> 2. I see we removed the `_propose` internal function? Not sure if it was causing any issue, but seemed more readable?

The current propose function (with _propose merged in) is quite clear and easy to understand.

> 3. I see we are still returning `type(uint48).max` when there are no forced inclusions. I think we are now handling it correctly by casting it to `uint256` before multiplying it, but I think this might be error prone in the sense that we are running inside `unchecked` blocks, and the next person touching the contract might inadvertently use that value for some calculations. Maybe returning `block.timestamp` instead is safer?

This is not the only case where developers needs to care about overflow, we are using a lot more `uncheck`s, I'll add a comment to the file level instead (in another PR: https://github.com/taikoxyz/taiko-mono/pull/20315/commits/4d9ed6c962285631b4fca62db3cc32400a317ba3)

packages/protocol/contracts/layer1/shasta/impl/Inbox.sol
dantaik: Its too simple and only contains a for-loop

ggonzalez94:
> I think most of the code still represents the core behavior of the inbox, and since the inbox’s code size is still within the limit, it makes sense to keep everything in a single file.

I still believe it is a bit awkward to have a `LibForcedInclusion` library and that it only partially has the logic to forced inclusion, while the rest is on the inbox. It does not feel like the best encapsulation IMO(e.g. decide how many should be processed and calculate the fees). And since we can still make the functions internal in the library I don't see a big downside of moving them there(but maybe I'm missing something since I didn't have the time to fully review this). If you still believe they are better in the inbox we can merge this and revisit at a later point if needed.

> The current propose function (with _propose merged in) is quite clear and easy to understand.
After reading it again I agree

> This is not the only case where developers needs to care about overflow, we are using a lot more unchecks, I'll add a comment to the file level instead (in another PR: https://github.com/taikoxyz/taiko-mono/commit/4d9ed6c962285631b4fca62db3cc32400a317ba3)

Makes sense. I like the new warnings

ggonzalez94 (approved):
Still a bit undecided on what logic should go in the inbox vs the library, but I don't think this should be a blocker.

davidtaikocha (approved):

PR #20320, opened 2025-09-30
fix(balance-monitor): close RPC clients on shutdown

sky-coderay:
call Close() on the L1 and L2 ethclient instances so their RPC connections are cleaned up
keep the lifecycle consistent with SubcommandAction, which expects Close to release resources


davidtaikocha:
Closed due to unnecessary changes, clients will be closed by `ctx` already.

PR #20322, opened 2025-09-30
docs(protocol): more important updates to the derivation doc

### LINKED ISSUE #20283: Update Derivation.md for multi-source derivation support
After this PR https://github.com/taikoxyz/taiko-mono/pull/20216 merged, we support multi-source derivation, but the `Derivation.md`  doc is outdated, we need to reach consensus on the approach for handling multi-source derivation.

ggonzalez94:
full context: https://www.notion.so/taikoxyz/Update-Shasta-Derivation-Doc-2759673143d68095a586cc678b249431?source=copy_link

**I pushed this on a rush before leaving, so there are some missing details I'm gonna leave in the comments**

## Changes

- **Restructured manifest hierarchy**: Introduced `ProposalManifest` containing proposal-level `proverAuthBytes` and an array of `DerivationSourceManifest[]`. Previously, `proverAuthBytes` was duplicated in each source manifest but only the first was used, wasting blob space and creating semantic confusion.

- **Renamed struct for clarity**: `ProposalManifest` → `DerivationSourceManifest` to accurately reflect that each represents one derivation source's blocks, not the entire proposal.

- **Documented invalid source handling**: Clarified that invalid `DerivationSourceManifest` objects are replaced with default manifests (single empty block) without invalidating the entire proposal. This prevents forced inclusion censorship—a malicious proposer cannot invalidate valid forced inclusions by including bad data in another source.

- **Documented block index monotonicity**: Added explicit requirement that `_blockIndex` must be globally monotonic across all sources in a proposal (not reset per source). This ensures proposal-level operations (prover designation, bond processing) execute exactly once, preventing financial incorrectness and hash chain violations.

---

> [!NOTE]
> Refactors manifest to proposal-level `proverAuthBytes` plus per-source manifests, and rewrites derivation docs to add source-level metadata, per-source validation/defaulting, forced inclusion handling, and globally monotonic block index.
> 
> - **Contracts (LibManifest.sol)**:
>   - **Manifest Hierarchy**: `ProposalManifest` now contains proposal-level `proverAuthBytes` and `DerivationSourceManifest[] sources`.
>   - **New Struct**: `DerivationSourceManifest` with `blocks` array; added notes on globally monotonic `_blockIndex` across all sources.
> - **Docs (Derivation.md)**:
>   - **Metadata Model**: Expanded to three tiers: proposal-level, derivation source-level, and block-level; updated assignment tables.
>   - **Manifests & Extraction**:
>     - Introduces `ProposalManifest` aggregating per-source `DerivationSourceManifest`s.
>     - Defines per-source blob validation/decoding pipeline and limits; failures default only that source (single anchor-only block).
>     - Details forced inclusion publishing and required zeroed fields for FI blocks.
>   - **Validation Rules**:
>     - Timestamp/anchor validations applied per source; non-forced sources without valid anchors defaulted (FI exempt).
>     - Coinbase selection clarified; gas limit rules unchanged in method but re-referenced per source.
>   - **Prover System**:
>     - `proverAuthBytes` is proposal-level (processed once at `_blockIndex == 0`).
>     - Low-bond handling replaces only the proposer’s source manifest; FI sources still processed.
>   - **Block Indexing**: `_blockIndex` must be globally monotonic across all sources within a proposal to ensure single execution of proposal-level ops.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f54723687eb70f9dbc99e9f7b19dff809fbc37ba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
+++ b/packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
@@ -84,7 +84,7 @@ library LibManifest {
     /// @notice Represents a derivation source manifest containing blocks for one source
     /// @dev Each proposal can have multiple DerivationSourceManifests (one per DerivationSource).
     /// If a DerivationSourceManifest is invalid, it is replaced with a default manifest
-    /// (single empty block), but the entire proposal is NOT invalidated. This design prevents
+    /// (single block with only an anchor transaction), but the entire proposal is NOT invalidated. This design prevents
     /// censorship of forced inclusions: a malicious proposer cannot invalidate their entire
     /// proposal (including valid forced inclusions) by including bad data in one source.
     ///
--- a/packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
+++ b/packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
@@ -84,7 +84,8 @@ library LibManifest {
     /// @notice Represents a derivation source manifest containing blocks for one source
     /// @dev Each proposal can have multiple DerivationSourceManifests (one per DerivationSource).
     /// If a DerivationSourceManifest is invalid, it is replaced with a default manifest
-    /// (single block with only an anchor transaction), but the entire proposal is NOT invalidated. This design prevents
+    /// (single block with only an anchor transaction), but the entire proposal is NOT invalidated.
+    /// This design prevents
     /// censorship of forced inclusions: a malicious proposer cannot invalidate their entire
     /// proposal (including valid forced inclusions) by including bad data in one source.
     ///
--- a/packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
+++ b/packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
@@ -70,14 +70,29 @@ library LibManifest {
         SignedTransaction[] transactions;
     }
 
-    /// @notice Represents a derivation source manifest
+    /// @notice Represents a proposal manifest containing proposal-level metadata and all sources
+    /// @dev The ProposalManifest aggregates all DerivationSources' blob data for a proposal.
+    /// The proverAuthBytes is proposal-level (one designated prover per proposal), while the
+    /// sources array contains per-source block data.
+    struct ProposalManifest {
+        /// @notice Prover authentication data (proposal-level, shared across all sources).
+        bytes proverAuthBytes;
+        /// @notice Array of derivation source manifests (one per DerivationSource).
+        DerivationSourceManifest[] sources;
+    }
+
+    /// @notice Represents a derivation source manifest containing blocks for one source
     /// @dev Each proposal can have multiple DerivationSourceManifests (one per DerivationSource).
     /// If a DerivationSourceManifest is invalid, it is replaced with a default manifest
     /// (single empty block), but the entire proposal is NOT invalidated. This design prevents
     /// censorship of forced inclusions: a malicious proposer cannot invalidate their entire
     /// proposal (including valid forced inclusions) by including bad data in one source.
+    ///
+    /// IMPORTANT: The _blockIndex parameter in updateState() must be globally monotonic across
+    /// all DerivationSourceManifests within a proposal (not reset per source). This ensures
+    /// proposal-level operations (prover designation, bond processing) execute exactly once.
     struct DerivationSourceManifest {
-        bytes proverAuthBytes;
+        /// @notice The blocks for this derivation source.
         BlockManifest[] blocks;
     }
 }
--- a/packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
+++ b/packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
@@ -70,8 +70,13 @@ library LibManifest {
         SignedTransaction[] transactions;
     }
 
-    /// @notice Represents a proposal manifest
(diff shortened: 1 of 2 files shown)


ggonzalez94:
@davidtaikocha there's one thing that I believe we should figure out(or at least I wasn't completely sure):

1. with the new `ProposalManifest` having the proverAuthBytes for the entire proposal now we cannot recover this from the individual derivation. I see 2 options:
  a) we get it from the **last** source of this proposal, since this is the one that the proposer constructs. The others don't need to encode it.
  b) we change the protocol to have this field in the Derivation struct

davidtaikocha:
> @davidtaikocha there's one thing that I believe we should figure out(or at least I wasn't completely sure):
> 
> 1. with the new `ProposalManifest` having the proverAuthBytes for the entire proposal now we cannot recover this from the individual derivation. I see 2 options:
>    a) we get it from the **last** source of this proposal, since this is the one that the proposer constructs. The others don't need to encode it.
>    b) we change the protocol to have this field in the Derivation struct

seems `b` sounds better to me, and we will also make `proverAuth` part of the `ProposeInput`? Then we can avoid introducing another optional field in`DerivationSourceManifest` (`a`).

dantaik (approved):

packages/protocol/contracts/layer1/shasta/libs/LibManifest.sol
```diff
@@ -70,9 +70,29 @@ library LibManifest {
...
+        /// @notice Prover authentication data (proposal-level, shared across all sources).
         bytes proverAuthBytes;
+        /// @notice Array of derivation source manifests (one per DerivationSource).
+        DerivationSourceManifest[] sources;
+    }
+
+    /// @notice Represents a derivation source manifest containing blocks for one source
+    /// @dev Each proposal can have multiple DerivationSourceManifests (one per DerivationSource).
+    /// If a DerivationSourceManifest is invalid, it is replaced with a default manifest
+    /// (single empty block), but the entire proposal is NOT invalidated. This design prevents
```
davidtaikocha:
```suggestion
    /// (single block with only an anchor transaction), but the entire proposal is NOT invalidated. This design prevents
```

packages/protocol/docs/Derivation.md
```diff
@@ -177,37 +187,64 @@ struct BlobSlice {
 }
 ```
 
-The `BlobSlice` struct represents binary data distributed across multiple blobs. Processing involves the following steps:
+The `BlobSlice` struct represents binary data distributed across multiple blobs. Each `DerivationSource` in the `derivation.sources[]` array is processed independently to extract its `DerivationSourceManifest`.
+
+**TODO**: The proposal-level `proverAuthBytes` field needs a storage location. Current `Derivation` struct only has per-source `blobSlice` fields. Option: add dedicated `proposalBlob` field to `Derivation` struct (requires protocol change).?
```
davidtaikocha: If we want to avoid protocol changes, maybe we can let proposer submit `ProposalManifest` in `BlobSlice`, and force-inclusion submitter submit `DerivationSourceManifest`?

packages/protocol/docs/Derivation.md
```diff
@@ -177,37 +187,64 @@ struct BlobSlice {
...
+For each `DerivationSource[i]`, the following steps are performed independently:
 
-A default manifest is returned when any of the following validation criteria fail:
+1. **Blob Validation**: Verify `blobSlice.blobHashes.length > 0`
+2. **Offset Validation**: Verify `blobSlice.offset <= BLOB_BYTES * blobSlice.blobHashes.length - 64`
+3. **Version Extraction**: Extract version from bytes `[offset, offset+32)` and verify it equals `0x1`
+4. **Size Extraction**: Extract data size from bytes `[offset+32, offset+64)`
+5. **Decompression**: Apply ZLIB decompression to bytes `[offset+64, offset+64+size)`
+6. **Decoding**: RLP decode the decompressed data:
+   - If `i == 0`: Decode into a structure containing `proverAuthBytes` and `DerivationSourceManifest[0]`
```
davidtaikocha: Under the current `propose()` implementation, i think `i == len(DerivationSource) - 1` should contain `proverAuthBytes` instead of `i == 0`?

packages/protocol/docs/Derivation.md
ggonzalez94: this makes sense to me

packages/protocol/docs/Derivation.md
```diff
@@ -168,7 +173,7 @@ To maintain the integrity of the proposal process, the `proposer` address must p
 - Confirming that the proposer holds a sufficient balance in the L2 BondManager contract.
 - Ensuring the proposer is not waiting for exiting.
 
-Should any of these validation checks fail (as determined by the `updateState` function returning `isLowBondProposal = true`), the proposal is replaced with the default manifest, which contains a single empty block with no transactions.
+Should any of these validation checks fail (as determined by the `updateState` function returning `isLowBondProposal = true`), the proposal is replaced with the default manifest, which contains a single block with only an anchor transaction.
```
davidtaikocha: > the proposal is replaced with the default manifest

Maybe here needs to be sth like: the proposer's `DerivationSource` is replaced with the default manifest?

If we invalidate the whole proposal, then all forced-included sources will be discarded as well (if it's a low bond proposal)?

PR #20329, opened 2025-10-01
chore(protocol): optimize away 1 SSTORE in LibCheckpointStore

dantaik:
## Summary
This PR includes two optimizations to reduce gas costs and eliminate compiler warnings:

### 1. LibCheckpointStore optimization - Reduces 1 SSTORE per checkpoint ⛽
- **Before**: Stored full `Checkpoint` struct (blockNumber, blockHash, stateRoot) = 3 storage slots
- **After**: Store only `CheckpointRecord` (blockHash, stateRoot) = 2 storage slots
- **Optimization**: `blockNumber` is now derived during reads as `latestCheckpointBlockNumber - offset`
- **Gas savings**: ~20,000 gas per checkpoint saved (1 SSTORE eliminated)

### 2. Test function mutability fixes
- Fixed two compiler warnings (Warning 2018) by restricting function state mutability to `pure`
- `_expectedBlobHashes`: Changed from `view` to `pure` (only operates on memory)
- `_assertForcedSource`: Changed from `view` to `pure` (only uses assertions on memory)

## Changes
- `packages/protocol/contracts/shared/shasta/libs/LibCheckpointStore.sol`
  - Introduced `CheckpointRecord` struct (blockHash + stateRoot only)
  - Modified `saveCheckpoint` to store 2 fields instead of 3
  - Modified `getCheckpoint` to derive `blockNumber` from `latestCheckpointBlockNumber - offset`
- `packages/protocol/test/layer1/shasta/inbox/suite2/propose/AbstractPropose.t.sol`
  - Line 613: `_expectedBlobHashes` mutability → `pure`
  - Line 659: `_assertForcedSource` mutability → `pure`

## Testing
- ✅ Forge build completes successfully without warnings
- ✅ All existing tests pass (no functional changes)
- ✅ Storage optimization verified: 1 fewer SSTORE per checkpoint

### DIFF
--- a/packages/protocol/contracts/shared/shasta/iface/ICheckpointStore.sol
+++ b/packages/protocol/contracts/shared/shasta/iface/ICheckpointStore.sol
@@ -12,9 +12,9 @@ interface ICheckpointStore {
     /// @notice Represents a synced checkpoint
     struct Checkpoint {
         uint48 blockNumber;
-        /// @notice The block hash for the end (last) L2 block in this proposal.
+        /// @notice The block hash for the end (last) block in this proposal.
         bytes32 blockHash;
-        /// @notice The state root for the end (last) L2 block in this proposal.
+        /// @notice The state root for the end (last) block in this proposal.
         bytes32 stateRoot;
     }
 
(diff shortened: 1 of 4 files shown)


dantaik:
@ggonzalez94 I just realized this won't work -- we have to save the block id as part of the Checkpoint in the mapping **if the mapping is a ring buffer**

But the technique here can be used in your PR -- using CheckpointRecord

PR #20340, opened 2025-10-02
fix(protocol): use SHA-256 precompile in hash

radik878:
Replace abi.encodePacked with assembly-based SHA-256 (precompile 0x02) in hash to avoid allocations


dantaik:
@claude please review

radik878:
claude knows the deal

AnshuJalan (approved):

dantaik:
We shall always use keccak256 in stead of sha-256 unless we must produce standard SHA-256 for protocol interoperability.

PR #20342, opened 2025-10-02
fix(protocol): remove misleading treasury/anchorGasLimit comments

Galoretka:
- Remove outdated references to treasury and anchorGasLimit from _getConfig.
- Clarify that treasury is handled at L2/circuits level, not via ITaikoInbox.Config.
- Note that anchorGasLimit is not part of ITaikoInbox.Config; use blockMaxGasLimit and derivation rules instead.


dantaik:
Lets keep Pacaya code as-is as they are going to deprecate very soon.

PR #20345, opened 2025-10-02
fix(bridge-ui): prevent SSR crash by guarding theme store localStorage access

maradini77:
### Summary
- Make theme store initialization SSR-safe by avoiding module-level localStorage access.

### Changes
- Initialize `theme` store with a safe default on the server.
- Read and apply saved theme only in the browser (guarded by runtime check).

### Why
- Prevents SSR/ prerender crashes (ReferenceError on `localStorage`).
- Reduces FOUC risk and stabilizes tests/CI.

### DIFF
--- a/packages/bridge-ui/src/stores/theme.ts
+++ b/packages/bridge-ui/src/stores/theme.ts
@@ -5,4 +5,13 @@ export const enum Theme {
   LIGHT = 'light',
 }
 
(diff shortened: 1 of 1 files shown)


KorbinianK:
@maradini77 I've not seen any of the ssr crashes you mention. Where do you experience those? Stores can't be serverside anyway, so local storage wouldn't cause an issue either way. Plus the fact that it doesn't use svelte's own browser check makes this PR a bit sus aka AI slop.

PR #20346, opened 2025-10-02
refactor(ejector): avoid cloning beacon URL when constructing client

rnkrtt:
Description:
- Reuse the existing `Url` passed to `BeaconClient::new` by handing references to the helper fetch functions.
- Removes two redundant `Url::clone` allocations while keeping behaviour identical.


davidtaikocha:
Closed https://github.com/taikoxyz/taiko-mono/blob/main/CONTRIBUTING.md.

PR #20347, opened 2025-10-02
refactor(ejector): migrate metrics module from once_cell to std::sync::LazyLock

rnkrtt:
Migrate metrics module from once_cell to std::sync::LazyLock
Replaces once_cell::Lazy with std::sync::LazyLock for all static metrics variables. Uses standard library solution instead of external dependency.


davidtaikocha:
Closed: https://github.com/taikoxyz/taiko-mono/blob/main/CONTRIBUTING.md .

PR #20353, opened 2025-10-03
chore: fix wrong usage of `fsPromises.readdir`

zkpepe:
turns out `fsPromises.readdir` doesn’t take a callback..
now it’s just awaited in a `try/catch`, so `filesName` properly comes back as an array like `["1.png", "2.png", "3.jpg"]`.
this was breaking file loading before, so it’s kind of a big deal.


davidtaikocha:
Closed due to deprecated package.

PR #20355, opened 2025-10-03
chore: Fix Swagger UI preset import for latest versions

defiberrys:
Updated the import to use `SwaggerUIStandalonePreset` directly instead of `SwaggerUIBundle.SwaggerUIStandalonePreset`.
In recent Swagger UI versions, `SwaggerUIStandalonePreset` is no longer a property of `SwaggerUIBundle`, so this change ensures compatibility and prevents runtime errors.

### DIFF
--- a/packages/taiko-client/docs/index.html
+++ b/packages/taiko-client/docs/index.html
@@ -1,18 +1,30 @@
-<script>
-    window.onload = function () {
-        // Begin Swagger UI call region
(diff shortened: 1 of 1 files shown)


RogerLamTd:
Fix does not work,
```
(index):20 Uncaught ReferenceError: SwaggerUIStandalonePreset is not defined
    at window.onload ((index):20:25)
```

PR #20370, opened 2025-10-05
chore: Remove Uncatchable os.Kill Signal from Balance Monitor CLI

aso20455:
Drop os.Kill from the signal.Notify list in the balance monitor CLI utility because SIGKILL cannot be intercepted in Go; the change removes unreachable code without altering runtime behavior.


davidtaikocha:

PR #20386, opened 2025-10-08
chore(protocol): remove unused checkpointManager and MockCheckpointProvider

radik878:
- Delete unused checkpointManager field and its initialization in _setupMocks.
- Drop MockCheckpointProvider from imports.
- This code was not used anywhere: the Inbox persists checkpoints internally via LibCheckpointStore, and tests construct ICheckpointStore.Checkpoint structs directly without a manager instance. Keeping the mock created dead code and added confusion


davidtaikocha:
@ggonzalez94 shall we let AI check all these kinds of issues and push a PR to address all of them? Not for everyone, but i think most of them are also using AI and farming.

dantaik:
Please merge with upstream changes.

PR #20389, opened 2025-10-08
fix: prevent ethclient resource leak in InitFromConfig

hawkadrian:
Added proper cleanup for ethclient connections when initialization fails. 

Implemented Close method to properly release resources on application shutdown. 

This prevents connection leaks that could accumulate over multiple initialization attempts.


davidtaikocha:
We shouldn't close clients in `InitFromConfig`'s `defer`.

PR #20397, opened 2025-10-09
chore(protocol): slightly refactor signal service 

dantaik:
## Summary
Refactors Inbox to use `_checkpointStore` (internal immutable) and optimizes SignalServiceShasta with `require` statements for better gas efficiency.

## Key Changes

**Inbox.sol**
- `signalService` → `_checkpointStore` (internal immutable, moved to immutable section)

**SignalServiceShasta.sol**  
- Converted all if-revert to `require` with custom errors
- Inlined checkpoint retrieval logic (removed `_getCheckpoint` helper)
- Optimized `_verifySignalReceived` to use storage pointer with local `stateRoot`

**Tests**
✅ All 511 layer1 tests passing


davidtaikocha (dismissed):

packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
```diff
@@ -18,7 +18,7 @@ contract SignalServiceShasta is EssentialContract, ISignalServiceShasta {
 
     /// @notice Storage-optimized checkpoint record with only persisted fields
     struct CheckpointRecord {
-        /// @notice The block hash for the end (last) block in this proposal.
+        /// @notice The block hash for he end (last) block in this proposal.
```
dantaik:
```suggestion
        /// @notice The block hash for the end (last) block in this proposal.
```

packages/protocol/contracts/shared/shasta/impl/SignalServiceShasta.sol
```diff
@@ -159,33 +160,21 @@ contract SignalServiceShasta is EssentialContract, ISignalServiceShasta {
...
-
-    /// @dev Gets a checkpoint by block number
-    /// @param _blockNumber The block number of the checkpoint
-    /// @return checkpoint The checkpoint
-    function _getCheckpoint(uint48 _blockNumber)
-        private
-        view
-        returns (Checkpoint memory checkpoint)
     {
         CheckpointRecord storage record = _checkpoints[_blockNumber];
```
dantaik:
```suggestion
        CheckpointRecord memory record = _checkpoints[_blockNumber];
```

dantaik:
@claude please review

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -326,7 +326,7 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
         config_ = IInbox.Config({
             codec: _codec,
             bondToken: address(_bondToken),
-            signalService: address(signalService),
+            signalService: address(_checkpointStore),
```
ggonzalez94: if we are naming it signalService on the config then it might make sense to use the same for the immutable variable? I prefer signalService since we know that will always be the case, but I'm fine if we change it in both places to checkPointStore as well, not a strong opinion

packages/protocol/contracts/shared/signal/ISignalServiceShasta.sol
```diff
@@ -55,7 +55,7 @@ interface ISignalServiceShasta is ICheckpointStore {
     /// @param _signal The signal (message) to send.
     /// @param _proof Merkle proof that the signal was persisted on the
     /// source chain. If this proof is empty, then we check if this signal has been marked as
-    /// received by TaikoL2.
+    /// received by calling proveSignalReceived.
```
ggonzalez94: nit for clarity:
```suggestion
    /// received before. This happens if someone previously called `proveSignalReceived` for the same message.
```

packages/protocol/contracts/shared/signal/ISignalServiceShasta.sol
```diff
@@ -74,7 +74,7 @@ interface ISignalServiceShasta is ICheckpointStore {
...
```
ggonzalez94: same as a above. I think this comment should specify that proof can be empty as an optimization if the signal has been marked as received before
```suggestion
    /// received before. This happens if someone previously called `proveSignalReceived` for the same message.
```

packages/protocol/contracts/shared/signal/SignalServiceShasta.sol
```diff
@@ -160,32 +161,20 @@ contract SignalServiceShasta is EssentialContract, ISignalServiceShasta {
...
-    }
-
-    // ---------------------------------------------------------------
-    // Internal Functions
-    // ---------------------------------------------------------------
-
-    /// @dev Gets a checkpoint by block number
-    /// @param _blockNumber The block number of the checkpoint
-    /// @return checkpoint The checkpoint
-    function _getCheckpoint(uint48 _blockNumber)
```
ggonzalez94: do we really save gas by inlining `_getCheckpoint` instead of reusing it? I know it is only a few lines of code, but I noticed this is a pattern we have been moving towards to inline private functions. And as far as I knew private functions shouldn't increase gas usage since they get inlined anyways(or the gas difference is probably minimal and dependent on the compiler)

ggonzalez94 (commented):
Looks good overall. Left a few comments

packages/protocol/contracts/shared/signal/SignalServiceShasta.sol
dantaik: inlining it will allow use to avoid a SLOAD in certain cases.

PR #20404, opened 2025-10-09
feat(protocol): L2 upgrade script to Shasta

YoGhurt111:


davidtaikocha (approved):

packages/protocol/script/layer2/based/UpgradeShastaL2.s.sol
```diff
@@ -0,0 +1,34 @@
...
+        vm.startBroadcast(privateKey);
+        _;
+        vm.stopBroadcast();
+    }
+
+    function run() external broadcast {
+        UUPSUpgradeable(taikoAnchor).upgradeTo(
+            address(
+                new TaikoAnchor(
+                    125e9, 125e9, signalService, pacayaForkHeight, shastaForkHeight, bondManager
```
ggonzalez94, line 29: shall we put the first two parameters as constants somewhere? also the provability bond should be higher than the liveness bond(but since this is for testnet this is not important for now).
Also the signal service has changed as well, should we include tit in this script?

ggonzalez94 (commented):
left one comment

packages/protocol/script/layer2/based/UpgradeShastaL2.s.sol
YoGhurt111: Currently this script is just a template for the upgrade process.
The contract is only deployed for the compatible version of the node to be able to launch Taiko Hoodi, and will never trigger the logic of Shasta.
So I think we can leave it there and change it later when we decide what value we actually want to use.
And for the new signal service, we'd like to test it on devnet first.

PR #20412, opened 2025-10-09
fix(balance-monitor): add missing L2 ERC-20 balance gauge

wedjob0X:
Added missing L2 ERC-20 balance gauge. Previously all ERC-20 balances were incorrectly recorded in the L1 metric.

- Add `l2Erc20BalanceGauge` for proper L2 ERC-20 monitoring
- Split balance tracking logic to use correct gauges for L1/L2
- Separate logging for L1 and L2 ERC-20 balances


davidtaikocha:
No L2 ERC-20 balance gauge is by design.

PR #20414, opened 2025-10-09
fix(taiko-client): initialize random seed to prevent predictable jitter

anim001k:
### Description

Fixes predictable proposer interval jitter by initializing the global `math/rand` seed in `InitFromConfig`. 

**Problem**: The proposer's random interval jitter (12-120 seconds) was deterministic across process restarts, leading to:
- Synchronized proposal attempts from multiple instances
- Increased collision probability 
- Potential MEV exploitation through predictable timing

**Solution**: Added `rand.Seed(time.Now().UnixNano())` during proposer initialization to ensure truly random jitter between restarts.

**Impact**: Improves network stability and reduces proposal conflicts by making each proposer instance's timing unpredictable.

### DIFF
--- a/packages/taiko-client/proposer/proposer.go
+++ b/packages/taiko-client/proposer/proposer.go
@@ -95,6 +95,9 @@ func (p *Proposer) InitFromConfig(
 	p.Config = cfg
 	p.lastProposedAt = time.Now()
 
+	/ Seed the global PRNG once to avoid predictable jitter across restarts.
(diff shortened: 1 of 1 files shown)


davidtaikocha:
It's only a fallback proposer when there is no functional preconfers, and this random interval feature won't be opened, so its ok here. And probly we will switch to use https://github.com/taikoxyz/taiko-mono/tree/main/packages/taiko-client-rs/crates/proposer later :)

PR #20416, opened 2025-10-10
refactor(protocol): make ShastaAnchor inherit EssentialContract directly

dantaik:
## Summary

Refactors ShastaAnchor to inherit directly from EssentialContract instead of PacayaAnchor, consolidating the inheritance hierarchy and improving code organization.

**Key changes:**
- Changed ShastaAnchor inheritance from PacayaAnchor to EssentialContract
- Moved necessary state variables and functions from PacayaAnchor to ShastaAnchor
- Converted State struct from storage to return-only struct with optimized slot packing (5 slots + 45 gap = 50 total)
- Renamed `signalService` to `checkpointStore` using ICheckpointStore interface
- Renamed `updateState` to `anchor` for clarity
- Merged `getDesignatedProver` and `_getDesignatedProver` into single public function
- Made TaikoAnchor inherit from PacayaAnchor instead

**Cleanup:**
- Removed unused code: `getBasefeeV2`, `skipFeeCheck`, `IBlockHashProvider`, `_trackParentBlockHash`, `getState`, `_blockhashes`
- Updated error naming convention from `L2_ABC_XYZ` to `AbcXyz` format
- Improved BondManager documentation for `withdrawalDelay`

**Storage optimization:**
- Optimal slot packing: `address(20) + uint48(6) + uint48(6) = 32 bytes`
- Maintains 50 total storage slots for upgradability

## Test plan

- [x] Verify all contracts compile successfully
- [ ] Run protocol test suite: `pnpm --filter @taiko/protocol test`
- [ ] Verify ShastaAnchor functionality with integration tests
- [ ] Validate storage layout matches expectations

### DIFF
--- a/.github/workflows/ejector--docker.yml
+++ b/.github/workflows/ejector--docker.yml
@@ -17,8 +17,7 @@ concurrency:
 env:
   # where to push
   REGISTRY_IMAGE: us-docker.pkg.dev/evmchain/images/ejector
-  # build context MUST be repo root for cargo workspace
-  BUILD_CONTEXT: .
+  BUILD_CONTEXT: packages/ejector
   DOCKERFILE: packages/ejector/Dockerfile
 
 jobs:
--- a/packages/ejector/Dockerfile
+++ b/packages/ejector/Dockerfile
@@ -4,17 +4,14 @@ WORKDIR /app
 RUN apt-get update && apt-get install -y --no-install-recommends \
     pkg-config libssl-dev ca-certificates && \
     rm -rf /var/lib/apt/lists/*
-COPY Cargo.toml Cargo.lock ./
-COPY packages ./packages
-RUN cargo build -p ejector --release
 
-# Use the repo root as build context so we can copy the workspace files:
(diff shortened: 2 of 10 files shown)


dantaik:
- High – publicInputHash is never verified or advanced in anchor, so the guard that previously prevented multiple
  anchor calls per block is gone (contracts/layer2/based/ShastaAnchor.sol:227-284). We read the stored value into
  state.publicInputHash but never recompute it with _calcPublicInputHash or check it against an expected current
  value; the write-back on line 279 just stores the same value again. Because the golden-touch private key is public,
  an attacker can re-invoke anchor later in the same L2 block (for example with _blockIndex = 1) and overwrite
  checkpointStore.saveCheckpoint with arbitrary L1 data or force extra bond transfers without triggering any revert. This
  is the regression that PacayaAnchor._verifyAndUpdatePublicInputHash and _trackParentBlockHash previously prevented.

  Open questions

  - Was the intent to compute and enforce the public-input hash inside this refactored contract, or is another contract
  expected to supply that guard now? If the former, we’ll need to reintroduce

PR #20421, opened 2025-10-10
fix(protocol): refactor Anchor.sol to remove Pacaya fork references

dantaik:
This is a big refractory of Shasta's Anchor contract assuming it will be deployed at a different address.

@ggonzalez94 the contract may still be buggy and it seems we are still assuming a block height based fork.


packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -30,12 +30,13 @@ contract Anchor is EssentialContract {
...
-        bytes32 publicInputHash;
+        // Proposal-level state
         bytes32 bondInstructionsHash;
         address designatedProver;
-        uint48 anchorBlockNumber;
-        uint48 endOfSubmissionWindowTimestamp;
         bool isLowBondProposal;
+        // Block-level state
+        uint48 anchorBlockNumber;
+        bytes32 ancestorsHash;
```
ggonzalez94, line 39: is `ancestorHash` new?

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -30,12 +30,13 @@ contract Anchor is EssentialContract {
 
     /// @notice State containing all non-mapping state variables.
     struct State {
-        bytes32 publicInputHash;
+        // Proposal-level state
         bytes32 bondInstructionsHash;
         address designatedProver;
-        uint48 anchorBlockNumber;
-        uint48 endOfSubmissionWindowTimestamp;
```
ggonzalez94, line 37: did we remove `endOfSubmissionWindowTimestamp`?

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -328,208 +271,272 @@ contract Anchor is EssentialContract {
...
+        bytes calldata _proverAuth,
+        address _currentDesignatedProver
     )
         public
         view
         returns (bool isLowBondProposal_, address designatedProver_, uint256 provingFeeToTransfer_)
     {
-        return _getDesignatedProver(_proposalId, _proposer, _proverAuth, designatedProver);
-    }
+        (address candidate, uint48 provingFeeGwei) =
```
ggonzalez94, line 293: not necessary that we address it on this PR, but I don't think we need the prover fee to be in gwei and is less error prone if we just use wei instead

ggonzalez94 (approved):
LGTM. Left a couple questions I was unsure about and:
1. for block vs timestamp fork created an issue #20426 
2. I think there are a couple other refactors we can do to make the code more clear, made an attempt at #20427

packages/protocol/contracts/layer2/core/Anchor.sol
dantaik: it's called public input hash, which is less accurate than ancestorHash

dantaik:
Given that you have another issue opened and another PR to be merged on top of this PR. I'll go ahead merge this PR into yours.

PR #20413, opened 2025-10-09
refactor(protocol): clean up folder structure and remove Pacaya contracts

ggonzalez94:
Fixes #20014 

## Summary
This PR removes deprecated Pacaya-specific contracts and cleans up the protocol folder structure by flattening the layer1 directory hierarchy and consolidating the verifier architecture. This significantly reduces codebase complexity (~12K lines removed) while maintaining all current functionality.

## Changes
- Removed Pacaya contracts (TaikoInbox, DevnetInbox, MainnetInbox) and associated tests (~8K lines)
- Removed deprecated infrastructure (ForkRouter, PreconfRouter, PreconfSlasher, ProverSet, ForcedInclusionStore)
- Flattened layer1 structure by moving contracts from `layer1/shasta/` to `layer1/` root
- Simplified verifier architecture by consolidating verifier contracts and removing legacy implementations
- Updated deployment scripts and storage layout generation for new structure

---

> [!NOTE]
> Removes Pacaya-era contracts and infra, flattens L1 (shasta → core) with new SignalService/Inbox/Anchor paths, consolidates verifiers, renames DAO controller, updates scripts/layouts/tests, and drops deprecated docs/sims.
> 
> - **Structure & Removals**:
>   - Drop Pacaya-era contracts and infra: `TaikoInbox/DevnetInbox/MainnetInbox`, `ForkRouter`, `PreconfRouter/Slasher`, `ProverSet`, `ForcedInclusionStore`, related tests/docs/simulation.
>   - Flatten L1: migrate `layer1/shasta/*` → `layer1/core/*` (e.g., `Inbox`, `Codec*`, libs) and update imports.
>   - Move shared vaults: `shared/tokenvault/*` → `shared/vault/*`.
> - **Signal/Inbox/Anchor**:
>   - Introduce `SignalService` with `authorizedSyncer`/`remoteSignalService` and `ICheckpointStore`; adapt tests to checkpoint-based proving.
>   - Replace L2 `TaikoAnchor` with `core/Anchor` and update tests.
> - **Verifiers**:
>   - Consolidate verifiers (`Risc0Verifier`, `SP1Verifier`, `SgxVerifier`) and remove legacy `Taiko*Verifier` variants.
> - **Governance**:
>   - Rename `TaikoDAOController` → `MainnetDAOController`; update tests and docs.
> - **Build/Tooling**:
>   - Update deploy scripts (`script/layer1/core/*`) and layout generator paths; adjust foundry `profile.lint`.
[...]

PR #20432, opened 2025-10-13
refactor(protocol): convert verifiers to view-only functions

YoGhurt111:
Consider removing the dependency on TaikoInbox address on verifiers after the design of separating into `ShastaInbox` and `PacayaInbox`.


packages/protocol/contracts/layer1/verifiers/TaikoSgxVerifier.sol
```diff
@@ -85,12 +84,11 @@ contract TaikoSgxVerifier is EssentialContract, IVerifier {
     error SGX_INVALID_PROOF();
 
     constructor(
-        address _taikoInbox,
+        uint64 _taikoChainId,
         address _taikoProofVerifier,
```
dantaik: we can remove taikoProofVerifier as it is not used anywhere

YoGhurt111: Will do

YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/20432/commits/4df4736baf78d57447180c7dc420f6f0f96e16ee

davidtaikocha (approved):

packages/protocol/contracts/layer1/devnet/verifiers/OpVerifier.sol
```diff
@@ -13,15 +11,9 @@ import "../../verifiers/IVerifier.sol";
...
-    address public immutable taikoInbox;
-    address public immutable proofVerifier;
-
     uint256[50] private __gap;
 
-    constructor(address _taikoInbox, address _proofVerifier) {
-        taikoInbox = _taikoInbox;
-        proofVerifier = _proofVerifier;
-    }
+    constructor() { }
```
dantaik:
```suggestion
```

dantaik (changes requested):

dantaik (approved):

ggonzalez94 (changes requested):
@YoGhurt111 all the new verifiers already use view functions(e.g. SP1 verifier [here](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/shasta/verifiers/ShastaSP1Verifier.sol#L49)) and the previous ones are being removed from the repo as part of #20413 
Is there anything we should port from here to the most recent changes?

YoGhurt111:
> @YoGhurt111 all the new verifiers already use view functions(e.g. SP1 verifier [here](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/shasta/verifiers/ShastaSP1Verifier.sol#L49)) and the previous ones are being removed from the repo as part of #20413 Is there anything we should port from here to the most recent changes?

Great! Removing some unused import would be better, like `import "src/shared/libs/LibNames.sol";`

YoGhurt111:
Closed by https://github.com/taikoxyz/taiko-mono/pull/20413

PR #20435, opened 2025-10-13
fix(relayer): correctly handle decodeABI errors in types.go

prestoalvarez:
### Description
- **Problem**: `decodeABI` errors were gated by `&&` with length checks, allowing invalid decodes to slip through when `err != nil` but lengths matched.
- **Solution**: Replace `&&` with `||` to fail fast on any decode error or unexpected value count.
- **Why**: Prevents using corrupted/partial data, improves correctness and reliability.


cyberhorsey:
this doesnt look like the correct fix. there should be a separate check for the values length and an error returned. otherwise, err could be non-nil but the values could be the wrong legnth?

PR #20437, opened 2025-10-14
fix(protocol): align the bond name and usage

YoGhurt111:


YoGhurt111:
@claude Plz review again

PR #20441, opened 2025-10-14
fix(taiko-client): prevent panic in NewFixedKSigner by using hexutil.Decode

sashass1315:
- Problem: NewFixedKSigner used hexutil.MustDecode(privKey), which panics on invalid/malformed hex, violating the function’s safe (FixedKSigner, error) contract and potentially crashing the process.
- Fix: Replace MustDecode with hexutil.Decode and propagate the returned error. Keep overflow/zero checks intact and avoid echoing the full private key in error messages.


davidtaikocha:
That should never happen, and if that happens, a panic is fine for the software.

PR #20444, opened 2025-10-14
fix(bridge-ui): provide SSR-safe storage fallback to prevent localStorage errors

kilavvy:
### Summary
- Replace direct `localStorage` usage with an SSR-safe fallback in `packages/bridge-ui/src/libs/storage/index.ts` to avoid runtime errors during server-side rendering.

### Why
- SvelteKit renders on the server where `localStorage` is undefined, causing crashes when services are instantiated at import-time.

### Changes
- Introduced a no-op `safeStorage` fallback and passed it to `BridgeTxService` and `CustomTokenService`.

### DIFF
--- a/packages/bridge-ui/src/libs/storage/index.ts
+++ b/packages/bridge-ui/src/libs/storage/index.ts
@@ -1,6 +1,18 @@
 import { BridgeTxService } from './BridgeTxService';
 import { CustomTokenService } from './CustomTokenService';
 
-export const bridgeTxService = new BridgeTxService(globalThis.localStorage);
+// Provide a no-op storage fallback for SSR environments where localStorage is unavailable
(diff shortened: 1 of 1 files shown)


KorbinianK:
Closing this PR. While I understand the intent to avoid SSR crashes by adding a no-op storage fallback, this approach isn’t suitable for our codebase:

- **SSR stores are not possible in SvelteKit.** Adding fallback logic for localStorage here is unnecessary.
- **Direct casting with `as any` is poor practice** and makes the code harder to maintain and reason about.
- **Instantiating services at import-time with environment detection** clutters the index file and is not a scalable solution.
- This change feels like a generic AI-generated workaround rather than a well-considered architectural improvement.

Let’s keep the code clean and only add logic that’s genuinely required for our framework and runtime.  
Thanks for the effort, but this isn’t the right direction for our project.

PR #20445, opened 2025-10-14
fix(taiko-client): handle reorgs and send to preconf api

cyberhorsey:


packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -1014,34 +1014,81 @@ func (s *PreconfBlockAPIServer) LatestSeenProposalEventLoop(ctx context.Context)
...
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	if s.latestSeenProposal != nil {
+		log.Info("Reset latest seen proposal due to chain update")
+	}
+
+	s.latestSeenProposal = nil
+	s.highestUnsafeL2PayloadBlockID = 0
+	metrics.DriverLastSeenBlockInProposalGauge.Set(0)
```
MatusKysel, line 1055: shouldn't we also set also set `metrics.DriverHighestPreconfUnsafePayloadGauge.Set(0)` or call `updateHighestUnsafeL2Payload(0)` ?

packages/taiko-client/driver/chain_syncer/event/syncer.go
```diff
@@ -535,6 +549,182 @@ func (s *Syncer) processPacayaBatch(
...
+
+// cloneLastSeenProposalForSyncer clones the pointer to a proposal and sets PreconfChainReorged
+// to true on the clone
+func cloneLastSeenProposalForSyncer(src *encoding.LastSeenProposal) *encoding.LastSeenProposal {
+	if src == nil {
+		return nil
+	}
+
+	cloned := *src
+	cloned.PreconfChainReorged = true
```
MatusKysel, line 828: as this always set PreconfChainReorged = true after insert, so `DriverReorgsByProposalCounter` counter will track proposals, not reorgs

MatusKysel (approved):
LGTM - left 2 comments

packages/taiko-client/driver/chain_syncer/event/syncer.go
cyberhorsey: but its only emitted after a reorg, right?

cyberhorsey: but its only emitted after a reorg, right?

packages/taiko-client/driver/chain_syncer/event/syncer.go
```diff
@@ -484,6 +496,7 @@ func (s *Syncer) processPacayaBatch(
 			s.state.SetL1Current(reorgCheckResult.L1CurrentToReset)
 			s.lastInsertedBatchID = reorgCheckResult.LastHandledBatchIDToReset
 			s.reorgDetectedFlag = true
+			s.handlePacayaReorg(reorgCheckResult)
```
YoGhurt111, line 511: IMO, it doesn't help that case.Because we can't receive the event after rerog to trigger this reorg check.

packages/taiko-client/driver/chain_syncer/event/syncer.go
MatusKysel: to me it seems it's used also for normal insertion of blocks https://github.com/taikoxyz/taiko-mono/blob/preconf_handle_l1_reorg/packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go#L259

packages/taiko-client/driver/chain_syncer/event/syncer.go
cyberhorsey: please review changes

packages/taiko-client/driver/chain_syncer/event/syncer.go
cyberhorsey: Ah ,thats maybe deceptive.

https://github.com/taikoxyz/taiko-mono/blob/preconf_handle_l1_reorg/packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go#L201

if the batch was succesfully preconfirmed, *we return nil out of this function early*.

If the batch was NOT succesfully preceonfirmed, then it continues, inserts all new blocks, and sends the latest seen proposal and marks it as a PreconfChainReorged.

https://github.com/taikoxyz/taiko-mono/blob/preconf_handle_l1_reorg/packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go#L113

we send the latestSeenProposal at the start before checking as well, so this one is expected (no reorg). In the case of a reorg, we re-send, and mark it as a reorg.

MatusKysel: ahh, you are right

PR #20448, opened 2025-10-14
feat(blobindexer-rs): blobindexer in rust

cyberhorsey:
https://github.com/taikoxyz/taiko-mono/issues/20446

we will need this as we should run a `supernode` anad then provide URL to our preconfirmation sidecars in order to fetch blob data, since this will not be available for regular beacon nodes anymore after Fusaka. Others can also run thier own `blobindexer-rs` and attach it to the driver, if they have a supernode, but the resource requirements are very large so we shouldn't expect it (and they could call the supernode directly). 

Will try running it tonight.


cyberhorsey:
wip, adding features to only listen for blobs on specific addresses (minimize resource usage) plus bug fixes and dB optimizations

davidtaikocha (approved):

davidtaikocha:
Also opened a PR in our k8s-config: https://github.com/taikoxyz/k8s-configs/pull/734

packages/blobindexer-rs/src/indexer.rs
```diff
@@ -0,0 +1,559 @@
...
+    pub async fn run(self) -> Result<()> {
+        let poll_interval = self.config.poll_interval;
+
+        loop {
+            if self.shutdown.is_cancelled() {
+                break;
+            }
+
+            if let Err(err) = self.tick().await {
+                tracing::error!(error = ?err, "indexer tick failed");
```
MatusKysel, line 163: i would add more context, eg slot

packages/blobindexer-rs/src/http.rs
```diff
@@ -0,0 +1,243 @@
...
+
+#[derive(Serialize)]
+struct ErrorBody {
+    message: String,
+}
+
+type RestResult<T> = std::result::Result<T, ApiError>;
+
+async fn health() -> &'static str {
+    "ok"
```
MatusKysel, line 116: maybe we should check if beacon node is sync also ? or something more sophisticated ?

packages/blobindexer-rs/src/http.rs
```diff
@@ -0,0 +1,243 @@
...
+    let response = BlobServerResponse::try_from(blob)?;
+    Ok(Json(response))
+}
+
+async fn list_blobs_by_slot(
+    State(state): State<AppState>,
+    Path(slot): Path<u64>,
+) -> RestResult<Json<Vec<BlobServerResponse>>> {
+    let slot_i64 = i64::try_from(slot).map_err(|_| ApiError::bad_request("slot out of range"))?;
+    let _block = state
```
MatusKysel: unused

MatusKysel: maybe let's add comment that we want to fetch it or something

packages/blobindexer-rs/src/storage/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+        canonical: bool,
+    ) -> Result<()> {
+        for slot in slots {
+            sqlx::query(
+                "UPDATE blocks SET canonical = ?, updated_at = CURRENT_TIMESTAMP WHERE slot = ?",
+            )
+            .bind(canonical)
+            .bind(slot)
+            .execute(&mut **tx)
+            .await?;
```
MatusKysel: maybe this could be reworked to batch for better performnace `UPDATE with WHERE slot IN (...)`

MatusKysel (approved):

packages/blobindexer-rs/src/http.rs
cyberhorsey: this should just be the helath check for the HTTP server itself - we should probably add a different endpoint to check the status of the beacon node itself that can return something more sophisticated

packages/blobindexer-rs/src/http.rs
cyberhorsey: removed

packages/blobindexer-rs/src/indexer.rs
cyberhorsey: i added some logs to the processing / refreshing

packages/blobindexer-rs/src/storage/mod.rs
cyberhorsey: tried that, please take a look

packages/blobindexer-rs/src/http.rs
cyberhorsey: i added a status endpoint

```
curl http://localhost:9000/v1/status
{"beaconSynced":true,"beacon":{"headSlot":1535900,"finalizedSlot":1535808,"syncDistance":0,"isSyncing":false,"isOptimistic":false,"elOffline":false},"lastProcessedSlot":1529204}
```

RogerLamTd (approved):

PR #20456, opened 2025-10-15
fix(taiko-client): initialize `reorgRewindDepth` from config in `BlockBatchIterator`

Galoretka:
The BlockBatchIterator never initialized reorgRewindDepth from BlockBatchIteratorConfig, leaving it at 0 even when callers intended a non-zero rewind. As a result, on reorg detection the iterator did not rewind to earlier stable blocks and could miss events that occurred before the fork point. This change reads ReorgRewindDepth from the config during construction so the iterator rewinds by the configured number of blocks on reorg. If the field is nil, the previous behavior (no rewind) is preserved.


davidtaikocha (approved):

MatusKysel (approved):
lgtm

PR #20461, opened 2025-10-16
feat(protocol): benchmark and improve lookahead store

AnshuJalan:


packages/protocol/contracts/layer1/preconf/impl/LookaheadStore.sol
```diff
@@ -1,68 +1,58 @@
...
+    /// is stored.
+    function checkProposer(address _proposer, bytes calldata _lookaheadData)
         external
         returns (uint48)
     {
+        // Must only be called the inbox
+        // The inbox should set `_proposer` to the current L2 batch proposer's address
         require(msg.sender == inbox, NotInbox());
 
         LookaheadData memory data = abi.decode(_lookaheadData, (LookaheadData));
```
ggonzalez94, line 55: wouldn't it make sense to use the optimized library to decode the entire LookaheadData struct instead of doing `abi.decode` here and then optimizing decoding on individual fields? or is there anything specific about the shape of the struct that makes it better to do individual decoding?

PR #20462, opened 2025-10-16
fix(prover): close HTTP response body on non-200 to prevent connection leaks

sashaodessa:
### Summary
Closes HTTP response body on error path to avoid leaking connections and ensure the client’s connection pool is reused properly.


MatusKysel:
it's done here https://github.com/taikoxyz/taiko-mono/blob/blob_indexer_rs/packages/taiko-client/prover/proof_producer/common.go#L77

PR #20464, opened 2025-10-16
fix(taiko-client): monitor the latest proposal on chain

YoGhurt111:
refer to https://github.com/taikoxyz/taiko-mono/pull/20445/files
A more direct way to solve the lastseen proposal reorg problem.


packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -1026,17 +1029,67 @@ func (s *PreconfBlockAPIServer) GetSequencingEndedForEpoch(epoch uint64) (common
...
+		log.Error("Failed to filter batch proposed event", "err", err)
+		return
+	}
+	defer iterPacaya.Close()
+
+	for iterPacaya.Next() {
+		s.recordLatestSeenProposal(&encoding.LastSeenProposal{
+			TaikoProposalMetaData: metadata.NewTaikoDataBlockMetadataPacaya(iterPacaya.Event),
+			PreconfChainReorged:   true,
+		})
```
davidtaikocha: Shall we check if the batch id in this `iterPacaya.Event` is less than the current `latestSeenProposal`? To prevent this `monitorLatestPacayaProposalOnChain` process the same event faster than the event-syncer?

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -1026,17 +1029,67 @@ func (s *PreconfBlockAPIServer) GetSequencingEndedForEpoch(epoch uint64) (common
...
+		return
+	}
+
+	latestSeenBatchID := proposal.Pacaya().GetBatchID()
+	latestOnChainBatchID := new(big.Int).SetUint64(numBatches - 1)
+	if latestSeenBatchID.Cmp(latestOnChainBatchID) <= 0 {
+		return
+	}
+
+	iterPacaya, err := s.rpc.PacayaClients.TaikoInbox.FilterBatchProposed(
```
davidtaikocha: Shall we also fetch the latest Shasta proposal from `s.shastaIndexer`, and check if its reorged or not by calling `inbox.getProposalHash`?

YoGhurt111: I think Shasta's changes to preconf haven't started yet, so I'll wait until Shasta's changes to preconf start.
Actually we need to add extra `lastSeenProposal`, since the shasta proposal ID starts from 0,right?

packages/taiko-client/driver/preconf_blocks/server.go
YoGhurt111: I think no matter who is faster, they will push the last seen proposal to the same level as the on-chain contract.
So even if there are duplicate `recordLatestSeenProposal`, it shouldn't matter?

packages/taiko-client/driver/preconf_blocks/server.go
YoGhurt111: You are right, will change

packages/taiko-client/driver/preconf_blocks/server.go
davidtaikocha: But this message contains `PreconfChainReorged:   true`? It might not will be a reorg for event-syncer?

packages/taiko-client/driver/preconf_blocks/server.go
YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/20464/commits/72a813c1fc37b8bab6859c1cc26d11c44f6b2a2d

cyberhorsey (approved):

packages/taiko-client/driver/preconf_blocks/server.go
cyberhorsey: looks better than mine, approved, we will need a separate PR based off the taiko-alethia-client-v1.11.0 branch though with only pacaya changes so we can deploy it to mainnet as a fix

packages/taiko-client/driver/preconf_blocks/server.go
YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/20464/commits/4072a149149d59291dd52a7ba70d66ff499f3165

davidtaikocha:
@claude plz review.

MatusKysel (approved):
lgtm

packages/taiko-client/driver/preconf_blocks/server.go
YoGhurt111: Created branch `client-1.11.0-hotfix` and cherry-picked changes, will test it on Hoodi tomorrow

cyberhorsey: thank you, LMK how it goes and we should also roll that version out to mainnet as well.

PR #20469, opened 2025-10-17
fix(taiko-client): return sub in ResubscribeErr callback; remove waitSubErr and local Unsubscribe

phrwlk:
This change corrects our use of go-ethereum’s event.ResubscribeErr by removing the blocking waitSubErr call and the local defer sub.Unsubscribe() from the resubscribe callback, and by returning the created subscription immediately. The previous pattern violated the contract of ResubscribeErr, which expects the wrapper—not the callback—to wait on sub.Err() and manage the subscription lifecycle; blocking in the callback and unsubscribing locally could delay cancellation and introduce unsubscribe races. With this update, the wrapper correctly handles retries and lifecycle, external callers manage Unsubscribe(), and the now-dead waitSubErr helper has been removed.

### DIFF
--- a/packages/taiko-client/pkg/rpc/subscription.go
+++ b/packages/taiko-client/pkg/rpc/subscription.go
@@ -41,9 +41,7 @@ func SubscribeBatchesVerifiedPacaya(
 			return nil, err
 		}
 
-		defer sub.Unsubscribe()
-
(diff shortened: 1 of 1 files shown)


davidtaikocha:
Don't think these changes are necessary...

PR #20471, opened 2025-10-17
rpc: fix data race in GetProtocolStateVariablesPacaya

vtjl10:
### Summary
Eliminates a data race when fetching protocol state variables by avoiding shared writes from concurrent goroutines.

### Problem
Both goroutines wrote to shared `err` and mutated `states` concurrently, causing potential error overwrites and inconsistent results.

### Fix
Collect results in local variables inside goroutines and assign to `states` only after `g.Wait()` succeeds.


davidtaikocha:
Don't think there is a data-race, it's assigned to two different fields.

PR #20472, opened 2025-10-18
feat(protocol): allow reading arbitrary Inbox & Anchor storage slot

dantaik:
Add a convinient method to read any slots in the Inbox contract.


dantaik:
@claude please review

ggonzalez94 (commented):
PR looks good. But we were already near the inbox size limits(had to use the optimized profile and split the view functions to the codec implementation). I wonder if it is worth adding this when this can already be queried off-chain(with less convenience, but still totally possible)

dantaik:
Lets decide later then.

PR #20496, opened 2025-10-21
fix(taiko-client): check the shasta last seen proposal after reorg

### LINKED ISSUE #20465: Monitor reorg in shasta indexer

YoGhurt111:


davidtaikocha (approved):

packages/taiko-client/bindings/metadata/interface.go
```diff
@@ -57,4 +58,5 @@ type TaikoProposalMetaDataShasta interface {
 	GetBlobTimestamp(int) uint64
 	GetRawBlockHeight() *big.Int
 	GetRawBlockHash() common.Hash
+	GetLog() types.Log
```
davidtaikocha:
```suggestion
	GetLog() *types.Log
```
Let's only return a reference here?

packages/taiko-client/pkg/state_indexer/indexer.go
```diff
@@ -33,6 +36,7 @@ type ProposalPayload struct {
 	Derivation     *shastaBindings.IInboxDerivation
 	RawBlockHash   common.Hash
 	RawBlockHeight *big.Int
+	Log            types.Log
```
davidtaikocha: ditto.

YoGhurt111: Fixed

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -1051,77 +1052,123 @@ func (s *PreconfBlockAPIServer) monitorLatestProposalOnChain(ctx context.Context
...
+
+	// Find the last valid proposal by searching backwards
+	maxIterations := latestSeenProposalID.Int64()
+	for i := int64(1); i <= maxIterations; i++ {
+		currentProposalID := new(big.Int).Sub(latestSeenProposalID, big.NewInt(i))
+
+		onChainHash, err := s.rpc.GetShastaProposalHash(&bind.CallOpts{Context: ctx}, currentProposalID)
 		if err != nil {
-			log.Error("Failed to filter batch proposed event", "err", err)
+			log.Error("Failed to get shasta proposal on chain", "proposalId", currentProposalID, "err", err)
```
MatusKysel, line 1138: should we return here on error? just asking as we are iterating there, shouldn't with re-try it with lower `currentProposalID` ?

YoGhurt111: Here is the RPC error, we exit the loop directly and rely on the next retry.
We re-try it with lower `currentProposalID` when we find that `onChainHash` does not match `recordedProposalHash`.

MatusKysel (approved):

PR #20497, opened 2025-10-21
fix(eventindexer): correct error messages to match actual V2 function names

ALEX-WWU:
Fix inconsistent error wrapping where V2 function calls were logged without the "V2" suffix

Consistent with existing code in [save_transition_proved_event.go:114](https://github.com/taikoxyz/taiko-mono/blob/eb7adbe20665b84b0a00f3927b8151370f75e86f/packages/eventindexer/indexer/save_transition_proved_event.go#L114) where V2 functions include "V2" in error messages.


davidtaikocha:
It's OK to keep them atm, thanks for the PR.

PR #20519, opened 2025-10-23
fix: remove unused ERC20 interface

Lil-Duckling-22:
Removed the unused ERC20 interface that was defined but never implemented or used in the codebase. The balance monitor uses dynamic contract calls via bind.NewBoundContract instead of static interface implementations.


davidtaikocha:

Issue #20486, opened 2025-10-20
[Post whitelist preconfs] Update driver for new PreconfComittment


cyberhorsey:
1) we no longer gossip blocks over P2P.

2) we instead gossip the PreconfirmationCommitment to the Envelope for the P2P network, to be published for every L2 block.

 a) check the slasherAddress *is* the correct slasher address as part of validation

 3) the driver can reconstruct the block and advance the head with the fields in this new PreconfirmationCommitment struct.

 4) anchor transaction must be constructed by the driver now.

 5) driver makes block, advances head.

PR #20529, opened 2025-10-24
feat(preconf): gossip sidecar-signed SignedCommitment; disable block gossip

MatusKysel:
ref https://github.com/taikoxyz/taiko-mono/issues/20486


packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -48,6 +48,8 @@ type BuildPreconfBlockRequestBody struct {
 	ExecutableData    *ExecutableData `json:"executableData"`
 	EndOfSequencing   *bool           `json:"endOfSequencing"`
 	IsForcedInclusion *bool           `json:"isForcedInclusion"`
+	// Optional preconfirmation SignedCommitment from the sidecar, to be gossiped as-is.
+	Preconfirmation *p2p.SignedCommitment `json:"preconfirmation"`
```
cyberhorsey, line 59: we should use an entirely new struct for this, and a new API, PreconfBlockV2 or whatever. Because all of the existing PreconfBlock data above, the `ExecutableData`, will be gone after the transition.

So, we should have an API that accepts the original struct, and a new API that accepts the PreconfirmationCommitment. Sidecars post-transition will call the new API, and use the old API until then.

After, we can delete the old API and struct.

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -275,31 +277,29 @@ func (s *PreconfBlockAPIServer) BuildPreconfBlock(c echo.Context) error {
...
-					c,
-					http.StatusInternalServerError,
-					fmt.Errorf("failed to update L1 origin signature: %w", err),
+			var sigBytes *[65]byte
+			// For pre-Shasta, we still persist the signature to L1Origin.
+			if header.Number.Uint64() < s.rpc.ShastaClients.ForkHeight.Uint64() && s.p2pSigner != nil {
+				b, err := s.p2pSigner.Sign(
+					ctx,
+					p2p.SigningDomainBlocksV1,
+					s.rpc.L2.ChainID,
```
cyberhorsey: yeah lets keep the original API unchanged here and just add a new post-shasta post-whitelist API endpoint

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -318,12 +318,15 @@ func (s *PreconfBlockAPIServer) BuildPreconfBlock(c echo.Context) error {
...
 				Signature:         sigBytes,
 			}
-
-			// Cache locally so this node can perform orphan handling without relying on receiving our own gossip.
 			s.tryPutEnvelopeIntoCache(env, s.p2pNode.Host().ID())
 
-			if err := s.p2pNode.GossipOut().PublishL2Payload(ctx, env, s.p2pSigner); err != nil {
-				log.Warn("Failed to propagate the preconfirmation block to the P2P network", "error", err)
+			// Publish sidecar-provided SignedCommitment if present; do not sign here.
+			if reqBody.Preconfirmation != nil {
```
cyberhorsey: same as above, new API endpoint that just publishes the preconfirmationCommitment, and the old API uses PublishL2Payload.

cyberhorsey (changes requested):

packages/taiko-client/cmd/flags/driver.go
```diff
@@ -72,6 +72,15 @@ var (
 		Category: driverCategory,
 		EnvVars:  []string{"TAIKO_WRAPPER"},
 	}
+	// preconfirmation transition timestamp (unix seconds). Before this timestamp, old APIs + block gossip are used.
+	// At or after this timestamp, new commitment-based API and logic are enabled and old APIs are disabled.
+	PreconfTransitionTimestamp = &cli.Uint64Flag{
```
cyberhorsey, line 84: unsure of whether itll be a block number or a timestamp right now but we can leave this and confer with the partners

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -357,6 +378,127 @@ func (s *PreconfBlockAPIServer) BuildPreconfBlock(c echo.Context) error {
...
+	defer func() {
+		elapsedMs := time.Since(start).Milliseconds()
+		metrics.DriverPreconfBuildPreconfBlockDuration.Observe(float64(elapsedMs) / 1_000)
+		log.Debug("BuildPreconfBlockV2 completed", "elapsed", fmt.Sprintf("%dms", elapsedMs))
+	}()
+
+	ctx := context.Background()
+
+	// Check if preconfirmation is enabled.
+	if s.rpc.PacayaClients.TaikoWrapper != nil {
```
cyberhorsey: i dont think we will have a preconfRouter anymore post-Shasta and thus post-Whitelist. it is removed from the repository. i think preconfirmations dont have a way to be "disabled" and only fallback proposing will exist anymore. @ggonzalez94 correct me if im wrong?

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -357,6 +378,127 @@ func (s *PreconfBlockAPIServer) BuildPreconfBlock(c echo.Context) error {
...
+	} else {
+		// No P2P configured; still store preimage for reconstruction
+		s.txlistMu.Lock()
+		s.txlist[pc.RawTxListHash] = reqBody.Transactions
+		s.txlistMu.Unlock()
+	}
+
+	// Process the commitment locally to reconstruct and insert the block.
+	// Note: OnUnsafePreconfirmationCommitment handles concurrency (locks internally).
+	var zeroPeer peer.ID
```
cyberhorsey, line 455: what does having a zeroPeer here do?

packages/taiko-client/driver/preconf_blocks/api.go
```diff
@@ -357,6 +378,127 @@ func (s *PreconfBlockAPIServer) BuildPreconfBlock(c echo.Context) error {
...
+	header, err := s.rpc.L2.HeaderByNumber(ctx, pc.BlockNumber)
+	if err != nil || header == nil {
+		return s.returnError(c, http.StatusInternalServerError, fmt.Errorf("failed to fetch inserted header: %w", err))
+	}
+
+	// Update unsafe payload tracking
+	s.updateHighestUnsafeL2Payload(header.Number.Uint64())
+
+	// If EOP marker, update cache
+	if pc.EOP && s.rpc.L1Beacon != nil {
```
cyberhorsey, line 470: i believe if we receive an EOP marker, we need to "advance the lookahead" here as well (ongoing discussions in partner channel), rather than wait for the pre-determined lookahead advancement.

cyberhorsey (changes requested):

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -177,6 +250,236 @@ func (s *PreconfBlockAPIServer) SetP2PSigner(p2pSigner p2p.Signer) {
...
+		return nil
+	}
+
+	if sc == nil {
+		log.Warn("Empty preconfirmation commitment received", "peer", from)
+		return nil
+	}
+
+	// Validate slasher address matches expected committer for current window
+	expected := s.CurrentPreconferCommitter()
```
cyberhorsey: this is incorrect, the slasher address is the address *of the actual slasher contract*, we only need to verify the SlasherAddress on the commitment is equal to the actual slasher address (which we can pass in to the driver via a flag)

cyberhorsey (changes requested):

cyberhorsey:
we will need to update the driver's lookahead building to support the whitelisted version and the post-whitelist version which doesnt have a `getCurrent` and `getNext` functions on a whitelist contract, and instead the lookahead can be reconstructed.

packages/taiko-client/cmd/flags/driver.go
MatusKysel: I though it will be time-based since we moved Shasta to time based one

packages/taiko-client/driver/preconf_blocks/api.go
MatusKysel: https://github.com/taikoxyz/taiko-mono/blob/2f8e9b44472a651ed3daed6b3180f00a88b70505/packages/taiko-client/driver/preconf_blocks/server.go#L925 it's following old logic, but before you used `""` I find zeroPeer more clear, but basically this function handles both flows p2p and local and we do same for old blocks

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -177,6 +263,252 @@ func (s *PreconfBlockAPIServer) SetP2PSigner(p2pSigner p2p.Signer) {
...
+			"peer", from,
+			"expected", expected,
+			"got", sc.Commitment.SlasherAddress,
+		)
+		return nil
+	}
+
+	pc := sc.Commitment.Preconf
+	// Derive committer from signature and cache by epoch
+	committer, err := s.recoverCommitter(sc)
```
MatusKysel, line 321: @cyberhorsey you mean reconstruct like that ?

packages/taiko-client/driver/preconf_blocks/api.go
MatusKysel: fixed

packages/taiko-client/driver/preconf_blocks/api.go
MatusKysel: please check now

packages/taiko-client/driver/preconf_blocks/server.go
MatusKysel: fixed

PR #20532, opened 2025-10-24
feat(overseer-rs): overseer and urc indexer

### LINKED ISSUE #20487: [Post whitelist preconfs] Overseer
`overseer` - sole responsibility is blacklisting. It should monitor L2 block times, and then call `blacklist` if a preconfirmer is not living up to the criteria of being a preconfirmer: publishing VALID blocks every X seconds (block time). This must be run by the DAO or Taiko, the overseer will be whitelisted, but it is an EOA that can submit transactions.

- if there are valid L2 mempool transactions (in the pending queue, not queued), and the preconfimer is not including them, they can also be blacklisted.  Should have several flags for parameters to decide how to blacklist. Blacklisting is immediate and a new preconfirmer will be selected.

cyberhorsey:
closes #20487


davidtaikocha:
@claude plz review.

davidtaikocha (approved):

packages/overseer-rs/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# syntax=docker/dockerfile:1.6
+
+## Build stage: compile the overseer binary
+FROM rust:1.77-slim AS builder
```
mskrzypkows: Why not newest 1.90?

packages/overseer-rs/src/criteria/block_timeliness.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl BlacklistCriterion for BlockTimelinessCriterion {
+    /// Flags a violation when no valid block is observed within the allowed delay.
+    async fn evaluate(&self, ctx: &EvaluationContext<'_>) -> Result<Option<Violation>> {
+        let last_block_time = ctx.observation.latest_block.timestamp;
+        let now = SystemTime::now();
+        let Ok(delay) = now.duration_since(last_block_time) else {
```
mskrzypkows, line 25: I think that error should be propagated here:
`let delay = now.duration_since(last_block_time)?;`
other way, there is no reason to return `Result` from the function, because it always will be `Ok`.

packages/overseer-rs/src/ethereum/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+            .ok_or_else(|| anyhow!("missing result field in RPC response"))
+    }
+}
+
+#[async_trait]
+impl EthereumClient for RpcEthereumClient {
+    /// Fetches the latest block metadata using `eth_getBlockByNumber`.
+    async fn latest_block(&self) -> Result<BlockSnapshot> {
+        let block: RpcBlock = self
+            .rpc_call("eth_getBlockByNumber", serde_json::json!(["latest", false]))
```
mskrzypkows: I wonder why don't you use alloy for accessing Ethereum RPC. E.g. block can be read with `get_block_by_number` from alloy provider: https://github.com/NethermindEth/Catalyst/blob/master/common/src/shared/execution_layer.rs#L98

packages/overseer-rs/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+reqwest = { version = "0.11", default-features = false, features = ["json", "rustls-tls"] }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+tokio = { version = "1.37", features = ["macros", "rt-multi-thread", "signal", "net"] }
+tracing = "0.1"
+tracing-subscriber = { version = "0.3", features = ["fmt", "env-filter"] }
+thiserror = "1.0"
+prometheus = "0.13"
+hyper = { version = "0.14", features = ["server", "http1"] }
+sqlx = { version = "0.8", default-features = false, features = ["runtime-tokio", "sqlite", "macros"] }
```
MatusKysel: is this used ?

packages/overseer-rs/src/monitor.rs
```diff
@@ -0,0 +1,269 @@
...
+                Err(err)
+            }
+        }
+    }
+
+    async fn identify_stalled_transactions(
+        &self,
+        pending: &[PendingTransaction],
+    ) -> Vec<StalledTransaction> {
+        let mut tracker = self.pending_tracker.lock().await;
```
MatusKysel, line 250: method tracks transaction hashes across cycles, but the tracker clears entries for transactions that aren’t seen in the current snapshot immediately. If a transaction is briefly included in a block and then reorged back into the pool, its previous tracking history is lost and its “age” resets. Is this expected ?

cyberhorsey: yeah this is by design as it DID get included at one point so it means the preconfirmer is not censoring

MatusKysel (approved):

release-please-config.json
```diff
@@ -76,6 +76,18 @@
     "packages/docs-site": {},
     "packages/eventindexer": {},
     "packages/fork-diff": {},
+    "packages/overseer-rs": {
+      "package-name": "overseer-rs",
+      "release-type": "simple",
+      "component": "overseer-rs",
```
RogerLamTd, line 82:
```suggestion
      "component": "overseer-rs",
      "release-as": "0.1.0",
```

release-please-config.json
```diff
@@ -100,6 +112,18 @@
           "jsonpath": "$.workspace.package.version"
         }
       ]
+    },
+     "packages/urcindexer-rs": {
+      "package-name": "urcindexer-rs",
+      "release-type": "simple",
+      "component": "urcindexer-rs",
```
RogerLamTd, line 120:
```suggestion
      "component": "urcindexer-rs",
      "release-as": "0.1.0",
```

packages/overseer-rs/Cargo.toml
cyberhorsey: ah no, its a hangover from before i added MySQL support to the urcindexing crate directly in netherminds catalyst repo.

packages/overseer-rs/src/ethereum/mod.rs
cyberhorsey: this part written by AI, we can use alloy provider!

packages/overseer-rs/src/criteria/block_timeliness.rs
cyberhorsey: agreed, the reason it returns `Result` is because it implements a trait though, which has a return type of Result, so it has to return Result like the other criteria.

packages/overseer-rs/Dockerfile
cyberhorsey: bumped it

RogerLamTd (changes requested):

RogerLamTd (approved):

PR #20538, opened 2025-10-25
chore(protocol): add zero address check in SigVerifyLib constructor

IrvinGa38:
The SigVerifyLib constructor accepts zero address for es256Verifier, which breaks all signature verifications. Added validation to fail early at deployment instead of silently failing at runtime.


packages/protocol/contracts/layer1/automata-attestation/utils/SigVerifyLib.sol
```diff
@@ -13,6 +13,7 @@ contract SigVerifyLib is ISigVerifyLib {
     address private immutable __es256Verifier;
 
     constructor(address es256Verifier) {
+        require(es256Verifier != address(0), "Invalid es256 verifier");
```
dantaik: We need to use a custom error here, not a string please

IrvinGa38: Hi @dantaik, done, pls check, thanks!

dantaik (approved):

smtmfft (approved):

PR #20540, opened 2025-10-26
perf(protocol): optimize hashing functions using EfficientHashLib

dantaik:
Use EfficientHashLib for hashing.


dantaik:
@claude please review

davidtaikocha (dismissed):

packages/protocol/contracts/layer1/verifiers/LibPublicInput.sol
```diff
@@ -24,11 +24,32 @@ library LibPublicInput {
...
+            // Store string at offset 0xa0
+            // "VERIFY_PROOF" = 12 bytes
+            mstore(add(ptr, 0xa0), 0x0c) // length = 12
+            mstore(add(ptr, 0xc0), "VERIFY_PROOF") // string data
+
+            // Total size: 160 + 64 = 224 = 0xe0
+            hash := keccak256(ptr, 0xe0)
+
+            // Update free memory pointer to prevent memory corruption
+            mstore(0x40, add(ptr, 0xe0))
```
AnshuJalan: Shouldn't it be safe to go back to the original pointer value and reclaim the full memory?

dantaik:
```suggestion
            mstore(0x40,ptr)
```

dantaik:
@AnshuJalan please take another look

packages/protocol/contracts/layer1/verifiers/LibPublicInput.sol
dantaik: please take another look

AnshuJalan (approved):

YoGhurt111 (approved):

packages/protocol/contracts/layer1/verifiers/LibPublicInput.sol
```diff
@@ -24,11 +24,33 @@ library LibPublicInput {
...
-        return keccak256(
-            abi.encode(
-                "VERIFY_PROOF", _chainId, _verifierContract, _aggregatedProvingHash, _newInstance
-            )
-        );
+        // Original: return keccak256(abi.encode(
+        //   "VERIFY_PROOF", _chainId, _verifierContract, _aggregatedProvingHash, _newInstance));
+        // Optimized using inline assembly: saves 281 gas (49.5% reduction: 568 gas -> 287 gas)
+        bytes32 hash;
+        assembly {
```
YoGhurt111: @smtmfft Maybe you need to make corresponding changes to this and the following changes on raiko side.

packages/protocol/contracts/layer1/verifiers/LibPublicInput.sol
```diff
@@ -24,11 +24,33 @@ library LibPublicInput {
         returns (bytes32)
     {
         require(_aggregatedProvingHash != bytes32(0), InvalidAggregatedProvingHash());
-        return keccak256(
-            abi.encode(
-                "VERIFY_PROOF", _chainId, _verifierContract, _aggregatedProvingHash, _newInstance
-            )
-        );
+        // Original: return keccak256(abi.encode(
```
ggonzalez94: a few questions before we decide to go with this approach(which I'm not opposed to), just wanna make sure we think about it:
1. zk proof verification costs between 200k-300k gas, is the 281 gas savings(~0.1%) worth the reduced code readability?
2. **If** we decide it is still worth pursuing, should we consider using solady's EfficientHashLib library instead as recommended for production in the [foundry book](https://getfoundry.sh/forge/linting/#gas-optimizations)? Less room for error, and less code to audit

ggonzalez94 (commented):
LGTM, but left an open question in terms of wether this is worth it given zk proofs already cost 200k-300k to verify + a recommendation to use solady's library instead. Let me know what you think

packages/protocol/contracts/layer1/verifiers/LibPublicInput.sol
dantaik: I don't think you should. The new code shall have the same results, only with less gas.

packages/protocol/contracts/layer1/verifiers/LibPublicInput.sol
dantaik: I think we should try to optimize gas, these efforts will add up. 

EfficientHashLib is a good idea, I'll give it a try.

dantaik: Done, please take another look.

dantaik:
@​smtmfft the semantics of hashPublicInputs have changed:
	•	Previously: we used a string `"VERIFY_PROOF"`
	•	Now: we use a bytes32 value `bytes32("VERIFY_PROOF")`

You’ll need to update the prover accordingly.

packages/protocol/contracts/layer1/verifiers/LibPublicInput.sol
```diff
@@ -24,10 +26,12 @@ library LibPublicInput {
...
-            abi.encode(
-                "VERIFY_PROOF", _chainId, _verifierContract, _aggregatedProvingHash, _newInstance
-            )
+        return EfficientHashLib.hash(
+            bytes32("VERIFY_PROOF"),
+            bytes32(uint256(_chainId)),
+            bytes32(uint256(uint160(_verifierContract))),
+            _aggregatedProvingHash,
+            bytes32(uint256(uint160(_newInstance)))
         );
```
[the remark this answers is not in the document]
dantaik: @smtmfft

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
```diff
@@ -136,16 +136,13 @@ contract SgxVerifier is IProofVerifier, Ownable2Step {
...
         address instance = address(bytes20(_proof[4:24]));
 
-        // Collect public inputs
-        bytes32[] memory publicInputs = new bytes32[](2);
-        // First public input is the current instance public key
-        publicInputs[0] = bytes32(uint256(uint160(instance)));
-
-        publicInputs[1] = LibPublicInput.hashPublicInputs(
-            _aggregatedProvingHash, address(this), address(0), taikoChainId
+        bytes32 signatureHash = EfficientHashLib.hash(
```
dantaik, line 139: @smtmfft please also notice the change here.

packages/protocol/contracts/layer1/verifiers/LibPublicInput.sol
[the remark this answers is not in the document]
smtmfft: It should be ok, as the Shasta implementation has no backward compatibility requirements.

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
smtmfft: noted. synced with client, we will align current impl this week, to let client complete the proving logic first, and can update to new hash impl next week

smtmfft (approved):

YoGhurt111 (approved):

PR #20543, opened 2025-10-27
fix(protocol): fork router logic for SignalService and fix anchor storage layout

ggonzalez94:
## Summary
Fixes two issues with the for transition:

1. The Anchor layout is not compatible with the previous fork(In pacaya since the contract was upgradable and inherited from `EssentialContract`it used 251 slots from that + other 104 from `TaikoAnchorDeprecated` and only after that it had its own storage variables). This PR fixes that gap
2. The SignalService we created for Shasta, while compatible in terms of storage layout and abi(same function names and parameters), the checkpoints are retrieved from a different location, which means if we were to upgrade the contract before the fork height all message verifications would fail, unless the upgrade happens at the exact fork timestamp.

## Description

This PR introduces a fix based on the fork router pattern, which allows the codebase to remain clean while still being able to fork to the right implementation contract at runtime.

**Additional Notes:**
- We don't reuse any slot in the SignalService to avoid issues and so that the storage layout is cleaner
- Removed the `MainnetSignalService` since we don't actually have reentrancy guards in the SS, and now that we removed the inheritance from EssentialContract it was not compiling

## TODOs
**If** we were to use this approach, there's still a couple things to finalize this PR:

- [x] Fix all the tests and scripts that expect an upgradable `SignalService`to use the immutable version OR the fork router pattern, depending on the case.
- [x] Inherit from the new `EssentialContractStorage` to make storage layout cleaner

---

> [!NOTE]
> Introduces a fork router for SignalService and reworks Anchor storage to be Pacaya-compatible, updating deployments, layouts, and tests accordingly.
> 
> - **Core Contracts**:
>   - **Anchor storage**: Inherit `PacayaAnchorStorage` (via new `EssentialContractStorage`) to match Pacaya layout; update `Anchor_Layout` accordingly.
>   - **SignalService**: Adjust storage layout (`_slotsUsedByPacaya` size, `_receivedSignals` slot), replace modifiers with `require`, add `init`, and align for fork-router compatibility.
>   - Remove `layer1/mainnet/MainnetSignalService.sol` and its layout.
> - **Fork Routing**:
>   - Add `ForkRouter` (with layout) and `SignalServiceForkRouter` (timestamp-based routing between old/new implementations).
> - **Deployment/Tooling**:
>   - Update `DeployProtocolOnL1.s.sol` to deploy/upgrade `SignalService` around inbox deployment; register if missing; transfer ownership.
>   - Include `ForkRouter` in `gen-layouts.sh`; drop `MainnetSignalService` from generation.
> - **Tests**:
>   - Refactor tests to deploy `SignalService` with immutables and use mock variant without proof; update genesis and bridge/vault suites; adjust inbox and anchor tests to upgraded patterns.
> - **Docs**:
>   - Tidy `Derivation.md` constants table formatting.
> - **Reports**:
>   - Refresh gas report snapshots.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c05a160c4484e785c0433cfb22981065bfe8a578. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/shared/signal/SignalService.sol
```diff
@@ -34,14 +33,15 @@ contract SignalService is EssentialContract, ISignalService {
...
     // ---------------------------------------------------------------
     // Pre shasta storage variables
     // ---------------------------------------------------------------
 
-    /// @dev Deprecated slots used by the old SignalService
+    /// @dev Deprecated slots used by the old SignalService. This includes the slots from `EssentialContract` and:
     // - `topBlockId`
     // - `authorized`
-    uint256[2] private _slotsUsedByPacaya;
+    uint256[253] private _slotsUsedByPacaya;
```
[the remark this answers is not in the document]
ggonzalez94: The `EssentialContract` used 251 slots. I already fixed the PR description

packages/protocol/contracts/shared/fork-router/ForkRouter.sol
```diff
@@ -13,6 +13,8 @@ import { UUPSUpgradeable } from "@openzeppelin/contracts/proxy/utils/UUPSUpgrade
 /// PROXY -> FORK_ROUTER--|
 ///                         +--> oldFork
 ///
+/// @dev WARNING: This contract uses 151 slots [0..150]. Routed contracts should reserve 151 slots to avoid collisions.
```
dantaik: I think this comment doesn't belong here as ForkRouter itself doesn't even require oldFork and newFork to be storage compatible.

packages/protocol/contracts/shared/common/EssentialContractStorage.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+/// @title EssentialContractStorage
+/// @notice Mirrors the storage footprint of `EssentialContract` so that contracts
+///         replacing it can keep the exact slot layout without inheriting its logic.
+contract EssentialContractStorage {
```
dantaik: Since this contract is only used by Anchor.sol, so maybe make sense to move it to Anchor.sol?

dantaik:
```suggestion
abstract contract EssentialContractStorage {
```

dantaik:
To handle the transition between the old and new signal-checking logic in SignalService, we can introduce a temporary mechanism as follows:

- **Introduce a new immutable variable**: Add an immutable variable checkPreviousSignalReceivedDeadline.
    - If the current timestamp is greater than this value, we only check for a signal using the new method.
    - If the timestamp is less than or equal to this value, we first check the old method, and if not found, then fall back to the new method.
   
- **Set a long enough deadline**: The deadline should be set far enough in the future to safely cover the fork timestamp, which may be unknown at the time of deployment.

- **Post-fork cleanup**: In a follow-up upgrade after the fork, remove both the checkPreviousSignalReceivedDeadline variable and all logic related to the old signal-checking path.

dantaik:
Regarding Anchor.sol:

I did a layout comparison between the current branch and the [Pacaya 2.3.0 release](https://github.com/taikoxyz/taiko-mono/tree/f599a91e2d6936351a0c92ffe7e04ef28ce8b48c).
It appears that the storage layouts are not compatible: [diff link](https://www.diffchecker.com/tqBOjg9B/).

Please also make sure that all deprecated storage slots are explicitly declared (e.g., using reserved gaps) so we don’t accidentally reuse “dirty” slots for new state variables in future upgrades.

dantaik (changes requested):

packages/protocol/contracts/shared/fork-router/ForkRouter.sol
ggonzalez94: But even in that case you don't want the routed contracts(oldFork and newFork) to have storage collision with the ForkRouter itself. Otherwise they can potentially overwrite the router's storage variables and vice versa

packages/protocol/contracts/shared/common/EssentialContractStorage.sol
ggonzalez94: my goal was to make something more reusable for any contract that we want to migrate from inheriting from EssentialContract to not doing it. Anchor is the first example of this, but if we go with this patter with the SignalService that's another, and you can see this happening more. Feels like a good reusable component?

ggonzalez94:
> Regarding Anchor.sol:
> 
> I did a layout comparison between the current branch and the [Pacaya 2.3.0 release](https://github.com/taikoxyz/taiko-mono/tree/f599a91e2d6936351a0c92ffe7e04ef28ce8b48c). It appears that the storage layouts are not compatible: [diff link](https://www.diffchecker.com/tqBOjg9B/).
> 
> Please also make sure that all deprecated storage slots are explicitly declared (e.g., using reserved gaps) so we don’t accidentally reuse “dirty” slots for new state variables in future upgrades.

You are correct, thanks for catching this. The problem was with the inheritance. I put the old storage variables in the base contract, and now everything should be fine(plus it is easier to follow). Here's the new [diff](https://www.diffchecker.com/SGnJLPRE/). The new storage variables only start from slot 255, which was previously the empty `gap` of the pacaya anchor

packages/protocol/contracts/shared/signal/SignalServiceForkRouter.sol
```diff
@@ -0,0 +1,23 @@
...
+    uint64 public immutable shastaForkTimestamp;
+
+    constructor(address _oldFork, address _newFork, uint64 _shastaForkTimestamp)
+        ForkRouter(_oldFork, _newFork)
+    {
+        shastaForkTimestamp = _shastaForkTimestamp;
+    }
+
+    function shouldRouteToOldFork(bytes4) public view override returns (bool) {
+        return block.timestamp < shastaForkTimestamp;
```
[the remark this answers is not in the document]
ggonzalez94: this is a known tradeoff we are making. Users will just need to regenerate their proofs

packages/protocol/contracts/layer1/mainnet/MainnetSignalService.sol
```diff
@@ -1,35 +0,0 @@
-// SPDX-License-Identifier: MIT
-pragma solidity ^0.8.24;
-
-import "src/layer1/mainnet/LibFasterReentryLock.sol";
```
dantaik, line 4: I think we shall restore MainnetSignalService as it will save gas by using TSTORE

dantaik:
@ggonzalez94 I still think SignalService should inherit from EssentialContract or EssentialResolvableContract. Using ForkRouter is only a temporary measure — once the fork is complete, we’ll upgrade the implementation again from ForkRouter to the new SignalService (or MainnetSignalService).

ForkRouter adds extra overhead since calls go through proxy → forkRouter → new/old_impl, whereas without it, they go directly via proxy → impl.

If we want the new SignalService to work under the simpler proxy → impl pattern, it needs to inherit from one of the Essential* base contracts. Otherwise, we’d have to modify its code again later just to remove the router.

What do you think?

packages/protocol/script/layer1/core/DeployProtocolOnL1.s.sol
```diff
@@ -175,8 +176,30 @@ contract DeployProtocolOnL1 is DeployCapability {
...
+        }
+
+        if (!signalServiceExists) {
+            uint256 nonceBefore = vm.getNonce(msg.sender);
+            // Upcoming creations before the proxy:
+            // 1) SignalService, 2) CodecOptimized, 3) DevnetInbox impl, 4) ERC1967Proxy
+            address predictedShastaInbox = vm.computeCreateAddress(msg.sender, nonceBefore + 3);
+            signalService = address(
+                new SignalService(predictedShastaInbox, config.remoteSigSvc, config.contractOwner)
+            );
```
davidtaikocha: Shall we just do another proxy upgrade for `signalService` after we deployed the `shastaInbox`? And pass the `predictedShastaInbox` here a place holder like `address(0)`? Which will make this logic doesn't rely on `nonceBefore + 3` assumption anymore.

packages/protocol/test/genesis/generate/taikoAnchor.ts
```diff
@@ -1029,6 +1016,16 @@ function replaceImmutableValues(artifact: any, maps: Array<any>): any {
...
+
+    // If no offsets found, the immutable might be declared but not used in this artifact's bytecode
+    // This can happen with inherited immutables. Return artifact unchanged.
+    if (!offsets || !Array.isArray(offsets) || offsets.length === 0) {
+        console.log(
+            `No immutable reference offsets found for id ${id} in bytecode, skipping replacement`,
+        );
+        return artifact;
+    }
+
```
davidtaikocha: Shall we explicitly removed the code here: https://github.com/taikoxyz/taiko-mono/blob/signal-fork-router/packages/protocol/test/genesis/generate/taikoAnchor.ts#L620-L633 if new signal service is no longer a `UUPS` and `EssentialContract`, instead of skip the checks here?

ggonzalez94:
@dantaik @davidtaikocha I updated the code to address comments and make the SignalService inherit from EssentialContract(we reuse the same storage slot for owner, so that we don't have to init the contract again). Please take another look.

@davidtaikocha please check the deploy and genesis scripts to make sure I didn't do anything rare there 😅

packages/protocol/test/genesis/generate/taikoAnchor.ts
```diff
@@ -657,13 +654,6 @@ async function generateContractConfigs(
             deployedBytecode:
                 contractArtifacts.SignalService.deployedBytecode.object,
             variables: {
-                // EssentialContract
-                __reentry: 1, // _FALSE
-                __paused: 1, // _FALSE
-                // EssentialContract => UUPSUpgradeable => Initializable
-                _initialized: 1,
-                _initializing: false,
-                // EssentialContract => Ownable2StepUpgradeable
```
davidtaikocha, line 666: Since we have made `SignalService` inherit from `EssentialContract` again, maybe we should still initialize these variables here?

dantaik: +1

packages/protocol/contracts/layer2/core/Anchor_Layout.sol
```diff
@@ -7,11 +7,12 @@ pragma solidity ^0.8.24;
...
-//   _owner                         | address                                            | Slot: 0    | Offset: 0    | Bytes: 20  
-//   _pendingOwner                  | address                                            | Slot: 1    | Offset: 0    | Bytes: 20  
-//   _status                        | uint256                                            | Slot: 2    | Offset: 0    | Bytes: 32  
-//   _pacayaSlots                   | uint256[4]                                         | Slot: 3    | Offset: 0    | Bytes: 128 
-//   _proposalState                 | struct Anchor.ProposalState                        | Slot: 7    | Offset: 0    | Bytes: 64  
-//   _blockState                    | struct Anchor.BlockState                           | Slot: 9    | Offset: 0    | Bytes: 64  
-//   blockHashes                    | mapping(uint256 => bytes32)                        | Slot: 11   | Offset: 0    | Bytes: 32  
-//   __gap                          | uint256[41]                                        | Slot: 12   | Offset: 0    | Bytes: 1312
+//   __essentialContractStorage     | uint256[251]                                       | Slot: 0    | Offset: 0    | Bytes: 8032
```
dantaik: I think the current layout is still different from the Pacaya anchor: https://raw.githubusercontent.com/taikoxyz/taiko-mono/f599a91e2d6936351a0c92ffe7e04ef28ce8b48c/packages/protocol/layout/layer2-contracts.md


```
╭-----------------------------+-----------------------------+------+--------+-------+----------------------------------------------------╮
| Name                        | Type                        | Slot | Offset | Bytes |
+========================================================================================================================================+
| _initialized                | uint8                       | 0    | 0      | 1     |
| _initializing               | bool                        | 0    | 1      | 1     |
| __gap                       | uint256[50]                 | 1    | 0      | 1600  |
| _owner                      | address                     | 51   | 0      | 20    |
| __gap                       | uint256[49]                 | 52   | 0      | 1568  |
| _pendingOwner               | address                     | 101  | 0      | 20    |
| __gap                       | uint256[49]                 | 102  | 0      | 1568  |
| __gapFromOldAddressResolver | uint256[50]                 | 151  | 0      | 1600  |
| __reentry                   | uint8                       | 201  | 0      | 1     |
| __paused                    | uint8                       | 201  | 1      | 1     |
| __gap                       | uint256[49]                 | 202  | 0      | 1568  |
| _blockhashes                | mapping(uint256 => bytes32) | 251  | 0      | 32    |
| publicInputHash             | bytes32                     | 252  | 0      | 32    |
| parentGasExcess             | uint64                      | 253  | 0      | 8     |
| lastSyncedBlock             | uint64                      | 253  | 8      | 8     |
| parentTimestamp             | uint64                      | 253  | 16     | 8     |
| parentGasTarget             | uint64                      | 253  | 24     | 8     |
| l1ChainId                   | uint64                      | 254  | 0      | 8     |
| __gap                       | uint256[46]                 | 255  | 0      | 1472  |
╰-----------------------------+-----------------------------+------+--------+-------+----------------------------------------------------╯
```

packages/protocol/contracts/shared/common/EssentialContractStorage.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+/// @title EssentialContractStorage
```
dantaik: I think we should merge this contract into PacayaAnchorStorage

packages/protocol/contracts/shared/signal/SignalService_Layout.sol
```diff
@@ -18,7 +18,7 @@ pragma solidity ^0.8.24;
...
 //   __paused                       | uint8                                              | Slot: 201  | Offset: 1    | Bytes: 1   
 //   __gap                          | uint256[49]                                        | Slot: 202  | Offset: 0    | Bytes: 1568
-//   _slotsUsedByPacaya             | uint256[2]                                         | Slot: 251  | Offset: 0    | Bytes: 64  
-//   _receivedSignals               | mapping(bytes32 => bool)                           | Slot: 253  | Offset: 0    | Bytes: 32  
+//   _slotsUsedByPacaya             | uint256[3]                                         | Slot: 251  | Offset: 0    | Bytes: 96  
 //   _checkpoints                   | mapping(uint48 => struct SignalService.CheckpointRecord) | Slot: 254  | Offset: 0    | Bytes: 32  
-//   __gap                          | uint256[44]                                        | Slot: 255  | Offset: 0    | Bytes: 1408
+//   _receivedSignals               | mapping(bytes32 => bool)                           | Slot: 255  | Offset: 0    | Bytes: 32  
+//   __gap                          | uint256[44]                                        | Slot: 256  | Offset: 0    | Bytes: 1408
```
dantaik: Layout for Pacaya SignalService looks like this:

```
| _initialized                | uint8                                         | 0    | 0      | 1     |
| _initializing               | bool                                          | 0    | 1      | 1     |
| __gap                       | uint256[50]                                   | 1    | 0      | 1600  |
| _owner                      | address                                       | 51   | 0      | 20    |
| __gap                       | uint256[49]                                   | 52   | 0      | 1568  |
| _pendingOwner               | address                                       | 101  | 0      | 20    |
| __gap                       | uint256[49]                                   | 102  | 0      | 1568  |
| __gapFromOldAddressResolver | uint256[50]                                   | 151  | 0      | 1600  |
| __reentry                   | uint8                                         | 201  | 0      | 1     |
| __paused                    | uint8                                         | 201  | 1      | 1     |
| __gap                       | uint256[49]                                   | 202  | 0      | 1568  |
| topBlockId                  | mapping(uint64 => mapping(bytes32 => uint64)) | 251  | 0      | 32    |
| isAuthorized                | mapping(address => bool)                      | 252  | 0      | 32    |
| _receivedSignals            | mapping(bytes32 => bool)                      | 253  | 0      | 32    |
| __gap                       | uint256[47]                                   | 254  | 0      | 1504  |
```

The issue is with `_receivedSignals`, in the new layout, it takes slot 253, in the new one, it takes slot 255. 

We either reuse slot 253, or we remove it from _receivedSignals from the new impl and related functions.

packages/protocol/contracts/layer2/core/Anchor_Layout.sol
dantaik: Please see this PR for storage fixes: https://github.com/taikoxyz/taiko-mono/pull/20614

packages/protocol/contracts/shared/signal/SignalService_Layout.sol
dantaik: Please see this PR for storage fixes: https://github.com/taikoxyz/taiko-mono/pull/20614

ggonzalez94: good point. We'll need this if we want people to be able to verify signals that were sent before the fork

packages/protocol/contracts/shared/common/EssentialContractStorage.sol
ggonzalez94: I see it is already removed as part of your PR, and no need to have it if all contracts are going to continue to be upgradable and inherit from EssentialContract. The original idea was that this was a reusable base contract to help with storage layout when transitioning out of the EssentailContract

dantaik (approved):

davidtaikocha (approved):

dantaik (approved):

dantaik (approved):

PR #20549, opened 2025-10-27
feat(protocol): simulate beacon chain proof verification and improve eip4788 library

AnshuJalan:
This PR uses beacon chain data from Fusaka Devnet 3 (https://beacon.fusaka-devnet-3.ethpandaops.io/) to simulate the full verification of the validator's inclusion within the proposer lookahead (EIP7917).

`LibEIP4788` - required quite some modifications to make it compatible with how beacon state serialises the data. The previous implementation was not sufficient.

`LookaheadSlasher` - I commented out the calls to the library since the proof structure has changed. This will be addressed in a separate slasher PR.

With some play of assembly, the proof verification cost now stands at a very reasonable `50929` (was about `75K` before)


packages/protocol/contracts/layer1/preconf/libs/LibEIP4788.sol
```diff
@@ -3,111 +3,253 @@
...
+                    mstore(ptr, proofElement)
+                    mstore(add(ptr, 0x20), h)
+                }
+
+                // sha256 precompile
+                if iszero(staticcall(gas(), 0x02, ptr, 0x40, ptr, 0x20)) { revert(0, 0) }
+                h := mload(ptr)
+
+                // Reclaim memory
+                mstore(0x40, ptr)
```
AnshuJalan, line 233: Redundant. Memory is never expanded.

PR #20550, opened 2025-10-27
chore(taiko-client-rs): remove reth as dependency of event-indexer

merklefruit:
Currently, the event-indexer depends from the `rpc` crate which depends on `alethia-reth`. This is not great for downstream consumers of this crate because it adds a lot of crate version clashes and compilation time.

The only type used from the `rpc` crate is `SubscriptionSource`. 

With this PR, I removed the `rpc` dependency by parsing the connection string directly in the event-indexer crate instead.
This is just one option; a probably more robust option would be to move `SubscriptionSource` to a different crate that doesn't depend on `alethia-reth`.

Please let me know if the current approach is acceptable or if you prefer to have a different crate, such as `primitives` or similar, where to put `SubscriptionSource`. 

cc @davidtaikocha

### DIFF
--- a/packages/taiko-client-rs/Cargo.lock
+++ b/packages/taiko-client-rs/Cargo.lock
@@ -3362,7 +3362,6 @@ dependencies = [
  "dashmap 6.1.0",
  "event-scanner",
  "metrics 0.24.2",
- "rpc",
  "thiserror 2.0.16",
  "tokio",
  "tokio-retry",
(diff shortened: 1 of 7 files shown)


davidtaikocha:
> Please let me know if the current approach is acceptable or if you prefer to have a different crate, such as primitives or similar, where to put SubscriptionSource.

Acceptable to me for now, the CI tests may is failing due to some recent protocol changes, will merge this PR after fixing this.

davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/pull/20595

PR #20556, opened 2025-10-27
docs: explain bond distribution formula in contestable validity rollup

hawkadrian:
I've added an explanation for the /4 coefficient used in bond distribution calculations. 

It clarifies that 75% of losing party's bond is burned while 25% is distributed as rewards to prevent collusion and ensure economic security.


RogerLamTd:
We've already transitioned away from contestable validity, this article can be considered deprecated/archived.

PR #20557, opened 2025-10-28
refactor(protocol): add storage layout comments to contracts

dantaik:
## Summary

Updated the gen-layouts.sh script to automatically append each contract’s storage layout as inline comments at the end of the corresponding source file (for all whitelisted contracts). This removes the need for separate layout text files.

Embedding layout data directly in the verified source code ensures that when contracts are deployed and verified on Etherscan, their storage layout is preserved and easily comparable across future upgrades.


.github/workflows/protocol.yml
```diff
@@ -92,15 +92,15 @@ jobs:
         run: pnpm snapshot:l1 && pnpm layout:l1
 
       - name: Generate coverage report
-        if: ${{ secrets.CODECOV_TOKEN != '' }}
+        if: github.event.pull_request.head.repo.fork == false
```
dantaik:
```suggestion
        if: ${{ secrets.CODECOV_TOKEN != '' }}
```

.github/workflows/protocol.yml
```diff
@@ -92,15 +92,15 @@ jobs:
...
         working-directory: ./packages/protocol
         run: |
           forge coverage --report lcov
           mkdir -p coverage
           mv lcov.info coverage/lcov.info
 
       - name: Upload coverage to Codecov
-        if: ${{ secrets.CODECOV_TOKEN != '' }}
+        if: github.event.pull_request.head.repo.fork == false
```
dantaik:
```suggestion
        if: ${{ secrets.CODECOV_TOKEN != '' }}
```

PR #20559, opened 2025-10-28
ci(protocol): add storage layout comments to contracts

dantaik:
This PR enhances the smart contract upgrade verification process by generating separate `*_Layout.sol` files containing storage layout information for each upgradeable contract.

## Background

Before upgrading a proxy to use a new implementation, we need to verify that:
1. The intended source code changes are correct
2. The new implementation is initialized with correct constructor parameters  
3. The storage layouts are compatible

To help Security Council members review upgrades, a verification tool was created at codediff.taiko.xyz. This tool enables reviewers to compare old and new contract implementations without running compilation steps locally. While the tool handles verification points 1 and 2 effectively, point 3 (storage layout compatibility) requires the layout information to be embedded in the source code for comparison.

## Solution Evolution

Initially, this PR embedded layout comments directly in contract files. However, after reconsideration, the approach was changed to move storage layouts to separate `*_Layout.sol` files. This preserves contract source code integrity while maintaining comparability for upgrade verification.

## Implementation

- Updated `gen-layouts.sh` script (simplified from 215 to 137 lines) automatically generates layout files from Forge's output
- Git's skip-worktree configuration allows developers to ignore these files locally while CI tracks them for verification
- Included `setup-local-ignores.sh` helper script for easy local configuration  
- All 29 contracts properly import their corresponding `*_Layout.sol` files with `// DO NOT DELETE` comments
- Root-level scripts also updated to handle pattern-based `*_Layout.sol` file management

## Benefits

By moving storage layouts to separate imported files rather than embedding them as inline comments, we maintain clean contract sources while ensuring Security Council members can verify storage compatibility during upgrade reviews directly through the codediff tool, without external dependencies.

## Example

You can take a look at [this page on codediff.taiko.xyz](https://codediff.taiko.xyz/?addr=0x10dea67478c5f8c5e2d90e5e9b26dbe60c54d800&newimpl=0x7e53c8C4045081c274624984Ad12E7BD96927e5B&chainid=1&filter=changed) which shows the diff between the new and the old layout files based on data fetched from  Etherescan.


davidtaikocha (dismissed):

dantaik:
I understand there might be some concerns about this change, since it’s not something typically done by other projects. Let me explain why I think it could be a good idea.  

Before upgrading our a proxy to use a new implementation, we need to verify that:  

1. The intended source code changes are correct.  
2. The new implementation is initialized with correct constructor parameters.  
3. The storage layouts are compatible. 

It would be ideal to have a tool that makes these verifications easy — and allows Security Council members to review the diffs without running any git or compilation steps. 

With the help of AI, I built this tool: [https://codediff.taiko.xyz](https://codediff.taiko.xyz). It helps quickly identify differences between old and new implementations. For example, the following upgrades will be proposed for SC review:  

- https://codediff.taiko.xyz/?addr=0x73Ee496dA20e5C65340c040B0D8c3C891C1f74AE&newimpl=0xDF6327caafC5FeB8910777Ac811e0B1d27dCdf36&chainid=1&filter=changed  
- https://codediff.taiko.xyz/?addr=0xD5AA0e20e8A6e9b04F080Cf8797410fafAa9688a&newimpl=0xafCEDDe020dB8D431Fa86dF6B14C20f327382709&chainid=1&filter=changed  

As you can see, the tool already provides clear verification for points (1) and (2), but not for (3) — storage layout compatibility.  

This PR introduces a change that embeds storage layout information directly as comments in the contract source code. This allows old and new implementations to be compared for layout compatibility without relying on external tools.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -17,6 +17,9 @@ import { LibBonds } from "src/shared/libs/LibBonds.sol";
 import { LibMath } from "src/shared/libs/LibMath.sol";
 import { ICheckpointStore } from "src/shared/signal/ICheckpointStore.sol";
 
+import "./Inbox_Layout.sol"; // DO NOT DELETE
+import "./InboxOptimized1_Layout.sol"; // DO NOT DELETE
```
```suggestion
```

YoGhurt111 (approved):

davidtaikocha (approved):

ggonzalez94 (approved):

PR #20563, opened 2025-10-28
feat(protocol): store historical block hashes in Anchor

dantaik:
Store historical (not from genesis) block hashes in case the slasher needs to use these values.


davidtaikocha (approved):

ggonzalez94 (commented):
Looks good, but I wonder if we need it. Here's my reasoning:
1. We're going to do slashing using the raw tx list and the anchor block id, so no need to store the blockhash for this.
2. I still think this may be useful information for application builders, but we are already storing this on the SignalService right? We can just make the ring buffer really large on the L2 to avoid it from wrapping.

Let me know what you think

dantaik:
> Looks good, but I wonder if we need it. Here's my reasoning:
> 
> 1. We're going to do slashing using the raw tx list and the anchor block id, so no need to store the blockhash for this.

**Is this the new direction? I thought our goal has always been to verify whether a transaction is included in a block, not just in a raw transaction list. Maybe we should discuss this.

From a user’s perspective, a transaction doesn’t really feel “included” unless it appears in a block. The inclusion guarantee for “in-proposal” is much weaker than “in-block.”**

> 2. I still think this may be useful information for application builders, but we are already storing this on the SignalService right? We can just make the ring buffer really large on the L2 to avoid it from wrapping.
> 
> Let me know what you think

SignalService only keep track of some L1 block hashes (as checkpoints), not historical L2 block hashes.

AnshuJalan:
> From a user’s perspective, a transaction doesn’t really feel “included” unless it appears in a block. The inclusion guarantee for “in-proposal” is much weaker than “in-block.”

My understanding is - the user won't get a positive transaction confirmation receipt if the transaction is in the proposal, but doesn't get included in the block. The proposer won't receive the tip either. This exists at the derivation level.

`rawTxList` vs `blockhash` discussion I feel is oriented more toward what vector the protocol should use to decide if a certain preconfirmation was respected by the preconfer or not - and `rawTxList` provides the same guarantees because it is the "delta" that creates the transition from blockhash B1 to B2

dantaik:
> > From a user’s perspective, a transaction doesn’t really feel “included” unless it appears in a block. The inclusion guarantee for “in-proposal” is much weaker than “in-block.”
> 
> My understanding is - the user won't get a positive transaction confirmation receipt if the transaction is in the proposal, but doesn't get included in the block. The proposer won't receive the tip either. This exists at the derivation level.
> 
> `rawTxList` vs `blockhash` discussion I feel is oriented more toward what vector the protocol should use to decide if a certain preconfirmation was respected by the preconfer or not - and `rawTxList` provides the same guarantees because it is the "delta" that creates the transition from blockhash B1 to B2

I’m not sure I follow. But to me, preconfirmation means the much earlier arrival of new blocks — before the proposal even reaches L1.

From a user’s perspective, if their transaction is executed in one of these preconfirmed blocks, it should be considered "preconfirmed". It would feel odd to the user if their transaction is still labeled as “preconfirmed” but doesn’t actually appear in a block. Details like derivation, blob data, or proposal structures should be abstracted away from users. Explaining that a transaction was “preconfirmed in the proposal data” but somehow never executed would not be an acceptable user experience.

YoGhurt111 (approved):

PR #20568, opened 2025-10-28
refactor(protocol): move Inbox errors inside contract and add MainnetSignalService

dantaik:
## Summary
- Moves error declarations inside the Inbox contract body (instead of after the closing brace)
- Adds MainnetSignalService contract for mainnet deployment (without storage layout comments)
- Updates all test files to use the `Inbox.ErrorName.selector` format for error references
- Delete `packages/protocol/test/layer1/preconf/lookahead-store/LookaheadStore_PostedByOperator.t.sol` as fixing it involves reformatting the whole file which makes further merging hard. Counting @AnshuJalan to re-add the tests in his new PRs.


packages/protocol/script/gen-layouts.sh
```diff
@@ -3,6 +3,7 @@
 # Define the list of contracts to inspect
 # Please try not to change the order
 # Contracts shared between layer 1 and layer 2
+# Contracts shared between layer 1 and layer 2
```
dantaik:
```suggestion
```

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -1035,36 +1035,36 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+    // ---------------------------------------------------------------
+    // Errors
+    // ---------------------------------------------------------------
+
+    error CannotProposeInCurrentBlock();
+    error CheckpointMismatch();
+    error CheckpointNotProvided();
+    error DeadlineExceeded();
+    error EmptyProposals();
+    error ForkNotActive();
```
davidtaikocha: Seems some inbox errors like this `error ForkNotActive();` are not used anymore, shall we also remove these unused errors inside this PR?

packages/protocol/contracts/layer1/mainnet/MainnetSignalService.sol
```diff
@@ -0,0 +1,35 @@
...
+        address remoteSignalService
+    )
+        SignalService(authorizedSyncer, remoteSignalService)
+    { }
+
+    // ---------------------------------------------------------------
+    // Internal Functions
+    // ---------------------------------------------------------------
+
+    function _storeReentryLock(uint8 _reentry) internal override {
```
ggonzalez94, line 28: I may be missing some context, but what's the benefit using `LibFasterReentry`? is it gas savings?

packages/protocol/contracts/layer1/mainnet/MainnetSignalService.sol
ggonzalez94, line 1: shall we also create one for L2? Actually... wouldn't it make sense to just implement the reentracy lock on the main `SignalService` and deploy that on both chains instead of having this overwrites? I don't think there's a reason it cannot be on the base contract and it would simplify the deployment right?

packages/protocol/test/layer1/preconf/lookahead-store/LookaheadStore_PostedByOperator.t.sol
ggonzalez94, line 1: I'm having a hard time tracking what the change here is? maybe we just revert he change on this file and let @AnshuJalan update it once he brings back the tests as part of his PRs

ggonzalez94 (approved):
Looks good, left a few comments. The main one is my suggestion to move the optimized reentrancy logic to the base SignalService and just deploy that directly instead of having specific L1 and L2 implementations

packages/protocol/contracts/layer1/mainnet/MainnetSignalService.sol
dantaik: Yes. And we cannot move LibFasterReentry into shared implementation (SignalService) as our L2 EVM doesn't support the TSTORE opcode. Once we upgraded our EVM, then we should apply these optimisations globally.

packages/protocol/contracts/layer1/mainnet/MainnetSignalService.sol
dantaik: See above

packages/protocol/test/layer1/preconf/lookahead-store/LookaheadStore_PostedByOperator.t.sol
dantaik: Deleted the file so Anshu can merge his new tests easier.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: Removed

davidtaikocha (approved):

PR #20569, opened 2025-10-28
fix(protocol): another approach to fork routing

ggonzalez94:
This is another approach to #20543

## Summary
- Fixes storage layout of the Anchor contract(same as the other PR)
- The SignalService is deployed as a single upgradable contract instead of using the fork router logic. This minimizes the risk of storage layout conflicts. The PR introduces `SignalServiceLegacy`, which implements all the public and internal functions that the old SS needs, and the main contract inherits from it and routes calls based on `block.timestamp`

Please review carefully, one thing that got removed is the resolver pattern, and we just the hardcoded `_remoteSignalService`.

---

> [!NOTE]
> Make SignalService a single upgradeable, fork-aware contract (pre/post-shasta routing) and update Anchor to preserve Pacaya storage layout; adjust deployments, layout, and tests.
> 
> - **SignalService (fork handling)**:
>   - Introduces `LegacySignalService` with legacy signal verification, hop caching, and chain data sync; `SignalService` now inherits it.
>   - Adds `_shastaForkTimestamp` and routes `prove/verifySignalReceived` pre-fork to legacy logic and post-fork to new checkpoint-based verification; caches `_receivedSignals` post-fork.
>   - Removes resolver usage for remote service; uses immutable `_remoteSignalService` and `_authorizedSyncer`.
>   - Persists checkpoints in `_checkpoints`; exposes `saveCheckpoint/getCheckpoint` unchanged.
>   - Updates constructors and all call sites (scripts/tests) to pass `authorizedSyncer`, `remoteSignalService`, and `shastaForkTimestamp`.
> - **Anchor (storage compatibility)**:
>   - Inherits `PacayaAnchorStorage` (with `EssentialContractStorage`) to mirror Pacaya anchor layout; removes local `_pacayaSlots`.
>   - Layout doc updated to reflect new slots.
> - **Infrastructure**:
>   - Notes storage slot usage in `ForkRouter` comments.
>   - Deployment script upgrades `SignalService` with new constructor; tests and helpers updated accordingly.
>   - Minor gas snapshot updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b4e411e2db7f7b5284d1d83dd7e8af565687fb49. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/shared/signal/SignalService.sol
```diff
@@ -96,8 +82,22 @@ contract SignalService is EssentialContract, ISignalService {
...
+        // Pre shasta logic
+            CacheAction[] memory actions = // actions for caching
+            _verifySignalReceivedLegacy(_chainId, _app, _signal, _proof, true);
+
+            uint256 numCacheOps;
+            for (uint256 i; i < actions.length; ++i) {
+                numCacheOps += _cacheLegacy(actions[i]);
+            }
+            return numCacheOps;
+        }
```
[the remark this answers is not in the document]
ggonzalez94: this is ok, we changed the behavior for shasta. In pacaya we do intermediate caching of hops, after shasta when calling this function the whole signal gets cached

packages/protocol/script/layer1/core/DeployProtocolOnL1.s.sol
```diff
@@ -205,7 +205,7 @@ contract DeployProtocolOnL1 is DeployCapability {
             .resolve(uint64(block.chainid), LibNames.B_SIGNAL_SERVICE, false);
 
         // Upgrade with actual inbox address
-        address newImpl = address(new SignalService(shastaInbox, config.remoteSigSvc));
+        address newImpl = address(new SignalService(shastaInbox, config.remoteSigSvc, 0));
         SignalService(signalService).upgradeTo(newImpl);
```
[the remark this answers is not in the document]
ggonzalez94: should be fine for deploying to devnet? cc @davidtaikocha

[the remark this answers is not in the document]
davidtaikocha: Should be fine, i can make it an env var later.

dantaik:
> Please review carefully, one thing that got removed is the resolver pattern, and we just the hardcoded _remoteSignalService.

The original idea was that if we deploy two Signal Services — X and Y — on L2 A and L2 B respectively, and one Signal Service Z on L1, then any pair of Signal Services should be able to verify messages directly.

In this setup, both X and Z act as remote counterparts for Y. Therefore, we need to include the chain ID as part of the lookup key to resolve which chain a message belongs to.

If we want to simplify this design (and maybe we should) and assume that Signal Services are only paired between L1 and L2, then the same simplification should also apply to vaults and bridges, since they follow the same design pattern.

## Concern
I tried to remove resolver from Bridge (https://github.com/taikoxyz/taiko-mono/pull/20578), it's doable but the changes are more risky. I image removing resolver from Vaults are even more risky as there are mapping-storage data that also have chain ID as part of the look up keys.

## Suggestion
I would suggest to make as small changes as possible to the SignalService by restoring using the resolver.

packages/protocol/contracts/layer2/core/PacayaAnchorStorage.sol
```diff
@@ -0,0 +1,18 @@
...
+pragma solidity ^0.8.24;
+
+import { EssentialContractStorage } from "src/shared/common/EssentialContractStorage.sol";
+
+/// @title PacayaAnchorStorage
+/// @dev This contract is used for maintaining compatibility with the storage layout of the Pacaya anchor.
+/// @dev 255 slots were used in Pacaya (EssentialContract, TaikoAnchorDeprecated, etc.).
+/// For a full layout of the Pacaya anchor please refer to
+/// [the layout table](https://github.com/taikoxyz/taiko-mono/blob/taiko-alethia-protocol-v2.3.1/packages/protocol/layout/layer2-contracts.md#taikoanchor)
+contract PacayaAnchorStorage is EssentialContractStorage {
```
dantaik, line 11: if EssentialContractStorage is only used here, shall we merge them into one contract?

packages/protocol/contracts/shared/signal/LegacySignalService.sol
```diff
@@ -0,0 +1,406 @@
...
+
+import "../common/EssentialContract.sol";
+import "../libs/LibTrieProof.sol";
+import "./ISignalService.sol";
+
+/// @title LegacySignalService
+/// @notice SignalService contract pre shasta for compatibility reasons during the fork.
+///          It keeps only the minimal functionality needed during the transition.
+/// @custom:security-contact security@taiko.xyz
+abstract contract LegacySignalService is EssentialContract {
```
dantaik, line 12: It would be helpful to see how this file differs from the current in-production SignalService.
Would it be possible to add the in-production version to the main branch first, renamed as LegacySignalService.sol?
That way, we can easily review the PR diff to understand how the new implementation diverges from the existing one.

packages/protocol/contracts/shared/signal/SignalService.sol
```diff
@@ -96,13 +88,30 @@ contract SignalService is EssentialContract, ISignalService {
...
+
+        // Post shasta logic
         _verifySignalReceived(_chainId, _app, _signal, _proof);
-        _receivedSignals[getSignalSlot(_chainId, _app, _signal)] = true;
+        _receivedSignals[slot] = true;
```
dantaik, line 106: `_receivedSignals` is only used to support the same slot message passing, I think we can remove the function.

dantaik (changes requested):

packages/protocol/contracts/shared/signal/LegacySignalService.sol
dantaik: I would suggest to bring back the  in-production version back, use it as the starting point so that we can have minimal changes... what do you think?

ggonzalez94:
closing in favor of #20543

PR #20545, opened 2025-10-27
refactor(protocol,taiko-client-rs,taiko-client): move LibManifest constants to Derivation.md

dantaik:
Moved existing constants in LibManifest.sol to Derivation.md and added for more constants:

| Constant | Value | Description |
|----------|-------|-------------|
| **INITIAL_BASE_FEE** | `0.025 gwei` (25,000,000 wei) | The initial base fee for the first Shasta block. |
| **MIN_BASE_FEE** | `0.005 gwei` (5,000,000 wei) | The minimum base fee (inclusive) after Shasta fork. |
| **MAX_BASE_FEE** | `1 gwei` (1,000,000,000 wei) | The maximum base fee (inclusive) after Shasta fork. |
| **BLOCK_TIME_TARGET** | `2 seconds` | The block time target. |

PR #20575, opened 2025-10-29
feat(protocol): updates related to `INITIAL_BASE_FEE` in derivation doc

davidtaikocha:
Since we already have `MIN_BASE_FEE` and `MAX_BASE_FEE` after https://github.com/taikoxyz/taiko-mono/pull/20545 , I think the previous consensus `Set first three Shasta block to INITIAL_BASE_FEE` will no longer be needed. So we can change to only let basefee to be `INITIAL_BASE_FEE` for the first Shasta block only when Shasta is from genesis,  to simplify the consensus.

Another benefit we will have is, after switch to activate the Shasta fork based on timestamp, the check for the "first X shasta blocks" will be not as easy and direct as now, and if we have the simplification inside this PR, that check also won't be needed anymore.

### DIFF
--- a/packages/protocol/docs/Derivation.md
+++ b/packages/protocol/docs/Derivation.md
@@ -514,8 +514,6 @@ The following block header fields are also set before transaction execution but 
 
 Note: Fields like `stateRoot`, `transactionsRoot`, `receiptsRoot`, `logsBloom`, and `gasUsed` are populated after transaction execution.
 
-For Shasta specifically, consensus pins the base fee to `INITIAL_BASE_FEE` for the first three blocks after the fork. EIP-4396 derives the next base fee from the parent block time (`parent.timestamp - parent.parent.timestamp`); when the fork starts from genesis this delta may be enormous. Maintaining a three-block fixed-fee window absorbs that anomaly and prevents an outsized jump—the fourth Shasta block and onward go to the normal EIP-4396 calculation.
-
 ### Anchor Transaction
 
 The anchor transaction serves as a privileged system transaction responsible for L1 state synchronization and bond instruction processing. It invokes the `anchorV4` function on the ShastaAnchor contract with precisely defined parameters:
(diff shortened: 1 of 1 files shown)


davidtaikocha:
@claude plz review.

YoGhurt111 (approved):

dantaik (approved):

packages/protocol/docs/Derivation.md
```diff
@@ -569,6 +567,7 @@ The anchor transaction executes a carefully orchestrated sequence of operations:
 
 The calculation of block base fee shall follow [EIP-4396](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4396.md#specification).
 
+The consensus engine pins the base fee at `INITIAL_BASE_FEE` for the very first block when the Shasta fork starts from genesis, because the parent block time (`parent.timestamp - parent.parent.timestamp`) needed for calculation is unavailable. If the fork activates later or once the block height exceeds `1`, base fee computation should follow [EIP-4396](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4396.md#specification), and the calculated value must be clamped within `MIN_BASE_FEE` and `MAX_BASE_FEE`.
```
ggonzalez94, line 570: this means on mainnet and hoodi we won't use `INITIAL_BASE_FEE` since we have previous block history right? we will only use it on devnets were genesis already starts from shasta

ggonzalez94 (approved):

packages/protocol/docs/Derivation.md
davidtaikocha: Yes correct, for mainnet and hoodi, we just use the EIP-4396 to calculate the Shasta basefee (if its the first Shasta block, 2 parent pacaya blocks timestamp will be included in the calculation), and the result will be clamped within `MIN_BASE_FEE` and `MAX_BASE_FEE`.

PR #20581, opened 2025-10-29
feat(protocol): bring back legacy signal service and interface

ggonzalez94:
Context: https://github.com/taikoxyz/taiko-mono/pull/20569#discussion_r2473074727

This PR brings back the old SignalService(taken from this [branch](https://github.com/taikoxyz/taiko-mono/blob/taiko-alethia-protocol-v2.3.1/packages/protocol/contracts/shared/signal/SignalService.sol)) so that it is easier to compare. Bringing back the interface is temporary, and we can remove it as part of #20569 when we move all the structs to the contract itself.

2 modifications were needed:
- We don't have the `LibStrings` library anymore, so those values have been added to the contract directly
- `EssentialContract` does not have a resolve method anymore, so we now inherit from `EssentialResolverContract` instead

**I think it is worth considering, is this simpler than just using the fork router? I'm leaning to think fork router might be simpler if we are careful about the storage slots(which I think we can be)**

---

> [!NOTE]
> Adds `LegacySignalService` and `ISignalServiceLegacy`, and updates storage layout generation to include the legacy contract.
> 
> - **Contracts**:
>   - **`contracts/shared/signal/LegacySignalService.sol`**: Adds legacy signal service implementing cross-chain signal send/verify with hop proofs, caching (state/signal roots), authorization, and events; includes storage for `topBlockId`, `isAuthorized`, `_receivedSignals`.
>   - **`contracts/shared/signal/ISignalServiceLegacy.sol`**: Introduces legacy interface defining `HopProof`, `CacheOption`, signal send/verify APIs, chain data sync/query, and related events.
> - **Layouts/Tooling**:
>   - **`layout/shared-contracts.txt`**: Adds storage layout for `LegacySignalService`.
>   - **`script/gen-layouts.sh`**: Includes `LegacySignalService` in shared contracts for layout generation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 08f0b0635bf3f35bfb1e6cce9cb6d7721ab1e9ab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


dantaik:
> I think it is worth considering, is this simpler than just using the fork router? I'm leaning to think fork router might be simpler if we are careful about the storage slots(which I think we can be)

@ggonzalez94 after seeing the changes in this and previous PRs, I agree, we shall use ForkRouter -- the changes are simpler :D

But in the new implementation, can we still keep data structs names consistent with old implementation to reduce bridge relayer work?

ggonzalez94:
closing in favor of #20543

PR #20584, opened 2025-10-29
feat(protocol): emit bond instructions as part of the Proposed event

ggonzalez94:
Full context can be found in [this](https://taikolabs.slack.com/archives/C083PBWGTK5/p1761646765868189) slack convo and this [message](https://taikolabs.slack.com/archives/C083PBWGTK5/p1761744550907319)

## Summary
Building the bond instructions that need to be sent to the anchor requires decoding `TransitionRecord` from the calldata of propose events that include instructions to finalize records. This logic is complex, and needs to be implemented across all 3 side cars. This PR proposes emitting bond instructions as part of the event to make it easier for off-chain software.

Pros:
- Easier off-chain implementation for side cars
- Less error prone

Cons:
- More complexity in the protocol
- Higher gas usage when emitting the Proposed event, which now grows linearly with the amount of bond instructions. This would usually be a big concerns, but since bond instructions should be fairly rare(they only happen when a prover does not prove on time), this seems like a much lower issue

---

> [!NOTE]
> Emit finalized bond instructions in Proposed events and update encoder/Inbox to compute, aggregate, encode, and decode them.
> 
> - **Core protocol**:
>   - Add `bondInstructions` to `IInbox.ProposedEventPayload` in `IInbox`.
>   - Update `Inbox`:
>     - `_finalize` now aggregates finalized `LibBonds.BondInstruction[]` and returns them.
>     - `_emitProposedEvent` and `propose` emit Proposed with `bondInstructions` (genesis emits empty array).
> - **Encoding**:
>   - Extend `LibProposedEventEncoder` to encode/decode `bondInstructions` and adjust `calculateProposedEventSize`.
> - **Tests**:
>   - Update unit, fuzz, and gas benchmark tests to cover `bondInstructions` in Proposed payloads.
> - **Snapshots**:
>   - Refresh gas snapshots (`shasta-*.json`) reflecting added event data.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 41291c5d4e4fc223288336fc2e6b64f7435c1122. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -950,7 +962,22 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
 
-            return coreState;
+            if (totalBondInstructionCount > 0) {
+                bondInstructions_ = new LibBonds.BondInstruction[](totalBondInstructionCount);
+                uint256 bondInstructionIndex;
+
+                for (uint256 i; i < finalizedCount; ++i) {
+                    LibBonds.BondInstruction[] memory instructions =
+                    _input.transitionRecords[i].bondInstructions;
+                    uint256 instructionsLen = instructions.length;
```
dantaik, line 972: maybe we can simply inline `instructions.length` in the for-loop below as `instructions` in a memory array?

dantaik (approved):

davidtaikocha (approved):

PR #20585, opened 2025-10-29
fix(taiko-client): correct remaining duration in proving window helpers

radik878:
Fix unsigned underflow in remaining duration calculation for both Pacaya and Shasta proving window helpers in util.go by computing a signed difference and clamping at zero when expired. This aligns with the documented function contract (“time remaining”) and prevents returning a huge positive duration when already expired. Added unit tests to validate not-expired (positive remaining) and expired (zero remaining) behavior and to guard against regressions.


davidtaikocha:
@claude plz review.

davidtaikocha (approved):

MatusKysel (approved):

PR #20587, opened 2025-10-29
feat(protocol): add more validations to the anchor

ggonzalez94:
Implements @smtmfft suggestions for the anchor

## Summary

To reduce off-chain complexity and eliminate divergence between the prover/client code and protocol rules, we moved the proposal/block sequencing checks directly into Anchor. The contract now persists the active proposal’s ID, proposer, and next expected block index, rejecting any replay or out-of-order block before running proposal-level side effects, and we documented the stronger guarantee plus added tests covering the new revert paths and the healthy sequential flow.

---

> [!NOTE]
> Persist active proposal progress and enforce blockIndex/proposalId/proposer continuity in Anchor, with docs update and tests for new revert paths.
> 
> - **Contracts (`contracts/layer2/core/Anchor.sol`)**
>   - Add `ProposalProgress` struct and `_proposalProgress` state to track active `proposalId`, `proposer`, and `nextBlockIndex`.
>   - Introduce `_validateProgress` to enforce sequencing:
>     - On `blockIndex == 0`: require strictly increasing `proposalId`; initialize progress.
>     - On subsequent blocks: require matching `proposalId`/`proposer` and exact `blockIndex` continuity; increment `nextBlockIndex`.
>   - Gate `_validateProposal` by new `isFirstBlock` flag; call before block-level validation.
>   - Adjust storage gap to accommodate new state.
> - **Docs (`docs/Derivation.md`)**
>   - Document on-chain enforcement of global, monotonic `_blockIndex` with persisted proposal progress.
>   - Minor table reformatting for constants.
> - **Tests (`test/layer2/core/Anchor.t.sol`)**
>   - Add tests for new sequencing rules and revert paths: `NonZeroBlockIndex`, `ProposalIdMismatch`, `InvalidBlockIndex`.
>   - Add happy-path test for sequential blocks within a proposal.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9d011909fa2e3da4d93784f8ce72a8bd8ad980a6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -405,6 +418,53 @@ contract Anchor is Ownable2Step, ReentrancyGuard {
...
+        ProposalParams calldata _proposalParams,
+        BlockParams calldata _blockParams
+    )
+        private
+        returns (bool isFirstBlock_)
+    {
+        ProposalProgress memory progress = _proposalProgress;
+
+        if (_blockParams.blockIndex == 0) {
+            if (progress.proposalId != 0 && _proposalParams.proposalId <= progress.proposalId) {
```
dantaik, line 435: lets we try to use require-statements to replace if-revert for consistency

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -405,6 +418,53 @@ contract Anchor is Ownable2Step, ReentrancyGuard {
...
+                // Ensure the proposal ID is increasing on the first block
+                revert ProposalIdMismatch();
+            }
+            _proposalProgress = ProposalProgress({
+                proposalId: _proposalParams.proposalId,
+                nextBlockIndex: 1,
```
dantaik, line 441: in the current impl, nextBlockIndex is never assigned to 0. But sometimes, the next blocks index shall be 0. Does this means a bug?

dantaik: when we have nextBlockIndex, we need to **predict** what the next value shall be. How about we use "currentBlockIndex" instead?

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -405,6 +418,53 @@ contract Anchor is Ownable2Step, ReentrancyGuard {
...
+    function _validateProgress(
+        ProposalParams calldata _proposalParams,
+        BlockParams calldata _blockParams
+    )
+        private
+        returns (bool isFirstBlock_)
+    {
+        ProposalProgress memory progress = _proposalProgress;
+
+        if (_blockParams.blockIndex == 0) {
```
davidtaikocha, line 434: I think a goal we wanted to reach is to remove the `_blockParams.blockIndex` and let the `anchorV4` itself checks if this is the first block, and seems it can be done by checking `proposalParams.proposalId` with `proposalProgress.proposalId`.

ggonzalez94:
Closing in favor of #20604

PR #20604, opened 2025-10-30
feat(protocol): simplify anchorV4 by removing the blockIndex

ggonzalez94:
Replaces #20587

This PR aligns the anchor with the simplifications made to the Derivation doc. Instead of relying on the `blockIndex` parameter we just check if the proposalId has increased or not to detect if this is the first block of a proposal or not,

---

> [!NOTE]
> Anchor now gates proposal initialization by increasing proposalId (removing blockIndex), adds proposalId to state, updates docs and tests accordingly.
> 
> - **Layer2/Anchor**:
>   - Remove `BlockParams.blockIndex`; add `_proposalState.proposalId`.
>   - `anchorV4` now:
>     - Validates non-decreasing `proposalId`; reverts on decrease (`ProposalIdMismatch`).
>     - Runs proposal init (prover designation, bond processing) only when `proposalId` increases.
>     - Persists `proposalId` after processing.
> - **Docs (`Derivation.md`)**:
>   - Replace block index semantics with `proposalId` monotonicity; update prover designation and execution flow.
>   - Remove `blockIndex` from `anchorV4` parameters and related narratives.
> - **Tests (`Anchor.t.sol`)**:
>   - Drop `blockIndex` from setups.
>   - Add coverage for same-proposal subsequent blocks, backward `proposalId` revert, transition to new proposal, and multi-block no double-processing.
> - **Layer1/LibManifest**:
>   - Remove comment enforcing global `blockIndex` monotonicity across sources.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8c1d94f93e3de9f33b9ac964eeff7739a15668e3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -222,10 +221,18 @@ contract Anchor is Ownable2Step, ReentrancyGuard {
...
-            _validateProposal(_proposalParams);
+
+        uint48 lastProposalId = _proposalState.proposalId;
+
+        if (_proposalParams.proposalId < lastProposalId) {
+        // Proposal ID cannot go backward
+            revert ProposalIdMismatch();
         }
 
+        // We do not need to account for proposalId = 0, since that's genesis
```
ggonzalez94, line 233: @davidtaikocha is this assumption reasonable? The genesis proposal(when the inbox is activated), will use proposalId = 0, so any blocks produced by proposers should already have proposalId = 1 or greater

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -222,10 +221,17 @@ contract Anchor is Ownable2Step, ReentrancyGuard {
         onlyValidSender
         nonReentrant
     {
-        if (_blockParams.blockIndex == 0) {
-            _validateProposal(_proposalParams);
+        uint48 lastProposalId = _proposalState.proposalId;
+
+        if (_proposalParams.proposalId < lastProposalId) {
```
mikhailUshakoff, line 228:
```suggestion
        if (
            _proposalParams.proposalId != lastProposalId &&
            _proposalParams.proposalId != lastProposalId + 1
        ) {
```
We have only two valid states for `proposalId`: the current one or increased by one. All other changes are not possible to include on L1 since
https://github.com/taikoxyz/taiko-mono/blob/a223e2285c69084e85568246913ebe0322a3a1bb/packages/protocol/contracts/layer1/core/impl/Inbox.sol#L252

mikhailUshakoff: It’s probably not true, since we need to skip some proposals

dantaik (approved):

packages/protocol/contracts/layer2/core/Anchor.sol
dantaik: Really???

packages/protocol/contracts/layer2/core/Anchor.sol
davidtaikocha: Yes, thats correct. Since in `_activateInbox`, we have:
```solidity
coreState.nextProposalId = 1;
```

davidtaikocha (approved):

PR #20614, opened 2025-11-01
fix(protocol): fix SignalService storage layout issue

dantaik:
- Ensure storage layout remains binary-compatible with the [Pacaya fork](https://github.com/taikoxyz/taiko-mono/tree/f599a91e2d6936351a0c92ffe7e04ef28ce8b48c/packages/protocol/contracts/shared/signal).
- Update SignalService to reuse the original struct names to preserve compatibility with the relayer and bridge UI.


packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -123,15 +120,21 @@ contract Anchor is PacayaAnchorStorage, Ownable2Step, ReentrancyGuard {
     // State variables
     // ---------------------------------------------------------------
 
+    /// @notice Mapping from block number to block hash.
+    mapping(uint256 blockNumber => bytes32 blockHash) public blockHashes;
```
ggonzalez94, line 124: I guess it makes sense to reuse the same storage slot for block hashes, that way people can still query the old ones

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -27,7 +24,7 @@ import "./Anchor_Layout.sol"; // DO NOT DELETE
 ///      - Cumulative bond instruction processing with integrity verification
 ///      - State tracking for multi-block proposals
 /// @custom:security-contact security@taiko.xyz
-contract Anchor is PacayaAnchorStorage, Ownable2Step, ReentrancyGuard {
+contract Anchor is EssentialContract {
```
ggonzalez94, line 27: Originally we had decided to make the anchor immutable, and just route trough the proxy. This PR takes it back to using a proxy, similar to the SS. This would allow us to later upgrade it and remove the router.
I'm on board with that direction, just wanted to clarify that that's why we had PacayaAnchorStorage before

packages/protocol/contracts/shared/signal/ISignalService.sol
```diff
@@ -12,22 +12,35 @@ import "./ICheckpointStore.sol";
...
+    /// @dev DEPRECATED
+    /// @dev Caching is no longer supported
+    enum CacheOption {
+        CACHE_NOTHING,
+        CACHE_SIGNAL_ROOT,
+        CACHE_STATE_ROOT,
+        CACHE_BOTH
+    }
+
+    struct HopProof {
```
ggonzalez94, line 24: I'm still not quite sure I understand why we need to restore the old struct names for bridge and relayers to work. As far as I understand:
- the bindings have already been generated, so this won't cause compile issues(but maybe I'm missing something here?)
- as long as we keep a compatible struct(note that was the goal of the struct `Proof`) any on-chain or off-chain software should continue functioning as expected

ggonzalez94 (commented):
Looks good. Just left one comment on having to restore the struct names

packages/protocol/contracts/layer2/core/Anchor.sol
dantaik: Sure, BTW, I still prefer to use ForkRouter for Anchor so in the new impl, we don't have all the old code. Later we can upgrade again to get rid of ForkRouter if we want.

packages/protocol/contracts/shared/signal/ISignalService.sol
dantaik: I see your point about renaming "HopProof" for better naming conventions. I checked the codebase and noticed it's still used in [the **relayer** and **bridge-ui** packages]( https://github.com/search?q=repo%3Ataikoxyz%2Ftaiko-mono%20HopProof&type=code). Maybe it's not worthy our valuable time for what might be a small name change.

PR #20618, opened 2025-11-02
Chore: remove redundant Vec import from BlobCoder

Doryu001:
Drop the unnecessary `use std::vec::Vec;` from `BlobCoder`rely on the standard prelude import of `Vec`, keeping the module tidy


davidtaikocha:
If the original code passes the checks in  `cargo fmt` and `clippy` in ci, then it should be fine :)

PR #20622, opened 2025-11-03
feat(taiko-client-rs): index `anchor::BondInstruction` instead of `codec::BondInstruction`

davidtaikocha:
Feedback in slack: 
> btw related to bond_instructions on taiko-client-rs
event indexer use bindings::codec_optimized::LibBonds
but anchor use bindings::anchor::LibBonds


pamungkaski (commented):
LGTM, tho the best solutions would be to make sure that there is only 1 LibBonds declarations but seems like it is not possible yeah

davidtaikocha:
> tho the best solutions would be to make sure that there is only 1 LibBonds declarations but seems like it is not possible yeah

Yes by using the current `forge bind`, its not easy to achieve.

MatusKysel (approved):
lgtm

YoGhurt111 (approved):

PR #20623, opened 2025-11-03
feat(taiko-client): add support of access list

MatusKysel:


davidtaikocha:
After discussing with @MatusKysel , we will impl this feature in rust client later.

PR #20630, opened 2025-11-03
Chore: drop redundant eslint suppression in whitelist mint count

Doryu001:
Remove the unused-variable suppression comment from totalWhitelistMintCount rely on tuple destructuring to ignore the first entry without lint overrides keep the loop logic intact while aligning with eslint expectations


davidtaikocha:
Closed since `packages/taikoon-ui` is not used atm.

PR #20641, opened 2025-11-04
feat(protocol): revert pacaya proposals after shasta fork timestamp

### LINKED ISSUE #20636: Upgrade pacaya inbox to reject proposals after shasta fork timestamp
For added safety the old inbox should reject any proposal that happens after the fork timestamp

Should we upgrade the pacaya inbox to avoid accidental proposals(that will never be able to be carried over to shasta) after the fork timestamp? If yes, then we need to modify the current deployed branch since we don't have the source code for the inbox on main anymore

_Originally posted by @ggonzalez94 in https://github.com/taikoxyz/taiko-mono/pull/20629#discussion_r2487670664_

ggonzalez94:
Fixes #20636

But reverts on the fork router instead of the inbox, since it is already the entrypoint so easier to just do it there and not having to deploy the inbox again.

I also updated all existing scripts and tests to set `_shastaForkTimestamp = type(uint64).max` to signal fork will not happen so that existing tests and more work.

---

> [!NOTE]
> PreconfRouter now reverts proposals at/after a configurable Shasta fork timestamp; all deploy/upgrade scripts and tests pass type(uint64).max by default.
> 
> - **Contracts**
>   - **`PreconfRouter`**: 
>     - Add immutable `shastaForkTimestamp` and `ShastaForkAlreadyActivated` error.
>     - Guard `_proposeBatch` to revert once `block.timestamp >= shastaForkTimestamp`.
>     - Update constructor to accept `_shastaForkTimestamp`.
> - **Scripts**
>   - Update all preconf deployment/upgrade scripts to pass `type(uint64).max` for `shastaForkTimestamp`.
>   - based deploy script: add `GENESIS_TIMESTAMP` env var wiring to whitelist deployment.
> - **Tests**
>   - Adjust tests to use new `PreconfRouter` constructor arg and ensure router/wrapper/fork upgrades align.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 30752c9aa8968b58d6812ae631f05e4e4499ac7a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
+++ b/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
@@ -12,8 +12,10 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
     IProposeBatch public immutable proposeBatchEntrypoint;
     IPreconfWhitelist public immutable preconfWhitelist;
     address public immutable fallbackPreconfer;
+    uint64 public immutable shastaForkTimestamp;
 
     error InvalidLastBlockId(uint96 _actual, uint96 _expected);
+    error ShastaForkAlreadyActivated();
 
     uint256[50] private __gap;
 
@@ -29,7 +31,8 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
     constructor(
         address _proposeBatchEntrypoint, // TaikoInbox or TaikoWrapper
         address _preconfWhitelist,
-        address _fallbackPreconfer
+        address _fallbackPreconfer,
+        uint64 _shastaForkTimestamp
     )
         nonZeroAddr(_proposeBatchEntrypoint)
         nonZeroAddr(_preconfWhitelist)
@@ -38,6 +41,7 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
         proposeBatchEntrypoint = IProposeBatch(_proposeBatchEntrypoint);
         preconfWhitelist = IPreconfWhitelist(_preconfWhitelist);
         fallbackPreconfer = _fallbackPreconfer;
+        shastaForkTimestamp = _shastaForkTimestamp;
     }
 
     function init(address _owner) external initializer {
@@ -84,6 +88,11 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
         onlyFromPreconferOrFallback
         returns (ITaikoInbox.BatchMetadata memory meta_, uint64 lastBlockId_)
     {
+        uint64 currentTimestamp = uint64(block.timestamp);
+        if (currentTimestamp >= shastaForkTimestamp) {
+            revert ShastaForkAlreadyActivated();
+        }
+
         // Both TaikoInbox and TaikoWrapper implement the same ABI for proposeBatch.
         (meta_, lastBlockId_) = IProposeBatch(proposeBatchEntrypoint).proposeBatch(_params, _txList);
 
--- a/packages/protocol/contracts/shared/tokenvault/BaseNFTVault.sol
+++ b/packages/protocol/contracts/shared/tokenvault/BaseNFTVault.sol
@@ -19,7 +19,7 @@ abstract contract BaseNFTVault is BaseVault {
         string name;
     }
 
-    /// @devStruct representing the details of a bridged token transfer operation.
+    /// @dev representing the details of a bridged token transfer operation.
     /// 5 slots
     struct BridgeTransferOp {
         // Destination chain ID.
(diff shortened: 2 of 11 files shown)


packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -38,6 +41,7 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
         proposeBatchEntrypoint = IProposeBatch(_proposeBatchEntrypoint);
         preconfWhitelist = IPreconfWhitelist(_preconfWhitelist);
         fallbackPreconfer = _fallbackPreconfer;
+        shastaForkTimestamp = _shastaForkTimestamp;
```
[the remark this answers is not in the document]
ggonzalez94: fixed

ggonzalez94:
@YoGhurt111 we'll need to create an upgrade script for this. Once we review this do you wanna do it?

YoGhurt111 (approved):

YoGhurt111:
> @YoGhurt111 we'll need to create an upgrade script for this. Once we review this do you wanna do it?

Yep, I can do it

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
```diff
@@ -12,8 +12,10 @@ contract PreconfRouter is EssentialContract, IPreconfRouter {
     IProposeBatch public immutable proposeBatchEntrypoint;
     IPreconfWhitelist public immutable preconfWhitelist;
     address public immutable fallbackPreconfer;
+    uint64 public immutable shastaForkTimestamp;
```
davidtaikocha, line 15: Shall we move this `immutable shastaForkTimestamp` into `TaikoInbox`? In case we want to run based-proposing during the transition?

dantaik: In Shasta, we no longer have such a router, so I think having shastaForkTimestamp here or the old TaikoInbox will both work. @ggonzalez94 correct?

dantaik: Correct me if I'm wrong: my understanding is that the protocol entry point in Pacaya is this PreconfRouter, in Shasta, it's the Inbox contract.

dantaik (approved):

packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol
ggonzalez94: Correct. And even if we wanna run based proposing before the fork we can do it via the `PreconfRouter` using the [fallbackProposer](https://github.com/taikoxyz/taiko-mono/blob/taiko-alethia-protocol-v2.3.1/packages/protocol/contracts/layer1/preconf/impl/PreconfRouter.sol#L22), we just need to turn off the preconfer if I remember correctly.

But my idea for how this should happen is:
1. A few minutes before the fork we remove all the other preconfers. Only taiko labs is the preconfer for a few epochs before the fork
2. We continue preconfing as usual and submit a batch a few slots before the fork(to give us room for the tx to land)
3. We stop preconfing and proposing for a few slots. This will cause the chain to not produce blocks for a few minutes, but I think that's acceptable
4. We activate the shasta inbox after the fork has happened
5. We start preconfing as usual using the new shasta structure and submit our first proposal to the shasta inbox (we could even start preconfing before we activate as long as we make sure the proposal lands **after** the activation)

What do you think @davidtaikocha ? do you see any benefit from running based proposing instead of preconfs during this period?

davidtaikocha: > What do you think @davidtaikocha ? do you see any benefit from running based proposing instead of preconfs during this period?

Not 100% sure, but imo in current sidecar code impl, after one operator submits its proposal to L1, the next operator (in this case it will be still us) won't wait but immediately start the next round of preconf (in handover window). So maybe we need to do some extra changes in sidecar for that? Otherwise at the fork time, probly some latest Pacaya blocks will be reorged, and we need to check that before send the correct canonical L2 hash to `shastaInbox.activate`?

ggonzalez94: Let's discuss during today's call

Regardless of what direction we take both can be achieved via the `PreconfRouter` more easily than upgrading the inbox and removing preconf entirely(since that can only happen via a contract upgrade). If we wanna run based proposing the only thing we need to do is remove all operators(including Taiko's preconfer) and use the fallback preconfer instead. But I think we will have the same issue as what you are describing that there will be some reorgs(unless you time it perfectly to remove the operator at the handover window)

PR #20645, opened 2025-11-05
fix(protocol): handle L1 reorg for `activateInbox`

dantaik:
After activateInbox is called, the last Pacaya block proposal transaction could still be reorged out of L1, which would invalidate the parameters previously used by activateInbox.
At that point, the activator has already been cleared (set to zero), leaving the chain in an invalid state.

This PR updates the logic to allow activateInbox to be called multiple times safely.


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -394,11 +421,17 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
     /// @dev Activates the inbox with genesis state so that it can start accepting proposals.
-    /// Sets up the initial proposal and core state with genesis block
-    /// @param _genesisBlockHash The hash of the genesis block
-    function _activateInbox(bytes32 _genesisBlockHash) internal {
+    /// Sets up the initial proposal and core state with genesis block.
+    /// Can be called multiple times to handle L1 reorgs or correct incorrect values.
+    /// Resets state variables to allow fresh start.
+    /// @param _lastPacayaBlockHash The hash of the last Pacaya block
+    function _activateInbox(bytes32 _lastPacayaBlockHash) internal {
+        require(_lastPacayaBlockHash != 0, InvalidLastPacayaBlockHash());
```
```suggestion
```

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -184,13 +197,26 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+    ///      exist and `propose` will revert with `ProposalHashMismatch()`.
+    ///      This function can be called multiple times to handle L1 reorgs where the last Pacaya
+    ///      block may change after this function is called.
+    /// @param _lastPacayaBlockHash The hash of the last Pacaya block
+    function activate(bytes32 _lastPacayaBlockHash) external {
         require(msg.sender == _shastaInitializer, ACCESS_DENIED());
-        _activateInbox(_genesisBlockHash);
+        require(_lastPacayaBlockHash != 0, InvalidLastPacayaBlockHash());
+        _activateInbox(_lastPacayaBlockHash);
+        emit InboxActivated(_lastPacayaBlockHash);
```
ggonzalez94, line 219: I don't think we need a new event since we still emit a `Proposed` event in `_activate`? And this way we avoid increasing the bytecode of the contract

ggonzalez94 (commented):
@dantaik I like this idea that we have this reorg protection, but I don't think we should allow the owner to reset the initializer without restrictions:
- The owner is likely to be a multisig, so it won't be able to sign the tx fast enough to overwrite the genesis hash in the case of an L1 reorg, thus being innefective
- Having this indefinite power to overwrite the initial hash might hurt our classification to stage 1 or being considered more trustelss
- Once we are sure a reorg has not happened, there's no reason to keep this option open


I have a proposal to do this more trustless. Please check #20649

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: I do think we need a new one here to make activation more obvious. We also should remove `activate` and related code later with another minor upgrade.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -394,11 +419,15 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
     /// @dev Activates the inbox with genesis state so that it can start accepting proposals.
-    /// Sets up the initial proposal and core state with genesis block
-    /// @param _genesisBlockHash The hash of the genesis block
-    function _activateInbox(bytes32 _genesisBlockHash) internal {
+    /// Sets up the initial proposal and core state with genesis block.
+    /// Can be called multiple times to handle L1 reorgs or correct incorrect values.
+    /// Resets state variables to allow fresh start.
+    /// @param _lastPacayaBlockHash The hash of the last Pacaya block
+    function _activateInbox(bytes32 _lastPacayaBlockHash) internal {
+        conflictingTransitionDetected = false;
```
ggonzalez94, line 428: this starts at false by default. Any reason to set it to true explicitely?

dantaik:
@claude please review

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -113,8 +130,9 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     // State Variables
     // ---------------------------------------------------------------
 
-    /// @dev The address responsible for calling `activate` on the inbox.
-    address internal _shastaInitializer;
+
+    /// @notice The timestamp when the first activation occurred.
+    uint48 public activationTimestamp;
```
ggonzalez94, line 135: maybe make it private or internal to avoid creating a view function which will further bloat the contract size? (already close to its limit). The timestamp can be checked on-chain based on when the tx was executed or the event emitted

ggonzalez94 (approved):
LGTM, just left a couple small suggestions

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: Just to make sure when _activateInbox is called, all states reverts to defaults.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: How about we do another minor-upgrade to remove all activation related code before we transfer the ownership back to the DAO?

YoGhurt111 (approved):

PR #20646, opened 2025-11-05
fix(protocol): update signature hash calculation and refine instance validation

smtmfft:
if there is no specific reason to calculate hash like `signatureHash = hash(instance, publicInputHash)`, we can use instance directly in public input hash, e.g.
```
        bytes32 signatureHash = LibPublicInput.hashPublicInputs(
                _aggregatedProvingHash, address(this), address(instance), taikoChainId
            );
```
which saves 1 hash and seems has no side effect.


YoGhurt111 (approved):

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
```diff
@@ -134,21 +134,17 @@ contract SgxVerifier is IProofVerifier, Ownable2Step {
...
+        uint32 id = uint32(bytes4(_proof[:4]));
         address instance = address(bytes20(_proof[4:24]));
+        require(_isInstanceValid(id, instance), SGX_INVALID_INSTANCE());
 
-        bytes32 signatureHash = EfficientHashLib.hash(
-            bytes32(uint256(uint160(instance))),
-            LibPublicInput.hashPublicInputs(
-                _aggregatedProvingHash, address(this), address(0), taikoChainId
-            )
+        bytes32 signatureHash = LibPublicInput.hashPublicInputs(
```
YoGhurt111, line 141: Looks like we can remove the param `_newInstance` in `LibPublicInput.hashPublicInputs` and use `EfficientHashLib.hash`

YoGhurt111 (approved):

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
smtmfft: that's ok, but this way has another benefit that saves 1 hash.

smtmfft:
@codex review

davidtaikocha (approved):

dantaik (approved):

ggonzalez94 (approved):
LGTM

PR #20649, opened 2025-11-05
feat(protocol): a proposal to handle reorgs on activate trustlessly

ggonzalez94:
Builds on top of #20645, but instead of allowing the owner to activate the inbox many times without restrictions it uses the blockhash to detect if there was an L1 fork. Activate can only be called the first time OR when the blockhash of the block number when the last pacaya batch was submitted is different

---

> [!NOTE]
> Makes Inbox.activate trustlessly reorg-aware by requiring genesis hash and pacaya block number, adding blockhash checks and state; updates storage layouts, deployment script, tests, and snapshots.
> 
> - **Core (Inbox)**:
>   - Change `activate` to `activate(bytes32 genesisBlockHash, uint256 pacayaBlockNumber)` with reorg detection via `blockhash(pacayaBlockNumber)` and new state (`_pacayaBlockHash`, `_pacayaBlockNumber`).
>   - Remove `resetInitializer` and related events; add errors `NoForkDetected`, `InvalidActivateParams`.
>   - Update `_activateInbox` to take `_genesisBlockHash`.
> - **Storage Layouts**:
>   - Add `_pacayaBlockHash` and `_pacayaBlockNumber` to `Inbox`, `InboxOptimized1`, `InboxOptimized2`, `DevnetInbox`, `MainnetInbox` layouts; adjust `__gap`s.
> - **Scripts**:
>   - `DeployProtocolOnL1.s.sol`: load `PACAYA_BLOCK_NUMBER` (default `block.number - 1`) and pass to `Inbox.activate`.
> - **Tests/Helpers**:
>   - Add `_preparePacayaBlock` helper; update deployers and init tests for new `activate` API and reorg scenarios (fork detected/mismatch/param checks/unauthorized).
> - **Snapshots/Gas**:
>   - Refresh gas reports and `shasta-*.json` snapshots with minor deltas.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 68a719766797a0224c4c905d02c4f8022fefc7dd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -150,7 +150,14 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
 
-    uint256[37] private __gap;
+    /// @dev L1 block hash when the last Pacaya batch was proposed
+    ///      Only used during activation to safeguard against reorgs
+    bytes32 private _pacayaBlockHash;
+    /// @dev L1 block number when the last Pacaya batch was proposed
+    ///      Only used during activation to safeguard against reorgs
+    uint256 private _pacayaBlockNumber;
+
+    uint256[50] private __gap;
```
ggonzalez94, line 152: since this is a fresh deployment to a new address it didn't make sense we use a smaller gap I believe

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -203,20 +210,24 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
-        emit InitializerReset();
+        require(genesisBlockHash != 0, InvalidActivateParams());
+        require (pacayaBlockNumber != 0, InvalidActivateParams());
+
+        // This returns the blockhash for the last 256 blocks, which should be more than enough to detect a reorg.
+        bytes32 pacayaBlockHash = blockhash(pacayaBlockNumber);
+        
+        // Detect if there was a reorg
+        require(pacayaBlockHash != 0 && pacayaBlockHash != _pacayaBlockHash, NoForkDetected());
+        require(_pacayaBlockNumber == 0 || _pacayaBlockNumber == pacayaBlockNumber, NoForkDetected());
```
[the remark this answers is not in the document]
ggonzalez94: I believe this is a feature not a bug 😃 Blocks are considered final after 64 slots, so 256 is more than enough. And we want to make this feature unusable afterwards

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -203,20 +202,26 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     ///      exist and `propose` will revert with `ProposalHashMismatch()`.
     ///      This function can be called multiple times to handle L1 reorgs where the last Pacaya
     ///      block may change after this function is called.
-    /// @param _lastPacayaBlockHash The hash of the last Pacaya block
-    function activate(bytes32 _lastPacayaBlockHash) external {
+    /// @param genesisBlockHash The hash of the genesis block(as of the latest pacaya block)
```
```suggestion
    /// @param genesisBlockHash The hash of the genesis block (as of the latest pacaya block)
```

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -203,20 +202,26 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     ///      exist and `propose` will revert with `ProposalHashMismatch()`.
     ///      This function can be called multiple times to handle L1 reorgs where the last Pacaya
     ///      block may change after this function is called.
-    /// @param _lastPacayaBlockHash The hash of the last Pacaya block
-    function activate(bytes32 _lastPacayaBlockHash) external {
+    /// @param genesisBlockHash The hash of the genesis block (as of the latest pacaya block)
+    /// @param pacayaBlockNumber The number of the latest pacaya block when the `genesisBlockHash` was calculated.
+    /// This is used for revert protection.
+    function activate(bytes32 genesisBlockHash, uint256 pacayaBlockNumber) external {
```
dantaik, line 208: use a leading "_" for input params please

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -1151,5 +1156,6 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     error TransitionRecordHashMismatchWithStorage();
     error TransitionRecordNotProvided();
     error UnprocessedForcedInclusionIsDue();
-    error InvalidLastPacayaBlockHash();
+    error NoForkDetected();
```
dantaik, line 1159: sort errors

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: naming them _lastPacayaBlockNumber and _lastPacayaBlockHash is less confusing

dantaik:
@ggonzalez94 the PR make the activate function less trust-based, but given that the two parameters `genesisBlockHash` and `pacayaBlockNumber` are both picked by the activator, making the code less trust-based doesn't offer much benefit.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: I did this because we also have a naming convention for internal/private variables and `_pacayaBlockNumber` is used for that.
Regarding `_lastPacayaBlockHash` I agree it is a better name. Will update accordingly

ggonzalez94:
> @ggonzalez94 the PR make the activate function less trust-based, but given that the two parameters `genesisBlockHash` and `pacayaBlockNumber` are both picked by the activator, making the code less trust-based doesn't offer much benefit.

While I do agree it doesn't remove any trust assumptions on the activator being able to set any blockhash they choose(this was a necessary trust assumption for our fork) I would argue that it does eliminate additional trust assumptions. Let me try to convince you why:
- under an owner based reactivation with no restrictions they privileged address can reactivate the inbox as many times as they wish. This is further worsen by the fact that given that we are using a ring buffer for proposals, they could not only set the genesis block hash, but replace the proposal at index 0 forever until the inbox is upgraded and this function removed.
- the new activation mechanism does solve this. Once the activator has chosen the block number which they claim corresponds to the latest pacaya batch proposal, they need to prove there was a reorg if they want to call this function again.

But maybe we don't even need to do any of this. In practice waiting 10-12 blocks is enough to be convinced with a very very high degree of confidence there will be no reorg. If we wait 2-3 minutes between the last pacaya batch proposal and the moment we activate the inbox it is very unlikely we'll have an issue(e.g. Circle waits 12 blocks to [mint USDC](https://developers.circle.com/circle-mint/blockchain-confirmations)) and in the worst case there is a reorg we can upgrade the inbox to solve the issue.

dantaik:
Gustavo, thanks for sharing your additional thoughts. My main point is that since this function relies on trust, its input parameters are manually provided and therefore prone to human error. For instance, the genesisBlockHash value could be incorrect (e.g., a wrong hash was copied). In such a case, the operator would need to re-run the function — not due to a reorg, but simply because of human error.

ggonzalez94:
after discussing we are closing it in favor of the simpler approach with a time restriction of only 2 hours for reactivation

PR #20653, opened 2025-11-06
fix(taiko-client): log correct error from godotenv.Load in LoadEnv

Snezhkko:
Capture and log the actual error returned by godotenv.Load in LoadEnv() instead of reusing the earlier os.Getwd error variable.
This ensures failures to load the test .env file are diagnosed correctly. Other usages either ignore the error explicitly or handle it directly; this was an isolated bug, not intentional.


MatusKysel (approved):
lgtm

davidtaikocha (approved):

PR #20663, opened 2025-11-06
fix(taiko-client-rs): use explicit index for blob version byte

hawkadrian:
Fix blob version byte indexing by replacing VERSIONED_HASH_VERSION_KZG with explicit index constant and adding bounds check.


davidtaikocha:
Can't see any benefit from this, the new constant value is just the same from the one in `alloy_eips::eip4844::VERSIONED_HASH_VERSION_KZG`.

PR #20302, opened 2025-09-29
fix(protocol): apply randomnessDelay in PreconfWhitelist operator selection

GarmashAlex:
- Use randomnessDelay when computing the randomness timestamp in _getOperatorForEpoch by shifting the epoch start back by randomnessDelay * SECONDS_IN_EPOCH (saturating at zero shift to avoid underflow).
- This makes the existing randomnessDelay state variable functional; previously it was initialized but unused.
- Motivation: avoids unintended bias from querying the exact epoch boundary, aligns with the documented intent of delayed randomness and existing tests that reference randomnessDelay, and improves resilience without changing _getRandomNumber semantics or public interfaces.

---

> [!NOTE]
> Apply `randomnessDelay` when deriving epoch randomness in `PreconfWhitelist`, set production/test configs to delay=2 epochs, add stability test, and update gas snapshots.
> 
> - **Contracts**:
>   - `PreconfWhitelist._getOperatorForEpoch`: use `randomnessDelay` (in epochs) to shift the randomness timestamp before calling `_getRandomNumber`.
> - **Deployment**:
>   - `DeployProtocolOnL1.s.sol`: initialize `PreconfWhitelist` with `randomnessDelay = 2`.
> - **Tests**:
>   - `PreconfWhitelistSetup.sol`: initialize test whitelist with `randomnessDelay = 2`.
>   - `PreconfWhitelist.t.sol`:
>     - Add test ensuring operator selection remains stable across an epoch boundary.
>     - Beacon root mock returns zero for future timestamps; advance time to account for delay.
> - **Snapshots**:
>   - Update Shasta gas snapshots (`shasta-finalize.json`, `shasta-propose.json`, `shasta-prove.json`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8a177e383a44a2d3a780c9f8168c31b01996a1de. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #20671, opened 2025-11-07
feat(protocol): a much simpler whitelist

ggonzalez94:
**This should be merged after #20302**

## Summary

Our current `PreconfWhitelist` is too complicated and the PR that fixes the bug made me remember that. 

This PR significantly simplifies it while maintaining storage compatibility, and avoids having to call `consolidate` manually. **This is a major benefit, since the gas difference between when we know all operators are active and not is large(it saves at least 4 `SLOADs`)**, and we are currently not taking advantage of it since we are not always calling `consolidate`.

The benfits are:
- much simpler code
- we always remove operators immediately(because they are ejected when they don't produce a block), so we can avoid having to manually consolidate and having all that conditional logic
- currently if we don't call consolidate the function that checks the proposer is very expensive(because we need to iterate over all operators). I don't think we are doing this often on mainnet. With this PR this is not a concern anymore since we use the timestamp to detect that.

## Changes
- Simplify the whitelist to remove a lot of redundant and unnecessary code.
- Use a timestamp detection mechanism to know if all validators are active or not
- Remove validators immediately(since we only did that under ejection scenarios)

---

> [!NOTE]
> Reworks `PreconfWhitelist` to a simpler, storage-compatible design using timestamp-based detection, removing consolidation logic and immediately ejecting inactive operators, with updates to tests, deploy script, snapshots, and gas reports.
> 
> - **Layer1 Contracts (Preconf Whitelist)**
>   - Simplifies `PreconfWhitelist` implementation in `contracts/layer1/preconf/impl/PreconfWhitelist.sol` with a leaner API and logic.
>   - Introduces/updates storage layout in `impl/PreconfWhitelist_Layout.sol` to remain storage-compatible.
>   - Updates `iface/IPreconfWhitelist.sol` to match the new interface (removes consolidation-related flows).
>   - Switches to timestamp-based detection for “all validators active” and immediate operator ejection/removal.
> - **Tooling and Integration**
>   - Updates deploy script `script/layer1/core/DeployProtocolOnL1.s.sol` to use the new whitelist.
>   - Refreshes snapshots in `snapshots/shasta-*.json` and gas report `gas-reports/layer1-contracts.txt`.
>   - Adjusts tests in `test/layer1/preconf/whitelist/PreconfWhitelist.t.sol` and setup helpers `test/layer1/core/inbox/common/PreconfWhitelistSetup.sol`; updates test config `test/genesis/tsconfig.json`.
>   - Minor package metadata changes in `packages/protocol/package.json`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 480e4cc6d4e83527ec537f8f6f00c4959cc145c6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -234,98 +183,76 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist, IProposerChec
...
-            return operatorMapping[rand % _operatorCount];
-        } else {
-            address[] memory candidates = new address[](_operatorCount);
-            uint256 count;
+            uint256 target = _getRandomNumber(randomnessTs) % _operatorCount;
+            
             for (uint256 i; i < _operatorCount; ++i) {
-                address operator = operatorMapping[i];
+                address operator = operatorMapping[target];
                 if (isOperatorActive(operator, _epochTimestamp)) {
```
ggonzalez94, line 269: the fact that we need to check if an operator is active makes this more expensive. We do this now because when adding an operator you want to avoid that changing the selection mid-epoch.

ggonzalez94: but maybe since operatorCount changes anyways this is fine? that already changes the operator selection anyways

ggonzalez94: the old logic actually prevented this, will have to think more about this

ggonzalez94:
> If we can launch phase2 preconf in Q1, maybe we can simply live with the current whitelist as it will not be used for long?

@dantaik here's why I think it is worth doing it(besides the fact that is much simpler)
1. We cannot use the `PreconfWhitelist` currently deployed on mainnet since it is not compatible with shasta(it does not have the `checkProposer` function), so we need to upgrade regardless.
2. The version in `main` already has optimizations that the currently deployed one doesn't. Please check this [PR](https://github.com/taikoxyz/taiko-mono/pull/20201/files#diff-4701131abf8c45a1eed585742b1684600922d316516cb62dd6d8e46b07fc39ca) we merged back in September, where we already trimmed ~10k gas per proposal
3. The previous lookahead relied on someone calling `consolidate` for being optimal(not having to scan trough the entire operator set). I don't think we are always doing that when adding/removing operators on mainnet, or it may be buggy because `havingPerfectOperators` is false every time I've queried it. This PR achieves the same gas efficiency, without having to manually call it(by using a timestamp mechanism)
4. Even after permisionless preconfs, we will still use this contract(it serves as the fallback preconfer when no validator has opted in), so we'll have it as part of our system for the forseable future

**So TL;DR: We need to upgrade the current implementation to make it compatible with shasta, and to save significant gas. So we might as well simplify it and audit it**

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -18,13 +18,19 @@ interface IPreconfWhitelist {
         address indexed proposer, address indexed sequencer, uint256 inactiveSince
     );
 
+    /// @notice Emitted when an ejecter is updated.
+    /// @param ejecter The address of the ejecter.
+    /// @param isEjecter Whether the address is an ejecter.
+    event EjecterUpdated(address indexed ejecter, bool isEjecter);
```
cyberhorsey, line 24: its spelled `ejector`

cyberhorsey (dismissed):

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
```diff
@@ -18,13 +18,19 @@ interface IPreconfWhitelist {
...
+
     error CannotRemoveLastOperator();
     error InvalidOperatorIndex();
     error InvalidOperatorCount();
     error InvalidOperatorAddress();
     error OperatorAlreadyExists();
     error OperatorAlreadyRemoved();
     error OperatorNotAvailableYet();
+    error NoActiveOperatorRemaining();
```
dantaik: lets move all errors to implementations as in other files

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -34,13 +41,19 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist, IProposerChec
...
+    /// This includes both active and inactive operators.
     uint8 public operatorCount;
-    /// @dev The number of epochs to delay for operator changes.
-    uint8 public operatorChangeDelay;
-    /// @dev The number of epochs to delay for randomness seed source.
-    uint8 public randomnessDelay;
-    /// @dev all operators in operatorMapping are active and none of them are to be deactivated.
-    bool public havingPerfectOperators;
+    /// @dev Deprecated variable. Kept for storage compatibility.
+    uint8 private deprecatedOperatorChangeDelay;
```
dantaik:
```suggestion
    uint8 private _deprecatedOperatorChangeDelay;
```

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -34,13 +41,19 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist, IProposerChec
...
+    /// @dev Deprecated variable. Kept for storage compatibility.
+    uint8 private deprecatedRandomnessDelay;
```
dantaik:
```suggestion
    uint8 private _deprecatedRandomnessDelay;
```

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -34,13 +41,19 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist, IProposerChec
...
+    /// @dev Deprecated variable. Kept for storage compatibility.
+    bool private deprecatedHavingPerfectOperators;
```
dantaik:
```suggestion
    bool private _deprecatedHavingPerfectOperators;
```

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -51,95 +64,28 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist, IProposerChec
...
-        havingPerfectOperators = true;
-    }
-
-    function setOperatorChangeDelay(uint8 _operatorChangeDelay) external onlyOwner {
-        operatorChangeDelay = _operatorChangeDelay;
-        emit OperatorChangeDelaySet(_operatorChangeDelay);
     }
 
     /// @inheritdoc IPreconfWhitelist
+    /// @dev NOTE: The operator only becomes active after `OPERATOR_CHANGE_DELAY` epochs.
```
dantaik:
```suggestion
    /// @dev The operator only becomes active after `OPERATOR_CHANGE_DELAY` epochs.
```

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -15,15 +14,23 @@ import "./PreconfWhitelist_Layout.sol"; // DO NOT DELETE
...
-    event EjecterUpdated(address indexed ejecter, bool isEjecter);
-
+    // ---------------------------------------------------------------
+    // Constants
+    // ---------------------------------------------------------------
+    /// @dev The number of epochs before a newly added operator becomes active.
+    uint8 public constant OPERATOR_CHANGE_DELAY = 2;
+    /// @dev The number of epochs to use as delay when selecting an operator.
+    ///      This needs to be 2 epochs or more to ensure the randomness seed source is stable across epochs.
+    uint8 public constant RANDOMNESS_DELAY = 2;
```
dantaik, line 29: should we change these constants to MIN_XXXX, then introduce immutable configurable DELAYs?

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -200,30 +134,49 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist, IProposerChec
...
-        if (_proposer == address(0)) return false;
-        OperatorInfo memory info = operators[_proposer];
-        if (_epochTimestamp < info.activeSince) {
-            return false;
-        } else if (info.inactiveSince != 0 && _epochTimestamp >= info.inactiveSince) {
-            return false;
-        } else {
-            return true;
+        unchecked {
+            OperatorInfo storage info = operators[_proposer];
```
dantaik, line 138:
```suggestion
            OperatorInfo memory info = operators[_proposer];
```

dantaik: Since OperatorInfo uses only 1 slot, we should load it directly in to memory, then we can use "iinfo.activeSince" directly without "uint32 activeSince = info.activeSince"

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -200,30 +134,49 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist, IProposerChec
         view
         returns (bool)
     {
-        if (_proposer == address(0)) return false;
-        OperatorInfo memory info = operators[_proposer];
-        if (_epochTimestamp < info.activeSince) {
-            return false;
-        } else if (info.inactiveSince != 0 && _epochTimestamp >= info.inactiveSince) {
```
dantaik, line 207: is this original code buggy? shall 'true' be returned if `info.inactiveSince != 0 && _epochTimestamp >= info.inactiveSince` is true?

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -200,30 +134,49 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist, IProposerChec
...
-            return false;
-        } else {
-            return true;
+        unchecked {
+            OperatorInfo storage info = operators[_proposer];
+
+            uint32 activeSince = info.activeSince;
+            return activeSince != 0 && _epochTimestamp >= activeSince;
```
dantaik, line 141: the logics doesn't match the old code, but it seems the old code is buggy

dantaik (changes requested):

dantaik:
> From Claude

## Review Complete - Key Findings

I've created a comprehensive security review document at: /claudedocs/PreconfWhitelist_Security_Review.md

### Critical Issues (2):

- C-1: Operator Re-Addition Vulnerability - Operators can be re-added after removal without proper tracking, potentially causing index inconsistencies since the entire struct is deleted on removal.

- C-2: uint8 operatorCount Hard Limit - The system is limited to 255 operators maximum due to uint8 type, with no graceful handling or documentation of this limit.

### High Severity Issues (3):

- H-1: Stale latestActivationEpoch - The "fast path" optimization breaks after operator removals because latestActivationEpoch is never updated when operators are removed.

- H-2: No Genesis Validation - Contract doesn't validate that beacon genesis is configured for the current chain, could malfunction on unsupported chains.

- H-3: Gas DoS on Removals - The _hasAnotherActiveOperator function has O(n) complexity and could become prohibitively expensive with many operators.

### Medium Severity Issues (4):

- M-1: Unchecked Beacon Root - Returns 0 for randomness if beacon root unavailable, always selecting index 0
- M-2: Randomness Delay Edge Cases - Early epoch handling may allow operator selection before proper delay
- M-3: Limited Ejecter Auditing - No way to enumerate ejecters or track their actions
- M-4: Immediate Removal Asymmetry - Additions are delayed but removals are immediate, creating timing vulnerabilities

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
```diff
@@ -234,70 +187,80 @@ contract PreconfWhitelist is EssentialContract, IPreconfWhitelist, IProposerChec
...
 
         emit OperatorAdded(_proposer, _sequencer, activeSince);
     }
 
-    function _removeOperator(address _proposer, uint8 _operatorChangeDelay) internal {
+    /// @dev Removes an operator immediately and backfills its slot with the last proposer so
+    ///      operatorMapping stays packed from 0..operatorCount-1.
+    /// IMPORTANT: Reverts if no other operator is active.
+    /// @param _proposer The proposer address of the operator to remove.
+    function _removeOperator(address _proposer) internal {
```
dantaik, line 209: This immediate removal of operator maybe a problem. If Alice is active and selected as the preconfer at time X, then and time Y (Y>X), he is removed, then at time Z (Z>Y>X) Alice want to propose her preconfered transactions, her tx will revert.

dantaik:
> > If we can launch phase2 preconf in Q1, maybe we can simply live with the current whitelist as it will not be used for long?
> 
> @dantaik here's why I think it is worth doing it(besides the fact that is much simpler)
> 
> 1. We cannot use the `PreconfWhitelist` currently deployed on mainnet since it is not compatible with shasta(it does not have the `checkProposer` function), so we need to upgrade regardless.
> 2. The version in `main` already has optimizations that the currently deployed one doesn't. Please check this [PR](https://github.com/taikoxyz/taiko-mono/pull/20201/files#diff-4701131abf8c45a1eed585742b1684600922d316516cb62dd6d8e46b07fc39ca) we merged back in September, where we already trimmed ~10k gas per proposal
> 3. The previous lookahead relied on someone calling `consolidate` for being optimal(not having to scan trough the entire operator set). I don't think we are always doing that when adding/removing operators on mainnet, or it may be buggy because `havingPerfectOperators` is false every time I've queried it. This PR achieves the same gas efficiency, without having to manually call it(by using a timestamp mechanism)
> 4. Even after permisionless preconfs, we will still use this contract(it serves as the fallback preconfer when no validator has opted in), so we'll have it as part of our system for the forseable future
> 
> **So TL;DR: We need to upgrade the current implementation to make it compatible with shasta, and to save significant gas. So we might as well simplify it and audit it**

I understand point 1. But we don't need to upgrade the current preconf-whitelist, we can deploy a new one without carrying about the existing state -- we can always re-register these operators as there are only a few.

dantaik:
We could adopt an even simpler design: make the whitelist upgradable, but avoid using state variables to track operators.
Instead, hardcode them directly in the code, for example:
```
function getOperators() public pure returns (Operator[] memory operators) {
    operators = new Operator[4] ;
    operators[0] = Operator({ ... }); // hardcoded
    operators[1] = Operator({ ... }); // hardcoded
    operators[2] = Operator({ ... }); // hardcoded
    operators[3] = Operator({ ... }); // hardcoded
}
```
When we need to update the operators, we simply change the code and deploy a new implementation, then upgrade the whitelist proxy.

This approach makes reads extremely cheap since no storage lookups are required.

dantaik (dismissed):

davidtaikocha (dismissed):

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
ggonzalez94: I thought about it, but then went with direct constants because it is not something that we will change our mind before deployment(these have been the values we've had since launching preconfs)
1. RANDOMNESS_DELAY: this just ensures that we always read the beacon hash from an old enough value that it is available
2. OPERATOR_CHANGE_DELAY: ensures we don't change this and next epoch preconfers mix epoch, so 2 is the minimum you can have that works.

If we ever wanted to change these(which I kinda doubt), we'll just do a new deployment same as an immutable, but I don't think we need the flexibility to have them as ctor parameters, it would just be a bit more error prone

ggonzalez94:
> urrent preconf-whitelist, we can deploy a new one without carrying about the existing state -- we can always re-register these operators as there are only a few.

This is actually a pretty neat idea, I'll do benches tomorrow and see how much we save if we make it happen

EDIT: After second thought I'm a bit more hesitant. Our the current design, under our ejector control we keep the ability to add and remove operators from the whitelist.
If we did this, and ever wanted to add a 5th operator OR an existing one wanted to change addresses we would have to wait 21+ days for DAO voting. This might be too much

ggonzalez94:
> I understand point 1. But we don't need to upgrade the current preconf-whitelist, we can deploy a new one without carrying about the existing state -- we can always re-register these operators as there are only a few.

Agree, and it would allow us to clean up the contract even more. The only downside I see is that preconf sidecars, ejector and other off-chain software would  have to keep track of both addresses. @davidtaikocha @YoGhurt111 thoughts?

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
ggonzalez94: after speaking about it we decided to stick with the typo to avoid changing all off chain software

packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol
ggonzalez94: I don't thin operatorInfo actually uses 3 slots right? 
```
struct OperatorInfo {
        uint32 activeSince; // Epoch when the operator becomes active.
        uint32 deprecatedInactiveSince; // Deprecated. Kept for storage compatibility.
        address sequencerAddress; // Sequencer address for this operator (for off-chain use).
    }
```

I just run it and the compiler still optimizes it, but to be safe I think it is better to stick with storage

packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol
ggonzalez94: done!

dantaik (approved):

ggonzalez94:
will fix all the merge conflicts later and merge it

YoGhurt111:
> > I understand point 1. But we don't need to upgrade the current preconf-whitelist, we can deploy a new one without carrying about the existing state -- we can always re-register these operators as there are only a few.
> 
> Agree, and it would allow us to clean up the contract even more. The only downside I see is that preconf sidecars, ejector and other off-chain software would have to keep track of both addresses. @davidtaikocha @YoGhurt111 thoughts?

Both fine to me, but one small concern is that having two different contract addresses might require all parties to make changes, and there would need to be a coordination process for that.

ggonzalez94:
> > > I understand point 1. But we don't need to upgrade the current preconf-whitelist, we can deploy a new one without carrying about the existing state -- we can always re-register these operators as there are only a few.
> > 
> > 
> > Agree, and it would allow us to clean up the contract even more. The only downside I see is that preconf sidecars, ejector and other off-chain software would have to keep track of both addresses. @davidtaikocha @YoGhurt111 thoughts?
> 
> Both fine to me, but one small concern is that having two different contract addresses might require all parties to make changes, and there would need to be a coordination process for that.

Planning to keep the same address for now, so no changes should be needed on the sidecars, driver and ejector side for this

PR #20676, opened 2025-11-07
fix(taiko-client): update Shasta extraData size from 2 to 32 bytes

mikhailUshakoff:
Increase Shasta extraData size from 2 to 32 bytes to align with block.extraData size


mikhailUshakoff:
I can handle it in a sidecar logic

PR #20678, opened 2025-11-09
docs(protocol):  sets `anchorBlockNumber`, `anchorBlockHash`, `anchorStateRoot` to `0` if manifest is invalid

smtmfft:
For the normal manifest flow, the L2's anchor contract enforces that the anchor is always >= the previous anchor, so we don’t need to validate it inside prover.
However, For the exceptional case, previously the anchor must equal the parent anchor, so we have to retrieve it by tracing the parent anchor transaction (via calldata → tx root → parent header hash).
If anchor == 0 is allowed, then we can skip this entire parent-anchor verification.  Meanwhile, as Derivation doc mentioned before

| Parameter            | Type              | Description                                              |
| -------------------- | ----------------- | -------------------------------------------------------- |
| anchorBlockNumber    | uint48            | L1 block number to anchor (0 to skip anchoring)          |

Since 0 already represents skipping the anchor, I think it makes more sense to treat invalid anchor/manifest cases the same way—i.e., fall back to skip anchoring instead of enforcing parent-anchor verification.

Any suggestion?


dantaik:
@smtmfft I removed myself from the PR as I lost track of the Derivation doc updates. Count on @ggonzalez94, @davidtaikocha, and @YoGhurt111 to review.

packages/protocol/docs/Derivation.md
```diff
@@ -276,7 +276,7 @@ Timestamp validation is performed collectively across all blocks and may result
 
 Anchor block validation ensures proper L1 state synchronization and may trigger manifest replacement:
 
-**Invalidation conditions** (sets `anchorBlockNumber` to `parent.metadata.anchorBlockNumber`):
+**Invalidation conditions** (sets `anchorBlockNumber`, `anchorBlockHash`, `anchorStateRoot` to `0`):
```
ggonzalez94, line 279: if we do this how should nodes interpret it? AFAIK we were setting it to the previous blocks anchor and still keeping the block valid?
@davidtaikocha has a more complete understanding here

davidtaikocha: We can modify the nodes accordingly, if we all think that makes sense.

davidtaikocha (approved):

YoGhurt111:
One question for the case that set to 0:
If there are 2 blocks in the proposal, we set the first block's anchorBlockNumber to 0 since the invalidation conditions, how can we handle the next one?
How do we handle the rule `Non-monotonic progression**: manifest.blocks[i].anchorBlockNumber < parent.metadata.anchorBlockNumber` ? 
It means [8,0,6] is acceptable?
@ggonzalez94 @smtmfft @davidtaikocha

YoGhurt111:
> One question for the case that set to 0: If there are 2 blocks in the proposal, we set the first block's anchorBlockNumber to 0 since the invalidation conditions, how can we handle the next one? How do we handle the rule `Non-monotonic progression**: manifest.blocks[i].anchorBlockNumber < parent.metadata.anchorBlockNumber` ? It means [8,0,6] is acceptable? @ggonzalez94 @smtmfft @davidtaikocha

Or we use the ancestor's anchor blocknum?

davidtaikocha:
> One question for the case that set to 0:
If there are 2 blocks in the proposal, we set the first block's anchorBlockNumber to 0 since the invalidation conditions, how can we handle the next one?
How do we handle the rule Non-monotonic progression**: manifest.blocks[i].anchorBlockNumber < parent.metadata.anchorBlockNumber ?
It means [8,0,6] is acceptable?
@ggonzalez94 @smtmfft @davidtaikocha

I think This is an issue for the current derivation process if we change to this @smtmfft

smtmfft:
> > One question for the case that set to 0: If there are 2 blocks in the proposal, we set the first block's anchorBlockNumber to 0 since the invalidation conditions, how can we handle the next one? How do we handle the rule `Non-monotonic progression**: manifest.blocks[i].anchorBlockNumber < parent.metadata.anchorBlockNumber` ? It means [8,0,6] is acceptable? @ggonzalez94 @smtmfft @davidtaikocha
> 
> Or we use the ancestor's anchor blocknum?

[8, 0, 6] are the input calldata of anchor, based on my understanding, current this monotonic is checked inside the anchor with a stored state, so, the only problem is how to get the current state in normal/correct cases, for client, it seems easy as it reads the L2 directly.

PR #20685, opened 2025-11-10
feat(taiko-client): introduce `--shasta.useLocalDecoder` flag

davidtaikocha:
Introduce an optional `--shasta.useLocalDecoder` flag, to use the Golang version of [`OptimizedCodec.sol`](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/core/impl/CodecOptimized.sol) in protocol, to save one RPC call each time there is a need for decoding the event payloads. Should be helpful when initializing the Shasta event indexer (lots of event payloads decoding are required).


packages/taiko-client/pkg/chain_iterator/event_iterator/batch_proposed_iterator.go
```diff
@@ -173,16 +173,14 @@ func assembleBatchProposedIteratorCallback(
...
 
-			proposedEventPayload, err := rpcClient.DecodeProposedEventPayload(&bind.CallOpts{Context: ctx}, event.Data)
+			payload, err := rpcClient.DecodeProposedEventPayload(&bind.CallOpts{Context: ctx}, event.Data)
 			if err != nil {
 				log.Error("Failed to decode proposed event data", "error", err)
 				return err
 			}
-			if proposedEventPayload == nil {
-				return errors.New("decoded proposed event payload is nil")
-			}
```
MatusKysel, line 183: shouldn't we keep this check?

davidtaikocha: Added back: https://github.com/taikoxyz/taiko-mono/pull/20685/commits/1bf722f6601a87fda3c1c76deb37fca6bdc3be05

MatusKysel (approved):
lgtm

packages/taiko-client/pkg/chain_iterator/event_iterator/event_decoder/shasta.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if bondInstructionsLen > 0 {
+		payload.BondInstructions = make([]shastaBindings.LibBondsBondInstruction, bondInstructionsLen)
+		for i := range payload.BondInstructions {
+			if payload.BondInstructions[i].ProposalId, err = r.readBigUint48(); err != nil {
+				return nil, err
+			}
+			if payload.BondInstructions[i].BondType, err = r.readUint8(); err != nil {
```
MatusKysel, line 131: shouldn't we do same check as below ?
 ```		
var bondType uint8
			if bondType, err = r.readUint8(); err != nil {
				return nil, err
			}
			if bondType > maxBondTypeValue {
				return nil, ErrInvalidBondType
			}
```

File: packages/taiko-client/pkg/chain_iterator/event_iterator/event_decoder/shasta.go [hunk 3]
```diff
@@ -0,0 +1,336 @@
...
+				return nil, err
+			}
```
MatusKysel, line 133:
```suggestion
           bt, err := r.readUint8()
           if err != nil { return nil, err }
           if bt > maxBondTypeValue { return nil, ErrInvalidBondType }
           payload.BondInstructions[i].BondType = bt
```

packages/taiko-client/pkg/chain_iterator/event_iterator/event_decoder/shasta.go
davidtaikocha: It's translated from the [`OptimizedCodec.sol`](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/core/impl/CodecOptimized.sol), and in `LibProvedEventEncoder`, it does have checks like: https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/core/libs/LibProvedEventEncoder.sol#L107 , but in `LibProposedEventEncoder`, it also doesn't check it: https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/core/libs/LibProposedEventEncoder.sol#L151-L161 . So its ok to leave here as it is?

YoGhurt111 (approved):

packages/taiko-client/pkg/chain_iterator/event_iterator/event_decoder/shasta.go
MatusKysel: @ggonzalez94 maybe we have a bug in sol? shouldn't we do same check in `ProposedEvent` as we do in `ProvedEvent` on `bondType` ?

PR #20688, opened 2025-11-10
fix: seed proposer jitter PRNG to avoid synchronized timers

prestoalvarez:
## Summary
- seed the proposer’s non-cryptographic PRNG during configuration to keep jitter random across restarts


davidtaikocha:
Closed since lacking `rand.Seed` is acceptable in proposer's use case.

PR #20690, opened 2025-11-10
feat(taiko-client): check valid proof before submission

YoGhurt111:


packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
```diff
@@ -56,12 +58,22 @@ func NewProofSubmitterShasta(
...
 	proofPollingInterval time.Duration,
 	proofBuffers map[proofProducer.ProofType]*proofProducer.ProofBuffer,
 	forceBatchProvingInterval time.Duration,
 ) (*ProofSubmitterShasta, error) {
+	anchorValidator, err := validator.New(
+		taikoAnchorAddress,
+		senderOpts.RPCClient.L2.ChainID,
+		senderOpts.RPCClient,
+	)
+
```
davidtaikocha:
```suggestion
```

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
```diff
@@ -426,3 +450,125 @@ func (s *ProofSubmitterShasta) WaitParentShastaTransitionHash(
...
+		}
+		proposalID := proof.BatchID
+		if !ok {
+			log.Error("A valid proof for this batch has already been submitted", "proposalID", proposalID)
+			invalidProposalIDs = append(invalidProposalIDs, proposalID.Uint64())
+			continue
+		}
+
+		// Validate each block in each proposal.
+		for _, blockHeader := range proof.Opts.ShastaOptions().Headers {
```
davidtaikocha: Maybe we only need to verify the last header in `proof.Opts.ShastaOptions().Headers` by `s.rpc.L2.BlockByHash`? For `block.Transactions().Len()` and `s.anchorValidator.ValidateAnchorTx` actually we can do nothing if that happened?

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
```diff
@@ -289,15 +302,26 @@ func (s *ProofSubmitterShasta) BatchSubmitProofs(ctx context.Context, batchProof
...
-	)
 	proofBuffer, exist := s.proofBuffers[batchProof.ProofType]
 	if !exist {
 		return fmt.Errorf("unexpected proof type from raiko to submit: %s", batchProof.ProofType)
 	}
+
+	// Check if there is any invalid batch proofs in the aggregation, if so, we ignore them.
+	invalidProposalIDs, err := s.validateBatchProofs(ctx, batchProof)
+	if err != nil {
+		return fmt.Errorf("failed to validate batch proofs: %w", err)
```
MatusKysel, line 300: shouldn't we still clear whole buffer here ?

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
YoGhurt111: I'm considering removing this check because it's useless and inefficient.

davidtaikocha: I think its ok only check if the last block's hash matches the one in local chain, which is enough imo?

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
YoGhurt111: We only need to clear the invalids.

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
YoGhurt111: Actually the content of Headers for Shasta is never used.

davidtaikocha: emm we are not using it to verify if the proofs indexed is valid?

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
MatusKysel: but if you return on error, will it be access again/later ? I don't know whole flow, so that's why I am asking. And it's also cleared on valid proofs

YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/20690/commits/649af88d657d102f547c7a74578e1db806a48389 .
The workflow is if we have invalid proofs detected, we remove these invalid proofs and skip this retry.
We'll try again once the buffer is full again.

davidtaikocha (approved):

MatusKysel (approved):

PR #20693, opened 2025-11-10
fix(taiko-client,taiko-client-rs): remove the `finalizationGracePeriod` check in `getTransitionsForFinalization`

davidtaikocha:
In the current protocol impl: https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/core/impl/Inbox.sol#L977-L985 , we can finalize a proposal as soon as a transition has been submitted, no need to care about `finalizationGracePeriod` anymore.


MatusKysel (approved):
lgtm

YoGhurt111 (approved):

PR #20695, opened 2025-11-10
fix(taiko-client): fix LRU Keys iteration in `OnUnsafeL2Request`

Fibonacci747:
The end-of-sequencing detection loop used range over Keys() incorrectly:
it iterated indices instead of the key values. As a result, the code passed an index (0..n-1) into Get(uint64(epoch)), which rarely matched real epoch keys and caused endOfSequencing to remain false even when the epoch->hash entry existed. This change switches the loop to iterate over values: for , epoch : range cache.Keys() { Get(epoch) }, restoring correct EoS detection behavior.


davidtaikocha:
@claude plz review.

davidtaikocha (approved):

YoGhurt111 (approved):

PR #20704, opened 2025-11-11
feat(taiko-client): remove duplicate buffer-size stop condition

Bashmunta:
This change deletes a redundant check in fetchHistoricalProposals that duplicated the existing buffer-size stop condition. The function already stops when Count() >= bufferSize (passed as s.bufferSize by Start), so the additional check using !historicalFetchCompleted && Count() >= s.bufferSize was functionally equivalent during the initial historical fetch and never provided additional safety.


davidtaikocha:
@claude plz review.

MatusKysel (approved):

davidtaikocha (approved):

YoGhurt111 (approved):

PR #20706, opened 2025-11-11
feat(protocol): add prevAnchorBlockNumber in anchor event

smtmfft:
Just came up with a simpler solution rather than https://github.com/taikoxyz/taiko-mono/pull/20678, which leads to changes in both protocol (derivation doc) & client. 
Now, if anchor emits the prev number, it's easy for prover to check if there is anything wrong. so the validation in prover is like:
if anchor_block_number is not correct, then prevBlockNumber must == _blockState.anchorBlockNumber(i.e., currBlockNumber), which has no historical access in prover side. And even better, no more changes in both client & derivation document.

### DIFF
--- a/packages/protocol/contracts/layer2/core/Anchor.sol
+++ b/packages/protocol/contracts/layer2/core/Anchor.sol
@@ -145,6 +145,7 @@ contract Anchor is EssentialContract {
         bytes32 bondInstructionsHash,
         address designatedProver,
         bool isLowBondProposal,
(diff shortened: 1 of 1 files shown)


davidtaikocha (dismissed):
changes lgtm

ggonzalez94 (dismissed):
LGTM. @smtmfft should we target the audit branch and then cherry pick to main? since we want this to be part of what we will deploy on hoodi and then mainnet

PR #20709, opened 2025-11-11
fix(protocol): prover auth using eip-712

ggonzalez94:
I think we topic we forgot to discuss is if we need EIP-712 or similar for the proverAuth in the anchor to both prevent replay attacks across chains(i.e. testnet and mainnet) and to make this more friendly for provers signing this message via a wallet. Since signatures are scoped to a specific `proposalId` and have a time they expire this is likely not an issue, but perhaps still worth fixing?

### Description

`Anchor.validateProverAuth` accepts arbitrary signer signatures and hashes the payload with `EfficientHashLib.hash(proposalId, proposer, provingFee)` (`packages/protocol/contracts/layer2/core/Anchor.sol:322-357`, `_hashProverAuthMessage` at :511-516). The message omits a domain separator—there is no chain ID, contract address, fork identifier, nor proposal hash. A signature created for `(proposalId, proposer, provingFee)` on any Taiko deployment (e.g., testnet/devnet) is therefore valid on every other deployment once `proposalId` matches. An attacker can capture a prover’s signature elsewhere and replay it on mainnet to (i) force that prover to become the designated prover for a proposal they never agreed to, and (ii) siphon the proving fee via the bond transfer that `anchorV4` performs when the signature is accepted.

### Reproduction Steps

1. Obtain a legitimate designated-prover signature for some `(proposalId = N, proposer = X, provingFee = Y)` on any network (the signer need not interact with the target chain).
2. Wait until mainnet proposal `N` is created for proposer `X`.
3. Submit an `anchorV4` transaction (from the golden touch address) that sets `proposalParams.proverAuth =` the captured bytes.
4. `validateProverAuth` recovers the signer and assigns them as the designated prover on mainnet, even though they never signed for this chain, proving the replay.

### Impact

Replay lets adversaries impersonate provers across forks/chains, causing unauthorized bond debits/credits and forcing honest provers to handle proposals they never opted into. Because bond debits happen immediately in `anchorV4`, replaying signatures can directly steal a designated prover’s funds.

### Remediation

Hash prover authorizations over a domain-separated message. An EIP‑712 struct that embeds `(chainId, address(this), forkId, proposalHash)` would prevent cross-chain reuse. Existing signatures should be considered invalid once domain separation ships; clients must regenerate signer payloads.

---

> [!NOTE]
> Switches prover authorization to EIP-712 typed-data with domain separation and updates tests to cover validity and replay protection cases.
> 
> - **Contracts (Anchor)**:
>   - Use EIP-712 for `ProverAuth` hashing: add domain/type hashes, domain separator, and struct hash helpers; replace raw hashing with `ECDSA.toTypedDataHash` in `_hashProverAuthMessage`.
>   - Change `validateProverAuth` from `pure` to `view` (uses chainId/contract in domain separator).
>   - Remove `EfficientHashLib` usage related to prover auth.
> - **Tests (Anchor.t.sol)**:
>   - Add EIP-712 signing utilities mirroring contract domain/type hashes.
>   - Update flows to sign/verify EIP-712 `ProverAuth`.
>   - New cases: accept valid EIP-712 signature; reject invalid/empty signature; reject signatures for wrong chain, proposal ID, or proposer; ensure `anchorV4` ignores invalid auth and fee not transferred.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4d2b8864a4bfff6e94ed9a52def9476d3539a302. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/protocol/contracts/layer2/core/Anchor.sol
+++ b/packages/protocol/contracts/layer2/core/Anchor.sol
@@ -5,7 +5,6 @@ import { IBondManager } from "./IBondManager.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { ECDSA } from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
-import { EfficientHashLib } from "solady/src/utils/EfficientHashLib.sol";
 import { EssentialContract } from "src/shared/common/EssentialContract.sol";
 import { LibAddress } from "src/shared/libs/LibAddress.sol";
 import { LibBonds } from "src/shared/libs/LibBonds.sol";
@@ -95,6 +94,14 @@ contract Anchor is EssentialContract {
 
     /// @dev Length of a standard ECDSA signature (r: 32 bytes, s: 32 bytes, v: 1 byte).
     uint256 private constant ECDSA_SIGNATURE_LENGTH = 65;
+
+    /// @dev EIP-712 domain/type hashes for prover authorization signatures.
+    bytes32 private constant PROVER_AUTH_DOMAIN_TYPEHASH =
+        keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
+    bytes32 private constant PROVER_AUTH_TYPEHASH =
+        keccak256("ProverAuth(uint48 proposalId,address proposer,uint256 provingFee)");
+    bytes32 private constant PROVER_AUTH_DOMAIN_NAME_HASH = keccak256("TaikoAnchorProverAuth");
+    bytes32 private constant PROVER_AUTH_DOMAIN_VERSION_HASH = keccak256("1");
 
     // ---------------------------------------------------------------
     // Immutables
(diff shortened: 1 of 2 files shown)


dantaik (approved):

dantaik:
@claude please review

davidtaikocha (approved):

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -96,6 +95,15 @@ contract Anchor is EssentialContract {
     /// @dev Length of a standard ECDSA signature (r: 32 bytes, s: 32 bytes, v: 1 byte).
     uint256 private constant ECDSA_SIGNATURE_LENGTH = 65;
 
+    /// @dev EIP-712 domain/type hashes for prover authorization signatures.
+    bytes32 private constant PROVER_AUTH_DOMAIN_TYPEHASH = keccak256(
```
nahimdhaney, line 99: I would use `EIP712_DOMAIN_TYPEHASH` instead

nahimdhaney (commented):
the implementation looks correct, 

but you should also expose the function DOMAIN_SEPARATOR()

so clients/consumer know what the actual domain separator is.

PR #20712, opened 2025-11-12
fix(taiko-client): skip responding on empty L1Origin signature and fix misleading return

Bashmunta:
This change fixes a misleading return path in OnUnsafeL2Request where an empty L1Origin signature triggered return err even though err was nil at that point. It now explicitly returns nil, making the intent (drop response without handler error) clear and robust.
Additionally, OnUnsafeL2EndOfSequencingRequest now mirrors this behavior by checking for an empty L1Origin signature and returning nil early. This ensures consistent response handling and avoids attempting to publish responses without a valid signature.
These changes align handler semantics with P2P expectations: missing local prerequisites should not surface as protocol errors, but responses should be withheld.


MatusKysel:
@claude plz review

davidtaikocha:
@claude plz review.

MatusKysel (approved):

davidtaikocha (approved):

PR #20714, opened 2025-11-12
test(taiko-client,protocol): introduce tests with bond instructions && improve `DeployProtocolOnL1` deployment script

davidtaikocha:


packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -900,29 +900,31 @@ func (c *Client) checkSyncedL1SnippetFromAnchor(
...
+		}
+		if lastPacayaBlockID.Cmp(common.Big0) == 0 {
+			return &rawdb.L1Origin{BlockID: common.Big0}, nil
 		}
 
+		l1Origin, err := c.L2.L1OriginByID(ctxWithTimeout, lastPacayaBlockID)
+		if err != nil {
+			return nil, fmt.Errorf("L1Origin not found for last block in batch ID %d: %w", batchID, err)
+		}
 		return l1Origin, nil
```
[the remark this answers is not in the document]
davidtaikocha: Fixed in https://github.com/taikoxyz/taiko-mono/pull/20714/commits/aba09c98cca4e643809f1b5de115eaeceee7f3ee

davidtaikocha:
@claude plz review.

packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -900,29 +900,31 @@ func (c *Client) checkSyncedL1SnippetFromAnchor(
...
-			return nil, fmt.Errorf("L1Origin not found for batch ID %d: %w", batchID, err)
+			return nil, fmt.Errorf("failed to fetch last Pacaya block ID: %w", err)
+		}
+		if lastPacayaBlockID.Cmp(common.Big0) == 0 {
+			return &rawdb.L1Origin{BlockID: common.Big0}, nil
 		}
 
+		l1Origin, err := c.L2.L1OriginByID(ctxWithTimeout, lastPacayaBlockID)
+		if err != nil {
+			return nil, fmt.Errorf("L1Origin not found for last block in batch ID %d: %w", batchID, err)
```
YoGhurt111:
```suggestion
			return nil, fmt.Errorf("L1Origin not found for last Pacaya block ID %d: %w", lastPacayaBlockID, err)
```
We can always know the batchID here is 0.

packages/taiko-client/prover/prover_test.go
```diff
@@ -200,6 +202,67 @@ func (s *ProverTestSuite) TestOnBatchProposed() {
...
+	state, err = s.RPCClient.ShastaClients.Anchor.GetProposalState(nil)
+	s.Nil(err)
+	s.NotZero(state.BondInstructionsHash)
+
+	l2Head, err := s.RPCClient.L2.BlockByNumber(context.Background(), nil)
+	s.Nil(err)
+	s.Greater(l2Head.Transactions().Len(), 0)
+	receipt, err := s.RPCClient.L2.TransactionReceipt(context.Background(), l2Head.Transactions()[0].Hash())
+	s.Nil(err)
+	s.Equal(types.ReceiptStatusSuccessful, receipt.Status)
```
YoGhurt111: IMO, this check will always be true.
Because it's Anchor tx.
And we don't mint bond token in the genesis, it will be concidered as LowBondProposal, we can't check the balance in bondManager?

davidtaikocha: Actually here I just want to ensure that, no `BondInstructionsHashMismatch` error in `anchorV4`. Since we have:
```Go
state, err = s.RPCClient.ShastaClients.Anchor.GetProposalState(nil)
s.Nil(err)
s.NotZero(state.BondInstructionsHash)
```
And bond transfer is not part of this test, here we just need to test that driver will insert the correct bond instructions after `BondProcessingDelay`.

YoGhurt111: Got it

YoGhurt111 (approved):

smtmfft (approved):

PR #20724, opened 2025-11-12
fix(balance-monitor): use context cancellation for RPC calls in balance monitor

hawkadrian:
- Replace context.Background() with b.ctx in Start() method to ensure RPC calls respect context cancellation during graceful shutdown. 

- Added timeout to prevent hanging operations when network issues occur.


davidtaikocha (approved):

davidtaikocha:
@claude plz review.

cyberhorsey (approved):

PR #20725, opened 2025-11-13
fix(taiko-client): fix repeated event process

YoGhurt111:


packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -372,6 +372,11 @@ func (c *Client) WaitL2Header(ctx context.Context, blockID *big.Int) (*types.Hea
 	return waitHeader(ctx, c.L2, blockID)
 }
 
+// WaitL2Block keeps waiting for the L2 block of the given block ID.
+func (c *Client) WaitL2Block(ctx context.Context, blockID *big.Int) (*types.Block, error) {
+	return waitBlock(ctx, c.L2, blockID)
+}
+
 // waitHeader keeps waiting for the block header of the given block ID.
 func waitHeader(ctx context.Context, ethClient *EthClient, blockID *big.Int) (*types.Header, error) {
```
davidtaikocha: Maybe we can also reuse `waitBlock(ctx, c.L2, blockID)` here? and only return `block.header`?

packages/taiko-client/driver/chain_syncer/event/syncer.go
```diff
@@ -440,11 +440,18 @@ func (s *Syncer) processShastaProposal(
 		); err != nil {
 			return fmt.Errorf("failed to insert Shasta blocks: %w", err)
 		}
-		if parent, err = s.rpc.L2.BlockByNumber(ctx, new(big.Int).Add(parent.Number(), common.Big1)); err != nil {
-			return fmt.Errorf("failed to fetch the new parent block: %w", err)
+		if parent, err = s.rpc.WaitL2Block(ctx, new(big.Int).Add(parent.Number(), common.Big1)); err != nil {
+			log.Warn("Failed to fetch the new parent block, keep retrying", "error", err)
```
davidtaikocha: We are waiting for `new(big.Int).Add(parent.Number(), common.Big1)`, not `parent block` in this log, right?

packages/taiko-client/pkg/rpc/methods.go
MatusKysel: this is much cheaper call, I dunno how we use it, but maybe right way would be to write wait function that accept rpc function to call as a param

YoGhurt111: > Maybe we can also reuse waitBlock(ctx, c.L2, blockID) here? and only return block.header?

Sure, will do

packages/taiko-client/driver/chain_syncer/event/syncer.go
YoGhurt111: We use the current block as the `parent` of next block manifest, so we call it as `new parent block` in the log

packages/taiko-client/pkg/rpc/methods.go
YoGhurt111: > this is much cheaper call, I dunno how we use it, but maybe right way would be to write wait function that accept rpc function to call as a param

Actually, we only use the `waitBlock` now, so I think it is not necessary to abstract it into a way of passing in an RPC function at this stage. WDYT?

davidtaikocha (approved):

packages/taiko-client/pkg/rpc/methods.go
MatusKysel: but why ? block is several X heavier RPC call then header, why do we want to use it ? doesn't make sense

YoGhurt111: You are right, will refactor it.

YoGhurt111: @MatusKysel Plz take another look. [d0069e2](https://github.com/taikoxyz/taiko-mono/pull/20725/commits/d0069e27c0853d67f628d0ad8acec64d7851fd1b)

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main,fix/repeated_event_processed]
```
```suggestion
    branches: [main]
```

MatusKysel (approved):

packages/taiko-client/pkg/rpc/methods.go
MatusKysel: looks good

PR #20727, opened 2025-11-13
fix(taiko-client): remove unreachable L2-specific error from waitHeader

Forostovec:
The final error return in waitHeader was unreachable due to the infinite polling loop, yet contained an L2 execution engine-specific message. This could mislead readers and future maintainers, and it could become visible if the loop structure changed. Removing it eliminates dead code and prevents confusion without altering runtime behavior, since waitHeader already returns the header on success or the context error (deadline/cancel) on failure.


davidtaikocha:
It's ok to leave it there as long as there is no error in Golang lint :)

PR #20730, opened 2025-11-13
feat(protocol): relax TIMESTAMP_MAX_OFFSET to allow next proposer to recover

ggonzalez94:
This was raised by @mikhailUshakoff during our preconf sync today. In our current derivation doc we set the `TIMESTAMP_MAX_OFFSET` to 384 seconds(1 epoch). This works well for when the proposer submits during their epoch. But it has 2 cases where this is too restrictive and problematic:
1. the current proposer fails tu submit during their window. The next one can recover that proposal and avoid a reorg, but if the timestamp of the blocks is too old, this will not be possible.
2. the current proposer has 2 consecutive epochs where he is the proposer. He may decide to batch 2 epochs worth of blocks before proposing. With the current restriction this would not be possible.

This PR aligns `TIMESTAMP_MAX_OFFSET` with `MAX_ANCHOR_OFFSET`, so that both can be 128 blocks in the past, which should allow both of the two issues above to be resolved.

Aditionally we clarify that `lastProposalBlockId` is an L1 block in the inbox corestate(which was causing some confusion).

---

> [!NOTE]
> Raises TIMESTAMP_MAX_OFFSET to 1536 (12*128) in derivation docs and clarifies that CoreState.lastProposalBlockId refers to an L1 block ID.
> 
> - **Docs** (`packages/protocol/docs/Derivation.md`):
>   - Increase `TIMESTAMP_MAX_OFFSET` from `384` to `1536` (12 * 128) with rationale to enable next proposer recovery without reorgs.
> - **Contracts** (`packages/protocol/contracts/layer1/core/iface/IInbox.sol`):
>   - Clarify comment for `CoreState.lastProposalBlockId` to specify it is an L1 block ID.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3b957fbae92d4dd77c7b6b93e17194cba2fe5ed7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mikhailUshakoff (approved):
LGTM

dantaik (approved):

davidtaikocha (approved):

PR #20732, opened 2025-11-13
refactor(taiko-client): make StatusRegistered a const

sashass1315:
Converted StatusRegistered from a package var to a const to reflect its immutable protocol token nature and avoid accidental mutation. This change is safe because StatusRegistered is only used in string comparisons (not assigned or address-taken) and is referenced by RaikoRequestProofBodyResponseV2.Validate in common.go to map the “registered” status to ErrRetry. Error sentinels remain vars since errors.New is not a constant expression. The update clarifies intent, slightly reduces runtime mutability, and aligns with idiomatic Go usage for fixed status strings without impacting behavior.


MatusKysel:
this is not improving anything

PR #20740, opened 2025-11-14
fix(relayer): correct ChainDataSynced log text to reference signal

MozirDmitriy:
Fix misleading logging in handle_chain_data_synced_event.go where the message said “for msgHash” but the logged value was the event’s signal. The ChainDataSynced event does not have a msgHash field (per ABI); it exposes Signal [32]byte. We now say “for signal” while keeping the "signal" key and value unchanged.


davidtaikocha:
Closed since can't see too much valuable improvements there.

PR #20741, opened 2025-11-14
fix: split L1/L2 ERC-20 balance metrics

Bilogweb3:
Keep L1 and L2 ERC-20 balances separate so each Prometheus gauge reflects the correct network, and add the missing l2_erc20_balance metric to expose L2 data that was previously invisible.


davidtaikocha:
Closed since we are not tracking L2 token balance.

PR #20744, opened 2025-11-15
fix(eventindexer): include genesis block by adjusting startingBlock to GenesisHeight-1

Snezhkko:
The event indexer was skipping the genesis block on L1 because the initial cursor was set to GenesisHeight while the filter loop starts from latestIndexedBlockNumber+1. As a result, the first processed block became GenesisHeight+1, dropping genesis events. This change aligns the event indexer with the relayer’s initialization pattern and the taiko-client’s usage, subtracting one from GenesisHeight (safely clamped to zero) when deriving the initial startingBlock from L1 contract state across v1, v2, and v3. With this fix, the first batch processed begins exactly at GenesisHeight, ensuring genesis events are indexed.


davidtaikocha:
We don't need to process events in genesis block.

PR #20745, opened 2025-11-16
fix(taiko-client,taiko-client-rs): ensure Shasta low-bond proposals set `block.extradata` bit even for default payloads

davidtaikocha:
- Always query `getDesignatedProver` in the Go syncer and propagate the `IsLowBondProposal` flag even when the manifest fetcher already returned a default payload or a forced-inclusion segment. Only swap the manifest for the synthetic default when a non-default payload turns out to be low-bond.
  - Mirror the same behavior in the Rust driver pipeline so every segment calls the low-bond detector, ensuring the extra-data encoder sets the bit for default/forced manifests as well.

To match the description in https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/docs/Derivation.md


MatusKysel (approved):

davidtaikocha:
@claude plz review.

YoGhurt111 (approved):

PR #20746, opened 2025-11-16
fix(taiko-client,taiko-client-rs): update gas-limit clamp

davidtaikocha:
Switch both clients to clamp gas limits using the correct denominator from `manifest.GasLimitChangePermyriad`, expose `GasLimitChangeDenominator` in the bindings/protocol constants, and update the Go validator/tests plus the Rust validator to share that constant.

To match the description in https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/docs/Derivation.md

### DIFF
--- a/packages/taiko-client/driver/chain_syncer/event/manifest/shasta.go
+++ b/packages/taiko-client/driver/chain_syncer/event/manifest/shasta.go
@@ -464,11 +464,13 @@ func validateGasLimit(
 	}
 	for i := range sourcePayload.BlockPayloads {
 		lowerGasBound := max(
-			parentGasLimit*(10000-manifest.MaxBlockGasLimitChangePermyriad)/10000,
+			parentGasLimit*(manifest.GasLimitChangeDenominator-manifest.MaxBlockGasLimitChangePermyriad)/
+				manifest.GasLimitChangeDenominator,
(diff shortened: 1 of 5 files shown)


MatusKysel (approved):

packages/taiko-client/driver/chain_syncer/event/manifest/shasta.go
```diff
@@ -464,11 +464,13 @@ func validateGasLimit(
 	}
 	for i := range sourcePayload.BlockPayloads {
 		lowerGasBound := max(
-			parentGasLimit*(10000-manifest.MaxBlockGasLimitChangePermyriad)/10000,
+			parentGasLimit*(manifest.GasLimitChangeDenominator-manifest.MaxBlockGasLimitChangePermyriad)/
```
YoGhurt111, line 467: Maybe we need to double check this value with protocol side, 0.1% makes more sense to me.
0.001% means 10,000,000 gas limit only can increase 100 gas limit in next block, isn't this too small?

davidtaikocha: image
Yep we can, and it's `0.001%` in protocol for now.

YoGhurt111 (approved):

PR #20748, opened 2025-11-17
fix(taiko-client,taiko-client-rs): only apply default low-bond manifests for normal proposer sources

davidtaikocha:
To match the description in [Derivation Process](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/docs/Derivation.md):
>
**Default Manifest Replacement**: When `isLowBondProposal = true`, the **only the proposer's manifest** is replaced with the default manifest, minimizing proving costs and disincentivizing spam. Forced inclusion manifests are still processed.

### DIFF
--- a/packages/taiko-client/driver/chain_syncer/event/syncer.go
+++ b/packages/taiko-client/driver/chain_syncer/event/syncer.go
@@ -766,6 +766,7 @@ func applyLowBondProposalRules(
 		Default:           true,
 		IsLowBondProposal: true,
 		ParentBlock:       payload.ParentBlock,
+		ProverAuthBytes:   payload.ProverAuthBytes,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


davidtaikocha:
@claude plz review.

packages/taiko-client/driver/chain_syncer/event/syncer.go
```diff
@@ -748,6 +742,34 @@ func (s *Syncer) checkReorgShasta(
...
+	isProposerSource bool,
+	isLowBondProposal bool,
+) *shastaManifest.ShastaDerivationSourcePayload {
+	// If not a low-bond proposal, return directly.
+	if payload == nil || !isLowBondProposal {
+		return payload
+	}
+
+	// For low-bond proposals, we always enforce the default manifest rules.
+	// If the source is a forced inclusion source, we keep the original payload but
```
YoGhurt111, line 757: Why is the force inclusion always low bond proposal?

davidtaikocha: We have:
```Go
	// If not a low-bond proposal, return directly.
	if payload == nil || !isLowBondProposal {
		return payload
	}
```
right above.

YoGhurt111 (approved):

johntaiko (approved):

PR #20749, opened 2025-11-17
chore(protocol): use custom errors in SgxVerifier constructor

IrvinGa38:
SgxVerifier already defines custom errors (SGX_* at lines 70-73) but its constructor still uses string errors. Replaced constructor string validation with INVALID_CHAIN_ID() and INVALID_AUTOMATA_DCAP_ATTESTATION() to match the contract's existing error pattern and save gas (4 bytes vs ~20+ bytes per error).


davidtaikocha:
Could u move all these kinds of similar updates in https://github.com/taikoxyz/taiko-mono/pull/20742 ? @IrvinGa38

IrvinGa38:
will do. @davidtaikocha

PR #20753, opened 2025-11-17
fix(taiko-client): respect derivation origin block for anchor checks

davidtaikocha:
- Anchor-window validation was using the raw L1 inclusion height (`meta.GetRawBlockHeight`) but the protocol defines `originBlockNumber = l1 height - 1`, so the check was off by one.
  - Pass `meta.GetDerivation().OriginBlockNumber` into `ValidateMetadata` so anchor bounds match the Inbox contract/spec.

To match the [Derivation process](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/docs/Derivation.md):
> 
**Invalidation conditions** (sets `anchorBlockNumber` to `parent.metadata.anchorBlockNumber`):
- **Non-monotonic progression**: `manifest.blocks[i].anchorBlockNumber < parent.metadata.anchorBlockNumber`
- **Future reference**: `manifest.blocks[i].anchorBlockNumber >= proposal.originBlockNumber - MIN_ANCHOR_OFFSET`
- **Excessive lag**: `manifest.blocks[i].anchorBlockNumber < proposal.originBlockNumber - MAX_ANCHOR_OFFSET`

### DIFF
--- a/packages/taiko-client/driver/chain_syncer/event/blocks_inserter/common.go
+++ b/packages/taiko-client/driver/chain_syncer/event/blocks_inserter/common.go
@@ -222,10 +222,11 @@ func isKnownCanonicalBatchPacaya(
 
 	// Check each block in the batch, and if the all blocks are preconfirmed, return the header of the last block.
 	for i := 0; i < len(metadata.Pacaya().GetBlocks()); i++ {
+		idx := i
 		g.Go(func() error {
-			parentHeader, err := rpc.L2.HeaderByNumber(ctx, new(big.Int).SetUint64(parent.Number.Uint64()+uint64(i)))
+			parentHeader, err := rpc.L2.HeaderByNumber(ctx, new(big.Int).SetUint64(parent.Number.Uint64()+uint64(idx)))
 			if err != nil {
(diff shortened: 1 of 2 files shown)


YoGhurt111 (approved):

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/common.go
```diff
@@ -222,10 +222,11 @@ func isKnownCanonicalBatchPacaya(
 
 	// Check each block in the batch, and if the all blocks are preconfirmed, return the header of the last block.
 	for i := 0; i < len(metadata.Pacaya().GetBlocks()); i++ {
+		idx := i
```
MatusKysel: I don't think we need to do this with go 1.22+ right ?

MatusKysel (approved):

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/common.go
davidtaikocha: reverted in https://github.com/taikoxyz/taiko-mono/pull/20753/commits/767a23f1b0127e8bba74c74bbcabe556af924fbd

MatusKysel (approved):

PR #20755, opened 2025-11-17
fix(taiko-client): eliminate errgroup data races by using local error variables

ANtutov:
This change removes data races caused by concurrent writes to a captured outer err variable inside multiple errgroup goroutines. In GetProtocolStateVariablesPacaya and GetForcedInclusionPacaya within packages/taiko-client/pkg/rpc/methods.go, and in TxBuilderWithFallback.BuildPacaya within packages/taiko-client/proposer/transaction_builder/fallback.go, separate goroutines previously wrote to the same err value, which is racy under the Go memory model and reproducible with the race detector. The fix adopts a standard errgroup pattern: each goroutine uses a local error variable and returns it, while results are staged in local temporaries and only assigned to shared outer variables after successful completion or after g.Wait confirms no error. This preserves behavior, avoids synchronization overhead, and ensures correctness under -race without altering external APIs or semantics.


davidtaikocha:
It's ok to return any error if there is.

PR #20765, opened 2025-11-18
fix(taiko-client): guard proof buffer state with locks

Bilogweb3:
FirstItemAt, IsAggregating, and MarkAggregating were accessing shared state without locks while Write/ClearItems hold the mutex. This causes data races when proofs are submitted concurrently. Added the missing RLock/Lock calls.


davidtaikocha:
@claude plz review.

MatusKysel (approved):

YoGhurt111 (approved):

PR #20766, opened 2025-11-18
fix(taiko-client): bound inbox event retries, add per-attempt logging

Forostovec:
This change addresses a failure mode where the indexer could stall indefinitely on a single bad inbox event due to an unbounded retry that ignored its final result and lacked per-attempt logging. We now add per-attempt warning logs inside the retry closure for both Proposed and Proved handlers, cap the exponential backoff retries to a finite number to prevent livelock, and on exhaustion, log a terminal error and increment a dropped-events counter before moving on to the next log. Additional metrics are introduced to track proposed/proved handling errors and dropped events, improving observability and enabling alerting. This aligns the indexer’s behavior with the driver’s logging pattern and ensures the system remains forward-progressing and diagnosable under persistent failures.


davidtaikocha:
We can't skip any log there, it's by design.

PR #20767, opened 2025-11-18
feat(protocol): add `SHASTA_FORK_TIME` to Shasta block `timestamp` lower bound calculation

davidtaikocha:
To address:
> I'm thinking about one validation edge case, rn in the current Shasta derivation pipeline, we have a TIMESTAMP_MAX_OFFSET (384 atm and will be updated to 1536) in timestamp [lower bound validation](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/docs/Derivation.md#timestamp-validation) , so it seems there is a chance (or edge case) that, during [SHASTA_FORK_TIME, SHASTA_FORK_TIME+TIMESTAMP_MAX_OFFSET), a preconfer / proposer can produce a Shasta block whose timestamp is earlier than the SHASTA_FORK_TIME , but still can pass the current validation rule?


ggonzalez94 (approved):
Lgtm

YoGhurt111 (approved):

PR #20770, opened 2025-11-19
feat(protocol): reject invalid block values instead of auto-setting in derivation pipeline

davidtaikocha:
- clarify that consensus no longer clamps `metadata.timestamp` or `gasLimit`; out-of-bound values now trigger default source replacement
  - update forced-inclusion requirements to note timestamps inherit `parent.timestamp + 1` and tweak the table formatting for clarity
  - remove the rule of fallback to `proposer` address when `block.coinbase == 0`


packages/protocol/docs/Derivation.md
```diff
@@ -266,11 +266,11 @@ With the extracted `ProposalManifest`, metadata computation proceeds using both
...
+Timestamp validation is performed collectively across all blocks:
 
-1. **Upper bound enforcement**: If `metadata.timestamp > proposal.timestamp`, set `metadata.timestamp = proposal.timestamp`
-2. **Lower bound calculation**: `lowerBound = max(parent.metadata.timestamp + 1, proposal.timestamp - TIMESTAMP_MAX_OFFSET)`
-3. **Lower bound enforcement**: If `metadata.timestamp < lowerBound`, set `metadata.timestamp = lowerBound`
+1. **Upper bound**: `proposal.timestamp`
+2. **Lower bound**: `lowerBound = max(parent.metadata.timestamp + 1, proposal.timestamp - TIMESTAMP_MAX_OFFSET)`
+3. **Out-of-bounds handling**: If any block's `manifest.blocks[i].timestamp` is outside `[lowerBound, proposal.timestamp]`, the entire derivation source is replaced with the default source manifest.
 
 #### `anchorBlockNumber` Validation
```
YoGhurt111, line 284: IMO, we also need to strictly check the `anchorBlockNumber`.
Because if we adjust the `anchorBlockNumber` in current preconf stage, it would cause the reorg, it's meaningless.
Let they decide the value, we only check and adjust the 0 value.
WDYT?

davidtaikocha: i don't think we are adjusting `anchorBlockNumber`? no?

packages/protocol/docs/Derivation.md
```diff
@@ -248,13 +248,13 @@ manifest.sources = [sourceManifest0, sourceManifest1, ...];  // With defaults fo
...
-| Field               | Required Value | Reason                                                  |
-| ------------------- | -------------- | ------------------------------------------------------- |
-| `timestamp`         | `0`            | Protocol assigns based on proposal timing               |
-| `coinbase`          | `address(0)`   | Protocol uses `proposal.proposer` for forced inclusions |
-| `anchorBlockNumber` | `0`            | Protocol inherits from parent block                     |
-| `gasLimit`          | `0`            | Protocol inherits from parent block                     |
-| `transactions`      | User-provided  | The actual transactions to be forcibly included         |
+| Field               | Required Value | Reason                                                                |
+| ------------------- | -------------- | --------------------------------------------------------------------- |
+| `timestamp`         | `0`            | Protocol inherits `parent.timestamp + 1` to maintain monotonicity     |
```
ggonzalez94: since forced inclusions come first in the proposal, don't we run the risk that if the latest blocks are old enough for whatever reason the forced inclusion blocks become invalid as well because of the timestamp constrains?

ggonzalez94: shouldn't we instead set it to the lower bound?
`max(parent.metadata.timestamp + 1, proposal.timestamp - TIMESTAMP_MAX_OFFSET)`

packages/protocol/docs/Derivation.md
YoGhurt111: We set `anchorBlockNumber` to `parent.metadata.anchorBlockNumber`.
When the followings met:
- **Non-monotonic progression**: `manifest.blocks[i].anchorBlockNumber < parent.metadata.anchorBlockNumber`
- **Future reference**: `manifest.blocks[i].anchorBlockNumber >= proposal.originBlockNumber - MIN_ANCHOR_OFFSET`
- **Excessive lag**: `manifest.blocks[i].anchorBlockNumber < proposal.originBlockNumber - MAX_ANCHOR_OFFSET`

Within the strictly checks, we should return default.

packages/protocol/docs/Derivation.md
```diff
@@ -293,25 +293,21 @@ The anchor hash and state root must always correspond to the actual L1 block ref
 The L2 coinbase address determination follows a hierarchical priority system:
 
 1. **Forced inclusions**: Always use `proposal.proposer`
-2. **Regular proposals**: Use `orderedBlocks[i].coinbase` if non-zero
-3. **Fallback**: Use `proposal.proposer` if manifest coinbase is `address(0)`
+2. **Regular proposals**: Use `orderedBlocks[i].coinbase`
```
ggonzalez94, line 305: what happens if the proposer sets it to zero? do we invalidate it or just burn the fees(i.e. send to the zero address)? do we make this changes because node or prover becomes simpler?

packages/protocol/docs/Derivation.md
```diff
@@ -293,25 +293,21 @@ The anchor hash and state root must always correspond to the actual L1 block ref
...
-**Calculation process**:
+**Validation process**:
 
 1. **Define bounds**:
 
    - `lowerBound = max(parent.metadata.gasLimit * (10000 - BLOCK_GAS_LIMIT_MAX_CHANGE) / 10000, MIN_BLOCK_GAS_LIMIT)`
    - `upperBound = min(parent.metadata.gasLimit * (10000 + BLOCK_GAS_LIMIT_MAX_CHANGE) / 10000, MAX_BLOCK_GAS_LIMIT)`
 
-2. **Apply constraints**:
-   - If `manifest.blocks[i].gasLimit == 0`: Inherit parent value
```
ggonzalez94, line 311: I think this was useful, wasn't it?

packages/protocol/docs/Derivation.md
ggonzalez94: I think we are if I understand gavin's concern correctly. At least according to the doc:

```
Anchor block validation ensures proper L1 state synchronization and may trigger manifest replacement:

**Invalidation conditions** (sets `anchorBlockNumber` to `parent.metadata.anchorBlockNumber`):

- **Non-monotonic progression**: `manifest.blocks[i].anchorBlockNumber < parent.metadata.anchorBlockNumber`
- **Future reference**: `manifest.blocks[i].anchorBlockNumber >= proposal.originBlockNumber - MIN_ANCHOR_OFFSET`
- **Excessive lag**: `manifest.blocks[i].anchorBlockNumber < proposal.originBlockNumber - MAX_ANCHOR_OFFSET`
```

I believe this can have 2 problems:
1. it is confusing which validations rules invalidate the derivation source, and which ones just cause replacement. Maybe we should just be consistent and choose a path for all conditions that are under the proposer's control(i.e. like we did with the timestamp validation)
2. once we launch permisionless preconfs if preconfer signs a block with an anchor block number, and then it gets changed on their proposal they will be slashed

ggonzalez94 (commented):
Left a few questions

packages/protocol/docs/Derivation.md
davidtaikocha: pushed an update: https://github.com/taikoxyz/taiko-mono/pull/20770/commits/ae1f5fe69b707576f1a5c1d95d4f3874143e8329 , we still allow forced inclusion to skip the monotonic progression check, let me know if you also want to remove this check skip.

packages/protocol/docs/Derivation.md
```diff
@@ -266,23 +266,23 @@ With the extracted `ProposalManifest`, metadata computation proceeds using both
...
 
-Timestamp validation is performed collectively across all blocks and may result in block count reduction:
+Timestamp validation is performed collectively across all blocks:
 
-1. **Upper bound enforcement**: If `metadata.timestamp > proposal.timestamp`, set `metadata.timestamp = proposal.timestamp`
-2. **Lower bound calculation**: `lowerBound = max(parent.metadata.timestamp + 1, proposal.timestamp - TIMESTAMP_MAX_OFFSET)`
-3. **Lower bound enforcement**: If `metadata.timestamp < lowerBound`, set `metadata.timestamp = lowerBound`
+1. **Upper bound**: `proposal.timestamp`
+2. **Lower bound**: `lowerBound = max(parent.metadata.timestamp + 1, proposal.timestamp - TIMESTAMP_MAX_OFFSET)`
+3. **Out-of-bounds handling**: If any block's `manifest.blocks[i].timestamp` is outside `[lowerBound, proposal.timestamp]`, the entire derivation source is replaced with the default source manifest.
```
YoGhurt111, line 282: And after returning default, I supposed to change the rule how we handle the default manifest's timestamp: `parent's timestamp + 1`, current impl is `proposal.timestamp`.

packages/protocol/docs/Derivation.md
davidtaikocha: updated: https://github.com/taikoxyz/taiko-mono/pull/20770/commits/530466e8456f00d8ab7ec4a97f03f729f454e7c7

packages/protocol/docs/Derivation.md
davidtaikocha: > what happens if the proposer sets it to zero? do we invalidate it or just burn the fees(i.e. send to the zero address)?

We just allow it (it's also possible on L1 imo).

> do we make this changes because node or prover becomes simpler?

made this change to follow the consensus in slack that: we don't want to let the consensus overwrite preconfer / proposer's value, which is inconvenient to prove.

packages/protocol/docs/Derivation.md
davidtaikocha: It was for forced inclusion iiuc. And we already have: 
```
| `gasLimit`          | `0`            | Protocol inherits from parent block                                   |
```

packages/protocol/docs/Derivation.md
YoGhurt111: https://github.com/taikoxyz/taiko-mono/pull/20770#discussion_r2540255808 This approach is better.

packages/protocol/docs/Derivation.md
```diff
@@ -293,25 +301,21 @@ The anchor hash and state root must always correspond to the actual L1 block ref
...
    - `lowerBound = max(parent.metadata.gasLimit * (10000 - BLOCK_GAS_LIMIT_MAX_CHANGE) / 10000, MIN_BLOCK_GAS_LIMIT)`
    - `upperBound = min(parent.metadata.gasLimit * (10000 + BLOCK_GAS_LIMIT_MAX_CHANGE) / 10000, MAX_BLOCK_GAS_LIMIT)`
 
-2. **Apply constraints**:
-   - If `manifest.blocks[i].gasLimit == 0`: Inherit parent value
-   - If below `lowerBound`: Clamp to `lowerBound`
-   - If above `upperBound`: Clamp to `upperBound`
-   - Otherwise: Use manifest value unchanged
+2. **Source validation**:
+   - If `manifest.blocks[i].gasLimit` is non-zero and falls outside `[lowerBound, upperBound]`: Replace the entire derivation source with the default source manifest.
```
YoGhurt111:
```suggestion
   - If `manifest.blocks[i].gasLimit` falls outside `[lowerBound, upperBound]`: Replace the entire derivation source with the default source manifest.
```
Follows the rule that we do not modify the normal proposal’s value.

packages/protocol/docs/Derivation.md
```diff
@@ -248,13 +256,13 @@ manifest.sources = [sourceManifest0, sourceManifest1, ...];  // With defaults fo
...
 Users submit forced inclusion transactions directly to L1 by posting blob data containing a `DerivationSourceManifest` struct. To ensure valid forced inclusions that pass validation, the following `BlockManifest` fields must be set to zero, allowing the protocol to assign appropriate values:
 
-| Field               | Required Value | Reason                                                  |
-| ------------------- | -------------- | ------------------------------------------------------- |
-| `timestamp`         | `0`            | Protocol assigns based on proposal timing               |
-| `coinbase`          | `address(0)`   | Protocol uses `proposal.proposer` for forced inclusions |
-| `anchorBlockNumber` | `0`            | Protocol inherits from parent block                     |
-| `gasLimit`          | `0`            | Protocol inherits from parent block                     |
-| `transactions`      | User-provided  | The actual transactions to be forcibly included         |
+| Field               | Value                                                                       |
```
YoGhurt111, line 260: Need to add the rule for FI: only allow 1 blob and 1 block.

davidtaikocha: Let's do it in another PR later, since it's not part of this PR's scope.

ggonzalez94 (approved):

johntaiko (approved):

PR #20776, opened 2025-11-19
fix(tools): resume from last processed chunk

strmfos:
When the scanner restarts it now resumes from the last processed block instead of rereading old chunks. The regex pulls the chunk end height, so we skip the duplicated ranges that used to happen on reruns.


davidtaikocha:
Closed due to being outdated.

PR #20777, opened 2025-11-19
docs(bridging): use Inbox/Anchor and correct vault/proof flow

Snezhkko:
1. Replaced outdated mentions of TaikoInbox/TaikoAnchor with current contracts Inbox (packages/protocol/contracts/layer1/core/impl/Inbox.sol) and Anchor (packages/protocol/contracts/layer2/core/Anchor.sol).
2. Updated checkpointing explanation to the current mechanism: authorized syncers persist checkpoints via SignalService.saveCheckpoint(...) as ICheckpointStore (packages/protocol/contracts/shared/signal/SignalService.sol).
3. Corrected vault references and flows:
- Use ERC20Vault, ERC721Vault, ERC1155Vault (packages/protocol/contracts/shared/vault/).
- Clarify that vaults construct Bridge.Message and call Bridge.sendMessage (packages/protocol/contracts/shared/bridge/Bridge.sol); proofs are generated off-chain by relayers (not by the vaults).
- For bridging back, instruct to call sendToken on the Vault with the bridged token; the Vault burns and canonical vault releases/mints upon proof.

### DIFF
--- a/packages/docs-site/src/content/docs/taiko-alethia-protocol/protocol-architecture/bridging.md
+++ b/packages/docs-site/src/content/docs/taiko-alethia-protocol/protocol-architecture/bridging.md
@@ -19,12 +19,10 @@ Taiko Alethia uses a **trust-minimized**, **Ethereum-equivalent** design for sec
 
 Two smart contracts are responsible for maintaining cross-chain state:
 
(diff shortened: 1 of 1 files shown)


davidtaikocha:
The doc for `Pacaya` design atm.

PR #20780, opened 2025-11-19
fix(taiko-client-rs): set safe/finalized to zero and accept VALID/SYNCING forkchoice

phrwlk:
This change aligns `beacon.rs::submit_remote_block()` with the Engine API specification and internal project semantics. The Engine API permits `safeBlockHash` and `finalizedBlockHash` to be zero unless the transition block is finalized, and mandates that non-zero values belong to the head chain. The previous code unconditionally set `safe_block_hash` to the new head and `finalized_block_hash` to the parent, which can violate finality semantics and trigger invalid forkchoice state errors. We now set both to `B256::ZERO`, matching the conservative approach used in engine.rs after insertion, until reliable safe/finalized values are known. Additionally, broaden forkchoice outcome handling to accept both VALID and SYNCING responses, as per the spec; only INVALID (or any unexpected status) results in an error. This prevents spurious failures when the engine already knows and validates the head.

### DIFF
--- a/packages/taiko-client-rs/crates/driver/src/sync/beacon.rs
+++ b/packages/taiko-client-rs/crates/driver/src/sync/beacon.rs
@@ -3,7 +3,7 @@
 use std::{borrow::Cow, marker::PhantomData, time::Duration};
 
 use alethia_reth_primitives::engine::types::TaikoExecutionDataSidecar;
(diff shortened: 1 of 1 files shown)


davidtaikocha:
It's not aligned with what we're doing in Go.

PR #20787, opened 2025-11-20
fix(protocol): update MAX_BLOCK_GAS_LIMIT to 45,000,000 in derivation doc

smtmfft:
To reduce the prover’s exposure to spam


davidtaikocha (approved):

YoGhurt111 (approved):

ggonzalez94 (approved):
Looks good. We might want to rethink this if we use the prover market and can afford to have more expensive proofs

PR #20795, opened 2025-11-20
fix(relayer): correct destination chain sync in multihop bridge

reject-i:
Fixed a critical bug in multihop bridge message processing where the final header sync check incorrectly validated the last hop chain against itself instead of the destination chain.

Bug introduced in commit 1039a96 (PR #15669, March 2024) during two-step bridge refactoring. Changed hopChainID.Uint64() to p.destChainId.Uint64() and removed unused hopChainID variable.

This ensures the relayer properly waits for destination chain sync before processing cross-chain messages, preventing proof generation failures.


davidtaikocha:
@claude plz review.

cyberhorsey:
multi-hop is going away with Shasta and this section of code should be removed then. and even then ,we never use multi-hop because we dont have a layer 3. regardless, we can approve and merge it.

davidtaikocha:
Closed since we are not using multi-hop and will deprecate it in Shasta.

PR #20804, opened 2025-11-22
feat(protocol): simplified token unlock

RogerLamTd:


packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,146 @@
...
+/// @notice Manages the unlocking of Taiko tokens after 6 months.
+/// Tokens granted off-chain are deposited into this contract directly from the `msg.sender`
+/// address. Token withdrawals are permitted only at the 6 month mark.
+/// A separate instance of this contract is deployed for each recipient.
+/// @custom:security-contact security@taiko.xyz
+contract SimpleTokenUnlock is EssentialContract {
+    using SafeERC20 for IERC20;
+    using LibMath for uint256;
+
+    uint256 public constant SIX_MONTHS = 182.5 days;
```
dantaik: you can make this an immutable and initiate it in the constructor.

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,146 @@
...
+    /// @param oldRecipient The old recipient address.
+    /// @param newRecipient The new recipient address.
+    event RecipientChanged(address indexed oldRecipient, address indexed newRecipient);
+
+    error INVALID_PARAM();
+    error NOT_WITHDRAWABLE();
+    error PERMISSION_DENIED();
+
+    modifier onlyRecipient() {
+        if (msg.sender != recipient) revert PERMISSION_DENIED();
```
dantaik: we now prefer require-statements.
```suggestion
        require (msg.sender == recipient, PERMISSION_DENIED());
```

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,146 @@
...
+        _;
+    }
+
+    modifier onlyRecipientOrOwner() {
+        if (msg.sender != recipient && msg.sender != owner()) revert PERMISSION_DENIED();
+        _;
+    }
+
+    constructor(address _resolver) EssentialContract(_resolver) { }
```
dantaik: we now also prefer NOT to use the resolver, you can use immutable for taiko token. But if such changes require a lot of work, using resolver is still fine.

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,146 @@
...
+/// @custom:security-contact security@taiko.xyz
+contract SimpleTokenUnlock is EssentialContract {
+    using SafeERC20 for IERC20;
+    using LibMath for uint256;
+
+    uint256 public constant SIX_MONTHS = 182.5 days;
+
+    uint256 public amountGranted; // slot 1
+    address public recipient; // 20 bytes
+    uint64 public grantTimestamp; // 1764460800 for 30 Nov 2025, 00:00:00 UTC. 8 bytes
```
dantaik: I suggest to change _grantTimestamp to immutable as it is shared by all deployments.

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,146 @@
...
+    function withdraw(
+        address _to,
+        uint256 _amount
+    )
+        external
+        nonZeroAddr(_to)
+        nonZeroValue(_amount)
+        onlyRecipient
+        nonReentrant
+    {
```
dantaik: lets check if _to is address(0), if so, use the recipient address.

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,146 @@
...
+        if (_amount > amountWithdrawable()) revert NOT_WITHDRAWABLE();
```
dantaik: lets check if _amount is 0, if so, use amountWithdrawable() directly.

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,146 @@
...
+        emit TokenWithdrawn(_to, _amount);
+        IERC20(resolve(LibStrings.B_TAIKO_TOKEN, false)).safeTransfer(_to, _amount);
```
dantaik: it seems amountWithdrawable() will not return a different value after this call, there is no amountWithdrawn state variable to track how much has been withdrwawan?

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,146 @@
...
+        nonZeroAddr(_recipient)
+        nonZeroValue(_grantTimestamp)
+        initializer
+    {
+        if (_owner == _recipient) revert INVALID_PARAM();
+
+        __Essential_init(_owner);
+
+        recipient = _recipient;
+        grantTimestamp = _grantTimestamp;
```
dantaik: we also need to delegate voting power to the recipient by default

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,146 @@
...
+        onlyRecipient
+        nonReentrant
+    {
+        if (_amount > amountWithdrawable()) revert NOT_WITHDRAWABLE();
+        emit TokenWithdrawn(_to, _amount);
+        IERC20(resolve(LibStrings.B_TAIKO_TOKEN, false)).safeTransfer(_to, _amount);
+    }
+
+    /// @notice Withdraws all tokens to the recipient address.
+    function withdraw() external nonReentrant {
```
dantaik: we can actually merge this function with the previous one.

dantaik (changes requested):

dantaik:
@claude please review

RogerLamTd:
@claude please review

RogerLamTd:
@claude please review

RogerLamTd:
@claude review please

RogerLamTd:
@claude please review

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,143 @@
...
+
+        amountGranted += _amount;
+        emit TokenGranted(_amount);
+
+        IERC20(TAIKO_TOKEN).safeTransferFrom(
+            msg.sender, address(this), _amount
+        );
+    }
+
+    /// @notice Withdraws tokens by the recipient.
```
Brechtpd:
```suggestion
    /// @notice Withdraws tokens by the recipient or owner.
```

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,143 @@
...
+
+    /// @notice Initializes the contract.
+    /// @param _owner The contract owner address.
+    /// @param _recipient Who will be the grantee for this contract.
+    function init(
+        address _owner,
+        address _recipient
+    )
+        external
+        nonZeroAddr(_recipient)
```
Brechtpd, line 71: `nonZeroAddr(_owner)`? Shouldn't matter but done for the other address so why not.

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,143 @@
...
+import "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
+import "src/shared/common/EssentialContract.sol";
+import "src/shared/libs/LibStrings.sol";
+import "src/shared/libs/LibMath.sol";
+
+/// @title SimpleTokenUnlock
+/// @notice Manages the unlocking of Taiko tokens 6 months after GRANT_TIMESTAMP.
```
Brechtpd:
```suggestion
/// @notice Fully unlocks Taiko tokens 6 months after GRANT_TIMESTAMP.
```

packages/protocol/script/layer1/team/simpletokenunlock/Deploy.s.sol
```diff
@@ -0,0 +1,32 @@
...
+    function run() external broadcast {
+        require(SIMPLE_TOKEN_UNLOCK_IMPL != address(0), "SIMPLE_TOKEN_UNLOCK_IMPL not set");
+
+        string memory path = "/script/simpletokenunlock/Deploy.data.json";
+        address[] memory recipients = abi.decode(
+            vm.parseJson(vm.readFile(string.concat(vm.projectRoot(), path))), (address[])
+        );
+
+        for (uint256 i; i < recipients.length; i++) {
+            address proxy = deploy({
```
Brechtpd, line 28: Could use a `require(recipients[i] != address(0), "zero recipient");` to fail before deploying the proxy instead of after.

packages/protocol/script/layer1/team/simpletokenunlock/Grant.s.sol
```diff
@@ -0,0 +1,58 @@
...
+            SimpleTokenUnlock target = SimpleTokenUnlock(proxy);
+
+            require(target.recipient() == recipient, "recipient mismatch");
+            require(target.owner() == 0x9CBeE534B5D8a6280e01a14844Ee8aF350399C7F, "owner mismatch");
+
+            total += SafeCastUpgradeable.toUint128(items[i].grantAmount * 1e18);
+        }
+
+        console2.log("total:", total / 1e18);
+        require(tko.balanceOf(msg.sender) >= total, "insufficient TKO balance");
```
Brechtpd: Is `msg.sender` the correct address here?

RogerLamTd: should be, only the owner/admin.taiko.eth should call this unless I'm mistaken

Brechtpd: image

Brechtpd: unless run is also called by the same address, then all good.

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,143 @@
...
+
+    /// @notice Withdraws tokens by the recipient.
+    /// @param _to The address the token will be sent to.
+    /// @param _amount The amount of tokens to withdraw.
+    function withdraw(
+        address _to,
+        uint256 _amount
+    )
+        external
+        onlyRecipientOrOwner
```
adaki2004: maybe `onlyRecipient()` instead of `onlyRecipientOrOwner()` ?

RogerLamTd:
@claude please review

Brechtpd:
Claude is really much worse at reviews than codex, but okay I guess chatgpt 5.1 pro is cheating. In any case, strongly recommend using it to review the new changes. I'll just paste the whole output here because it explains it well:

---

Looking at the latest PR (including commit `4eec508…`) and comparing it to the version Brecht reviewed (`3ad5ecc…`), here’s what changed and what still needs work, focusing on behavior, security, and simplicity.

## 1. `SimpleTokenUnlock.sol`

### 1.1. New `amountWithdrawable` / `_getAmountLocked` logic – **critical bug**

**What changed**

Previously, `amountWithdrawable` was:

```solidity
function amountWithdrawable() public view returns (uint256) {
    IERC20 tko = IERC20(TAIKO_TOKEN);
    uint256 balance = tko.balanceOf(address(this));
    if (block.timestamp < GRANT_TIMESTAMP + SIX_MONTHS){
        return 0;
    } else {
        return balance.min(amountGranted);
    }
}
```

Now it is:

```solidity
function amountWithdrawable() public view returns (uint256) {
    IERC20 tko = IERC20(TAIKO_TOKEN);
    uint256 balance = tko.balanceOf(address(this));
    uint256 locked = _getAmountLocked();
    return balance.max(locked) - locked;
}

function _getAmountLocked() internal view returns (uint256) {
    if (block.timestamp < GRANT_TIMESTAMP + SIX_MONTHS) {
        return amountGranted;
    }
    return 0;
}
```

([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])

The intent (as shown by the new test) seems to be:

* Before 6 months, **only “extra” tokens** above `amountGranted` are withdrawable.
* After 6 months, **the full balance** is withdrawable.

This formula does match that *for a single call*:

* Pre‑cliff: withdrawable = `max(balance, amountGranted) − amountGranted`
  → `balance > amountGranted` → `balance − amountGranted` (extra only)
  → `balance ≤ amountGranted` → `0`
* Post‑cliff: locked = 0 → withdrawable = `balance`.

So far so good.

**Where it goes wrong**

The bug comes from the **interaction with the updated `withdraw` logic**:

```solidity
function withdraw(address _to, uint256 _amount)
    external
    onlyRecipientOrOwner
    nonReentrant
{
    if (_to == address(0)) _to = recipient;
    if (_amount == 0) _amount = amountWithdrawable();
    require(_amount <= amountWithdrawable(), NOT_WITHDRAWABLE());

    emit TokenWithdrawn(_to, _amount);
    IERC20(TAIKO_TOKEN).safeTransfer(_to, _amount);

    // only the portion coming from the grant bucket should reduce the grant counter; extra deposits don't touch amountGranted
    uint256 grantedReduction = _amount.min(amountGranted);
    amountGranted -= grantedReduction;
}
```

([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])

The comment is what you *want* (“extra deposits don’t touch `amountGranted`”), but the implementation doesn’t do that.

Take exactly the scenario you added in the test:

* Pre‑cliff.
* `amountGranted = 100 ether`.
* Contract balance = `100.5 ether` (`grant(100)` + direct transfer `0.5`).
* So `amountWithdrawable()` returns `0.5` (only the extra tokens).

Now call `withdraw` once with `_amount = amountWithdrawable() = 0.5`:

1. Pre‑call state:

   * `B = 100.5` (balance)
   * `G = 100` (amountGranted)
   * `locked = G` (pre‑cliff)
   * `withdrawable = B − G = 0.5`.

2. After transfer:

   * `safeTransfer` sends `0.5` to user → new `B' = 100`.
   * `grantedReduction = min(0.5, 100) = 0.5`.
   * `G' = 100 − 0.5 = 99.5`.

3. New withdrawable pre‑cliff:

   * `locked' = G' = 99.5`.
   * `withdrawable' = max(B', locked') − locked' = 100 − 99.5 = 0.5`.

So **withdrawable stays at 0.5**, even though there are now **no extra tokens** – the entire 100 remaining are grant tokens. You can keep calling `withdraw(0.5)` pre‑cliff until the entire 100‑token grant has been drained, which defeats the cliff entirely.

This is a **high‑severity logic bug**: a recipient can fully unlock their grant *before* `GRANT_TIMESTAMP + SIX_MONTHS` by repeatedly withdrawing only the “currently withdrawable” amount.

The test you added (`test_simpletokenunlock_single_vest_withdrawal`) doesn’t catch this because it **never calls `withdraw`**, only `amountWithdrawable`.([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])

---

**Suggested robust fix**

The core issue is that `grantedReduction` needs to account for how much of `_amount` came from the *extra* bucket vs the grant bucket.

You can do that directly from current state without tracking extra storage:

```solidity
function amountWithdrawable() public view returns (uint256) {
    IERC20 tko = IERC20(TAIKO_TOKEN);
    uint256 balance = tko.balanceOf(address(this));

    // pre-cliff: only tokens above the granted amount are withdrawable
    if (block.timestamp < GRANT_TIMESTAMP + SIX_MONTHS) {
        if (balance > amountGranted) {
            return balance - amountGranted;
        }
        return 0;
    }

    // post-cliff: all tokens in the contract are withdrawable
    return balance;
}
```

And in `withdraw`:

```solidity
function withdraw(address _to, uint256 _amount)
    external
    onlyRecipientOrOwner
    nonReentrant
{
    if (_to == address(0)) _to = recipient;
    if (_amount == 0) _amount = amountWithdrawable();

    uint256 available = amountWithdrawable();
    require(_amount <= available, NOT_WITHDRAWABLE());

    IERC20 tko = IERC20(TAIKO_TOKEN);
    uint256 balance = tko.balanceOf(address(this));

    // Extra = any balance above the grant bucket
    uint256 extra = balance > amountGranted ? balance - amountGranted : 0;

    // First consume from extra, then from the grant bucket
    uint256 grantPortion = _amount > extra ? _amount - extra : 0;
    if (grantPortion > 0) {
        // Safety guard if something weird happened to balance/accounting
        uint256 reduction = grantPortion > amountGranted
            ? amountGranted
            : grantPortion;
        amountGranted -= reduction;
    }

    emit TokenWithdrawn(_to, _amount);
    tko.safeTransfer(_to, _amount);
}
```

Key properties of this approach:

* **Pre‑cliff:**
  `extra = balance - amountGranted`, so any pre‑cliff withdrawal is entirely from `extra` → `grantPortion = 0` → `amountGranted` unchanged.
* **Post‑cliff:**
  `amountWithdrawable = balance`, but `grantPortion` is `max(_amount − extra, 0)` → `amountGranted` decreases only when you’re actually consuming grant tokens.
* Prevents the “withdrawable stays constant, drain in chunks” bug.

At minimum, you need to change the `grantedReduction` logic so that pre‑cliff withdrawals **never** decrease `amountGranted`.

---

**Tests to add for this bug**

You should add tests that explicitly try to exploit this:

1. **Pre‑cliff multi‑withdraw attack**

   * Setup same as current test: `grant(100)`, direct transfer `0.5`.
   * While `block.timestamp < grantTimestamp + SIX_MONTHS`:

     * loop: `withdraw(recipient, amountWithdrawable())` until `amountWithdrawable() == 0`.
   * Assert:

     * `amountGranted() == 100 ether` (pre‑cliff, grants untouched).
     * `taikoToken.balanceOf(address(target)) == 100 ether`.
     * Only the extra 0.5 was withdrawn.

2. **Pre‑ then post‑cliff correctness**

   * Pre‑cliff: withdraw extra tokens.
   * Warp past cliff: withdraw everything.
   * Assert that final `amountGranted() == 0` and contract balance is zero.

These tests will fail on the current implementation and pass once the bug is fixed.

---

### 1.2. `nonZeroAddr(_owner)` added to `init` – good, but re‑check behavior

Change:

```solidity
function init(
    address _owner,
    address _recipient
)
    external
    nonZeroAddr(_recipient)
    nonZeroAddr(_owner)
    initializer
{
    require(_owner != _recipient, INVALID_PARAM());

    __Essential_init(_owner);

    recipient = _recipient;
    ERC20VotesUpgradeable(TAIKO_TOKEN).delegate(_recipient);
}
```

([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])

This:

* Fixes the missing zero‑check for `_owner` that Brecht pointed out.
* Keeps the invariant `_owner != _recipient`, which is consistent with the rest of the contract (since `onlyRecipientOrOwner` treats them as separate roles).

No behavior or security issues here; this is an improvement.

---

### 1.3. `withdraw` comment and NatSpec mismatch

NatSpec now says:

```solidity
/// @notice Fully unlocks Taiko tokens 6 months after GRANT_TIMESTAMP.
/// ...
/// Token withdrawals are permitted only at the 6 month mark.
```

([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])

But the test and new `amountWithdrawable` logic intend to allow **pre‑cliff withdrawals of extra tokens** (balance above `amountGranted`). That contradicts “withdrawals are permitted only at the 6 month mark”.

Once you fix the logic, you should align docs to something like:

* “Grant tokens are fully unlocked 6 months after GRANT_TIMESTAMP.”
* “Any tokens sent to this contract *in excess* of the granted amount can be withdrawn immediately.”

Right now comments are misleading for reviewers and auditors.

---

### 1.4. Grant accounting and `grant` openness

`grant` is still:

```solidity
function grant(uint128 _amount) external nonReentrant {
    require(_amount != 0, INVALID_PARAM());
    amountGranted += _amount;
    emit TokenGranted(_amount);

    IERC20(TAIKO_TOKEN).safeTransferFrom(
        msg.sender, address(this), _amount
    );
}
```

([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])

The new test:

```solidity
vm.prank(Carol);
vm.expectRevert();
target.grant(10 ether);
```

reverts only because Carol doesn’t have TKO or allowance, not because of an access control check. The comment in the test (`"revert: Ownable: caller is not the owner"`) is misleading.

Design‑wise, letting *anyone* call `grant` is arguably fine (it lets other parties top up the grant), but then:

* `amountGranted` increases whenever **anyone** funds the contract via `grant`.
* Combined with your cliff logic, this means third‑party donors can **lock** tokens for the recipient until 6 months elapse (since they increase `amountGranted`).

If that’s intentional, this is fine. If you only want the admin/owner to create official “grant” allocations, consider adding `onlyOwner` to `grant` and adjust the test to check the correct revert reason.

Either way, I’d update the test comment to reflect the actual reason it reverts.

---

## 2. `DeploySimpleTokenUnlock.s.sol`

**New change**

You added a zero‑recipient guard in the loop:

```solidity
for (uint256 i; i < recipients.length; i++) {
    require(recipients[i] != address(0), "zero recipient");
    address proxy = deploy({
        name: "",
        impl: SIMPLE_TOKEN_UNLOCK_IMPL,
        data: abi.encodeCall(SimpleTokenUnlock.init, (OWNER, recipients[i]))
    });
    ...
}
```

([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])

This is a good, simple robustness improvement.

**Ongoing issues (unchanged since Brecht’s version)**

1. **`SIMPLE_TOKEN_UNLOCK_IMPL` is still undefined**

   The script still has:

   ```solidity
   // NOT YET DEPLOYED
   // address public SIMPLE_TOKEN_UNLOCK_IMPL =
   function run() external broadcast {
       require(SIMPLE_TOKEN_UNLOCK_IMPL != address(0), "SIMPLE_TOKEN_UNLOCK_IMPL not set");
       ...
   }
   ```

   There is no definition of `SIMPLE_TOKEN_UNLOCK_IMPL` anywhere in the repo. The script will not compile and is unusable as is.([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])

   Recommended:

   * Either define it in this script:

     ```solidity
     address public constant SIMPLE_TOKEN_UNLOCK_IMPL =
         0x...; // real deployed implementation
     ```

   * Or pass it via `BaseScript` / `vm.envAddress` and rename this script accordingly so it’s obvious how to set it.

2. **`name: ""` in `deploy`**

   `deploy({ name: "", impl: ..., data: ... })` gives no identifier for the proxy in your deployment tooling. For maintenance, it’s nicer to give these a name like `"simple_token_unlock"` or `"simple_token_unlock_recipient"` (you already use `"simple_token_unlock"` in the tests).([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])

These are not security issues, but they affect operability and clarity.

---

## 3. `GrantSimpleTokenUnlock.s.sol`

**Change since Brecht’s review**

Previously:

```solidity
console2.log("total:", total / 1e18);
require(tko.balanceOf(msg.sender) >= total, "insufficient TKO balance");

vm.startBroadcast();
for (uint256 i; i < items.length; i++) {
    uint128 grantAmount = uint128(items[i].grantAmount * 1e18);
    tko.approve(items[i].proxy, grantAmount);
    SimpleTokenUnlock(items[i].proxy).grant(grantAmount);
}
vm.stopBroadcast();
```

Now:

```solidity
console2.log("total:", total / 1e18);
vm.startBroadcast();
require(tko.balanceOf(msg.sender) >= total, "insufficient TKO balance");
for (uint256 i; i < items.length; i++) {
    uint128 grantAmount = uint128(items[i].grantAmount * 1e18);
    tko.approve(items[i].proxy, grantAmount);
    SimpleTokenUnlock(items[i].proxy).grant(grantAmount);
}
vm.stopBroadcast();
```

([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])

This is precisely what Brecht flagged: outside the broadcast, `msg.sender` is the **script contract**, not the EOA. Inside a `vm.startBroadcast()` block, `msg.sender` becomes the configured sender account (EOA), as per the Foundry docs and discussion.([[getfoundry.sh](https://getfoundry.sh/reference/cheatcodes/start-broadcast/?utm_source=chatgpt.com)][3])

So this change:

* Fixes the balance check to correctly look at the EOA’s TKO balance, not the script’s.
* Aligns the check with the account that actually does the `approve` and `grant` calls.

**Behaviorally & security‑wise:**

* `total` is computed as a sum of `SafeCastUpgradeable.toUint128(items[i].grantAmount * 1e18)`. If this overflows `uint128` it will revert, which is a nice safety guard.
* After `vm.startBroadcast`, all calls (`balanceOf`, `approve`, `grant`) run as the same `msg.sender` EOA, so allowances and transfers line up correctly.

This change is correct and fixes the earlier concern.

**Small robustness / simplicity suggestions**

Not strictly required, but would make it clearer:

1. Explicitly choose the broadcaster:

   ```solidity
   address deployer = vm.envAddress("GRANT_ADMIN");
   vm.startBroadcast(deployer);
   require(tko.balanceOf(deployer) >= total, "insufficient TKO balance");
   ```

   This makes the code’s intent obvious and avoids relying on default broadcast semantics.

2. Consider using `SafeERC20` in the script too, to mirror the contract’s use of `SafeERC20` and fail more gracefully on non‑standard ERC‑20s.

---

## 4. `SimpleTokenUnlock.t.sol` (new tests)

You added a new dedicated test suite:([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])

* Uses a local `MyERC20` (ERC20Votes) and `vm.etch` / `vm.store` to stand in for the real TKO at the canonical address.
* Deploys `SimpleTokenUnlock` via the same `deploy()` helper as the rest of the protocol.
* Registers `"taiko_token"` and `"taiko"` in the resolver, so the contract behaves more realistically.

**The tests:**

1. `test_simpletokenunlock_single_vest_withdrawal`

   * Checks:

     * Only Alice can successfully call `grant` (via ERC‑20 balance/allowance, not access control).
     * Extra tokens sent directly to the contract (`taikoToken.transfer(address(target), 0.5 ether)`) are immediately reflected in `amountWithdrawable` pre‑cliff.
     * Pre‑cliff, `amountWithdrawable` is constant across time warps.
     * Post‑cliff, the full balance is withdrawable.
   * This encodes the new desired semantics (grant locked until cliff; extra tokens free).

   **Issue:** As mentioned above, it **never calls `withdraw`**, so it doesn’t test the actual withdrawal path and doesn’t catch the critical accounting bug.

2. `test_simpletokenunlock_delegate`

   * Verifies:

     * `grant` funds the contract.
     * `amountWithdrawable` is zero pre‑cliff when there are no extra tokens.
     * Delegation can be changed by the recipient, and that `taikoToken.delegates(address(target))` equals the delegatee.
   * This aligns with ERC20Votes semantics (the contract address is the token holder; it delegates to the chosen address).

3. `test_simpletokenunlock_delegate_change_recipient`

   * Same as above but also:

     * Allows recipient to change `recipient` to `David`.
     * Keeps the existing delegation (`delegates(address(target)) == Carol`) stable across `changeRecipient`.

   This is good: it demonstrates that governance voting power follows the contract, not the `recipient` field, which is exactly what you want.

**Recommendations for test coverage**

To really make this feature robust, I’d add at least:

* A test that calls `withdraw` pre‑cliff **twice+** and asserts that `amountGranted` remains unchanged and only the extra tokens are withdrawn (this will fail right now).
* A test that after the cliff, multiple `withdraw` calls cannot withdraw more than the total deposited (grants + extras).
* A test that:

  * Owner can call `withdraw` (using `onlyRecipientOrOwner`).
  * A random address cannot (reverts with `PERMISSION_DENIED()`).

---

## 5. Summary of what’s fixed vs still open

**Fixed / improved since Brecht’s review**

* `DeploySimpleTokenUnlock`: guard against zero recipients in the loop.
* `GrantSimpleTokenUnlock`: `tko.balanceOf(msg.sender)` check moved into the `vm.startBroadcast()` block, so it uses the broadcaster EOA instead of the script contract.
* `SimpleTokenUnlock.init`: added `nonZeroAddr(_owner)` to match `_recipient` and preserve invariants.
* New tests capture the intended “extra tokens withdrawable pre‑cliff, grants post‑cliff” behavior and delegation behavior.

**New / remaining issues**

1. **High‑severity bug in `withdraw` + `amountWithdrawable`**

   * Pre‑cliff, repeated calls to `withdraw(amountWithdrawable())` can drain the entire grant over many calls, completely bypassing the cliff.
   * Fix requires changing how `grantedReduction` is computed (see section 1.1).

2. **Docs / NatSpec are now inconsistent with behavior**

   * Comments still claim withdrawals “only at the 6 month mark” while tests and `amountWithdrawable` allow pre‑cliff withdrawals of extra tokens.
   * Should be updated once logic is finalized.

3. **`DeploySimpleTokenUnlock` script still non‑functional**

   * `SIMPLE_TOKEN_UNLOCK_IMPL` remains undefined anywhere in the repo.
   * `name: ""` is unhelpful for deployment tooling.

4. **Test suite doesn’t exercise `withdraw` yet**

   * Tests only read `amountWithdrawable`; they don’t assert correct grant accounting and invariants along the withdrawal path.

5. **`grant` access pattern is a bit opaque**

   * Behavior is “anyone can grant if they’re willing to transfer tokens into the contract”.
   * This *might* be intended; if so, test comments should be updated to reflect the real reason for the revert.

If you fix the pre‑cliff withdrawal accounting and add tests that enforce the intended invariants, this will be much more robust and much easier to reason about in an audit.

RogerLamTd:
@claude please review

RogerLamTd:
@claude please review

RogerLamTd:
> Claude is really much worse at reviews than codex, but okay I guess chatgpt 5.1 pro is cheating. In any case, strongly recommend using it to review the new changes. I'll just paste the whole output here because it explains it well:
> 
> Looking at the latest PR (including commit `4eec508…`) and comparing it to the version Brecht reviewed (`3ad5ecc…`), here’s what changed and what still needs work, focusing on behavior, security, and simplicity.
> 
> ## 1. `SimpleTokenUnlock.sol`
> ### 1.1. New `amountWithdrawable` / `_getAmountLocked` logic – **critical bug**
> **What changed**
> 
> Previously, `amountWithdrawable` was:
> 
> ```solidity
> function amountWithdrawable() public view returns (uint256) {
>     IERC20 tko = IERC20(TAIKO_TOKEN);
>     uint256 balance = tko.balanceOf(address(this));
>     if (block.timestamp < GRANT_TIMESTAMP + SIX_MONTHS){
>         return 0;
>     } else {
>         return balance.min(amountGranted);
>     }
> }
> ```
> 
> Now it is:
> 
> ```solidity
> function amountWithdrawable() public view returns (uint256) {
>     IERC20 tko = IERC20(TAIKO_TOKEN);
>     uint256 balance = tko.balanceOf(address(this));
>     uint256 locked = _getAmountLocked();
>     return balance.max(locked) - locked;
> }
> 
> function _getAmountLocked() internal view returns (uint256) {
>     if (block.timestamp < GRANT_TIMESTAMP + SIX_MONTHS) {
>         return amountGranted;
>     }
>     return 0;
> }
> ```
> 
> ([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])
> 
> The intent (as shown by the new test) seems to be:
> 
> * Before 6 months, **only “extra” tokens** above `amountGranted` are withdrawable.
> * After 6 months, **the full balance** is withdrawable.
> 
> This formula does match that _for a single call_:
> 
> * Pre‑cliff: withdrawable = `max(balance, amountGranted) − amountGranted`
>   → `balance > amountGranted` → `balance − amountGranted` (extra only)
>   → `balance ≤ amountGranted` → `0`
> * Post‑cliff: locked = 0 → withdrawable = `balance`.
> 
> So far so good.
> 
> **Where it goes wrong**
> 
> The bug comes from the **interaction with the updated `withdraw` logic**:
> 
> ```solidity
> function withdraw(address _to, uint256 _amount)
>     external
>     onlyRecipientOrOwner
>     nonReentrant
> {
>     if (_to == address(0)) _to = recipient;
>     if (_amount == 0) _amount = amountWithdrawable();
>     require(_amount <= amountWithdrawable(), NOT_WITHDRAWABLE());
> 
>     emit TokenWithdrawn(_to, _amount);
>     IERC20(TAIKO_TOKEN).safeTransfer(_to, _amount);
> 
>     // only the portion coming from the grant bucket should reduce the grant counter; extra deposits don't touch amountGranted
>     uint256 grantedReduction = _amount.min(amountGranted);
>     amountGranted -= grantedReduction;
> }
> ```
> 
> ([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])
> 
> The comment is what you _want_ (“extra deposits don’t touch `amountGranted`”), but the implementation doesn’t do that.
> 
> Take exactly the scenario you added in the test:
> 
> * Pre‑cliff.
> * `amountGranted = 100 ether`.
> * Contract balance = `100.5 ether` (`grant(100)` + direct transfer `0.5`).
> * So `amountWithdrawable()` returns `0.5` (only the extra tokens).
> 
> Now call `withdraw` once with `_amount = amountWithdrawable() = 0.5`:
> 
> 1. Pre‑call state:
>    
>    * `B = 100.5` (balance)
>    * `G = 100` (amountGranted)
>    * `locked = G` (pre‑cliff)
>    * `withdrawable = B − G = 0.5`.
> 2. After transfer:
>    
>    * `safeTransfer` sends `0.5` to user → new `B' = 100`.
>    * `grantedReduction = min(0.5, 100) = 0.5`.
>    * `G' = 100 − 0.5 = 99.5`.
> 3. New withdrawable pre‑cliff:
>    
>    * `locked' = G' = 99.5`.
>    * `withdrawable' = max(B', locked') − locked' = 100 − 99.5 = 0.5`.
> 
> So **withdrawable stays at 0.5**, even though there are now **no extra tokens** – the entire 100 remaining are grant tokens. You can keep calling `withdraw(0.5)` pre‑cliff until the entire 100‑token grant has been drained, which defeats the cliff entirely.
> 
> This is a **high‑severity logic bug**: a recipient can fully unlock their grant _before_ `GRANT_TIMESTAMP + SIX_MONTHS` by repeatedly withdrawing only the “currently withdrawable” amount.
> 
> The test you added (`test_simpletokenunlock_single_vest_withdrawal`) doesn’t catch this because it **never calls `withdraw`**, only `amountWithdrawable`.([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])
> 
> **Suggested robust fix**
> 
> The core issue is that `grantedReduction` needs to account for how much of `_amount` came from the _extra_ bucket vs the grant bucket.
> 
> You can do that directly from current state without tracking extra storage:
> 
> ```solidity
> function amountWithdrawable() public view returns (uint256) {
>     IERC20 tko = IERC20(TAIKO_TOKEN);
>     uint256 balance = tko.balanceOf(address(this));
> 
>     // pre-cliff: only tokens above the granted amount are withdrawable
>     if (block.timestamp < GRANT_TIMESTAMP + SIX_MONTHS) {
>         if (balance > amountGranted) {
>             return balance - amountGranted;
>         }
>         return 0;
>     }
> 
>     // post-cliff: all tokens in the contract are withdrawable
>     return balance;
> }
> ```
> 
> And in `withdraw`:
> 
> ```solidity
> function withdraw(address _to, uint256 _amount)
>     external
>     onlyRecipientOrOwner
>     nonReentrant
> {
>     if (_to == address(0)) _to = recipient;
>     if (_amount == 0) _amount = amountWithdrawable();
> 
>     uint256 available = amountWithdrawable();
>     require(_amount <= available, NOT_WITHDRAWABLE());
> 
>     IERC20 tko = IERC20(TAIKO_TOKEN);
>     uint256 balance = tko.balanceOf(address(this));
> 
>     // Extra = any balance above the grant bucket
>     uint256 extra = balance > amountGranted ? balance - amountGranted : 0;
> 
>     // First consume from extra, then from the grant bucket
>     uint256 grantPortion = _amount > extra ? _amount - extra : 0;
>     if (grantPortion > 0) {
>         // Safety guard if something weird happened to balance/accounting
>         uint256 reduction = grantPortion > amountGranted
>             ? amountGranted
>             : grantPortion;
>         amountGranted -= reduction;
>     }
> 
>     emit TokenWithdrawn(_to, _amount);
>     tko.safeTransfer(_to, _amount);
> }
> ```
> 
> Key properties of this approach:
> 
> * **Pre‑cliff:**
>   `extra = balance - amountGranted`, so any pre‑cliff withdrawal is entirely from `extra` → `grantPortion = 0` → `amountGranted` unchanged.
> * **Post‑cliff:**
>   `amountWithdrawable = balance`, but `grantPortion` is `max(_amount − extra, 0)` → `amountGranted` decreases only when you’re actually consuming grant tokens.
> * Prevents the “withdrawable stays constant, drain in chunks” bug.
> 
> At minimum, you need to change the `grantedReduction` logic so that pre‑cliff withdrawals **never** decrease `amountGranted`.
> 
> **Tests to add for this bug**
> 
> You should add tests that explicitly try to exploit this:
> 
> 1. **Pre‑cliff multi‑withdraw attack**
>    
>    * Setup same as current test: `grant(100)`, direct transfer `0.5`.
>    * While `block.timestamp < grantTimestamp + SIX_MONTHS`:
>      
>      * loop: `withdraw(recipient, amountWithdrawable())` until `amountWithdrawable() == 0`.
>    * Assert:
>      
>      * `amountGranted() == 100 ether` (pre‑cliff, grants untouched).
>      * `taikoToken.balanceOf(address(target)) == 100 ether`.
>      * Only the extra 0.5 was withdrawn.
> 2. **Pre‑ then post‑cliff correctness**
>    
>    * Pre‑cliff: withdraw extra tokens.
>    * Warp past cliff: withdraw everything.
>    * Assert that final `amountGranted() == 0` and contract balance is zero.
> 
> These tests will fail on the current implementation and pass once the bug is fixed.
> 
> ### 1.2. `nonZeroAddr(_owner)` added to `init` – good, but re‑check behavior
> Change:
> 
> ```solidity
> function init(
>     address _owner,
>     address _recipient
> )
>     external
>     nonZeroAddr(_recipient)
>     nonZeroAddr(_owner)
>     initializer
> {
>     require(_owner != _recipient, INVALID_PARAM());
> 
>     __Essential_init(_owner);
> 
>     recipient = _recipient;
>     ERC20VotesUpgradeable(TAIKO_TOKEN).delegate(_recipient);
> }
> ```
> 
> ([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])
> 
> This:
> 
> * Fixes the missing zero‑check for `_owner` that Brecht pointed out.
> * Keeps the invariant `_owner != _recipient`, which is consistent with the rest of the contract (since `onlyRecipientOrOwner` treats them as separate roles).
> 
> No behavior or security issues here; this is an improvement.
> 
> ### 1.3. `withdraw` comment and NatSpec mismatch
> NatSpec now says:
> 
> ```solidity
> /// @notice Fully unlocks Taiko tokens 6 months after GRANT_TIMESTAMP.
> /// ...
> /// Token withdrawals are permitted only at the 6 month mark.
> ```
> 
> ([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])
> 
> But the test and new `amountWithdrawable` logic intend to allow **pre‑cliff withdrawals of extra tokens** (balance above `amountGranted`). That contradicts “withdrawals are permitted only at the 6 month mark”.
> 
> Once you fix the logic, you should align docs to something like:
> 
> * “Grant tokens are fully unlocked 6 months after GRANT_TIMESTAMP.”
> * “Any tokens sent to this contract _in excess_ of the granted amount can be withdrawn immediately.”
> 
> Right now comments are misleading for reviewers and auditors.
> 
> ### 1.4. Grant accounting and `grant` openness
> `grant` is still:
> 
> ```solidity
> function grant(uint128 _amount) external nonReentrant {
>     require(_amount != 0, INVALID_PARAM());
>     amountGranted += _amount;
>     emit TokenGranted(_amount);
> 
>     IERC20(TAIKO_TOKEN).safeTransferFrom(
>         msg.sender, address(this), _amount
>     );
> }
> ```
> 
> ([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])
> 
> The new test:
> 
> ```solidity
> vm.prank(Carol);
> vm.expectRevert();
> target.grant(10 ether);
> ```
> 
> reverts only because Carol doesn’t have TKO or allowance, not because of an access control check. The comment in the test (`"revert: Ownable: caller is not the owner"`) is misleading.
> 
> Design‑wise, letting _anyone_ call `grant` is arguably fine (it lets other parties top up the grant), but then:
> 
> * `amountGranted` increases whenever **anyone** funds the contract via `grant`.
> * Combined with your cliff logic, this means third‑party donors can **lock** tokens for the recipient until 6 months elapse (since they increase `amountGranted`).
> 
> If that’s intentional, this is fine. If you only want the admin/owner to create official “grant” allocations, consider adding `onlyOwner` to `grant` and adjust the test to check the correct revert reason.
> 
> Either way, I’d update the test comment to reflect the actual reason it reverts.
> 
> ## 2. `DeploySimpleTokenUnlock.s.sol`
> **New change**
> 
> You added a zero‑recipient guard in the loop:
> 
> ```solidity
> for (uint256 i; i < recipients.length; i++) {
>     require(recipients[i] != address(0), "zero recipient");
>     address proxy = deploy({
>         name: "",
>         impl: SIMPLE_TOKEN_UNLOCK_IMPL,
>         data: abi.encodeCall(SimpleTokenUnlock.init, (OWNER, recipients[i]))
>     });
>     ...
> }
> ```
> 
> ([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])
> 
> This is a good, simple robustness improvement.
> 
> **Ongoing issues (unchanged since Brecht’s version)**
> 
> 1. **`SIMPLE_TOKEN_UNLOCK_IMPL` is still undefined**
>    The script still has:
>    ```solidity
>    // NOT YET DEPLOYED
>    // address public SIMPLE_TOKEN_UNLOCK_IMPL =
>    function run() external broadcast {
>        require(SIMPLE_TOKEN_UNLOCK_IMPL != address(0), "SIMPLE_TOKEN_UNLOCK_IMPL not set");
>        ...
>    }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    There is no definition of `SIMPLE_TOKEN_UNLOCK_IMPL` anywhere in the repo. The script will not compile and is unusable as is.([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/files)][1])
>    Recommended:
>    
>    * Either define it in this script:
>      ```solidity
>      address public constant SIMPLE_TOKEN_UNLOCK_IMPL =
>          0x...; // real deployed implementation
>      ```
>    * Or pass it via `BaseScript` / `vm.envAddress` and rename this script accordingly so it’s obvious how to set it.
> 2. **`name: ""` in `deploy`**
>    `deploy({ name: "", impl: ..., data: ... })` gives no identifier for the proxy in your deployment tooling. For maintenance, it’s nicer to give these a name like `"simple_token_unlock"` or `"simple_token_unlock_recipient"` (you already use `"simple_token_unlock"` in the tests).([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])
> 
> These are not security issues, but they affect operability and clarity.
> 
> ## 3. `GrantSimpleTokenUnlock.s.sol`
> **Change since Brecht’s review**
> 
> Previously:
> 
> ```solidity
> console2.log("total:", total / 1e18);
> require(tko.balanceOf(msg.sender) >= total, "insufficient TKO balance");
> 
> vm.startBroadcast();
> for (uint256 i; i < items.length; i++) {
>     uint128 grantAmount = uint128(items[i].grantAmount * 1e18);
>     tko.approve(items[i].proxy, grantAmount);
>     SimpleTokenUnlock(items[i].proxy).grant(grantAmount);
> }
> vm.stopBroadcast();
> ```
> 
> Now:
> 
> ```solidity
> console2.log("total:", total / 1e18);
> vm.startBroadcast();
> require(tko.balanceOf(msg.sender) >= total, "insufficient TKO balance");
> for (uint256 i; i < items.length; i++) {
>     uint128 grantAmount = uint128(items[i].grantAmount * 1e18);
>     tko.approve(items[i].proxy, grantAmount);
>     SimpleTokenUnlock(items[i].proxy).grant(grantAmount);
> }
> vm.stopBroadcast();
> ```
> 
> ([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])
> 
> This is precisely what Brecht flagged: outside the broadcast, `msg.sender` is the **script contract**, not the EOA. Inside a `vm.startBroadcast()` block, `msg.sender` becomes the configured sender account (EOA), as per the Foundry docs and discussion.([[getfoundry.sh](https://getfoundry.sh/reference/cheatcodes/start-broadcast/?utm_source=chatgpt.com)][3])
> 
> So this change:
> 
> * Fixes the balance check to correctly look at the EOA’s TKO balance, not the script’s.
> * Aligns the check with the account that actually does the `approve` and `grant` calls.
> 
> **Behaviorally & security‑wise:**
> 
> * `total` is computed as a sum of `SafeCastUpgradeable.toUint128(items[i].grantAmount * 1e18)`. If this overflows `uint128` it will revert, which is a nice safety guard.
> * After `vm.startBroadcast`, all calls (`balanceOf`, `approve`, `grant`) run as the same `msg.sender` EOA, so allowances and transfers line up correctly.
> 
> This change is correct and fixes the earlier concern.
> 
> **Small robustness / simplicity suggestions**
> 
> Not strictly required, but would make it clearer:
> 
> 1. Explicitly choose the broadcaster:
>    ```solidity
>    address deployer = vm.envAddress("GRANT_ADMIN");
>    vm.startBroadcast(deployer);
>    require(tko.balanceOf(deployer) >= total, "insufficient TKO balance");
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    This makes the code’s intent obvious and avoids relying on default broadcast semantics.
> 2. Consider using `SafeERC20` in the script too, to mirror the contract’s use of `SafeERC20` and fail more gracefully on non‑standard ERC‑20s.
> 
> ## 4. `SimpleTokenUnlock.t.sol` (new tests)
> You added a new dedicated test suite:([[GitHub](https://github.com/taikoxyz/taiko-mono/pull/20804/commits/4eec5084d31f833b843b387cb115027aa2147304)][2])
> 
> * Uses a local `MyERC20` (ERC20Votes) and `vm.etch` / `vm.store` to stand in for the real TKO at the canonical address.
> * Deploys `SimpleTokenUnlock` via the same `deploy()` helper as the rest of the protocol.
> * Registers `"taiko_token"` and `"taiko"` in the resolver, so the contract behaves more realistically.
> 
> **The tests:**
> 
> 1. `test_simpletokenunlock_single_vest_withdrawal`
>    
>    * Checks:
>      
>      * Only Alice can successfully call `grant` (via ERC‑20 balance/allowance, not access control).
>      * Extra tokens sent directly to the contract (`taikoToken.transfer(address(target), 0.5 ether)`) are immediately reflected in `amountWithdrawable` pre‑cliff.
>      * Pre‑cliff, `amountWithdrawable` is constant across time warps.
>      * Post‑cliff, the full balance is withdrawable.
>    * This encodes the new desired semantics (grant locked until cliff; extra tokens free).
>    
>    **Issue:** As mentioned above, it **never calls `withdraw`**, so it doesn’t test the actual withdrawal path and doesn’t catch the critical accounting bug.
> 2. `test_simpletokenunlock_delegate`
>    
>    * Verifies:
>      
>      * `grant` funds the contract.
>      * `amountWithdrawable` is zero pre‑cliff when there are no extra tokens.
>      * Delegation can be changed by the recipient, and that `taikoToken.delegates(address(target))` equals the delegatee.
>    * This aligns with ERC20Votes semantics (the contract address is the token holder; it delegates to the chosen address).
> 3. `test_simpletokenunlock_delegate_change_recipient`
>    
>    * Same as above but also:
>      
>      * Allows recipient to change `recipient` to `David`.
>      * Keeps the existing delegation (`delegates(address(target)) == Carol`) stable across `changeRecipient`.
>    
>    This is good: it demonstrates that governance voting power follows the contract, not the `recipient` field, which is exactly what you want.
> 
> **Recommendations for test coverage**
> 
> To really make this feature robust, I’d add at least:
> 
> * A test that calls `withdraw` pre‑cliff **twice+** and asserts that `amountGranted` remains unchanged and only the extra tokens are withdrawn (this will fail right now).
> * A test that after the cliff, multiple `withdraw` calls cannot withdraw more than the total deposited (grants + extras).
> * A test that:
>   
>   * Owner can call `withdraw` (using `onlyRecipientOrOwner`).
>   * A random address cannot (reverts with `PERMISSION_DENIED()`).
> 
> ## 5. Summary of what’s fixed vs still open
> **Fixed / improved since Brecht’s review**
> 
> * `DeploySimpleTokenUnlock`: guard against zero recipients in the loop.
> * `GrantSimpleTokenUnlock`: `tko.balanceOf(msg.sender)` check moved into the `vm.startBroadcast()` block, so it uses the broadcaster EOA instead of the script contract.
> * `SimpleTokenUnlock.init`: added `nonZeroAddr(_owner)` to match `_recipient` and preserve invariants.
> * New tests capture the intended “extra tokens withdrawable pre‑cliff, grants post‑cliff” behavior and delegation behavior.
> 
> **New / remaining issues**
> 
> 1. **High‑severity bug in `withdraw` + `amountWithdrawable`**
>    
>    * Pre‑cliff, repeated calls to `withdraw(amountWithdrawable())` can drain the entire grant over many calls, completely bypassing the cliff.
>    * Fix requires changing how `grantedReduction` is computed (see section 1.1).
> 2. **Docs / NatSpec are now inconsistent with behavior**
>    
>    * Comments still claim withdrawals “only at the 6 month mark” while tests and `amountWithdrawable` allow pre‑cliff withdrawals of extra tokens.
>    * Should be updated once logic is finalized.
> 3. **`DeploySimpleTokenUnlock` script still non‑functional**
>    
>    * `SIMPLE_TOKEN_UNLOCK_IMPL` remains undefined anywhere in the repo.
>    * `name: ""` is unhelpful for deployment tooling.
> 4. **Test suite doesn’t exercise `withdraw` yet**
>    
>    * Tests only read `amountWithdrawable`; they don’t assert correct grant accounting and invariants along the withdrawal path.
> 5. **`grant` access pattern is a bit opaque**
>    
>    * Behavior is “anyone can grant if they’re willing to transfer tokens into the contract”.
>    * This _might_ be intended; if so, test comments should be updated to reflect the real reason for the revert.
> 
> If you fix the pre‑cliff withdrawal accounting and add tests that enforce the intended invariants, this will be much more robust and much easier to reason about in an audit.

Thanks @Brechtpd ! I've applied the review comments, will definitely use 5.1 to review new changes going forward

dantaik:
@RogerLamTd lets fix these small issues:


- [x] Missing nonReentrant on changeRecipient (Medium)

Location: SimpleTokenUnlock.sol:137

The changeRecipient function lacks the nonReentrant modifier. While the function doesn't directly transfer tokens, changing the recipient could be exploited in combination with other operations if reentrancy is possible through external calls elsewhere.

- [ ] No Access Control on grant() Function (Low/Info)

Location: SimpleTokenUnlock.sol:90

Anyone can call grant() to deposit tokens. While this isn't necessarily a vulnerability (since it increases locked tokens), it could be used to:

Artificially inflate amountGranted, locking more tokens than intended by the original grantor
Create accounting confusion
The test at line 80 expects a revert from Carol calling grant(), but the contract doesn't actually restrict this. The test passes only because Carol doesn't have token approval, not because of access control.


- [x] Voting Delegation Not Updated on Recipient Change (Medium)

Location: SimpleTokenUnlock.sol:137-142

When changeRecipient is called, the voting delegation (ERC20VotesUpgradeable.delegate) is NOT updated. The new recipient does not automatically receive voting rights - they must manually call delegate() afterward. This could lead to:

Stale voting delegation to old recipient
Confusion about who controls votes

- [ ]  Missing Input Validation for _to in withdraw (Low)

Location: SimpleTokenUnlock.sol:109

While _to = recipient fallback exists for address(0), there's no check preventing withdrawal to malicious/unintended addresses if the recipient mistypes or is tricked into using a wrong address.


- [x]  Inconsistent Modifier Usage

withdraw: has nonReentrant
grant: has nonReentrant
delegate: has nonReentrant
changeRecipient: missing nonReentrant

dantaik:
## Deployment Scripts Review

### 1. Deploy.s.sol Issues

#### **1.1 Empty Name Parameter Causes JSON Write Failure** (High)
**Location:** `Deploy.s.sol:23-27`

```solidity
address proxy = deploy({
    name: "",
    impl: SIMPLE_TOKEN_UNLOCK_IMPL,
    data: abi.encodeCall(SimpleTokenUnlock.init, (OWNER, recipients[i]))
});
```

The `deploy()` function in BaseScript (line 57-60) conditionally registers the address when `name != ""`. While this skips registration, it still writes to JSON using:
```solidity
string memory _name = Strings.toString(uint256(name));
```
Empty bytes32 (`""`) converts to `"0"`, meaning all 22 proxies will overwrite the same JSON key, losing all but the last deployment address.

#### **1.2 No Duplicate Recipient Check** (Medium)
**Location:** `Deploy.s.sol:21-30`

The loop doesn't verify that `recipients[i]` isn't already in a previous iteration. Deploying multiple contracts for the same recipient would be wasteful and could cause confusion.

#### **1.3 Hardcoded Implementation Address** (Info)
**Location:** `Deploy.s.sol:11`

`SIMPLE_TOKEN_UNLOCK_IMPL` is hardcoded. If the implementation needs updating before deployment, the script must be modified. Consider using an environment variable instead.

---

### 2. Grant.s.sol Issues

#### **2.1 Wrong JSON Path** (Critical)
**Location:** `Grant.s.sol:22`

```solidity
string memory path = "/script/simpletokenunlock/Grant.data.json";
```

But the actual file is at:
```
/script/layer1/team/simpletokenunlock/Grant.data.json
```

This script will **fail to execute** because the file won't be found.

#### **2.2 Empty Proxy Addresses in Grant.data.json** (Critical)
**Location:** `Grant.data.json:3,8,13,...`

All `proxy` fields are empty strings:
```json
"proxy": "",
```

The script at line 38 does `SimpleTokenUnlock(proxy)` which will call `address(0)`, causing the transaction to fail or behave unexpectedly. The proxy addresses must be filled in after `Deploy.s.sol` runs.

#### **2.3 Inefficient Approval Pattern** (Low)
**Location:** `Grant.s.sol:53-54`

```solidity
tko.approve(items[i].proxy, grantAmount);
SimpleTokenUnlock(items[i].proxy).grant(grantAmount);
```

Calling `approve` for each proxy in a loop is gas-inefficient. The script approves the exact amount, so any approval reverts if there's any issue. Consider using `type(uint256).max` approval once or using `safeIncreaseAllowance`.

#### **2.4 Potential Truncation in SafeCast** (Low)
**Location:** `Grant.s.sol:43,52`

```solidity
total += SafeCastUpgradeable.toUint128(items[i].grantAmount * 1e18);
```

If `grantAmount * 1e18` exceeds `type(uint128).max` (~340 billion tokens), `SafeCast` will revert. The largest single grant is 500,000 TKO which is safe, but there's no explicit check that the sum fits in uint128 (line 43 adds to uint256 `total`, but line 52 casts to uint128).

#### **2.5 No Validation That Proxies Match Recipients Order** (Medium)
The script relies on proxy addresses matching the order of recipients in the JSON. If the data is manually filled incorrectly, grants could go to wrong contracts.

---

### 3. ProposifyDeploy.s.sol Issues

#### **3.1 Hardcoded Transaction Count** (Medium)
**Location:** `ProposifyDeploy.s.sol:34`

```solidity
require(count == 22, "does not match number of recipients");
```

This hardcoded `22` must match `Deploy.data.json` which has 22 addresses. If recipients are added/removed, both files must be updated manually.

#### **3.2 No Validation of Broadcast Path Existence** (Low)
**Location:** `ProposifyDeploy.s.sol:28`

If `Deploy.s.sol` dry-run hasn't been executed, the file won't exist and the script will fail with an unclear error.

#### **3.3 View Function Cannot Execute Transactions** (Info)
**Location:** `ProposifyDeploy.s.sol:27`

The function is `view` (correct for generating calldata), but ensure operators understand this just generates data and doesn't execute anything.

---

### 4. Data Files Issues

#### **4.1 Mixed Case Addresses** (Low)
**Location:** `Deploy.data.json` and `Grant.data.json`

Addresses use mixed checksumming (some lowercase, some mixed case). While Solidity handles this, it can cause issues with JSON parsing or string comparisons.

#### **4.2 Grant.data.json Not Ready for Use** (Critical)
All proxy addresses are empty strings. This file is a template that must be populated after deployment.

---

## Summary Table

| File | Severity | Issue |
|------|----------|-------|
| Grant.s.sol | **Critical** | Wrong JSON path (`/script/simpletokenunlock/` instead of `/script/layer1/team/simpletokenunlock/`) |
| Grant.data.json | **Critical** | Empty proxy addresses - script will fail |
| Deploy.s.sol | High | Empty name causes all deployments to overwrite same JSON key |
| Deploy.s.sol | Medium | No duplicate recipient check |
| Grant.s.sol | Medium | No validation that proxy order matches recipient order |
| ProposifyDeploy.s.sol | Medium | Hardcoded transaction count (22) |
| Grant.s.sol | Low | Inefficient per-proxy approvals |
| Grant.s.sol | Low | Potential truncation in SafeCast |
| ProposifyDeploy.s.sol | Low | No validation broadcast file exists |
| Data files | Low | Mixed case addresses |
| Deploy.s.sol | Info | Hardcoded implementation address |

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,173 @@
...
+/// Tokens granted off-chain are deposited into this contract directly from the `msg.sender`
+/// address. Token withdrawals from the grant bucket are permitted only at the 6 month mark.
+/// Any tokens sent directly to the contract in excess of amountGranted are immediately withdrawable.
+/// A separate instance of this contract is deployed for each recipient.
+/// @custom:security-contact security@taiko.xyz
+contract SimpleTokenUnlock is EssentialContract {
+    using SafeERC20 for IERC20;
+    using LibMath for uint256;
+
+    uint256 public immutable SIX_MONTHS; // 183 days
```
dantaik: use constant not immutable here

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,173 @@
...
+        _;
+    }
+
+    modifier onlyRecipientOrOwner() {
+        require((msg.sender == recipient || msg.sender == owner()), PERMISSION_DENIED());
+        _;
+    }
+
+    constructor(address _resolver) EssentialContract(_resolver) {
+        SIX_MONTHS = 183 days;
```
dantaik: use immutable if you want to pass the value from the constructor, otherwise, use constant

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,173 @@
...
+    /// @param _amount The amount of tokens to withdraw.
+    function withdraw(address _to, uint256 _amount)
+    external
+    onlyRecipient
+    nonReentrant
+{
+    if (_to == address(0)) _to = recipient;
+    if (_amount == 0) _amount = amountWithdrawable();
+
+    uint256 available = amountWithdrawable();
```
dantaik: why call `amountWithdrawable` twice?

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,173 @@
...
+    require(_amount <= available, NOT_WITHDRAWABLE());
+
+    IERC20 tko = IERC20(TAIKO_TOKEN);
+    uint256 balance = tko.balanceOf(address(this));
+
+    // extra = any balance above the grant bucket
+    uint256 extra = balance > amountGranted ? balance - amountGranted : 0;
```
dantaik: I'd ignore this extra part to make the contract simpler. No one will send extra tokens to this contract if we don't handle it.

dantaik (changes requested):

.github/workflows/repo--typo-check.yml
```diff
@@ -1,6 +1,9 @@
 name: Typo Check
 
-on: [pull_request]
+on:
+  pull_request:
+    branches:
```
dantaik, line 5: better to have a separate PR

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,145 @@
...
+        IERC20(TAIKO_TOKEN).safeTransferFrom(msg.sender, address(this), _amount);
+    }
+
+    /// @notice Withdraws tokens by the recipient.
+    /// @param _to The address the token will be sent to.
+    /// @param _amount The amount of tokens to withdraw.
+    function withdraw(address _to, uint256 _amount) external onlyRecipient nonReentrant {
+        if (_to == address(0)) _to = recipient;
+
+        uint256 withdrawable = amountWithdrawable();
```
dantaik: I think defining a `canWithdraw() returns bool` is better than having a `amountWithdrawable`. Then once the token can be withdrawn, withdraw all of them. Partial withdrawal doesn't make sense.

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,145 @@
...
+
+        // pre-cliff: no tokens
+        if (block.timestamp < GRANT_TIMESTAMP + SIX_MONTHS) {
+            return 0;
+        }
+        // post-cliff: all tokens in the contract are withdrawable
+        return balance;
+    }
+
+    function _getAmountLocked() internal view returns (uint256) {
```
dantaik: seems this function is not used.

dantaik (changes requested):

.github/workflows/repo--typo-check.yml
RogerLamTd: This was just to disable the workflow on this branch to allow CI, I can move it to a separate PR later!

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,130 @@
...
+        require(msg.sender == recipient, PERMISSION_DENIED());
+        _;
+    }
+
+    modifier onlyRecipientOrOwner() {
+        require((msg.sender == recipient || msg.sender == owner()), PERMISSION_DENIED());
+        _;
+    }
+
+    constructor(address _resolver) EssentialContract(_resolver) { }
```
davidtaikocha: I see we have `address public constant TAIKO_TOKEN = 0x10dea67478c5F8C5E2D90e5E9B26dBe60c54d800;` as a constant now, and this contract doesn't use `resolve()`, do we really need to pass an `address _resolver` to the constructor? If not, i think we can simplify to:
```solidity
constructor() EssentialContract(address(0)) { }
```
like [here](https://github.com/taikoxyz/taiko-mono/blob/simpletokenunlock/packages/protocol/contracts/layer1/forced-inclusion/TaikoWrapper.sol#L66).

packages/protocol/contracts/layer1/team/SimpleTokenUnlock.sol
```diff
@@ -0,0 +1,130 @@
...
+    /// @notice Withdraws tokens to an address or recipient.
+    /// @param _to The address the token will be sent to.
+    function withdraw(address _to) external onlyRecipient nonReentrant {
+        if (_to == address(0)) _to = recipient;
+        uint256 withdrawable = canWithdraw();
+        require(withdrawable > 0, NOT_WITHDRAWABLE());
+
+        IERC20 tko = IERC20(TAIKO_TOKEN);
+        emit TokenWithdrawn(_to, withdrawable);
+        tko.safeTransfer(_to, withdrawable);
```
davidtaikocha, line 102: Do we still need to reset `amountGranted` to `0` here? Since it looks like `grant` is allowed to be called multiple times?

davidtaikocha:
@claude plz review.

davidtaikocha (approved):

packages/protocol/script/layer1/team/simpletokenunlock/Deploy.s.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import "src/layer1/team/SimpleTokenUnlock.sol";
+import "script/BaseScript.sol";
+
+contract DeploySimpleTokenUnlock is BaseScript {
+    using stdJson for string;
+
+    address public OWNER = 0x9CBeE534B5D8a6280e01a14844Ee8aF350399C7F; // admin.taiko.eth
```
dantaik, line 10: correct owner

dantaik:
Roger, maybe we don't want to merge this contract into our main branch. If you stil prefer to check it in, please merge it to https://github.com/taikoxyz/taiko-mono/tree/main/packages/supplementary-contracts

RogerLamTd:
> Roger, maybe we don't want to merge this contract into our main branch. If you stil prefer to check it in, please merge it to https://github.com/taikoxyz/taiko-mono/tree/main/packages/supplementary-contracts

I believe it depends on libraries that no longer exist on main, we can just close the branch and not merge it!

PR #20807, opened 2025-11-22
fix(balance-monitor): separate L1 and L2 ERC-20 balance metrics

anim001k:
The `l1_erc20_balance` metric was incorrectly aggregating ERC-20 token balances from both L1 and L2 networks, making it impossible to monitor L1 balances independently.


davidtaikocha:
We don't need to monitor L2 ERC-20 balance.

PR #20809, opened 2025-11-24
feat(taiko-client-rs): wire finalized Shasta proposal into forkchoice updates

davidtaikocha:
- plumb `last_finalized_proposal_id` from proposal core state into bundle metadata
  - resolve finalized block hash via `last_block_id_by_batch_id` and use it when updating forkchoice `safe_block_hash`/`finalized_block_hash`, falling back to `zero` if unavailable
  - keep payload application non-failing when the mapping/header lookup is missing


YoGhurt111 (approved):

MatusKysel (approved):
lgtm

PR #20825, opened 2025-11-24
fix(protocol): disable forced inclusions until after the first proposal

ggonzalez94:
It is in theory possible that a forced inclusion block can be the first block after fork (even the first block after genesis if shasta time is 0). If so, the rule of its anchor number has some issues. This PR fixes it by not allowing forced inclusions before the first proposal(not counting activation).
The cost of doing a forced inclusion goes up slightly, but since this is not part of the path we wanna optimize for this should be fine.

image


davidtaikocha (approved):

dantaik (approved):

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -321,7 +321,12 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     }
 
     /// @inheritdoc IForcedInclusionStore
+    /// @dev This function will revert if called before the first non-activation proposal is submitted
+    /// to make sure blocks have been produced already and the derivation can use the parent's block timestamp.
     function saveForcedInclusion(LibBlobs.BlobReference memory _blobReference) external payable {
+        bytes32 proposalHash = _proposalHashes[1];
```
dantaik, line 327: `_proposalHashes[1]` involves a SLOAD, is it possible to use the `activationTimestamp` immutable to avoid reading storage?

dantaik: After a second thought, this change is OK -- we don't encourage people to use forced inclusion, so this function being not optimized is actually fine.

dantaik (approved):

PR #20826, opened 2025-11-24
feat(relayer): use the new signal service

cyberhorsey:
also adds a new flag to pause processing around the shastaForktimestamp.

### DIFF
--- a/packages/relayer/indexer/indexer.go
+++ b/packages/relayer/indexer/indexer.go
@@ -93,7 +93,7 @@ type Indexer struct {
 	destBridge relayer.Bridge
 
 	signalService   relayer.ChainDataSignalService
-	signalServiceV2 *v4.SignalService
+	signalServiceV4 *v4.SignalService
 
 	blockBatchSize      uint64
 	numGoroutines       int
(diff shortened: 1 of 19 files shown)


packages/relayer/indexer/indexer.go
```diff
@@ -430,6 +444,13 @@ func (i *Indexer) filter(ctx context.Context) error {
 					slog.Error("i.indexChainDataSyncedEvents", "error", err)
```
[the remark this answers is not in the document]
cyberhorsey: SrcSignalService will always be set, and then removed after the Shasta transition to only have one input.

davidtaikocha (approved):

packages/relayer/indexer/indexer.go
```diff
@@ -91,7 +92,8 @@ type Indexer struct {
 	bridge     relayer.Bridge
 	destBridge relayer.Bridge
 
-	signalService relayer.SignalService
+	signalService   relayer.ChainDataSignalService
+	signalServiceV2 *v4.SignalService
```
MatusKysel: I know the version is of the bindings, but it's kind of confusing to use v2 if the bindings are v4, maybe `signalServiceV4` would be better name. WDYT?

packages/relayer/indexer/indexer.go
```diff
@@ -202,7 +204,9 @@ func InitFromConfig(ctx context.Context, i *Indexer, cfg *Config) (err error) {
 		}
 	}
 
-	var signalService relayer.SignalService
+	var signalService relayer.ChainDataSignalService
+
+	var signalServiceV2 *v4.SignalService
```
MatusKysel: same here

MatusKysel (approved):

packages/relayer/indexer/indexer.go
cyberhorsey: yeah definitely, i will update this. and we will just delete the old one after the shasta transition so it will come back to just being called `signalService`. so i will make these changes

RogerLamTd (approved):

PR #20827, opened 2025-11-24
fix(taiko-client-rs): align pool and manifest gas limits with L2 head and anchor budget

VolodymyrBg:
This change fixes a gas limit mismatch in the proposer. Previously the tx pool fetch constrained pre-built tx lists using MIN_BLOCK_GAS_LIMIT while the manifest always set MAX_BLOCK_GAS_LIMIT, which conflicts with Shasta’s gas limit validation (ppm delta bounds and clamping) and can cause the manifest to default. We now use the latest L2 head’s gas limit when requesting pool content so the EE builds lists against the correct block capacity, and set BlockManifest.gas_limit from the same L2 head’s gas limit minus the anchor gas budget (ANCHOR_V3_GAS_LIMIT). This mirrors the reference Go client and matches the Derivation spec: the manifest excludes anchor gas and the driver adds it back when assembling payload metadata. The result is consistent, protocol-compliant behavior, better pool filling, and avoidance of defaulted manifests during derivation.

### DIFF
--- a/packages/taiko-client-rs/crates/proposer/src/proposer.rs
+++ b/packages/taiko-client-rs/crates/proposer/src/proposer.rs
@@ -11,7 +11,7 @@ use alloy::{
 use alloy_network::TransactionBuilder;
 use event_indexer::indexer::{ShastaEventIndexer, ShastaEventIndexerConfig};
 use metrics::{counter, gauge, histogram};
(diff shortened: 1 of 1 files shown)


davidtaikocha:
It's ok to use `MIN_BLOCK_GAS_LIMIT ` for that devnet Shasta only proposer.

PR #20830, opened 2025-11-25
Enforce forced inclusion activation delay

dantaik:
Adds a 24-hour activation delay for forced inclusion to prevent immediate usage after the first proposal. This ensures blocks have been produced and the derivation pipeline can safely use parent block timestamps.

Changes the saveForcedInclusion function validation logic and renames the ACTIVATION_WINDOW constant to follow private naming conventions.

🤖 Generated with Claude Code

### DIFF
--- a/packages/protocol/contracts/layer1/core/impl/Inbox.sol
+++ b/packages/protocol/contracts/layer1/core/impl/Inbox.sol
@@ -40,7 +40,9 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     // ---------------------------------------------------------------
     // Constants
     // ---------------------------------------------------------------
-    uint256 private constant ACTIVATION_WINDOW = 2 hours;
+    uint256 private constant _ACTIVATION_WINDOW = 2 hours;
+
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -324,8 +326,9 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     /// @dev This function will revert if called before the first non-activation proposal is submitted
     /// to make sure blocks have been produced already and the derivation can use the parent's block timestamp.
     function saveForcedInclusion(LibBlobs.BlobReference memory _blobReference) external payable {
-        bytes32 proposalHash = _proposalHashes[1];
-        require(proposalHash != bytes32(0), IncorrectProposalCount());
+        unchecked {
+        require(block.timestamp > activationTimestamp + _FORCE_INCLUSION_ACTIVATION_DELAY, ForceInclusionNotActivated());
+        }
```
davidtaikocha, line 331: Shall we also require `activationTimestamp > 0` here?

PR #20833, opened 2025-11-25
fix(relayer): validate chain IDs are non-negative uint64 in GetBlockInfo

Snezhkko:
This change hardens GetBlockInfo by rejecting negative or out-of-range chain IDs before converting big.Int to uint64. EIP-155 defines chainId as an unsigned integer, so accepting negative values is invalid. Previously, negative inputs would flow to Uint64(), which is undefined for values that cannot be represented as uint64 and, in practice, can coerce -1 to 1, causing silent misrouting of DB queries and misleading responses. The handler now validates parsed srcChainID and destChainID from query parameters using Sign() and IsUint64(), returning 422 for invalid input, and performs a final guard after resolving IDs (including RPC-derived values) to ensure only valid non-negative uint64 values are used. This eliminates undefined conversions, aligns behavior with EIP-155 semantics, and prevents incorrect repository lookups.


davidtaikocha:
Don't think its really necessary. They are loaded from the configs.

PR #20835, opened 2025-11-25
fix(protocol): fix chain continuity violation

ggonzalez94:
## Issue
`InboxOptimized1._buildAndSaveAggregatedTransitionRecords` aggregated consecutive proposal proofs without verifying that each transition’s `parentTransitionHash` matched the hash of the previously aggregated transition. An attacker could submit transitions `[B(parent=A), C(parent=X)]` with consecutive proposal IDs. The contract would store a single aggregated record keyed by `(id(B), A)` but whose `transitionHash` pointed to `hash(C)` and `span=2`. During finalization the inbox would advance directly from `A` to `hash(C)`, effectively “teleporting” to a fork rooted at `X`.


dantaik (approved):

dantaik:
From AI:

The fix is correct. However, there's one thing missing from the PR that I discovered:

The existing test helper functions use keccak256(abi.encode(transitions[i])) to compute transition hashes, but the actual codec implementations (CodecOptimized vs CodecSimple) use different hash functions. This causes some tests to fail for InboxOptimized2 (which uses CodecOptimized) because the hash mismatch prevents aggregation.

Your teammate should also update:

- AbstractProve.t.sol:_createProveInputForMultipleProposals - line 420/432 should use _codec().hashTransition(transitions[i])
- AbstractFinalize.t.sol:_proveConsecutiveProposalsAndGetRecords - line 565 should use _codec().hashTransition(transitions[i])

Without these test fixes, the existing tests for InboxOptimized2 will fail after the fix is applied.

davidtaikocha (approved):

dantaik:
This PR won't be needed if we accept https://github.com/taikoxyz/taiko-mono/pull/20850

ggonzalez94:
closing this as we've decided to go with sequential proving #20884

PR #20836, opened 2025-11-25
fix(protocol): fix innefective conflict handling allows finalization of conflicted transitions

ggonzalez94:
# Ineffective Conflict Handling Allows Finalization of Conflicted Transitions

The conflict handling mechanism used between **InboxOptimized1** and **Inbox** fails to prevent finalization of transitions that have been explicitly flagged as conflicted. When `InboxOptimized1` detects a hash collision for the same proposal and parent (i.e., same key in the ring buffer but different transition data), it overwrites the slot with the new transition and sets its `finalizationDeadline` to the maximum value, intending to mark the entry as permanently non-finalizable.

The enforcement of this flag is delegated to the `_finalize` logic in **Inbox**. However, `_finalize` only considers the `finalizationDeadline` in the path where no transition data is supplied by the caller (implicit finalization). When the caller explicitly provides a `TransitionRecord` in the calldata, `_finalize` only checks that the supplied record matches the stored hash and immediately updates the chain state, ignoring the stored `finalizationDeadline`. As a result, the “never finalize” marker applied to conflicting transitions is not honored in the explicit finalization path.

An attacker can exploit this mismatch as follows:

1. A legitimate prover stores a valid transition for a given proposal and parent state.
2. A malicious prover submits a conflicting transition with the same proposal and parent, causing `InboxOptimized1` to overwrite the ring buffer slot and set its deadline to the maximum value.
3. A proposer, acting in coordination with the malicious prover, then calls the propose entrypoint and includes this conflicting transition in the list of transition records passed as input.
4. During finalization, the contract observes that the supplied record matches the stored hash and finalizes it, never checking the stored `finalizationDeadline`.

The conflicting transition becomes canonical, and the original valid transition is effectively discarded.

## Storage Layout
The storage layout should still be fine, since this occupied slot 251(packed with `activationTimestamp`) and the next variable occupies slot 252.


dantaik (approved):

davidtaikocha (approved):

dantaik:
This PR won't be needed if we accept https://github.com/taikoxyz/taiko-mono/pull/20850

ggonzalez94:
closing this as we've decided to go with sequential proving #20884

PR #20839, opened 2025-11-26
refactor(protocol): additional transition conflict handling in finalization

dantaik:
@ggonzalez94 I suggest we make the following additional changes on top of your PR:

- When attempting to finalize a conflicting transaction, do not revert; simply stop processing at that point.
- Remove TransitionDuplicateDetected.
- Add four fields to TransitionConflictDetected to assist with further diagnostics.


ggonzalez94 (approved):
LGTM. Just did some slight refactor in the `_storeTransitionRecord` function so that it returns early instead of having more else if.

PR #20840, opened 2025-11-26
feat(protocol): introduce `bondProcessingProposalHash` in `TransitionMetadata`

davidtaikocha:
To solve the proving issue raised by @smtmfft : https://www.notion.so/taikoxyz/Bond-Processing-in-Prover-2b69673143d680718715cc4fb8b6315d , and avoid changing the current protocol / derivation consensus.

## Summary

Add delayed bond-processing hash to metadata and enforce validation in `Inbox`.

## Description

- expose `BOND_PROCESSING_DELAY` in `LibManifest` and carry `bondProcessingProposalHash` through `TransitionMetadata` encoding/decoding and hashing (simple/optimized).
- in Inbox prove flow, require each provided bond-processing proposal hash to match the stored hash of `id - BOND_PROCESSING_DELAY`, reverting with `BondProcessingProposalHashMismatch` otherwise; keep hashing unchanged.

## TODO

Fix tests if it looks good to the reviewers.

### DIFF
--- a/packages/protocol/contracts/layer1/core/iface/IInbox.sol
+++ b/packages/protocol/contracts/layer1/core/iface/IInbox.sol
@@ -110,6 +110,8 @@ interface IInbox {
         address designatedProver;
         /// @notice The actual prover who submitted the proof.
         address actualProver;
+        /// @notice Hash of the proposal at `id - BOND_PROCESSING_DELAY` used for bond processing.
+        bytes32 bondProcessingProposalHash;
     }
 
     /// @notice Represents a record of a transition with additional metadata.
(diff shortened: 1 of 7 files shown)


dantaik:
I need some time to think about it ...

smtmfft (approved):

ggonzalez94 (commented):
I do not have a better idea yet for how to solve this, but a few points I'm concerned about if we go with this approach:
- It increases the cost of proving, and we add one additional `SLOAD` per transition being proven, which goes against one of our existing principles(avoid costs that scale linearly with the amount of blocks, proofs, etc).
- The contract size becomes  23,830 bytes, which gives us a margin of only 746. I'm a bit hesitant to keep adding code to the inbox contracts.

@smtmfft @davidtaikocha I would encourage you to keep thinking about other possible solutions, even if it complicates the prover(within a reasonable complexity). If we cannot find a better approach, we'll merge this.

dantaik (commented):
I think I have figured out a better solution without extra SLOAD. Will create a PR over the weekend.

dantaik:
See https://github.com/taikoxyz/taiko-mono/pull/20868

davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/pull/20868

PR #20801, opened 2025-11-21
fix(protocol): InboxOptimized1's ProvedEventPayload has last aggregated transition's checkpoint

dantaik:
## Summary
Modified InboxOptimized1 to emit ProvedEventPayload with the last transition's checkpoint instead of the first when aggregating consecutive proposals. This ensures the Proved event correctly reflects the final L2 state of the aggregated span. Added comprehensive batch prove comparison test between Inbox and InboxOptimized1 implementations.

## Changes
- Modified `_buildAndSaveAggregatedTransitionRecords` to construct aggregated transitions with the first proposal's hash but the last transition's checkpoint
- Added `InboxOutputComparison.t.sol` test demonstrating the difference between individual proofs and batch aggregated proofs
- Updated gas snapshots for affected prove operations

## Testing
- All 52 existing inbox prove tests pass
- New comparison test verifies batch prove outputs and payload formatting
- Gas costs updated to reflect the new logic

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #20843, opened 2025-11-26
fix(protocol): avoid mutating prove input

ggonzalez94:
Ref: https://github.com/taikoxyz/taiko-mono/pull/20801#discussion_r2565523431

## Summary
- Fix aggregated prove path to avoid mutating the provided transitions before hashing, which previously replaced the first transition’s checkpoint with the last one and broke aggregated proving hashes.
- Emit/store aggregated checkpoints using a local copy of the transition, leaving the original prove input untouched so the on-chain aggregatedProvingHash matches off-chain proofs.
- Add a targeted test verifier and coverage to ensure the aggregated proving hash is computed over unmodified transitions.

## Motivation
- For aggregated spans, the contract was mutating `_input.transitions[_firstIndex].checkpoint` in place before hashing. Off-chain proofs built over the original transitions no longer matched the on-chain aggregatedProvingHash, causing verification failures and liveness risk.
- The change preserves hash correctness while keeping the optimized event payload behavior for aggregated spans.


dantaik:
What do you think about fixing this by moving the `aggregatedProvingHash` calculation above the `_buildAndSaveTransitionRecords` call? This should let us avoid the additional memory copy.

Pleas see https://github.com/taikoxyz/taiko-mono/pull/20847

ggonzalez94:
> What do you think about fixing this by moving the aggregatedProvingHash calculation above the _buildAndSaveTransitionRecords call? This should let us avoid the additional memory copy.

While in theory this should be more efficient, I think modifying reference variables inside a function is always dangerous for maintenance(makes the code harder to reason about), and makes it more likely that we introduce a bug in the future. Also in terms of gas benches, the compiler seems to be doing a better with allocating this memory variable because the gas cost are slightly lower if I checked correctly.
Anyways, I'll first review #20850 and if we can simplify this whole logic I believe it is much better

PR #20850, opened 2025-11-27
fix(protocol): remove transition aggregation and conflict detection

### LINKED ISSUE #20851: Bug: Finalization DoS due to Forced Record Hash Mismatch
A bug was reported by Sebastian Fabry as follows:

----

The **InboxOptimized1** contract reduces gas costs by allowing provers to aggregate transition records for consecutive proposals into a single submission. To ensure state correctness, the system includes a conflict detection mechanism that sets the finalization deadline to the maximum integer value `(type(uint48).max)` if a submitted transition record hash differs from an existing one for the same proposal and parent. This mechanism is designed to freeze finalization if contradictory state transitions are detected.

However, a valid aggregated transition record and a valid single-step transition record for the same starting proposal will inherently have different hashes. The aggregated record includes the transitionHash and checkpointHash of the final proposal in the batch and a larger span, whereas the single-step record uses the data from the first proposal and a span of 1. Consequently, the contract treats these two valid proofs as conflicting. Although the current finalization logic contains a separate defect that bypasses the deadline check, fixing that defect would expose this Denial of Service vector. The attack proceeds as follows:

1. An honest prover submits an aggregated proof for a sequence of proposals (e.g., `P1` through `P5`), storing a record hash `H_agg` derived from P5's end state.
1. An attacker submits a valid single-step proof for P1, generating a record hash H_single derived from P1's end state.
1. The contract compares `H_single` against `H_agg`.
1. Because the hashes differ, the contract flags a conflict and sets the finalization deadline to `type(uint48).max`.
1. Once the accompanying finalization logic is corrected, this action would permanently prevent the aggregated batch from finalizing.

Consider updating the conflict detection logic to support the coexistence of valid aggregated and single-step proofs. This could be achieved by verifying if the single-step transition is a consistent prefix of the stored aggregated transition before flagging a conflict. Alternatively, consider enforcing a canonical proving strategy (e.g., only allowing the longest known span) to prevent ambiguity between valid records.

dantaik:
## Summary

This PR proposes one possible way to fix the bugs reported in https://github.com/taikoxyz/taiko-mono/issues/20851:

- Remove the `span` field from the `TransitionRecord` struct and **disable transition aggregation** in `InboxOptimized1`
- Remove bond instruction merging functions (no longer necessary)
- **Remove conflicting transition detection** — the first transition is deemed correct and cannot be overwritten
- Remove `compositeKeyVersion`
- Refactor code in `Inbox.sol` and `InboxOptimized.sol` to use shared helper function(s)
- Add parentProposalHash to Proposal struct so proposals can linked up for easy verifications


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -458,106 +472,127 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+        bool _overwrittenByOwner
+    )
+        internal
+        returns (bool isOwnerSaved_, bool isDuplicate_, bool isConflicting_)
+    {
+        bytes26 existingRecordHash = _entry.recordHash;
+
+        if (existingRecordHash == 0) {
+            _entry.finalizationDeadline = _hashAndDeadline.finalizationDeadline;
+            _entry.recordHash = _hashAndDeadline.recordHash;
```
davidtaikocha: In some cases if there is a proving system related bug that leads to no proof can be generated / submitted, will we still create a `setTransitionWithoutProof` proposal to force-prove it?  If so, shall we check `_overwrittenByOwner` and return `isOwnerSaved_  = true` here?

dantaik: If there is a bug that has to be fixed using `setTransitionWithoutProof`, we need to get approval from Security Council. Inside `setTransitionWithoutProof`, `true` is passed into the inner function so isOwnerSaved_ will be true.

dantaik: Code updated

davidtaikocha (approved):

davidtaikocha (dismissed):

smtmfft (dismissed):

PR #20857, opened 2025-11-27
Fail fast when watchdog pause transactions revert

sashaodessa:
return explicit errors when mined pause receipts report failure for src/dest bridges so the watchdog can retry or alert instead of silently succeeding.


davidtaikocha:

PR #20867, opened 2025-11-28
fix(taiko-client-rs): rename proposal_id to batch_id

Forostovec:
Renamed the proposal_id parameter to batch_id in last_l1_origin_by_batch_id and last_block_id_by_batch_id within 
packages/taiko-client-rs/crates/rpc/src/l1_origin.rs to align with the execution engine’s public RPC method names taiko_lastL1OriginByBatchID and taiko_lastBlockIDByBatchID and to maintain consistency with 
packages/taiko-client-rs/crates/rpc/src/auth.rs where batch_id is already used for engine-facing operations. Although Shasta semantics use “proposal” as the domain term, the engine API retains the legacy “batch” terminology, and the rest of the codebase (including the Shasta derivation pipeline) already passes proposal IDs to these “batch” endpoints.

### DIFF
--- a/packages/taiko-client-rs/crates/rpc/src/l1_origin.rs
+++ b/packages/taiko-client-rs/crates/rpc/src/l1_origin.rs
@@ -56,22 +56,22 @@ impl<P: Provider + Clone> Client<P> {
     }
 
     /// Fetch the last L1 origin associated with the given batch id via the public engine API.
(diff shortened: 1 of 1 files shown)


davidtaikocha:
Its ok to use proposal_id in taiko-client-rs.

PR #20868, opened 2025-11-28
fix(protocol): add parentProposalHash to Proposal struct

dantaik:
## Summary

This PR introduces a cleaner fix for the [issue reported by @smtmfft ](https://www.notion.so/native/taikoxyz/Bond-Processing-in-Prover-2b69673143d680718715cc4fb8b6315d?deepLinkOpenNewTab=true).

With the newly added `parentProposalHash` field to the `Proposal` struct, all previous proposals verifiable with a new proposal.


packages/protocol/test/layer1/core/libs/LibBondInstruction.t.sol
```diff
@@ -128,7 +128,8 @@ contract LibBondInstructionTest is Test {
             endOfSubmissionWindowTimestamp: _timestamp + 1,
             proposer: PROPOSER,
             coreStateHash: bytes32(uint256(0x11)),
-            derivationHash: bytes32(uint256(0x22))
+            derivationHash: bytes32(uint256(0x22)),
+            parentProposalHash: bytes32(0)
```
dantaik:
```suggestion
            parentProposalHash: bytes32(uint256(0x33))
```

davidtaikocha (commented):
Changes LGTM and I feel like this can be a fix, but still need @smtmfft 's double-check.

davidtaikocha (approved):

smtmfft:
A bond is provable if there is a path from the MAX_OFFSET proposal to the current proposal.
The previous solution (#20840) tries to link these two proposals explicitly on L1.
The new approach links them implicitly, through a backward (parent-based) reference chain.

From the prover’s point of view:

-. Previous approach:
Requires only one hash, which proves that the bonds are included in the MAX_OFFSET proposal.
Downsides:
It changes the public input.
MAX_OFFSET becomes fixed and cannot be adjusted flexibly.

-. Current approach:
Requires MAX_OFFSET hashes, essentially verifying something similar to a Merkle-like proof, and therefore increases API data transfer.
Benefit: MAX_OFFSET can be adjusted dynamically, and the public input remains unchanged.

If bonds occur frequently and MAX_OFFSET does not change often, I think the previous approach is easier to handle.

dantaik:
> A bond is provable if there is a path from the MAX_OFFSET proposal to the current proposal. The previous solution (#20840) tries to link these two proposals explicitly on L1. The new approach links them implicitly, through a backward (parent-based) reference chain.
> 
> From the prover’s point of view:
> 
> -. Previous approach: Requires only one hash, which proves that the bonds are included in the MAX_OFFSET proposal. Downsides: It changes the public input. MAX_OFFSET becomes fixed and cannot be adjusted flexibly.
> 
> -. Current approach: Requires MAX_OFFSET hashes, essentially verifying something similar to a Merkle-like proof, and therefore increases API data transfer. Benefit: MAX_OFFSET can be adjusted dynamically, and the public input remains unchanged.
> 
> If bonds occur frequently and MAX_OFFSET does not change often, I think the previous approach is easier to handle.

The previous approach is more expensive. The new one doesnt' have significant gas cost. So lets use this new one.

PR #20870, opened 2025-11-28
fix(protocol): fix finalization DOS while preserving transition aggregation

### LINKED ISSUE #20851: Bug: Finalization DoS due to Forced Record Hash Mismatch

ggonzalez94:
Fixes #20851

## Summary
This is a long PR, but tackles several related issues. I'm pushing it while it is still a bit of a draft, so we can decide if we like the direction before finalizing it.

- Removes on-chain transaction conflict detection entirely, which was causing a lot of direct and indirect issues. We now rely fully on multi-proofs for security
- Fixes the finalization DOS issues and the problematic conflict resolution issues, **while preserving transition aggregation**.
- Merges `InboxOptimized1` with the `Inbox` to avoid logic duplication, and since the inbox already needs to be aware of spans if we want to keep transition aggregation and fix the reported issues(this will become obvious once you start reviewing the PR)

## Description

**Remove on-chain transaction conflict**

Fairly straight forward, I believe there was strong consensus after the audit issues reported that this was causing more harm than good.  The whole point of multi proof is that you have backup security guarantees. I think conflict resolution is still important, but at the individual proof level to satisfy stage 2 requirements(which is not what we are targeting at the moment, so no need to build an on-chain mechanism for it now).

**Fix finalization DOS issues and other problematic conflict resolution issues**

Keeping transition aggregation, even after removing on-chain conflict detection was challenging, since we don't keep the full transition history on-chain it becomes hard to detect if a new transition for the same proposal and parentHash arrives(but with a different hash) what we should do.

For example imagine:
- Prover 1 submits a proof proving P1->P3 with parent transition linked to proposal 0. The contract accepts the proof and stores the transition record.
- Prover 2 submits a proof proving P1->P2 with the same parent transition. Since there's already a record stored for this combination, what should the contract do? From outside it is obvious that it should not replace it, since the new proof leads to a "shorter" chain.
- Prover 3 submits a proof for P1->P5 with the same parent transition. Since there's already a record stored for this combination, what should the contract do? From outside it is obvious that it SHOULD replace it, since this new proof leads to a "longer" chain. But bond decisions should stay the same, since proof timing for P1,P2 and P3 have already been decided

**This PR incorporates a detection mechanism that always chooses the longer chain of proofs for the same proposal and parent, and maintains the existing bond decisions and merges the new ones**

_Note this does not resolve the issue for two different proofs P1->P3 and P2->P5 since the contract is not aware how to connect them. This can be solved via off-chain prover coordination. Otherwise it results in wasted prover work._

**Merge InboxOptimized1 with Inbox**

In reality we will only deploy the most optimized version of the inbox, both on testnet and mainnet. The reason we kept them separate is so that anyone reading them can have an understanding of the protocol rules without needing to dive into the optimizations. But if we need the storing of the transition records to be aware of span length the abstraction breaks, and we end with a lot of code repetition. This is why I suggest we merge both inbox under the same contract

## TODOs

- [ ] Decide if we like this design
- [ ] Remove `InboxOptimized1` entirely and adjust the rest of the codebase and tests
- [ ] Refactor the code to make it more readable. I feel there are better abstractions to some of this that I haven't figured out yet


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -458,11 +465,56 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
 
     /// @dev Builds and persists transition records for batch proof submissions
     /// Validates transitions, calculates bond instructions, and stores records
-    /// @dev Virtual function that can be overridden for optimization (e.g., transition aggregation)
+    /// @dev Aggregates consecutive proposals to reduce storage and event costs.
```
dantaik: @ggonzalez94 maybe we overlooked a potential optimization for this function in the first place -- why don't we enable the off-chain prover to prove aggregated proofs across multiple consecutive proposals, by **adding `span` directly into the `Transition` record**. Then we don't need to work on aggregation on-chain.

dantaik:
@ggonzalez94 I have the following suggestions:

## 1. Support Multi-Span Transitions

We may have been overlooking this optimization in the first place -- why don't we enable the off-chain prover to prove aggregated proofs across multiple consecutive proposals by adding `span` directly into the `Transition` struct?

This eliminates the need for on-chain aggregation logic, significantly simplifying the code. With this approach, a `Pn → P(n+1)` transition is simply a `span=1` transition, and the code always assumes `span >= 1` for every record.

From the prover's perspective, multiple multi-span transitions can still be verified with a single proof. The off-chain proof aggregation pipeline becomes:

```
block proofs → proposal proofs → multi-span transition proofs → single multi-transition proof
```

## 2. Verifiable Intermediate Checkpoints

Consider two (multi-span) transitions: `P1 → P3` from Alice and `P2 → P5` from Bob.

To allow Bob to write a new transition record `P3 → P5` without re-verifying the proof, when `P2 → P5` is written, we hash the intermediate checkpoints (`P3`, `P4`) into an on-chain `bytes32`. When Bob later needs to write `P3 → P5`, the `P3` checkpoint can be verified via hash comparison rather than re-verifying the already-verified proof.

This also eliminates the need for proof regeneration.

## 3. Proactive Tail Forwarding During Proving

For now, in a transition `P3 → P5`, lets call `P3` the *head* and `P5` the *tail*.

When Alice submits transition `P3 → P5`, the protocol should first look up `P5 → P?`.If a `P5 → P9` record already exists on-chain (and no further `P9 → P?` is available), the protocol writes `P3 → P9` directly instead.

While proactive tail forwarding costs additional `SLOAD`s, earlier proving also means cheaper transactions—slightly penalizing late proving.

## 4. Optional Parent Transition Provisioning

For any given transition (e.g. `P7 → P10`), the prover can optionally supply its parent transition (e.g. `P2 → P7`). Once the parent transition is found, validated, and yet-to-be-finalized, the contract can  update `P2 → P7` to `P2 → P10` in the same storage slot.

If no parent transition is provided (or found), the new transition `P7 → P10` is likely written into a new storage slot, which costs significantly more gas.

## 5. Max-Span Configuration

We could introduce a `maxSpan` immutable in the protocol, initially set to `1`. This ensures off-chain software doesn't need to handle overlapping transitions like `P1 → P3` & `P2 → P5` until it's ready. Once the software matures, we can increase `maxSpan` accordingly.


-- 

what do you think?

ggonzalez94:
closing this since it became to complex and it still had issues around it

PR #20876, opened 2025-11-30
 fix(relayer): propagate submitBridgeTx error in Start

ANtutov:
The bridge Start method previously discarded the error from submitBridgeTx, causing the CLI to report successful startup even when the transaction submission failed. This change returns the error so the SubcommandAction harness can surface and handle failures consistently with other components (indexer, processor, watchdog).


dantaik:
@claude please review

PR #20882, opened 2025-12-01
[Draft] feat(protocol): Core state hash in inbox to optimize gas and simplify bond instruction imports

linoscope:
**In draft state, will revisit after initial feedback.** 

### Summary

Start storing `coreStateHash` in inbox.

Motivation is two-fold:
- Gas optimization by removing need to pass `parentProposals` at proposing time (~20% when ring buffer is full).
- Simplify L1->L2 bond instruction imports by removing `BOND_PROCESSING_DELAY` and letting bond instructions be imported via the anchor block.

Note on latter point: As @smtmfft described in his [Bond Instruction Processing doc](https://www.notion.so/taikoxyz/Bond-Instruction-Processing-2719673143d680fb9563ee00a98a8ef1), the blocker for using anchor block for bond instructions was that the finalized `bondInstructionHash` of a given anchor block could not be easily determined just by the anchor hash, and you needed to traverse the proposals. By storing `coreStateHash` in the inbox (where `coreState` contains `bondInstructionHash`), we can now prove the `bondInstructionHash` of any anchor block using Merkle proofs and core state preimages. This removes the traversal requirement and unblocks anchor-based imports.

### Commits

- **f90d7bbf6: Store `coreStateHash` in inbox contract**
  - Add `coreStateHash` to `Inbox.sol`
  - Remove `parentProposals[]` from `ProposeInput`
  - Remove `_verifyChainHead()`

- **31c50d2be: Remove `coreStateHash` from `Proposal` struct**
  - Delete redundant `coreStateHash` field from `Proposal`

- **5f40dfdf0: Update anchor to process bonds per-block**
  - Move bond instructions from per-proposal to per-block

### Gas Impact

Test Case | Old Gas | New Gas | Diff 
-- | -- | -- | --
propose_reuse_ring_buffer_Inbox | 64,264 | 49,445 | −14,819 (−23.1%)
propose_reuse_ring_buffer_InboxOptimized1 | 64,619 | 49,812 | −14,807 (−22.9%)
propose_reuse_ring_buffer_InboxOptimized2 | 49,433 | 38,810 | −10,623 (−21.5%)
propose_single_empty_ring_buffer_Inbox | 73,422 | 73,637 | +215 (+0.29%)
propose_single_empty_ring_buffer_InboxOptimized1 | 73,416 | 73,637 | +221 (+0.30%)
propose_single_empty_ring_buffer_InboxOptimized2 | 67,569 | 67,971 | +402 (+0.59%)

### Compatibility

This breaks client code. Requires fixes if we are going to take this route.


smtmfft:
From the prover’s point of view, I think this approach works to some extent, but it is not better than the currently preferred solution #20868. The main concern is that both the MKL proof and the anchor block introduce dynamic hashes. In practice, the prover’s complexity stays the same: it must first verify the MKL proof against the anchor block, and then traverse the chain all the way up to the current proposal block. This means performing anywhere from a few dozen hashes to well over a hundred hashes (considering the anchor offset can be up to 96, and the MKL depth grows over time).
In contrast, the current solution requires a fixed 6 hashes and does not require MKL verification at all.

Recently we have evaluated three solutions:
#20840
#20868 (the current preferred option)
This proposal

Solution #20840 is the simplest for both prover and client, but introduces extra SLOAD and changes the public input.
This proposal seems beneficial for the client/sidecar, but adds additional complexity on the prover side.

So overall, the second solution appears to strike the best balance.


BTW: the easiest way for me is like we save corestate in inbox, but we don't process any anchor related things, instead, we add corestate into the public input, kind of combination of this one & #20840 :)

linoscope:
> it must first verify the MKL proof against the anchor block, and then traverse the chain all the way up to the current proposal block.

Why does it need to traverse the chain up to the current proposal block? Isn't the range you are interested not the "anchor to proposed block" but rather the "parent anchor to current anchor". And for that, think what you need is:
1. Store the latest `bondInstructionHash` of the anchor block in the anchor contract.
2. In the anchor parameter, pass the new `bondInstructionHash` of the new anchor block, plus the list of `bondInstructions`.  In prover you just need one MKP to check that the `bondInstructionHash` indeed matches that stored in the anchor block's inbox core state hash.
3. In the anchor contract, check that the `hash(parent.bondInstructionHash, [bondInstruction]) == param.bondInstructionHash`.

smtmfft:
> Why does it need to traverse the chain up to the current proposal block? Isn't the range you are interested not the "anchor to proposed block" but rather the "parent anchor to current anchor". And for that, think what you need is:

Because it's not part of public input, so, one can give whatever value in prover if you don't link them together.

linoscope:
> > Why does it need to traverse the chain up to the current proposal block? Isn't the range you are interested not the "anchor to proposed block" but rather the "parent anchor to current anchor". And for that, think what you need is:
> 
> Because it's not part of public input, so, one can give whatever value in prover if you don't link them together.

Actually, don't you already have to do this traversal to verify that the anchor hash of the anchor block ID is correct? So we need to traverse starting from `originBlockHash` of `originBlockNumber` back to the `anchorBlockNumber` and check that the `anchorBlockHash`es are correct.

smtmfft:
> Actually, don't you already have to do this traversal to verify that the anchor hash of the anchor block ID is correct? 

oh, yes, you reminds me this, I am thinking of adding anchor block hash back into the public input like PACAYA did, the anchor traversal in prover does not make sense, but I did not come up with a good place to hold it. By doing that, it's easier to do sth within a anchor block.

linoscope:
> > Actually, don't you already have to do this traversal to verify that the anchor hash of the anchor block ID is correct?
> 
> oh, yes, you reminds me this, I am thinking of adding anchor block hash back into the public input like PACAYA did, the anchor traversal in prover does not make sense, but I did not come up with a good place to hold it. By doing that, it's easier to do sth within a anchor block.

Oh, I see, I didn't know we were considering removing that traversal.

By
> adding anchor block hash back into the public input like PACAYA did

Do you mean we call `blockhash(anchorBlockNumber)` in the L1 inbox contract and store the list in the `Propose` struct? As done in Pacaya.

If this is the case, then you already have the `anchorBlockHash` of anchor block in the prover, hence the additional MKP to prove the core state of that anchor block shouldn't be too much work? Or am I missing something?

smtmfft:
> If this is the case, then you already have the anchorBlockHash of anchor block in the prover, hence the additional MKP to prove the core state of that anchor block shouldn't be too much work? Or am I missing something?

Yes, that's the case, no other dependency, although I do not recommend verify MKL in prover since I think it's expensive than current 6 hashes.


> Do you mean we call blockhash(anchorBlockNumber) in the L1 inbox contract and store the list in the Propose struct? As done in Pacaya.

Yes, I think only prove() needs some changes to support that, so it should be fine

linoscope:
> Yes, I think only prove() needs some changes to support that, so it should be fine

The `blockhash` opcode can only fetch up to 256 blocks in history, if we call them at proving time won't we risk the chain becoming unprovable when there is a huge delay in proofs landing?

linoscope:
> Yes, that's the case, no other dependency, although I do not recommend verify MKL in prover since I think it's expensive than current 6 hashes.

Do you think the gas cost difference is significant enough to influence the design choice? My intuition is that MKP verification isn’t that expensive, though I haven’t analyzed properly. We also need to weigh this against other factors. For example, a 6-proposal delay is quite long, and it increases how long a preconfer’s bond remains locked. This becomes especially problematic for preconfers who preconf many slots, since a longer bond-processing delay forces them to lock up a large amount of TAIKO tokens (well, this might be a good thing for tokenomics :slightly_smiling_face:) . Also the `BOND_PROCESSING_DELAY` feels pretty artificial to me and think it's cleaner if we can remove, but that is more of a personal aesthetic opinion.

smtmfft:
I was told that 1 is not recommended because of sidecar's impl, it's not the case??
actually using recent anchor is kind of the same as BOND_PROCESSING_DELAY == 1, which is way cheaper then either 6 or MKL.

smtmfft:
> since a longer bond-processing delay forces them to lock up a large amount of TAIKO tokens

I think in normal case, if everyone proves in time, there will be no bond instruction, so it should have less impact to those honest proposal/prover? let me double check.

linoscope:
Closing, similar optimization already done in https://github.com/taikoxyz/taiko-mono/pull/20884

PR #20884, opened 2025-12-01
feat(protocol): use sequential proving to fix audit issues and gain maximum simplicity and efficiency

ggonzalez94:
This is a long PR, but the simplicity and efficiency we gain might be worth going trough the effort of the refactor. It solves all the issues we had with transition aggregation in a way that is maximum simple and efficient.

It is still in draft given the amount of changes. More tests, libraries and optimizations still need to be properly implemented. ** But the simplification and gas results are very promising**.

## Context

A lot of the complexity and issues in our codebase comes from trying to do parallel proving(anyone can submit a proof for their transition before the parent has landed) and proving aggregation at the same time. This two goals are in tension:
- Parallel proofs tries to give maximum accountability to the proposer. The liveness window starts from the moment you propose and you have a set window to submit their proof. This incentivices proposers to generate individual proofs and submit them on-chain as soon as they have them.
- Aggregation tries to maximize utility and reduce costs. It encourages proposers to coordinate and submit their proofs together.

As we can see this two goals are in tension. And while accountability is important, having the proof submitted on chain is not very useful if the parent proposal has not been proven and you cannot finalize the chain.

## Summary
This PR is a radical simplification. Instead of allowing out of order proofs and try to layer aggregation on top, it embraces sequential proving.
- Proofs are always expected in sequence. You can only submit a proof for proposal P5 if P4 has been proven
- Aggregation happens naturally, since you can generate a proof for several sequential proposals
- `_provingWindow` and `_extendedProvingWindow` are now counted from the moment a proposal becomes provable. This is the `max(proposalTimeStamp, priorFinalizedTimestamp)` and is only relevant for the first proposal in a batch of proofs
- Finalization now happens inside `prove`, since now all the data necessary to finalize the chain is available. Thus, propose becomes simpler and cheaper

This allowed to simplify the data structures greatly. And since `CoreState` is now smaller, we can actually store it on-chain, which reduces some of the race conditions we had before, and allows us to remove data from calldata and events.

## Open Questions
- [x] Incentives: Let's say a prover is generating proofs for P7->P15, which requires a substantial amount of computation. The someone comes and proofs only P7, and now the lastFinalizedProposal is P7, so the large proof now becomes invalid. While this does not compromise correctness from the protocol perspective, it does not create the best incentive.
- Proof delay: We are switching the proving windows to now count from when the proposal is submitted to when it can be actually proven. Under honest provers, this is fine. But are there important attack/delay vectors we should consider?

## TODOs

- [x] Discuss this new design
- [x] Implement proper test structure and cases(the current set was fully AI generated from the previous one)
- [x] Implement proper gas benches
- [ ] Update Derivation
- [x] Improve the proving window calculation based on @linoscope suggestion
    https://github.com/taikoxyz/taiko-mono/pull/20926
- [x] Review library encoding
- [x] Review library hashing


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -134,23 +116,14 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     /// @notice The timestamp when the first activation occurred.
     uint48 public activationTimestamp;
 
-    /// @notice Flag indicating whether a conflicting transition record has been detected
-    bool public conflictingTransitionDetected;
+    /// @notice Persisted core state.
+    CoreState internal _state;
```
dantaik: CoreState takes 3 slots in storage. I think most times, all 3 slots will be updated, costing 15K gas. I believe we can do better.

packages/protocol/contracts/layer1/core/impl/InboxOptimized.sol
```diff
@@ -8,64 +8,34 @@ import { LibProposedEventEncoder } from "../libs/LibProposedEventEncoder.sol";
 import { LibProveInputDecoder } from "../libs/LibProveInputDecoder.sol";
 import { LibProvedEventEncoder } from "../libs/LibProvedEventEncoder.sol";
 import { Inbox } from "./Inbox.sol";
-import { InboxOptimized1 } from "./InboxOptimized1.sol";
 import { ICheckpointStore } from "src/shared/signal/ICheckpointStore.sol";
 
-import "./InboxOptimized2_Layout.sol"; // DO NOT DELETE
+import "./InboxOptimized_Layout.sol"; // DO NOT DELETE
```
dantaik: Suggest to delete `./InboxOptimized2_Layout.sol `and rename `./InboxOptimized1_Layout.sol` to `./InboxOptimized_Layout.sol`.

You also need to change gen-layout.sh accordingly.

packages/protocol/contracts/layer1/core/libs/LibBondInstruction.sol
```diff
@@ -5,202 +5,67 @@ import { IInbox } from "../iface/IInbox.sol";
...
-                : _loopCopyBondInstructions(_existingInstructions, _newInstructions);
-        }
-    }
-
-    // ---------------------------------------------------------------
-    // Internal Functions
-    // ---------------------------------------------------------------
-
-    /// @notice Calculates bond instructions based on proof timing and prover identity
+    /// @notice Calculates all bond instructions for a sequential prove call.
```
dantaik: With only this one single function remaining, I suggest to move it directly into Inbox.sol, what do you think?

packages/protocol/contracts/layer1/core/iface/IInbox.sol
dantaik: if there is no proof aggregation, shall we remove `span` here?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -268,56 +223,67 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+            state.lastProposalBlockId = uint48(block.number);
+
             Proposal memory proposal = Proposal({
-                id: coreState.nextProposalId++,
+                id: proposalId,
                 timestamp: uint48(block.timestamp),
                 endOfSubmissionWindowTimestamp: endOfSubmissionWindowTimestamp,
                 proposer: msg.sender,
-                coreStateHash: _hashCoreState(coreState),
+                coreStateHash: _hashCoreState(state),
```
dantaik: If we store the full core state in storage, `coreStateHash` may no longer be necessary. That said, I’m inclined to keep only the hash on-chain: persisting just the core state hash still preserves the original, elegant idea of storing a compact commitment instead of the full state.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -379,221 +345,162 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+                    _extendedProvingWindow,
+                    _input.proposals[0],
+                    _input.metadata[0],
+                    firstReadyTimestamp_
+                );
+
+            record_.bondInstructions = bondInstructions;
+
+            if (bondInstructions.length != 0) {
+                newState_.bondInstructionsHash = LibBonds.aggregateBondInstruction(
```
dantaik: With sequential proving, “proved” effectively means “finalized”. In that model, we can drop `bondInstructionsHash` entirely from the core state and instead:

- Build a bondInstructions array (each instruction including the corresponding proposal ID in its hash).
- Hash this array into a single value if the array is not empty
- Send that hash as a bond-signal to L2.

On L2, we build a contract that honors these signals to process the corresponding bond instructions with best effort. These bond-signals can even be consumed out of order. 

This is what I mentioned as "decoupling bond processing from Anchor"

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -379,221 +345,162 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+                    newState_.bondInstructionsHash, bondInstructions[0]
+                );
+            }
+
+            newState_.lastFinalizedProposalId = expectedId - 1;
+            newState_.lastFinalizedTransitionHash = parentHash;
+            newState_.lastFinalizedTimestamp = uint48(block.timestamp);
```
dantaik: Is there a potential race condition here, given that both propose() and prove() now attempt to modify the core state?

dantaik (changes requested):

packages/protocol/contracts/layer1/core/iface/IInbox.sol
ggonzalez94: ended up removing the entire TransitionRecord, but I left a TODO that maybe it is worth considering adding some info for off-chain actors to know how many proposals were proven or the first and last of the batch.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: Now that Corestate is smaller it ends up being a better decision to store it on-chain, as each function only modifies 1 slot I believe. This allowed to simplify a lot of the other data structures and remove some data from events. So it ends up being cheaper

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: This is a great idea. I implemented it as part of the PR

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: I don't think this is the case. Can you elaborate more? If anything I think given corestate is now on-chain race conditions are eliminated

packages/protocol/contracts/layer1/core/impl/Inbox.sol
linoscope: > keep only the hash on-chain: persisting just the core state hash still preserves the original

I think if we do this, since the `prove` function now modifies the core state, proposers will face a race condition where proofs can modify the core state and invalidate the core state preimage in their parameters.

YoGhurt111:
> Proof delay: We are switching the proving windows to now count from when the proposal is submitted to when it can be actually proven. Under honest provers, this is fine. But are there important attack/delay vectors we should consider?

I have discussed this question with @davidtaikocha yesterday, and I have a concern about this: without any constraints, this would cause the proving window to extend indefinitely from a design perspective.
For example, around time T1, proposals A, B, and C are made by proposers A, B, and C respectively.
If proposer A behaves dishonestly and submits their proof only at T1 + 2h, then B’s proving window shifts from [T1, T1 + 2h] to [T1 + 2h, T1 + 4h].
Similarly, if B does the same, C’s proving window is pushed from [T1, T1 + 2h] to [T1 + 4h, T1 + 6h], and so on.

In other words, any dishonest prover can delay the proving window for everyone else.
In this example, we assume the `provingWindow` is 2 hours and that the prover is the proposer.

Although our own prover implementation does not exhibit this issue, in theory this could be a potential vulnerability.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: Now it only takes 2. And `propose` only updates one slot. `prove` still updates 2.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -164,18 +147,16 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     /// @notice Initializes the Inbox contract
     /// @param _config Configuration struct containing all constructor parameters
     constructor(IInbox.Config memory _config) {
-        require(_config.checkpointStore != address(0), ZERO_ADDRESS());
+        require(_config.signalService != address(0), ZERO_ADDRESS());
         require(_config.ringBufferSize != 0, RingBufferSizeZero());
 
         _codec = _config.codec;
         _bondToken = IERC20(_config.bondToken);
```
davidtaikocha: Shall we remove this `IERC20 internal immutable _bondToken;` here? Looks like it's not used in the inbox anymore.

ggonzalez94:
> > Proof delay: We are switching the proving windows to now count from when the proposal is submitted to when it can be actually proven. Under honest provers, this is fine. But are there important attack/delay vectors we should consider?
> 
> I have discussed this question with @davidtaikocha yesterday, and I have a concern about this: without any constraints, this would cause the proving window to extend indefinitely from a design perspective. For example, around time T1, proposals A, B, and C are made by proposers A, B, and C respectively. If proposer A behaves dishonestly and submits their proof only at T1 + 2h, then B’s proving window shifts from [T1, T1 + 2h] to [T1 + 2h, T1 + 4h]. Similarly, if B does the same, C’s proving window is pushed from [T1, T1 + 2h] to [T1 + 4h, T1 + 6h], and so on.
> 
> In other words, any dishonest prover can delay the proving window for everyone else. In this example, we assume the `provingWindow` is 2 hours and that the prover is the proposer.
> 
> Although our own prover implementation does not exhibit this issue, in theory this could be a potential vulnerability.

Addressed [here](https://github.com/taikoxyz/taiko-mono/pull/20926)

packages/protocol/contracts/layer1/core/impl/CodecSimple.sol
```diff
@@ -140,23 +140,11 @@ contract CodecSimple is ICodec {
     }
 
     /// @inheritdoc ICodec
-    function hashTransitionRecord(IInbox.TransitionRecord calldata _transitionRecord)
```
dantaik, line 143: How about we merge Inbox and InboxOptimized into a single Inbox.sol file, then remove CodecSimple and LibHashSimple.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: Agreed. also the IERC20 imports

packages/protocol/contracts/layer1/core/impl/InboxOptimized2_Layout.sol
```diff
@@ -7,23 +7,23 @@ pragma solidity ^0.8.24;
 /// @custom:security-contact security@taiko.xyz
 
 // solhint-disable max-line-length
-//   _initialized                   | uint8                                              | Slot: 0    | Offset: 0    | Bytes: 1   
-//   _initializing                  | bool                                               | Slot: 0    | Offset: 1    | Bytes: 1   
+//   _initialized                   | uint8                                              | Slot: 0    | Offset: 0    | Bytes: 1
```
dantaik: please remove InboxOptimized[1|2]_Layout.sol

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -26,7 +27,7 @@ import "./Inbox_Layout.sol"; // DO NOT DELETE
 ///      instead, Shasta uses a separate inbox address.
 /// @dev This contract implements the fundamental inbox logic including:
 ///      - Proposal submission with forced inclusion support
-///      - Proof verification with transition record management
+///      - Sequential proof verification
```
dantaik, line 31: rename ACTIVATION_WINDOW to _ACTIVATION_WINDOW as it is private.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -372,228 +299,273 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+            uint256 parentBlockNumber = block.number - 1;
+            derivation_ = Derivation({
+                originBlockNumber: uint48(parentBlockNumber),
+                originBlockHash: blockhash(parentBlockNumber),
+                basefeeSharingPctg: _basefeeSharingPctg,
+                sources: result.sources
+            });
+
+            proposal_ = Proposal({
+                id: _nextProposalId,
```
dantaik: If we prove **n** consecutive proposals, we can add a `parentProposalHash` in the `Proposal` struct, then in the ProveInput struct, we only need to provide the last proposal, not the whole list. The prover can verify these proposals can be chained by checking their `parentProposalHash`

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -372,228 +299,273 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
-        TransitionRecord memory _transitionRecord
-    )
-        internal
-        virtual
+    /// @dev Prepares proof verification inputs and resulting state updates.
+    /// @param _input The prove input data.
+    /// @return result_ Aggregated result containing new state, bond instruction, verifier hash, and indexes.
+    function _buildProofResult(ProveInput memory _input)
+        private
+        returns (ProofBuildResult memory result_)
```
dantaik: how about ProofResult or ProvingResult

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -1074,64 +829,66 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+        returns (bytes32 signal_)
+    {
+        signal_ = _bondSignalHash(_bondInstruction);
+        _signalService.sendSignal(signal_);
+    }
+
+    /// @dev Calculates the bond signal hash for a bond instruction.
+    /// @param _bondInstruction The bond instruction to hash.
+    /// @return The hash of the bond instruction.
+    function _bondSignalHash(LibBonds.BondInstruction memory _bondInstruction)
```
```suggestion
    function _hashBondInstruction(LibBonds.BondInstruction memory _bondInstruction)
```

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -223,101 +198,53 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
-        // Build transition records with validation and bond calculations
-        _buildAndSaveTransitionRecords(input);
+        ProofBuildResult memory result = _buildProofResult(input);
 
-        uint256 proposalAge;
-        if (input.proposals.length == 1) {
-            unchecked {
-                proposalAge = block.timestamp - input.proposals[0].timestamp;
-            }
+        _proofVerifier.verifyProof(result.proposalAge, result.aggregatedProvingHash, _proof);
```
dantaik: lets do this external call at the very end of `prove()` to save some gas.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -46,19 +47,21 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
         bool allowsPermissionless;
     }
 
+    /// @notice Result of preparing proof verification and finalization data
+    struct ProofBuildResult {
+        CoreState newState;
+        LibBonds.BondInstruction bondInstruction;
+        uint256 proposalAge;
+        uint256 firstProvenIndex;
+        bytes32 aggregatedProvingHash;
```
dantaik: aggregatedProvingHash
```suggestion
        bytes32 aggregatedTransitionsHash;
```

dantaik (changes requested):

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -372,228 +299,273 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+    /// @return result_ Aggregated result containing new state, bond instruction, verifier hash, and indexes.
+    function _buildProofResult(ProveInput memory _input)
+        private
+        returns (ProofBuildResult memory result_)
     {
-        (bytes26 transitionRecordHash, TransitionRecordHashAndDeadline memory hashAndDeadline) =
-            _computeTransitionRecordHashAndDeadline(_transitionRecord);
+        uint48 firstReadyTimestamp;
+        (result_.newState, result_.bondInstruction, firstReadyTimestamp, result_.firstProvenIndex) =
+            _processProof(_state, _input);
```
dantaik: does it make sense to merge _processProof into _buildProofResult

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -372,228 +299,273 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+            bondInstruction_ = LibBondInstruction.calculateBondInstruction(
+                _provingWindow,
+                _extendedProvingWindow,
+                _input.proposals[firstProvenIndex_],
+                _input.transitions[firstProvenIndex_],
+                firstReadyTimestamp_
+            );
+
+            newState_.lastFinalizedProposalId = expectedId - 1;
+            newState_.lastFinalizedTransitionHash = parentHash;
```
dantaik: writeing lastFinalizedTransitionHash in between `lastFinalizedProposalId` and `lastFinalizedTimestamp` may pevent optimizer/compiler to use one SSTORE for `lastFinalizedProposalId` and `lastFinalizedTimestamp`. 

Suggest to move `newState_.lastFinalizedTransitionHash = parentHash;` after `newState_.lastFinalizedTimestamp = uint48(block.timestamp);`

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -372,228 +299,273 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+                    );
+                }
+
+                parentHash = _hashTransition(transition);
+                ++expectedId;
+            }
+
+            // Only the first unproven proposal in a sequential prove can be late; later proposals
+            // become proveable when the previous one finalizes within this transaction.
+            bondInstruction_ = LibBondInstruction.calculateBondInstruction(
```
dantaik: how about we move calculateBondInstruction from LibBondInstruction to the Inbox contact.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -372,228 +299,273 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
-            entry.recordHash = _recordHash;
-            entry.finalizationDeadline = _hashAndDeadline.finalizationDeadline;
-        } else if (recordHash == _recordHash) {
-            emit TransitionDuplicateDetected();
-        } else {
-            emit TransitionConflictDetected();
-            conflictingTransitionDetected = true;
-            entry.finalizationDeadline = type(uint48).max;
+        unchecked {
+            newState_ = _stateBefore;
```
dantaik: maybe you don't have to make a copy. Updating _stateBefore directly will be fine.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -372,228 +299,273 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+
+            uint48 expectedId = _stateBefore.lastFinalizedProposalId + 1;
+            uint256 count = _input.proposals.length;
+            bytes32 parentHash = _stateBefore.lastFinalizedTransitionHash;
+
+            // Find the index of the first proposal that still has not been proven
+            // This help if the array includes older proposals that have already been proven
+            firstProvenIndex_ = _findFirstIndexToProve(_stateBefore, _input);
```
dantaik: `firstProvenIndex_` may not be an concern if we have `parentProposalId` inside `Proposal`?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -372,228 +299,273 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+
+                bytes32 proposalHash = _checkProposalHash(proposal);
+                Transition memory transition = _input.transitions[i];
+                require(
+                    proposalHash == transition.proposalHash, ProposalHashMismatchWithTransition()
+                );
+                require(transition.parentTransitionHash == parentHash, InvalidParentTransition());
+
+                if (i == firstProvenIndex_) {
+                    firstReadyTimestamp_ = _computeReadyTimestamp(
```
dantaik: I'm a bit lost on `firstReadyTimestamp_`...

dantaik (changes requested):

packages/protocol/contracts/layer1/core/impl/InboxOptimized2_Layout.sol
ggonzalez94: I remember having removed it. I don't know why it keeps coming back lol.

ggonzalez94: Oh it was because of the script that setups local ignores. It should be gone for good now

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: I just tried it and as expected the compiler is smart enough. But I agree it is a good idea to do it anyways so that the result does not change if we switch compiler versions or optimizer runs

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: this is necessary to support the scenario you mentioned in our slack. Where proverA submits P3->P5, before proverB submits P4->P8 for example. `_findFirstIndexToProve ` finds the first proposal that has not yet been proven in prover's B array so that the prover work is not wasted and he can still submit a proof

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: agree

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: By merging _buildProofResult with _processProof the need to make a copy went away

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: it is the timestamp when the first proposal from this batch(that has not yet been proven) became provable. This is the max(proposal.timestamp and the lastFinalizedTimestamp). Happy to change the name to something more meaningful if you have suggestions 😄

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: I suggest we first review these 2 PRs and then decide.
- https://github.com/taikoxyz/taiko-mono/pull/20926/files
- https://github.com/taikoxyz/taiko-mono/pull/20927

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: I like ProofResult

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: in case of a revert you mean? I just moved it to the end before the event. Let me know if you meant something else

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -223,101 +198,53 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
     ///         process at least `config.minForcedInclusionCount` if they are due.
-    ///      4. Updates core state and emits `Proposed` event
+    ///      3. Updates core state and emits `Proposed` event
     /// NOTE: This function can only be called once per block to prevent spams that can fill the ring buffer.
     function propose(bytes calldata _lookahead, bytes calldata _data) external nonReentrant {
         unchecked {
-            // Decode and validate input data
             ProposeInput memory input = _decodeProposeInput(_data);
-
             _validateProposeInput(input);
```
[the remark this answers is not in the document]
ggonzalez94: fixed: https://github.com/taikoxyz/taiko-mono/pull/20884/commits/54869914813f0d07526ee0de3245e195603361a5

dantaik (approved):

dantaik (approved):

PR #20887, opened 2025-12-01
fix(taiko-client): use retry logic for checking canonical batches and retry on timeouts pacaya

cyberhorsey:
not sure if its better to retry within the for loop or retry the entire for loop here - it depends on whether we get a timeout on a specific call or batch-wide. A separate PR can be done for Shasta, so this PR can be cherry-picked onto a tag that can be used for Hoodi and then Mainnet.


packages/taiko-client/driver/chain_syncer/chain_syncer.go
```diff
@@ -51,12 +51,24 @@ func New(
...
+	eventSyncer, err := event.NewSyncer(
+		ctx,
+		rpc,
+		indexer,
+		state,
+		tracker,
+		blobServerEndpoint,
+		latestSeenProposalCh,
+		backOffMaxRetries,
+		backOffRetryInterval,
```
davidtaikocha, line 70: I think the current driver high-level retry design is:
- When inserting a batch / proposal from the on-chain event, the [whole proposal needs to be retried until there is no error from the event handler](https://github.com/taikoxyz/taiko-mono/blob/main/packages/taiko-client/pkg/chain_iterator/block_batch_iterator.go#L232).
- When inserting a preconf block API, the server will fail fast with a 500 status code response to let the sidecar handle the retry.

So I feel like maybe we don't need to touch something related to retry to fix this problem since they are impled at the higher level pipeline already, maybe it's better to change [here](https://github.com/taikoxyz/taiko-mono/blob/main/packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go#L184)? 

In `isKnownCanonicalBatchPacaya`'s return values, shall we distinguish if there is an (network) error happened (which can be retried like what changed in this PR rn) or the checks found out it's not known from the canonical (which is safe to continue the execution)? And then the higher-level pipeline will determine how to retry automatically instead of introducing another internal retry inside the pipeline?

packages/taiko-client/driver/chain_syncer/event/blocks_inserter/pacaya.go
```diff
@@ -74,6 +81,9 @@ func (i *Pacaya) InsertBlocks(
 	i.mutex.Lock()
 	defer i.mutex.Unlock()
 
+	ctx, cancel := rpc.CtxWithTimeoutOrDefault(ctx, rpc.DefaultRpcTimeout)
+	defer cancel()
```
MatusKysel, line 85: are you sure about this ? We having several rpc call inside this function are you sure we are not breaking something with this timeout ?

cyberhorsey: we are probably breaking something with this timeout :D. preconf partners have been hitting a timeout here occasionally. I will submit a better update to this PR to handle the timeout error

packages/taiko-client/driver/chain_syncer/chain_syncer.go
davidtaikocha: Opened a PR to solve the issue by this way here: https://github.com/taikoxyz/taiko-mono/pull/20925 , also feel free to take a look whenever having time :)

davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/pull/20925

PR #20888, opened 2025-12-01
chore(protocol): update derivation structs

ggonzalez94:
update `Derivation.md` to reflect that `proverAuthBytes` is per proposal and no longer by DerivationSourceManifest. Now it matches `LibManifest.sol`

### DIFF
--- a/packages/protocol/docs/Derivation.md
+++ b/packages/protocol/docs/Derivation.md
@@ -165,17 +165,15 @@ struct BlockManifest {
 /// @notice Represents a proposal manifest containing proposal-level metadata and all sources
 /// @dev The ProposalManifest aggregates all DerivationSources' blob data for a proposal.
 struct ProposalManifest {
-  /// @notice Prover authentication data (proposal-level).
+  /// @notice Prover authentication data (proposal-level, shared across all sources).
   bytes proverAuthBytes;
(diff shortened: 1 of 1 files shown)


packages/protocol/docs/Derivation.md
```diff
@@ -165,17 +165,15 @@ struct BlockManifest {
...
-  /// @notice Array of derivation source manifests (one per derivation source).
+  /// @notice Array of derivation source manifests (one per DerivationSource).
   DerivationSourceManifest[] sources;
 }
 
 /// @notice Represents a derivation source manifest containing blocks for one source
 /// @dev Each proposal can have multiple DerivationSourceManifests (one per DerivationSource).
 struct DerivationSourceManifest {
-  /// @notice Proposal-level prover authentication data; ignored when the derivation source is a forced inclusion.
-  bytes proverAuthBytes;
```
davidtaikocha, line 178: I remember that we pushed this PR: https://github.com/taikoxyz/taiko-mono/pull/20590 , to make both normal proposal and forced-inclusion proposal use `DerivationSourceManifest` as the data structure to encode inside their blobs, to simplify the protocol. Has the design been changed?

ggonzalez94: no, I just got confused because `LibManifest` was not updated. I'll clarify this to the auditors, thanks

PR #20893, opened 2025-12-02
[CONCETPUAL] feat(protocol): Add anchor info back into the public input

smtmfft:
This PR tries to add anchor hash back into public input as PACAYA did (not the same way through), there are a few reasons:
-. It's easier in L1 to get the anchor block hash, if we don't do that, prover needs to traversal back to max anchor offset, and roll over all parents until it reaches the parent block, which bring lots of unnecessary workload.
-. for security reason, once prover & client switches to a L1 fork without anchor hash checked on-chain, the proof will be accepted without any conflict.
-. perhaps it's also helpful for some design that based on anchor, like https://github.com/taikoxyz/taiko-mono/pull/20882

Please note: this is just a conceptual impl, please review first and take it into consideration as we are now re-designing the proving part.


dantaik:
what is "anchor access list"?

packages/protocol/contracts/layer1/core/libs/LibHashOptimized.sol
```diff
@@ -346,19 +346,62 @@ library LibHashOptimized {
...
+        returns (bytes32)
+    {
+        unchecked {
+            uint256 length = _anchorAccessList.length;
+            if (length == 0) {
+                return EMPTY_BYTES_HASH;
+            }
+
+            if (length == 1) {
+                bytes32 anchorBlockHash = blockhash(_anchorAccessList[0]);
```
linoscope, line 381: The `blockhash` opcode returns `0`  when the passed block ID is older than 256 blocks before ([specs](https://github.com/ethereum/execution-specs/blob/602f102878307c51c79b673ce96d2c514bfefcdd/src/ethereum/prague/vm/instructions/block.py#L53-L54)). And since we are calling this at proving time (not proposing time as done in Pacaya), this means that `blockhash` can return `0` here if the proofs does not land on time for some reason.

What do we do in that case? If we are going this route, only thing I can see is that we reorg the L2 blocks with anchor hash `0`  when proofs don't hit L1 on time. But that will potentially be a massive reorg. And in general, it's good to say that proofs can be delayed but are always provable eventually without disrupting the L2 execution.

Alternative would be to call `blockhash` in proposal time, as done in Pacaya. Do you see downsides on that? Intuitively the cost seems the same, as we need to call `blockhash` once for each new anchor ID anyways.

smtmfft: > Alternative would be to call blockhash in proposal time, as done in Pacaya. Do you see downsides on that? Intuitively the cost seems the same, as we need to call blockhash once for each new anchor ID anyways.

yes, submitting anchor block during proposal (PACAYA did so) is better than proving, as long as it is less likely to refer to a far away (>256) block considering the max anchor offset is 128. But I did not come up with a good solution since it somehow needs to change the current protocol. This PR is only for conceptual showing, and as we are able to re-design the proving part now, hopefully we will find a better solution.

smtmfft:
> what is "anchor access list"?

The anchor blocks this transaction refers to. Considering the real use case, it seems we just need the last anchor (the last one is the most close to the proposal in preconfer's point of view), but as long as each L2 block has independent anchor height, all data there need to be proven, that's why it's a list.

packages/protocol/contracts/layer1/core/libs/LibHashOptimized.sol
linoscope: > as long as it is less likely to refer to a far away (>256) block considering the max anchor offset is 128.

It's not just "less likely", you can actually make it impossible to propose (= proposal reverts) if the anchor block is too old (older then max anchor offset). And as soon as anchor offset is expired, preconfers can trigger reorg in P2P and limit the size of reorg.

> But I did not come up with a good solution since it somehow needs to change the current protocol.

Did you find specific potential difficulties? Or is it more, "we need to look into it more"?

> This PR is only for conceptual showing, and as we are able to re-design the proving part now, hopefully we will find a better solution.

Makes sense, thanks for starting the discussion!

smtmfft: A simple solution is to change proposal to contain the anchor block number, and store the blockhash onchain at that time, we need to store it since it may exceeds 256 when we prove the block as you already pointed out. However, this seems a conflict with current design?

linoscope: Yeah, thats what I also had in mind. "Conflict with current design" as in the interface of `propose` and the internal storage of the contract changes, right? Think it's pretty much unaviodable if we want to do on-chain fetching of anchor block hashes. In terms of whether it's feasable at this stage considering the state of audits, I have no idea. I heard we plan to have another upgrade shortly after Shasta so maybe we can add it there?

smtmfft:
KECCAK inside ZKVM seems acceptable, we will first use parent block hash linkage approach.
close for now.

PR #20895, opened 2025-12-02
fix(taiko-client-rs): added commit hash to the alethia-reth dependencies

mskrzypkows:
This is needed when we have taiko-client-rs as a dependency in Catalyst node. So we won't have dependencies conflict.

### DIFF
--- a/packages/ejector/Cargo.lock
+++ b/packages/ejector/Cargo.lock
@@ -1797,7 +1797,7 @@ dependencies = [
 
 [[package]]
 name = "ejector"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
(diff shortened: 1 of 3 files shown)


davidtaikocha (approved):
Seems `6ff4dea33acf33851d1c5d320d064c5bd4ad1d50` is a commit 10 days ago, shall we change it to the latest one?

mskrzypkows:
I've just followed the commit, which you have in Cargo.lock. Can be updated if needed.

davidtaikocha:
Closed since it will be merged in https://github.com/taikoxyz/taiko-mono/pull/20921 @mskrzypkows

PR #20802, opened 2025-11-21
fix(protocol): correct MIN_PROVER_AUTH_LENGTH to 256

GarmashAlex:
This change corrects the MIN_PROVER_AUTH_LENGTH from 225 to 256 in Anchor.sol to reflect Solidity ABI padding rules for a 65-byte ECDSA signature within a ProverAuth payload. The previous 225 threshold was derived from summing raw field sizes without accounting for dynamic bytes padding, which allowed calldata lengths in the 225–255 range to pass the guard and then potentially revert during abi.decode. The new value uses the accurate breakdown—head 4×32=128 bytes plus tail 32 length + 96 padded data=128 bytes—for a total of 256 bytes, ensuring the predecode length check aligns with actual ABI-encoded size and avoids decode attempts on undersized payloads. The accompanying comment is updated to document this rationale.

PR #20912, opened 2025-12-03
fix(protocol): avoid revert on anchor

### LINKED ISSUE #20911: Denial of Service via Unsafe ABI Decoding in Anchor Contract
Originally reported by seppi during the audit:

The Anchor contract facilitates L2 block processing through the `anchorV4` function, which accepts an ABI-encoded `ProverAuth` struct as part of its calldata. This data is processed within the `validateProverAuth` function, where `abi.decode` is employed to parse the byte array into a structured format for signature verification and prover designation.

The `validateProverAuth` function attempts to decode the provided `proverAuth` bytes without verifying that they form a valid ABI encoding. This `proverAuth` data is part of the `ProposalManifest` and `DerivationSourceManifest` struct that is provided through the blob during regular proposals or a forced inclusion, respectively. If these bytes are malformed (specifically data that satisfies the minimum length requirement but fails decoding) the `abi.decode` operation will revert. Since the successful execution of the `anchorV4` transaction is a system-imposed requirement for L2 block validity, this revert causes the execution engine to mark the block as invalid.

This rejection creates a critical deadlock in the L2 chain derivation process. The off-chain driver, observing the block rejection, enters an infinite retry loop attempting to process the same invalid payload, resulting in a permanent L2 chain halt. With the L2 chain stuck, the state transitions required to prove the L1 proposal can unlikely occur. Consequently, the proposal containing the malformed data cannot be proved or finalized, stalling the finalization of subsequent valid proposals that depend on the halted state.

Consider implementing a robust decoding pattern within the Anchor contract by wrapping the `abi.decode` operation in an external call to the contract itself (e.g., `try this.decodeProverAuth(...)`). This architecture allows the contract to catch reverts caused by malformed input. In the catch block, the logic should apply safe fallback values (such as designating the proposer with a zero proving fee) to ensure the `anchorV4` transaction executes successfully.

ggonzalez94:
Fixes #20911

~This also ports another fix from #20802 that identified the length of signature was not correctly calculated~
EDIT: after this [suggestion](https://github.com/taikoxyz/taiko-mono/pull/20912#issuecomment-3637389226) we instead rely on the try/catch to safeguard against this and check for MAX_PROVER_AUTH_LENGTH to prevent reverts before the try/catch

---

> [!NOTE]
> Prevents anchor reverts by decoding `ProverAuth` via try/catch with an external helper and enforcing a max payload length; adds tests for malformed encodings and fallback behavior.
> 
> - **Core (`contracts/layer2/core/Anchor.sol`)**:
>   - Add `MAX_PROVER_AUTH_LENGTH = 4_096` and reject oversized `_proverAuth`.
>   - Introduce `decodeProverAuth(bytes)` and use `try/catch` in `validateProverAuth` to avoid reverts on malformed payloads; fallback to proposer with zero fee.
>   - Retain ECDSA signature length check in `validateProverAuth`.
> - **Tests (`test/layer2/core/Anchor.t.sol`)**:
>   - Add tests covering malformed `proverAuth` encodings in `anchorV4` and `validateProverAuth`, and trimmed/undersized payload behavior, asserting fallback to proposer and zero fee.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2ea78d749bed022c6c44be5cfe70f50f387358b7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ggonzalez94:
@codex please review again

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -560,6 +579,12 @@ contract Anchor is EssentialContract {
         /// forge-lint: disable-end
     }
 
+    /// @dev Decodes `ProverAuth` calldata; kept public so `validateProverAuth` can try/catch via
+    /// `this._decodeProverAuth` to swallow malformed encodings.
+    function _decodeProverAuth(bytes calldata _proverAuth) public pure returns (ProverAuth memory) {
+        return abi.decode(_proverAuth, (ProverAuth));
+    }
```
davidtaikocha: Seems this one is duplicated with the `decodeProverAuth` above?

ggonzalez94: removed, thanks

davidtaikocha (approved):

dantaik (approved):

dantaik:
@ggonzalez94 with the try-catch approach, maybe we can delete  `MIN_PROVER_AUTH_LENGTH` and rely on the try-catch?

Also, do you think we should add a  MAX_PROVER_AUTH_LENGTH to prevent failing memory expansion? I'm not sure about this.

dantaik:
An alternative to use try-catch is to use low-level staticcall with manual handling (decodeProverAuth still has to be public or external):

```
function _tryDecodeProverAuth(bytes calldata _proverAuth) 
    internal 
    view 
    returns (bool success, ProverAuth memory proverAuth) 
{
    (success, bytes memory returnData) = address(this).staticcall(
        abi.encodeCall(this.decodeProverAuth, (_proverAuth))
    );
    if (success && returnData.length > 0) {
        proverAuth = abi.decode(returnData, (ProverAuth));
    }
}
```

Not sure which approach is better :D

ggonzalez94:
> @ggonzalez94 with the try-catch approach, maybe we can delete `MIN_PROVER_AUTH_LENGTH` and rely on the try-catch?
> 
> Also, do you think we should add a MAX_PROVER_AUTH_LENGTH to prevent failing memory expansion? I'm not sure about this.

I belive this is the correct way to handle it. I've implemented this

PR #20915, opened 2025-12-04
feat(taiko-client): client updates based on sequential proving

davidtaikocha:
ref: https://github.com/taikoxyz/taiko-mono/pull/20884

Also updated the anvil node version tag in `docker-compose` to fixed `v1.3.2` for testing stability in github CI.


.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, consecutive-proofs-client]
```
```suggestion
    branches: [main]
```

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
```diff
@@ -532,3 +476,31 @@ func (s *ProofSubmitterShasta) ValidateProof(
...
+
+	return backoff.Retry(func() error {
+		coreState, err := s.rpc.GetCoreStateShasta(&bind.CallOpts{Context: ctx})
+		if err != nil {
+			log.Error("Failed to get Shasta core state", "error", err)
+			return fmt.Errorf("failed to get Shasta core state: %w", err)
+		}
+		if new(big.Int).Add(coreState.LastFinalizedProposalId, common.Big1).Cmp(transitionID) >= 0 {
+			log.Info(
+				"Transition verified",
```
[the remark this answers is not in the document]
davidtaikocha: Fixed: https://github.com/taikoxyz/taiko-mono/pull/20915/commits/da7e074e26f711ac1a6ac6864bbd319177c1fce1

davidtaikocha:
@claude plz review.

YoGhurt111 (approved):

packages/taiko-client/prover/event_handler/batches_proved.go
```diff
@@ -90,26 +85,7 @@ func (h *BatchesProvedEventHandler) HandleShasta(
 		return fmt.Errorf("failed to get header by number: %w", err)
 	}
 
-	if header.Hash() == payload.Transition.Checkpoint.BlockHash {
-		log.Info("New valid proven Shasta batch received", "batchID", payload.ProposalId, "lastBatchID", header.Number)
-		return nil
-	}
```
MatusKysel, line 96: doesn't it mean we are now accepting all proved events as valid ?

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
```diff
@@ -543,3 +487,31 @@ func (s *ProofSubmitterShasta) ValidateProof(
...
 	return true, nil
 }
+
+// WaitTransitionVerified waits until the given transition ID is verified on L1.
+func (s *ProofSubmitterShasta) WaitTransitionVerified(ctx context.Context, transitionID *big.Int) error {
+	ctx, cancel := rpc.CtxWithTimeoutOrDefault(ctx, rpc.DefaultRpcTimeout)
+	defer cancel()
+
+	return backoff.Retry(func() error {
+		coreState, err := s.rpc.GetCoreStateShasta(&bind.CallOpts{Context: ctx})
```
MatusKysel, line 497: what if the rpc timeout is higher than `DefaultRpcTimeout` ? maybe we should kind of use multiplier of rpc defined by user ?

MatusKysel (approved):

packages/taiko-client/prover/event_handler/batches_proved.go
davidtaikocha: Yes since for the new design, each time there is a new proof arrived on-chain, it will be immediately finalized. So it's meaningless to let the client double-check it here anymore. So the new assumption is, if a proof (and its block hash) is accepted by the on-chain zk verifier, then it must be correct, no chance there is a confilction.

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
davidtaikocha: we could, but let's do it in a separated PR later :) Since i believe if we want to change so, maybe we should not only change this one.

PR #20918, opened 2025-12-04
fix(beacon): accept VALID/ACCEPTED/SYNCING from forkchoiceUpdated

phrwlk:
Relax the forkchoice status check in beacon sync to accept VALID, ACCEPTED, and SYNCING, and error only on INVALID. The Engine API allows VALID responses when the head is immediately known and valid; requiring SYNCING caused false failures. This change brings the beacon path in line with the engine flow and avoids unnecessary errors on healthy nodes.


davidtaikocha:
In that beacon sync process, it should be `SYNCING `.

PR #20927, opened 2025-12-05
feat(protocol): remove extended window and implement slashing and better bond distribution for late proofs

ggonzalez94:
This PR addresses 2 things:

1. It fixes #20883 
2. It removes `extendedProvingWindow`: The rationale is that I don't think we were getting much of it in the first place, and now with sequential proving  it adds another variable at the formula that makes it harder to reason about the finalization times.

## TODOs
- [ ] What should we do with the slashed amount? We have 3 options
         1. Burn it by just leaving it on the contract(no changes required). Could be rescued if necessary by upgrading the `BondManager`
         2. Properly burn it by sending it to an unused address(e.g. zero address)
         3. Send it to the treasury

I don't have a super strong opinion on this, but I feel 2 is the most appropiate.

---

> [!NOTE]
> Removes extended proving window and provability bonds; late proofs now always trigger a liveness bond from the designated prover with 50% slashing and updated L2 distribution.
> 
> - **L1/Core**:
>   - Remove `extendedProvingWindow` from `IInbox.Config`, `Inbox` immutables/constructor, `getConfig`, and config validation.
>   - Update `DevnetInbox`/`MainnetInbox` constructors to omit `extendedProvingWindow`.
>   - `LibBondInstruction`: simplify to a single late-proof rule using `provingWindow` and `maxProofSubmissionDelay`; emit `LIVENESS` with `payer=designatedProver`, `payee=actualProver` (even if equal); remove proposer/extended window logic.
> - **L2/Core**:
>   - `Anchor`: drop `provabilityBond` immutable and constructor arg.
>   - `BondManager`: remove `provabilityBond` support; `processBondSignal` slashes 50%:
>     - `payer!=payee`: credit 50% to payee.
>     - `payer==payee`: refund 40% to payer, 10% to caller; burn rest.
>   - `_bondAmountFor` only supports `LIVENESS`.
> - **Shared Libs**:
>   - `LibBonds`: remove `PROVABILITY` from `BondType`.
> - **Tests**:
>   - Update all inbox, bonds, anchor, and bond manager tests to new config and bonding rules; remove extended-window/provability cases; add self-slash and 50% distribution assertions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 99f620123e6c72b748fa607ea02e7b60edae463b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


davidtaikocha (approved):
Changes lgtm. I don't have a strong preference among those 3 options either. And if we burn it, maybe we should burn it after a while without prover bug reports. Since if there is something inside our prover that leads to the slashes, we might will open a proposal to refund the slashed tokens?

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -23,8 +23,6 @@ interface IInbox {
         address proposerChecker;
         /// @notice The proving window in seconds
         uint48 provingWindow;
-        /// @notice The extended proving window in seconds
-        uint48 extendedProvingWindow;
         /// @notice Maximum delay allowed between consecutive proofs to still be on time.
         /// @dev Must be shorter than the expected proposal cadence to prevent backlog growth.
```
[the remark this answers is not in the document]
davidtaikocha: Will update the bindings once it merges.

linoscope:
@ggonzalez94 I like the idea of removing the extended window! It simplifies the protocol, and the extended window is less needed now that we are implementing proper prover-killer mitigations anyway.

For:

> What should we do with the slashed amount? We have 3 options

I like 2., seems good to have option to refund while prover may be unstable.

packages/protocol/contracts/layer1/core/impl/Inbox_Layout.sol
```diff
@@ -7,19 +7,19 @@ pragma solidity ^0.8.24;
 /// @custom:security-contact security@taiko.xyz
 
 // solhint-disable max-line-length
-//   _initialized                   | uint8                                              | Slot: 0    | Offset: 0    | Bytes: 1
```
dantaik, line 10: lets remove Inbox.sol from gen-layout, as MainnetInbox.sol is already included.

packages/protocol/contracts/layer1/core/libs/LibBondInstruction.sol
```diff
@@ -9,72 +9,46 @@ import { LibMath } from "src/shared/libs/LibMath.sol";
...
     )
         internal
         view
         returns (LibBonds.BondInstruction memory bondInstruction_)
     {
         uint256 proofTimestamp = block.timestamp;
         uint256 proposalDeadline = uint256(_proposalTimestamp) + _provingWindow;
-        uint256 sequentialDeadline =
-            uint256(_priorFinalizedTimestamp) + _maxProofSubmissionDelay;
+        uint256 sequentialDeadline = uint256(_priorFinalizedTimestamp) + _maxProofSubmissionDelay;
```
dantaik, line 40: now this lib function is very simple, I suggest we move it to Inbox.sol, then we don't have to pass in the two immutables (_maxProofSubmissionDelay &  _provingWindow) into the funciton.

dantaik (approved):

PR #20928, opened 2025-12-05
chore(protocol): add storage layout files for anchor and ss fork routers

ggonzalez94:
This is a result of the audit concerns that they didn't fully understand the layout for the Anchor and its fork router. Printing the layout as in other contracts should help detect any drifts


davidtaikocha (approved):

dantaik (approved):

packages/protocol/script/gen-layouts.sh
```diff
@@ -25,6 +25,7 @@ contracts_shared=(
 "contracts/shared/common/DefaultResolver.sol:DefaultResolver"
 "contracts/shared/signal/SignalService.sol:SignalService"
 "contracts/shared/fork-router/ForkRouter.sol:ForkRouter"
```
dantaik, line 27: so maybe we remove ForkRouter.sol and Inbox.sol as they not the actual contracts that we will deploy? Keep (Mainnet|DevnetInbox.sol), AnchorForkRouter, etc.

dantaik (changes requested):

dantaik (approved):

PR #20933, opened 2025-12-06
chore(protocol): add whenNotPaused modifier to signal operations

dantaik:
## Summary

Add `whenNotPaused` modifier to signal operations in SignalService contract to enable emergency pause capabilities for critical incident response. This addresses the Missing Paused State Check vulnerability.

## Changes

- Add `whenNotPaused` modifier to `sendSignal` function
- Add `whenNotPaused` modifier to `proveSignalReceived` function  
- Add `whenNotPaused` modifier to `verifySignalReceived` function


davidtaikocha (approved):

packages/protocol/contracts/shared/signal/SignalService.sol
```diff
@@ -83,7 +83,7 @@ contract SignalService is EssentialContract, ISignalService {
     // ---------------------------------------------------------------
 
     /// @inheritdoc ISignalService
-    function sendSignal(bytes32 _signal) external returns (bytes32) {
+    function sendSignal(bytes32 _signal) external whenNotPaused returns (bytes32) {
```
davidtaikocha: Not sure if this is part of the expectation and post here for a double-confirm: after we paused the `SignalService`, this change will also [pause the proof submission of `Inbox.sol`](https://github.com/taikoxyz/taiko-mono/blob/consecutive-proofs/packages/protocol/contracts/layer1/core/impl/Inbox.sol#L233) if there is a bond instruction for the very next proof (even if we haven't paused the `Inbox.sol`), which might will lead to the full ring-buffer.

dantaik: Deleted this `whenNotPaused` for sendSignal

davidtaikocha (approved):

ggonzalez94 (approved):

PR #20934, opened 2025-12-06
chore(protocol): validate HopProof.chainId in SignalService

dantaik:
## Summary
- Add validation to ensure HopProof.chainId matches the current chain ID (block.chainid) in _verifySignalReceived
- Prevents proofs intended for other chains from being accepted
- Update tests to explicitly set the correct chainId in proof structures

## Test Plan
- All 110 existing SignalService and protocol tests pass
- New test_proveSignalReceived_RevertWhen_ChainIdMismatch validates the chainId check


davidtaikocha (approved):

packages/protocol/contracts/shared/signal/SignalService.sol
```diff
@@ -247,6 +247,8 @@ contract SignalService is EssentialContract, ISignalService {
 
         HopProof memory proof = proofs[0];
 
+        if (proof.chainId != uint64(block.chainid)) revert SS_INVALID_CHAIN_ID();
```
ggonzalez94, line 250: Assuming a 2 chain only messaging system L1 <-> L2 do we need this? the storage slot is already calculated by doing `keccak256(abi.encodePacked("SIGNAL", _chainId, _app, _signal));` so the origin chainId is already implicit, meaning you cannot replay the same message on the other direction.

I was actually under the assumption we had deprecated `chainId` from the struct. 

Or is your idea that we wanna keep it so that we potentially support hop messages again?(in which case I think we still need to reintroduce the logic)

ggonzalez94 (commented):
I believe I'm confused about this assumption. Left one comment

packages/protocol/contracts/shared/signal/SignalService.sol
dantaik: I think we shall keep chain id in the struct for readability and security, even if multi-hop is removed.

PR #20935, opened 2025-12-06
feat(protocol): prevent checkpoint overwrite vulnerability

dantaik:
## Summary

Prevents the authorized syncer from overwriting existing checkpoints by silently ignoring overwrite attempts and emitting a `CheckpointOverwriteIgnored` event with both existing and attempted values for full transparency.

## Test plan

- All existing 13 SignalService tests pass
- New test `test_saveCheckpoint_IgnoresOverwriteAndEmitsEvent` verifies the event is emitted and the original checkpoint remains unchanged
- Overwrite attempts no longer revert, maintaining syncer resilience


davidtaikocha (commented):
Also think this can't happen since we [have checks in `Anchor.sol`](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer2/core/Anchor.sol#L442), but approved for extra safety validation (with tradeoff of some gas cost).

PR #20937, opened 2025-12-06
chore(bridge): add validation for non-zero _message.from

dantaik:
## Summary

Add defense-in-depth validation to ensure `_message.from` is never zero when processing bridge messages. This prevents potential downstream issues in contracts relying on `context().from` even though the field is normally set during `sendMessage`. Also updates test messages to properly initialize the `from` field.

## Test Plan

- [x] All existing Bridge tests pass (36 tests)
- [x] Full protocol test suite passes (109 tests)
- [x] New validation correctly rejects messages with zero `from` address


davidtaikocha (approved):
Seems it [can't happen](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/shared/bridge/Bridge.sol#L169), but approved for safety validation (with tradeoff of some gas cost) here.

PR #20939, opened 2025-12-06
feat(bridge): remove isContract check for EIP-7702

dantaik:
## Summary

Removes the `isContract()` check from `_unableToInvokeMessageCall` to address the L-4 audit finding. With EIP-7702 allowing EOAs to have delegated code, the bridge now treats all addresses uniformly when validating function calls.

## Changes

- Remove `_message.to.isContract()` condition that allowed calls to EOAs with arbitrary data
- Update test expectations to reflect that arbitrary function calls are now prohibited for all addresses
- Value is routed to `destOwner` when invocation is prohibited

## Test Plan

- All 42 Bridge contract tests pass
- Verified EOA-specific tests still work (they use empty calldata)
- Verified arbitrary function call rejection for both contracts and EOAs


davidtaikocha (approved):

packages/protocol/contracts/shared/bridge/Bridge.sol
```diff
@@ -595,8 +595,7 @@ contract Bridge is EssentialResolverContract, IBridge {
         if (_message.to == address(_signalService)) return true;
 
         return _message.data.length >= 4
-            && bytes4(_message.data) != IMessageInvocable.onMessageInvocation.selector
-            && _message.to.isContract();
```
davidtaikocha, line 598: And looks like:
```solidity
using Address for address;
```
are also not used in `Bridge.sol` anymore, shall we remove them too in this PR?

dantaik: fixed

davidtaikocha (approved):

ggonzalez94 (approved):
LGTM. Where is this audit finding? Since the bridge was not in scope for this audit I guess we are refering to an older audit report?

PR #20943, opened 2025-12-06
chore(taiko-client): remove Shasta `Proved` event handling in prover

davidtaikocha:
Since for the new protocol design: https://github.com/taikoxyz/taiko-mono/pull/20884 , all proofs will be finalized when landed on-chain, so the prover client doesn't need to monitor that `Proved` event anymore to see if there is a need to submit a new proof to contest.


davidtaikocha:
@claude plz review.

MatusKysel (approved):

PR #20946, opened 2025-12-07
fix(eventindexer): use BytesToAddress for event log address extraction

hawkadrian:
### What
Replaced fmt.Sprintf("0x%v", common.Bytes2Hex(...)) with common.BytesToAddress(...).Hex() for extracting addresses from event log topics.
### Why
Uses the idiomatic go-ethereum approach, reduces allocations, and matches patterns used elsewhere in the codebase.


davidtaikocha:
Can't see too much valuable improvements here: https://github.com/taikoxyz/taiko-mono/blob/main/CONTRIBUTING.md

PR #20947, opened 2025-12-07
fix: use RFC6979 nonce in driver fixed-k signer

anim001k:
Replace hardcoded k=1/2 path for main signing with RFC6979-derived nonce to prevent private-key recovery from deterministic nonces.


davidtaikocha:
It's part of the network consensus. We can't change this for now unless we have a hardfork.

PR #20949, opened 2025-12-08
feat(protocol): consolidate Inbox.sol and reimplemented prove() func

dantaik:
## Summary

- **Merge InboxOptimized into Inbox.sol** - single Inbox implementation
- **Rename CodecOptimized to Codex** - cleaner naming, delete CodecSimple.sol
- **Delete LibHashSimple.sol** - no longer needed after merge
- **Move calculateBondInstruction to Inbox.sol** - delete LibBondInstruction.sol
- **Rename codec libraries to consistent naming**:
  - LibProposedEventEncoder → LibProposedEventCodec
  - LibProposeInputDecoder → LibProposeInputCodec
  - LibProvedEventEncoder → LibProvedEventCodec
  - LibProveInputDecoder → LibProveInputCodec
- **Remove unused hash functions from LibHashOptimized and ICodec**:
  - hashCheckpoint (not used by Inbox)
  - hashCoreState (not used by Inbox)
  - hashBlobHashesArray (dead code)
  - _hashDerivationSource (dead private function)
- **Delete LibHashParity.t.sol** - tested removed code
- **Optimize hashBondInstruction** with new tests
- **Move hashBondInstruction from LibHashOptimized to LibBonds**
- **replace string errors with custom errors**
- **add parentProposalHash to Proposal**
- **remove `Transition` and implement `prove()` differently**
- **require prover to prove enough proposals or the last proposal proved is young enough**


packages/protocol/contracts/layer1/core/impl/Codex.sol
```diff
@@ -4,18 +4,18 @@ pragma solidity ^0.8.24;
...
-import { LibProvedEventEncoder } from "../libs/LibProvedEventEncoder.sol";
-import { ICheckpointStore } from "src/shared/signal/ICheckpointStore.sol";
+import { LibProposeInputCodec } from "../libs/LibProposeInputCodec.sol";
+import { LibProposedEventCodec } from "../libs/LibProposedEventCodec.sol";
+import { LibProveInputCodec } from "../libs/LibProveInputCodec.sol";
+import { LibProvedEventCodec } from "../libs/LibProvedEventCodec.sol";
+import { LibBonds } from "src/shared/libs/LibBonds.sol";
 
-/// @title CodecOptimized
+/// @title Codex
```
davidtaikocha: Shall we call it `Codec` instead of `Codex` here?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -812,21 +710,10 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
         private
         returns (bytes32 signal_)
     {
-        signal_ = _hashBondInstruction(_bondInstruction);
+        signal_ = LibHashOptimized.hashBondInstruction(_bondInstruction);
```
davidtaikocha: If we change so, we might also need to update [`_bondSignalHash` in `BondManager.sol`](https://github.com/taikoxyz/taiko-mono/blob/consecutive-proofs/packages/protocol/contracts/layer2/core/BondManager.sol#L297-L303)?

dantaik: Done

packages/protocol/contracts/layer1/core/impl/Codex.sol
dantaik: Done

davidtaikocha (approved):

packages/protocol/contracts/shared/libs/LibBonds.sol
```diff
@@ -27,19 +29,26 @@ library LibBonds {
...
     // ---------------------------------------------------------------
-    // Internal Functions
+    // Functions
     // ---------------------------------------------------------------
 
-    function aggregateBondInstruction(
-        bytes32 _bondInstructionsHash,
-        BondInstruction memory _bondInstruction
-    )
+    /// @dev Hashing for BondInstruction structs using EfficientHashLib
```
ggonzalez94: I wonder if using `EfficientHashLib` for the bonds is not adding unnecessary complexity. Since bonds are only sent during slashing conditions, maybe we can just use `keccak256` and reduce the amount of code?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -859,12 +746,65 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+    ///         - Late (within extendedProvingWindow): Liveness bond transfer if prover differs
+    ///           from designated prover of the first transition.
+    ///         - Very late (after extendedProvingWindow): Provability bond transfer if prover
+    ///           differs from proposer of the first transition.
+    /// @param _firstProposal The first proposal proven in the batch.
+    /// @param _firstTransition The transition for the first proposal.
+    /// @param _readyTimestamp Timestamp when the first proposal became proveable.
+    /// @return bondInstruction_ A bond transfer instruction, or a BondType.NONE instruction when
+    ///         no transfer is required.
+    function _calculateBondInstruction(
```
ggonzalez94: I believe I still prefer to have this logic encapsulated in  a separate `LibBondInstruction` library. But in any case shall we wait for PRs ##20926 and #20927 to be resolved and merge it if we are going to move it? it will make the merge conflicts easier to manage

packages/protocol/test/layer1/core/inbox/InboxRingBuffer.t.sol
```diff
@@ -1,21 +1,19 @@
...
-
+contract InboxRingBufferTest is InboxTestBase {
     function _buildConfig() internal override returns (IInbox.Config memory cfg) {
         cfg = super._buildConfig();
         // Need headroom for 10-item batches in ring-buffer tests.
         cfg.ringBufferSize = 16;
         return cfg;
     }
 
-    /// forge-config: default.isolate = true
```
ggonzalez94, line 18: why remove this? it is necessary for foundry to capture accurate gas snapshots according to their [docs](https://getfoundry.sh/forge/gas-tracking/gas-section-snapshots/)

packages/protocol/test/layer1/core/inbox/InboxRingBuffer.t.sol
```diff
@@ -46,12 +44,68 @@ abstract contract RingBufferTestBase is ProveTestBase {
             inbox.getProposalHash(p6.proposal.id), codec.hashProposal(p6.proposal), "proposal hash"
         );
     }
-}
 
-contract InboxRingBufferProveTest is RingBufferTestBase {
-    constructor() RingBufferTestBase(InboxVariant.Simple) { }
-}
+    // ---------------------------------------------------------------------
```
ggonzalez94: I guess we need all these helpers since `InboxProveTest` is not abstract anymore. Maybe we can still move them to a base clase to avoid the code repetition?

packages/protocol/test/shared/libs/LibBonds.t.sol
```diff
@@ -0,0 +1,245 @@
...
+        });
+
+        bytes32 hash = LibBonds.hashBondInstruction(instruction);
+
+        // Verify the hash is not zero
+        assertTrue(hash != bytes32(0), "hash should not be zero");
+
+        // Verify determinism - same input should produce same output
+        bytes32 hash2 = LibBonds.hashBondInstruction(instruction);
+        assertEq(hash, hash2, "hash should be deterministic");
```
ggonzalez94: is this actually adding any value?

packages/protocol/test/shared/libs/LibBonds.t.sol
```diff
@@ -0,0 +1,245 @@
...
+        LibBonds.BondInstruction memory instruction = LibBonds.BondInstruction({
+            proposalId: 0, bondType: LibBonds.BondType.NONE, payer: address(0), payee: address(0)
+        });
+
+        bytes32 hash = LibBonds.hashBondInstruction(instruction);
+
+        assertTrue(hash != bytes32(0), "hash of zero values should not be zero");
+    }
+
+    function test_hashBondInstruction_maxValues() public pure {
```
ggonzalez94: not sure if valuable?

packages/protocol/test/shared/libs/LibBonds.t.sol
```diff
@@ -0,0 +1,245 @@
...
+            payer: address(type(uint160).max),
+            payee: address(type(uint160).max)
+        });
+
+        bytes32 hash = LibBonds.hashBondInstruction(instruction);
+
+        assertTrue(hash != bytes32(0), "max values should hash correctly");
+    }
+
+    function test_hashBondInstruction_samePayer_andPayee() public pure {
```
ggonzalez94: same: not sure if valuable? obviously just setting the same payer and payee won't result in a zero hash

packages/protocol/test/shared/libs/LibBonds.t.sol
```diff
@@ -0,0 +1,245 @@
...
+        bytes32 hash = LibBonds.hashBondInstruction(instruction);
+
+        assertTrue(hash != bytes32(0), "same payer and payee should hash correctly");
+    }
+
+    // ---------------------------------------------------------------
+    // Fuzz Tests for hashBondInstruction
+    // ---------------------------------------------------------------
+
+    function testFuzz_hashBondInstruction_deterministic(
```
ggonzalez94: I don't think this test is valuable, same as the one above without hashing. It just checks that hashing the same thing twice is deterministic. I don't know how this can not be the case :D

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -235,9 +238,8 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
 
         _state = result.newState;
 
-        _proofVerifier.verifyProof(result.proposalAge, result.aggregatedTransitionHash, _proof);
-
         _emitProvedEvent(input, result, bondSignal);
+        _proofVerifier.verifyProof(result.proposalAge, result.aggregatedTransitionHash, _proof);
```
ggonzalez94: I'm perfectly fine with this, but I'm just curious why calling the verifier after emitting the event is preferable?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -859,12 +746,65 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+            });
+        }
+    }
+
     // ---------------------------------------------------------------
     // Errors
     // ---------------------------------------------------------------
 
     error ActivationPeriodExpired();
-    error ActivationRequired();
```
ggonzalez94, line 868: we might need to bring this error back. I think this is one check that my big PR is missing(ensuring you cannot propose before the activation)

ggonzalez94 (commented):
Left a few comments, and I think some of the tests that were generated are too redundant.

But I do like merging both inbox and just having a single contract

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: I had the impression that calling another contract at the end of a function is better/safer. But maybe i'm wrong.

packages/protocol/test/shared/libs/LibBonds.t.sol
dantaik: deleted with a few other tests.

packages/protocol/test/shared/libs/LibBonds.t.sol
dantaik: Deleted

packages/protocol/test/shared/libs/LibBonds.t.sol
dantaik: Deleted

packages/protocol/test/shared/libs/LibBonds.t.sol
dantaik: Deleted

packages/protocol/test/layer1/core/inbox/InboxRingBuffer.t.sol
dantaik: added back

dantaik:
@ggonzalez94 your comments addressed, please take another look.

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -124,16 +111,32 @@ interface IInbox {
...
 
+    /// @notice Metadata for a proposal used in prove
+    struct ProposalState {
+        /// @notice Address of the proposer.
+        address proposer;
+        /// @notice Address of the designated prover.
+        address designatedProver;
+        /// @notice Timestamp of the proposal.
+        uint48 timestamp;
+        /// @notice Transition hash for the proposal.
```
davidtaikocha:
```suggestion
        /// @notice Last block hash for the proposal.
```

dantaik:
@claude please review

packages/protocol/contracts/layer1/core/libs/LibBondInstruction.sol
```diff
@@ -1,71 +1,64 @@
...
         uint48 _provingWindow,
-        uint48 _extendedProvingWindow,
-        IInbox.Proposal memory _firstProposal,
-        IInbox.Transition memory _firstTransition,
-        uint48 _readyTimestamp
+        uint48 _extendedProvingWindow
     )
         internal
-        view
+        pure
```
dantaik, line 39: @ggonzalez94 if you further remove `_extendedProvingWindow`, this function will be even simpler, so maybe it makes sense to move it inside Inbox.sol?

dantaik:
@claude please review -- ignore storage layout compatibility issues.

MatusKysel (approved):

dantaik:
@ggonzalez94 It would be great if you could merge this PR into your consecutive-proofs branch and then rebase your other PRs on top of it, so I can switch myself into review-only mode :D

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: oh I think to avoid reentrancy(if the contract is untrusted) you want to do it after you run the checks and update state. But I think before or after the event it should make no difference

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -42,6 +41,8 @@ interface IInbox {
         /// @notice The multiplier to determine when a forced inclusion is too old so that proposing
         /// becomes permissionless
         uint8 permissionlessInclusionMultiplier;
+        /// @notice The minimum number of proposals that must be finalized in a single prove call
+        uint8 minProposalsToFinalize;
```
ggonzalez94, line 45: Does the design need this? I think we will better off if we don't introduce this constrain. While we do want to favor aggregation in most cases, in an adversarial environment we still want to allow proposers to submit individual proofs

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -109,7 +96,7 @@ interface IInbox {
         /// @dev In genesis block, this is set to 0 to allow the first checkpoint to be saved.
         uint48 lastCheckpointTimestamp;
         /// @notice The hash of the last finalized transition.
```
ggonzalez94:
```suggestion
        /// @notice The block hash of the last finalized L2 block.
```

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -16,8 +21,6 @@ import { LibMath } from "src/shared/libs/LibMath.sol";
 import { ICheckpointStore } from "src/shared/signal/ICheckpointStore.sol";
 import { ISignalService } from "src/shared/signal/ISignalService.sol";
 
-import "./Inbox_Layout.sol"; // DO NOT DELETE
```
ggonzalez94, line 19: I don't think we wanna remove this?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -158,6 +149,15 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
         _forcedInclusionFeeDoubleThreshold = _config.forcedInclusionFeeDoubleThreshold;
         _minCheckpointDelay = _config.minCheckpointDelay;
         _permissionlessInclusionMultiplier = _config.permissionlessInclusionMultiplier;
+        _minProposalsToFinalize = _config.minProposalsToFinalize;
+    }
+
+    // ---------------------------------------------------------------
+    // Modifiers
+    // ---------------------------------------------------------------
+    modifier onlyWhenActivated() {
```
ggonzalez94: I'm not sure I like the idea of using a modifier. While it does make the code slightly cleaner it causes another `SLOAD` before the function runs(maybe not if the compiler is smart enough?)

Also I think we only need this on `propose`, it is probably not needed on `prove` since yo need an existing proposal to exist

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -211,39 +221,144 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+            _proofVerifier.verifyProof(proposalAge, hashToProve, _proof);
+        }
     }
 
     /// @inheritdoc IForcedInclusionStore
-    /// @dev This function will revert if called before the first non-activation proposal is submitted
-    /// to make sure blocks have been produced already and the derivation can use the parent's block timestamp.
+    /// @dev This function will revert if called before the first non-activation proposal is
+    /// submitted to make sure blocks have been produced already and the derivation can use the
+    /// parent's block timestamp.
```
ggonzalez94, line 402: I see we've split multiple natspec comment sentences into different lines, but I'm not sure if they were actually hitting the length limit. Was this intentional or forge fmt or claude did it?

packages/protocol/contracts/layer1/core/iface/IInbox.sol
ggonzalez94: I actually see that in the `MainnetInbox` configuration we've set this to 1. So maybe we can just remove it? I think it complicates the reasoning for the incentives and how the unfinalized proposals might grow

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -211,39 +221,144 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+                    _provingWindow,
+                    _extendedProvingWindow
+                );
+            if (bondInstruction.bondType != LibBonds.BondType.NONE) {
+                _signalService.sendSignal(LibBonds.hashBondInstruction(bondInstruction));
+                emit BondInstructionCreated(bondInstruction.proposalId, bondInstruction);
+            }
+
+            // ---------------------------------------------------------
+            // 5. Sync checkpoint if minimum delay has passed
```
ggonzalez94: don't you think we should still provide a way for the prover to still sync the checkpoint if they want to do so to enable faster withdrawals and L2->L1 messages? I agree we don't want to force them, but as a general rule I think we should always allow them to do it if it will help the network. And they might have incentives to still do it(e.g. they are or partner with a solver)

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -194,10 +196,18 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
-    function propose(bytes calldata _lookahead, bytes calldata _data) external nonReentrant {
+    /// NOTE: This function can only be called once per block to prevent spams that can fill the
+    /// ring buffer.
+    function propose(
+        bytes calldata _lookahead,
+        bytes calldata _data
+    )
+        external
+        onlyWhenActivated
+        nonReentrant
```
ggonzalez94: nit: now that there are no funds being moved we can safely remove the `nonReentrant` modifier. Same for the `prove` function

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -211,39 +221,144 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+                    proposalAge,
+                    firstProposal.proposer,
+                    firstProposal.designatedProver,
+                    input.actualProver,
+                    _provingWindow,
+                    _extendedProvingWindow
+                );
+            if (bondInstruction.bondType != LibBonds.BondType.NONE) {
+                _signalService.sendSignal(LibBonds.hashBondInstruction(bondInstruction));
+                emit BondInstructionCreated(bondInstruction.proposalId, bondInstruction);
```
ggonzalez94: do we need an extra event? Maybe there are a couple more efficients way to do this(although I don't have a strong opinion on this).
1. Provers interested on this can subscribe to the SS events and just filter based on messages coming from this contract.
2. We could incoporate it into the final `ProvedEvent`

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -75,26 +76,12 @@ interface IInbox {
         uint48 endOfSubmissionWindowTimestamp;
         /// @notice Address of the proposer.
         address proposer;
+        /// @notice Hash of the parent proposal (zero for genesis).
+        bytes32 parentProposalHash;
```
ggonzalez94, line 80: we don't need to check the `parentProposalHash`on-chain when calling`prove` because the prover enforces this right?

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -109,7 +96,7 @@ interface IInbox {
         /// @dev In genesis block, this is set to 0 to allow the first checkpoint to be saved.
         uint48 lastCheckpointTimestamp;
         /// @notice The hash of the last finalized transition.
-        bytes32 lastFinalizedTransitionHash;
+        bytes32 lastFinalizedBlockHash;
```
ggonzalez94: would it make sense to use lastFinalizedCheckpointHash here? that we can use a single abstraction across the codebase?

dantaik:
@claude please review

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -194,56 +192,214 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
     ///      2. Process `input.numForcedInclusions` forced inclusions. The proposer is forced to
     ///         process at least `config.minForcedInclusionCount` if they are due.
     ///      3. Updates core state and emits `Proposed` event
-    /// NOTE: This function can only be called once per block to prevent spams that can fill the ring buffer.
+    /// NOTE: This function can only be called once per block to prevent spams that can fill the
+    /// ring buffer.
     function propose(bytes calldata _lookahead, bytes calldata _data) external nonReentrant {
```
ggonzalez94, line 197: I just noted this change, but I actually think the `propose` function was simpler to follow before

ggonzalez94 (commented):
Leaving some comments I already had before we merged the PR. I'll be submitting a new PR to address some of my thoughts/concerns

PR #20951, opened 2025-12-08
feat(protocol): another way to prove and finalize proposals [do-not-merge]

dantaik:
@ggonzalez94 lets discuss this new approach of proving and finalizing proposals.

We need this [PR](https://github.com/taikoxyz/taiko-mono/pull/20952) so that prover can rely on the last proposal's hash to verify the ProveInput2 object.


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -245,6 +246,98 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+            // - Must start at or before the last finalized proposal (to include its transition hash)
+            // - Must advance past the current finalization point
+            // - Cannot prove proposals that haven't been proposed yet
+            require(
+                firstProposalParentId <= state.lastFinalizedProposalId, ProofRangeStartTooLate()
+            );
+            require(
+                input.lastProposalId >= state.lastFinalizedProposalId + _minProposalsToFinalize,
+                InsufficientProposalsToFinalize()
+            );
```
davidtaikocha, line 262: If we introduce this hard limit, is there a chance that, a user only wants to prove and finalize his / her own proposals? Like the current `lastFinalizedProposalId`  is `100`, Alice proposed `101`, and Bob proposed `102`, but Alice only wants to submit the proof of her own `101` in time ASAP and doesn't want to generate the proofs for Bob?

dantaik: then the chain will finalize very slowly and eventual the ring buffer will be exhausted. we have to enforce each proof to prove enough proposals so finalization can catch up with new proposals.

PR #20967, opened 2025-12-09
feat(protocol): remove MIN_ANCHOR_OFFSET from Shasta derivation

linoscope:
## Context

`MIN_ANCHOR_OFFSET` was introduced (https://github.com/taikoxyz/taiko-mono/issues/20059) to prevent unintended slashing in earlier preconfirmation designs, which preconfirmed blockhash rather than rawTxList. At the time, a preconfer could be forced to reference a very recent anchor block because of how the previous preconfer selected their anchor block. If that recent anchor block was later reorged, the preconfer could be unfairly slashed. `MIN_ANCHOR_OFFSET` ensured that anchor blocks were always sufficiently old to avoid this situation.

However, now that we use the `rawTxList`+`anchorBlockNumber` approach, we do not face this issue. Hence `MIN_ANCHOR_OFFSET` is no longer needed.

On top of enabling faster L1->L2 messaging, removing `MIN_ANCHOR_OFFSET` also has the benefit of simplifying "per-block" derivation. If we had `MIN_ANCHOR_OFFSET` in the protocol, then the following can happen:
- Preconfer preconfs L2 block with anchor block `100` when the latest head is `100`.
- Driver accepts that block. Note that driver cannot know when this L2 block will be proposed to L1 at this point.
- Preconfer, immediately propose that L2 block to L1 in L1 block `101`.
- Cause reorg in L2, as the block is accepted at preconf time while being filtered out at proposal time.
- Preconfer will not be slashed as he proposed the exact `rawTxList` and `anchorBlockNumber` that he preconfed.

## Changes

- Remove `MIN_ANCHOR_OFFSET` from protocol: https://github.com/taikoxyz/taiko-mono/pull/20967/commits/6fd1db5d3def9aa9cb323bdcfcc83566c08f7be1
- Introduce `anchorOffset` as a proposer config: https://github.com/taikoxyz/taiko-mono/pull/20967/commits/ede642b2aa4dc0e2eb35167844b174bfee314f5d

## Notes

- If we merge this, we must communicate with preconf implementors as they have `MIN_ANCHOR_OFFSET` checks at sidecar side (at least Nethermind does).


smtmfft (approved):

MatusKysel (approved):
lgtm

dantaik (approved):

dantaik:
@linoscope thank you for creating this PR ❤️

davidtaikocha (approved):

ggonzalez94 (approved):
LGTM. Thanks!

davidtaikocha (approved):

PR #20973, opened 2025-12-10
Implement lastProposalHash field and optimize hashProveInput

dantaik:
## Summary

- Added `lastProposalHash` field to `ProveInput` struct for explicit verification of the last proposal hash during proof validation
- Re-implemented `hashProveInput()` using `EfficientHashLib` for optimized gas usage instead of ABI encoding
- Updated all codec libraries (`LibProveInputCodec`, `LibProvedEventCodec`) to properly encode/decode the new field
- Moved `lastFinalizedBlockHash` out of `CoreState` into its own storage slot for gas optimization
- Enabled single SLOAD/SSTORE pattern in `propose()` and `prove()` functions
- Added `getLastFinalizedBlockHash()` getter function to interface

## Test plan

- [x] 69 inbox tests pass
- [x] 12 codec tests pass
- [x] All encoding/decoding roundtrips verified
- [x] Hash verification logic tested in proof validation path


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -441,10 +432,14 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
         });
     }
 
-    /// @notice Returns the core state as a struct (convenience wrapper for tests).
-    /// @dev The public `coreState` variable returns a tuple; this returns the struct.
-    function getCoreState() external view returns (CoreState memory) {
-        return coreState;
```
davidtaikocha: I think right now `getCoreState` is really widely used for the current off-chain client software / preconf sidecar, which is a part of the reason that they can get rid of the off-chain indexer component. If possible, let's keep this method public.

dantaik: added back.

PR #20926, opened 2025-12-05
fix(protocol): fix unbounded finalization by introducing maxProofSubmissionDelay

ggonzalez94:
## Context

The major concern of introducing sequential proving was that it makes the number of unfinalized proposals grow unbounded, **if all proposer wait until the very end of their proving window and no "altruistic" prover comes and proves bigger chunks**.  This is because the cadence of proposals(around 6min is much faster than the `provingWindow` - currently configured to 4 hours).
In practice Taiko will operate a proving service that will generate and submit proofs for all proposers, and we expect other participants that care about the health of the network to do the same. **But we still want to provide an incentive compatible solution, that even in the extreme case that no one does it, the network is still resilient**.

This is the growth un unfinalized proposals over time under the conditions described above:

image

_(you can play with the simulation generated by claude [here](https://claude.ai/public/artifacts/848ff69e-bbdf-4e7b-bbbf-eaf4009b4ca8))_
## Summary 

This PR fixes this issue by introducing `maxProofSubmissionDelay`(Props to @linoscope  for the magnific idea), which if chosen to be lower than the expected average proving cadence changes the shape of the curve of unfinalized proposals. It goes from a linear curve that grows unbounded to a curve that grows until `provingWindow` is hit, and then becomes constant.

This is the expected behavior after introducing `maxProofSubmissionDelay` and using the formula `proof_deadline(proposal_N) = max(timestamp_of(proposal_N) + livenessWindow, timestamp_of(proof_of(proposal_N-1)) + maxProofSubmissionDelay)` 

image

_(you can play with the simulation generated by claude [here](https://claude.ai/public/artifacts/0b094e33-3417-4d4a-8480-95ab71263a40))_

**KEY INSIGHT: YOU HAVE TO CHOSE A `maxProofSubmissionDelay` SMALLER THAN THE EXPECTED PROPOSAL CADENCE**

## TODO
- [x] Implement #20883 
        https://github.com/taikoxyz/taiko-mono/pull/20927
- [x] Remove `extendedProvingWindow`. It adds another variable at the formula that makes it harder to reason about, and arguably it does not add as much benefit(we can make `livenessBond` larger)
       https://github.com/taikoxyz/taiko-mono/pull/20927

---

> [!NOTE]
> Adds maxProofSubmissionDelay to bound sequential proving cadence and updates bond instruction logic, configs, and tests accordingly.
> 
> - **Core (Inbox)**:
[...]

PR #20981, opened 2025-12-10
feat(protocol): remove minProposalsToFinalize, allow checkpoint syncing ahead of time and several renames and refactors

ggonzalez94:
This PR addresses several of the comments I left on #20949 that were too late before it got merged.

## Summary
- Remove `minProposalsToFinalize` since it is not compatible with the incentive system proposed in #20926. More details [here](https://github.com/taikoxyz/taiko-mono/pull/20949/changes#r2604166452)
- Rename `ProposalState` to `Transition` to better reflect the previous terminology. ProposalState suggests on-chain storage, which is missleading
- Use `checkpointHash` instead of `blockHash` and use checkpoint instead of its individual fields. This makes the struct more concise to read and avoids having multiple abstractions
- Re-introduce the ability for provers to sync the checkpoint before `minCheckpointDelay` has passed(reasoning [here](https://github.com/taikoxyz/taiko-mono/pull/20949#discussion_r2604331998))
- Revert to the previous `propose` flow. I have less of a strong opinion here, but I think it was in general more readable. Using step by step strict comments IMO should only be used when there's a complex function that is very hard to follow, which is not the case of the propose function
- Split `prove` into smaller private functions. We were very close to the limit of stack too deep, which became apparent when adding back the check if the checkpoint should be synced or not
- Other improvements in natspec and readability

---

> [!NOTE]
> Replaces block-hash based proving with checkpoint-based `Transition`s, removes `minProposalsToFinalize`, refactors `prove/propose` and hashing/codec layers, and updates state/events/tests accordingly.
> 
> - **Core (Inbox)**:
>   - Replace block-hash continuity with checkpoint continuity using `Transition{checkpointHash}` and `CoreState.lastFinalizedCheckpointHash`.
>   - Remove `minProposalsToFinalize` config/logic and related errors/branches.
>   - Allow checkpoint syncing before `minCheckpointDelay` if `lastCheckpoint` is provided; otherwise enforce delay.
>   - Refactor `prove` into helpers (`_validateBatchBoundsAndCalculateOffset`, `_verifyProof`) and split `propose` building into `_buildProposal`; add `_validateProposeInput`.
>   - Remove `lastFinalizedBlockHash`; rename activation event payload to `lastPacayaCheckpointHash`.
> - **Interfaces/Structs**:
>   - Rename `ProposalState` → `Transition`; `ProveInput` now has `firstProposalParentCheckpointHash`, `transitions[]`, and `lastCheckpoint`.
>   - `CoreState` adds `lastFinalizedCheckpointHash`.
> - **Hashing/Codec**:
>   - Add `hashCheckpoint` and new `hashProveInput(lastProposalHash, ProveInput)`; update `LibHashOptimized` accordingly.
>   - Update `LibProveInputCodec`/`LibProvedEventCodec` to encode/decode new `ProveInput`/`Transition`/checkpoint fields.
> - **Configs/Deployments**:
>   - Remove `minProposalsToFinalize` from `Config`; update Devnet/Mainnet constructors and layout docs.
> - **Tests & Snapshots**:
>   - Rewrite prove-related tests to use checkpoints and `Transition`s; add new validations (parent checkpoint mismatch, checkpoint hash mismatch, missing after delay).
>   - Update gas snapshots for propose/prove benchmarks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e329bb2c13eccb82b7b9928036e43d467643b6ff. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/core/iface/ICodec.sol
```diff
@@ -125,7 +125,11 @@ interface ICodec {
         returns (bytes32);
 
     /// @notice Hashing for prove input data combining proposal hash and prove input
+    /// @param _lastProposalHash The hash of the last proposal in the batch
     /// @param _input The prove input to hash
-    /// @return The hash of the prove input
-    function hashProveInput(IInbox.ProveInput calldata _input) external pure returns (bytes32);
+    /// @return The hash of the prove input payload
+    function hashProveInput(bytes32 _lastProposalHash, IInbox.ProveInput calldata _input)
```
ggonzalez94: I actually liked the single input version, but I saw it meant we should add it to the `ProveInput` struct and still check that it matches on-chain. So it seems like a better tradeoff to have this function with 2 inputs?

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -183,31 +182,8 @@ interface IInbox {
     function propose(bytes calldata _lookahead, bytes calldata _data) external;
 
     /// @notice Verifies a batch proof covering multiple consecutive proposals and finalizes them.
-    /// @dev The proof covers a contiguous range of proposals. The input contains an array of
```
ggonzalez94, line 186: all this is implementation specific and is better to have it in the implementation contract IMO

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -102,22 +101,16 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
 
     /// @notice The timestamp when the first activation occurred.
     uint48 public activationTimestamp;
 
-    /// @notice Persisted core state (all uint48 fields packed into single slot).
+    /// @notice Persisted core state.
     CoreState internal _coreState;
 
-    /// @notice The hash of the last finalized block (separate slot for gas optimization).
-    bytes32 public lastFinalizedBlockHash;
```
ggonzalez94, line 119: I believe `lastFinalizedBlockHash`(or now `lastFinalizedCheckpointHash`) is conceptually part of the CoreState of the chain. So it should be part of the same struct, we can just be mindful when retrieving it to optimize gas

packages/protocol/snapshots/shasta-prove.json
```diff
@@ -1,6 +1,7 @@
 {
-  "prove_batch_2": "20041",
-  "prove_consecutive_10": "35322",
```
ggonzalez94, line 3: I think there was something wrong with these previous benches. I don't think they were correct

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -196,203 +187,131 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
-                        proposalAge,
-                        firstProposal.proposer,
-                        firstProposal.designatedProver,
-                        input.actualProver,
-                        _provingWindow,
-                        _extendedProvingWindow
-                    );
-                if (bondInstruction.bondType != LibBonds.BondType.NONE) {
-                    _signalService.sendSignal(LibBonds.hashBondInstruction(bondInstruction));
-                    emit BondInstructionCreated(bondInstruction.proposalId, bondInstruction);
```
ggonzalez94, line 361: I removed the BondInstructionCreated event. Provers should be able to listen to the SS event and filter for messages coming from the Inbox contract,

dantaik (approved):

davidtaikocha (approved):

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: But on L2, there is no easy way to recover the bondInstruction associated with an event, the only way is to use tracing, which is very inconvenient. Given that bond instruction is not frequent, we should emit the event -- same reason we didn't optimize hashBondInstruction

PR #20986, opened 2025-12-11
feat(protocol): proposal a few additional changes

dantaik:
- **Remove on-chain checkpoint verification**  
  The validity and integrity of `ProveInput` is now assumed to be fully enforced by the off-chain prover. The `hashCheckpoint` field and its on-chain verification logic have been removed.

- **Add `forceCheckpointSync` to `ProveInput`**  
  Introduce a `bool forceCheckpointSync` field in the `ProveInput` struct. This allows callers to explicitly force checkpoint synchronization even if the minimum checkpoint delay has not yet elapsed. Codec size calculations have been updated to account for the extra 1-byte field.

- **Introduce `Commitment` struct**  
  Add a `Commitment` struct to encapsulate all proof-related data that the verifier actually cares about. Configuration / option parameters that are not part of the proof itself remain in `ProveInput`.

- **Add `lastProposalHash` to `Commitment`**  
  `lastProposalHash` is now part of `Commitment` instead of being read from storage. This makes the commitment self-contained and simplifies off-chain prover implementation by avoiding mixed on-chain / off-chain sources of truth.

- **Reuse `getProposalHash`**  
  Reuse the existing `getProposalHash` helper to slightly simplify the code and avoid duplication.

- **Restore `BondInstructionCreated` event**  
  The `BondInstructionCreated` event has been reintroduced because L2 still relies on it to process bond instructions. Reconstructing all bond instructions purely from signals would be difficult. Since we expect this event to be emitted infrequently, keeping it provides better ergonomics with negligible overhead.

---

> [!NOTE]
> Replaces checkpoint-based prove inputs with a `Commitment` hashed on-chain, validates block-hash linkage and last proposal hash, adds optional forced checkpoint sync, restores bond instruction event, and updates codecs/verifiers/tests accordingly.
> 
> - **Core (Inbox/Interfaces)**:
>   - Replace checkpoint continuity with block-hash continuity: `CoreState.lastFinalizedBlockHash` (was `lastFinalizedCheckpointHash`).
>   - Introduce `Commitment` in `IInbox.ProveInput`; prove validates parent `blockHash` and `lastProposalHash` via `getProposalHash(uint256)`.
>   - Add `forceCheckpointSync` to allow early checkpoint saving; checkpoint saved via `ICheckpointStore.Checkpoint{blockNumber,endStateRoot,blockHash}`.
>   - Revamp `prove`: `_validateCommitment`, compute bond instruction, emit `Proved` with new payload, and call verifier with `hashCommitment`.
>   - Restore `BondInstructionCreated` event.
> - **Codec/Hashing**:
>   - Add `ICodec.hashCommitment`; remove checkpoint/prove-input hashing from codec/lib; update `LibHashOptimized` to hash `Commitment` and transitions' `blockHash`.
>   - Update `LibProveInputCodec` and `LibProvedEventCodec` to encode/decode `Commitment` and `forceCheckpointSync` (size constants: 131 + 78/transition).
> - **Verifiers**:
>   - `IProofVerifier.verifyProof` now takes `_commitmentHash`; update `OpVerifier` and composed verifier to pass through.
> - **Misc**:
>   - `Inbox.activate` and event use last Pacaya block hash.
>   - Foundry config: exclude `test/genesis/*`.
>   - Tests and gas snapshots updated to new structures and checks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 71e100286b0b7b1d813c8191ba60b3e97324aef4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


davidtaikocha (approved):
Change wise overall looks correct to me, left one comment.

packages/protocol/contracts/layer1/core/impl/Codec.sol
```diff
@@ -121,15 +120,6 @@ contract Codec is ICodec {
...
     }
 
-    /// @inheritdoc ICodec
-    function hashCheckpoint(ICheckpointStore.Checkpoint calldata _checkpoint)
-        external
-        pure
-        returns (bytes32)
-    {
-        return LibHashOptimized.hashCheckpoint(_checkpoint);
-    }
```
davidtaikocha, line 131: If we remove this from `Codec` and `LibHashOptimized`,but when submitting a proof, in `Commitment.transitions`, we still need to calculate each `transition.checkpointHash`?

dantaik: See https://github.com/taikoxyz/taiko-mono/pull/20988

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -243,75 +243,74 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
             // -------------------------------------------------------------------------------
             // 1. Validate batch bounds and calculate offset of the first unfinalized proposal
             // -------------------------------------------------------------------------------
+            Commitment memory c = input.commitment;
```
ggonzalez94: nit: single letter variable names are harder to reason about
```suggestion
            Commitment memory commitment = input.commitment;
```

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -243,75 +243,74 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
+            // 4. Sync checkpoint
             // -----------------------------------------------------------------------------
-            if (input.lastCheckpoint.blockHash != 0) {
-                require(
-                    input.transitions[numProposals - 1].checkpointHash
-                        == LibHashOptimized.hashCheckpoint(input.lastCheckpoint),
-                    CheckpointHashMismatch()
-                );
-                _signalService.saveCheckpoint(input.lastCheckpoint);
+            if (
```
ggonzalez94, line 291: just curious why you think the extra flag is necessary. Less error prone than relying on an "empty" checkpoint?

packages/protocol/snapshots/shasta-prove.json
```diff
@@ -1,7 +1,7 @@
 {
-  "prove_batch_10": "91114",
-  "prove_batch_2": "75991",
-  "prove_batch_3": "78006",
-  "prove_batch_5": "81768",
-  "prove_single": "74268"
+  "prove_batch_10": "36201",
```
ggonzalez94: now the benches only count for the path where we don't store the checkpoint. I'll duplicate them for when we do.

ggonzalez94 (approved):
LGTM. Just left one small suggestion

packages/protocol/contracts/layer1/core/impl/Codec.sol
ggonzalez94: should be resolved now that the other PR is merged

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: It seems to me that `ProveInput` contains two categories of data:

- **Commitment data** – this is what the prover is actually proving, and it must be fully determined before the proof is generated.  
- **Usage options** – e.g. whether to proactively write a checkpoint; these can be decided after the proof is generated.

The prover’s guest program only depends on the `Commitment`. Our prover system can choose or adjust the usage options later, using additional data, either during or after proof generation.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: done in https://github.com/taikoxyz/taiko-mono/pull/21005/changes

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik: improve the comments of ProverInput in https://github.com/taikoxyz/taiko-mono/pull/21005/changes

ggonzalez94 (approved):

PR #20987, opened 2025-12-11
feat(protocol): add whitelistProver field to Inbox config

dantaik:
Before our new Shasta guest code is audited, maybe we want to be the only prover?


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -590,6 +582,44 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
         }
     }
 
+    function _isWhitelistedProver() private view returns (bool) {
```
ggonzalez94, line 585: Nit: add natspec
```suggestion
    /// @dev Checks if the caller is a whitelisted prover.
    /// If `_whitelistProver` is different from address(0), it will revert if the caller is not the whitelisted prover.
    /// @return true if the caller is a whitelisted prover, false otherwise.
    function _isWhitelistedProver() private view returns (bool) {
```

ggonzalez94 (commented):
This actually makes sense to me. My only suggestion is that we might want to have a mapping of whitelisted provers instead of a single entity for redundancy and to allow a gradual rollout until we make it permisionless.
It will slightly increase the cost of the `prove` function by 2.1k gas, but considering current gas prices and the fact that verifying a snark is already around 300k gas the small extra cost is worth the flexibility

PR #20988, opened 2025-12-11
refactor(protocol): use block hash instead of checkpoint hash

dantaik:
@ggonzalez94 

If the change **[Remove on-chain checkpoint verification](https://github.com/taikoxyz/taiko-mono/pull/20986)** makes sense to you, then we should also use block hash instead of checkpoint hash. Block hash is more intuitive.


davidtaikocha (approved):
Changes lgtm.

ggonzalez94 (approved):
LGTM. Sorry going back to checkpointHash and now you had to update to blockHash again 😅 . I thought that having a single abstraction(i.e. checkpoint) made sense, but now agree this is better

PR #20992, opened 2025-12-11
fix(protocol): use getBeaconBlockRootAtOrAfter in `PreconfWhitelist` to account for L1 missed slots

### LINKED ISSUE #20991: L-05 Beacon Block Root Retrieval Lacks Fallback for Missed Slots
The PreconfWhitelist contract determines the active operator for a given epoch by [deriving randomness](https://github.com/taikoxyz/taiko-mono/blob/503445678a4bd875d761e56ba80a29a5b8e68d6e/packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol#L280) from the beacon block root associated with the epoch's start timestamp. This randomness is retrieved via the _getRandomNumber function, which currently queries the beacon roots contract for the exact timestamp of the epoch boundary.

The use of [LibPreconfUtils.getBeaconBlockRootAt](https://github.com/taikoxyz/taiko-mono/blob/503445678a4bd875d761e56ba80a29a5b8e68d6e/packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol#L60) strictly requires a block to exist at the specific calculation timestamp. If the slot at the epoch boundary is missed, the function returns bytes32(0), resulting in a randomness value of zero. This causes the operator selection logic to deterministically default to the operator at index 0, which introduces a selection bias.

Consider replacing the call to getBeaconBlockRootAt with [getBeaconBlockRootAtOrAfter](https://github.com/taikoxyz/taiko-mono/blob/503445678a4bd875d761e56ba80a29a5b8e68d6e/packages/protocol/contracts/layer1/preconf/libs/LibPreconfUtils.sol#L39) within the _getRandomNumber function. This change ensures that if the boundary slot is missed, the system correctly locates the next available valid beacon block root, thereby preserving the intended randomness distribution and removing the dependency on a single operator during network anomalies.

ggonzalez94:
Fixes #20991 

---

> [!NOTE]
> Switches preconf operator selection to use `getBeaconBlockRootAtOrAfter`, adds a beacon-root mock and a test for skipped slots, and refreshes gas/snapshot metrics.
> 
> - **Preconf**:
>   - Update `PreconfWhitelist._getRandomNumber` to use `LibPreconfUtils.getBeaconBlockRootAtOrAfter` for epoch randomness.
> - **Tests**:
>   - Add `MockBeaconBlockRoot` with configurable roots and deterministic fallback.
>   - Inject beacon-root mock in `InboxTestBase` via `_mockBeaconBlockRoot`.
>   - Add `TestPreconfWhitelist.test_getOperatorForEpoch_handlesSkippedRandomnessSlot` to validate selection when the primary randomness slot is missed.
> - **Metrics**:
>   - Refresh gas report entries in `gas-reports/layer1-contracts.txt`.
>   - Update snapshot values in `snapshots/shasta-propose.json`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cd72387148c980aa9cf212ba6d952f76f4c96fb3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/snapshots/shasta-propose.json
```diff
@@ -1,5 +1,5 @@
 {
-  "propose_after_ring_buffer_wrap": "39746",
-  "propose_forced_inclusion": "50973",
-  "propose_single": "62846"
-}
\ No newline at end of file
+  "propose_after_ring_buffer_wrap": "40787",
+  "propose_forced_inclusion": "54014",
+  "propose_single": "65887"
```
ggonzalez94: I believe the gas increase looks higher than it actually will be because of the fact that the mock accesses storage. The real BEACON_ROOT contract will not

davidtaikocha (approved):

PR #20959, opened 2025-12-09
feat(protocol): make taiko cheaper

ggonzalez94:
Current base fee = 0.025 gwei(https://taikoscan.io/gastracker)

Reduce minimum base fee to `0.0025 gwei` now that L1 is much cheaper and remove hardcoded `BASEFEE_MIN_VALUE` since the minimum calculated by the EIP-1559 curve is now smaller.

Script for the DAO proposal coming soon.

@davidtaikocha @dantaik please review the changes carefully since this is a part of the codebase I'm not familiar with

---

> [!NOTE]
> Lowers the base fee floor to 0.0025 gwei by reducing `BASEFEE_MIN_VALUE` and updating `minGasExcess` in L1 configs, with tests adjusted accordingly.
> 
> - **Protocol**:
>   - **L2**:
>     - Reduce `BASEFEE_MIN_VALUE` in `packages/protocol/contracts/layer2/based/TaikoAnchor.sol` from `25_000_000` (0.025 gwei) to `2_500_000` (0.0025 gwei).
>   - **L1 Configs**:
>     - Update `baseFeeConfig.minGasExcess` to `1_289_447_652` (was `1_344_899_430`) in:
>       - `packages/protocol/contracts/layer1/mainnet/MainnetInbox.sol`
>       - `packages/protocol/contracts/layer1/hekla/HeklaInbox.sol`
>       - `packages/protocol/contracts/layer1/devnet/DevnetInbox.sol`
> - **Tests**:
>   - Adjust `test_mainnet_min_basefee` in `packages/protocol/test/layer2/LibEIP1559.t.sol` to use new `minGasExcess` value.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f9779eae0c7884ee0e7d48434f16d1b14cd5f185. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #21004, opened 2025-12-11
feat(protocol): proposal 7 to update the minimum base fee

ggonzalez94:
DAO proposal for PR #21048 (original was #20959)

- [CodeDiff](https://codediff.taiko.xyz/?addr=0x1670000000000000000000000000000000010001&newimpl=0xf381868dd6b2ac8cca468d63b42f9040de2257e9&chainid=167000)

EDIT: After discussions we decided to only do an upgrade on the L2 anchor to avoid further changes on driver and sidecars, and potential re-orgs. This still allows a 2.5x reduction on the floor base fee.


packages/protocol/script/layer1/proposals/Proposal0007.s.sol
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    function buildL2Actions()
+        internal
+        pure
+        override
+        returns (uint64 l2ExecutionId, uint32 l2GasLimit, Controller.Action[] memory actions)
+    {
+        l2ExecutionId = 7;
+        l2GasLimit = 500_000;
```
ggonzalez94: are these 2 correct?

packages/protocol/script/layer1/proposals/Proposal0007.s.sol
```diff
@@ -0,0 +1,38 @@
...
+// To print the proposal action data: `P=0007 pnpm proposal`
+// To dryrun the proposal on L1: `P=0007 pnpm proposal:dryrun:l1`
+// To dryrun the proposal on L2: `P=0007 pnpm proposal:dryrun:l2`
+contract Proposal0007 is BuildProposal {
+    // L1 TaikoInbox (proxy) and new implementation with 0.0025 gwei curve floor
+    address public constant TAIKO_INBOX = 0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a;
+    address public constant TAIKO_INBOX_NEW_IMPL = 0x0000000000000000000000000000000000000000; // TODO: set deployed impl
+
+    // L2 TaikoAnchor (proxy) and new implementation without BASEFEE_MIN_VALUE hard floor
+    address public constant TAIKO_L2 = 0x1670000000000000000000000000000000010001;
```
dantaik: maybe we can use addresses defined in https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer2/mainnet/LibL2Addrs.sol?

packages/protocol/script/layer1/proposals/Proposal0007.s.sol
dantaik: These two actions will not be executed automatically, L1 actions will execute first, then L2 later through bridge. I wonder if this is OK for this upgrade.

dantaik:
@davidtaikocha this proposal has both L1 and L2 actions. Do you think it will work if L1 actions get executed first to upgrade the gas limit, then followed by L2 actions (hours later) to change the base fee???

davidtaikocha:
> @davidtaikocha this proposal has both L1 and L2 actions. Do you think it will work if L1 actions get executed first to upgrade the gas limit, then followed by L2 actions (hours later) to change the base fee???

It'll result in the node will use the new `MinGasExcess` to call the old `getBasefeeV2` for a few hrs, as long as it's acceptable, or our contract tests have covered this already, then it's fine.

ggonzalez94:
> this proposal has both L1 and L2 actions. Do you think it will work if L1 actions get executed first to upgrade the gas limit, then followed by L2 actions (hours later) to change the base fee???

I believe they could be independently executed in any order right? 
- If L1 gets executed first, the calculated fee for the EIP-1559 curve will be lower than the `BASEFEE_MIN_VALUE` enforced by the anchor. So it will still be capped at a higher value until L2 executes.
- If L2 gets executed first, then `BASEFEE_MIN_VALUE` will be lower, but the curve will still output a higher value, so the higher value will be used.

Am I missing something?

ggonzalez94:
@dantaik @davidtaikocha the contracts have been deployed both on L1 and L2, and the proposal has been updated.
- L1 [code diff](https://codediff.taiko.xyz/?addr=0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a&newimpl=0x7b4dBB1bCF546e181958B8c053806a95332553dD&chainid=1)
- L2 [code diff](//https://codediff.taiko.xyz/?addr=0x1670000000000000000000000000000000010001&newimpl=0xeaB88687835b87f9a9DF22fA42E16153fE7b25BB&chainid=167000)

dantaik:
> > this proposal has both L1 and L2 actions. Do you think it will work if L1 actions get executed first to upgrade the gas limit, then followed by L2 actions (hours later) to change the base fee???
> 
> I believe they could be independently executed in any order right?
> 
> * If L1 gets executed first, the calculated fee for the EIP-1559 curve will be lower than the `BASEFEE_MIN_VALUE` enforced by the anchor. So it will still be capped at a higher value until L2 executes.
> * If L2 gets executed first, then `BASEFEE_MIN_VALUE` will be lower, but the curve will still output a higher value, so the higher value will be used.
> 
> Am I missing something?

The L1 actions must be executed first followed by a signal sent to L2 so L2 actions can be executed later.

**@davidtaikocha , can you double-confirm this will work? **

dantaik (approved):

packages/protocol/script/layer1/proposals/Proposal0007.s.sol
```diff
@@ -0,0 +1,32 @@
...
+        actions[0] = buildUpgradeAction(L1.INBOX, INBOX_NEW_IMPL);
+    }
+
+    function buildL2Actions()
+        internal
+        pure
+        override
+        returns (uint64 l2ExecutionId, uint32 l2GasLimit, Controller.Action[] memory actions)
+    {
+        l2ExecutionId = 0;
```
dantaik, line 23: @ggonzalez94 since the L2 actions don't have a precondition of other L2 actions, I changed the l2ExecutionId to 0.

packages/protocol/script/layer1/proposals/Proposal0007.s.sol
```diff
@@ -0,0 +1,32 @@
...
+import "../governance/BuildProposal.sol";
+
+// To print the proposal action data: `P=0007 pnpm proposal`
+// To dryrun the proposal on L1: `P=0007 pnpm proposal:dryrun:l1`
+// To dryrun the proposal on L2: `P=0007 pnpm proposal:dryrun:l2`
+contract Proposal0007 is BuildProposal {
+    // https://codediff.taiko.xyz/?addr=0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a&newimpl=0x7b4dBB1bCF546e181958B8c053806a95332553dD&chainid=1
+    address public constant INBOX_NEW_IMPL = 0x7b4dBB1bCF546e181958B8c053806a95332553dD;
+
+    // https://codediff.taiko.xyz/?addr=0x1670000000000000000000000000000000010001&newimpl=0x7b4dBB1bCF546e181958B8c053806a95332553dD&chainid=167000
```
dantaik: @ggonzalez94 I notice the two impl addresses are the same. I might be that you use deployed these two contracts using the same nonce. I think it will be good to have a different address.

dantaik: I also noticed that the library node_modules/solady/src/utils/FixedPointMathLib.sol we use has a different version. For SC to approve this proposal quickly. Do you think it make sense to copy the old library code to the local file to re-compile and re-deploy?

dantaik (changes requested):

dantaik:
Readiness checklist:

- [x] @davidtaikocha verifies the order of contract upgrade works
- [x] @smtmfft verifies prover will support this change without changes
- [x] @ggonzalez94 use different addresses for the 2 impl
- [x] @ggonzalez94 use old library to re-compile Anchor.sol?

packages/protocol/script/layer1/proposals/Proposal0007.s.sol
ggonzalez94: > I notice the two impl addresses are the same. I might be that you use deployed these two contracts using the same nonce. I think it will be good to have a different address.

Yes, I used a fresh deployed and both got deployed with the same nonce. I deployed a new [Anchor](https://taikoscan.io/address/0xeaB88687835b87f9a9DF22fA42E16153fE7b25BB) with the same solady version as the currently deployed. So both issues should be fixed now

ggonzalez94:
@codex please review

ggonzalez94:
Updated the proposal after discussions to follow PR #21048 where we only upgrade the L2 side

dantaik (approved):

davidtaikocha (approved):

PR #21006, opened 2025-12-12
feat(protocol): introduce IProverWhitelist for flexible prover management

dantaik:
## Summary
- Introduce IProverWhitelist to allow only trusted provers before our Prover Guest program is thoroughly audited again.
- I sorted function in Inbox.sol (transactional -> view -> pure)


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -263,26 +272,15 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
...
-                    commitment.actualProver,
-                    _provingWindow,
-                    _extendedProvingWindow
-                );
-            if (bondInstruction.bondType != LibBonds.BondType.NONE) {
-                _signalService.sendSignal(LibBonds.hashBondInstruction(bondInstruction));
-                emit BondInstructionCreated(bondInstruction.proposalId, bondInstruction);
+            // `proposalAge` is the age of the next-to-finalize proposal.
+            uint256 proposalAge = block.timestamp
+                - commitment.transitions[offset].timestamp.max(state.lastFinalizedTimestamp);
```
dantaik: @ggonzalez94 maybe we should simply use: `block.timestamp - commitment.transitions[offset].timestamp`?

Anyway, this may related to the PR you will rebase.

PR #21009, opened 2025-12-12
feat(protocol):  recreated the @gustavo's same PR with upstream changes merged

dantaik:
Original: https://github.com/taikoxyz/taiko-mono/pull/20926

## Additional changes
- merge IBondProcessor into IBondManager

---

> [!NOTE]
> Replaces `extendedProvingWindow` with `maxProofSubmissionDelay` and unifies bond logic to LIVENESS-only with updated L2 BondManager processing, removing provability paths and updating configs/genesis/tests.
> 
> - **Layer 1 (Inbox)**:
>   - Replace `extendedProvingWindow` with `maxProofSubmissionDelay` in `IInbox.Config` and `Inbox`.
>   - Update proof age calculation and proof verification; compute `proposalAge` relative to `lastFinalizedTimestamp` and transition timestamp.
>   - Revise bond signaling: emit LIVENESS bond when late (beyond `provingWindow` + sequential grace), including when payer==payee; remove extended-window/provability logic and `LibBondInstruction`.
>   - Update `getConfig`; adjust devnet/mainnet constructors to set `maxProofSubmissionDelay`.
>   - Clean up validation errors in `LibInboxSetup`.
> - **Layer 2 (Bonds/Anchor)**:
>   - Remove `provabilityBond` from `Anchor` (constructor/immutables/docs) and related genesis wiring.
>   - `BondManager` now processes bond instructions (`processBondInstruction`): slash 50%; when payer!=payee credit 50% to payee, when payer==payee refund 40% and reward 10% to caller; unify deposit events to `BondDeposited(depositor, recipient, amount)`; rearrange view/txn fns and errors.
>   - `IBondManager` interface updated to match deposits/withdrawals and views; `IBondProcessor` event/method renamed to instruction-based.
>   - `LibBonds` enum drops `PROVABILITY` (LIVENESS-only).
> - **Genesis/Tooling/Tests**:
>   - Remove `provabilityBond` from genesis config, codegen, and storage layouts; link updates for immutables.
>   - Update gas snapshots and tests across L1/L2 to new timing and bond semantics (late liveness, payer==payee case, whitelisted prover skip).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ab06824f9a680e8412f55e40c1ace41157f2cb95. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -312,6 +308,7 @@ contract Inbox is IInbox, IForcedInclusionStore, EssentialContract {
             // ---------------------------------------------------------
             // 6. Verify the proof
             // ---------------------------------------------------------
+            uint256 proposalAge = block.timestamp - commitment.transitions[offset].timestamp;
```
ggonzalez94: don't you think we should send to the verifier the time when the proposal became provable? not when it was created. Because that is the actual moment when we should start counting if the proposal is "too old".

Imagine this scenario:
- proposalA is posted at time t0
- but it only becomes provable after 2 days(because provers took too long, even if they got slashed, etc)
- proposalAge should already be 2 days? I don't think so, since if you want to take the action on the verifier to allow this proposal to be proven with a weaker guarantee because it is too old, you want to do it from the moment it becomes provable, not when it gets posted.

I know in practice the moment when a proposal gets posted and when it becomes provable are not going to have too much of a difference(only a few hours at max), but I still think it is more correct to use the moment when the proposal becomes provable


```suggestion
            uint256 proposalAge = block.timestamp - transitionAtOffset.timestamp.max(state.lastFinalizedTimestamp);
```

packages/protocol/contracts/layer2/core/IBondProcessor.sol
```diff
@@ -1,31 +0,0 @@
-// SPDX-License-Identifier: MIT
-pragma solidity ^0.8.24;
-
-import { LibBonds } from "src/shared/libs/LibBonds.sol";
-
-/// @title IBondProcessor
-/// @notice Interface for processing L1 bond signals on L2
-/// @custom:security-contact security@taiko.xyz
-interface IBondProcessor {
```
ggonzalez94, line 9: a bit of a nit, but:
while in our case this is implemented in the same `BondManager` contract, I think conceptually bond management(e.g. deposits, withdrawals, etc.) and bond processing are different. So it might be worth to keep them separated in two interfaces, especially for consumers(e.g. the relayer only needs to care about IBondProcessor methods, not the management)

packages/protocol/contracts/layer2/core/IBondManager.sol
```diff
@@ -73,30 +75,47 @@ interface IBondManager {
...
-
-    /// @notice Deposit ERC20 bond tokens for another address.
+    /// @notice Deposit ERC20 bond tokens for an address.
     /// @dev Does not cancel the recipient's pending withdrawal; the recipient must call
     /// `cancelWithdrawal` to reactivate their bond status.
-    /// @param _recipient The address to credit the bond to.
+    /// @param _recipient The address to credit the bond to. If zero, credits msg.sender.
     /// @param _amount The amount to deposit.
-    function depositTo(address _recipient, uint256 _amount) external;
+    function deposit(address _recipient, uint256 _amount) external;
```
ggonzalez94: while functionally correct, in my experience this kind of interfaces tend to give more problems that just having two separate functions that clearly reflect the intent(it is easier to use and less error prone).
Are you open to just keeping them as `deposit` and `depositTo`? I don't think it adds much code repetition since the core of the logic is still implemented in `_creditBond`

packages/protocol/contracts/layer2/core/IBondManager.sol
```diff
@@ -30,15 +32,10 @@ interface IBondManager {
...
-    /// @param account The account that deposited the bond
-    /// @param amount The amount deposited
-    event BondDeposited(address indexed account, uint256 amount);
-
-    /// @notice Emitted when a bond is deposited for another address
     /// @param depositor The account that made the deposit
     /// @param recipient The account that received the bond credit
     /// @param amount The amount deposited
-    event BondDepositedFor(address indexed depositor, address indexed recipient, uint256 amount);
+    event BondDeposited(address indexed depositor, address indexed recipient, uint256 amount);
```
ggonzalez94, line 36: I do agree it is a good idea to merge the events 🔝

ggonzalez94 (commented):
Thanks for porting the changes and merging upstream. I left a few suggestions, if you agree with them I can also create a PR on top of this one to address them

ggonzalez94 (approved):

PR #21015, opened 2025-12-13
feat(protocol): implement EIP-7201 namespaced storage in SignalService

dantaik:
## Security Improvement
EIP-7201 namespaced storage prevents collisions with contract state variables through high-entropy domain separation, reducing collision probability to ~1/2^256 compared to legacy keccak256(abi.encodePacked(...)) slots.

Replacing an old PR: https://github.com/taikoxyz/taiko-mono/pull/20940


davidtaikocha (approved):

dantaik:
@claude please review

ggonzalez94 (changes requested):
I'm still thinking about it. But my first impression(summarized by claude is) that maybe the risk is negligible enough that it is not worth the complexity + extra cost? (at least during this upgrade)

 **Why the legacy approach (keccak256(abi.encodePacked("SIGNAL", chainId, app, signal))) is probably safe:**

  1. It starts with "SIGNAL" - this already provides domain separation from other contract storage
  2. SignalService has minimal storage - only _receivedSignals mapping at slot 253
  3. The formula uses encodePacked which produces different results than Solidity's internal slot calculation (which uses encode)

  My assessment:

  The risk is more theoretical than practical for SignalService specifically. The PR description claims:
  "reducing collision probability to ~1/2^256 compared to legacy keccak256(abi.encodePacked(...)) slots"

  This is misleading - the legacy approach ALSO has ~1/2^256 collision probability. The real benefit is standardization and tooling compatibility, not meaningfully improved collision resistance.

**IF** we were to still decide to implement it I think we should use a standard library(i.e. OZ) to calculate the slots instead of doing our own implementation. Please check the [minimal rollup implementation](https://github.com/OpenZeppelin/minimal-rollup/blob/main/src/libs/LibSignal.sol) for an example of how it could be implemented.

Once we discuss if it is worth the complexity(maybe we get the auditors involved?) I can migrate to using the standard library implementation for slot calculation

dantaik:
Let me think about it too.

PR #21016, opened 2025-12-13
fix(taiko-client): remove redundant IsShasta guard in HandleShasta

lisenokdonbassenok:
Removed an unreachable !meta.IsShasta() guard from HandleShasta, since the public Handle method already routes only Shasta metadata to this helper.


davidtaikocha:
It’s ok to keep it here for a safety.

PR #21021, opened 2025-12-14
fix(bridge-ui): swap gas limit values for ERC1155 bridge

Bilogweb3:
The gas limit values for deployed vs not-deployed tokens were swapped in ERC1155Bridge. When bridging a token that hasn't been deployed on the destination chain yet, we need more gas (2.6M) to cover the deployment cost. The code was using the lower deployed limit (1.1M) instead, which could cause transactions to fail.

Fixed to match the logic in ERC20Bridge, ERC721Bridge and recommendProcessingFee.ts.


dantaik (approved):

packages/bridge-ui/src/libs/bridge/ERC1155Bridge.ts
```diff
@@ -216,8 +216,8 @@ export class ERC1155Bridge extends Bridge {
       gasLimit = 0;
     } else {
       gasLimit = !isTokenAlreadyDeployed
```
dantaik, line 218: maybe removing "!" is a better change

Bilogweb3: > maybe removing "!" is a better change

the "!" is intentional - higher gas limit when token isn't deployed yet (false), lower when it is (true)

KorbinianK (approved):

PR #21029, opened 2025-12-15
fix(protocol): redundant code and unused components

### LINKED ISSUE #21028: N-05 Redundant Code and Unused Components Identification
Several instances of redundant code and unused components were identified across the codebase, indicating opportunities for simplification and optimization.

- **Redundant `totalFees > 0` Check**:  
  The `_dequeueAndProcessForcedInclusions` function includes an explicit check (`if (totalFees > 0)`) before calling `_feeRecipient.sendEtherAndVerify`. This check is redundant because the `sendEtherAndVerify` function itself handles a zero amount by returning without performing a transfer. Consider removing the `if (totalFees > 0)` check to streamline the control flow.

- **Unused `BLOB_BYTES` Constant**:  
  The `BLOB_BYTES` constant is defined in `LibBlobs.sol` as `BLOB_FIELD_ELEMENTS * FIELD_ELEMENT_BYTES` but is not referenced anywhere within the contract. Unused code contributes to cognitive overhead and can create confusion regarding its purpose. Consider removing the `BLOB_BYTES` constant to clean up the codebase.

- **Redundant Encoding/Decoding Logic**:  
  Across the core libraries, there is duplicated code for encoding and decoding specific structs. For instance, there are two `_encodeProposal` functions [1, 2], while other encode/decode functions handle structs inline. This duplication leads to code bloat and inconsistencies. Consider consolidating common encoding and decoding functions into a shared, dedicated library (e.g., `LibCodec` or `LibTypes`) that can be imported and utilized uniformly across the protocol. This would centralize logic and improve consistency.

- **Redundant `BlockParams` Struct**:  
  The `BlockParams` struct in `Anchor.sol` (`anchorBlockNumber`, `anchorBlockHash`, `anchorStateRoot`) duplicates the fields and purpose of the `Checkpoint` struct defined in `ICheckpointStore.sol`. This redundancy introduces unnecessary complexity and an additional struct definition for essentially the same data, particularly since `Anchor.sol` already interacts with `ICheckpointStore`. Consider replacing the `BlockParams` struct with `ICheckpointStore.Checkpoint` directly wherever `BlockParams` is used, reducing redundancy and improving type consistency.

- **Unused `LibSignals.sol` Library**:  
  The `LibSignals.sol` library, which defines constants like `SIGNAL_ROOT` and `STATE_ROOT`, appears to be entirely unused throughout the codebase. Similar to other unused code, its presence adds unnecessary overhead. Consider removing `LibSignals.sol` if it is not intended for future use, or integrate its functionality if it serves a specific, currently unfulfilled purpose.

Consider addressing the above instances to enhance code quality, reduce maintenance effort, and improve overall system clarity.

ggonzalez94:
Fixes #21028

---

> [!NOTE]
> Extracts a shared transition codec, updates Anchor.anchorV4 to accept ICheckpointStore.Checkpoint, cleans up blob utilities/docs, removes an unused lib, tweaks forced-inclusion fee transfer, and updates tests/snapshots.
> 
> - **L2 Anchor**:
>   - Change `anchorV4` signature to accept `ICheckpointStore.Checkpoint` instead of `BlockParams`; update `_validateBlock` and callers/tests accordingly.
> - **L1 Codecs**:
>   - Introduce `LibTransitionCodec` (`encodeTransition`/`decodeTransition`, `TRANSITION_SIZE`).
>   - Refactor `LibProveInputCodec` to use `LibTransitionCodec` and its `TRANSITION_SIZE`.
> - **Blobs**:
>   - Remove unused size constants from `LibBlobs`; keep/clarify `BlobReference` and `BlobSlice` usage.
> - **Inbox**:
>   - Always call `_feeRecipient.sendEtherAndVerify(totalFees * 1 gwei)` after processing forced inclusions.
> - **Docs**:
>   - Add explicit `BLOB_BYTES = 4096 * 32` note in derivation rules.
> - **Cleanup**:
>   - Remove unused `LibSignals` library.
> - **Tests & Snapshots**:
>   - Update `Anchor.t.sol` to pass `Checkpoint` and adjust assertions.
>   - Refresh `shasta-*` snapshot values.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5fb0312c0a566d5692ebc10583e5727c5256f7ad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/core/libs/LibBlobs.sol
```diff
@@ -5,17 +5,6 @@ pragma solidity ^0.8.26;
 /// @notice Library for handling blobs.
 /// @custom:security-contact security@taiko.xyz
 library LibBlobs {
-    // ---------------------------------------------------------------
-    // Constants
-    // ---------------------------------------------------------------
-    uint256 internal constant FIELD_ELEMENT_BYTES = 32;
```
ggonzalez94, line 11: @davidtaikocha is it ok to get rid of these constants or is the node reading directly from them? I updated the derivation doc to use them directly instead

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -43,13 +43,6 @@ contract Anchor is EssentialContract {
         bytes proverAuth; // Encoded ProverAuth for prover designation
     }
 
-    /// @notice Block-level data specific to a single block within a proposal.
-    struct BlockParams {
```
ggonzalez94, line 47: these values were indeed redundant to those on the `Checkpoint` struct, but will likely require an update in the preconf sidecars. @davidtaikocha please check if it is not inconvinient, otherwise we can just keep it

packages/protocol/contracts/layer1/core/libs/LibBlobs.sol
davidtaikocha: Yes its OK. Clients are only using the same constants defined in geth / reth.

packages/protocol/contracts/layer2/core/Anchor.sol
davidtaikocha: I think it's easy to change, if we want this simplification in `anchorv4`.

davidtaikocha (approved):

PR #21039, opened 2025-12-16
fix(protocol): align derivation spec with shasta implementation

MoNyAvA:
Align the Shasta derivation spec with the current implementation. The manifest struct examples now match LibManifest on L1, and the gas limit section documents the effective parent gas limit approach instead of the old +1,000,000 adjustment for the anchor transaction. The doc also clarifies that timestamp, anchor and gas limit validation is applied per derivation source and falls back to the default manifest on failure. The values of TIMESTAMP_MAX_OFFSET and MAX_BLOCK_GAS_LIMIT in the constants table are updated to reflect the actual protocol limits.

### DIFF
--- a/packages/protocol/docs/Derivation.md
+++ b/packages/protocol/docs/Derivation.md
@@ -129,8 +129,6 @@ The `derivation.sources` array contains `DerivationSource` objects, each with a 
 The manifest data structures are defined as follows:
 
 ```solidity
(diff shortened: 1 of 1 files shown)


davidtaikocha:
Thanks to the contribution and our latest derivation doc is under https://github.com/taikoxyz/taiko-mono/tree/taiko-alethia-protocol-v3.0.0 :)

PR #21043, opened 2025-12-17
fix(eventindexer): use correct event name for BlockVerifiedV2

reject-i:
Fixed wrong event name in saveBlockVerifiedEventV2 function. It was saving BlockVerifiedV2 events as "BatchesVerified" instead of "BlockVerified". This was a copy-paste mistake that caused V2 block verification events to be stored with incorrect type in the database.


davidtaikocha:
it's `EventNameBatchesVerified` for current Pacaya fork.

PR #21046, opened 2025-12-18
fix(ejector): simplify parent_missing logic in ChainReorgTracker

hawkadrian:
## What

Removed dead code in `ChainReorgTracker::apply()` that checked `last.hash != block.parent_hash` after exiting the loop via `break`. This check was always `false` since we only reach it when the parent hash matches.

## Why

The redundant `match` expression with an unreachable branch made the code harder to understand and maintain. Simplified logic now directly checks if history is empty and reorged blocks exist to determine if parent is missing.

### DIFF
--- a/packages/ejector/src/monitor_reorg.rs
+++ b/packages/ejector/src/monitor_reorg.rs
@@ -64,10 +64,9 @@ impl ChainReorgTracker {
             break;
         }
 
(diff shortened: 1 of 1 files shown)


davidtaikocha:
Can't see too many improvements here: https://github.com/taikoxyz/taiko-mono/blob/main/CONTRIBUTING.md

PR #21057, opened 2025-12-19
chore(protocol): update derivation pipeline doc for the latest protocol implementation

davidtaikocha:


davidtaikocha:
@claude plz review.

packages/protocol/docs/Derivation.md
```diff
@@ -28,11 +28,10 @@ Throughout this document, metadata references follow the notation `metadata.fiel
 | **isLowBondProposal**  | Indicates if the proposer has insufficient bond or is exiting |
 | **designatedProver**   | The prover responsible for proving the block                  |
 | **timestamp**          | The timestamp when the proposal was accepted on L1            |
-| **originBlockNumber**  | The L1 block number in which the proposal was accepted        |
+| **originBlockNumber**  | The L1 block number **just before** the proposal was accepted |
```
ggonzalez94: nit: better to be more specific about it being exactly one block before
```suggestion
| **originBlockNumber**  | The L1 block number from **one block before** the proposal was accepted |
```

packages/protocol/docs/Derivation.md
```diff
@@ -28,11 +28,10 @@ Throughout this document, metadata references follow the notation `metadata.fiel
 | **isLowBondProposal**  | Indicates if the proposer has insufficient bond or is exiting |
 | **designatedProver**   | The prover responsible for proving the block                  |
```
ggonzalez94: this was not accurate before
```suggestion
| **designatedProver**   | The prover responsible for proving the proposal                  |
```

packages/protocol/docs/Derivation.md
ggonzalez94: maybe we link to the actual event definition in `IInbox`? this way we avoid the definition from becoming stale

packages/protocol/docs/Derivation.md
```diff
@@ -72,7 +72,7 @@ event Proposed(
 
 The other fields can be derived by querying the contract:
```
ggonzalez94: maybe it is more accurate to say the come from querying the L1 and the contract? because timestamp and originblocHash and number are not directly retrieved from the contract itself
```suggestion
The other fields can be derived by querying the L1 and the inbox contract:
```

packages/protocol/docs/Derivation.md
ggonzalez94: we do not mention `endOfSubmissionWindowTimestamp` as one of the metadata fields for proposal in the initial section. Should we do that or remove it from here since we are actually not using it before we enable permisionless preconfs?

packages/protocol/docs/Derivation.md
```diff
@@ -95,7 +95,7 @@ The following metadata fields are extracted directly from the event payload:
 The `proposal.sources` array contains `DerivationSource` objects, each with a `blobSlice` field that serves as the primary mechanism for locating and validating proposal metadata. Responsibilities are split as follows:
 
 - **Forced inclusion submitters** publish their `DerivationSourceManifest` blob on L1 directly as blobs.
-- **The proposer** gathers every required derivation source—both their own blocks and any outstanding forced inclusions—and publishes a single `ProposalManifest` within their own `DerivationSource` (**appended last**). The inbox contract guarantees that forced inclusions are proposed as they were posted(without metadata being maniupalted).
+- **The proposer** gathers every required derivation source—both their own blocks and any outstanding forced inclusions—and publishes a single `ProposalManifest` within their own `DerivationSource` (**appended last**). The inbox contract guarantees that forced inclusions are proposed as they were posted (without metadata being manipulated).
```
ggonzalez94: is the proposer still posting the actual `ProposalManifest` or is they only post their DerivationSource and the actual ProposalManifest is constructed at derivation time? This is something I often get confused about, but from our previous discussion and asking AI in our codebase it seems the proposer does not actually submit ProposalManifest?

ggonzalez94: **IF** this is correct we may wanna write something like:

- **Forced inclusion submitters** publish blob data for a `DerivationSourceManifest` and then call `Inbox.saveForcedInclusion(blobReference)`; the inbox stores the resulting `blobSlice` in a queue.
- **The proposer** publishes blob data for their own `DerivationSourceManifest` and calls `Inbox.propose(...)` with a `blobReference` to it plus `numForcedInclusions`. The inbox dequeues that many forced inclusions and appends the proposer's source **last**.

So sources are processed in order: forced inclusions first (if any), then the proposer source last.

packages/protocol/docs/Derivation.md
```diff
@@ -95,7 +95,7 @@ The following metadata fields are extracted directly from the event payload:
...
 
 The manifest data structures are defined as follows:
```
ggonzalez94, line 90: Maybe we wanna reference that the canonical definitions are in `LibManifest`? Or alternatively we should delete that file if it is not useful, and just keep the definitions here.

packages/protocol/docs/Derivation.md
ggonzalez94, line 99: does the `ProposalManifest` still have a `proverAuthBytes` or it is just a leftover of our previous design?

packages/protocol/docs/Derivation.md
```diff
@@ -154,12 +154,7 @@ struct BlockManifest {
 
 ### Proposer and `isLowBondProposal` Validation
 
-To maintain the integrity of the proposal process, the `proposer` address must pass specific validation checks, which include:
-
-- Confirming that the proposer holds a sufficient balance in the L2 BondManager contract.
-- Ensuring the proposer is not waiting for exiting.
-
-Should any of these validation checks fail (as determined by the `anchorV4` function returning `isLowBondProposal = true`), the proposer's derivation source is replaced with the default manifest, which contains a single block with only an anchor transaction.
+The proposer must have sufficient L2 BondManager balance; if `anchorV4` flags `isLowBondProposal = true`, the proposer’s derivation source is replaced with the default manifest.
```
ggonzalez94: should we explain a bit more how this is done? i.e. mentioning the specific event?

Also maybe I forgot... but what happens with forced inclusions, they are not replaced with a default manifest right? if so we should call this out and explain how that is determined

packages/protocol/docs/Derivation.md
ggonzalez94: nit: link to the actual struct definition?

ggonzalez94, line 168: did we ended up implementing the block length validation in the driver? I don't remember what our final decision on this was.

If we ship zkgas this seems less necessary perhaps?

packages/protocol/docs/Derivation.md
ggonzalez94: After reading the code for a bit it seems that we would also invalidate the forced inclusions since they have the same proposer and this check would return true? If that's correct maybe it is more accurate to say that the whole proposal is invalidated?

ggonzalez94: on a separate, but related topic. I think on the rust implementation of the driver we are passing the proverAuthBytes from the proposer's source as the proposal level metadata(which is correct), but on the go implementation we pass empty bytes. This means we could mistakenly not use the right prover if there was a designation made. But maybe I'm wrong here as I'm not very knowledgable yet about the driver's code

packages/protocol/docs/Derivation.md
ggonzalez94: maybe this can be more clear?

packages/protocol/docs/Derivation.md
```diff
@@ -227,7 +222,7 @@ manifest.sources = [sourceManifest0, sourceManifest1, ...];  // With defaults fo
 
 #### Forced Inclusion Submission Requirements
 
-Users submit forced inclusion transactions directly to L1 by posting blob data containing a `DerivationSourceManifest` struct. To ensure valid forced inclusions that pass validation, the following `BlockManifest` fields must be set to zero, allowing the protocol to assign appropriate values:
+Users submit forced inclusion transactions directly to L1 by posting blob data containing a `DerivationSourceManifest` struct. To ensure valid forced inclusions that pass validation, the following `BlockManifest` fields are ignored and replaced by inherited values from the parent/proposer during metadata application:
```
ggonzalez94: nit: we say that the following values are replaced, but then we include the transactions in the list. Maybe something like this is more general?

```suggestion
Users submit forced inclusion transactions directly to L1 by posting blob data containing a `DerivationSourceManifest` struct. To ensure valid forced inclusions that pass validation, the following `BlockManifest`rules are applied:
```

packages/protocol/docs/Derivation.md
```diff
@@ -239,7 +234,7 @@ Users submit forced inclusion transactions directly to L1 by posting blob data c
 
 This design ensures forced inclusions integrate properly with the chain's metadata while allowing users to specify only their transactions without requiring knowledge of chain state parameters.
 
-If any of `gasLimit`, `coinbase`, `anchorBlockNumber`, or `timestamp` is non-zero, the decoder overwrites that field with zero so the derivation source remains valid rather than being downgraded to the default manifest.
+Any non-zero `gasLimit`, `coinbase`, `anchorBlockNumber`, or `timestamp` is overwritten during metadata application with inherited proposer/parent values, keeping the source valid and avoiding a fallback to the default manifest.
```
ggonzalez94, line 215: what are the proposer values? aren't all the ones listed above inherited from the parent block?

ggonzalez94 (commented):
Did a first pass and left a bunch of comments. A lot of them are around things we had not improved before and can be more clear. But I think the new main ones are:
- Clarify if we are using `ProposalManifest` and how. AFAIK the proposer is not posting it, so we should at least change that. But I could not find it in the derivation code either, so if that's the case should we remove it?
- Low bond proposals and forced inclusion sources: It is not clear how we deal with them(check the comments on that section)

Also should we update `LibManifest` as well? and if not used and we are ok with just having these structs defined in the Derivation doc directly maybe we just remove it?

ggonzalez94:
@codex please review carefully the derivation doc and compare it with the driver implementation. The driver is the source of truth.
- what things are not accurate?
- what things are not clear?
- what things should be improved?

Think deeply about this

packages/protocol/docs/Derivation.md
davidtaikocha: Updated in: https://github.com/taikoxyz/taiko-mono/pull/21057/commits/3da8b7cd728bde2e4d4a92e09b8c0de092e4927c

packages/protocol/docs/Derivation.md
davidtaikocha: Updated in https://github.com/taikoxyz/taiko-mono/pull/21057/commits/3da8b7cd728bde2e4d4a92e09b8c0de092e4927c

davidtaikocha: yeah it makes sense to remove it for now, removed in: https://github.com/taikoxyz/taiko-mono/pull/21057/commits/9f4240b16ca56d3ae8212055899f4653b8a1e77a

packages/protocol/docs/Derivation.md
davidtaikocha: Proposer is also posting `DerivationSource` as well, like FI submitter. Updated in: https://github.com/taikoxyz/taiko-mono/pull/21057/commits/1b0b68a495e1e4462b6cb2544bc796ef23f29e26

packages/protocol/docs/Derivation.md
davidtaikocha: Was thinking the same, deleted in: https://github.com/taikoxyz/taiko-mono/pull/21057/commits/9f4240b16ca56d3ae8212055899f4653b8a1e77a

packages/protocol/docs/Derivation.md
davidtaikocha: I think yes, the `proverAuthBytes` is used in `#### Prover Designation Process` later?

davidtaikocha: Changed.

davidtaikocha: We do have this implementation in driver: https://github.com/taikoxyz/taiko-mono/blob/main/packages/taiko-client/driver/chain_syncer/event/manifest/shasta.go#L135-L142 / https://github.com/taikoxyz/taiko-mono/blob/main/packages/taiko-client-rs/crates/driver/src/derivation/pipeline/shasta/pipeline/mod.rs#L67-L74. And I agree if we have ZK Gas for each block it will be less necessary.

davidtaikocha: Added the link to the rule: https://github.com/taikoxyz/taiko-mono/pull/21057/commits/b6e5a7fc50604bc28c741bb6aab9ab971c4d3307

packages/protocol/docs/Derivation.md
davidtaikocha: `coinbase` is not inherited from the parent block, and we are using the `proposal.proposer`:

```md
| `coinbase`          | Protocol substitutes `proposal.proposer`                        |
```

packages/protocol/docs/Derivation.md
davidtaikocha: > but what happens with forced inclusions, they are not replaced with a default manifest right? if so we should call this out and explain how that is determined

The current driver [won't replace it with a default manifest](https://github.com/taikoxyz/taiko-mono/blob/main/packages/taiko-client/driver/chain_syncer/event/syncer.go#L734-L739), and looks like the section `Immediate Mitigations` later should have explained this?
```md
##### Immediate Mitigations

- **Default Manifest Replacement**: When `isLowBondProposal = true`, **only the proposer's manifest** is replaced with the default manifest, minimizing proving costs and disincentivizing spam. Forced inclusion manifests are still processed.
```

davidtaikocha: > on a separate, but related topic. I think on the rust implementation of the driver we are passing the proverAuthBytes from the proposer's source as the proposal level metadata(which is correct), but on the go implementation we pass empty bytes. This means we could mistakenly not use the right prover if there was a designation made. But maybe I'm wrong here as I'm not very knowledgable yet about the driver's code

I think Go driver also sets the `proverAuthBytes` from the proposer's source to every source, but it's not in `ShastaDerivationSourceFetcher`, but [here](https://github.com/taikoxyz/taiko-mono/blob/main/packages/taiko-client/driver/chain_syncer/event/syncer.go#L321-L322), the implementation details are slightly different between Go and Rust, but should all align with the derivation process.

davidtaikocha:
> Clarify if we are using ProposalManifest and how. AFAIK the proposer is not posting it, so we should at least change that. But I could not find it in the derivation code either, so if that's the case should we remove it?

It's just a terminology used inside the derivation process now, our code is using some different names. But imo it's OK inside the doc for explanation?

packages/protocol/docs/Derivation.md
ggonzalez94: I think it is ok to keep it for now until we implement zk gas

packages/protocol/docs/Derivation.md
```diff
@@ -94,8 +83,8 @@ The following metadata fields are extracted directly from the event payload:
 
 The `proposal.sources` array contains `DerivationSource` objects, each with a `blobSlice` field that serves as the primary mechanism for locating and validating proposal metadata. Responsibilities are split as follows:
```
ggonzalez94:
```suggestion
The `sources` array in the `Proposed` event contains `DerivationSource` objects(see `IInbox.DerivationSource`[https://github.com/taikoxyz/taiko-mono/blob/update-derivation/packages/protocol/contracts/layer1/core/iface/IInbox.sol#L48]), each with a `blobSlice` field that serves as the primary mechanism for locating and validating proposal metadata. Responsibilities are split as follows:
```

packages/protocol/docs/Derivation.md
ggonzalez94, line 94:
```suggestion
/// @dev The ProposalManifest aggregates all DerivationSources' blob data for a proposal.
/// The ProposalManifest is conceptual, and used at derivation time only(i.e. it is not posted in blobs)
```

ggonzalez94, line 97:
```suggestion
  /// @notice Prover authentication data (proposal-level).
  /// It is retrieved from the proposer's derivation source manifest(the last manifest in the `sources` array)
```

ggonzalez94:
```suggestion
 /// @notice The anchor block number. If set to zero it will use the parent's anchor.
```

packages/protocol/docs/Derivation.md
```diff
@@ -154,29 +143,11 @@ struct BlockManifest {
 
 ### Proposer and `isLowBondProposal` Validation
 
-To maintain the integrity of the proposal process, the `proposer` address must pass specific validation checks, which include:
-
-- Confirming that the proposer holds a sufficient balance in the L2 BondManager contract.
-- Ensuring the proposer is not waiting for exiting.
-
-Should any of these validation checks fail (as determined by the `anchorV4` function returning `isLowBondProposal = true`), the proposer's derivation source is replaced with the default manifest, which contains a single block with only an anchor transaction.
+The proposer must have sufficient L2 BondManager balance; if `anchorV4` flags `isLowBondProposal = true`, the proposer’s derivation source is replaced with the default manifest.
```
ggonzalez94:
```suggestion
The proposer must have sufficient balance in the L2 [BondManager](https://github.com/taikoxyz/taiko-mono/blob/update-derivation/packages/protocol/contracts/layer2/core/BondManager.sol) contract. If the `anchorV4` transaction flags `isLowBondProposal = true`, the proposer’s derivation source is replaced with the default manifest.
```

packages/protocol/docs/Derivation.md
```diff
@@ -293,16 +264,9 @@ After all calculations above, an additional `1_000_000` gas units will be added
...
+Late-proof handling on L1 may emit at most one bond instruction for the first proven proposal. When triggered, the Inbox emits `BondInstructionCreated` and forwards the hash via `SignalService.sendSignal(keccak256(abi.encode(bondInstruction)))`.
 
-Bond instructions are emitted in the `Proposed` event, requiring clients to index these events. This indexing allows for the off-chain aggregation of bond instructions, which are then provided to the anchor transaction as input. Transitions that are proved but not used for finalization will be excluded from the anchor process and should be removed from the index.
-
-A parent proposal L1 transaction may revert, potentially causing the subsequent proposal's anchor transaction to revert due to differing bond instructions. To reduce such reverts, the anchor transaction processes bond instructions from an ancestor proposal that is `BOND_PROCESSING_DELAY` proposals prior to the current one. If `BOND_PROCESSING_DELAY` is set to 1, it effectively processes the parent proposal's instructions.
+Bond instruction settlement is **entirely on L2**: `BondManager.processBondInstruction` will consume the signal, which can be called by everyone.
```
ggonzalez94:
```suggestion
Bond instruction settlement is **entirely on L2** and not part of the derivation process. `[BondManager.processBondInstruction](https://github.com/taikoxyz/taiko-mono/blob/update-derivation/packages/protocol/contracts/layer2/core/BondManager.sol#L180)` consumes the signals emitted on the L1, and can be called by anyone. It has its own incentive system to incentivize slashing faulty provers.
```

packages/protocol/docs/Derivation.md
```diff
@@ -293,16 +264,9 @@ After all calculations above, an additional `1_000_000` gas units will be added
...
 
 ### Designated Prover System
```
ggonzalez94, line 273: maybe the designated prover section can be moved to its own H2 section since it is not directly related with Metadata Preparation?

packages/protocol/docs/Derivation.md
ggonzalez94: I don't remember why do we require that the same `proverAuthBytes` must be supplied for all subsequent blocks if it will be ignored?

packages/protocol/docs/Derivation.md
```diff
@@ -327,52 +291,73 @@ struct ProverAuth {
...
 - **Proposer Bond Check**: Verifies if the proposer maintains sufficient L2 bonds to cover the proving fee
-- **Low-Bond Detection**: Sets `metadata.isLowBondProposal = true` when:
-  - The proposer's bond balance falls below the protocol threshold
-  - The proposer is in the process of exiting the system
+- **Low-Bond Detection**: Sets `metadata.isLowBondProposal = true` when the proposer's bond balance falls below the protocol threshold.
 - **Designated Prover Bond Check**: If a different prover is designated, verifies they have sufficient bonds
 
 ##### 3. Prover Assignment Logic
 
 The final prover assignment follows a hierarchical fallback mechanism:
```
ggonzalez94: maybe it would be smarter to reference the code in the anchor instead of copying it here? I imagine this is something that otherwise can get out of sync pretty quickly

packages/protocol/docs/Derivation.md
ggonzalez94: Now that I read this again I think we should remove these sections, as we are not actively considering other alternatives to just designating the last prover(at least at the moment).
And it should actually happen fairly irregularly and the cost of proving the default manifest should be small enough that this is ok

ggonzalez94 (commented):
Left a few more comments. I'll create a PR on top of this to address them

ggonzalez94:
@davidtaikocha I created #21061 on top of this, please take a look. 

> It's just a terminology used inside the derivation process now, our code is using some different names. But imo it's OK inside the doc for explanation?

I added some comments that make it clear that `ProposalManifest` is just conceptual, and only used during derivation(the main confusion I wanted to avoid is people reading this expecting it should be posted on blobs). So I think it is fine for now, but if we are not using this abstraction on the code, maybe we shouldn't use it on the docs either(and use the actual abstraction we are using on the code)

packages/protocol/docs/Derivation.md
davidtaikocha: i remember both should be fine, but we chose to require the same `proverAuthBytes` for all blocks (just) for the simplicity of the implementation on both client and prover.

PR #21060, opened 2025-12-19
fix(taiko-client): correct parent block parameters in Shasta base fee calculation log

prestoalvarez:
Fixes incorrect logging in `CalculateBaseFeeShasta` where parent block parameters were logged using the current block's values instead of the actual parent block data.


davidtaikocha:
Here it is `parent` block.

PR #21063, opened 2025-12-21
feat(protocol): move proving fee + bond processing back to L1

dantaik:
## Move proving fee + bond processing back to L1.

 I think it’s worth considering because:

- It will significantly simplify the Anchor function on L2 .
- It avoids creating a new issue around “low-bond proposals.”
- Paying proving fees in ETH on L1 is cheap in practice (liveness-bond penalties should be rare and can be handled lazily).
- L1 prover auction is something we need to cope with the sequential proving design we have in Shasta -- ProverAuth assumes each proposer can pick their own prover, but with sequential proving, this assumption is no longer valid.

## TODO
- [ ] Remove `ProverAuth `, `LibBonds` and `BondManager` from L2
- [ ] Simplify the Anchor transaction path
- [ ] Implement a minimal `ProverAuction` contract that effectively acts as a whitelist (no actual auction logic)

**Note:** I won’t be available to drive this PR to merge, but I do think the current `ProverAuth`-based approach is fundamentally incompatible with sequential proving. It’s better to address this now rather than later.


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -201,6 +201,9 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, EssentialContract {
             _coreState.lastProposalBlockId = uint48(block.number);
             _setProposalHash(proposal.id, LibHashOptimized.hashProposal(proposal));
 
+            // Transfer proving fee to ProverAuction contract
```
```suggestion
           
```

PR #21066, opened 2025-12-22
chore(protocol): validate Blacklist overseers

IrvinGa38:
Ensure the provided overseers list is not empty, and verify that each overseer address is non-zero.


davidtaikocha:
Can't see too much improvement here.

PR #21069, opened 2025-12-23
chore(taiko-client): reduce monitor interval

YoGhurt111:
ref: https://github.com/taikoxyz/taiko-mono/pull/21033

### DIFF
--- a/packages/taiko-client/prover/proof_submitter/proof_buffer_monitor.go
+++ b/packages/taiko-client/prover/proof_submitter/proof_buffer_monitor.go
@@ -9,20 +9,18 @@ import (
 	proofProducer "github.com/taikoxyz/taiko-mono/packages/taiko-client/prover/proof_producer"
 )
 
+const monitorInterval = 5 * time.Minute
+
 // startProofBufferMonitors launches a monitor goroutine per proof type so we can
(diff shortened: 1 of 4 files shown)


packages/taiko-client/prover/proof_submitter/proof_buffer_monitor.go
```diff
@@ -9,20 +9,18 @@ import (
 	proofProducer "github.com/taikoxyz/taiko-mono/packages/taiko-client/prover/proof_producer"
 )
 
+const monitorInterval = 5 * time.Minute
```
davidtaikocha, line 12: Would it be better if we made it configurable? If there is a chance the we want to change it in the future.

YoGhurt111: In my view, a 5-minute interval is reasonable and shouldn’t cause noticeable overhead. On the other hand, exposing an unreasonable configurable value could be risky. I don’t have a strong opinion either way, though—if you think configurability is better, I’m happy to make the change. Let me know your thoughts.

davidtaikocha (approved):

johntaiko (approved):

PR #21071, opened 2025-12-23
feat(taiko-client): the strategy of proof submission

YoGhurt111:
未命名文件


YoGhurt111:
> The condition at line 232-233 checks toID.Cmp(fromID) == 0 || toID.Cmp(...LastInsertID()+1) == 0. This logic won't properly handle sequential gap filling. If buffer has ID 2 and you receive ID 5 (goes to cache), then ID 3 arrives, it will ALSO cache (LastInsertID still 2), preventing cache flushes.

@claude When the ID 3 arrives, we don't want the ID 3 cached, we need it inserted into the buffer and then `TryAggregate` later.
And plz review again.

YoGhurt111:
@claude plz review again

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
```diff
@@ -216,20 +224,34 @@ func (s *ProofSubmitterShasta) RequestProof(ctx context.Context, meta metadata.T
...
+				return fmt.Errorf(
+					"failed to add proof into buffer (id: %d) (current buffer size: %d): %w",
+					meta.GetProposalID(),
+					bufferSize,
+					err,
+				)
+			}
+		} else {
+			cacheMap.set(toID, proofResponse)
+			if proofRangeCached(fromID, toID, cacheMap) {
```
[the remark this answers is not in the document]
YoGhurt111: fixed

YoGhurt111:
@claude plz review again

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
```diff
@@ -216,20 +224,35 @@ func (s *ProofSubmitterShasta) RequestProof(ctx context.Context, meta metadata.T
...
+			toID == toBeInsertedID {
+			bufferSize, err := proofBuffer.Write(proofResponse)
+			if err != nil {
+				return fmt.Errorf(
+					"failed to add proof into buffer (id: %d) (current buffer size: %d): %w",
+					meta.GetProposalID(),
+					bufferSize,
+					err,
+				)
+			}
```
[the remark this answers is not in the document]
YoGhurt111: The periodic flushing is enough.

packages/taiko-client/prover/proof_submitter/proof_cache.go
```diff
@@ -0,0 +1,44 @@
...
+// Thread-safety: All operations are protected by a RWMutex to allow concurrent
+// reads and exclusive writes.
+//
+// Lifecycle:
+// - Proofs are cached when they arrive out of order (not sequential with buffer)
+// - Cached proofs are flushed to buffer when the gap is filled
+// - Finalized proofs are cleaned up periodically by the background monitor
+type ProofCache struct {
+	mu    sync.RWMutex
+	cache map[uint64]*proofProducer.ProofResponse
```
davidtaikocha: Maybe we can remove the mutex and all the locking / unlocking below by introducing https://github.com/orcaman/concurrent-map ?

packages/taiko-client/prover/proof_submitter/proof_buffer_monitor.go
```diff
@@ -49,3 +54,134 @@ func monitorProofBuffer(
...
+			}
+			lastFinalizedProposalID := coreState.LastFinalizedProposalId
+			// remove stale cache
+			removeFinalizedProofsFromCache(cacheMap, lastFinalizedProposalID)
+			// flush cached proofs
+			toID := lastFinalizedProposalID.Uint64() + buffer.MaxLength
+			if err := flushProofCacheRange(lastFinalizedProposalID.Uint64(), toID, buffer, cacheMap); err != nil {
+				if !errors.Is(err, ErrCacheNotFound) {
+					log.Error("Failed to flush proof cache range",
+						"error", err,
```
davidtaikocha:
```suggestion
					log.Error(
					    "Failed to flush proof cache range",
						"error", err,
```

Let's maintain a consistent logging style in Go client.

packages/taiko-client/prover/proof_submitter/proof_buffer_monitor.go
```diff
@@ -49,3 +54,134 @@ func monitorProofBuffer(
...
+	for currentID <= toID {
+		cachedProof, ok := cacheMap.cache[currentID]
+		if !ok {
+			log.Error("cached proof not found for proposal", "proposalID", currentID)
+			return ErrCacheNotFound
+		}
+		if _, err := proofBuffer.Write(cachedProof); err != nil {
+			if errors.Is(err, proofProducer.ErrBufferOverflow) {
+				log.Info("Buffer overflow during cache flush, stop flushing",
+					"proposalID", currentID,
```
davidtaikocha:
```suggestion
				log.Info(
				    "Buffer overflow during cache flush, stop flushing",
					"proposalID", currentID,
```

packages/taiko-client/prover/init.go
```diff
@@ -139,10 +139,17 @@ func (p *Prover) initShastaProofSubmitter(ctx context.Context, txBuilder *transa
 	}
 
 	// Init proof buffers.
-	var proofBuffers = make(map[producer.ProofType]*producer.ProofBuffer, proofSubmitter.MaxNumSupportedProofTypes)
+	var (
+		proofBuffers = make(map[producer.ProofType]*producer.ProofBuffer, proofSubmitter.MaxNumSupportedProofTypes)
+		cacheMaps    = make(
```
davidtaikocha, line 144: Let's create enough unit tests to test the new strategy as well.

packages/taiko-client/prover/proof_submitter/proof_buffer_monitor.go
```diff
@@ -49,3 +54,134 @@ func monitorProofBuffer(
...
+			log.Error("Proof buffer for proof type not found", "proofType", proofType)
+			return
+		}
+		go cleanUpStaleCacheAndFlush(ctx, rpc, cacheMap, monitorInterval, buffer)
+	}
+}
+
+// cleanUpStaleCacheAndFlush periodically removes cached proofs that have already been
+// finalized so stale entries do not accumulate and flush cache into buffer.
+func cleanUpStaleCacheAndFlush(
```
davidtaikocha, line 76: If we just want to clean up finalized proofs in cache, shall we just do it in `prover/event_handler/batches_proved.go` each time there is a `Proved` event? Instead of keeping scanning it.

packages/taiko-client/prover/init.go
YoGhurt111: Will do.

packages/taiko-client/prover/proof_submitter/proof_buffer_monitor.go
YoGhurt111: I recommend keeping the scanning approach for two reasons:
1. It minimizes dependencies on other parts of the codebase, which helps avoid unexpected behavior or errors in edge cases.
2. We already have a similar code path that we can reference and reuse.

packages/taiko-client/prover/proof_submitter/proof_cache.go
YoGhurt111: Done

packages/taiko-client/prover/proof_submitter/proof_buffer_monitor.go
```diff
@@ -49,3 +56,102 @@ func monitorProofBuffer(
...
+) {
+	if lastFinalizedProposalID == nil {
+		return
+	}
+
+	for _, proposalID := range cacheMap.Keys() {
+		id, fail := new(big.Int).SetString(proposalID, 10)
+		if fail {
+			return
+		}
```
MatusKysel:
```suggestion
		id, ok := new(big.Int).SetString(proposalID, 10)
		if !ok {
			return
		}
```

this is wrong

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
```diff
@@ -509,3 +534,38 @@ func (s *ProofSubmitterShasta) WaitTransitionVerified(ctx context.Context, trans
...
+	}
+	var fromID *big.Int
+	if buffer.LastInsertID() > 0 {
+		fromID = new(big.Int).SetUint64(buffer.LastInsertID() + 1)
+	} else {
+		fromID = new(big.Int).Add(coreState.LastFinalizedProposalId, common.Big1)
+	}
+	toID := new(big.Int).Add(fromID, new(big.Int).SetUint64(buffer.AvailableCapacity()))
+	if err := flushProofCacheRange(fromID, toID, buffer, cacheMap); err != nil {
+		if !errors.Is(err, ErrCacheNotFound) {
```
MatusKysel, line 570: are we sure we don't want to log it ? maybe as a debug?

packages/taiko-client/prover/proof_producer/proof_buffer.go
```diff
@@ -115,6 +131,7 @@ func (pb *ProofBuffer) ClearItems(blockIDs ...uint64) int {
 	pb.buffer = newBuffer
 	if len(pb.buffer) == 0 {
 		pb.firstItemAt = time.Time{}
+		pb.lastInsertID = 0
```
MatusKysel, line 133: lastInsertID isn't updated when clearing the newest entry; can skip IDs after reorgs

packages/taiko-client/prover/prover.go
```diff
@@ -128,6 +129,7 @@ func InitFromConfig(
 	p.proveNotify = make(chan struct{}, 1)
 	p.batchesAggregationNotifyPacaya = make(chan proofProducer.ProofType, proofSubmitter.MaxNumSupportedProofTypes)
 	p.batchesAggregationNotifyShasta = make(chan proofProducer.ProofType, proofSubmitter.MaxNumSupportedProofTypes)
+	p.flushCacheNotify = make(chan proofProducer.ProofType, 1)
```
MatusKysel, line 132: buffer size 1 + blocking sends can deadlock when RequestProof enqueues flushes

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
```diff
@@ -509,3 +534,38 @@ func (s *ProofSubmitterShasta) WaitTransitionVerified(ctx context.Context, trans
...
+	if err != nil {
+		return fmt.Errorf("failed to get Shasta core state: %w", err)
+	}
+	var fromID *big.Int
+	if buffer.LastInsertID() > 0 {
+		fromID = new(big.Int).SetUint64(buffer.LastInsertID() + 1)
+	} else {
+		fromID = new(big.Int).Add(coreState.LastFinalizedProposalId, common.Big1)
+	}
+	toID := new(big.Int).Add(fromID, new(big.Int).SetUint64(buffer.AvailableCapacity()))
```
MatusKysel, line 568: `AvailableCapacity` is a count; inclusive range should use `fromID+capacity-1`

YoGhurt111: There is no error here. When capacity is zero, `toID` should be equal to `fromID`.

packages/taiko-client/prover/proof_producer/proof_buffer.go
YoGhurt111: IDs in the proof buffer are contiguous. If cleanup reaches the newest ID, it means the buffer has been fully cleared.

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
YoGhurt111: We don't treat this as an error when we try to flush the cache into the buffer and don't found one of them.

packages/taiko-client/prover/prover.go
YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/21071/commits/49f728b01f4ae6f044737d82223c7e64ea65e845

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
MatusKysel: yeah I know, but maybe we should still print a log

packages/taiko-client/prover/proof_producer/proof_buffer.go
MatusKysel: Contiguity doesn’t guarantee “newest removed ⇒ buffer empty.” ClearItems can remove a suffix (reorg case) while leaving earlier IDs intact. ValidateProof explicitly treats L1 reorgs as invalid, so if a reorg hits later proposals, you’ll clear the newest IDs and still keep older ones. That leaves lastInsertID stale and makes toBeInsertedID skip missing IDs, so cached proofs won’t flush. I think we should set it to 0, only if it's empty. e.g.:

  - Buffer has [10,11,12], lastInsertID=12.
  - Reorg invalidates 12 → ClearItems(12) leaves [10,11], lastInsertID still 12.
  - Next proof for 12 goes to cache because toBeInsertedID=13.
  - FlushCache starts at 13, so 12 is never flushed; when buffer empties, flush starts at lastFinalized+1 and fails on missing 10, so it still won’t reach 12.

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
MatusKysel: you are right. the issue is just an extra attempt/log: with an inclusive loop and toID = fromID + capacity, you try capacity+1 IDs. When capacity is 0 you still attempt one write and log an overflow; when the cache doesn’t have the extra ID you hit ErrCacheNotFound. If you want to avoid that, guard capacity or use capacity-1 for the inclusive end.

packages/taiko-client/prover/proof_producer/proof_buffer.go
YoGhurt111: Actually, we have multiple mechanisms in place to prevent reorg cases. That said, from a code-logic completeness standpoint, I can optimize this part.

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
YoGhurt111: IMO, we can ignore this, since it's expected and frequently occurs.

packages/taiko-client/prover/proof_producer/proof_buffer.go
YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/21071/commits/37cfc8ace3ab9e799c66de70d176a3b20732ec11

packages/taiko-client/prover/proof_submitter/proof_submitter_shasta.go
YoGhurt111: We can accept that

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main,feature/new_strategy_of_proof_submission]
```
```suggestion
    branches: [main]
```

davidtaikocha:
@claude plz review.

MatusKysel (approved):

davidtaikocha (approved):

davidtaikocha (approved):

PR #21073, opened 2025-12-23
feat(protocol): move bonds to l1

ggonzalez94:
A simpler proposal instead of #21063 , where we keep the existing incentive system(instead of formalizing the idea of prover auction/market) but do it directly on L1.

## Context
We originally moved bond processing to L2 in order to make propose and prove cheaper. This has worked, but has created complexity in derivation(low bond proposals), driver, prover, relayer and sidecars.
While originally seemed like a good tradeoff, the current version of the protocol, and the L1 fees are cheap enough that it makes sense to consider moving bonds back to L1.
Furthermore, we'll likely rollout shasta with a whitelisted prover, so no bond slashing will take place. We will consider a prover market and incentive mechanism for permisionless preconfs.

## Summary
This PR moves the existing bond system back to L1. Bonds are handled directly on the inbox contract, and slashing is only active is whitelisted proving is not enabled.
It also removes prover delegation and `ProverAuth` entirely, since doing those checks on the L1 would become gas intensive, and it is not a feature currently used or demanded. Moreover, until we think more deeply about the prover market Taiko is likely to be the whitelisted prover, so introducing the complexity and gas cost seems unnecessary and wasteful.

## TODOs
- [ ] Review Derivation pipeline after the changes
- [ ] Decide how to deal with changing bonds. The amount can change between propose and prove time, causing an unbalance and accounting errors in the contract
         https://github.com/taikoxyz/taiko-mono/pull/21082
- ~[ ] Merge LibBonds into the Inbox directly? The library is small enough that we can consider removing it and just introducing the code into the inbox itself~
     After thinking more about it, this is probably not a good idea
- [x] Optimize bond refunds by processing batches

---

> [!NOTE]
> Moves bond lifecycle to L1 and simplifies prover flow.
> 
> - Introduces `IBondManager` on L1 and implements it in `Inbox` with `deposit`, `depositTo`, `withdraw`, and `getBondBalance`; adds `bondToken` and `livenessBond` to `Inbox.Config`
> - New L1 `LibBonds` handles balances, batched credits, deposits/withdrawals, and liveness slashing; `propose` debits bond, `prove` refunds or slashes via `_settleProposalBonds` (no L1→L2 bond signals)
> - Removes L2 `BondManager`, `IBondManager`, `IBondProcessor`, and shared `LibBonds`; deletes bond instruction event and `ICodec.hashBondInstruction`
> - Simplifies `Anchor`: drops prover delegation/`ProverAuth`, proposal-level prover state, and bond checks; keeps proposal ID monotonicity and checkpoint anchoring
> - Updates `DevnetInbox`/`MainnetInbox` constructors, deployment script (`DeployProtocolOnL1.s.sol`) to pass `bondToken`/`livenessBond`; genesis/test configs drop L2 bond settings
> - Refreshes derivation docs to reflect L1 bonding and removal of low-bond/prover auth paths; adds design notes for Inbox/LibBonds
> - Test suite overhauled to assert L1 bond flows, liveness slashing behavior, whitelist bypass, and updated gas snapshots
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 932d8520ddfa18fc186a5f4de642f4cc3a2d3b6b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


dantaik:
https://github.com/taikoxyz/taiko-mono/pull/21074

@ggonzalez94 maybe the ProverAuction can replace the BondManager on L1

ggonzalez94:
> #21074
> 
> @ggonzalez94 maybe the ProverAuction can replace the BondManager on L1

@dantaik  I'll check the new changes. In general I really like the idea of a prover market(we had the idea of designing it since the first time we met), and still see it as the final destination. But I'm afraid that for shasta this might be too big of a change.
- It is too large of a change to fit the current audit
- It changes the incentive mechanism completely. I think we should spend time thinking about this before considering its deployment

This is why I propose we just move the current incentive system to L1(protocol becomes simpler, we get rid of the concept of low bond proposals entirely, and the node and prover also get simplified), and work on a prover market for the next hard fork.
What do you think?

ggonzalez94:
I'll still iterate and do some reviews, but this should be ready for review now.

packages/protocol/contracts/layer2/core/BondManager_Layout.sol
```diff
@@ -7,17 +7,17 @@ pragma solidity ^0.8.24;
 /// @custom:security-contact security@taiko.xyz
 
 // solhint-disable max-line-length
-//   _initialized                   | uint8                                              | Slot: 0    | Offset: 0    | Bytes: 1   
-//   _initializing                  | bool                                               | Slot: 0    | Offset: 1    | Bytes: 1   
+//   _initialized                   | uint8                                              | Slot: 0    | Offset: 0    | Bytes: 1
```
dantaik, line 10: lets also delete this file.

packages/protocol/script/gen-layouts.sh
```diff
@@ -50,7 +50,6 @@ contracts_layer2=(
 "contracts/layer2/mainnet/BridgedTaikoToken.sol:BridgedTaikoToken"
 "contracts/layer2/governance/DelegateController.sol:DelegateController"
 "contracts/layer2/core/Anchor.sol:Anchor"
-"contracts/layer2/core/BondManager.sol:BondManager"
```
dantaik, line 53: in line 99, please change to use `pragma solidity ^0.8.26;` instead

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -18,6 +17,10 @@ interface IInbox {
         address proverWhitelist;
         /// @notice The signal service contract address
         address signalService;
+        /// @notice The ERC20 bond token address
+        address bondToken;
+        /// @notice The liveness bond amount
+        uint256 livenessBond;
```
dantaik, line 23: We should prefer a smaller type like uint64 and use gwei or even 1 TAIKO token as bond's base unit. We generally follow that pattern already, so livenessBond probably shouldn’t be an exception.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -193,6 +208,10 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, EssentialContract {
             uint48 lastFinalizedProposalId = _coreState.lastFinalizedProposalId;
             require(nextProposalId > 0, ActivationRequired());
 
+            uint256 bondAmount = _livenessBond;
+            // During permisionless proving we can set the bond to 0, avoiding the cost of this check
+            if (bondAmount > 0) LibBonds.debitBond(_bondingStorage, msg.sender, bondAmount);
```
dantaik, line 213:
```suggestion
            if (bondAmount > 0) {LibBonds.debitBond(_bondingStorage, msg.sender, bondAmount);}
```

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -193,6 +208,10 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, EssentialContract {
             uint48 lastFinalizedProposalId = _coreState.lastFinalizedProposalId;
             require(nextProposalId > 0, ActivationRequired());
 
+            uint256 bondAmount = _livenessBond;
```
dantaik, line 211: I think we can use _livenessBond directly below without extra cost?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
dantaik, line 213: One idea I really like is **lazy bond processing**: at proposal time, we only verify that the actor has enough bond locked and that it stays locked long enough. Any penalties can be applied later (i.e., not during proposal submission).

That means proposing only needs a single SLOAD, and we can skip the “debit then refund” liveness bond flow, which should cut gas costs significantly.

dantaik: **Lazy bond processing** is also implemented by the ProverAuction in https://github.com/taikoxyz/taiko-mono/pull/21074, and there is only 1 SLOAD extra cost by using it by the inbox.

dantaik, line 213: if there is a white list, then we should also skip debiting the bond.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -464,6 +473,30 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, EssentialContract {
         return _proposalHashes[_proposalId % _ringBufferSize];
     }
 
+    // ---------------------------------------------------------------
+    // Bond Manager Functions
+    // ---------------------------------------------------------------
+
+    /// @inheritdoc IBondManager
+    function deposit(uint256 _amount) external nonReentrant {
```
dantaik, line 481: A potentially cleaner alternative is to introduce a BondManagerBase.sol that implements IBondManager, and then have Inbox inherit from BondManagerBase.

That said, I still think adopting ProverAuction directly is the better approach overall. It would also let us deprecate IProverWhitelist.

The main selling point of ProverAuction is that it completely hides bond accounting from Inbox.

packages/protocol/script/layer1/core/DeployProtocolOnL1.s.sol
```diff
@@ -103,6 +105,8 @@ contract DeployProtocolOnL1 is DeployCapability {
         config.sharedResolver = vm.envAddress("SHARED_RESOLVER");
         config.remoteSigSvc = vm.envOr("REMOTE_SIGNAL_SERVICE", msg.sender);
         config.preconfWhitelist = vm.envOr("PRECONF_WHITELIST", address(0));
+        config.bondToken = vm.envAddress("BOND_TOKEN");
```
davidtaikocha, line 108: Shall we just use `config.taikoToken = vm.envAddress("TAIKO_TOKEN");` as the bond token in this script? and remove the `require(config.bondToken != address(0), "BOND_TOKEN not set");` check since we may still need to deploy the `TaikoToken` in this script.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: I was thinking this would re-introduce low bond proposals(which was one of the main complexities we were trying to eliminate), but now that I think about it more I don't think that's the case. 
I'll try to work on it, what I'm thinking:
- we'll have to re-introduce the delay with bond withdrawals
- bond slashing, etc. will have to be best effort again(since you cannot fully guarantee that no more proposals were created that bonds outstanding)

This will re-introuduce some complexity, but it might be worth it

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: > A potentially cleaner alternative is to introduce a BondManagerBase.sol that implements IBondManager, and then have Inbox inherit from BondManagerBase

In general agree, but we've been followingthe library pattern in the inbox(e.g. with forced inclusions), so it feels more consistent if we do the same for others

ggonzalez94:
currently considering #21085 a better option. Will close this once we have a final decision

PR #21076, opened 2025-12-24
fix(eventindexer): correct transition contested error context

Snezhkko:
The error wrapping messages in saveTransitionContestedEvents and saveTransitionContestedEventsV2 were copy-pasted from a different handler and referenced i.saveBlockProvenEvent. This makes logs and traces misleading when a transition contested event fails. The messages are updated to reference the correct functions, so operational debugging and metrics correlation point to the actual failing code paths.


davidtaikocha:
Can't see too many improvements there: https://github.com/taikoxyz/taiko-mono/blob/main/CONTRIBUTING.md

PR #21077, opened 2025-12-24
fix(relayer): reuse cached chain IDs in HTTP handlers

Fibonacci747:
Use the srcChainID and destChainID values cached on the Server instead of issuing new ChainID RPC calls on every HTTP request in GetRecommendedProcessingFees and GetBlockInfo. Chain IDs are immutable for a given network and already validated during server initialization, so calling ChainID again per request adds unnecessary latency and load without improving correctness.


davidtaikocha:
@claude plz review.

davidtaikocha (approved):

MatusKysel (approved):

PR #21081, opened 2025-12-26
fix(taiko-client): correctly handle wrong blob data

YoGhurt111:
When blob data was filled with arbitrary bytes, which violates EIP-4844 field element constraints and causes blob decoding failures, we need to correctly handle this case(produce empty block).


YoGhurt111:
@claude Changed the description. plz review again.

YoGhurt111:
@claude plz review again

YoGhurt111:
> CRITICAL: Blob Server Path Lacks Hash Validation (Security Issue)
@claude If the returned blobs don't match the requested blobHashes,it won't pass the following check
```
if len(sidecars) != len(blobHashes) {
		return nil, fmt.Errorf("blob sidecar count mismatch: expected %d, got %d", len(blobHashes), len(sidecars))
	}
```

davidtaikocha (approved):

davidtaikocha:
@claude plz review.

johntaiko (approved):

PR #21082, opened 2025-12-26
feat(protocol): account for changing bonds during inbox upgrades

ggonzalez94:
Ref #21073 

During contract upgrades, the value for `livenessBond` can change. This is a mechanism that the SC or the DAO can use to reduce or increase barriers to entry or penalties. The problem is that if the value changes(via a contract upgrade) after a proposal is made and the time it gets proven, we will have an accounting issue.
- if the value is reduced, the proposer won't be failry compensated. While not ideal, this is less of an issue.
- if the value is increased, the contract will use the balance the pay the proposer, which can lead to the contract becoming insolvent.

This PR proposes a solution to this problem, creating a commitment at proposal time and using the transition struct(enforced by the prover) to pass this value, aligning bond values at proposal and prove time. @davidtaikocha @smtmfft please review it and let me know what you think.

---

> [!NOTE]
> Aligns bond accounting between propose and prove by carrying `livenessBond` per proposal through commitments.
> 
> - Add `livenessBond` to `IInbox.Proposal` and `IInbox.Transition`; `_buildProposal` stores current `_livenessBond`; `propose` debits bond only when > 0
> - Update commitment hashing (`LibHashOptimized.hashCommitment`) and transition codec (`LibTransitionCodec`, `TRANSITION_SIZE=110`) to include `livenessBond`
> - Settle bonds in `Inbox._settleProposalBonds` using `transitions[_offset].livenessBond` for late slashing and refund remaining via `LibBonds.creditBonds`
> - Introduce `LibBonds.creditBonds` to aggregate proposer refunds from `Transition[]`; remove fixed-amount refund path
> - Extend tests to cover per-proposal bond across upgrades, zero-bond proposing, and updated prove paths; refresh gas reports and snapshots (including `propose_single_zero_bond`)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1f7864764a21923319c11bf9ab2824ad2be9637d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -72,6 +73,8 @@ interface IInbox {
         bytes32 originBlockHash;
         /// @notice The percentage of base fee paid to coinbase.
         uint8 basefeeSharingPctg;
+        /// @notice The liveness bond amount reserved for this proposal at propose time.
+        uint256 livenessBond;
```
dantaik, line 77: can we use a smaller type such as uint96?

dantaik:
Lets merge this PR to its base and keep working on the base branch.

dantaik (approved):

davidtaikocha (approved):

packages/protocol/contracts/layer1/core/iface/IInbox.sol
davidtaikocha, line 77: Let's also make this part of the `Proposed` event payload? Would be great for off-chain software to track.

ggonzalez94:
currently considering #21085 a better option. Will close this once we have a final decision

PR #21085, opened 2025-12-29
feat(protocol): optimistic bond processing on L1

ggonzalez94:
Similar goals to #21073 but using optimistic bond checks.

We don't debit the bonds on propose, but instead ask the proposer to hold a minimum amount of bond on L1, and only debit from that bond when slashing occurs.
This is the same mechanism we were using on L2, but now moved to L1 - which still eliminates the complexity of low bond proposals.

This approach achieves most of the benefits of #21073 , but with less code and more gas efficient.

---

> [!NOTE]
> Introduces optimistic, ERC20-backed bond accounting directly in L1 `Inbox`, eliminating L2 bond signaling and Anchor prover-designation/low-bond logic.
> 
> - Adds `IBondManager` and `LibBonds` on L1; `Inbox` now implements bond ledger: `deposit`, `depositTo`, `withdraw`, `requestWithdrawal`, `cancelWithdrawal`, `getBond`
> - New `Inbox` config/immutables: `bondToken`, `minBond`, `livenessBond`, `withdrawalDelay`; enforce `minBond` in `_buildProposal` (permissionless inclusions exempt); withdrawal request disables proposing
> - Liveness handling: settle on L1 during `prove` via `_processLivenessBond` (50% credit to prover, 50% burn); removes L2 bond instruction signaling and `hashBondInstruction`
> - Removes L2 `BondManager`, `IBondProcessor`, and shared `LibBonds`; simplifies `Anchor.anchorV4` (drops prover auth/low-bond) and related docs
> - Updates constructors (Devnet/Mainnet Inbox), deploy scripts, genesis configs, and tests to use L1 bonds; adjusts forced inclusion storage calls; bumps pragmas/layout docs; refreshes gas reports/snapshots
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dca5448ade13346e09d800ce0aa84899db2c6c25. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/core/libs/LibBonds.sol
```diff
@@ -0,0 +1,207 @@
...
+        if (
+            bond_.withdrawalRequestedAt == 0
+                || block.timestamp < bond_.withdrawalRequestedAt + _withdrawalDelay
+        ) {
+            require(balance - amount >= _minBond, MustMaintainMinBond());
+        }
+
+        debited_ = _debitBond($, _from, amount);
+        _bondToken.safeTransfer(_to, _toTokenAmount(debited_));
+        emit IBondManager.BondWithdrawn(_from, debited_);
```
davidtaikocha, line 88: Shall we reset `bond_.withdrawalRequestedAt = 0;` here in `withdraw`? Otherwise, after a withdrawal, if a proposer wants to re-join again later, he needs to manually call `cancelWithdrawal` (although the withdrawal was finished, and there would be one `WithdrawalCancelled` event) again before calling `deposit`? Otherwise, if he only just calls `deposit`, looks like the `hasSufficientBond` check will still fail?

packages/protocol/docs/Derivation.md
```diff
@@ -321,14 +315,13 @@ Note: Fields like `stateRoot`, `transactionsRoot`, `receiptsRoot`, `logsBloom`,
 
 ### Anchor Transaction
 
-The anchor transaction serves as a privileged system transaction responsible for L1 state synchronization. It invokes the `anchorV4` function on the ShastaAnchor contract with precisely defined parameters:
+The anchor transaction serves as a privileged system transaction responsible for L1 state synchronization. It invokes the `anchorV4` function on the ShastaAnchor contract with the L1 checkpoint fields:
 
-| Parameter         | Type    | Description                                      |
-| ----------------- | ------- | ------------------------------------------------ |
-| proposalId        | uint48  | Unique identifier of the proposal being anchored |
```
davidtaikocha: Let's add it to the derivation rule in `block.extradata`?

packages/protocol/docs/Derivation.md
```diff
@@ -321,14 +315,13 @@ Note: Fields like `stateRoot`, `transactionsRoot`, `receiptsRoot`, `logsBloom`,
...
-| anchorBlockNumber | uint48  | L1 block number to anchor (0 to skip anchoring)  |
-| anchorBlockHash   | bytes32 | L1 block hash at anchorBlockNumber               |
-| anchorStateRoot   | bytes32 | L1 state root at anchorBlockNumber               |
+| Parameter         | Type    | Description                                     |
+| ----------------- | ------- | ----------------------------------------------- |
+| anchorBlockNumber | uint48  | L1 block number to anchor (0 to skip anchoring) |
```
smtmfft, line 323: Do we still need to support skipping? I think 0 is the same as invalid anchor block number, which will be corrected by client to the last anchor, is it?

smtmfft (approved):
LGTM!

AnshuJalan (approved):
LGTM! 

A small nit you may want to consider - since `LibBonds` has a storage pointer in its functions, you may use the `_bondStorage.<function>()` syntax

packages/protocol/docs/Derivation.md
ggonzalez94: maybe still ok for convenience? i.e. what value to set when you don't want to set a new anchor.
Not a strong opinion here

packages/protocol/contracts/layer1/core/libs/LibBonds.sol
ggonzalez94: Here's what I ended up doing:
1. cancel withdraw request when calling deposit(not depositTo since we don't want to cancel the request on behalf of other proposers)
2. cancel withdraw request, but only when withdrawing the full amount of the bond(e.g. you don't want to cancel the request if the person is exiting and they withdraw an excess amount before)

Let me know what you think

packages/protocol/docs/Derivation.md
ggonzalez94: added. Let me know if that's what you had in mind

davidtaikocha (approved):

PR #21092, opened 2025-12-30
feat(protocol): replace BondManager with ProverAuction

dantaik:
Replaced by https://github.com/taikoxyz/taiko-mono/pull/21102


dantaik:
@claude please review

ggonzalez94:
Very interesting... a few questions while I continue looking into the code:
1. when underbid are provers in the pool now expected to work with the new fee? and then expected to monitor the events and exit if they don't agree?
2. why is the prover selection weighted? so that the one who actually made the lower bid gets more work?

> Allow proposers to either use the auction winner (current prover) or self-prove their own proposals.

this is planned but not yet implemented right? is the idea that this is a flag that the proposer sets and they can decide?

dantaik:
> Very interesting... a few questions while I continue looking into the code:
> 
> 1. when underbid are provers in the pool now expected to work with the new fee? and then expected to monitor the events and exit if they don't agree?

Provers that are out-bided will force to exist immediately. They need to monitor the events to decide if they want to follow the new winner's bid or even further outbid the new winner to lead another round.

> 2. why is the prover selection weighted? so that the one who actually made the lower bid gets more work?
The idea is that if Alice, Bob, and Charley are the 3 provers with the same proving fees, If Alice is the auction winner, Bob and Charley simply joins the pool, Alice will have the biggest change to be selected, followed by Bob, then Charley. This encourage Alice to bid with a lower proving fees first.

> > Allow proposers to either use the auction winner (current prover) or self-prove their own proposals.
> 
> this is planned but not yet implemented right? is the idea that this is a flag that the proposer sets and they can decide?

This is now implemented.

dantaik:
@ggonzalez94 @davidtaikocha @claude 
This PR is ready for review. I will also review the ProverAuction.sol code line-by-line myself (I need 48 hours). Until this point, the code is reviewed by AI only.

If you guys like this direction, maybe we should merge this PR before merging v3.0.0 back to main.

ggonzalez94:
> @ggonzalez94 @davidtaikocha @claude This PR is ready for review. I will also review the ProverAuction.sol code line-by-line myself (I need 48 hours). Until this point, the code is reviewed by AI only.
> 
> If you guys like this direction, maybe we should merge this PR before merging v3.0.0 back to main.

Will start a more deep dive review this weekend. I don't think we should merge to the audit branch, because we will likely not have enough time to audit and I also don't think we should rush a prover market/auction feature before intensive testing and more thinking on the incentives. We will start with the whitelisted prover anyways, so we don't need to rush on shipping this with shasta. Would you agree?
My more detailed answer is in this [slack message](https://taikolabs.slack.com/archives/C0970375FFE/p1766977357229709) on our internal channel

PR #21102, opened 2026-01-05
feat(protocol): replace BondManager with ProverAuction (new PR)

dantaik:
## Summary

This PR is a recreation of https://github.com/taikoxyz/taiko-mono/pull/21092 with a new branch name but same code.

This PR introduces a **ProverAuction** (a continuous reverse auction with **lazy bond processing**) to replace BondManager for better code decoupling from the Inbox.

## Key ideas

- Proving fees are paid in ETH to minimize payment-processing overhead on L1.
- Bonds remain denominated in TAIKO but managed entirely by ProverAuction with Inbox knows as little as possible about bonds -- Liveness bonds are managed by ProverAuction, so the Inbox does not need to track or reason about them (they’re effectively transparent to the Inbox).
- Allow proposers to either use the auction winner (current prover) or self-prove their own proposals.
- The **Proposed** event now emits the entire encoded Proposal
- Add **nonReentrance** to `prove()`


dantaik:
@claude please review

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -542,29 +519,41 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+            // Get designated prover and fee from auction
+            (address designatedProver, uint32 feeInGwei) = _proverAuction.getProver();
+            if (designatedProver == address(0)) {
+                // No auction winner - proposer becomes the prover but must have sufficient bond
+                designatedProver = msg.sender;
+                require(
+                    _proverAuction.checkBondDeferWithdrawal(msg.sender), InvalidSelfProverBond()
+                );
+            }
+            proverFee_ = uint256(feeInGwei) * 1 gwei;
```
ggonzalez94: nit: should we handle all the fees in gwei and only convert at the moment of transfering? if there are places where we do gwei accounting and others where we use wei, there's more room for error

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,755 @@
...
+        movingAverageWindow = _movingAverageWindow;
+        maxFeeDoublings = _maxFeeDoublings;
+        initialMaxFee = _initialMaxFee;
+        movingAverageMultiplier = _movingAverageMultiplier;
+
+        unchecked {
+            uint128 ejectionThreshold = uint128(_livenessBondAmount) * _bondMultiplier;
+            _ejectionThreshold = ejectionThreshold;
+            _requiredBond = ejectionThreshold * 2;
+        }
```
ggonzalez94: I imagine what we wanna represent here is: the bond we require should be several times larger than the liveness bond.

If so I think we can simplify this by just having a single `ejectionThreshold` parameter and doing `require(ejectionThreshold > _livenessBondAmount)` and not trying to come up with specific multipliers(we can do the math outside the contract and just input it here at construction).

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,755 @@
...
+        __Essential_init(_owner);
+        _contractCreationTime = uint48(block.timestamp);
+    }
+
+    // ---------------------------------------------------------------
+    // External Transactional Functions
+    // ---------------------------------------------------------------
+
+    /// @inheritdoc IProverAuction
+    function deposit(uint128 _amount) external nonReentrant {
```
ggonzalez94, line 171: nit: if we swap the order of the state update and the transfer we don't need a reentrancy check and it follows the general check-effects-interact pattern in contracts.

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,755 @@
...
+        bondToken.safeTransferFrom(msg.sender, address(this), _amount);
+        _bonds[msg.sender].balance += _amount;
+        emit Deposited(msg.sender, _amount);
+    }
+
+    /// @inheritdoc IProverAuction
+    function withdraw(uint128 _amount) external nonReentrant {
```
ggonzalez94: nit: No need for reentrancy check?

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,755 @@
...
+
+    // ---------------------------------------------------------------
+    // Structs
+    // ---------------------------------------------------------------
+
+    struct PoolState {
+        uint32 feeInGwei;
+        uint8 poolSize;
+        uint48 vacantSince;
+        uint8 everHadPool;
```
ggonzalez94: maybe I'm missing something, but can't we just check `poolSize` and `vacantSince` to decide if the contract ever had a pool? If the answer is no, both are zero

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,755 @@
...
+        BondInfo storage bond = _bonds[msg.sender];
+        PoolState memory pool = _pool;
+        address oldLeader = _activeProvers[0];
+
+        bool isVacant = pool.poolSize == 0 || pool.vacantSince > 0;
+        PoolMember memory member = _members[msg.sender];
+        bool isMember = member.active;
+
+        if (isMember) {
+            require(block.timestamp >= MIN_SELF_BID_INTERVAL + _lastAvgUpdate, SelfBidTooFrequent());
```
ggonzalez94: maybe it is better/cleaner to allow frequent self bid, but not update the moving average if not enough time has happened? otherwise we hurt an existing prover's ability to reclaim the top of the queue if they are outbid

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,755 @@
...
+            return (_activeProvers[0], pool.feeInGwei);
+        }
+        // Use a hash of the block prevrandao to avoid predictable slot cycling.
+        // SLOT_TABLE_SIZE is 256 so the uint8 cast is an implicit modulo.
+        uint8 slot = uint8(uint256(keccak256(abi.encodePacked(block.prevrandao, address(this)))));
+        uint8 idx = _readSlotTable(slot);
+        return (_activeProvers[idx], pool.feeInGwei);
+    }
+
+    /// @notice Get the maximum allowed bid fee at the current time.
```
ggonzalez94: nit: the logic in this function is complex, it took me a while to wrap myself around the different cases. So maybe we should improve the natspec at least:


```suggestion
   /// @notice Returns the maximum fee a prover can bid at the current time.
    /// @dev Behavior depends on pool state:
    ///      - Active pool: Returns current fee reduced by `minFeeReductionBps` (enforces minimum undercut).
    ///      - Vacant pool: Implements a reverse Dutch auction where the max fee doubles every
    ///        `feeDoublingPeriod` (capped at `maxFeeDoublings`), starting from
    ///        `max(lastFee, initialMaxFee, movingAvgFee × multiplier)`.
```

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,755 @@
...
+        PoolState memory pool = _pool;
+        address oldLeader = _activeProvers[0];
+
+        bool isVacant = pool.poolSize == 0 || pool.vacantSince > 0;
+        PoolMember memory member = _members[msg.sender];
+        bool isMember = member.active;
+
+        if (isMember) {
+            require(block.timestamp >= MIN_SELF_BID_INTERVAL + _lastAvgUpdate, SelfBidTooFrequent());
+            require(_feeInGwei < pool.feeInGwei, FeeMustBeLower());
```
ggonzalez94: when someone is a member should we still force that they undercut the best bid by a minimum %? otherwise provers will continuously outbid each other, wasting gas, at least until an equilibrium is reached(which will change as gas costs, and proving costs change)

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,755 @@
...
+            uint8 joinOrder = pool.poolSize + 1;
+            require(bond.balance >= getRequiredBond(), InsufficientBond());
+            if (bond.withdrawableAt != 0) bond.withdrawableAt = 0;
+
+            _addToPool(msg.sender, joinOrder);
+            emit PoolJoined(msg.sender, joinOrder);
+            return;
+        }
+
+        require(_feeInGwei < pool.feeInGwei, FeeMustBeLower());
```
ggonzalez94: redundant since we are checking that `_feeInGwei <= maxAllowedFee` below?

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,755 @@
...
+        return uint16(10_000 - decay);
+    }
+
+    /// @dev Resets the pool to a single leader and rebuilds the slot table.
+    function _resetPool(address leader, uint32 fee) internal {
+        _pool.feeInGwei = fee;
+        _pool.poolSize = 1;
+        _pool.vacantSince = 0;
+        // Sticky flag: once a pool exists, we never revert to "never had pool" state.
+        _pool.everHadPool = 1;
```
ggonzalez94: I don't think this is the right place to set `everHadPool` since this is never called the very first time someone bids?

ggonzalez94 (commented):
Just finished a first pass and left a few comments. But here are a few general observations/questions:

1. Why do we treat outbids by existing provers in the pool different than by new bidders? Shouldn’t we just use the same mechanism and just push everyone one place in the queue?
2. Why do we need weighted selection if all the provers in the pool are charging the same fee? would it be simpler to just do random selection?
3. Does this mechanism achieve optimal aggregation? what happens if provers outbid each other often(the expected fee makes sense for them to bid), thus minimizing aggregation and pushing prices up for users. Under this assumption provers are acting rationally, but since they don’t expect to be able to aggregate for long enough they don’t bid lower. The counter argument to this is a prover can just bid lower, knowing that will grant them more time to aggregate. But the strategy seems complex.
4. (This is a point beyond the proposed implementation, and general to any prover market): Do provers have the necessary tools to estimate costs and bid efficiently? Since they are going to bid on expected value and without knowing the complexity of the batches being submitted. Currently too expensive to prove batches can be submitted, but soon(ish) zk gas should fix this last bit

---
On a separate note on code implementation:
- we use unchecked blocks in multiple parts of the code, but more sporadically(i.e. not in the entire function body). This makes thinking about unchecked cases/errors harder. Should we go back to using it in entire function’s bodies or not using it at all?

dantaik:
@ggonzalez94 all these questions from you are very good. I wonder if you want to take over this PR if you like this direction, or you prefer to use IBondManager for Shasta for now.

dantaik:
> Just finished a first pass and left a few comments. But here are a few general observations/questions:
> 
> 1. Why do we treat outbids by existing provers in the pool different than by new bidders? Shouldn’t we just use the same mechanism and just push everyone one place in the queue?
Maybe we can, but new bider and existing pool member may have different bond status. 

> 2. Why do we need weighted selection if all the provers in the pool are charging the same fee? would it be simpler to just do random selection?

The idea is to encourage people to bid a new low for a higher chance of being selected. The order of joining the pool make provers have different odds and thus different returns per capital locked, which makes collusion harder. But this does create complexity, we can remove weighted section.

> 3. Does this mechanism achieve optimal aggregation? what happens if provers outbid each other often(the expected fee makes sense for them to bid), thus minimizing aggregation and pushing prices up for users. Under this assumption provers are acting rationally, but since they don’t expect to be able to aggregate for long enough they don’t bid lower. The counter argument to this is a prover can just bid lower, knowing that will grant them more time to aggregate. But the strategy seems complex.

Gas fees are independent of proving fees, I don't think minimizing proof aggregation can push gas fees up.

> 4. (This is a point beyond the proposed implementation, and general to any prover market): Do provers have the necessary tools to estimate costs and bid efficiently? Since they are going to bid on expected value and without knowing the complexity of the batches being submitted. Currently too expensive to prove batches can be submitted, but soon(ish) zk gas should fix this last bit

Provers will have to learn from practice whether they are making a profit or suffering a loss. Regardless of how the auction is done, I believe this is always the case.

> 
> On a separate note on code implementation:
> 
> * we use unchecked blocks in multiple parts of the code, but more sporadically(i.e. not in the entire function body). This makes thinking about unchecked cases/errors harder. Should we go back to using it in entire function’s bodies or not using it at all?

Agreed.

dantaik:
@ggonzalez94 I think you’ll be more comfortable with a simplified auction implementation. I plan to simplify it by removing:

- weighted selection
- support for multiple active provers

I previously had a much simpler version, but as more ideas were added, the code became more complex than I’d like. I’ll address your concerns in the next few commits and then ask you to take another look.

dantaik:
@ggonzalez94 ready for another look.

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -103,18 +97,22 @@ interface IInbox {
     struct ProposeInput {
         /// @notice The deadline timestamp for transaction inclusion (0 = no deadline).
         uint48 deadline;
-        /// @notice Blob reference for proposal data.
-        LibBlobs.BlobReference blobReference;
+        /// @notice Maximum proving fee in Gwei the proposer is willing to pay.
+        /// @dev If 0, always self-prove without using the auction prover.
+        /// Use type(uint32).max to accept any fee.
+        uint32 maxProvingFeeGwei;
```
ggonzalez94, line 103: I really like this addition of giving the proposer the flexibility to set their max accepted fee or self prove

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,377 @@
...
+    /// @param _amount The amount of tokens to deposit.
+    function deposit(uint128 _amount) external nonReentrant {
+        BondInfo storage info = _bonds[msg.sender];
+        info.balance += _amount;
+        _bondToken.safeTransferFrom(msg.sender, address(this), _amount);
+        emit Deposited(msg.sender, _amount);
+    }
+
+    /// @notice Withdraws bond tokens after withdrawal delay has passed.
+    function withdraw() external nonReentrant {
```
ggonzalez94, line 179: nit: maybe we should allow to withdraw excess balance? i.e. anything above `_ejectionThreshold` or `_requiredBond`(which I think we should convert in a single paramenter, and the prover can decide if they deposit extra bonds, such that one slashing event does not make them ejectable)

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,377 @@
...
+        emit BidPlaced(msg.sender, _feeInGwei, oldProver);
+    }
+
+    /// @notice Requests to exit as the current prover.
+    function requestExit() external nonReentrant {
+        require(proverState.currentProver == msg.sender, NotCurrentProver());
+
+        uint48 withdrawableAt = _withdrawableAt();
+        _bonds[msg.sender].withdrawableAt = withdrawableAt;
+        _vacateProver();
```
ggonzalez94, line 234: I wonder if the impact of immediately setting the prover to zero is not too large. What I mean is:
- a proposer expects to have a prover from the auction be the one that proofs his proposal
- if they suddenly exit now they have to deposit a bond and self prove or find a prover off-chain

This likely results in an incentive system where proposers will have to be fully bonded in the prover auction contract, and have a backup prover ready in case the prover exists when it is their turn or the pool becomes vacant for whatever reason

packages/protocol/contracts/layer1/core/impl/ProverAuction.sol
```diff
@@ -0,0 +1,377 @@
...
+            uint128 actualReward;
+            if (_recipient != address(0)) {
+                actualReward = uint128(uint256(actualSlash) * _rewardBps / 10_000);
+            }
+
+            uint128 newBalance = bond.balance - actualSlash;
+            totalSlashedAmount += actualSlash - actualReward;
+
+            if (actualReward > 0) {
+                _bondToken.safeTransfer(_recipient, actualReward);
```
ggonzalez94, line 256: I guess here it is safe to revert in case that the transfer fails because this is the actual prover address right?

ggonzalez94: if so, maybe we add a short comment explaining it?

ggonzalez94 (commented):
I like this simpler design a lot more, since it is easier to reason about. But there are a few open questions that I don’t know the answer to.

- One benefit of having a prover auction is the proposer can rely on the market selecting a prover for them, and they don’t need to run backup infra or  find a prover themselves. But if the fee is too high, or there’s no prover you are forced to self prove.
      - How does this impact proposers. Where can they expect to get the backup proving?
- Concerns about immediate exit/ejection(I put them in the comments)
- Should we consider having fees paid in Taiko as well? how much overhead will it really add. We could make this just a balance change in some bond contract instead of actually transferring the funds
- Does the incentive and bidding strategy incentivize aggregation?

Let's discuss in our sync. In terms of when to ship this, I still believe it is better to keep the BondManager on L1 for shasta, and ship this in the next hard fork if you agree. This way we remain in time for the current audit, and gives us enough time to think properly about all the incentives and edge cases for the auction

PR #21105, opened 2026-01-05
feat(protocol): remove designatedProver in Transition

YoGhurt111:
@ggonzalez94 @davidtaikocha It seems that we can remove the `designatedProver` in `Transition` with the new approach?

I opened this PR just for suggestions, so I haven't remove the related code in Codec and test.


davidtaikocha:
Yep i also suggested this in the Slack thread earlier today.

ggonzalez94:
@YoGhurt111 yes, this is a good idea. Sorry I didn't see your PR before. I pushed one with the full set of changes [here](https://github.com/taikoxyz/taiko-mono/pull/21107). Please take a look and feel free to choose the one you prefer

PR #21108, opened 2026-01-05
feat(protocol): modularize the inbox by adding hooks and native eth bond support

AnshuJalan:
This PR refactors the inbox to add some hooks that we found useful to make the Inbox modular. Using these hooks, developers of rollup protocols can extend the inbox and add custom features without having to modify the base contract.

Examples of such custom features (from Surge repo) are linked below:
- [Finality Gadget](https://github.com/NethermindEth/surge-taiko-mono/blob/v3.0.0-deployment-scripts/packages/protocol/contracts/layer1/surge/features/FinalityGadgetInbox.sol): An extension to detect proof conflicts are allow forced upgrades of verifiers.
- [Chain Rollback](https://github.com/NethermindEth/surge-taiko-mono/blob/v3.0.0-deployment-scripts/packages/protocol/contracts/layer1/surge/features/RollbackInbox.sol): An extension that allows rolling back the chain incase finalizations have come to a halt due to a prover bug.

Developers can also "chain" multiple features in a single inbox like this: https://github.com/NethermindEth/surge-taiko-mono/blob/v3.0.0-deployment-scripts/packages/protocol/contracts/layer1/surge/deployments/internal-devnet/SurgeInbox.sol

Additionally, this PR also adds the support for using native eth as bonds, enabled by allowing `_bondToken` to be a `address(0)`


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -785,6 +797,54 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+        result_ = _consumeForcedInclusions(msg.sender, _input.numForcedInclusions);
+        result_.sources[result_.sources.length - 1] =
+            DerivationSource(false, LibBlobs.validateBlobReference(_input.blobReference));
+    }
+
+    /// @dev Handles proof verification by delegating to the proof verifier contract.
+    /// @param _proposalAge Seconds passed since the proposal was submitted
+    /// @param _commitment The commitment containing the batch transitions to verify.
+    /// @param _proof The encoded proof data to verify against the commitment.
+    function _handleProofVerification(
```
ggonzalez94, line 843: I wonder if this really needs to be virtual? it seems to me that you always want to call the verifier. If the verifier interface is not suitable for all use cases, perhaps we should improve `IVerifier` instead of making this overridable?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -785,6 +797,54 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+
+    /// @dev A post proving hook to execute extra logic after proving a proposal
+    function _afterProve() internal virtual { }
+
+    /// @dev Builds the consumption result that contains the derivation sources
+    /// @dev In certain extended inboxes, forced inclusion may be disabled when the chain
+    /// enters a feature specific "safety mode"
+    /// @param _input The propose input data
+    /// @return result_ The consumption result with sources array and permissionless flag
+    function _buildConsumptionResult(ProposeInput memory _input)
```
ggonzalez94: I'm a bit affraid of making this virtual, since it includes not only the forced inclusion logic, but also the logic of building the main derivation source as well. I also feel it is not a clean abstraction(i.e. before and after hooks are very clean, and they express meaning to the user), while this one I feel doesn't

I don't have yet a better suggestion, but I feel we can probably think about this a bit more and try to come up with a better approach

ggonzalez94 (commented):
I like this. I think the idea is clean and pretty helpful.

I left a couple comments on abstractions I believe we can improve, and will submit a small PR for an idea of how to improve one of them.

If we go this route, I think we should consider what other things make sense to make possible to override safely.

- the whole activation flow. New chains should not have a need for this, and even existing chains will likely want to deal with it differently.
- Event emission? other chains might require a different event structure

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: Here's a proposal: https://github.com/taikoxyz/taiko-mono/pull/21128

packages/protocol/contracts/layer1/core/impl/Inbox.sol
AnshuJalan: This was made vritual because the verifier we wanted to use had a completely different set of parameter types. A clean way to manage this would be to let `IVerifier`'s verify function take in `bytes` - but that would mean another layer of encoding and decoding, which felt a bit unnecessary

packages/protocol/contracts/layer1/core/impl/Inbox.sol
AnshuJalan: Hmm, I agree! Liking your suggested approach

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -785,6 +801,58 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+    // Hooks
+    // ---------------------------------------------------------------
+
+    /// @dev A pre proposal hook to execute extra logic before making a proposal
+    function _beforePropose() internal virtual { }
+
+    /// @dev A post proposal hook to execute extra logic after making a proposal
+    function _afterPropose() internal virtual { }
+
+    /// @dev A pre proving hook to execute extra logic before proving a proposal
```
ggonzalez94: nit: 
```suggestion
    /// @dev A pre proving hook to execute extra logic before proving a range of proposals
```

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -785,6 +801,58 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+    function _beforeProve() internal virtual { }
+
+    /// @dev A post proving hook to execute extra logic after proving a proposal
```
ggonzalez94: nit:

```suggestion
    /// @dev A post proving hook to execute extra logic after proving a range of proposal
```

packages/protocol/contracts/layer1/core/libs/LibBonds.sol
```diff
@@ -32,6 +35,7 @@ library LibBonds {
 
     /// @dev Deposits bond tokens in gwei units and credits the recipient.
     /// If `_cancelWithdrawal` is true, the pending withdrawal request is cleared.
+    /// When _bondToken is address(0), native ETH is used and msg.value must equal the deposit amount.
```
ggonzalez94, line 38: nit: 
```suggestion
    /// When _bondToken is address(0), native ETH is used and msg.value must equal the deposit amount(after conversion to wei).
```

packages/protocol/contracts/layer1/core/libs/LibInboxSetup.sol
```diff
@@ -26,7 +26,7 @@ library LibInboxSetup {
         require(_config.proofVerifier != address(0), ProofVerifierZero());
         require(_config.proposerChecker != address(0), ProposerCheckerZero());
         require(_config.signalService != address(0), SignalServiceZero());
-        require(_config.bondToken != address(0), BondTokenZero());
+        // Note: bondToken can be address(0) for native ETH bonds
```
ggonzalez94: A comment here seems kinda awkward? I think we can remove it or in any case move it to the natspec

packages/protocol/contracts/layer1/core/impl/Inbox.sol
ggonzalez94: Understood. If we can get a single verifier interface, and avoid making this virtual I think we should. Here's a proposed interface that I think achieves some consistency + the flexibility needed, let me know what you think:

```sol
    /// @notice Verifies a validity proof for a state transition
    /// @dev This function must revert if the proof is invalid
    /// @param _commitmentHash Hash of the last proposal hash and commitment data
    /// @param _proof The proof data
    /// @param _data Extra data specific to each verifier
    function verifyProof(
        bytes32 _commitmentHash,
        bytes calldata _proof,
        bytes calldata _data
    )
        external
        view;
```

ggonzalez94 (commented):
Looks great! Left a few minor nits, and a suggestion for how to deal with the verifier without needing to make that function virtual

packages/protocol/contracts/layer1/core/libs/LibBonds.sol
AnshuJalan: I actually changed it to be wei everywhere in the latest commit, gwei felt odd

packages/protocol/contracts/layer1/core/impl/Inbox.sol
AnshuJalan: I actually liked this, but midway into making the change, I realised that if we do not have a separate virtual function, the dev who is extending the inbox needs to modify the prove function to accommodate the new value of `_data`.
Modifications to `propose` and `prove` are something we intend to avoid altogether

ggonzalez94: you are right. Ok, the current approach then seems like an acceptable trade-off until we think about a better way 😄

AnshuJalan:
Marked this as `do.not.merge` for now, since there are a couple more developments on our end regarding modularization. I'll get back to you on this.

CC: @ggonzalez94

ggonzalez94:
> Marked this as `do.not.merge` for now, since there are a couple more developments on our end regarding modularization. I'll get back to you on this.
> 
> CC: @ggonzalez94

Sounds good. If they are strictly additions of top I would still try to merge this so that the PR does not become too big. Also let's not forget to go back using gwei for the bond, since wei is too small and it would break our shasta deployment

AnshuJalan:
Yeah, so I am reverting to wei and not making any new changes in this PR. Will raise a separate one after a discussion

AnshuJalan:
@ggonzalez94 Need your re-approval

smartprogrammer93 (approved):

PR #21117, opened 2026-01-07
feat(taiko-client-rs): extend `PreconfSignerResolver` trait

davidtaikocha:
##  Summary
   Extend the lookahead resolver API to return both the signer and the canonical submissionWindowEnd, matching the descirption in the design doc:

> Each `LookaheadSlot` carries an `submissionWindowEnd`, the deadline by which the elected preconfer must submit their L2 block(s) for that slot.
If the preconfer is also the L1 proposer, this timestamp corresponds to their L1 proposing slot. If the preconfer is a fallback preconfer, it marks the end of the allocated epoch (or other agreed allocation of fallback preconfer).
Its primary role is slashing attribution ****(as we see in later slashing sections): it serves as the canonical “window ID” when comparing preconfed vs. submitted blocks, helping distinguish whether a mismatch is due to the current preconfer or previous preconfer.

##   Changes
   - Add PreconfSlotInfo (signer + submissionWindowEnd)
   - Extend PreconfSignerResolver with slot_info_for_timestamp
   - Implement LookaheadResolver::slot_info_for_timestamp:
     ·  opted-in slot: submissionWindowEnd = slot.timestamp
     ·  fallback: submissionWindowEnd = next_epoch_start - SECONDS_IN_SLOT
     ·  blacklisted slot: signer falls back, submissionWindowEnd stays slot.timestamp
   - Add unit tests for the above cases

### DIFF
--- a/packages/taiko-client-rs/crates/protocol/src/preconfirmation/lookahead/mod.rs
+++ b/packages/taiko-client-rs/crates/protocol/src/preconfirmation/lookahead/mod.rs
@@ -32,10 +32,22 @@ pub use resolver::{LookaheadBroadcast, LookaheadResolver};
 /// Convenience alias for the default provider stack used by lookahead clients/resolvers.
 pub type LookaheadResolverWithDefaultProvider = LookaheadResolver;
 
+/// Resolved signer plus canonical submission window end for a preconfirmation slot.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct PreconfSlotInfo {
+    /// Address allowed to sign the commitment for the slot.
+    pub signer: Address,
+    /// Canonical end of the submission window for the slot.
+    pub submission_window_end: U256,
+}
+
 /// Resolves the expected signer for a preconfirmation commitment at a given L2 block timestamp,
 /// matching the documented lookahead resolver behavior above.
 #[async_trait]
 pub trait PreconfSignerResolver {
     /// Return the address allowed to sign the commitment covering `l2_block_timestamp`.
     async fn signer_for_timestamp(&self, l2_block_timestamp: U256) -> Result<Address>;
+
+    /// Return the signer plus canonical submission window end for `l2_block_timestamp`.
+    async fn slot_info_for_timestamp(&self, l2_block_timestamp: U256) -> Result<PreconfSlotInfo>;
 }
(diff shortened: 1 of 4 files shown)


packages/taiko-client-rs/crates/protocol/src/preconfirmation/lookahead/resolver/core.rs
```diff
@@ -572,6 +593,11 @@ impl PreconfSignerResolver for LookaheadResolver {
     async fn signer_for_timestamp(&self, l2_block_timestamp: U256) -> Result<Address> {
         self.committer_for_timestamp(l2_block_timestamp).await
     }
+
+    /// Return the signer plus canonical submission window end for `l2_block_timestamp`.
+    async fn slot_info_for_timestamp(&self, l2_block_timestamp: U256) -> Result<PreconfSlotInfo> {
+        self.slot_info_for_timestamp(l2_block_timestamp).await
```
MatusKysel:
```suggestion
        LookaheadResolver::slot_info_for_timestamp(self, l2_block_timestamp).await
```
explicit call is easier to read IMO

packages/taiko-client-rs/crates/protocol/src/preconfirmation/lookahead/mod.rs
```diff
@@ -32,10 +32,22 @@ pub use resolver::{LookaheadBroadcast, LookaheadResolver};
...
+}
+
 /// Resolves the expected signer for a preconfirmation commitment at a given L2 block timestamp,
 /// matching the documented lookahead resolver behavior above.
 #[async_trait]
 pub trait PreconfSignerResolver {
     /// Return the address allowed to sign the commitment covering `l2_block_timestamp`.
     async fn signer_for_timestamp(&self, l2_block_timestamp: U256) -> Result<Address>;
+
+    /// Return the signer plus canonical submission window end for `l2_block_timestamp`.
```
MatusKysel, line 51: isn't it L1 timestamp ?

MatusKysel (approved):

YoGhurt111 (approved):

packages/taiko-client-rs/crates/protocol/src/preconfirmation/lookahead/mod.rs
davidtaikocha: I think it means the preconfed block's timestamp here, so should be L2 block timestamp?

PR #21124, opened 2026-01-07
fix(relayer): honor backoff config for queue subscriptions

kurahin:
Previously the relayer processor and watchdog parsed the common backoff.retryInterval and backoff.maxRetries flags into their configs but still used hardcoded retry intervals (1s) when subscribing to the RabbitMQ queue. This made the CLI backoff settings effectively dead code and inconsistent with the indexer, which already uses the configured backoff policy.

This change wires Processor.backOffRetryInterval / backOffMaxRetries and Watchdog.backOffRetryInterval / backOffMaxRetries into the backoff.Retry calls around queue.Subscribe by wrapping a ConstantBackOff with WithMaxRetries and WithContext, mirroring the pattern used in the indexer. The ScanBlocks retry interval remains a fixed 5s to keep behaviour aligned across components.

### DIFF
--- a/packages/relayer/processor/processor.go
+++ b/packages/relayer/processor/processor.go
@@ -483,6 +483,14 @@ func (p *Processor) Start() error {
 	}
 
 	go func() {
+		bo := backoff.WithContext(
+			backoff.WithMaxRetries(
+				backoff.NewConstantBackOff(p.backOffRetryInterval),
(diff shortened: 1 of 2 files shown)


MatusKysel:
@claude pls review

MatusKysel (approved):

davidtaikocha (approved):

PR #21128, opened 2026-01-08
feat(protocol): use a different hook for forced inclusions

ggonzalez94:
Rough idea for a better abstraction for addressing [this comment](https://github.com/taikoxyz/taiko-mono/pull/21108#discussion_r2673376746). Please feel free to improve the code.

The general idea is that we should be very conscious about what we allow to override, how can we make it easier to do so in a safe way, and that the abstractions are clean and meaningful(similar to how the before and after hooks are very clean)

---

> [!NOTE]
> Modernizes forced inclusion handling and proposal assembly in `Inbox.sol`.
> 
> - Introduces `_buildForcedInclusions(_input, _sources)` to generate derivation sources; removes `ConsumptionResult` and `_buildConsumptionResult`
> - Changes `_consumeForcedInclusions` to populate a caller-provided `sources` array and return `(processed, allowsPermissionless)`; proposal now appends the normal source and shrinks the array via `mstore`
> - Keeps proposer permission checks but gates them on the new `allowsPermissionless` flag
> - Updates snapshot gas figures with small reductions in `shasta-propose.json`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 75187a01f3fc9a449c59d4f49310409fc2814817. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


AnshuJalan (approved):
I like the approach! The naming scheme is cleaner too

PR #21131, opened 2026-01-09
chore(bridge-ui): remove empty files

massarinoaa:
Found 3 empty files that nothing references, so deleted them.


KorbinianK:
cross domain, bridge file will be "fixed" in next update. Non critical

PR #21141, opened 2026-01-12
fix: correct type reference in InitFromCli comment

gap-editor:
Fixed incorrect type reference in comment for `InitFromCli` method.

The comment referenced "Indexer" instead of "BalanceMonitor".


davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/blob/main/CONTRIBUTING.md

PR #21150, opened 2026-01-13
refactor(balance-monitor): eliminate duplicate decimals divisor calculation

anim001k:
Removes duplicate calculation of `decimalsDivisor` in `checkErc20Balance` by computing it once and reusing it for both token balance and bond balance conversions.


davidtaikocha:
Can't see much valuable improvements here.

PR #21152, opened 2026-01-13
fix(taiko-client): remove duplicate AnchorTxValidator construction

sashaodessa:
The function previously had two separate `return &AnchorTxValidator{...}` statements in different code paths (Shasta vs Pacaya clients), creating a maintenance risk. If the struct definition changes in the future, both return statements would need to be updated separately, which could lead to inconsistencies.


MatusKysel (approved):

davidtaikocha:
Looks like there is not much improvement, and the previous approach looks clearer to me, will close it, thanks for the PR.

PR #21153, opened 2026-01-13
chore(taiko-client-rs, taiko-client): address ci timeout

RogerLamTd:


MatusKysel (approved):

packages/taiko-client-rs/tests/deploy.sh
```diff
@@ -12,7 +12,6 @@ cd "${PROTOCOL_DIR}" &&
     --fork-url "$L1_HTTP" \
     --broadcast \
     --ffi \
-    -vvvvv \
```
davidtaikocha, line 15: Let's also change the deployment script in `taiko-client` as well.

packages/taiko-client/internal/version/version_test.go
```diff
@@ -0,0 +1,42 @@
+package version
```
davidtaikocha: Do we really need these tests in this PR?

RogerLamTd: removed

davidtaikocha (approved):

PR #21154, opened 2026-01-13
chore(relayer): fix correct order of dest/src

MatusKysel:
…er of dest/src

### DIFF
--- a/packages/relayer/processor/process_message.go
+++ b/packages/relayer/processor/process_message.go
@@ -337,7 +337,7 @@ func (p *Processor) generateEncodedSignalProof(ctx context.Context,
 	// we can grab the latestBlockID, create a singular "hop" of srcChain => destChain,
 	// and generate a proof.
 	if len(p.hops) == 0 {
-		latestBlockID, err := p.latestSyncedBlockID(ctx, p.srcChainId.Uint64(), p.destChainId.Uint64())
+		latestBlockID, err := p.latestSyncedBlockID(ctx, p.destChainId.Uint64(), p.srcChainId.Uint64())
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


packages/relayer/processor/process_message.go
```diff
@@ -337,7 +337,14 @@ func (p *Processor) generateEncodedSignalProof(ctx context.Context,
...
 	if len(p.hops) == 0 {
-		latestBlockID, err := p.latestSyncedBlockID(ctx, p.srcChainId.Uint64(), p.destChainId.Uint64())
+		chainID, syncedChainID := p.srcChainId.Uint64(), p.destChainId.Uint64()
+		if p.shastaOldForkAddress == relayer.ZeroAddress {
+			// Without the fork router (v4 not implemented), rely on the dest chain's synced data.
+			chainID = p.destChainId.Uint64()
+			syncedChainID = p.srcChainId.Uint64()
+		}
+
+		latestBlockID, err := p.latestSyncedBlockID(ctx, chainID, syncedChainID)
```
davidtaikocha: If we switch those two params, the `latestBlockID` will be a block number on a different layer? If so, then seems the changes are not that correct?

davidtaikocha: If we don't have a fork router or there is no old signal service, will `p.eventRepo.LatestChainDataSyncedEvent` only return `0` in `latestSyncedBlockID`? If so, seems we don't need to change anything since the `p.eventRepo.LatestCheckpointSyncedEvent` will return the correct number?

MatusKysel: image
it will not because order is wrong, this is how it was before v4 implementation

davidtaikocha (approved):

YoGhurt111 (approved):

PR #21155, opened 2026-01-13
fix(eventindexer): use Scan for unique proposers/provers queries

Forostovec:
Replace FirstOrInit with Scan in FindUniqueProposers and FindUniqueProvers so that GORM returns all grouped rows instead of implicitly adding LIMIT 1. This aligns the repository methods with their API contract (return all unique addresses with counts), ensures the HTTP endpoints expose the full set of unique proposers/provers, and keeps error wrapping consistent with the actual query method.


davidtaikocha:
@claude plz review.

MatusKysel (approved):

davidtaikocha (approved):

PR #21158, opened 2026-01-14
fix(relayer): keep queue message body format when requeuing

phrwlk:
Fixed the processor requeue path so that messages keep using the QueueMessageSentBody JSON format expected by processMessage. Previously, eventLoop marshalled the whole queue.Message struct and republished it, which broke JSON decoding on the next attempt and effectively caused such messages to be dropped into the DLX instead of being retried.

Now, when processMessage returns shouldRequeue == true, the processor only calls Nack(requeue=true) and lets the broker requeue the original delivery without touching its body.

### DIFF
--- a/packages/relayer/processor/processor.go
+++ b/packages/relayer/processor/processor.go
@@ -3,7 +3,6 @@ package processor
 import (
 	"context"
 	"crypto/ecdsa"
-	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


MatusKysel:
@claude plz review

MatusKysel (approved):

davidtaikocha (approved):

PR #21159, opened 2026-01-14
ci(taiko-client, taiko-client-rs): improve ci timeout

RogerLamTd:
previously failing on forge script, likely due to invisible failures where txs don't get mined; inferred from failed runs not reaching `ONCHAIN EXECUTION COMPLETE & SUCCESSFUL.` log. 

also extended timeout as we have encountered cases where compilation is the gating factor, ~15 should be enough for now

### DIFF
--- a/packages/taiko-client-rs/tests/docker/docker-compose.test.yaml
+++ b/packages/taiko-client-rs/tests/docker/docker-compose.test.yaml
@@ -3,7 +3,7 @@ services:
     container_name: l1_node
     image: ghcr.io/foundry-rs/foundry:stable
     restart: unless-stopped
-    pull_policy: if_not_present
+    pull_policy: always
     ports:
       - "18545:8545"
     entrypoint:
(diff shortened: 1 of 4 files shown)


packages/taiko-client-rs/tests/docker/docker-compose.test.yaml
```diff
@@ -3,19 +3,21 @@ services:
     container_name: l1_node
     image: ghcr.io/foundry-rs/foundry:stable
     restart: unless-stopped
-    pull_policy: always
+    pull_policy: if_not_present
```
davidtaikocha: We still need `always` here for `packages/taiko-client-rs/tests/docker/docker-compose.test.yaml`

RogerLamTd: reverted

davidtaikocha (approved):

MatusKysel (approved):

PR #21167, opened 2026-01-15
feat(taiko-client): dont use ws subscription and switch to http polling 

MatusKysel:


MatusKysel:
@claude plz review

MatusKysel:
@codex review

MatusKysel:
@claude plz review, @codex review

packages/taiko-client/cmd/flags/common.go
```diff
@@ -21,12 +21,12 @@ var (
 
 // Required flags used by all client software.
 var (
-	L1WSEndpoint = &cli.StringFlag{
-		Name:     "l1.ws",
-		Usage:    "Websocket RPC endpoint of a L1 ethereum node",
+	L1HTTPEndpoint = &cli.StringFlag{
```
YoGhurt111, line 26: Can we add HTTP support instead of replacing WebSocket?

MatusKysel: maybe it should be just L1 endpoint ? and it could be universal ? WDYT?

davidtaikocha: imo it's better to keep this change backward compatible, so let's keep the `l1.ws` flag here, wdyt?

MatusKysel: so do you want 2 flags? or we will keep old one for both ws/http?

davidtaikocha: Keep `l1.ws` (if we want to make it backward compatible) for both ws/http seems not direct... so let's create another `l1.http`, and only require one of them?

MatusKysel: fixed a42e34c315cfbd2750c28f0fee533935354e0077

packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -605,27 +605,17 @@ where
         let derivation = Arc::new(derivation_pipeline);
         let router = self.build_router(derivation.clone());
 
-        let mut scanner = self
-            .cfg
-            .client
-            .l1_provider_source
-            .to_event_scanner_from_tag(start_tag)
-            .await
-            .map_err(|err| SyncError::EventScannerInit(err.to_string()))?;
```
davidtaikocha, line 614: Seems this feature is not really friendly with the `event-scanner` our `taiko-client-rs` is using, let me think about it...

davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/pull/21450

PR #21178, opened 2026-01-16
fix(repo): shell injection in deploy workflow and tighten websocket origins

RinZ27:
Refactored the Vercel deployment workflow to use intermediate environment variables for user-supplied inputs. Direct interpolation into `run:` steps is a known vector for command injection.

Modified the WebSocket upgrader in the preconfirmation block server to enforce origin validation. The previous permissive policy allowed connections from any origin, introducing unnecessary CSRF risk to the driver component.

Also added directory name sanitization in the bridged-token-scanner tool to prevent potential path traversal when handling network names with special characters.

Impact:
- Mitigates runner takeover risks in GitHub Actions.
- Protects the preconfirmation server from cross-site WebSocket hijacking.
- Ensures local tool scripts operate within their designated data directories.

### DIFF
--- a/.github/workflows/repo--vercel-deploy.yml
+++ b/.github/workflows/repo--vercel-deploy.yml
@@ -43,10 +43,19 @@ jobs:
         run: pnpm add --global vercel@latest
 
       - name: Pull Vercel Environment Information
-        run: vercel pull --yes --environment=${{ inputs.environment }} --token=${{ secrets.vercel_token }}
+        env:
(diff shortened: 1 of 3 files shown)


RinZ27:
Addressed the security findings regarding shell injection and WebSocket origin checks. Also resolved any potential conflicts. Ready for review.

davidtaikocha:
Closed due to being outdated.

PR #21179, opened 2026-01-16
fix(bridge-ui): shadowed address validation state in AddCustomERC20

sashaodessa:
Fix a real UI logic bug caused by variable shadowing in `AddCustomERC20.svelte`, where the component-level `isValidEthereumAddress` flag was never updated.


davidtaikocha:
Closed due to being outdated.

PR #21190, opened 2026-01-19
feat(protocol): set inbox parameters for mainnet and devnet.

ggonzalez94:
This PR is the result of our first session going trough the parameters for the inbox for mainnet and devnet. There are also some that we collectively decided were not necessary anymore given proposal cadence and future zk gas.

This PR also updates `numForcedInclusions` to uint16 to prevent an edge case where if more than 255 inclusions are due proposing becomes stucked.

### Removed Parameters
- `minCheckpointDelay`: was originally introduced when we were finalizing in the prove function. Now that we are doing it when finalizing, and planning to aggregate aggressively, we can basically sync the checkpoint on ever prove. There's no need to have this parameter
- `minForcedInclusionCount`: originally developed as a rate limit to avoid overwhelming, but given we plan to introduce zk gas soon, this is not a concern. Currently there is also no attack vector on forced inclusions that is also not present by just submitting txs to the L2 mempool 

---

> [!NOTE]
> Introduces finalized mainnet/devnet Inbox parameters and simplifies proving/checkpointing.
> 
> - Removes `minForcedInclusionCount` and `minCheckpointDelay`; deletes `forceCheckpointSync` from `ProveInput`, codec, and tests
> - `prove` now always saves a checkpoint (updates `lastCheckpointTimestamp`) and drops `checkpointSynced` from `Proved` event
> - Tightens forced inclusion handling: after processing requested items, reverts if any older due inclusion remains unprocessed
> - Bond check on propose only when `minBond > 0`
> - Sets network configs:
>   - Mainnet: ring buffer `21_600`, `minBond/livenessBond=0`, `withdrawalDelay=1w`, `provingWindow=4h`, `permissionlessProvingDelay=5d`, `maxProofSubmissionDelay=3m`, `basefeeSharingPctg=75`, `forcedInclusionDelay=576s`, `forcedInclusionFeeInGwei=1_000_000`, `forcedInclusionFeeDoubleThreshold=50`, `permissionlessInclusionMultiplier=160`
>   - Devnet: similar values with ring buffer `100`
> - Deployment script updates constructors (drops bond/delay params) and removes related env vars
> - Updates gas snapshots and refactors tests to new API/behavior, adding checkpoint validity assertions
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ac577677b280fd9ab1ce2846160142495b481c78. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/contracts/layer1/mainnet/MainnetInbox.sol
```diff
@@ -40,26 +34,25 @@ contract MainnetInbox is Inbox {
...
                 proposerChecker: _proposerChecker,
                 proverWhitelist: _proverWhitelist,
                 signalService: LibL1Addrs.SIGNAL_SERVICE,
                 bondToken: _bondToken,
-                minBond: _minBond,
-                livenessBond: _livenessBond,
-                withdrawalDelay: _withdrawalDelay,
-                provingWindow: 4 hours,
-                permissionlessProvingDelay: 72 hours,
+                minBond: 0,
```
YoGhurt111: `0` for mainnet?

ggonzalez94: yes, during the time we have the prover whitelist we don't need a bond(since the proposers can't get slashed since we are the only ones proving anyways). Once we make proving permisionless again after we test and audit the guest program, then we need to set a min bond and an appropiate liveness bond

davidtaikocha (approved):

YoGhurt111 (approved):

PR #21191, opened 2026-01-19
refactor(taiko-client): deduplicate wei unit conversions

majtte:
This PR removes duplicated wei/unit conversion logic in `packages/taiko-client/pkg/utils/utils.go` by extracting small internal helpers, while keeping the exported API and behavior unchanged.


davidtaikocha:
Don't see too many valuable improvements here, thanks for the PR.

PR #21214, opened 2026-01-26
feat(protocol): prototype l2 fee vault using an EIP-1559 like mechanism

ggonzalez94:
This is the first iteration for the L2 fee [design](https://www.notion.so/nethermind/L2-Fee-Session-2ee360fc38d080c19413d4079d61cc3a?source=copy_link) we discussed. After a short description I'll include the TODO's that I still plan to work on:

### Properties
Self-balancing L1 cost reimbursement using EIP-1559 proportional controller, with accurate gas tracking and partial risk sharing for unprofitable proposals.                                  

### How It Works
I recommend starting with the `L2FeeVault`, then the `Anchor` and finally the `Inbox` when reviewing.                                                         
                                                                       
  1. Revenue Collection                                                
  - L2 basefees flow into L2FeeVault as ETH revenue (assets)           
                                                                       
  2. Liability Tracking                                                
  - When proposals are imported (via Anchor), vault calculates L1 cost  from accurate gas tracking                                           
  - Profitable proposals (L2 revenue ≥ L1 cost): 100% reimbursement    
  - Loss-making proposals (L2 revenue < L1 cost): Partial reimbursement (lossReimbursementBps)                                              
  - Reimbursements are tracked as liabilities until claimed            
                                                                       
  3. Fee Adjustment (EIP-1559 Style)                                   
  - Effective balance = ETH balance - total liabilities                
  - Error ratio = (target - effective) / target                        
  - Adjustment factor = 1 + errorRatio / 8 (max ±12.5% per update)     
  - New fee = clamp(oldFee × adjustmentFactor, minFee, maxFee)         
                                                                       
  Examples:                                                            
  - 100% deficit (broke) → fee increases 12.5%                         
  - At target → no change                                              
  - 100% surplus (2× target) → fee decreases 12.5%                     
                                                                       
  4. Basefee Integration                                               
  - L2 clients read feePerGasWei from vaul                  
  - Creates feedback loop: deficit → higher fees → more revenue →  convergence                                                          
                                                                       
### TODOs

- [ ] Make the inbox more gas efficient - maybe we don't need to emit all this information on the event. Check with @davidtaikocha 
- [ ] Incorporate revenue into the L2 vault fees. The current implementation assumes base fees should not generate a profit. In reality we do want some base free profit, which can flow to the protocol. Design a mechanism that takes that goal into account
- [ ] Review prover constrains with @smtmfft


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -56,6 +56,37 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
 
     event InboxActivated(bytes32 lastPacayaBlockHash);
 
+    // ---------------------------------------------------------------
+    // Gas Tracking Constants
+    // ---------------------------------------------------------------
+
+    /// @dev Base transaction cost (EIP-2028)
+    uint256 private constant _TX_BASE_GAS = 21_000;
```
dantaik, line 72: move these new constants to a new library?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -138,7 +166,7 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
     /// @dev Storage for bond balances.
     LibBonds.Storage private _bondStorage;
 
-    uint256[43] private __gap;
+    uint256[44] private __gap;
```
dantaik: was reserving 43 slots a bug, if so ,we should fix it on the 3.0.0 branch

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -216,14 +245,22 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
             uint48 lastFinalizedProposalId = _coreState.lastFinalizedProposalId;
             require(nextProposalId > 0, ActivationRequired());
 
-            Proposal memory proposal = _buildProposal(
+            proposal = _buildProposal(
```
dantaik: moving `Proposal memory` back here should work

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -216,14 +245,22 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
                 input, _lookahead, nextProposalId, lastProposalBlockId, lastFinalizedProposalId
             );
 
+            // Capture L1 cost inputs
+            proposal.cost.numBlobs = uint32(input.blobReference.numBlobs);
```
dantaik, line 313: will it be easier to just do 
```
proposal.cost = ProposalCost({...})
```

packages/protocol/contracts/layer1/core/iface/IInbox.sol
```diff
@@ -59,6 +57,18 @@ interface IInbox {
         LibBlobs.BlobSlice blobSlice;
     }
 
+    /// @notice Canonical L1 cost inputs for a proposal.
+    struct ProposalCost {
+        /// @notice Gas used by this proposal transaction on L1.
+        uint64 gasUsed;
+        /// @notice Number of blobs attached to this proposal transaction on L1.
+        uint32 numBlobs;
```
dantaik, line 62: maybe uint8 and uint32 is good enough for numBlobs and gasUsed respectiely.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -714,6 +750,132 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+
+    /// @dev Calculates intrinsic gas cost for this transaction.
+    /// @dev Intrinsic gas = base cost + calldata cost (4 gas per zero byte, 16 per non-zero byte).
+    /// @return gas_ Total intrinsic gas cost.
+    function _calcIntrinsicGas() private pure returns (uint256 gas_) {
+        unchecked {
+            gas_ = _TX_BASE_GAS;
+            bytes calldata data = msg.data;
+            uint256 len = data.length;
+            for (uint256 i; i < len; ++i) {
```
dantaik, line 836: maybe we simply use `len*_CALLDATA_NONZERO_BYTE_GAS` as an good approximate for simplicity and to avoid ADD ops.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -714,6 +750,132 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+
+    /// @dev Calculates gas cost for storing the proposal hash in the ring buffer.
+    /// @dev Cost = SSTORE (set or reset) + cold access cost.
+    /// @param _proposalId The proposal ID being stored.
+    /// @return gas_ Gas cost for the SSTORE operation.
+    function _calcProposalHashStorageGas(uint48 _proposalId) private view returns (uint256 gas_) {
+        unchecked {
+            // Determine if this is a SET (writing to new slot) or RESET (overwriting existing)
+            // Slots wrap around at ringBufferSize, so proposalId >= ringBufferSize means overwrite
+            bool isSet = uint256(_proposalId) < _ringBufferSize;
```
dantaik, line 850: maybe we don't need to be accurate, simply use _COLD_SLOAD_COST instead?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -714,6 +750,132 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+                // - inclusionProof offset: 32 bytes
+                // - blobHashes offset: 32 bytes
+                dataBytes += 192;
+
+                // inclusionProof bytes array (empty for now, just length)
+                dataBytes += 32;
+
+                // blobHashes bytes32[] array
+                uint256 numHashes = sources[i].blobSlice.blobHashes.length;
+                dataBytes += 32; // Array length
```
dantaik, line 914: not sure if the compiler will help us optimize these ADDs, or should we simply do the math here to sum them up instead?

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -98,7 +105,7 @@ contract Anchor is EssentialContract {
     mapping(uint256 blockNumber => PreconfMetadata metadata) internal _preconfMetadata;
 
     /// @notice Storage gap for upgrade safety.
-    uint256[40] private __gap;
+    uint256[41] private __gap;
```
dantaik: why 41?

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -216,14 +245,22 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+            proposal = _buildProposal(
                 input, _lookahead, nextProposalId, lastProposalBlockId, lastFinalizedProposalId
             );
 
+            // Capture L1 cost inputs
+            proposal.cost.numBlobs = uint32(input.blobReference.numBlobs);
+            proposal.cost.basefee = uint128(block.basefee);
+            proposal.cost.blobBasefee = uint128(block.blobbasefee);
+
+            // Calculate gas used (must be done before state changes and event emission)
```
linoscope, line 317: Why does this have to be `must be done before state changes and event emission`? Wouldn't `gasStart - gasleft()` capture the storage and event cost if we call it after  `_emitProposedEvent`?

packages/protocol/contracts/layer2/core/Anchor.sol
```diff
@@ -256,6 +268,27 @@ contract Anchor is EssentialContract {
...
+
+        uint48 lastProposalId = _lastProposalId;
+
+        // Same proposal - fee data already imported, skip
+        if (_proposalId == lastProposalId) return;
+
+        // New proposal - import fee data
+        require(_proposalId == lastProposalId + 1, InvalidProposalId());
+        feeVault.importProposalFee(_feeData);
+        _lastProposalId = _proposalId;
```
linoscope: nit: Feels kind of weird to set `lastProposalId` inside `_importFeeData`, better to set inside the parent `anchorV4`.

packages/protocol/snapshots/shasta-propose.json
```diff
@@ -1,5 +1,5 @@
 {
-  "propose_after_ring_buffer_wrap": "44615",
-  "propose_forced_inclusion": "58427",
-  "propose_single": "70915"
-}
\ No newline at end of file
+  "propose_after_ring_buffer_wrap": "71063",
```
linoscope: The gas overhead is pretty high. I think most of it is coming from `_calcGasUsed` - Likely we can optimize that further with the tradeoff of being less accurate. E.g., Don't scan through bytes in `msg.data` and make an assumption of the number of zero bytes based on actual proposals (e.g., 20% is zero bytes).

linoscope:
Need to think a little more on whether EIP-1559-style is the good first formula for the fee calculation - There is quite a large design space here.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -216,14 +245,22 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
+            proposal.cost.gasUsed = _calcGasUsed(gasStart, proposal);
+
             _coreState.nextProposalId = nextProposalId + 1;
             _coreState.lastProposalBlockId = uint48(block.number);
-            _setProposalHash(proposal.id, LibHashOptimized.hashProposal(proposal));
 
+            _setProposalHash(proposal.id, LibHashOptimized.hashProposal(proposal));
             _emitProposedEvent(proposal);
```
linoscope, line 324: I am wondering if we even need to emit the `cost` info in the even, considering that driver should read fee data from anchor block in the proposed block, not from the proposal itself. But if cost cut from removing fee data from `proposal` event is minimal, I am also fine with keeping it so we just have one `proposal` struct.

packages/protocol/docs/Derivation.md
```diff
@@ -316,7 +331,41 @@ Note: Fields like `stateRoot`, `transactionsRoot`, `receiptsRoot`, `logsBloom`,
...
+| l2BasefeeRevenue | uint256 | Sum of L2 basefee revenue for blocks in this proposal    |
+
+The validity proof must verify that `feeData` corresponds to canonical L1 Inbox state and L2 execution.
+In particular, it should recompute `hashProposal(proposal)` from the full proposal data (including
+cost fields) and verify it matches the proposal hash stored in the ring buffer for that `proposalId`,
+using the appropriate L1 state root. It must also verify that `l2BasefeeRevenue` is correctly computed
+from the L2 blocks in that proposal (excluding system transactions as specified by the protocol).
+
+Fee data is provided once per proposal (typically on the first L2 block of that proposal). The
+anchor tracks the last imported proposal id and only imports when `proposalId` increments; for other
```
linoscope: > The
anchor tracks the last imported proposal id and only imports when `proposalId` increments

This would mean that the preconfer of window `N` will typically only import the fee data of proposal from preconfer of window `N-2`, as when the preconfer `N` starts preconfing, it's typically in the handover window where `N-1` has not proposed yet. Which may be fine, but in general the faster you import feeData the better as you can react to L1 fee trend shifts more quickly (how much better is an open question).

_Ideally_, we would want to import the feeData as soon as the anchor block detects a new proposal in inbox, even if that L2 block is in the middle of a proposal ID. But the naiive way of doing this would require Merkle proofs at every increment in anchor block ID. Which might be fine depending on the cost of Merkle proofs.

So there is a tradeoff between:
- How quickly we import latest feeData from anchor.
- Merkle proof generation/verificaiton cost.

packages/protocol/contracts/layer1/core/impl/Inbox.sol
linoscope: One nice thing of moving `cost` out of proposal event is that we can move `_calcGasUsed` after emitting the proposal event, hence remove the need to "simulate" the event emission cost in the contract.

linoscope: Oh wait, I guess you still want to emit the full `proposal` (including `cost`) to later reconstruct the proposal from the hash.. Nevermind of these comments.

packages/protocol/docs/Derivation.md
linoscope: > But the naiive way of doing this would require Merkle proofs at every increment in anchor block ID.

Heard from @smtmfft that we don't need Merkle proofs to read proposals from anchor block - Instead we can traverse the proposal through parent proposals. So cost might not be too bad, especially if we limit the anchor block delay to be short (which we are already considering to have better L1->L2 message passing delay).

PR #21215, opened 2026-01-26
chore(taiko-client): double-check `sourcePayload.Default` in `InsertBlocksWithManifest`

davidtaikocha:
Add a double-check of `sourcePayload.Default` in `InsertBlocksWithManifest` for safety, if this is a default manifest, we can just skip the `isKnownInCanonical` check.


YoGhurt111 (approved):

MatusKysel (approved):

davidtaikocha:
After thinking twice, let's still keep ignoring the `payload.Default` in checks, since it should not hurt anything.

PR #21218, opened 2026-01-26
fix(bridge-ui): replace manual escape handler with standard action

sashaodessa:
Replaces manual Escape key event listener implementation with the existing `closeOnEscapeOrOutsideClick` action used throughout the codebase.


davidtaikocha:
Closed due to being outdated.

PR #21226, opened 2026-01-27
feat(taiko-client): add ratelimit http request

dipkakwani:
Currently the taiko-client driver ignores `x-ratelimit-reset` in the HTTP response from Beacon endpoint. We observed in the production that the client makes multiple consecutive requests, while the quicknode server returns 429 status code.

The fix adds a new `http.RoundTripper`. It parses the header `x-ratelimit-reset` when a 429 status code is returned, and applies exponential backoff.


jmadibekov (commented):
LGTM!

davidtaikocha:
@claude plz review.

MatusKysel:
@dipkakwani can you pls check AI suggestions

MatusKysel:
@claude review plz

dipkakwani:
Addressed comments. I don't think critical issues are actually issues: 
> Response Body Not Returned After Max Retries ❌

Body is not returned with http.transport, only error to indicate the operation failed (follows how standard library does)

> Race Condition in Body Reset Logic

Incorrect, we return only when attempt > 0. We've already closed the previous response at the end of the previous iteration.

Added the rest of the suggestions.

MatusKysel (approved):

dipkakwani:
@davidtaikocha  can you please review?

davidtaikocha (approved):

PR #21231, opened 2026-01-27
fix(relayer): move metrics increment after db update to prevent double counting on retry

massarinoaa:
Metrics were being incremented before UpdateStatus, so if the db write failed and the message got requeued, we'd count the same event twice. Moved the log and metrics update after the db operation.


davidtaikocha:
@claude plz review.

davidtaikocha (approved):

MatusKysel (approved):

PR #21249, opened 2026-02-02
feat(ejector): skip reorg ejection during re-anchoring

MatusKysel:


YoGhurt111 (approved):

packages/ejector/src/bindings.rs
```diff
@@ -27,3 +27,14 @@ sol! {
...
+sol! {
+    #[sol(rpc)]
+    interface Anchor {
+        struct BlockState {
+            uint48 anchorBlockNumber;
+            bytes32 ancestorsHash;
+        }
+        function getBlockState() external view returns (BlockState memory);
+    }
+}
```
davidtaikocha, line 40: iiuc, this will only work for Shasta blocks? For current Hoodi and mainnet, we still need to include [Pacaya anchor contract ABI](https://github.com/taikoxyz/taiko-mono/blob/taiko-alethia-protocol-v2.3.1/packages/protocol/contracts/layer2/based/TaikoAnchor.sol#L151)?

MatusKysel: yeah, you are right. Maybe we should wait with this feature ?

MatusKysel: sorry maybe better question is, do we even want it for pacaya ? this feature could be enabled only for shasta networks as it's enabled by flag. WDYT?

davidtaikocha: we can only have it for Shasta, but need to keep the `anchor_address` empty before the fork happens.

davidtaikocha (approved):

PR #21252, opened 2026-02-02
fix(relayer): correct recommended processing fees layers

radik878:
Before this change, GetRecommendedProcessingFees used L2 gas parameters for srcChainID entries and L1 gas parameters for destChainID entries, and passed mismatching layer flags to getCost. This produced incorrect fee estimates for each destination chain and made the API response inconsistent with how the bridge and processor compute costs.

This change swaps the arguments in GetRecommendedProcessingFees so that each DestChainID uses its own gas parameters and appropriate layer flag, and adds a unit test verifying that getCost applies the processingFeeMultiplier only for Layer1 while leaving Layer2 costs unscaled.


MatusKysel:
@codex & @claude  plz review

MatusKysel (approved):

davidtaikocha (approved):

PR #21262, opened 2026-02-03
balance-monitor: dedupe ERC-20 decimals divisor when normalizing balances

anim001k:
This PR removes a small duplication in the ERC-20 balance normalization logic by computing the `10**decimals` divisor once and reusing it for both the token balance and the bond balance paths.


davidtaikocha:
Don't see a lot of improvements here.

PR #21263, opened 2026-02-03
balance-monitor: remove ineffective `os.Kill` from signal handler

anim001k:
Removes `os.Kill` (SIGKILL) from the signal notification list in `SubcommandAction`, as it cannot be caught or handled by `signal.Notify`.

`os.Kill` is a non-catchable signal that cannot be intercepted by user-space handlers. Including it in `signal.Notify` has no effect and may mislead readers into thinking the application handles SIGKILL gracefully.


davidtaikocha:
Don't see a lot of improvements here.

PR #21267, opened 2026-02-04
fix(protocol): avoid corner case on operator removal

ggonzalez94:
This should not happen if we call consolidate on the whitelist contract before we upgrade it. But as we detected on Hoodi, if we don't then we cannot later remove inactive operators from the contract because their record is still there and marked with `activeSince = 0`. This PR changes the inequality comparison to just use a different field.

---

> [!NOTE]
> **Medium Risk**
> Touches L1 whitelist operator-removal logic; although the change is small, it alters the criteria for whether an operator record is considered valid and could affect governance/ejection flows if the new sentinel is incorrect.
> 
> **Overview**
> Fixes a corner case in `PreconfWhitelist._removeOperator` where inactive/stale operator entries with `activeSince == 0` could not be removed.
> 
> The removal precondition now treats an operator record as existing when `sequencerAddress != address(0)` (instead of checking `activeSince`), enabling cleanup of previously-added operators whose `activeSince` was cleared while keeping the “cannot remove last active operator” safety check intact.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dd87fdaf83f4d33fc4b6fed6e299be4d4fd44a7c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


YoGhurt111 (approved):

YoGhurt111:
LGTM

davidtaikocha (approved):

PR #21273, opened 2026-02-06
fix(bridge-ui): shasta fee overrides

RogerLamTd:


RogerLamTd:
@claude please review

ggonzalez94:
we actually realized that we were filtering transactions on the node directly based on an older(higher) base fee. So we're preparing a fix to `taiko-geth` and `taiko-reth`

PR #21275, opened 2026-02-06
fix(bridge-ui): empty gas values explicitely

ggonzalez94:
another approach to solving the hoodi L2->L1 issue

---

> [!NOTE]
> **Medium Risk**
> Changes the transaction submission path for all bridge operations; while behavior is a narrow parameter sanitization, it impacts every write and could affect wallets/networks that rely on explicit fee fields.
> 
> **Overview**
> Updates the bridge UI to submit all on-chain writes (ETH, ERC20, ERC721, ERC1155, and message `process/retry/recall`) via a new `safeWriteContract` wrapper instead of calling `@wagmi/core` `writeContract` directly.
> 
> `safeWriteContract` removes fee-related fields (`maxFeePerGas`, `maxPriorityFeePerGas`, `gasPrice`, `type`) from simulated requests before sending, aiming to avoid stale/underpriced transactions; includes a unit test for the stripping behavior and updates the wagmi mock to export `writeContract`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 089911079dc0673b076ad651e9a90998711b9a91. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/bridge-ui/__mocks__/@wagmi/core.ts
+++ b/packages/bridge-ui/__mocks__/@wagmi/core.ts
@@ -14,6 +14,8 @@ export const getToken = vi.fn();
 
 export const readContract = vi.fn();
 
+export const writeContract = vi.fn();
+
 const mockChains = [
   {
     id: 0,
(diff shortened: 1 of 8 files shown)


ggonzalez94:
we actually realized that we were filtering transactions on the node directly based on an older(higher) base fee. So we're preparing a fix to `taiko-geth` and `taiko-reth`

PR #21297, opened 2026-02-09
fix(taiko-client): make based proposer working with new WL impl

YoGhurt111:


packages/taiko-client/proposer/proposer.go
```diff
@@ -593,76 +599,47 @@ func (p *Proposer) GetParentMetaHash(ctx context.Context) (common.Hash, error) {
...
 // shouldPropose checks whether the proposer should propose at this time.
 func (p *Proposer) shouldPropose(ctx context.Context) (bool, error) {
-	ctxWithTimeout, cancel := rpc.CtxWithTimeoutOrDefault(ctx, rpc.DefaultRpcTimeout)
-	defer cancel()
-
-	preconfRouterAddr, err := p.rpc.GetPreconfRouterPacaya(&bind.CallOpts{Context: ctxWithTimeout})
+	// check the current epoch operator
+	operator, err := p.rpc.GetPreconfWhiteListOperator(&bind.CallOpts{
+		Context: ctx,
+	})
```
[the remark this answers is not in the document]
YoGhurt111: Fixed in https://github.com/taikoxyz/taiko-mono/pull/21297/changes/803f64b2a80b5e262a9d974df0fd932392567afc

YoGhurt111:
@claude About issue 1, it's a optional config, we have this check in txBuilder.

```

if preconfRouterAddress != rpc.ZeroAddress {
		to = &preconfRouterAddress
	}

```

davidtaikocha (approved):

MatusKysel (commented):
do we still need `GetAllActiveOperators` after this change ?

packages/taiko-client/proposer/proposer.go
```diff
@@ -130,6 +129,13 @@ func (p *Proposer) InitFromConfig(
 		}
 	}
 
+	if cfg.TaikoWrapperAddress.Hex() != rpc.ZeroAddress.Hex() {
+		p.preconfRouterAddress, err = p.rpc.GetPreconfRouterPacaya(&bind.CallOpts{Context: p.ctx})
```
MatusKysel, line 133: This now caches `preconfRouterAddress` only at startup, previously we re-resolved it inside `shouldPropose`

packages/taiko-client/proposer/proposer.go
```diff
@@ -593,76 +599,50 @@ func (p *Proposer) GetParentMetaHash(ctx context.Context) (common.Hash, error) {
...
 func (p *Proposer) shouldPropose(ctx context.Context) (bool, error) {
-	ctxWithTimeout, cancel := rpc.CtxWithTimeoutOrDefault(ctx, rpc.DefaultRpcTimeout)
-	defer cancel()
-
-	preconfRouterAddr, err := p.rpc.GetPreconfRouterPacaya(&bind.CallOpts{Context: ctxWithTimeout})
+	if p.rpc.PacayaClients.PreconfWhitelist == nil {
+		return true, nil
+	}
+	// check the current epoch operator
+	operator, err := p.rpc.GetPreconfWhiteListOperator(&bind.CallOpts{
```
MatusKysel, line 606: `GetPreconfWhiteListOperator` returns the sequencer address, but here we compare it with proposer identities (p.proposerAddress / p.ProverSetAddress). If proposer and sequencer keys differ, this check will reject a valid proposer. We should compare against the whitelist proposer address (e.g. `getOperatorForCurrentEpoch`) instead?

packages/taiko-client/proposer/proposer.go
```diff
@@ -593,76 +599,50 @@ func (p *Proposer) GetParentMetaHash(ctx context.Context) (common.Hash, error) {
...
+	if time.Since(p.l2HeadUpdate.updatedAt.UTC()) < p.FallbackTimeout {
+		log.Info(
+			"Fallback timeout not reached, skip proposing",
+			"l2HeadUpdate", p.l2HeadUpdate.updatedAt.UTC(),
+			"blockID", p.l2HeadUpdate.blockID,
+			"now", time.Now().UTC(),
+			"fallbackTimeout", p.FallbackTimeout,
+		)
+		return false, nil
+	} else {
```
MatusKysel, line 636: we don't need this else as we return on line above

packages/taiko-client/proposer/proposer.go
```diff
@@ -593,76 +599,50 @@ func (p *Proposer) GetParentMetaHash(ctx context.Context) (common.Hash, error) {
 
 // shouldPropose checks whether the proposer should propose at this time.
 func (p *Proposer) shouldPropose(ctx context.Context) (bool, error) {
-	ctxWithTimeout, cancel := rpc.CtxWithTimeoutOrDefault(ctx, rpc.DefaultRpcTimeout)
-	defer cancel()
-
-	preconfRouterAddr, err := p.rpc.GetPreconfRouterPacaya(&bind.CallOpts{Context: ctxWithTimeout})
+	if p.rpc.PacayaClients.PreconfWhitelist == nil {
```
MatusKysel, line 602: it seems `PreconfWhitelistAddress` is set only for driver not proposer right ?

MatusKysel (approved):

packages/taiko-client/proposer/proposer.go
YoGhurt111: We got it from PreconfRouter.

YoGhurt111: Yes, the implementation here isn’t ideal — the driver reads it from a flag, while the proposer derives it from the contract relationship.

packages/taiko-client/proposer/proposer.go
YoGhurt111: Since the current proposer implementation is only used by us and contains some workaround logic, we can ensure consistency between the proposer and the sequencer ourselves.

packages/taiko-client/proposer/proposer.go
YoGhurt111: Yes.

PR #21302, opened 2026-02-09
fix(eventindexer): correct V2 error messages in event handlers

HenmaSab0:
Fixed error messages in V2 event handlers that were missing the V2 suffix, making it clear which version failed during debugging.


davidtaikocha:
Can't see too many improvements there.

PR #21304, opened 2026-02-09
fix(taiko-client-rs): better caching based on nethermind code

MatusKysel:


MatusKysel:
@claude re-review plz

MatusKysel:
@claude re-review plz

davidtaikocha:
@claude plz review.

packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/importer/signer.rs
```diff
@@ -76,70 +116,509 @@ where
         )))
     }
 
-    /// Return (current, next) sequencer addresses, using the TTL cache when fresh.
+    /// Return (current, next) sequencer addresses, using cache when fresh.
     async fn cached_whitelist_sequencers(&mut self, now: Instant) -> Result<CachedSequencers> {
```
davidtaikocha, line 95: Looks like if we use `tokio::try_join!` + `calls with bindings` directly instead of using `batch JSON RPC`, the code modification will be hugely simplified, which would be 5 extra RPC calls every ~6mins, seems acceptable, wdyt?

MatusKysel: fixed https://github.com/taikoxyz/taiko-mono/pull/21304/commits/fff4bea00382e89da55f05d74330fc63b137239b

davidtaikocha (approved):

mskrzypkows (approved):

PR #21328, opened 2026-02-13
feat(taiko-client-rs): strict inbound preconf gossipsub validation

MatusKysel:

PR #21329, opened 2026-02-13
feat(taiko-client-rs): replace lookahead-based whitelist validation with configured sequencer allowlist

MatusKysel:


packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/network.rs
davidtaikocha, line 38: If I'm not wrong, changes here are duplicated with the ones in https://github.com/taikoxyz/taiko-mono/pull/21328/changes ?

MatusKysel: sorry this is still not ready, i mean this PR

davidtaikocha: i see, just converted it into a draft, feel free to open it again when ready.

MatusKysel:
@claude + @codex re-review

packages/taiko-client-rs/bin/client/src/flags/preconfirmation.rs
```diff
@@ -27,6 +28,19 @@ pub struct PreconfirmationArgs {
...
+    #[clap(
+        long = "p2p.sequencer-addresses",
+        env = "P2P_SEQUENCER_ADDRESSES",
+        value_delimiter = ','
+    )]
+    pub p2p_sequencer_addresses: Vec<Address>,
+
+    /// Single fallback sequencer address for legacy non-whitelist p2p signing checks.
+    #[clap(long = "p2p.sequencer-address", env = "P2P_SEQUENCER_ADDRESS")]
+    pub p2p_sequencer_address: Option<Address>,
```
davidtaikocha: Not sure why we need this `p2p_sequencer_address`, what's the issue of just having one `p2p_sequencer_addresses`?

packages/taiko-client-rs/bin/client/src/flags/preconfirmation.rs
```diff
@@ -27,6 +28,19 @@ pub struct PreconfirmationArgs {
...
     pub p2p_static_peers: Vec<String>,
 
+    /// Comma-separated list of allowed sequencer addresses for whitelist preconfirmation
+    /// gossipsub.
+    #[clap(
+        long = "p2p.sequencer-addresses",
+        env = "P2P_SEQUENCER_ADDRESSES",
+        value_delimiter = ','
+    )]
+    pub p2p_sequencer_addresses: Vec<Address>,
```
davidtaikocha, line 38: Seems in Go impl, there is no this flag? And if we want to introduce this allow list to simplify the Rust design, shall we just use this allow list (i mean, remove the whole real-time lookahead cache, since actually we only have a limited number of sequencers)?

packages/taiko-client-rs/bin/client/src/flags/preconfirmation.rs
MatusKysel: this is only to match go implementation, if we don't want to I can remove it and keep only `p2p_sequencer_addresses`

packages/taiko-client-rs/bin/client/src/flags/preconfirmation.rs
MatusKysel: yeah, that' something I wanted to discuss, but maybe we need to also talk to the partners what they think about it. IMO this is much better to have only this flag, but maybe you or partners are against this

packages/taiko-client-rs/bin/client/src/flags/preconfirmation.rs
davidtaikocha: Let's remove it then, we don't need to move the things we feel don't make sense from the Go impl.

packages/taiko-client-rs/bin/client/src/flags/preconfirmation.rs
davidtaikocha: I'm OK with that, let's discuss with partners?

MatusKysel: [9ec07b6](https://github.com/taikoxyz/taiko-mono/pull/21329/commits/9ec07b6a3618f493f73ad56f2985313dd4a5b99c)

packages/taiko-client-rs/bin/client/src/flags/preconfirmation.rs
```diff
@@ -27,6 +28,24 @@ pub struct PreconfirmationArgs {
...
+    /// Accept whitelist preconfirmation messages from all sequencers.
+    /// This bypasses sequencer allowlist checks and should only be used for
+    /// debugging or trusted deployments.
+    #[clap(
+        long = "p2p.allow-all-sequencers",
+        env = "P2P_ALLOW_ALL_SEQUENCERS",
+        default_value = "false"
+    )]
+    pub p2p_allow_all_sequencers: bool,
```
davidtaikocha, line 47: Shall we further remove this flag, and make it the only behavior in the current codebase, if we decide to accept this way? (which can hugely simplify the current Rust impl imo and it's always easy to add Golang parity stuffs back with code agent)

davidtaikocha (approved):

RogerLamTd (approved):

PR #21342, opened 2026-02-18
fix(ejector): call removeOperatorByAddress on whitelist

MatusKysel:


packages/ejector/src/utils/eject.rs
```diff
@@ -56,7 +56,7 @@ async fn eject_operator_internal(
 
     info!(reason = reason, operator = %operator_hex, "Sending removeOperator transaction");
 
-    let pending = preconf_whitelist.removeOperator(operator, true).send().await?;
+    let pending = preconf_whitelist.removeOperatorByAddress(operator).send().await?;
```
davidtaikocha: Seems i don't see this method in Shasta preconf whitelist contract: https://github.com/taikoxyz/taiko-mono/blob/taiko-alethia-protocol-v3.0.0/packages/protocol/contracts/layer1/preconf/iface/IPreconfWhitelist.sol

MatusKysel: I was using this one https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/layer1/preconf/impl/PreconfWhitelist.sol#L89 -> so I should use the branch and interface you posted ?

davidtaikocha: yep `taiko-alethia-protocol-v3.0.0` is the Shasta contract branch, rn `main` contains some whitelist preconfirmation changes.

MatusKysel: fixed [04ce528](https://github.com/taikoxyz/taiko-mono/pull/21342/commits/04ce5289301972c9ca5fccd75e0935e6e11bde52)

davidtaikocha (approved):

YoGhurt111 (approved):

PR #21347, opened 2026-02-19
fix(protocol): add evidence bounds validation in LookaheadSlasher

dantaik:
Fixes HIGH severity finding F1 (Taiko Shasta 3.0.0): `LookaheadSlasher.slash()` previously interpreted attacker-controlled evidence bytes as structs using inline-assembly pointer casting (e.g., `evidence := _bytes.offset`). This bypassed ABI decoding and therefore did not enforce length/bounds checks, which could allow malformed evidence to trigger out-of-bounds reads.

Changes:
- Decode the evidence tuple and all evidence structs via `abi.decode`.
- Remove the unsafe assembly-based evidence decoding helper.

Impact:
- Malformed evidence now reliably reverts during ABI decoding instead of being treated as structs.

---

> **Discovery:** This issue was identified during an [EVMBench](https://github.com/paradigmxyz/evmbench)-assisted security review of the Shasta 3.0.0 contracts (`taiko-alethia-protocol-v3.0.0` branch).

### DIFF
--- a/packages/protocol/contracts/layer1/preconf/impl/LookaheadSlasher.sol
+++ b/packages/protocol/contracts/layer1/preconf/impl/LookaheadSlasher.sol
@@ -90,7 +90,7 @@ contract LookaheadSlasher is ILookaheadSlasher, EssentialContract {
     // --------------------------------------------------------------------------
 
     function _validateLookaheadEvidence(
-        bytes calldata _evidenceLookaheadBytes,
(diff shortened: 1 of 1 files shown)


AnshuJalan:
This change will bring the calldata into memory, making the gas usage unreasonably high. Given that slashing will be done by a well-tested preconfirmation node and is not a user-facing op, such low-level calldata manipulation should not be a problem.

PR #21348, opened 2026-02-19
fix(taiko-client-rs): make Shasta event sync resilient on genesis and pre-finality startup

MatusKysel:
## Summary

  Fixes current Shasta driver startup behavior for two real-world bootstrap cases:

  1) Forkchoice payload handling in whitelist driver now uses explicit JSON pre-serialization in FCU calls, matching existing engine new payload behavior.
  2) Event sync bootstrap no longer fails fatally when finality is not yet available; it waits and retries instead of exiting.

### DIFF
--- a/packages/taiko-client-rs/crates/driver/src/sync/event.rs
+++ b/packages/taiko-client-rs/crates/driver/src/sync/event.rs
@@ -145,8 +145,8 @@ fn resolve_zero_target_start_block(
 ///
 /// - When finalization is available, target is bounded by `min(resume, finalized_safe)`.
 /// - When finalization is unavailable and `resume_proposal_id == 0` (genesis), target is 0.
-/// - When finalization is unavailable and `resume_proposal_id > 0`, finalization is required
-///   and a `MissingFinalizedL1Block` error is returned.
+/// - When finalization is unavailable and `resume_proposal_id > 0`, finalization is required and a
+///   `MissingFinalizedL1Block` error is returned.
 fn resolve_target_with_optional_finalization(
     resume_proposal_id: u64,
     finalized_safe_proposal_id: Option<u64>,
@@ -666,9 +666,7 @@ where
     ///
     /// Returns `None` when the L1 chain has not yet finalized (e.g. fresh devnets).
     #[instrument(skip(self), level = "debug")]
-    async fn try_finalized_l1_snapshot(
-        &self,
-    ) -> Result<Option<FinalizedL1Snapshot>, SyncError> {
+    async fn try_finalized_l1_snapshot(&self) -> Result<Option<FinalizedL1Snapshot>, SyncError> {
         let finalized_block = self
             .rpc
             .l1_provider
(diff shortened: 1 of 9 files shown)


packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -138,6 +148,7 @@ fn resolve_zero_target_start_block(
 }
 
 /// Convert missing finalized block responses into an explicit fail-closed sync error.
+#[cfg(test)]
 fn require_finalized_block<T>(value: Option<T>) -> Result<T, SyncError> {
     value.ok_or(SyncError::MissingFinalizedL1Block)
 }
```
davidtaikocha: If it's only for tests, let's move this to the `mod tests` block below.

packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -249,17 +260,12 @@ where
...
-                    Ok(None) => {
-                        warn!(
-                            block_number,
-                            "rejecting preconfirmation payload in ingress loop: head_l1_origin missing"
-                        );
-                        let _ = job.respond_to.send(Err(DriverError::PreconfIngressNotReady));
-                        gauge!(DriverMetrics::PRECONF_QUEUE_DEPTH).set(rx.len() as f64);
-                        continue;
-                    }
+                    Ok(None) => 0,
```
davidtaikocha, line 268: Seems I have another idea to solve this problem, not sure if that will be better, let me open a PR for that later, and then we can review together with this solution.

packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -53,6 +53,10 @@ use rpc::{RpcClientError, blob::BlobDataSource, client::Client};
 ///
 /// Covers both the enqueue operation and awaiting the processing response.
 const PRECONFIRMATION_PAYLOAD_SUBMIT_TIMEOUT: Duration = Duration::from_secs(12);
+/// Poll interval in seconds used to retry waiting for finalized L1 snapshots.
+const FINALITY_POLL_INTERVAL_SECS: u64 = 12;
+/// Poll interval used to retry waiting for finalized L1 snapshots.
+const FINALITY_POLL_INTERVAL: Duration = Duration::from_secs(FINALITY_POLL_INTERVAL_SECS);
```
davidtaikocha: If the finalized block is `None` for the new L1 devent,  shall we just modify `event_stream_start_block` to accept it as `None`? And simply use `resume_proposal_id` (probly in this case it's `0` as well) as the `target_proposal_id`?

MatusKysel: [fb86c33](https://github.com/taikoxyz/taiko-mono/pull/21348/commits/fb86c33054e40eb840ce7900897316ad0ef1495e)

packages/taiko-client-rs/crates/driver/src/sync/event.rs
MatusKysel: [fb86c33](https://github.com/taikoxyz/taiko-mono/pull/21348/commits/fb86c33054e40eb840ce7900897316ad0ef1495e)

packages/taiko-client-rs/crates/driver/src/sync/event.rs
davidtaikocha: and we can remove `FINALITY_POLL_INTERVAL` as well?

MatusKysel: no, it's still used when resume_proposal_id > 0 but L1 hasn't finalized yet (e.g., restarting the driver against a fresh L1 that already has proposals)

MatusKysel: [6fae8c3](https://github.com/taikoxyz/taiko-mono/pull/21348/commits/6fae8c3337305407be8cba54909160cb57ca27e5)

packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -675,30 +705,49 @@ where
...
+        let start_block = if target_proposal_id == 0 {
+            finalized_snapshot.as_ref().map_or(0, |snapshot| {
+                resolve_zero_target_start_block(
+                    snapshot.finalized_safe_proposal_id,
+                    snapshot.block_number,
+                )
+            })
+        } else {
+            0
+        };
```
davidtaikocha: Why when `target_proposal_id != 0`, we need to scan the L1 from genesis? (`start_block == 0`)

MatusKysel: fixed

davidtaikocha: hmm seems no new commit in this PR yet?

MatusKysel: [011a0ab](https://github.com/taikoxyz/taiko-mono/pull/21348/commits/011a0abb211a1107ad2cee3bd76bb529c730b60b) sorry I forgot to push

davidtaikocha (approved):

YoGhurt111 (approved):

PR #21361, opened 2026-02-23
fix(eventindexer): add defer Close() to event iterators

VolodymyrBg:
Event iterators returned from Filter* functions contain an internal subscription that must be released via Close(). Without calling Close(), the subscription's resources (channels, goroutines) are never freed, causing a memory leak.

This commit adds [defer events.Close()]https://github.com/taikoxyz/taiko-mono/blob/087a04c1c83504dbbb01ee30e5e5b4358e3f8b29/packages/eventindexer/contracts/shasta/inbox/Inbox.go#L3104-L3109 after the initial Next() check in all save*Events functions across the eventindexer package.


MatusKysel:
@VolodymyrBg thanks for the PR, your finding is correct but the way you fixing it very wrong, closing this PR

PR #21365, opened 2026-02-24
feat(protocol): add EIP-3009 support for gasless bridged token transfers

dantaik:
## Summary

Adds EIP-3009 (Transfer With Authorization) support to Taiko's bridged ERC20 tokens, enabling gasless transfers via x402.

**Why this matters:**
- Users can transfer bridged tokens without holding ETH for gas
- Relayers can submit transactions on behalf of users
- Enables better UX for bridged token transfers

> **Note:** The bridged USDC on Taiko ([0x07d83526730c7438048D55A4fc0b850e2aaB6f0b](https://taikoscan.io/address/0x07d83526730c7438048D55A4fc0b850e2aaB6f0b)) already supports gasless transfers via Circle's FiatTokenV2_2 implementation. This PR adds the same capability to other bridged tokens using Taiko's BridgedERC20 system.

## Test plan

- [x] 24 unit tests covering all EIP-3009 functions
- [x] Fuzz tests for amounts, nonces, and validity windows
- [x] Storage layout compatibility verified
- [x] Upgrade path from V2 to V3 tested

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


packages/protocol/contracts/shared/vault/BridgedERC20V3.sol
```diff
@@ -0,0 +1,236 @@
...
+        bytes32 _structHash,
+        uint8 _v,
+        bytes32 _r,
+        bytes32 _s
+    )
+        private
+        view
+    {
+        bytes32 hash = _hashTypedDataV4(_structHash);
+        address recovered = ECDSAUpgradeable.recover(hash, _v, _r, _s);
```
davidtaikocha: Looks like the current USDC also supports smart contract signer: https://github.com/circlefin/stablecoin-evm/blob/master/contracts/util/SignatureChecker.sol#L46 , shall we also support it here?

dantaik: Taiko’s USDC on L2 already supports EIP-3009.

PR #21368, opened 2026-02-24
feat(protocol): prepare mainnet shasta

ggonzalez94:
- Set the values for mainnet(fork, activator, prover)
- Updates the MainnetVerifier so that it requires ANY two proofs. This will allow us to deprecate Geth later without having to do another upgrade

---

> [!NOTE]
> **High Risk**
> Updates L1 proof-sufficiency logic in `MainnetVerifier`, which is security-critical and could change which proof combinations are accepted on mainnet. Also adjusts deployment-time role wiring and hardcodes mainnet parameters, increasing the blast radius of misconfiguration.
> 
> **Overview**
> **Verifier policy change:** `MainnetVerifier.areVerifiersSufficient` now accepts *either* `SGX_GETH` or `SGX_RETH` as the SGX proof, paired with one of the other verifiers (`SGX_*`, `RISC0`, or `SP1`), enabling future deprecation of Geth without another upgrade.
> 
> **Deployment changes:** Shasta L1 deploy config adds explicit `proverManager` and `ejectorManager` roles and wires them into `ProverWhitelist` and `PreconfWhitelist` constructors; mainnet and L2-mainnet scripts switch to hardcoded activator/prover list/fork timestamp and update mainnet verifier dependency addresses.
> 
> **Tests/metrics:** Extends `ComposeVerifier.t.sol` with `MainnetVerifier` cases and refreshes the layer1 gas report accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8866c26d0640404e052b59ac709e90f319349b6e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/protocol/script/layer1/core/DeployShastaMainnet.s.sol
```diff
@@ -32,8 +32,9 @@ contract DeployShastaMainnet is DeployShastaContracts {
         config.sp1PlonkVerifier = 0xcdCEBD75cDcb9DEd637D537776431Db563Ff0821;
 
         // Load deployment-specific values from environment
-        config.activator = vm.envAddress("ACTIVATOR");
-        config.provers = vm.envAddress("PROVERS", ",");
-        config.shastaForkTimestamp = uint64(vm.envUint("SHASTA_FORK_TIMESTAMP"));
+        config.activator = 0xF14Dc4EdDb43e9a6A440e6beC97ea2ea64f39Ef7;
+        config.provers = new address[](1); //TODO: validate we'll keep using the same prover. What other addresse(s) we want to whitelist?
```
davidtaikocha: I remember that we have one backup prover which always just generates Sgx proofs when the main prover's window expires, @YoGhurt111 are we still using it? If so, we might need to add it here.

YoGhurt111: Previously, we decided to stop using the back prover in order to increase zk coverage.

Is that still our current goal?

@ggonzalez94 @smtmfft

packages/protocol/contracts/layer1/mainnet/MainnetVerifier.sol
```diff
@@ -29,7 +29,9 @@ contract MainnetVerifier is ComposeVerifier {
     { }
 
     /// @notice Check if the provided verifiers are sufficient
-    /// @dev Requires exactly 2 verifiers: SGX-GETH + (SGX or RISC0 or SP1)
+    /// @dev Requires exactly 2 verifiers:
+    /// SGX-GETH + (SGX-RETH or RISC0 or SP1), or
+    /// SGX-RETH + (SGX-GETH or RISC0 or SP1)
```
YoGhurt111, line 34: We don't support SGX-RETH + (RISC0 or SP1) in our prover software.

ggonzalez94: yes, but we'll need to support it in order to deprecate Geth. Better to have it ready on the contracts

packages/protocol/script/layer1/core/DeployShastaMainnet.s.sol
smtmfft: I think we always need backup prover to cover zk failures.

packages/protocol/contracts/layer1/mainnet/MainnetVerifier.sol
YoGhurt111: Yes, this is a nice-to-have. However, based on my understanding, once we deprecate geth, we will no longer need combined proofs. We would only require a single proof type—either SGX, RISC0, or SP1.

packages/protocol/script/layer1/core/DeployShastaMainnet.s.sol
YoGhurt111: 1. We don't use the old backup prover address, we need to create a new one and I don't have the access to add it to mainnet cluster? @ggonzalez94 Could you do thay? If yes, I will share a command to you, you can use it to add the private key.
2. Do we need to do the same as on Hoodi—i.e., have only us run the prover, so sidecars don’t need to maintain their own prover?
3. Our current prover address is `0xa5cb34B75bD72f15290ef37A01F06183E8036875`

packages/protocol/script/layer1/core/DeployShastaMainnet.s.sol
```diff
@@ -32,8 +32,9 @@ contract DeployShastaMainnet is DeployShastaContracts {
...
+        config.provers[0] = 0xC72cB4489E8e294177E7A2FcCcc818F844Ab2a1D;
```
YoGhurt111: Where did this address `0xC72cB4489E8e294177E7A2FcCcc818F844Ab2a1D` come from?

packages/protocol/contracts/layer1/mainnet/MainnetVerifier.sol
ggonzalez94: we should still require 2. In case the proof system is buggy

packages/protocol/script/layer1/core/DeployShastaMainnet.s.sol
ggonzalez94: 1. we don't need to do this for now, since we added a proverManager role [here](https://github.com/taikoxyz/taiko-mono/pull/21382) we can add more entities to the prover list later if needed.
2. yes, we'll be running the prover for everyone
3. gotcha, replaced the old one with this

packages/protocol/script/layer1/core/DeployShastaMainnet.s.sol
ggonzalez94: I think it is one of our partner's prover. I hadn't realized we submit our proofs via a smart contract(Taiko Labs Prover), so ours don't actually show up as txs in the inbox. I already replaced it on the script with the one you shared

YoGhurt111 (approved):

packages/protocol/script/layer1/core/DeployShastaMainnet.s.sol
```diff
@@ -31,9 +28,11 @@ contract DeployShastaMainnet is DeployShastaContracts {
         config.sgxRethAutomataProxy = 0x8d7C954960a36a7596d7eA4945dDf891967ca8A3;
         config.sp1PlonkVerifier = 0xcdCEBD75cDcb9DEd637D537776431Db563Ff0821;
```
smtmfft: We decided to deprecate our self-deployment, so the verifiers come from vender's gateway contract:
For mainnet, they are: 
        config.r0Groth16Verifier = 0x8EaB2D97Dfce405A1692a21b3ff3A172d593D319;
        config.sp1PlonkVerifier = 0x3B6041173B80E77f038f3F2C0f9744f04837185e;

Reference:
https://dev.risczero.com/api/blockchain-integration/contracts/verifier
https://docs.succinct.xyz/docs/sp1/verification/contract-addresses

smtmfft (changes requested):

packages/protocol/script/layer1/core/DeployShastaMainnet.s.sol
YoGhurt111:
```suggestion
        config.r0Groth16Verifier = 0x8EaB2D97Dfce405A1692a21b3ff3A172d593D319;
```

packages/protocol/script/layer1/core/DeployShastaMainnet.s.sol
YoGhurt111:
```suggestion
        config.sp1PlonkVerifier = 0x3B6041173B80E77f038f3F2C0f9744f04837185e;
```

YoGhurt111 (approved):

davidtaikocha (approved):

PR #21396, opened 2026-02-27
 repo: replace placeholder vault names with actual contract names

cmende:
Two docs referenced vault contracts with placeholder names (`ERCXXX_Vault`, `TokenVault contract`) instead of the real implementations.

  The actual contracts — `ERC20Vault`, `ERC721Vault`, and `ERC1155Vault` — exist in `packages/protocol/contracts/shared/vault/`. Using placeholder names in docs is misleading and makes it harder for developers to locate the right contracts.

  This replaces both occurrences with the correct contract names.


davidtaikocha:
Closed due to i don't see many improvements here.

PR #21414, opened 2026-03-03
fix(relayer): return one-shot processor result

edgarr1986:
Return `processSingle` errors in `targetTxHash` mode instead of calling `os.Exit(0)`, skip interrupt waiting for one-shot processor runs.


edgarr1986:
Hi @MatusKysel, I’ve pushed a follow-up fix for the relayer lint issue in this PR. Could you please take another look when you have a moment? Thanks for your time.

MatusKysel (approved):

davidtaikocha (approved):

PR #21429, opened 2026-03-04
fix: enforce pause checks

mk0walsk:
This PR makes bridge pause checks actually block execution.

  Several bridge flows were calling `isBridgePaused().then(...)`, which does not gate the current async flow and can throw outside local `try/catch`. That means paused-state checks could be bypassed in practice.
  I replaced those call sites with awaited guards (`if (await isBridgePaused()) ...`) in the bridge/approval paths and used early returns in UI handlers where throwing would become an unhandled async error.

  I also added explicit `BridgePausedError` handling in `handleBridgeError`, so paused-state failures show a clear paused message instead of falling back to a generic unknown error.


davidtaikocha:
Closed due to being outdated.

PR #21433, opened 2026-03-05
refactor(ejector): avoid duplicated max_depth.max(1) calculation

Ocheretovich:
Avoids repeating max_depth.max(1) by normalizing the value once and reusing it for both history capacity and the max_depth field.


davidtaikocha:
Closed due to dont see to much improvements here.

PR #21442, opened 2026-03-08
refactor: avoid cloning client config URLs

jschnelder:
During provider initialization, we directly consume the RPC client configuration URLs—this eliminates redundant clones while preserving the original error reporting logic unchanged.


davidtaikocha:
closed due to can't see too much improvements here.

PR #21450, opened 2026-03-10
feat(taiko-client-rs): support configurable L1 transports

MatusKysel:


MatusKysel:
@codex plz review

MatusKysel:
@claude & @codex plz re-review

packages/taiko-client-rs/bin/client/src/flags/common.rs
```diff
@@ -3,21 +3,39 @@
...
-        env = "L1_WS",
-        required = true,
-        help = "Websocket RPC endpoint of a L1 ethereum node"
+        long = "l1.transport",
+        env = "L1_TRANSPORT",
+        value_enum,
+        help = "Select which L1 transport to use when multiple endpoints are configured"
     )]
-    pub l1_ws_endpoint: Url,
+    pub l1_transport: Option<L1Transport>,
```
davidtaikocha: Shall we just return an error message when `--l1.ws` and `--l1.http` are both set? Which should not happen usually if we make it clear in the flag description and we can avoid this extra complicity?

packages/taiko-client-rs/crates/protocol/src/subscription_source.rs
```diff
@@ -122,20 +123,27 @@ impl SubscriptionSource {
...
+                    .map_err(|e| format!("invalid http url: {e}")),
+                "ws" | "wss" => value
+                    .parse::<Url>()
+                    .map(SubscriptionSource::Ws)
+                    .map_err(|e| format!("invalid websocket url: {e}")),
+                _ => Err(format!("unsupported subscription source scheme: {scheme}")),
+            };
         }
+
+        Err("subscription source must use http://, https://, ws://, or wss://".to_string())
```
davidtaikocha: Let's define an explicit error type in `SubscriptionSourceError` instead of just create a string error here? Also for the `Err(format!("unsupported subscription source scheme: {scheme}"))` and `format!("invalid http url: {e}")` above.

.github/workflows/taiko-client-rs--test.yml
```diff
@@ -232,3 +232,11 @@ jobs:
...
           PROTOCOL_DIR: ${{ github.workspace }}/${{ env.PROTOCOL_FORK_DIR }}/packages/protocol
         run: just test
+
+      - name: Run WS driver tests
+        working-directory: packages/taiko-client-rs
+        env:
+          PROTOCOL_DIR: ${{ github.workspace }}/${{ env.PROTOCOL_FORK_DIR }}/packages/protocol
+          L1_TRANSPORT: ws
+          TEST_CRATE: driver
+        run: ./tests/entrypoint.sh
```
davidtaikocha: IMO we don't need to test both here, given the current CI workflow is slow enough? The changes are straight forward and we can keep the test as using ws, which is a better choice and should be used for the most users. Or we can just add one extra test case to cover the basic http client creation and query instead of one more step in CI here.

packages/taiko-client-rs/crates/protocol/src/subscription_source.rs
MatusKysel: [5433adc](https://github.com/taikoxyz/taiko-mono/pull/21450/commits/5433adc3fe03e973243cd5d510d0a91fc4acc7e5)

packages/taiko-client-rs/bin/client/src/flags/common.rs
MatusKysel: [5433adc](https://github.com/taikoxyz/taiko-mono/pull/21450/commits/5433adc3fe03e973243cd5d510d0a91fc4acc7e5)

.github/workflows/taiko-client-rs--test.yml
MatusKysel: [5433adc](https://github.com/taikoxyz/taiko-mono/pull/21450/commits/5433adc3fe03e973243cd5d510d0a91fc4acc7e5)

packages/taiko-client-rs/tests/entrypoint.sh
```diff
@@ -85,6 +85,19 @@ echo "L2_GENESIS_HASH: $L2_GENESIS_HASH"
...
+        unset L1_WS
+        ;;
+    ws)
+        unset L1_HTTP
+        ;;
+    *)
+        echo "unsupported L1_TRANSPORT $L1_TRANSPORT; expected http or ws"
+        exit 1
+        ;;
+esac
```
davidtaikocha: Since we have removed the WS CI step, shall we remove this extra check as well?

MatusKysel: That block in entrypoint.sh is now serving a different purpose than the removed WS CI step: it prevents the test process from seeing both L1_HTTP and L1_WS at the same time. Since common.rs now treats “both set” as an error, removing the case while still exporting both vars would reintroduce ambiguous env for anything in the suite that reads CLI config from env.

packages/taiko-client-rs/tests/entrypoint.sh
```diff
@@ -8,6 +8,8 @@ export PROTOCOL_DIR
 echo "Starting docker compose services..."
 
 export HARNESS_L1_HTTP=${HARNESS_L1_HTTP:-http://localhost:18545}
+export HARNESS_L1_WS=${HARNESS_L1_WS:-ws://localhost:18545}
+export HARNESS_L1_TRANSPORT=${HARNESS_L1_TRANSPORT:-ws}
```
davidtaikocha: Do we really need this? Since after we renamed to `HARNESS_L1_HTTP` and `HARNESS_L1_WS`, there is no conflict with the flag names?

packages/taiko-client-rs/crates/test-harness/src/shasta/env.rs
```diff
@@ -62,6 +62,25 @@ impl fmt::Debug for ShastaEnv {
 }
 
 impl ShastaEnv {
+    fn load_l1_source() -> Result<SubscriptionSource> {
+        let l1_http = env::var("L1_HTTP").ok();
+        let l1_ws = env::var("L1_WS").ok();
```
davidtaikocha: Shall we use `HARNESS_L1_HTTP` and `HARNESS_L1_WS` here?

packages/taiko-client-rs/crates/test-harness/src/shasta/env.rs
```diff
@@ -62,6 +62,25 @@ impl fmt::Debug for ShastaEnv {
...
+                let l1_http_url =
+                    RpcUrl::parse(l1_http.as_str()).context("invalid L1_HTTP endpoint")?;
+                Ok(SubscriptionSource::Http(l1_http_url))
+            }
+            (None, Some(l1_ws)) => {
+                let l1_ws_url = RpcUrl::parse(l1_ws.as_str()).context("invalid L1_WS endpoint")?;
+                Ok(SubscriptionSource::Ws(l1_ws_url))
+            }
+            (Some(_), Some(_)) => bail!("configure exactly one of L1_HTTP or L1_WS"),
+            (None, None) => bail!("L1_HTTP or L1_WS env var is required"),
```
davidtaikocha: Let's use WS as the default in tests since it's faster for subscription, and only test HTTP in the new test cases to make sure it works.

packages/taiko-client-rs/tests/entrypoint.sh
```diff
@@ -7,8 +7,9 @@ export PROTOCOL_DIR
 
 echo "Starting docker compose services..."
 
-export L1_HTTP=http://localhost:18545
-export L1_WS=ws://localhost:18545
+export HARNESS_L1_HTTP=${HARNESS_L1_HTTP:-http://localhost:18545}
+export HARNESS_L1_WS=${HARNESS_L1_WS:-ws://localhost:18545}
```
davidtaikocha, line 11: And since we are starting two nodes in docker-compose and the second node is using `L2_WS_1` and `L2_HTTP_1`, shall we just change to `L2_WS_0` and `L2_HTTP_0` here?

packages/taiko-client-rs/tests/entrypoint.sh
```diff
@@ -84,6 +85,22 @@ echo "L2_GENESIS_HASH: $L2_GENESIS_HASH"
 
 $DIR/deploy.sh
 
+# The test process should see exactly one L1 endpoint to match the client CLI rules.
+case "$HARNESS_L1_TRANSPORT" in
+    http)
+        export L1_HTTP="$HARNESS_L1_HTTP"
+        unset L1_WS
```
davidtaikocha: We are not setting the values by `export` anymore, do we really need to `unset` here?

davidtaikocha (approved):

YoGhurt111 (approved):

PR #21452, opened 2026-03-10
fix(relayer): standardize bridge gas estimation error log key

jschnelder:
Use the standard `error` field when gas estimation falls back, improving structured log filtering for relayer bridge operators.


davidtaikocha:
it's ok to leave b.estimateGas there imo.

PR #21453, opened 2026-03-11
fix(overseer-rs): remove redundant RPC URL clone

jschnelder:
This removes an unnecessary `clone()` when building the startup HTTP provider in `overseer-rs`. The parsed `rpc_url` is consumed only once and never reused, so passing it directly preserves behavior while simplifying the startup path; `cargo fmt`, `cargo test`, and `cargo clippy` all pass for `packages/overseer-rs`.


RogerLamTd:
overseer has been deprecated

PR #21459, opened 2026-03-12
fix(taiko-client): preserve Shasta base fee config

odupontt:
This change hardens Shasta base-fee config assembly by starting from the network defaults and then injecting the runtime `ShastaTime`, instead of relying on a partial config shape in the call site.
  That is necessary because this exact path recently regressed when switching to `NetworkIDToChainConfigOrDefault`, which dropped the fork-time input and needed a follow-up fix. The added regression test shows we now preserve the chain-specific EIP-4396 settings, keep the runtime fork time, and avoid mutating the shared default config.


davidtaikocha:
closed due to can't see many improvements here :)

PR #21464, opened 2026-03-13
Protect Shasta config assembly

odupontt:
This change makes Shasta base-fee config assembly explicit and safe by copying the network defaults before applying the runtime `ChainID` and `ShastaTime`.

  That guard is necessary because this path already regressed in `0070c5ac`, where switching to `NetworkIDToChainConfigOrDefault` dropped the fork-time input and needed an immediate follow-up fix. The new regression test proves we now preserve both the chain-specific defaults and the runtime Shasta settings without mutating the shared default config.


davidtaikocha:
https://github.com/taikoxyz/taiko-mono/pull/21459

PR #21467, opened 2026-03-16
fix(taiko-client-rs): query shasta inbox core state by block number

YoGhurt111:


MatusKysel (approved):

packages/taiko-client-rs/crates/driver/src/derivation/pipeline/shasta/pipeline/mod.rs
```diff
@@ -241,15 +241,15 @@ where
 
     /// Read the inbox core state at the proposal log's block to extract the last finalized id.
     async fn inbox_last_finalized_proposal_id(&self, log: &Log) -> Result<u64, DerivationError> {
-        let block_hash = log
-            .block_hash
-            .ok_or_else(|| DerivationError::Other(anyhow!("proposal log missing block hash")))?;
```
davidtaikocha, line 246: For here, i think we should keep using `block_hash` to prevent the `log` being reorged?

PR #21470, opened 2026-03-16
ci(taiko-client): add optional cache dependency paths to speedup dependency install

RogerLamTd:


packages/taiko-client/scripts/common.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+#abc
```
davidtaikocha:
```suggestion
```

.github/workflows/taiko-client--test.yml
```diff
@@ -167,6 +164,14 @@ jobs:
           path: ${{ env.SHASTA_FORK_TAIKO_MONO_DIR }}
           ref: taiko-alethia-protocol-v3.0.0
 
+      - name: Install pnpm dependencies
+        uses: ./.github/actions/install-pnpm-dependencies
+        with:
+          cache-dependency-path: |
+            pnpm-lock.yaml
+            ${{ env.PACAYA_FORK_TAIKO_MONO_DIR }}/pnpm-lock.yaml
+            ${{ env.SHASTA_FORK_TAIKO_MONO_DIR }}/pnpm-lock.yaml
```
davidtaikocha, line 173: Shall we update the one in `taiko-client-rs` as well?

RogerLamTd: > Shall we update the one in `taiko-client-rs` as well?

Screenshot 2026-03-17 at 6 58 43 AM

seems this solution is not foolproof, if cache is not hit pnpm dependencies can still time out

davidtaikocha (approved):

MatusKysel (approved):

PR #21474, opened 2026-03-17
perf(taiko-client-rs): avoid payload clones in engine submission

edgarr1986:
Move payload ownership directly in `envelope_into_submission` to avoid redundant clones, preserve block metadata extraction and post-Shanghai withdrawals semantics.


davidtaikocha:
can't see much improvements here.

PR #21480, opened 2026-03-18
feat(taiko-client-rs): configure event sync retry limit

YoGhurt111:
## Summary

Auto-generated PR based on current working tree changes.
- Title basis: feat(taiko-client-rs): configure event sync retry limit
- Changed files: 11

## What Changed

| File | +Lines | -Lines |
| --- | ---: | ---: |
| `packages/taiko-client-rs/bin/client/src/commands/driver.rs` | 1 | 0 |
| `packages/taiko-client-rs/bin/client/src/commands/preconfirmation_driver.rs` | 1 | 0 |
| `packages/taiko-client-rs/bin/client/src/commands/whitelist_preconfirmation_driver.rs` | 1 | 0 |
| `packages/taiko-client-rs/bin/client/src/flags/driver.rs` | 45 | 0 |
| `packages/taiko-client-rs/crates/driver/src/config.rs` | 38 | 0 |
| `packages/taiko-client-rs/crates/driver/src/sync/event.rs` | 26 | 2 |
| `packages/taiko-client-rs/crates/driver/tests/proposer_driver_e2e.rs` | 4 | 0 |
| `packages/taiko-client-rs/crates/driver/tests/shasta_event_sync.rs` | 2 | 0 |
| `packages/taiko-client-rs/crates/preconfirmation-driver/tests/dual_driver_e2e.rs` | 2 | 0 |
| `packages/taiko-client-rs/crates/preconfirmation-driver/tests/preconf_e2e.rs` | 2 | 0 |
| `packages/taiko-client-rs/crates/test-harness/src/preconfirmation/driver.rs` | 2 | 0 |

## Commit Context

- Branch generated from current HEAD with local modifications.
- Commit message and PR content inferred from file-level changes.

## Testing

- [ ] Not run (auto-generated; fill manually)
- [ ] Local tests passed (fill manually)

## Risks

- [ ] No breaking changes expected
- [ ] Potential behavior changes reviewed

## Checklist

- [x] Branch created from current modifications
- [x] Changes committed and pushed
- [x] PR description generated with per-file diff stats

## Operational Behavior Change

This PR changes the retry behavior for transient failures during event-sync proposal processing.

Before this change, proposal derivation retried indefinitely with exponential backoff capped at 12 seconds. The driver would continue retrying until the transient failure cleared.

After this change, proposal derivation retries are capped by event_sync_max_retries in DriverConfig and can be configured via `--driver.eventSyncMaxRetries / DRIVER_EVENT_SYNC_MAX_RETRIES`. Once the retry budget is exhausted, the error is propagated and the event sync loop exits instead of retrying forever.

This is an operational behavior change. Deployments that previously relied on the driver to self-heal through extended transient RPC or infrastructure outages may now observe driver termination after repeated failures, depending on the configured retry limit.

Default behavior in this PR:

`event_sync_max_retries` defaults to `10`
backoff starts at `10ms`
backoff is capped at `12s`
Operators should ensure the runtime environment has an appropriate restart policy or supervisor if bounded retries are used in production.


packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -346,8 +358,7 @@ where
                 "dispatching proposal log to derivation pipeline"
             );
             // Retry proposal processing on transient errors.
-            let retry_strategy =
-                ExponentialBackoff::from_millis(10).max_delay(Duration::from_secs(12));
+            let retry_strategy = event_sync_retry_strategy(self.cfg.event_sync_max_retries);
```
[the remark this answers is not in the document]
YoGhurt111: This is intentional.

We do not expect the driver to reliably self-recover from this class of repeated failure, so retrying forever in-process is not the behavior we want. Once the bounded retry budget is exhausted, we prefer to fail fast, exit the sync task/process, and rely on Kubernetes restart behavior to retry from a clean state.

So the bounded retries are deliberate, but we agree this is an operational behavior change and should be called out clearly in the PR description.

YoGhurt111:
@claude 
Review 1: Added `Operational Behavior Change` in the PR description.
Review 2: Fixed.
Review 3: This helper is private and only used locally in event.rs, so exposing Take<ExponentialBackoff> is not part of any public API surface. I agree impl Iterator<Item = Duration> would be slightly cleaner, but I’d treat that as optional refactoring rather than something we need to change in this PR.

MatusKysel (approved):

packages/taiko-client-rs/crates/driver/src/sync/event.rs
davidtaikocha, line 361: I think for all kinds of errors EXCEPT reorg, we must not miss any event log from canonical chain, no matter how many times we've retired?

packages/taiko-client-rs/crates/driver/src/config.rs
```diff
@@ -34,14 +39,72 @@ impl DriverConfig {
         l1_beacon_endpoint: Url,
         l2_checkpoint_url: Option<Url>,
         blob_server_endpoint: Option<Url>,
+        event_sync_max_retries: usize,
     ) -> Self {
+        assert!(event_sync_max_retries > 0, "event_sync_max_retries must be at least 1");
```
davidtaikocha: Let's throw an error instead of `assert!` here.

packages/taiko-client-rs/crates/driver/src/sync/event.rs
YoGhurt111: We can safely keep retrying to handle the event log until success, as long as the L1 state remains unchanged between fetching and processing the event log.
However, the observed behavior is not like this.

YoGhurt111:
Will fix lint and test later

packages/taiko-client-rs/crates/driver/src/config.rs
YoGhurt111: [598c704](https://github.com/taikoxyz/taiko-mono/pull/21480/commits/598c704928b990a5cfcc4ee4c84b5c37164800ee)
We can delete the assert directly, since we already have `NonZeroUsize` now.

PR #21487, opened 2026-03-19
feat(taiko-client-rs): make finalized proposal id optional in shasta derivation

davidtaikocha:
## Summary

This PR makes `last_finalized_proposal_id` optional across the Shasta derivation pipeline so proposal processing can continue when the block-pinned inbox `getCoreState()` lookup fails.

## What Changed

- added a best-effort helper to read `lastFinalizedProposalId` from inbox core state at the proposal log's L1 block
- changed `inbox_last_finalized_proposal_id` to return `Option<u64>` instead of failing proposal derivation on lookup errors
- updated `BundleMeta.last_finalized_proposal_id` to store `Option<u64>`
- threaded the optional finalized proposal id through manifest construction and proposal processing
- only updated the `DERIVATION_LAST_FINALIZED_PROPOSAL_ID` gauge when the finalized proposal id is present
- kept finalized block hash resolution best-effort, falling back to no finalized forkchoice hint when the finalized proposal id is unavailable
- added unit tests for:
  - preserving an absent finalized proposal id in bundle metadata
  - returning `Some(id)` when the block-pinned inbox lookup succeeds
  - returning `None` when the block-pinned inbox lookup fails

## Why

Before this change, the Shasta derivation path treated the finalized proposal id as required state during proposal processing. That made proposal handling fail on inbox core-state lookup errors even though missing finalization data is acceptable and should only disable safe/finalized forkchoice hints.

With this change, proposal derivation remains forward-progressing:
- proposal logs are still decoded and processed
- finalized forkchoice hints are only used when the finalized proposal id is actually available
- lookup failures degrade gracefully instead of aborting the proposal path

## Testing

- `just fmt`
- `just clippy-fix`
- `cargo test -p driver derivation::pipeline::shasta::pipeline::tests --lib`

### DIFF
--- a/packages/taiko-client-rs/crates/driver/src/derivation/pipeline/shasta/pipeline/mod.rs
+++ b/packages/taiko-client-rs/crates/driver/src/derivation/pipeline/shasta/pipeline/mod.rs
@@ -10,7 +10,6 @@ use alloy::{
 use alloy_consensus::TxEnvelope;
 use alloy_provider::RootProvider;
 use alloy_rpc_types::{Transaction as RpcTransaction, eth::Block as RpcBlock};
-use anyhow::anyhow;
 use async_trait::async_trait;
 use bindings::inbox::{IInbox::DerivationSource, Inbox::Proposed};
 use metrics::{counter, gauge};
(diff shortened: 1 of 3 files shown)


mikhailUshakoff (approved):

packages/taiko-client-rs/crates/driver/src/derivation/pipeline/shasta/pipeline/payload.rs
```diff
@@ -251,8 +251,9 @@ where
         let mut outcomes = Vec::new();
         // Best-effort lookup of the last finalized block hash; missing data should not block
         // payload application.
-        let finalized_block_hash =
-            self.finalized_block_hash_for(meta.last_finalized_proposal_id).await;
+        let finalized_block_hash = self
+            .finalized_block_hash_for(meta.last_finalized_proposal_id.unwrap_or_default())
```
MatusKysel: If we want last_finalized_proposal_id to be truly optional, I'd rather pass Option<u64> into finalized_block_hash_for instead of converting None back to 0 here

packages/taiko-client-rs/crates/driver/src/derivation/pipeline/shasta/pipeline/mod.rs
```diff
@@ -240,19 +287,16 @@ where
...
-            .inbox
-            .getCoreState()
-            .block(BlockId::Hash(RpcBlockHash { block_hash, require_canonical: Some(false) }))
-            .call()
-            .await?;
-        Ok(core_state.lastFinalizedProposalId.to::<u64>())
+    ///
+    /// Missing block hash or inbox-query failures are treated as absence of finalization data so
+    /// derivation can continue without finalized forkchoice hints.
+    async fn inbox_last_finalized_proposal_id(&self, log: &Log) -> Option<u64> {
```
MatusKysel: Both callers reach this only after decode_log_to_event_context(), which already rejects a missing block_hash. Taking &Log here reintroduces an impossible None path and duplicates the code

MatusKysel (approved):

packages/taiko-client-rs/crates/driver/src/derivation/pipeline/shasta/pipeline/payload.rs
davidtaikocha: updated: https://github.com/taikoxyz/taiko-mono/pull/21487/changes/bb38e29728300f7562c216db024a7be2745ca41f

packages/taiko-client-rs/crates/driver/src/derivation/pipeline/shasta/pipeline/mod.rs
davidtaikocha: removed: https://github.com/taikoxyz/taiko-mono/pull/21487/changes/c63523f8208314cad68c42f65f5ce4f38a9b2f6c

YoGhurt111 (approved):

PR #21489, opened 2026-03-19
fix(taiko-client): fix prover retry submission buffer cleanup

YoGhurt111:
## Summary

This change fixes how the prover handles failed proof submissions after aggregation.

When an aggregated proof submission hits a reverted transaction or an already-proven condition, users can otherwise get stuck with stale proof items still sitting in the in-memory buffers. That causes repeated retries against proofs that should have been dropped and can block later proof processing for the same proof type.

The root cause was that proof buffer cleanup happened inside submitter implementations at the wrong point in the flow. The buffer items were cleared too early on some error paths and not managed centrally alongside prover retry exhaustion handling, which made the retry lifecycle inconsistent.

The fix moves proof buffer cleanup behind an explicit `ClearProofBuffers` submitter interface and a shared helper used by both Pacaya and Shasta submitters. The prover now clears buffers after successful submission, clears them immediately on execution-reverted submissions that indicate the batch has already been handled, and uses the new `withRetry` callback to clear stale buffered proofs after retries are exhausted.

### DIFF
--- a/packages/taiko-client/prover/prover.go
+++ b/packages/taiko-client/prover/prover.go
@@ -360,9 +360,10 @@ func (p *Prover) submitProofAggregationOp(batchProof *proofProducer.BatchProofs)
 				"error", err,
 			)
 			return nil
-		} else if strings.Contains(err.Error(), vm.ErrExecutionReverted.Error()) {
+		} else if strings.Contains(err.Error(), vm.ErrExecutionReverted.Error()) ||
+			strings.Contains(err.Error(), transaction.ErrUnretryableSubmission.Error()) {
 			log.Error(
-				"Proof submission reverted",
+				"Proof submission reverted or unretryable",
 				"blockIDs", batchProof.BatchIDs,
 				"proofType", batchProof.ProofType,
 				"error", err,
--- a/packages/taiko-client/prover/prover.go
+++ b/packages/taiko-client/prover/prover.go
@@ -385,6 +385,7 @@ func (p *Prover) submitProofAggregationOp(batchProof *proofProducer.BatchProofs)
 	return nil
 }
 
+// clearProofBuffer clears the buffered proof items for the batch from the matching submitter.
 func (p *Prover) clearProofBuffer(batchProof *proofProducer.BatchProofs) error {
 	if batchProof == nil || len(batchProof.ProofResponses) == 0 {
 		return fmt.Errorf("empty batch proof")
@@ -409,7 +410,7 @@ func (p *Prover) ProverAddress() common.Address {
 	return p.txmgr.From()
 }
 
-// withRetry retries the given function with prover backoff policy.
+// withRetry retries the given function with prover backoff policy and runs callback if retries are exhausted.
 func (p *Prover) withRetry(f func() error, callback func() error) {
 	p.wg.Add(1)
 	go func() {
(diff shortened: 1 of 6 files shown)


packages/taiko-client/prover/prover.go
```diff
@@ -253,33 +253,36 @@ func (p *Prover) eventLoop() {
 		case <-p.ctx.Done():
 			return
 		case batchProof := <-p.batchProofGenerationCh:
-			p.withRetry(func() error { return p.submitProofAggregationOp(batchProof) })
+			p.withRetry(
+				func() error { return p.submitProofAggregationOp(batchProof) },
+				func() error { return p.clearProofBuffer(batchProof) },
```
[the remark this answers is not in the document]
YoGhurt111: Fixed in [46cbfe4](https://github.com/taikoxyz/taiko-mono/pull/21489/commits/46cbfe4b8d07129481cc55facc5c983433de768f)

YoGhurt111:
@claude Plz review again.
Bug 1: Fixed.
Concern 2: Added resending logic, so we won't miss any batch when retries are exhausted.
Minor 3: It's ok.

packages/taiko-client/prover/prover.go
```diff
@@ -374,10 +382,28 @@ func (p *Prover) submitProofAggregationOp(batchProof *proofProducer.BatchProofs)
 		)
 		return err
 	}
+	if err := submitter.ClearProofBuffers(batchProof, false); err != nil {
```
MatusKysel, line 378: do we need to return error here ? maybe we can just log it as proof was already submitted and we don't need to re-submmited right ?

packages/taiko-client/prover/prover.go
```diff
@@ -374,10 +382,28 @@ func (p *Prover) submitProofAggregationOp(batchProof *proofProducer.BatchProofs)
...
+	if batchProof == nil || len(batchProof.ProofResponses) == 0 {
+		return fmt.Errorf("empty batch proof")
+	}
+	submitter := p.proofSubmitterPacaya
+	if batchProof.ProofResponses[0].Meta.IsShasta() {
+		submitter = p.proofSubmitterShasta
+	}
+	if utils.IsNil(submitter) {
+		return fmt.Errorf("submitter not found: %s", batchProof.ProofType)
+	}
```
MatusKysel, line 405: this is the same code as begging of the `submitProofAggregationOp` -> maybe we can move it to helper e.g. -> `func (p *Prover) getSubmitter(batchProof *proofProducer.BatchProofs) (proofSubmitter.Submitter, error)`

packages/taiko-client/prover/prover.go
YoGhurt111: We need to ensure `ClearProofBuffers` successful, otherwise we won't do aggregate again

packages/taiko-client/prover/prover.go
YoGhurt111: Refactored in [fafa76c](https://github.com/taikoxyz/taiko-mono/pull/21489/commits/fafa76c1f8fcde6eb52cc8841a0c84f1a08da326)

MatusKysel (approved):

davidtaikocha (approved):

davidtaikocha (approved):

PR #21490, opened 2026-03-20
feat(taiko-client-rs): skip orphaned proposal logs during event sync

davidtaikocha:
## Summary

This PR prevents `process_log_batch` from retrying forever when an L1 proposal log comes from a block that has been reorged out of the canonical chain.

On proposal-processing failure, the driver now re-checks the log's `block_hash` against L1. If the source block no longer exists, the log is treated as permanently orphaned, a warn-level log and dedicated metric are emitted, and batch processing continues with later logs. All other failures remain retryable under the existing backoff behavior.

## Changes

- add `driver_event_orphaned_proposal_logs_total`
- classify failed proposal logs as permanently orphaned only when `get_block_by_hash` returns `None`
- skip confirmed orphaned logs instead of retrying endlessly
- continue processing later logs in the same batch after an orphaned-log skip
- move `ProposalLogProcessing` to module scope for clearer control-flow typing
- add tests covering:
  - missing L1 block => permanent orphan
  - existing L1 block => retryable
  - L1 RPC error => retryable
  - missing `block_hash` => not classified orphaned
  - batch continues after skipping an orphaned log

## Operator Impact

- orphaned proposal logs now produce a warn-level skip instead of an endless retry loop
- operators can track these skips via `driver_event_orphaned_proposal_logs_total`

## Verification

- `cargo test -p driver process_log_batch_skips_orphaned_proposal_log_and_continues_batch -- --nocapture`


packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -324,6 +332,25 @@ where
...
+        };
+
+        let block = self
+            .rpc
+            .l1_provider
+            .get_block_by_hash(block_hash)
+            .await
+            .map_err(|err| SyncError::Rpc(RpcClientError::Provider(err.to_string())))?;
+
+        Ok(block.is_none())
```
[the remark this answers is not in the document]
davidtaikocha: fixed in https://github.com/taikoxyz/taiko-mono/pull/21490/changes/a65484a5ad1e14a4f0af68be26f9ee14d4185303

davidtaikocha:
@claude plz review.

packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -42,13 +42,21 @@ use crate::{
...
-        ProductionInput, ProductionRouter,
+        ProductionInput, ProductionRouter, path::EngineBlockOutcome,
     },
 };
 
 use alloy_rpc_types_engine::PayloadId;
 use rpc::{RpcClientError, blob::BlobDataSource, client::Client};
 
+/// Result of processing a single proposal log inside `process_log_batch`.
+enum ProposalLogProcessing {
```
MatusKysel: maybe something like `ProposalLogResult` would be better name

MatusKysel (approved):

packages/taiko-client-rs/crates/driver/src/sync/event.rs
davidtaikocha: updated: https://github.com/taikoxyz/taiko-mono/pull/21490/commits/afe101434afa3cc54412408a255755ea047c6e1f

YoGhurt111 (approved):

PR #21492, opened 2026-03-20
chore(protocol): fix *_Layout.sol detection in skip-worktree cleanup

Olexandr88:
Fix unreliable glob matching for *_Layout.sol files by using git ls-files + grep for consistent behavior


Olexandr88:
Hey @RogerLamTd would love to get your review when you have a moment!

PR #21496, opened 2026-03-22
feat(taiko-client-rs): add receipt timeout and tip bumping for proposal transactions in proposer

davidtaikocha:
## Summary
  - Add configurable receipt timeout (`--propose.receiptTimeout`, default 180s) to prevent the proposer from waiting indefinitely for a transaction to be mined
  - Implement automatic tip bumping retry loop: on timeout, resubmit the proposal transaction with bumped priority fee (`--propose.tipBumpPercentage`, default +20%) up to a configurable max retries (`--propose.maxTipBumpRetries`, default 3)
  - After exhausting retries, log an error and continue the proposer loop instead of blocking forever

  ## Context
  Previously, the proposer called `pending_tx.get_receipt().await` with no timeout. If the L1 mempool didn't include the transaction (e.g. due to low gas fees or L1 congestion), the proposer would block forever, never submitting another proposal.


.github/workflows/taiko-client-rs--docker.yml
```diff
@@ -4,7 +4,7 @@ permissions:
 
 on:
   push:
-    branches: [main]
+    branches: [main, proposer-rs-retry]
```
davidtaikocha, line 7:
```suggestion
    branches: [main]
```

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -172,32 +172,85 @@ impl Proposer {
...
+                let bumped_priority_fee = fee_estimate.max_priority_fee_per_gas * multiplier / 100;
+                let bumped_max_fee = fee_estimate.max_fee_per_gas * multiplier / 100;
+
+                current_request = current_request
+                    .max_priority_fee_per_gas(bumped_priority_fee)
+                    .max_fee_per_gas(bumped_max_fee);
+            }
+
+            let pending_tx =
+                self.rpc_provider.l1_provider.send_transaction(current_request.clone()).await?;
```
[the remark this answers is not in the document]
davidtaikocha: fixed in https://github.com/taikoxyz/taiko-mono/pull/21496/changes/51094cc35787ce0950542a356f15d90c9e42509d

davidtaikocha:
@claude plz review.

.github/workflows/taiko-client-rs--docker.yml
davidtaikocha, line 7:
```suggestion
    branches: [main]
```

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -172,32 +172,147 @@ impl Proposer {
...
+                warn!(
+                    attempt,
+                    tip_multiplier_pct = tip_multiplier,
+                    nonce,
+                    "receipt timeout, resubmitting with bumped tip"
+                );
+
+                // Get current fee estimates, apply bump, and enforce minimum floor.
+                // Blob txs require >= 2x blob fee bump per geth's blobPriceBump.
+                let fee_estimate = self.rpc_provider.l1_provider.estimate_eip1559_fees().await?;
```
MatusKysel: Using a fresh estimate_eip1559_fees() as the base for every retry can make the replacement fees go down when network conditions cool off. Replacement rules compare against the last submitted tx, not the latest estimate, so bumped_priority_fee / bumped_max_fee can still be lower than the tx already in mempool and we end up looping on underpriced. The bump should be computed from the previous request values, or at least clamped so every retry is strictly above the last sent fees.

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -172,32 +172,147 @@ impl Proposer {
...
+                }
+            }
         }
 
+        error!(
+            max_retries = self.cfg.max_tip_bump_retries,
+            "proposal transaction not mined after all tip bump retries"
+        );
+        counter!(ProposerMetrics::PROPOSALS_FAILED).increment(1);
         Ok(())
```
MatusKysel: shouldn't we return error here ?

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -172,32 +172,147 @@ impl Proposer {
...
+                // Bump the priority fee for replacement transaction.
+                tip_multiplier += self.cfg.tip_bump_percentage;
+                warn!(
+                    attempt,
+                    tip_multiplier_pct = tip_multiplier,
+                    nonce,
+                    "receipt timeout, resubmitting with bumped tip"
+                );
+
+                // Get current fee estimates, apply bump, and enforce minimum floor.
```
MatusKysel: The fee-bump calculation is duplicated in the timeout path and again in the underpriced error path. I’d strongly prefer extracting one helper that takes the previous fees plus the latest floor and produces the next request.

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -172,32 +172,147 @@ impl Proposer {
...
+            let pending_tx = match self
+                .rpc_provider
+                .l1_provider
+                .send_transaction(current_request.clone())
+                .await
+            {
+                Ok(tx) => tx,
+                Err(err) => {
+                    let err_str = err.to_string();
+                    if err_str.contains("replacement transaction underpriced") ||
```
MatusKysel: AI reported -> Nethermind reports the replacement case as ReplacementNotAllowed instead of replacement transaction underpriced

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -172,32 +172,147 @@ impl Proposer {
...
+                    tip_multiplier_pct = tip_multiplier,
+                    nonce,
+                    "receipt timeout, resubmitting with bumped tip"
+                );
+
+                // Get current fee estimates, apply bump, and enforce minimum floor.
+                // Blob txs require >= 2x blob fee bump per geth's blobPriceBump.
+                let fee_estimate = self.rpc_provider.l1_provider.estimate_eip1559_fees().await?;
+                let multiplier = tip_multiplier as u128;
+                blob_fee_multiplier += 100; // +100% each retry (geth requires >= 2x)
```
MatusKysel: as we are increasing it here and also in lower part, if the RPC replies underprice we will bump it 2x

packages/taiko-client-rs/bin/client/src/flags/proposer.rs
```diff
@@ -45,4 +45,44 @@ pub struct ProposerArgs {
...
+    )]
+    pub min_max_fee_per_gas_gwei: u64,
+    /// Minimum priority fee per gas in gwei.
+    #[clap(
+        long = "propose.minPriorityFeePerGas",
+        env = "PROPOSE_MIN_PRIORITY_FEE_PER_GAS",
+        default_value = "1",
+        help = "Minimum priority fee per gas in gwei"
+    )]
+    pub min_priority_fee_per_gas_gwei: u64,
```
YoGhurt111, line 87: Does this mean we cannot set a tip lower than 1 gwei?

davidtaikocha: default is `1` gwei, but it can be overridden to `0`.

davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/pull/21513

PR #21501, opened 2026-03-25
refactor(taiko-client-rs): use AtomicU64 for highest unsafe id

edgarr1986:
Replace `Arc<Mutex<u64>>` with `Arc<AtomicU64>` for highest unsafe payload tracking, preserving existing store/max semantics while removing async lock contention.


MatusKysel:
@claude @codex review

davidtaikocha:
Can't see too many improvements here.

PR #21505, opened 2026-03-25
fix(taiko-client-rs): retry timed-out preconf submissions and reconnect event sync

MatusKysel:
## Summary

 This hardens the embedded `taiko-client-rs` preconfirmation flow against transient `reth` stalls and L1 WS interruptions.

  - reconnect the event syncer instead of exiting when the event scanner init/start/stream ends
  - retry timed-out preconfirmation submissions on `PreconfResponseTimeout` with bounded backoff
  - make retries idempotent by skipping already materialized preconfirmation payloads before enqueue and again in the ingress loop
  - add unit coverage for the preconfirmation timeout retry helper

  ## Why

During proposal catch-up, `reth` can stall on MDBX write-lock contention long enough for preconfirmation submissions to exceed the current response timeout. Separately, L1 WS interruptions were causing the event syncer to exit instead of recovering.

These changes make the driver tolerate both transient failure modes without crashing.

### DIFF
--- a/packages/taiko-client-rs/crates/driver/src/sync/event.rs
+++ b/packages/taiko-client-rs/crates/driver/src/sync/event.rs
@@ -90,9 +90,12 @@ struct EventStreamStartPoint {
 fn should_probe_confirmed_sync(
     preconfirmation_enabled: bool,
     preconf_ingress_spawned: bool,
+    preconf_ingress_ready: bool,
     scanner_live: bool,
 ) -> bool {
-    preconfirmation_enabled && !preconf_ingress_spawned && scanner_live
+    preconfirmation_enabled
+        && scanner_live
+        && (!preconf_ingress_spawned || !preconf_ingress_ready)
 }
 
 /// Resolve whether confirmed-sync readiness should open ingress.
@@ -1240,6 +1243,7 @@ where
                 if should_probe_confirmed_sync(
                     self.cfg.preconfirmation_enabled,
                     preconf_ingress_spawned,
+                    self.preconf_ingress_ready.load(Ordering::Acquire),
                     scanner_live,
                 ) {
                     let confirmed_sync_probe = self.confirmed_sync_snapshot().await;
@@ -1268,9 +1272,20 @@ where
                                 self.preconf_ingress_notify.clone(),
                             );
                             preconf_ingress_spawned = true;
+                        } else if !self.preconf_ingress_ready.swap(true, Ordering::AcqRel) {
+                            info!(
+                                "re-opened preconfirmation ingress after scanner reconnect"
+                            );
+                            self.preconf_ingress_notify.notify_waiters();
                         }
                     }
                 }
+            }
+
+            // A dropped scanner forces a historical replay window again, so close ingress until
+            // the next live scanner transition and confirmed-sync probe re-open it.
+            if self.preconf_ingress_ready.swap(false, Ordering::AcqRel) {
+                info!("closing preconfirmation ingress during event scanner reconnect");
             }
 
             if let Some(block_number) = last_seen_l1_block_number {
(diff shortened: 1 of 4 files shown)


MatusKysel:
@claude re-review

YoGhurt111:
I noticed that on Hoodi, there’s an event scanner–related error that can cause the driver to stall. Is this related to this PR?

MatusKysel:
> I noticed that on Hoodi, there’s an event scanner–related error that can cause the driver to stall. Is this related to this PR?

If the Hoodi issue is an event-scanner init/start/stream failure in driver::sync::event, then yes, that is directly in scope of this PR

packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -1016,85 +1144,161 @@ where
...
-                Ok(ScannerMessage::Notification(notification)) => {
-                    info!(?notification, "event scanner notification");
-                    if matches!(notification, Notification::SwitchingToLive) {
-                        // Scanner live is necessary but not sufficient: confirmed-sync readiness
-                        // must also pass before ingress opens.
-                        scanner_live = true;
-                    }
-                }
+        loop {
+            if !preconf_ingress_spawned {
```
YoGhurt111, line 1157: This reconnect loop resets `scanner_live`, but it never clears `preconf_ingress_ready` once the ingress task has been spawned. That means `submit_preconfirmation_payload_with_timeout()` will continue accepting preconfirmations during a scanner reconnect / historical replay window because its gate only checks `preconf_ingress_ready`.

In other words, the new comment says reconnect should wait for a fresh `SwitchingToLive` before ingress re-opens, but the actual ingress gate stays open forever after the first successful spawn. Since the ingress loop keeps running independently, preconfirmations can still be injected against stale canonical state while the event scanner is catching back up.

If the intended behavior is to re-close ingress on reconnect, I think this needs to clear/re-arm the readiness flag as part of the reconnect cycle, not just reset `scanner_live`.

packages/taiko-client-rs/crates/test-harness/src/preconfirmation/driver.rs
```diff
@@ -192,7 +183,7 @@ where
 {
     /// Create a new client backed by the driver event syncer.
     pub fn new(event_syncer: Arc<EventSyncer<P>>, client: Client<P>) -> Self {
-        Self { event_syncer, inbox: client.shasta.inbox, l2_provider: client.l2_provider }
+        Self { inner: RuntimeEventSyncerDriverClient::from_client(event_syncer, client) }
```
YoGhurt111: This test-harness wrapper now delegates to `RuntimeEventSyncerDriverClient::from_client()`, which hardcodes `DEFAULT_WAIT_EVENT_SYNC_POLL_INTERVAL = 12s`. Before this refactor the harness implementation polled every 200ms, so `wait_event_sync()` in E2E paths can now block for up to an extra ~12s even after sync is already ready.

That looks like an accidental behavior regression for tests rather than an intentional harness change. If the reuse is desirable, I think the harness should still construct the runtime client with a much shorter poll interval to preserve previous test latency / flakiness characteristics.

packages/taiko-client-rs/crates/driver/src/sync/event.rs
MatusKysel: [6cc554e](https://github.com/taikoxyz/taiko-mono/pull/21505/commits/6cc554e4f980796e45840b8e408611e4e160d7e5) fixed

packages/taiko-client-rs/crates/test-harness/src/preconfirmation/driver.rs
MatusKysel: fixed [4e03a98](https://github.com/taikoxyz/taiko-mono/pull/21505/commits/4e03a981a868ebe50e248f819ba5c96b3a246b53)

YoGhurt111 (approved):

davidtaikocha (approved):

PR #21506, opened 2026-03-25
fix(taiko-client): guard nil Shasta LastBlockIDByBatchID

odupontt:
Guard nil LastBlockIDByBatchID results across Shasta recovery paths, preventing panics when taiko-geth returns no batch mapping.


MatusKysel:
@odupontt closing this as this is impossible as function ALWAYS return error or block num

PR #21510, opened 2026-03-27
feat(taiko-client): add p2p allow-all-sequencers flag

MatusKysel:


MatusKysel:
@claude re-review

packages/taiko-client/driver/preconf_blocks/server.go
```diff
@@ -177,6 +180,22 @@ func (s *PreconfBlockAPIServer) SetP2PSigner(p2pSigner p2p.Signer) {
...
+
+	s.allowAllSequencers = allow
+}
+
+// UpdateActiveSequencerAddresses updates the cached active sequencer addresses used by P2P validation.
+func (s *PreconfBlockAPIServer) UpdateActiveSequencerAddresses(addresses []common.Address) {
+	s.lookaheadMutex.Lock()
+	defer s.lookaheadMutex.Unlock()
+
+	s.allActiveSequencerAddresses = append([]common.Address(nil), addresses...)
```
davidtaikocha, line 196: If the L1 node is very laggy, will the `ActiveSequencerAddresses` also be wrong, which then leads to the same issue?

MatusKysel: yeah, I could remove that but what I saw happing on testnet is current/next can be wrong every epoch handover because laggy L1. So if we are ok going in less safe option I would even remove check for active ...

davidtaikocha (approved):

YoGhurt111 (approved):

PR #21511, opened 2026-03-27
refactor(bridge-ui): deduplicate bridged token state logic in checkBalanceToBridge

anim001k:
- extract duplicated bridged token state logic from `checkBalanceToBridge`
- reuse a single helper in both ERC20 and ERC1155 paths
- preserve the existing early-return and `isTokenAlreadyDeployed` behavior


davidtaikocha:
Closed due to being outdated.

PR #21512, opened 2026-03-27
fix(taiko-client): tolerate transient L1 sync read failures

MatusKysel:
## Summary

  This PR makes preconfirmation sync gating resilient to transient L1 read failures in both `taiko-client` and `taiko-client-rs`.

  Previously, gossip/block import could stall if the driver briefly lost access to L1 while checking confirmed sync progress. That made L1 WS instability a liveness risk for local preconfirmation handling.

  ## What Changed

  - Keep local engine/auth checks hard-blocking.
  - Treat L1-derived sync target reads as best-effort and cached.
  - In `taiko-client`, refresh and cache the highest relevant L1 origin in the background and use that cache on transient L1 failures.
  - In `taiko-client-rs`, split the confirmed-sync probe into:
    - `next_proposal_id()` from L1
    - local confirmed-sync evaluation for a provided target proposal id
  - Cache the last successful L1 proposal target and reuse it during transient L1 failures.
  - Do not mask local engine/auth failures with cached fallback behavior.

  ## Why

  This preserves the intended safety boundary:

  - Local execution/auth state remains authoritative and must be healthy.
  - Temporary L1 WS/RPC issues no longer cause the driver to reject or stall otherwise valid gossip/import work.

### DIFF
--- a/packages/taiko-client/pkg/rpc/methods.go
+++ b/packages/taiko-client/pkg/rpc/methods.go
@@ -582,6 +582,61 @@ func (p *L2SyncProgress) IsSyncing() bool {
 	return p.CurrentBlockID.Cmp(p.HighestOriginBlockID) < 0
 }
 
+func latestPacayaBatchID(numBatches uint64) (uint64, bool) {
+	if numBatches == 0 {
+		return 0, false
+	}
+
+	return numBatches - 1, true
+}
+
+// fetchHighestOriginBlockIDFromContracts resolves the latest L1-derived origin block ID used by
+// both L2ExecutionEngineSyncProgress and FetchHighestOriginBlockIDFromL1 so the Shasta/Pacaya
+// dispatch and edge-case handling stay consistent.
+func (c *Client) fetchHighestOriginBlockIDFromContracts(ctx context.Context) (*big.Int, error) {
+	coreState, err := c.GetCoreStateShasta(&bind.CallOpts{Context: ctx})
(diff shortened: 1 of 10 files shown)


MatusKysel:
@claude re-review

packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -662,6 +684,50 @@ func (c *Client) L2ExecutionEngineSyncProgress(ctx context.Context) (*L2SyncProg
 	return progress, nil
 }
 
+// L2ExecutionEngineSyncProgressLocal fetches only the local L2 sync state (no L1 calls).
+// This is safe to call even when L1 is unreachable.
+func (c *Client) L2ExecutionEngineSyncProgressLocal(ctx context.Context) (*L2SyncProgress, error) {
```
davidtaikocha, line 689: Just thinking out loud while reviewing this and [that PR](https://github.com/taikoxyz/taiko-mono/pull/21510), maybe instead of introducing these patches / work-arounds and starting to add the extra layer of complexity, shall we take more time to investigate the root cause of why the L1 RPC was laggy? And making a reliable L1 RPC is an assumption of the client software? And the client will fail if the upstream L1 is not working as expected? (And seems other clients like [op-rust](https://github.com/ethereum-optimism/optimism/tree/develop/rust) are also assuming that, according to codex investigation)

PR #21513, opened 2026-03-28
feat(taiko-client-rs): adopt `tx-manager` for proposal submission

davidtaikocha:
To replace: https://github.com/taikoxyz/taiko-mono/pull/21496

## Summary

This PR switches proposer transaction submission to `tx-manager` while keeping proposal construction in the proposer crate.

It also fixes manifest gas-limit derivation for proposal transactions and preserves proposer-level handling for bounded retry exhaustion versus confirmed receipts.

## Changes

- adopt `tx-manager` as the backend for proposer transaction publishing
- introduce a proposer-owned built-transaction boundary and tx-manager adapter
- map proposer retry and fee-floor config into `TxManagerConfig`
- keep proposer success semantics tied to confirmed receipts
- keep retry exhaustion non-fatal to the outer proposer loop
- remove the legacy proposer-managed submission retry path
- fix manifest gas-limit derivation to use the effective parent path
- handle empty genesis `extra_data` in engine mode when deriving the first proposal id
- pin `base-tx-manager` to `base/base@1f1149cc` because newer upstream blob-sidecar behavior is incompatible with the current Shasta test environment

## Why

The proposer should not own its own transaction lifecycle state machine. `tx-manager` already handles nonce management, resubmission, fee bumping, and receipt polling, so this change moves that responsibility to the shared upstream component and keeps proposer focused on building proposal transactions and interpreting submission outcomes.

The pinned tx-manager revision is intentional. A newer upstream revision changed blob-sidecar handling in a way that caused Shasta publish failures in the current harness.

## Testing

- `just fmt`
- `just clippy-fix`
- `cargo test -p proposer --lib -- --nocapture`
- `just test`

## Results

- proposer unit tests pass
- `propose_shasta_batches` passes
- `propose_shasta_batches_engine_mode` passes
- full workspace test suite passes

## Notes

- proposer keeps a small proposer-specific config surface and translates it internally to tx-manager config
- `receipt_query_interval` remains optional; `None` uses the upstream tx-manager default, while tests can override it for faster harness feedback
- driver integration tests were updated to consume the new `BuiltProposalTx` boundary


davidtaikocha:
@claude plz review.

.github/workflows/taiko-client-rs--docker.yml
```diff
@@ -4,7 +4,7 @@ permissions:
 
 on:
   push:
-    branches: [main]
+    branches: [main, proposer-retry-2]
```
davidtaikocha:
```suggestion
    branches: [main]
```

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -135,15 +158,17 @@ impl Proposer {
             interval.tick().await;
             info!(epoch, "proposer epoch");
 
-            self.fetch_and_propose().await?;
+            if matches!(self.fetch_and_propose().await?, ProposalSendOutcome::RetryExhausted) {
```
MatusKysel: This still exits the proposer on any surfaced tx-manager error because of the ? here. That means NonceTooLow, FeeLimitExceeded, and provider/RPC failures currently terminate the loop even though they are operational conditions rather than startup/config failures. Can we match on fetch_and_propose() here and only return for truly fatal variants like invalid config/signing errors, while logging and continuing for the rest?

packages/taiko-client-rs/crates/proposer/src/tx_manager_adapter.rs
```diff
@@ -0,0 +1,509 @@
...
+    match result {
+        Ok(receipt) => Ok(ProposalSendOutcome::ConfirmedReceipt { receipt: Box::new(receipt) }),
+        Err(err) if is_retry_exhausted(&err) => Ok(ProposalSendOutcome::RetryExhausted),
+        Err(err) => Err(err.into()),
+    }
+}
+
+/// Return `true` when the tx-manager exhausted a bounded retry path and the outer proposer loop
+/// should continue.
+fn is_retry_exhausted(err: &TxManagerError) -> bool {
```
MatusKysel: The non-fatal classification still looks too narrow. is_retry_exhausted() only covers SendTimeout and MempoolDeadlineExpired, but the comment in proposer.rs says retry exhaustion also covers repeated pre-publish RPC failures. Right now TxManagerError::Rpc(_) still bubbles out as a hard error and trips the ? in start(). Either broaden the classification or tighten the comment/expected behavior so this is deliberate.

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -253,34 +252,33 @@ impl Proposer {
...
+        self.calculate_next_shasta_block_base_fee_for_parent(&parent).await
+    }
 
-        // Calculate the parent block time by subtracting its timestamp from its parent's timestamp.
+    /// Calculate the base fee for the next L2 block from a specific parent snapshot.
+    async fn calculate_next_shasta_block_base_fee_for_parent(
+        &self,
+        parent: &Block,
+    ) -> Result<U256> {
         let parent_number = parent.number();
```
MatusKysel, line 262: This still is not a fully fixed snapshot boundary. We freeze parent, but grandparent is fetched later by block number from the live canonical chain. If a reorg lands between those reads, we can combine a parent from the old branch with a grandparent from the new branch and compute the wrong EIP-4396 base fee. Can we capture the grandparent in the same snapshot flow, or at least validate the hash linkage, so the base-fee calculation stays on one branch?

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -521,4 +615,146 @@ mod tests {
...
+                Some(&grandparent),
+                1_000_000_000
+            )
+            .expect("parent snapshot should determine the next base fee"),
+            expected
+        );
+    }
+
+    /// Build a minimal receipt with the requested execution status.
+    fn receipt_with_status(success: bool, tx_hash: B256) -> TransactionReceipt {
```
MatusKysel: receipt_with_status() is duplicated verbatim across two test modules(tx_manager_adapter.rs:430). Can we move this into a shared #[cfg(test)] helper instead of maintaining two identical receipt builders?

packages/taiko-client-rs/crates/proposer/src/transaction_builder.rs
```diff
@@ -113,18 +215,388 @@ impl ShastaProposalTransactionBuilder {
...
+    };
+
+    impl BuiltProposalTx {
+        /// Return the internal blob payload for crate-local tests.
+        pub(crate) fn blob_payload(&self) -> &ProposalBlobPayload {
+            &self.blob_payload
+        }
+    }
+
+    impl ProposalBlobPayload {
```
MatusKysel: The test-only blob-payload construction is duplicated in two places

packages/taiko-client-rs/crates/proposer/src/transaction_builder.rs
```diff
@@ -113,18 +215,388 @@ impl ShastaProposalTransactionBuilder {
...
+        assert_eq!(sidecar.as_eip4844().expect("expected EIP-4844 sidecar").blobs.len(), 1);
+        assert_eq!(request.gas, Some(210_000));
+        assert_eq!(
+            request.blob_sidecar().expect("transaction request should retain sidecar").blobs.len(),
+            1
+        );
+    }
+
+    #[derive(Clone, Debug)]
+    struct ManifestTestTransport {
```
MatusKysel, line 309: This inline ManifestTestTransport harness is getting pretty heavy for a production module. It adds a lot of bespoke JSON-RPC mocking machinery to transaction_builder.rs, which makes the proposal-building code harder to scan. Could we move this into a dedicated test helper module so the production file stays focused on builder logic?

packages/taiko-client-rs/crates/proposer/src/transaction_builder.rs
```diff
@@ -25,6 +25,104 @@ use crate::{
...
+        &self.call_data
+    }
+
+    /// Return the optional gas-limit override that should be applied on submission.
+    pub fn gas_limit(&self) -> Option<u64> {
+        self.gas_limit
+    }
+
+    /// Return the number of blobs carried by this proposal transaction.
+    pub fn blob_count(&self) -> usize {
```
MatusKysel: does not appear to have any production callers right now

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -501,16 +498,113 @@ impl Proposer {
...
+                counter!(ProposerMetrics::PROPOSALS_SUCCESS).increment(1);
+
+                // Record gas used once the confirmed receipt shows successful execution.
+                histogram!(ProposerMetrics::GAS_USED).record(receipt.gas_used as f64);
+            } else {
+                error!(tx_hash = %receipt.transaction_hash, "proposal transaction failed");
+                counter!(ProposerMetrics::PROPOSALS_FAILED).increment(1);
+            }
+        }
+        ProposalSendOutcome::RetryExhausted => {
```
MatusKysel: RetryExhausted and confirmed reverted receipts both increment PROPOSALS_FAILED. A separate retry-exhausted counter would make it much easier to distinguish “never got confirmed” from “confirmed on-chain but reverted”

packages/taiko-client-rs/crates/proposer/src/proposer.rs
davidtaikocha: simplified: https://github.com/taikoxyz/taiko-mono/pull/21513/changes/51ac61fb0631f88b95f08c6d865e9ef7a29ac7d0

packages/taiko-client-rs/crates/proposer/src/tx_manager_adapter.rs
davidtaikocha: simplified: https://github.com/taikoxyz/taiko-mono/pull/21513/changes/51ac61fb0631f88b95f08c6d865e9ef7a29ac7d0

packages/taiko-client-rs/crates/proposer/src/proposer.rs
davidtaikocha: changed to use `parent_hash`: https://github.com/taikoxyz/taiko-mono/pull/21513/commits/00b97176db969d4fe9dcfa063861bc82aa0fedc1

packages/taiko-client-rs/crates/proposer/src/proposer.rs
davidtaikocha: removed duplicated `receipt_with_status()`

packages/taiko-client-rs/crates/proposer/src/transaction_builder.rs
davidtaikocha: updated: https://github.com/taikoxyz/taiko-mono/pull/21513/changes/d73721b6a4572bcecc93a87d9fd741a2736c8f09

packages/taiko-client-rs/crates/proposer/src/transaction_builder.rs
davidtaikocha: IMO it's fine here since it only in tests and the current proposer is pretty light-weight and only will be used in devnet.

packages/taiko-client-rs/crates/proposer/src/transaction_builder.rs
davidtaikocha: updated: https://github.com/taikoxyz/taiko-mono/pull/21513/changes/a34b10b6f402745d795b42c250c9ab4726733873

packages/taiko-client-rs/crates/proposer/src/proposer.rs
davidtaikocha: simplified in https://github.com/taikoxyz/taiko-mono/pull/21513/changes/51ac61fb0631f88b95f08c6d865e9ef7a29ac7d0 as well, and rn the distinguishing between`confirmed success` and `confirmed revert` is fine since it's used for devnet atm.

packages/taiko-client-rs/crates/proposer/src/proposer.rs
```diff
@@ -501,16 +497,112 @@ impl Proposer {
...
+        return Ok(1);
+    }
+
+    decode_shasta_proposal_id(parent_extra_data)
+        .map(|proposal_id| proposal_id + 1)
+        .ok_or(ProposerError::InvalidExtraData)
+}
+
+/// Return `true` when a surfaced proposer error should be logged and retried on the next epoch.
+fn is_operational_submission_error(err: &ProposerError) -> bool {
```
YoGhurt111, line 573: Should we switch to retrying only a limited set of retryable errors, and not retry the rest, instead of the current approach where only a few errors are excluded from retries?

davidtaikocha: updated: https://github.com/taikoxyz/taiko-mono/pull/21513/changes/2fa1b6db55b33ce7673ebae166310865f2d4c789

MatusKysel (approved):

YoGhurt111 (approved):

PR #21517, opened 2026-03-30
fix(taiko-client): avoid errgroup err data races

edgarr1986:
Use goroutine-local errors in errgroup branches to remove shared err races.


davidtaikocha:
can't see too many improvements.

PR #21528, opened 2026-04-01
docs(preconfirmation-p2p): sync net docs with current behavior

odupontt:
Sync the preconfirmation-p2p docs with the current networking implementation, removing stale feature references and correcting the rate-limit and event-delivery descriptions.


davidtaikocha:
can't see real improvements here.

PR #21549, opened 2026-04-09
taiko-client: reject overflowing Shasta manifest headers

odupontt:
Add strict range checks for Shasta manifest version and size fields, and add overflow tests to prevent silent truncation paths.


davidtaikocha:
Can't see much improvements since we do have `version := new(big.Int).SetBytes(data[offset : offset+32])`

PR #21550, opened 2026-04-09
feat(bridge-ui): add manual claim entry for none fee transfers

YoGhurt111:
## Summary

This change improves the bridge-ui manual-claim flow for transactions created with the `None` processing fee option.

When a user selects `None`, the processing fee modal now explains that the transfer must be claimed manually and links them to the Transactions page, where they can manage their transfer later.

The Transactions list now also exposes a `Try claim` action for zero-fee transfers that are still shown as `Processing`. This gives users an explicit way to open the existing claim dialog and attempt a manual claim, instead of waiting for the status to become claimable in the UI. If the message is not ready yet, the existing claim dialog error handling still explains that state.

## Root Cause

Previously, the `None` option only warned users that no relayer would claim on their behalf, but it did not provide an actionable path to continue. In addition, zero-fee transfers that remained in the `Processing` state had no direct manual-claim entry in the Transactions list, even though the user intent was to self-claim.

## Fix

- move the `None` option warning UI into a dedicated component path that can render a clear manual-claim CTA
- add a Transactions-page link for the `None` option warning state
- add a `Try claim` action in transaction status rendering when the transfer is `NEW`, still not processable, and has a zero processing fee
- keep existing claim dialog behavior so early attempts still surface the correct on-chain error state

## Validation

- `pnpm exec vitest run src/components/Bridge/SharedBridgeComponents/ProcessingFee/noneOption.test.ts src/components/Transactions/Status/status.test.ts`


YoGhurt111:
@claude Plz review again

PR #21555, opened 2026-04-11
feat(bridge-ui): bridge UI manual claim entry

YoGhurt111:


davidtaikocha (approved):

packages/bridge-ui/src/components/Transactions/Status/Status.svelte
```diff
@@ -71,6 +71,14 @@
     dispatch('openModal', 'claim');
   }
 
+  async function handleTryClaimClick() {
+    isBridgePaused().then((paused) => {
+      if (paused) throw new BridgePausedError('Bridge is paused');
+    });
```
kimo-ice: **Bug: pause check is non-blocking.**

```ts
async function handleTryClaimClick() {
  isBridgePaused().then((paused) => {
    if (paused) throw new BridgePausedError('Bridge is paused');
  });
  if (!$connectedSourceChain || !$account?.address) return;
  dispatch('openModal', 'try_claim');
}
```

The promise isn't awaited, so `throw` inside the `.then()` callback becomes an unhandled rejection — it never propagates to the caller. Execution falls through to `dispatch('openModal', 'try_claim')` regardless of pause state, so users can open the try-claim modal while the bridge is paused.

Suggested fix (matches the pattern used by `handleClaimClick` above):

```ts
async function handleTryClaimClick() {
  if (await isBridgePaused()) throw new BridgePausedError('Bridge is paused');
  if (!$connectedSourceChain || !$account?.address) return;
  dispatch('openModal', 'try_claim');
}
```

kimo-ice (changes requested):
Nice feature — the `try_claim` mode and error-normalization helper are clean and well-tested. Flagging one bug in the pause gate before approval; a couple of small nits in a separate comment.

packages/bridge-ui/src/components/Transactions/Status/Status.svelte
YoGhurt111: Fixed. The pause check now blocks before opening the modal, and I applied the same fix to the other status actions in this component (`retry`, `release`, `claim`) since they had the same fire-and-forget pattern. Included in c3159873bc45b351c1fb5d538e036ab9dcc37bf3.

kimo-ice (approved):

RogerLamTd (approved):

PR #21575, opened 2026-04-20
fix(bridge-ui): fall back to 500k gas when ERC20 sendToken estimation fails

dantaik:
## Summary

- Wrap `ERC20Bridge.estimateGas`'s call to `tokenVaultContract.estimateGas.sendToken` in a try/catch so a failed RPC estimation no longer bubbles up and breaks `estimateCostOfBridging` / the UI cost display.
- On failure, return `500_000n` as a safe default — matches the pattern already used in `Bridge.ts` for `processMessage` (which falls back to `1_300_000n`).

## Test plan

- [x] Bridge an ERC20 (e.g. USDC) from L2 to L1 on a network where gas estimation succeeds — cost display and broadcast behave identically to before.
- [x] Force an estimation failure (e.g. temporarily pointing at an RPC that rejects the call) — UI shows a cost derived from 500k gas instead of throwing.

### DIFF
--- a/packages/bridge-ui/src/app.config.ts
+++ b/packages/bridge-ui/src/app.config.ts
@@ -3,6 +3,7 @@ export const gasLimitConfig = {
   ethGasLimit: 100_000,
   erc20NotDeployedGasLimit: 750_000,
   erc20DeployedGasLimit: 500_000,
+  erc20SendTokenFallbackGasLimit: 500_000n, // used when eth_estimateGas for sendToken fails
   erc721NotDeployedGasLimit: 2_400_000,
   erc721DeployedGasLimit: 1_100_000,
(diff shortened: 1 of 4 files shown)


dantaik:
I took me a while before I can bridge USDC from L2 to L1 before this enhancement. Somehow gas estimation failed in my wallet and the default gas limit is too small for bridging USDC from L2 to L1.

YoGhurt111 (approved):

ggonzalez94 (approved):
Mostly looks good to me. I think there's a few cases we should think about(can be shipped in a separate PR).
- if the token is not deployed we need more gas(~750k gas). This would make this fix fail under this edge case.
- we don't have a similar fix for eth, erc721 and erc1155 bridges. This creates an inconsistent experience for people using the bridge

PR #21581, opened 2026-04-21
fix(bridge-ui): per-token fallback gas + extra headroom for undeployed tokens

dantaik:
## Summary
- Generalizes the sendToken gas-estimation helper into estimateBridgeGas (shared across ETH, ERC20, ERC721, ERC1155) with a caller-supplied fallback.
- Adds per-token-type fallback gas limits and a bridgeTxNotDeployedExtraGas headroom applied when the destination token isn't deployed yet.
- Passes an explicit gas value into simulateContract/writeContract in all four bridge flows so the fallback actually takes effect when estimation fails.

## Test plan
- [x] pnpm --filter @taiko/bridge-ui test (estimateBridgeGas unit tests cover success, rejection, non-Error rejection, number/bigint fallback coercion, deployed vs undeployed scaling)
- [x] Manual bridge of ETH, ERC20, ERC721, ERC1155 (both first-time/undeployed and already-deployed destinations)


YoGhurt111:
I found one blocking issue in the new fallback gas logic.

`bridgeTxNotDeployedExtraGas` is now being added inside `tokenBridgeFallbackGas(...)` and then reused as the fallback gas for the source-chain `sendToken` transaction. That seems to mix two different concerns:

- the destination message `gasLimit` / processing-fee headroom when the token is not yet deployed on the destination chain
- the source transaction gas limit used only when `eth_estimateGas` fails

The first one is already accounted for in `_prepareTransaction()` via `erc20NotDeployedGasLimit` / `erc721NotDeployedGasLimit` / `erc1155NotDeployedGasLimit`, so adding `bridgeTxNotDeployedExtraGas` again to the source tx fallback looks like double-counting the undeployed penalty.

In practice this will systematically overestimate the source tx cost in `estimateCostOfBridging()`, and it will also inflate the wallet-submitted `gas` value even though the source `sendToken` tx itself should not become ~750k more expensive just because the destination token is undeployed.

I think the token-type fallback helper should probably only return the per-token source fallback baseline, while the undeployed extra headroom remains only on the destination message `gasLimit` / fee side.

dantaik:
> I found one blocking issue in the new fallback gas logic.
> 
> `bridgeTxNotDeployedExtraGas` is now being added inside `tokenBridgeFallbackGas(...)` and then reused as the fallback gas for the source-chain `sendToken` transaction. That seems to mix two different concerns:
> 
> * the destination message `gasLimit` / processing-fee headroom when the token is not yet deployed on the destination chain
> * the source transaction gas limit used only when `eth_estimateGas` fails
> 
> The first one is already accounted for in `_prepareTransaction()` via `erc20NotDeployedGasLimit` / `erc721NotDeployedGasLimit` / `erc1155NotDeployedGasLimit`, so adding `bridgeTxNotDeployedExtraGas` again to the source tx fallback looks like double-counting the undeployed penalty.
> 
> In practice this will systematically overestimate the source tx cost in `estimateCostOfBridging()`, and it will also inflate the wallet-submitted `gas` value even though the source `sendToken` tx itself should not become ~750k more expensive just because the destination token is undeployed.
> 
> I think the token-type fallback helper should probably only return the per-token source fallback baseline, while the undeployed extra headroom remains only on the destination message `gasLimit` / fee side.

Sharp eyes. Please take another look.

YoGhurt111 (approved):

PR #21593, opened 2026-04-23
fix(taiko-client): refresh cached L2 head before beacon-sync decision

davidtaikocha:
## Summary

Fixes a deadlock in the driver's chain syncer where a newly-provisioned L2 execution engine that has already caught up to the checkpoint-URL's head (via its own L2 p2p / discv5) causes the driver to repeatedly re-trigger beacon sync and fatally error on every cycle with \`unexpected ForkchoiceUpdate response status: VALID\`. As a result event sync never runs, \`HeadL1Origin\` never gets written, and the preconfirmation block server stays \`NotReady\` indefinitely — every incoming preconf payload is cached instead of inserted.

## Root cause

Two compounding bugs in \`driver/chain_syncer\`:

1. **Stale cached \`state.L2Head\`** (this PR). \`driver/state/state.go\` initializes \`l2Head\` once via \`HeaderByNumber(nil)\` and afterward advances it only via the \`newHeads\` WebSocket subscription. That subscription does not fire for blocks the execution engine acquires through engine-API-driven beacon-sync backfill, so the cached head stays frozen at the pre-sync value. \`needNewBeaconSyncTriggered\` → \`AheadOfHeadToSync\` consults the stale cache and keeps deciding "behind the checkpoint head."

2. **Asymmetric engine-API status handling in \`beaconsync/syncer.go\`** (*left out of scope of this PR*). \`NewPayload\` tolerates both \`SYNCING\` and \`VALID\`, but \`ForkchoiceUpdate\` hard-errors on \`VALID\`. When bug #1 drives a spurious re-trigger, the engine legitimately returns \`VALID\` (block already present), and the driver kills the \`Sync()\` pass before \`UpdateMeta\`/\`MarkFinished\`/\`eventSyncer.ProcessL1Blocks\` can run — preventing any state that would let the next iteration recover.

Fixing bug #1 alone is sufficient to break the deadlock: with a fresh cache, \`AheadOfHeadToSync\` correctly returns true once the engine is caught up, \`needNewBeaconSyncTriggered\` returns false, \`TriggerBeaconSync\` is never called, and the \`VALID\` code path is never exercised.

## Change

In \`needNewBeaconSyncTriggered\`, immediately after the existing \`!p2pSync || Finished()\` short-circuit, refresh \`state.L2Head\` from \`rpc.L2.HeaderByNumber(s.ctx, nil)\` and push it into the cache via a new exported \`state.SetL2Head\`. On RPC error we log a warning and fall through with the existing cache (no new failure mode).

The refresh is gated behind the same short-circuit, so the extra RPC is only paid while beacon sync is armed and not yet complete. After \`MarkFinished\` runs the function returns early before hitting the refresh, so there's no steady-state cost.

2 files changed, 19 insertions:
- \`packages/taiko-client/driver/state/state.go\` — +7 lines, adds public \`SetL2Head\` wrapper over the existing private \`setL2Head\` (no rename, preserves existing tests' calls).
- \`packages/taiko-client/driver/chain_syncer/chain_syncer.go\` — +12 lines, adds the refresh block.

No new imports, no new goroutines, no new state. No flag changes, no storage changes, no protocol changes.

## Observed failure (masaya testnet)

Driver configured with \`--p2p.sync\` + \`--p2p.checkPointSyncUrl\` against a synced \`l2-node-archived\`. Local reth had already caught up via discv5 from \`l2ArchivedEnodes\` by the time the driver started.

\`\`\`
Checking whether the execution engine is ahead of the head to sync
  heightToSync=3,344,221 executionEngineHead=3,342,301
L2 execution engine is behind of the head to sync
Mark preconfirmation block server as not ready to insert blocks
ERROR Process new L1 blocks error
  error="trigger beacon sync error: unexpected ForkchoiceUpdate response status: VALID"
\`\`\`

Meanwhile, querying reth directly: \`eth_blockNumber = 0x33075d = 3,344,221\` — reth's actual head matched \`heightToSync\`. The driver's cached \`state.L2Head\` was frozen at \`3,342,301\` (reth's head at the instant \`state.Init\` ran, before the backfill completed).

Through ~12 minutes of logs spanning multiple L1 Proposed events (proposal IDs 20869 → 20878 → 20880 → 20882+), the loop never broke.

## Test plan

- [x] \`make lint\` — 0 issues
- [x] \`make test PACKAGE=driver/chain_syncer/...\` — pass (integration suite, Docker-backed)
- [x] \`go build ./...\` — pass
- [x] Deploy and verify preconf-observer node transitions from \`NotReady\` → \`Ready\` after the refresh fixes the decision loop (will update once rolled out)

## Out of scope (follow-up)

- Accept \`VALID\` in \`ForkchoiceUpdate\` status check in \`beaconsync/syncer.go\` (bug #2 above). Independent minor fix; happy to do in a follow-up PR if maintainers agree.
- Adding a periodic \`state.L2Head\` poll as subscription-miss safety-net.

### DIFF
--- a/.github/workflows/taiko-client--docker.yml
+++ b/.github/workflows/taiko-client--docker.yml
@@ -74,8 +74,7 @@ jobs:
           push: true
           tags: ${{ env.REGISTRY_IMAGE }}
           labels: ${{ steps.meta.outputs.labels }}
-          outputs: type=image,name=${{ env.REGISTRY_IMAGE
-            }},push-by-digest=true,name-canonical=true,push=true
+          outputs: type=image,name=${{ env.REGISTRY_IMAGE }},push-by-digest=true,name-canonical=true,push=true
 
       - name: Export digest
         run: |
--- a/.github/workflows/taiko-client--docker.yml
+++ b/.github/workflows/taiko-client--docker.yml
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main, fix/chain-syncer-stale-l2-head-deadlock]
+    branches: [main]
     tags:
       - "taiko-alethia-client-v*"
     paths:
--- a/.github/workflows/taiko-client--docker.yml
+++ b/.github/workflows/taiko-client--docker.yml
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, fix/chain-syncer-stale-l2-head-deadlock]
     tags:
       - "taiko-alethia-client-v*"
     paths:
@@ -74,7 +74,8 @@ jobs:
           push: true
           tags: ${{ env.REGISTRY_IMAGE }}
           labels: ${{ steps.meta.outputs.labels }}
-          outputs: type=image,name=${{ env.REGISTRY_IMAGE }},push-by-digest=true,name-canonical=true,push=true
+          outputs: type=image,name=${{ env.REGISTRY_IMAGE
+            }},push-by-digest=true,name-canonical=true,push=true
 
       - name: Export digest
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, fix/chain-syncer-stale-l2-head-deadlock]
```
```suggestion
    branches: [main]
```

YoGhurt111 (approved):

PR #21595, opened 2026-04-23
feat(protocol,taiko-client,taiko-client-rs): raise DERIVATION_SOURCE_MAX_BLOCKS to 768

ggonzalez94:
## Summary

Raises the per-derivation-source block limit from **192 → 768**.

At the configured `BLOCK_TIME_TARGET` of 2s, 192 blocks covered one Ethereum epoch (32 × 12s = 384s). Bumping to 768 leaves headroom to drop block time as low as **500ms** while still fitting one proposal per epoch (768 × 0.5s = 384s).

---

> [!NOTE]
> **Medium Risk**
> Touches Shasta derivation/manifest validation and fork gating, which can affect block acceptance and engine payload hashing; incorrect timestamp/chain handling could cause proposal fallback or rejection.
> 
> **Overview**
> Updates Shasta derivation-source manifest validation to use a **timestamp-aware per-source block limit**, keeping the pre-fork cap at `192` and allowing up to `768` blocks once **Unzen** is active.
> 
> Implements this across Rust (`protocol`/`driver`) and Go (`taiko-client`) by adding Unzen-aware constants/helpers, passing source `timestamp`/`chain_id` into manifest decoding, and enforcing the new limit in both manifest fetch/validation and proposer-side proposal sizing. Separately removes the devnet Unzen/Uzen timestamp override flags/OnceLock plumbing and renames remaining fork checks/logging from *Uzen* to *Unzen*, with updated tests/docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f50491f2aede2a3556bd06507fba0849aadc9c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/taiko-client-rs/crates/protocol/src/shasta/constants.rs
```diff
@@ -7,9 +7,9 @@ use alethia_reth_consensus::eip4396::{MAINNET_MIN_BASE_FEE, MIN_BASE_FEE};
 use alloy_eips::eip4844::BYTES_PER_BLOB;
 use alloy_hardforks::ForkCondition;
 
-/// The maximum number of blocks allowed in a proposal. If we assume block time is as
-/// small as one second, 192 blocks will cover an Ethereum epoch.
-pub const DERIVATION_SOURCE_MAX_BLOCKS: usize = 192;
+/// The maximum number of blocks allowed in a proposal. Allows blocks as fast as 500ms
+/// assuming one proposal per epoch.
+pub const DERIVATION_SOURCE_MAX_BLOCKS: usize = 768;
```
davidtaikocha: I think we can't bump the constant here directly, need to use different values before / after Uzen fork timestamp. I can create a PR for that.

davidtaikocha (changes requested):

YoGhurt111:
@smtmfft Is it acceptable for the prover side?

ggonzalez94:
> @smtmfft Is it acceptable for the prover side?

We already decided to go with 1s block times for Unzen(which is 384 per proposal). The extra 2x is to give us room to grow in the future without the need of another fork. So it should be fine

packages/taiko-client-rs/crates/protocol/src/shasta/constants.rs
ggonzalez94: @davidtaikocha I tried to apply a comprehensive fork logic equivalent to shasta + fix the spelling error we had uzen -> unzen. Take a look, and if you think it is in a good direction feel free to continue and apply any necessary changes on top. Otherwise, just take the Derivation update, and apply the rest separately

smtmfft:
> > @smtmfft Is it acceptable for the prover side?
> 
> We already decided to go with 1s block times for Unzen(which is 384 per proposal). The extra 2x is to give us room to grow in the future without the need of another fork. So it should be fine

I think 768 itself is not a big deal, the additional hashes overhead should not be a big problem. However, one concern from the zkgas side is that the hard limit would make each block ~4x smaller, which means we are choosing to have more (faster) blocks rather than more transactions per block. If that’s the intended design, I think it’s fine. But if the goal is to avoid future forks, that may not hold — since any adjustment to the zkgas limit would still require a hard fork.

packages/taiko-client-rs/crates/protocol/src/shasta/constants.rs
davidtaikocha: seems still something needs to be modified, so i opened another PR instead: https://github.com/taikoxyz/taiko-mono/pull/21600 , and ported the derivation doc changes there, plz review.

ggonzalez94:
superseed by #21600

PR #21617, opened 2026-05-01
feat(taiko-client): support HTTP-only L1/L2 endpoints in driver via polling in driver

davidtaikocha:
## Summary

- Adds an HTTP-polling backend behind the existing `Subscribe*` helpers in `pkg/rpc/subscription.go` so the **driver** can run with HTTP-only L1 / L2 RPC endpoints (no `eth_subscribe` required).
- Same external behavior, two backends: `EthClient.IsHTTP()` (set once from URL scheme) routes each subscription to either WS (`event.ResubscribeErr` + `Watch*` / `SubscribeNewHead`) or polling (`time.Ticker` + `HeaderByNumber` / `Filter*`).
- Adds `--l1.http`; reuses existing `--l2.http`. Driver requires exactly one of `--l1.ws` / `--l1.http` (same for L2). Prover & proposer are unchanged — they still require `--l1.ws` / `--l2.ws` (now enforced explicitly in their config files since the cli library no longer marks the flags as required).

## Architecture

```
                    SubscribeChainHead{,Interval} / SubscribeProposed / SubscribeProved
                              (returns event.Subscription)
                                     │
                  ┌──────────────────┴──────────────────┐
                  ▼                                     ▼
        WS backend (existing)                  HTTP polling backend (new)
        client.SubscribeNewHead                ticker → HeaderByNumber(nil)
        Watch{Proposed,Proved}                 ticker → Filter{Proposed,Proved}
```

- L1 head / Proposed / Proved poll at 6s; L2 head polls at 2s (driver passes the cadence via `SubscribeChainHeadInterval`).
- `Filter*` ranges are capped at 1000 blocks per tick to avoid oversized `eth_getLogs` calls during catch-up.
- Cursor bootstraps from current L1 head on first successful read; reorgs (head < cursor) snap the cursor back.

## Test plan

- [x] `PACKAGE=pkg/rpc make test` — full suite green, includes new `TestIsHTTPEndpoint`, `TestPollChainHead_DeliversAndStops`, `TestNextFilterRange`.
- [x] `PACKAGE=driver make test` — full suite green (`TestDriverTestSuite` + new `TestResolveEndpoints`).
- [x] `go build ./...` clean across the package.
- [ ] Run a driver against an HTTP-only L1/L2 in a staging environment and confirm head + Proposed / Proved events flow.

## Notes

- Prover behavior unchanged in this PR (per scope decision); the polling helpers are wired in but `--l1.ws` / `--l2.ws` remain required for prover/proposer. A follow-up can flip the prover flag requirement once HTTP-mode is validated for the driver.
- Dropped events under backpressure log at WARN with the affected block range so an operator can correlate downstream gaps.


packages/taiko-client/pkg/rpc/subscription.go
```diff
@@ -91,3 +130,128 @@ func waitSubErr(ctx context.Context, sub event.Subscription) (event.Subscription
...
+				return nil
+			case <-ticker.C:
+				h, err := client.HeaderByNumber(ctx, nil)
+				if err != nil {
+					log.Warn("pollChainHead: HeaderByNumber failed", "err", err)
+					continue
+				}
+				n := h.Number.Uint64()
+				if n == lastNumber {
+					continue
```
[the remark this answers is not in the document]
davidtaikocha: Fixed in 894d7f879 — `pollChainHead` now tracks `(lastNumber, lastHash)` and delivers whenever either changes, so same-height reorgs are forwarded. Matches `eth_subscribe("newHeads")` semantics from the WS path.

packages/taiko-client/pkg/rpc/subscription.go
```diff
@@ -91,3 +130,128 @@ func waitSubErr(ctx context.Context, sub event.Subscription) (event.Subscription
...
+					default:
+						log.Warn(
+							"pollProved: receiver channel full, dropping (cursor still advances; downstream may miss this event)",
+							"start", start,
+							"end", end,
+						)
+					}
+				}
+				_ = iter.Close()
+				cursor = end
```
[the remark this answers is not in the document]
davidtaikocha: Fixed in 894d7f879 with a small departure from the suggestion. We now check `iter.Error()` after iteration, but still advance the cursor — holding it on a permanent decode error (e.g. one bad log in the range) would deadlock the polling loop. Operator gets a loud warn; canonical recovery happens on the next chain-syncer pass.

packages/taiko-client/driver/state/state.go
```diff
@@ -18,6 +19,17 @@ import (
...
 
+const (
+	// l1HeadPollInterval is the cadence at which the driver polls the L1 head
+	// when running against an HTTP-only L1 endpoint. Tuned well below an L1
+	// slot so the driver picks up new heads within one slot.
+	l1HeadPollInterval = 3 * time.Second
+	// l2HeadPollInterval is the cadence at which the driver polls the L2 head
+	// when running against an HTTP-only L2 endpoint. Set faster than the L2
+	// block time so the driver tracks the L2 chain with sub-block latency.
+	l2HeadPollInterval = 1 * time.Second
```
YoGhurt111:
```suggestion
	l2HeadPollInterval = 500 * time.Millisecond
```
In some network, we have enabled the 1s block time

packages/taiko-client/prover/config.go
```diff
@@ -62,6 +62,20 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
...
+	// (the driver can fall back to HTTP polling), so we enforce the prover's
+	// requirement here. Placed after the private-key validation so existing
+	// error-precedence assertions in tests stay intact.
+	if c.String(flags.L1WSEndpoint.Name) == "" {
+		return nil, fmt.Errorf("flag --%s is required for the prover", flags.L1WSEndpoint.Name)
+	}
+	if c.String(flags.L2WSEndpoint.Name) == "" {
+		return nil, fmt.Errorf("flag --%s is required for the prover", flags.L2WSEndpoint.Name)
+	}
+	if c.String(flags.L2HTTPEndpoint.Name) == "" {
```
YoGhurt111: It seems that we can remove this requirement.

YoGhurt111 (approved):

packages/taiko-client/prover/config.go
davidtaikocha: yes, removed: https://github.com/taikoxyz/taiko-mono/pull/21617/changes/f2148634e0011819da66352156d0f2214728f450

YoGhurt111 (approved):

PR #21631, opened 2026-05-04
refactor(protocol): inline Inbox setup logic

ggonzalez94:
## Summary
Inline the small Inbox setup library into `Inbox` so constructor validation and genesis state initialization live with the contract that uses them.
`MIN_RING_BUFFER_SIZE`, the setup helpers, and setup-specific errors now follow the existing Inbox organization, and `LibInboxSetup` is removed.

## Tests
- `forge test --match-path test/layer1/core/inbox/InboxInit.t.sol`
- `forge fmt --check contracts/layer1/core/impl/Inbox.sol test/layer1/core/inbox/InboxInit.t.sol`
- `pnpm install --frozen-lockfile` in a clean temporary worktree
- `pnpm test` in a clean temporary worktree


davidtaikocha:
Closed since we've done Shasta setup on Mainnet.

PR #21640, opened 2026-05-05
fix(protocol): verify proof before state mutation in prove()

bravesasha:
State was finalized and checkpoint saved before proof verification. Moved verifyProof to run first to enforce CEI ordering

### DIFF
--- a/packages/protocol/contracts/layer1/core/impl/Inbox.sol
+++ b/packages/protocol/contracts/layer1/core/impl/Inbox.sol
@@ -295,8 +295,19 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
                 _processLivenessBond(commitment, offset);
             }
 
+            // ---------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dantaik:
The verifier is governed by the same owner (the DAO), so calling its mutation function after the state change is safe. Placing the external call at the end of the function also improves gas efficiency, which is why it’s implemented this way.

bravesasha:
Agreed it's safe today under DAO governance, but CEI violation means correctness depends on external trust assumptions. If the verifier is ever upgraded, this ordering becomes a real risk. Happy to defer to the team's call.

dantaik (approved):

bravesasha:
@dantaik deepseek-review fails due to GitHub not passing secrets to fork PRs, not related to the code. Could someone re-run it manually?

davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/pull/21640#issuecomment-4385675958

PR #21642, opened 2026-05-06
fix(relayer): harden message handling and config validation

YoGhurt111:
## Summary

This PR hardens the relayer Go services against several reliability and configuration issues found during review.

The processor now recovers from per-message panics, skips receipt logs that have no topics before reading `Topics[0]`, persists unprofitable retry counts in the queue message body, and avoids acknowledging original deliveries when retry publication fails. Transient processing errors are now explicitly nacked with requeue so RabbitMQ does not leave them stuck as unacked deliveries.

The watchdog now treats malformed queue payloads as permanent failures instead of requeueing them forever. The bridge CLI validates required bridge addresses before binding contracts. The shared DB helper now interprets `db.connMaxLifetime` as seconds, matching the flag documentation, and the relayer HTTP server validates that `TaikoL2` is present before registering routes that dereference it. Gas tip fallback handling now returns after setting fallback values so they are not overwritten with nil.

The shared Go service Dockerfile also pins the `golang:1.26.0` and `alpine:latest` base images to immutable manifest digests.

## Tests

```bash
go test -count=1 ./packages/relayer/pkg/utils ./packages/relayer/pkg/db ./packages/relayer/processor ./packages/relayer/watchdog ./packages/relayer/pkg/http ./packages/relayer/bridge
```

```bash
go test -count=1 $(go list ./packages/relayer/... | grep -v /pkg/repo)
```

I also ran:

```bash
go test -count=1 ./packages/relayer/...
```

That full command only failed in `packages/relayer/pkg/repo` because the local environment lacks rootless Docker for testcontainers (`rootless Docker not found`).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ARG PACKAGE=eventindexer
 
-FROM golang:1.26.0 AS builder
+FROM golang:1.26.0@sha256:fb612b7831d53a89cbc0aaa7855b69ad7b0caf603715860cf538df854d047b84 AS builder
 
 ARG PACKAGE
 
@@ -18,7 +18,7 @@ WORKDIR /taiko-mono/packages/${PACKAGE}
 
 RUN CGO_ENABLED=0 GOOS=linux go build -o /taiko-mono/packages/${PACKAGE}/bin/${PACKAGE} /taiko-mono/packages/${PACKAGE}/cmd/main.go
 
-FROM alpine:latest
+FROM alpine:latest@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
 
 ARG PACKAGE
 ENV PACKAGE=${PACKAGE}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ARG PACKAGE=eventindexer
 
-FROM golang:1.26.0
+FROM golang:1.26.0 AS builder
 
 ARG PACKAGE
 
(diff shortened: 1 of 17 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG PACKAGE=eventindexer
 
-FROM golang:1.26.0 AS builder
+FROM golang:1.26.0@sha256:fb612b7831d53a89cbc0aaa7855b69ad7b0caf603715860cf538df854d047b84 AS builder
```
YoGhurt111:
```suggestion
FROM golang:1.26.0
```

Dockerfile
```diff
@@ -18,7 +18,7 @@ WORKDIR /taiko-mono/packages/${PACKAGE}
 
 RUN CGO_ENABLED=0 GOOS=linux go build -o /taiko-mono/packages/${PACKAGE}/bin/${PACKAGE} /taiko-mono/packages/${PACKAGE}/cmd/main.go
 
-FROM alpine:latest
+FROM alpine:latest@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
```
YoGhurt111:
```suggestion
FROM alpine:latest
```

Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG PACKAGE=eventindexer
 
-FROM golang:1.26.0 AS builder
+FROM golang:1.26.0
```
YoGhurt111:
```suggestion
FROM golang:1.26.0 AS builder
```

davidtaikocha (approved):
Approved but plz fix the lint errors in ci.

davidtaikocha (approved):

PR #21670, opened 2026-05-13
fix(taiko-client): prevent genesis reset on event-sync handoff

jmadibekov:
This is the small driver-side fix that pairs with the NMC fix in https://github.com/NethermindEth/nethermind/pull/11648.

## The bug

On the tick right after beacon-sync finishes, the driver calls `SetUpEventSync(0)`. `HeaderByNumber(0)` then resolves to genesis, which resets the L1 cursor and reorgs the EE chain head all the way back to block 0.

## Behavior before / after

**Before** (patched NMC with the changes from https://github.com/NethermindEth/nethermind/pull/11648 + unpatched driver): 
```
12:45:13 driver | L2 EE has finished the P2P sync work  lastSyncedBlockID=17844
12:45:20 driver | Switch to insert pending proposals one by one  p2pOutOfSync=false
12:46:25 nmc    | ForkchoiceUpdatedHandler | Synced Chain Head to 0 (0x0ab9db...c132c4)   ← chain reset to genesis
```

**After** (patched NMC with the changes from https://github.com/NethermindEth/nethermind/pull/11648 + driver with these changes): 
```
13:02:59 nmc | Synced Chain Head to 17873
13:03:02 nmc | Synced Chain Head to 17874
13:03:06 nmc | Synced Chain Head to 17875
...
13:09:11 nmc | Synced Chain Head to 17905
```

Final state: NMC head matches devnet head, `eth_syncing: false`, zero genesis resets, zero `Not receiving ForkChoices` warnings.


packages/taiko-client/driver/chain_syncer/beaconsync/progress_tracker.go
```diff
@@ -100,7 +102,14 @@ func (t *SyncProgressTracker) track(ctx context.Context) {
...
 
 		if new(big.Int).SetUint64(headHeight).Cmp(t.lastSyncedBlockID) >= 0 {
+			// EE reports `eth_syncing: false` and its head is at or past the
+			// latest beacon pivot. Mark out-of-sync so chain_syncer.Sync's
+			// needNewBeaconSyncTriggered returns false on the next tick and the
+			// syncer falls through to MarkFinished + SetUpEventSync, instead of
+			// waiting out the full `--p2p.syncTimeout`.
+			t.lastSyncProgress = nil
 			t.lastProgressedTime = time.Now()
+			t.outOfSync = true
```
YoGhurt111: `MarkFinished` would be better here.

YoGhurt111:
Triggering beacon sync only once can also introduce an issue: you will only sync up to the height at the time the node started. If beacon sync takes a long time but still finishes before the timeout, the synced height may still be far behind the actual latest height.

YoGhurt111:
> Triggering beacon sync only once can also introduce an issue: you will only sync up to the height at the time the node started. If beacon sync takes a long time but still finishes before the timeout, the synced height may still be far behind the actual latest height.

@davidtaikocha WDYT about this?

davidtaikocha:
> > Triggering beacon sync only once can also introduce an issue: you will only sync up to the height at the time the node started. If beacon sync takes a long time but still finishes before the timeout, the synced height may still be far behind the actual latest height.
> 
> @davidtaikocha WDYT about this?

I think its true @jmadibekov , so in driver, during the beacon sync, it will call `needNewBeaconSyncTriggered()` to compare the local and target chain tip constantly every ~12 secs, and once `needNewBeaconSyncTriggered()` returns `false`, driver then will call `progressTracker.MarkFinished()`, so probly sth was wrong in this check? Could u also share more logs during that period to @YoGhurt111 to analyze as well?

jmadibekov:
@YoGhurt111 @davidtaikocha reimplemented to make my fix a bit cleaner. Also, I ran a fresh node against Taiko devnet locally and I see that driver still re-triggers beacon-sync repeatedly during bulk download.
   
Will share the entire log files via slack.

YoGhurt111 (approved):

davidtaikocha (approved):

PR #21693, opened 2026-05-18
fix(taiko-client): skip updating the finalized height when LastFinalizedProposalId is larger than proposalID

### LINKED ISSUE #21692: Finalized block progression has stopped

YoGhurt111:


davidtaikocha (approved):
And do we also have to update taiko-client-rs for this?

PR #21715, opened 2026-05-22
fix(taiko-client): allow preconf ingress before first proposal

YoGhurt111:
## Summary

- Align Go `taiko-client` post-beacon-sync preconf readiness so imports can open when the current head-L1-origin condition is met or when core state still has no proposal (`nextProposalId <= 1`).
- Keep the existing ready path unchanged: if `head_l1_origin` is already written, preconf imports open without requiring the new core-state check.
- Add a pure unit test for the readiness decision.

## Why

Before the first proposal exists, there is no event-confirmed proposal range to validate. Requiring `head_l1_origin` in that state can keep preconfirmation imports closed even though the contract core state indicates `nextProposalId <= 1`.

## Validation

- `go test ./driver/chain_syncer -run TestShouldEnablePreconfImportsAfterEventSync`

Notes:
- `go test ./driver/chain_syncer` was attempted, but the existing integration suite requires local L1/L2 env and failed during setup with empty env/private-key parsing before running those suite tests.

### DIFF
--- a/packages/taiko-client/driver/chain_syncer/chain_syncer.go
+++ b/packages/taiko-client/driver/chain_syncer/chain_syncer.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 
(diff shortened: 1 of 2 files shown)


packages/taiko-client/driver/chain_syncer/chain_syncer.go
davidtaikocha: Shall we update the comment here as well? It still says: `once head L1 origin has been written by the first successful event sync`

davidtaikocha (changes requested):
Golang changes LGTM with one small nitpick, but the Rust changes seem not really correct, let's just keep the Golang changes here and merge this PR.

packages/taiko-client/driver/chain_syncer/chain_syncer.go
```diff
@@ -149,14 +160,15 @@ func (s *L2ChainSyncer) Sync() error {
 	return nil
 }
 
+func shouldEnablePreconfImportsAfterEventSync(headL1OriginWritten bool, nextProposalID *big.Int) bool {
```
davidtaikocha: Let's add a proper comment for this method as well.

YoGhurt111: [59cc451](https://github.com/taikoxyz/taiko-mono/pull/21715/commits/59cc4512dd8dd6580ecad835da758354463ce877)
Added.

packages/taiko-client/driver/chain_syncer/chain_syncer.go
YoGhurt111: Updated in [59cc451](https://github.com/taikoxyz/taiko-mono/pull/21715/commits/59cc4512dd8dd6580ecad835da758354463ce877)

davidtaikocha (approved):

PR #21718, opened 2026-05-23
fix(eventindexer): harden combine-bridged-events pipeline

operagxoksana:
Fixed a critical bug where events from the same transaction were silently dropped during deduplication (was using txHash+vaultType, now txHash+logIndex). Also added guards for malformed records, missing RPC URLs, invalid bridgedToken addresses, and replaced all silent failures with proper logging.


davidtaikocha:
Closed since this tool is no longer used.

operagxoksana:
Hey @davidtaikocha, maybe we should remove this tool completely?

PR #21735, opened 2026-05-30
fix(taiko-client): resolve prover proposal last block from extraData

davidtaikocha:
## Summary

The prover resolved a proposal's last L2 block via reth's L1Origin table, which caused two failures:

1. **Premature last block (event-sync race).** reth writes `BatchToLastBlock[proposalID]` for every non-preconf block as it's built, so mid-insertion the mapping reflects "last block so far", not the complete proposal. `LastCertainL1OriginByBatchID` reads that cache with no completeness check, so the prover could build a proof against the wrong end block → `Inbox.prove` reverts with `ParentBlockHashMismatch`. Surfaces with the op prover (submits fast enough to hit the window).
2. **Beacon-synced proposals unprovable.** Beacon/snap-synced blocks carry full headers but no L1Origin row, so `LastCertainL1OriginByBatchID` returns nothing and `LastL1OriginInProposal(P-1)` returned `(nil, nil)` → nil deref in the proof submitter.

## Fix

Resolve proposal boundaries from L2 block-header `extraData` (the proposal ID) instead of the L1Origin table — `extraData` is present regardless of sync mode. Scoped to `packages/taiko-client` (no alethia-reth / protocol changes).

- `WaitProposalHeader` now resolves the candidate last block via `LastBlockIDByBatchID` (cache-then-extraData-scan; never returns a preconf block), reads headers by number, and only returns once the **boundary is sealed**: the block after the candidate exists and its extraData proposal ID is `> P`.
- Explicit prover-side **preconf guard**: a candidate whose L1Origin is present with zero L1 block height is rejected (proposal not yet landed by the driver). Beacon-synced candidates (L1Origin absent) are accepted.
- Previous-proposal lookups use a new beacon-safe `ProposalLastBlockID`; the aggregation builder derives `FirstProposalParentBlockHash` from the first block's `ParentHash()`.
- New `DecodeShastaExtraDataProposalID` helper; removed the now-unused `LastL1OriginInProposal`.

Tradeoff: at the chain tip (no next proposal yet) `WaitProposalHeader` waits until the next proposal's first block lands, then retries — self-heals in steady state.

## Test plan

- [x] `PACKAGE=bindings/encoding make test` — `TestDecodeShastaExtraDataProposalID` (round-trip + length validation)
- [x] `PACKAGE=pkg/rpc make test` — `TestEvaluateProposalSeal` (preconf / beacon-synced / event-synced / boundary / malformed) + existing suite
- [x] `PACKAGE=prover/proof_submitter make test` — existing prover suite
- [x] `go build ./...` / `go vet` clean
- [ ] Devnet validation: prove an event-synced proposal under load (premature-window) and a beacon-synced proposal


.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main, feat/prover-proposal-boundary-guard]
```
```suggestion
    branches: [main]
```

packages/taiko-client/prover/proof_submitter/proof_submitter.go
```diff
@@ -121,27 +121,33 @@ func (s *ProofSubmitter) RequestProof(ctx context.Context, meta metadata.TaikoPr
 		return fmt.Errorf("failed to wait for L2 header, blockID: %d, error: %w", proposalID, err)
 	}
 
-	lastOriginInLastProposal, err := s.rpc.LastL1OriginInProposal(
+	prevProposalLastBlockID, err := s.rpc.ProposalLastBlockID(
```
YoGhurt111, line 124: Why we choose to get block id here instead of L1 origin?
It makes one more rpc call in the following else condition?

packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -295,21 +324,55 @@ func (c *Client) WaitProposalHeader(ctx context.Context, proposalID *big.Int) (*
 			return nil, ctxWithTimeout.Err()
 		}
 
-		l1Origin, err := c.L2Engine.LastCertainL1OriginByBatchID(ctxWithTimeout, proposalID)
+		// Resolve the candidate last block of the proposal from block extraData. This works for
+		// both event-synced and beacon-synced blocks and never returns a preconfirmation block.
+		candidateID, err := c.L2Engine.LastBlockIDByBatchID(ctxWithTimeout, proposalID)
```
YoGhurt111: We can use `ProposalLastBlockID` here directly

packages/taiko-client/pkg/rpc/methods.go
```diff
@@ -295,21 +324,55 @@ func (c *Client) WaitProposalHeader(ctx context.Context, proposalID *big.Int) (*
...
 		}
+		if candidateID == nil {
+			continue
+		}
 
-		if l1Origin == nil {
+		// Fetch the candidate's L1Origin: absent (NotFound) for beacon-synced blocks, present
+		// with zero L1 height for preconfirmation blocks.
+		candidateL1Origin, err := c.L2.L1OriginByID(ctxWithTimeout, candidateID.ToInt())
+		if err != nil && err.Error() != ethereum.NotFound.Error() {
```
YoGhurt111, line 342: BTW, seems we missed the `HeadL1Origin` check in this PR. When candidate block is synced by beacon sync, we would keep retrying to get the L1 origin that should never occur until timeout.

packages/taiko-client/prover/proof_submitter/proof_submitter.go
davidtaikocha: Since for beacon synced blocks, there is no L1 Origin to fetch?

packages/taiko-client/pkg/rpc/methods.go
davidtaikocha: Since we are using `LastBlockIDByBatchID` rn, and waiting for one more proposal to land, seems `HeadL1Origin` check is no longer needed?

packages/taiko-client/pkg/rpc/methods.go
davidtaikocha: Done in 38337e15b — `WaitProposalHeader` now calls `ProposalLastBlockID` directly. It returns a `*big.Int` and folds the nil result into an error, so the separate nil check and the repeated `ToInt()` conversions are gone. `CtxWithTimeoutOrDefault` keeps the existing deadline, so no extra timeout is introduced.

YoGhurt111 (approved):

YoGhurt111 (approved):

PR #21751, opened 2026-06-03
feat(taiko-client-rs): metrics refactor

RogerLamTd:
align with ejector/reorg-monitor metrics, should make histograms useable


YoGhurt111:
Thanks for the metrics refactor. I found one blocker and one observability concern.

1. The lockfile is out of sync with the actual manifests. In this PR, `packages/taiko-client-rs/Cargo.lock` records `preconfirmation-net` as depending on `once_cell` / `prometheus`, but `packages/preconfirmation-p2p/crates/net/Cargo.toml` still depends on `metrics.workspace = true` and `metrics-exporter-prometheus.workspace = true`. When I ran `cargo check -p taiko-client` from `packages/taiko-client-rs`, Cargo modified the lockfile and restored `metrics 0.23.1`, `metrics-exporter-prometheus 0.15.3`, `metrics-util 0.17.0`, and `sketches-ddsketch 0.2.2`, and changed `preconfirmation-net` back to `metrics` / `metrics-exporter-prometheus`. Please regenerate and commit the lockfile, or update the `preconfirmation-net` manifest if that crate is also meant to move to direct `prometheus`.

2. Potential observability regression: the new metrics endpoint only serves `prometheus::gather()`, while the old `metrics_exporter_prometheus::PrometheusBuilder::install()` installed a global `metrics` facade recorder. The taiko-client crates touched here were migrated to direct `prometheus`, but the dependency graph still includes crates using the `metrics` facade, including reth-related crates and currently `preconfirmation-net`. Unless dropping those facade-emitted metrics from `/metrics` is intentional, we probably need a bridge/recorder or a clearer migration plan for dashboards.

Verification: `cargo check -p taiko-client` passed locally, but it produced the lockfile diff described above.

YoGhurt111 (approved):

davidtaikocha (approved):

PR #21767, opened 2026-06-10
fix(protocol): use address-keyed TDX proofs

smtmfft:
## Summary
- Switch GCP/Azure TDX proof verification to the 85-byte `address || signature` layout.
- Look up the verifier-assigned instance id from `instanceIdByAddress` instead of requiring the remote prover to include it.
- Add focused TDX verifier tests and update the TDX deployment doc.

## Test Plan
- `forge test --match-path test/layer1/verifiers/TdxVerifier.t.sol -vv`
- `forge test --match-path test/layer1/verifiers/SgxVerifier.t.sol -vv`


smtmfft:
Closing this stacked PR because the base TDX PR has absorbed the address-keyed 85-byte proof changes.

PR #21776, opened 2026-06-11
refactor(taiko-client-rs): consolidate shared assembly and collapse single-impl abstractions

davidtaikocha:
## Summary

Items 3–5 from the post-#21774 simplification backlog, as 7 bisectable commits. Net **−1,075 lines** (64 files, +1,328/−2,403), behavior-preserving by intent throughout.

**Consolidations (new shared code in `protocol`):**
- `protocol::metrics` — one copy of `DURATION_SECONDS_BUCKETS` + counter/gauge/histogram registration helpers, previously byte-identical in three crates. Driver metrics rewritten as typed-field handles (no more 29 name constants + per-call linear string lookup); **exported metric names unchanged verbatim**.
- `protocol::shasta::build_payload_attributes(_with_id)` — `TaikoPayloadAttributes` was hand-assembled in **five** crates, each re-stating the same invariants (`prev_randao == mix_hash`, fee recipient == beneficiary, dual u64/U256 timestamp, present-but-empty withdrawals, zeroed `l2_block_hash`). One builder now owns them; per-site differences (L1 origin heights, forced inclusion, signature, anchor tx, beacon root, payload-id stamping) stay explicit. The existing preconfirmation-driver unit test cross-checks the builder against a hand-built expected payload, including the payload-id stamp.
- `protocol::shasta::constants::calculate_next_block_eip4396_base_fee_for_parent` — the genesis rule, timestamp-delta direction, and missing-base-fee handling around the EIP-4396 math were triplicated; now single-sourced.

**Collapses (single-implementation abstractions removed):**
- `ManifestFetcher` and `DerivationPipeline` traits (one impl each, no mocks anywhere) inlined into `ShastaSourceManifestFetcher` / `ShastaDerivationPipeline`.
- `ProductionRouter` no longer scans a `Vec<Arc<dyn …>>` by `ProductionPathKind` for a fixed 1:1 mapping; it holds the canonical path and `Option<preconf path>`. The unreachable `MissingPath` case maps to the existing `DriverError::PreconfirmationDisabled`; `kind()`/`prepare()` (zero overrides, zero callers) and the `ProductionError` enum are gone. `BlockProductionPath` keeps its single `produce` method as the unit-test seam.
- `ProposalTxManager` generic wrapper deleted — the proposer holds `SimpleTxManager` directly; the mock-based tests it existed for only asserted compiler-derived `From` conversions (the real logic, config mapping + `proposal_candidate`, keeps direct tests).
- `TaikoAuthMethod`/`TaikoEngineMethod`/`TaikoOriginMethod` enums (each variant → one string → one call site, zero external users) inlined as literals.
- `PreconfIngressError` trait + both per-crate wrapper modules replaced by a concrete `PreconfIngressSyncError`; consumers absorb it via 1:1 `From` impls, so existing matching on their `Sync`/`Driver` variants is unchanged.

**Test dedup:** shared `test_support` mocks in preconfirmation-driver, a `start_preconf_client` boot helper replacing nine pasted setup blocks (now uniformly surfacing event-loop errors), test-harness delegation-wrapper removal, fourteen zero-caller pub helpers demoted to `pub(crate)`, and driver e2e tests reusing the env-built `ClientConfig`.

## ⚠️ Flagged for review

- **Proposer tx-manager metric shape change:** the `base_tx_manager_*` collectors lose their `name` label (only value ever was `"proposer"`). Exported names are unchanged, but any Grafana panel selecting on `name=` needs updating. Everything else keeps identical exported shapes.
- The preconf-input-without-registered-path case (unreachable through the runner wiring) now surfaces as `PreconfirmationDisabled` instead of a stringly `MissingPath` inside `DriverError::Other`.

## Verification

- `just clippy` (`-D warnings -D missing_docs -D clippy::missing_docs_in_private_items`): clean
- `just fmt-check` + cargo-sort: clean
- 355 unit tests green across all 7 crates; integration-test binaries compile clean
- Full docker e2e (`just test`) runs in CI — it exercises the consolidated payload assembly and base-fee math against a live devnet
- An adversarial cross-check of metric names, inlined RPC strings, and the five payload-builder call sites is running; findings (if any) will be addressed before marking ready


davidtaikocha:
**Adversarial wire-parity review (automated, cross-checked against `main`): no blocking findings.**

- **Prometheus metrics:** all exported names/types/buckets byte-identical across the four metrics files (driver 29/29, preconfirmation 17/17, whitelist, proposer). Only shape change is the documented `base_tx_manager_*` `name`-label drop.
- **RPC method strings:** 13/13 inlined literals byte-identical to the old enum mappings, each at its original call site with unchanged param tuples.
- **Payload assembly:** all 5 builder call sites reconstruct every field identically, and `payload_id_taiko` hashes the same zeroed-id struct with the same parent hash as before; the proposer and test-harness sites correctly remain unstamped.
- **Base fee / error paths:** delta direction, genesis semantics, clamp threading, and per-site error variants unchanged; the new `From` impls reproduce the deleted `map_driver_error` flattening input-for-input.

Three P3 notes, all display-equivalent or on practically unreachable paths: (1) `PreconfIngressSync::start` now flattens `DriverError::Sync` the same way the readiness path always did (variant identity differs on that one path, no matcher affected); (2) `compute_base_fee_per_gas` fetches the grandparent before the missing-base-fee check (one extra RPC only when a non-genesis parent lacks a base fee); (3) metric-registration panic wording standardized.

YoGhurt111 (approved):

PR #21791, opened 2026-06-15
feat(taiko-client): add local proposal metadata API

smtmfft:
## Summary
- add a disabled-by-default loopback-only driver API for Shasta proposal metadata
- expose `GET /internal/shasta/proposals/{proposal_id}` with reconstructed proposal, `hashProposal`, and L1 event coordinates
- wire `proposalApi.enabled` / `proposalApi.addr` flags and reject non-loopback binds

## Tests
- `go test ./packages/taiko-client/pkg/proposalapi ./packages/taiko-client/cmd/flags`
- `go test ./packages/taiko-client/driver -run 'TestNewConfigFromCliContextRejectsRemoteProposalAPIAddress|TestResolveEndpointsPrefersWSOverHTTP'`\n- `git diff --check`\n\n## Notes\n- Full `go test ./packages/taiko-client/driver` was not usable in this local environment: the existing driver suite expects devnet env vars/JWT and fails during `internal/testutils` setup with empty env parsing / invalid JWT length before reaching these changes.\n- `golangci-lint` is not installed in this environment (`command not found`).

### DIFF
--- a/packages/taiko-client/README.md
+++ b/packages/taiko-client/README.md
@@ -45,6 +45,25 @@ Review each sub-command's command line flags:
 bin/taiko-client <sub-command> --help
 ```
 
+### Local proposal metadata API
+
+The driver can expose a disabled-by-default, loopback-only Shasta proposal metadata API for co-located
+TDX provers:
+
+```sh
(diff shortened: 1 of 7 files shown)


davidtaikocha:
Closed since we are not moving this forward.

PR #21804, opened 2026-06-17
fix(taiko-client): fallback RISC0 proofs to SP1

YoGhurt111:
## Summary

- Remove the `zk_any` proof type and zk-any draw fallback path.
- Request RISC0 proofs directly when ZKVM proving is configured.
- Fall back from RISC0 to SP1 when the proposal distance exceeds `prover.maxRisc0ProofProposalDistance`.
- Keep RISC0 and SP1 on the same ZKVM proof producer by carrying the selected proof type in `ProposalProofRequestOptions`.
- Add `prover.forceSP1Proof` / `PROVER_FORCE_SP1_PROOF` to request SP1 from the ZKVM proof producer without trying RISC0 first. If no ZKVM proof producer is configured, the prover keeps using the base proof producer.
- Guard against nil proof responses from proof producers so malformed successful responses return errors instead of panicking later in proof handling.
- Remove the `maxProofRequestTimeout` SGX fallback behavior so configured ZKVM proving no longer falls back to base proofs after request timeout.
- Rename the proof-distance option to `prover.maxRisc0ProofProposalDistance` and remove legacy `prover.maxZKProofProposalDistance` flag/env compatibility.
- Rename RISC0/SP1 fallback and backlog files/types/metrics from generic ZK wording to RISC0-specific wording.

## Validation

- `go test -count=1 ./prover/proof_submitter`
- `go test -count=1 ./prover/proof_producer`
- `go test -race -count=1 ./prover/proof_submitter -run 'TestRisc0SP1FallbackTestSuite|TestRequestProposalProof'`
- `go test -count=1 ./prover -run 'TestNewConfigFromCliContextForceSP1Proof|TestNewConfigFromCliContextMaxRisc0ProofProposalDistance'`
- `go test -count=1 ./cmd/...`
- `make lint` from `packages/taiko-client`
- `git diff --check`

Note: full `go test ./prover` still depends on local integration-style test environment variables in this checkout and fails before reaching these change paths.


YoGhurt111:
About the Critical Issue 1, that's we expected.

YoGhurt111:
https://github.com/taikoxyz/raiko2/issues/98 About the changes of raiko api for clear and status, will wait the feature request finished.

davidtaikocha (approved):

PR #21806, opened 2026-06-22
feat(protocol): version the SignalService checkpoints mapping

dantaik:
Namespaces SignalService checkpoints by an immutable `checkpointVersion` (a new third constructor argument, currently set to `1` at every deploy/genesis site). The `_checkpoints` mapping becomes `version => blockNumber => CheckpointRecord`, so any future version bump cleanly invalidates all previously stored checkpoints without a storage migration. The `ICheckpointStore` interface and the public `saveCheckpoint`/`getCheckpoint` API are unchanged, so callers (`Inbox`, `Anchor`) are unaffected. The genesis immutable injector and all deploy scripts/test fixtures now pass version `1`, and the prove gas snapshot is updated to reflect the extra mapping level.


YoGhurt111 (approved):

davidtaikocha (approved):

packages/protocol/contracts/shared/signal/SignalService.sol
```diff
@@ -53,11 +61,16 @@ contract SignalService is EssentialContract, ISignalService {
...
+    /// @dev Deprecated unversioned checkpoint storage. Retained only to preserve the storage
+    /// layout; no longer read from or written to.
+    mapping(uint48 blockNumber => CheckpointRecord checkpoint) private _deprecatedCheckpoints;
+
     /// @notice Storage for checkpoints persisted via the SignalService.
-    /// @dev Maps block number to checkpoint data
-    mapping(uint48 blockNumber => CheckpointRecord checkpoint) private _checkpoints;
+    /// @dev Maps checkpoint version => block number => checkpoint data.
+    mapping(uint256 version => mapping(uint48 blockNumber => CheckpointRecord checkpoint)) private
+        _checkpoints;
```
davidtaikocha, line 69: Do we also need to cleanup _receivedSignals cache? In case the signals in the wrong checkpoints have been proved once and cached?

dantaik: Good catch, let me think about how to fix it.

dantaik: Done

davidtaikocha (approved):

dantaik:
ported to https://github.com/taikoxyz/taiko-mono/pull/21820

PR #21809, opened 2026-06-22
feat(protocol): add Bridge.init3() to reset stale RETRIABLE message statuses

dantaik:
Adds a one-time, owner-only `init3(bytes32[])` reinitializer to `Bridge` that resets the status of the listed messages from `RETRIABLE` back to the default `NEW`, allowing previously deployed contracts to clear stale message statuses. Each supplied hash must currently be in `Status.RETRIABLE`, otherwise the whole call reverts via `_checkStatus`. A dedicated `MessageStatusReset(bytes32 indexed msgHash)` event is emitted per reset instead of reusing `MessageStatusChanged`, so indexers can distinguish administrative resets from normal status transitions. Guarded by `reinitializer(3)` so it can only be executed once. Adds `Bridge1_init3.t.sol` covering the reset path, the owner check, the non-`RETRIABLE` revert, single-call enforcement, and the empty-array case.

The following message hashes shall be deleted from the L1 bridge's state:

- `0xf2994252987db0f55b6accdf1ff979bc508cd4acfcc229ad64b23a67f6fd984d`
- `0xea26be1009e743aec78e1f566e91db0b9fda29e16fec1e72e2d74c6983a68e70`
- `0x997216448ef88e6398e82b0f003abb8637d25441ca6d22b09a65f63ef077480a`

### DIFF
--- a/packages/protocol/contracts/shared/bridge/Bridge.sol
+++ b/packages/protocol/contracts/shared/bridge/Bridge.sol
@@ -126,6 +126,18 @@ contract Bridge is EssentialResolverContract, IBridge, IEthMinter {
         __reserved3 = 0;
     }
 
+    /// @notice Deletes the status of a list of messages, resetting each to the default
+    /// `Status.NEW`. Each message must currently be in `Status.RETRIABLE`. Intended to be
+    /// called once by previously deployed contracts to clear stale message statuses.
+    /// @param _msgHashes The hashes of the messages whose status should be deleted.
+    function init3(bytes32[] calldata _msgHashes) external onlyOwner reinitializer(3) {
+        for (uint256 i; i < _msgHashes.length; ++i) {
+            _checkStatus(_msgHashes[i], Status.RETRIABLE);
+            delete messageStatus[_msgHashes[i]];
+            emit MessageStatusReset(_msgHashes[i]);
+        }
+    }
+
     /// @inheritdoc IBridge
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/shared/bridge/Bridge.sol
```diff
@@ -126,6 +126,18 @@ contract Bridge is EssentialResolverContract, IBridge, IEthMinter {
         __reserved3 = 0;
     }
 
+    /// @notice Deletes the status of a list of messages, resetting each to the default
+    /// `Status.NEW`. Each message must currently be in `Status.RETRIABLE`. Intended to be
+    /// called once by previously deployed contracts to clear stale message statuses.
+    /// @param _msgHashes The hashes of the messages whose status should be deleted.
+    function init3(bytes32[] calldata _msgHashes) external onlyOwner reinitializer(3) {
+        for (uint256 i; i < _msgHashes.length; ++i) {
+            _checkStatus(_msgHashes[i], Status.RETRIABLE);
```
davidtaikocha, line 135: probly we need to do this: https://github.com/taikoxyz/taiko-mono/pull/21806#discussion_r3452125938 , before we call this method.

dantaik: fixed there

dantaik:
ported to https://github.com/taikoxyz/taiko-mono/pull/21820

PR #21813, opened 2026-06-22
fix(protocol): enable SGX local report checks by default

failsafesecurity:
## Summary

Make Automata DCAP attestation fail closed by enabling local enclave report checks during initialization, and add an explicit setter for operators/tests instead of relying only on a toggle.

This keeps MRENCLAVE/MRSIGNER checks on by default for new deployments, so quote verification requires the configured application enclave identity unless an owner deliberately disables the check.

## Changes

- Set `checkLocalEnclaveReport = true` in `AutomataDcapV3Attestation.init()`.
- Add `setLocalReportCheck(bool)` for explicit owner-controlled configuration.
- Update attestation tests to assert the default and avoid accidentally disabling checks in tests that intend to verify the checked path.

## Test plan

- Not run locally: this environment does not have Foundry/`forge` installed.
- Intended command: `cd packages/protocol && FOUNDRY_PROFILE=layer1 forge test --match-contract TestAutomataDcapV3Attestation`.

## Security note

This is a fail-closed hardening change for SGX attestation policy. It avoids new deployments starting in a permissive local-report-check mode before operator configuration is completed.


smtmfft:
Thanks for the hardening PR. One important detail for mainnet: the relevant Automata attestation contracts are already running with `checkLocalEnclaveReport == true`, so MRENCLAVE/MRSIGNER allowlist checks are enabled there.

This change is still useful as a fail-closed default for new deployments and tests. For mainnet, though, it should be treated as default-configuration hardening rather than a change to the currently active local report policy.

dantaik:
> checkLocalEnclaveReport == true

where is the code?

dantaik:
This variable is set to true by default in https://github.com/taikoxyz/taiko-mono/blob/ec069a140b7ea80828c229e02f0738b5b2894a98/packages/protocol/script/layer1/core/DeployShastaHoodi.s.sol

PR #21814, opened 2026-06-22
fix(protocol): make signal checkpoints immutable

failsafesecurity:
## Summary

Make SignalService checkpoints immutable per block number: once an authorized syncer saves a checkpoint for a block, a later call cannot overwrite it with a different root/hash.

## Why

`saveCheckpoint` currently accepts a later write to the same `blockNumber`, replacing the stored checkpoint. Since successful signal proofs can be cached and later reused with an empty proof, allowing checkpoint replacement increases incident blast radius: a temporary or accidental bad root can be used for proofs and then overwritten, while cached signal state remains.

Immutability keeps checkpoint history auditable and prevents silent replacement for the same block number.

## Changes

- Add `SS_CHECKPOINT_ALREADY_EXISTS`.
- Reject `saveCheckpoint` when a checkpoint already exists for the block number.
- Add a unit test for duplicate checkpoint writes.

## Test plan

- Static check: `git diff --check` passed.
- Not run locally: this environment does not have Foundry/`forge` installed.
- Intended command: `cd packages/protocol && FOUNDRY_PROFILE=shared forge test --match-contract TestSignalService`.

### DIFF
--- a/packages/protocol/contracts/shared/signal/SignalService.sol
+++ b/packages/protocol/contracts/shared/signal/SignalService.sol
@@ -154,6 +154,9 @@ contract SignalService is EssentialContract, ISignalService {
         if (msg.sender != _authorizedSyncer) revert SS_UNAUTHORIZED();
         if (_checkpoint.stateRoot == bytes32(0)) revert SS_INVALID_CHECKPOINT();
         if (_checkpoint.blockHash == bytes32(0)) revert SS_INVALID_CHECKPOINT();
+        if (_checkpoints[_checkpoint.blockNumber].blockHash != bytes32(0)) {
+            revert SS_CHECKPOINT_ALREADY_EXISTS();
+        }
 
         _checkpoints[_checkpoint.blockNumber] = CheckpointRecord({
             blockHash: _checkpoint.blockHash, stateRoot: _checkpoint.stateRoot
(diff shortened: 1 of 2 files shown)


dantaik (approved):

packages/protocol/contracts/shared/signal/SignalService.sol
```diff
@@ -154,6 +154,9 @@ contract SignalService is EssentialContract, ISignalService {
         if (msg.sender != _authorizedSyncer) revert SS_UNAUTHORIZED();
         if (_checkpoint.stateRoot == bytes32(0)) revert SS_INVALID_CHECKPOINT();
         if (_checkpoint.blockHash == bytes32(0)) revert SS_INVALID_CHECKPOINT();
+        if (_checkpoints[_checkpoint.blockNumber].blockHash != bytes32(0)) {
```
dantaik, line 157: After a second thought: I think it's OK for a new proposal to re-insert (save) the same checkpoint at the same block number.
Will the current behavior introduce any issue?

failsafesecurity:
Hi @dantaik 👋 — thanks for approving this PR! All CI checks are passing. Is there anything else blocking merge, or would you like me to rebase/squash to help get this landed?

davidtaikocha:
Closed since we are not moving this forward.

PR #21807, opened 2026-06-22
feat(protocol): add Inbox.init2() to reset Inbox core state

dantaik:
Adds a one-time `init2` to Inbox to reset chain state.

PR #21212, opened 2026-01-26
feat(protocol): add mintEth to Bridge for trusted protocols

dantaik:
This is to prepare the Bridge to be able to trust certain protocols, e.g. https://github.com/taikoxyz/shadow, to mint ETH.

PR #21820, opened 2026-06-22
feat(protocol): port hack recovery hooks to v3

ggonzalez94:
## Summary

This PR fixes all the structural issues the exploit caused. 
1. Implements a version  for signals and checkpoints in the `SignalService` so that any previous state transition cannot be used to withdraw funds from the bridge or to verify any L2 state on L1
2. adds a new  one time initializer function  `Bridge.init3(bytes32[])` to invalidate existing retriable messages the attacker had already proven, but was not able to execute. Without this, even with the fix in 1. he can re-execute the existing messages.
3. adds a new one time initializer function to the inbox `Inbox.init2(...)` to restore the state of the chain to a point before the attack.
4. removes the `IEthMinter` funcitionality that was never deployed on-chain, and was part of a [feature](https://github.com/taikoxyz/taiko-mono/pull/21212) Taiko decided not to deploy. This aligns the code with that decision

Ports the recovery changes from #21806, #21809, and #21807 onto `taiko-alethia-protocol-v3.0.0` in one branch.

## Intentional differences from source PRs

- Bridge recovery marks each listed `RETRIABLE` message as `DONE` instead of deleting the status back to `NEW`. This prevents both proofless retries and later re-processing as a fresh message. Because of that, this does not carry over #21809's `MessageStatusReset` event; it emits the existing `MessageStatusChanged` event.
- Inbox recovery is `onlyOwner` instead of a hardcoded recovery admin address, matching the DAO/governance execution path for the deployed owner-controlled proxy.


packages/protocol/contracts/shared/bridge/Bridge.sol
```diff
@@ -126,6 +126,18 @@ contract Bridge is EssentialResolverContract, IBridge, IEthMinter {
...
     }
 
+    /// @notice Invalidates stale retriable messages by marking them done.
+    /// @dev Intended for one-time recovery on already deployed contracts. Each message must
+    /// currently be in `Status.RETRIABLE`; marking it `DONE` prevents both proofless retries and
+    /// later re-processing as a new message.
+    /// @param _msgHashes The hashes of the messages to invalidate.
+    function init3(bytes32[] calldata _msgHashes) external onlyOwner reinitializer(3) {
+        for (uint256 i; i < _msgHashes.length; ++i) {
+            _checkStatus(_msgHashes[i], Status.RETRIABLE);
```
ggonzalez94: remove this. we should make setting the status unconditionally DONE

dantaik (approved):

davidtaikocha (approved):

PR #21822, opened 2026-06-22
fix(protocol): reject DEBUG SGX instances

smtmfft:
Summary:
- Reject SGX application enclave quotes with the DEBUG attribute bit set during `SgxVerifier.registerInstance`.
- Add coverage for DEBUG-mode quote rejection.
- Update the protocol gas report for the verifier change.

Tests:
- `git diff --check origin/taiko-alethia-protocol-v3.0.0..HEAD`
- `FOUNDRY_PROFILE=layer1 forge test --match-contract SgxVerifierTest`


YoGhurt111:
I still recommand to use `onlyOwner` modifier in `registerInstance` to avoid unexpected attributes, and we aim to 100% zk, the sgx is only used in emergency case.

dantaik:
please use [taiko-alethia-protocol-v3.0.0](https://github.com/taikoxyz/taiko-mono/tree/taiko-alethia-protocol-v3.0.0) as your PR's base branch.

dantaik (approved):

davidtaikocha (approved):

ggonzalez94 (approved):

ggonzalez94:
> I still recommand to use onlyOwner modifier in registerInstance to avoid unexpected attributes, and we aim to 100% zk, the sgx is only used in emergency case.

I think gavin's suggestion is probably a good idea. Let's discuss in our call

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
dantaik, line 139: add onlyFrom(registrant), if registrant is address(0), it shall be permissionless.

dantaik: @daniel will create another PR on top of this one.

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
```diff
@@ -116,6 +121,10 @@ contract SgxVerifier is IProofVerifier, Ownable2Step {
     {
         (bool verified,) = IAttestation(automataDcapAttestation).verifyParsedQuote(_attestation);
         require(verified, SGX_INVALID_ATTESTATION());
+        require(
```
dantaik, line 155: @smtmfft please add more check as we suggested, then we'll test it in a devnet.

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
dantaik: https://github.com/taikoxyz/taiko-mono/pull/21830

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
davidtaikocha: Here are the suggestions from Claude workflow:

```
So, "besides MRENCLAVE and DEBUG," the security-relevant fields are:

mrSigner — checked; the signing identity. The leaked key ⇒ this is what #21825 rotates.
attributes (the non-DEBUG bits) — still unchecked. Most notable is PROVISION_KEY 0x10: an enclave with provisioning-key access can derive platform-identifying keys; production enclaves normally have it clear. Worth rejecting alongside DEBUG.
isvSvn — still unchecked; this is the "freshness"/anti-downgrade knob from the earlier discussion. It only bites if you ever trust by mrSigner+isvProdId instead of exact mrEnclave.
isvProdId — still unchecked; binds to a specific enclave product.
cpuSvn / miscSelect — unchecked on the report, but cpuSvn's job (platform TCB level) is already covered via the PCK-cert TCB check.
```
Just sharing this here in case it’s helpful.

dantaik (approved):

packages/protocol/contracts/layer1/verifiers/SgxVerifier.sol
smtmfft: Implemented the attributes part by rejecting both `DEBUG(0x02)` and `PROVISION_KEY(0x10)` through a single forbidden attributes mask in `SgxVerifier`.

For the other fields, I do not think we should hard-enforce them in this PR:

- `isvProdId` / `isvSvn`: the current trust boundary is exact `MRENCLAVE && MRSIGNER` allowlisting. With exact `MRENCLAVE`, product/SVN policy is effectively tied to the selected enclave build. Hard-coding these fields here would add deployment/config churn and can easily block valid lanes, since raiko2 and gaiko2 already use different product IDs (`0` vs `1`). If we later move to signer-level trust (`MRSIGNER + isvProdId + minIsvSvn`) instead of exact MRENCLAVE trust, then configurable product/min-SVN checks make sense.
- `cpuSvn`: platform TCB freshness is already covered by the PCK/TCB validation in the attestation path, not by a separate app-report field check here.
- `miscSelect`: we do not currently have a defined expected app mask/value, so enforcing it would be under-specified.

I also checked current generated quotes against the new mask:

- raiko2 mainnet/hoodi attributes: `0x0500000000000000e700000000000000`
- gaiko2 mainnet attributes: `0x05000000000000000700000000000000`

All have `DEBUG=false`, `PROVISION_KEY=false`, and `forbidden_mask(0x12) == 0`.

dantaik (approved):

davidtaikocha (approved):

PR #21825, opened 2026-06-23
fix(protocol): add proposal 0015 SGX trust rotation

smtmfft:
Summary:
- Add Proposal0015 to rotate the trusted SGX MRSIGNER on the mainnet SGX attesters.
- Revoke retired SGX MRENCLAVE values.
- Delete currently registered SGX instances except the allowed SGXGETH id 4 and SGXRETH id 5.

Tests:
- `git diff --check`
- `P=0015 pnpm proposal`
- `FOUNDRY_PROFILE=layer1 forge build --contracts script/layer1/proposals/Proposal0015.s.sol`


dantaik:
We have two options here:
1) adopt the simple fix in https://github.com/taikoxyz/taiko-mono/pull/21822
2) upgrade automata dependency to the latest, audited release, and delete local hardcopies in https://github.com/taikoxyz/taiko-mono/pull/21827/

I'm considering 2).

Also, should we simply deploy a new SGXVerifier instance instead of fixing the existing one.

packages/protocol/script/layer1/proposals/Proposal0015.s.sol
```diff
@@ -0,0 +1,181 @@
...
+pragma solidity ^0.8.24;
+
+import "../governance/BuildProposal.sol";
+
+// To print the proposal action data: `P=0015 pnpm proposal`
+// To dryrun the proposal actions on L1: `P=0015 pnpm proposal:dryrun:l1`
+//
+// This proposal prunes stale SGX instances, rotates the trusted SGX MRSIGNER, and revokes
+// previously trusted SGX MRENCLAVE values from the existing mainnet SGX attesters.
+contract Proposal0015 is BuildProposal {
```
ggonzalez94, line 11: I'm reviewing. But I think:
1. we'll need to unify this proposal with the upgrades to the SS, Bridge and Inbox so that we submit a single emergency proposal
2. We should simplify the logic and structure so that it is easy to review. It currently has many indirections, and things that are calculated on the go. I would prefer to try to simplify the logic as much as we can

packages/protocol/script/layer1/proposals/Proposal0015.s.sol
```diff
@@ -0,0 +1,181 @@
...
+            )
+        });
+        actions[cursor++] = Controller.Action({
+            target: SGXRETH_VERIFIER,
+            value: 0,
+            data: abi.encodeWithSignature(
+                "deleteInstances(uint256[])",
+                _deleteIds(SGXRETH_NEXT_INSTANCE_ID, _sgxRethAllowedIds())
+            )
+        });
```
ggonzalez94, line 99: Do we need to touch the existing verifier contracts if we're going to deploy new ones?

dantaik:
@ggonzalez94 @smtmfft we shall use one single proposal to cover everything including layer 2 Bridge/ERC20Vault.

We should also use [taiko-alethia-protocol-v3.0.0](https://github.com/taikoxyz/taiko-mono/tree/taiko-alethia-protocol-v3.0.0) as the base

ggonzalez94:
superseeded by #21833 

@smtmfft please contribute there. 

@dantaik we've been merging proposals to main(since they don't directly reference contract code). It is a bit messy, that we have it separated from 3.0.0 branch, but I wouldn't change this now.

PR #21817, opened 2026-06-22
feat(protocol): migrate SGX attestation to Taiko-owned Automata DCAP v1.1.0 (npm deps), remove legacy, reject DEBUG enclaves

dantaik:
## Summary

Migrates Taiko&#39;s SGX remote-attestation off the in-tree legacy `AutomataDcapV3Attestation` onto the
current upstream **Automata DCAP attestation** (consumed as **pinned npm dependencies**), has Taiko
**deploy and own its own attestation entrypoint**, and **closes the SGX DEBUG-attribute gap**.

&gt; ✅ **Security fix:** `SgxVerifier.registerInstance` now rejects any quote whose enclave report has
&gt; the SGX **DEBUG** attribute bit set (revert `SGX_DEBUG_ENCLAVE`). A debug enclave&#39;s memory —
&gt; including the in-enclave signing key in `reportData` — is host-readable/writable, so its quotes
&gt; must never be trusted on-chain. This was the original motivation.

## Why Taiko deploys its own entrypoint (security rationale)

Given Automata&#39;s deployed entrypoint is **owner-controlled** — `AutomataDcapAttestationFee` is
`Ownable`, and the owner&#39;s `setQuoteVerifier(version → verifier)` decides which contract validates a
quote — whoever holds that owner key could point version 3 at a permissive verifier and make **any**
quote pass: a **full attestation bypass** that would let attacker-controlled SGX instances register.
The owner can also set fees (`setBp`), which could DoS `registerInstance`.

So rather than trust the owner of Automata&#39;s deployed entrypoint, **Taiko deploys and owns its own
`AutomataDcapAttestationFee` + `V3QuoteVerifier`**, pointed at Automata&#39;s deployed **on-chain PCCS**.
PCCS is lower-trust: it serves Intel-signed collateral that the verifier checks by signature, so a
malicious/compromised PCCS can only *deny* (stale/missing collateral), not *forge* a passing
attestation. Net: Taiko controls the security-critical verifier wiring; Automata only supplies
signature-verified collateral.

## What changed

- **Dependencies (pinned npm, not vendored):** `@automata-network/automata-dcap-attestation@1.1.0`,
  `@automata-network/on-chain-pccs@1.1.1`, remapped to `node_modules` in `foundry.toml`. The
  attestation repo has no root `package.json`, so npm — not a `github:` dep — is the correct source.
- **Legacy removed:** deleted `contracts/layer1/automata-attestation/` and the tests that
  exclusively covered it.
[...]

PR #21827, opened 2026-06-23
feat(protocol): use audited upstream automata code with DEBUG enclaves rejection

dantaik:
## Summary

Migrates Taiko's SGX remote-attestation off the in-tree legacy `AutomataDcapV3Attestation` onto the current upstream **Automata DCAP attestation** (consumed as **pinned npm dependencies**), has Taiko **deploy and own its own attestation entrypoint**, and adds several **SGX-verifier security hardenings** (DEBUG rejection, fail-closed allowlist, strict per-network TCB policy, per-MRENCLAVE ATTRIBUTES pin, policy-change instance revocation, optional registrar, non-payable registration).

> Migration of #21817 onto the `taiko-alethia-protocol-v3.0.0` line, preserving v3-specific deltas (the `hekla-*` verifier envs in `configure_sgx_verifier.sh`, the `cache_path`/profile layout in `foundry.toml`, and v3's `DeployProtocolOnL1`/`gen-layouts` changes).

## Why Taiko deploys its own entrypoint (security rationale)

Automata's deployed entrypoint is **owner-controlled** — `AutomataDcapAttestationFee` is `Ownable`, and the owner's `setQuoteVerifier(version → verifier)` decides which contract validates a quote — so whoever holds that owner key could point version 3 at a permissive verifier and make **any** quote pass: a **full attestation bypass**. The owner can also set fees (`setBp`), which could DoS `registerInstance`.

So rather than trust the owner of Automata's deployed entrypoint, **Taiko deploys and owns its own `AutomataDcapAttestationFee` + `V3QuoteVerifier`**, pointed at Automata's deployed **on-chain PCCS**. PCCS is lower-trust: it serves Intel-signed collateral that the verifier checks by signature, so a malicious/compromised PCCS can only *deny* (stale/missing collateral), not *forge* a passing attestation.

### Migration
- **Dependencies (pinned npm, not vendored):** `@automata-network/automata-dcap-attestation@1.1.0`, `@automata-network/on-chain-pccs@1.1.1`, remapped to `node_modules` in `foundry.toml`. Collateral (TCBInfo/QEIdentity/CRL) is now **Intel-signature-verified via on-chain PCCS**, replacing the legacy contract's owner-uploaded-without-verification setters.
- **Legacy removed:** deleted `contracts/layer1/automata-attestation/` and the tests that exclusively covered it.
- **`registerInstance`:** raw-quote → `IDcapAttestation.verifyAndAttestOnChain`. An on-chain **verified-body hash check** binds the fields read from the raw quote (MRENCLAVE/MRSIGNER/DEBUG/reportData) to the report the entrypoint actually authenticated, so they can't be spoofed in calldata.
- **Deployment (two profiles):** `DeployAutomataDcapAttestation` (under `profile.layer1o`) deploys `V3QuoteVerifier(P256Verifier, PCCS_ROUTER)` + `AutomataDcapAttestationFee`; `DeployProtocolOnL1` / `DeployShastaContracts` consume it via `DCAP_ATTESTATION` and share it across both SGX instances.
- **`ConfigureSgxVerifier`:** configures the verifier directly; on-chain TCB/QE-identity config dropped (now from PCCS). `gen-layouts.sh`, the `configure_sgx_verifier.sh` wrapper and the verifiers `README` updated to match.

### SGX-verifier security hardening (added during review)
- **Reject DEBUG enclaves** (the original motivation; closes a previously exploited gap): any quote with the SGX **DEBUG** attribute set reverts `SGX_DEBUG_ENCLAVE`. A debug enclave's memory — incl. the in-enclave signing key in `reportData` — is host-readable/writable, so its quotes must never be trusted on-chain.
- **Fail-closed MRENCLAVE/MRSIGNER allowlist**, enforced by default (`checkLocalEnclaveReport = true`): no instance can register until the owner trusts the audited enclave identity.
- **Strict, per-network TCB-status policy.** An abstract **`SgxVerifier`** holds the shared logic; concrete subclasses pick the policy **in bytecode** (immutable — no admin can weaken it at runtime):
  - **`SecureSgxVerifier`** accepts only the up-to-date statuses (`OK`, `TCB_SW_HARDENING_NEEDED`, `TCB_CONFIGURATION_AND_SW_HARDENING_NEEDED`), and additionally enforces a **per-MRENCLAVE ATTRIBUTES pin** (every allowlisted measurement must declare the exact ATTRIBUTES profile it may register with, on top of the universal DEBUG/PROVISION_KEY/EINITTOKEN_KEY deny-mask).
  - **`InsecureSgxVerifier`** additionally accepts out-of-date statuses (`TCB_OUT_OF_DATE`, `TCB_OUT_OF_DATE_CONFIGURATION_NEEDED`) for **local devnet** liveness on hardware lagging on microcode/config.
  - Mainnet **and all public testnets (including Hoodi)** deploy the strict **`SecureSgxVerifier`** (secure by default); the lenient **`InsecureSgxVerifier`** is selected only by an explicit `useInsecureSgxPolicy` opt-in and must never run on a public testnet or mainnet. Rejecting out-of-date matches reference DCAP verifiers (Gramine flags out-of-date as `_INSECURE`; Intel permits it only within a bounded grace window).
- **Policy/allowlist changes revoke already-registered instances.** Each instance now records its attested **MRENCLAVE/MRSIGNER** and the per-MRENCLAVE **policy version** in force at registration, and `verifyProof` re-checks the *current* policy at proof time (a fixed ~6 cold SLOADs, **O(1) regardless of instance count**, ~7,696 gas). Untrusting an MRENCLAVE/MRSIGNER, or editing/removing a `SecureSgxVerifier` ATTRIBUTES pin (the version is bumped monotonically and **never reset**, so a remove-then-re-add can't silently re-enable), now **revokes every instance registered under the old policy** — invalidated, not deleted. The `registrar` can call `removeEnclaveAttributePolicy` to fail-close a compromised enclave. Owner-added instances (`addInstances`, no attestation) are exempt and revoked via `deleteInstances`. This closes the gap where removing a policy only blocked *future* registration while stale instances kept proving until expiry (the Daybreak audit finding, see below).
- **Optional `registrar`** (immutable, ported from #21830): when set, only it may call `registerInstance` (reverts `SGX_NOT_REGISTRAR`); `address(0)` keeps registration permissionless (the default). Defense-in-depth on top of the allowlist.
- **`registerInstance` is non-payable** and forwards no value to the (feeless, Taiko-owned) entrypoint — no stray ETH can be trapped in the verifier.

### Security review
A full analysis of whether permissionless registration + permissionless proving can be abused to forge proofs is in **[this comment](https://github.com/taikoxyz/taiko-mono/pull/21827#issuecomment-4775753416)**. Conclusion: it can't, because security rests on the **attestation** (allowlisted MRENCLAVE/MRSIGNER + non-DEBUG + acceptable TCB + Intel-signed collateral), not on who may call register — the registered key is generated inside the genuine enclave and never leaves SGX.

A scoped **Daybreak/Codex security audit** of `SgxVerifier.sol` + `SecureSgxVerifier.sol` is committed at `packages/protocol/contracts/layer1/verifiers/sgx-verifier-daybreak-audit.md`. Its single finding — *removing an enclave policy did not revoke already-registered instances* (medium) — is **resolved** by the policy-change revocation hardening above; the audit's out-of-scope Hoodi observation (public testnet deploying a lenient verifier) no longer applies now that all public testnets default to `SecureSgxVerifier`.

## Build & deploy — `via_ir` isolation

The Automata DCAP verifier (`QuoteVerifierBase` X.509/ASN.1 parsing) needs the IR pipeline to compile, but enabling `via_ir` on `profile.layer1` alters codegen in a way that breaks unrelated Shasta tests. So:
- The only file importing the Automata concrete contracts is `DeployAutomataDcapAttestation`.
- `profile.layer1` stays non-IR and `skip`s that one script → the standard L1 build/test compiles without IR and the Shasta suite stays green.
- `DeployAutomataDcapAttestation` is compiled/run under `profile.layer1o`, which keeps `via_ir` and writes to an isolated `out/layer1o` so the IR artifacts never clobber the non-IR build.

## Local verification
- `FOUNDRY_PROFILE=layer1 forge build` — clean (only pre-existing lint warnings).
- `FOUNDRY_PROFILE=layer1o forge build` — compiles the DCAP deploy script + Automata contracts under `via_ir`.
- `FOUNDRY_PROFILE=layer1 forge test --match-path 'test/layer1/verifiers/*'` — green (**174 tests**); covers registration, DEBUG rejection, **both** TCB policies (Secure rejects / Insecure accepts out-of-date), the MRENCLAVE/MRSIGNER allowlist, the per-MRENCLAVE ATTRIBUTES pin, registrar gating, proof verification, and **policy-removal / allowlist-change revocation at proof time**.
- Full `layer1` suite green; gas snapshot refreshed.

## Follow-ups
- [ ] Off-chain prover/client must send the **raw** DCAP quote to `registerInstance(bytes)` (interface change).
- [ ] Operational for mainnet: keep `checkLocalEnclaveReport == true`; trust only the audited MRENCLAVE/MRSIGNER; owner = multisig/governance (the owner can `addInstances` without attestation and toggle the allowlist).


dantaik:
## Does the upstream release fix any bug in the in-tree (legacy) Automata contract?

**Yes — most importantly, the legacy contract trusts attestation collateral that is never signature-verified on-chain.**

### Bug — TCBInfo / QEIdentity / CRL collateral is owner-supplied, not Intel-signed

The in-tree `AutomataDcapV3Attestation` (a fork of `automata-network/automata-dcap-v3-attestation`) loads **every** piece of Intel collateral the verifier depends on from plain `onlyOwner` setters, with **no check that Intel actually signed it**:

| collateral | setter | stored as |
|---|---|---|
| TCBInfo | `configureTcbInfoJson` | `tcbInfo[fmspc] = tcbInfoInput` |
| QE identity | `configureQeIdentityJson` | `qeIdentity = qeIdentityInput` |
| Cert revocation (CRL) | `addRevokedCertSerialNum` / `removeRevokedCertSerialNum` | `serialNumIsRevoked[...] = true/false` |

This collateral drives `_checkTcbLevels`, `_verifyQEReportWithIdentity`, `_verifyCertChain` and `_attestationTcbIsValid`. Because the TCB Signing certificate's signature over the collateral is **never verified on-chain**, a compromised or careless owner can upload forged collateral — e.g. mark a revoked/out-of-date TCB as `OK`, whitelist an arbitrary Quoting Enclave, or un-revoke a certificate — and silently forge every SGX attestation. Automata's own writeup of the v3 design states this plainly:

> the manually uploaded collaterals lack TCB signature verification, rendering them insecure. Any compromise of the contract owner jeopardizes all verification.

### Fix in the upstream release

Automata replaced owner-configured collateral with **on-chain PCCS**, which cryptographically verifies Intel's signatures on the collateral before it can be used.

- **Introduced upstream in [automata-dcap-v3-attestation#8](https://github.com/automata-network/automata-dcap-v3-attestation/pull/8)** — *"On-Chain PCCS and RiscZero Integration"* (merged 2024-05-27): replaces statically configured TCB info and QE identity with dynamically verified on-chain PCCS data that cryptographically validates Intel's signatures on attestation collateral.
- **Productized in the release this PR migrates to:** `@automata-network/automata-dcap-attestation@1.1.0` + `@automata-network/on-chain-pccs@1.1.1` — Trail of Bits–audited ([report](https://github.com/trailofbits/publications/blob/master/reviews/2025-02-automata-dcap-attestation-onchain-pccs-securityreview.pdf)); v1.1 also carries the OpenZeppelin-audit PCCS-Router timestamp-validity remediation.

Our in-tree copy was forked from `main` + the `parse-quote-offline` gas branch and **does not include the PR #8 PCCS integration** (its collateral is still the owner-trusted setters above), so the gap exists in the legacy code and is closed by this migration.

### For completeness — already-applied upstream fixes (not outstanding in our copy)

Two earlier upstream correctness fixes are already present in the in-tree code, so they are *not* bugs we still carry:
- notBeforeTag ASN.1 tag check — `PEMCertChainLib` already accepts both `0x17`/`0x18` — [#9](https://github.com/automata-network/automata-dcap-v3-attestation/pull/9)
- PCK issuer-name accepts both *Platform CA* and *Processor CA* — [#1](https://github.com/automata-network/automata-dcap-v3-attestation/pull/1) (merged via #3)

> Note: the legacy contract also accepts `TCB_OUT_OF_DATE` / `TCB_OUT_OF_DATE_CONFIGURATION_NEEDED` in `_attestationTcbIsValid`. That is a deliberate liveness policy rather than an upstream bug — and in this PR that acceptance policy is moved into and enforced by `SgxVerifier` directly.

---

dantaik:
# SGX verifier security review — can permissionless `registerInstance` be abused to forge proofs?

**Short answer: No — provided the MRENCLAVE/MRSIGNER allowlist stays enforced and correctly configured.** Permissionless registration is safe *by design*, because the security comes from the **attestation**, not from gating who may call `registerInstance`. There is **one policy change I'd make before mainnet** (stop accepting `TCB_OUT_OF_DATE`), plus a short operational checklist.

## Why "anyone can call register" is not the attack surface

A forged proof needs a signature from a key the attacker controls that is **registered** as an instance. `registerInstance(rawQuote)` only registers a key if the quote simultaneously:

1. is authentic (Intel-signed, chained to the Intel SGX root, with current TCB/QE collateral) — checked by the Taiko-owned Automata entrypoint;
2. is **not** a DEBUG enclave;
3. has an **acceptable TCB status**;
4. carries the **allowlisted MRENCLAVE + MRSIGNER** (the exact Taiko prover enclave); and
5. embeds the instance address in `reportData`, **bound** to the verified report bytes.

The registered key is whatever the enclave wrote into `reportData[0:20]`. For the allowlisted MRENCLAVE/MRSIGNER that enclave is the genuine Taiko prover, which **generates its keypair inside SGX and never exposes the private key**. So whoever *submits* the quote neither learns nor controls that key — they just register the enclave's own key. Gating the caller (the new `registrar`) changes *who submits*, not *whose key is registered* — which is why permissionless registration is safe and `registrar` is defense-in-depth, not the primary control.

The attacker's only routes to register a key they control are to **break the allowlist** — run a *different* enclave (different MRENCLAVE) that leaks a known key, or disable the check. The controls below prevent that.

## Control-by-control verification (`SgxVerifier.sol`)

| # | Property / parameter | Where | Verdict |
|---|---|---|---|
| 1 | Quote authenticity (Intel chain, TCB, collateral) | `verifyAndAttestOnChain(_rawQuote)` → `require(verified)` | ✅ delegated to the Taiko-owned, ToB-audited Automata entrypoint + on-chain PCCS. **External trust dependency** (risk #3). |
| 2 | **Verified-body binding** | `keccak256(output[11:395]) == keccak256(_rawQuote[48:432])` | ✅ Sound. Every field below is read from `_rawQuote[48:432]`, proven byte-equal to the entrypoint-authenticated report. All offsets (attributes 48, MRENCLAVE 64, MRSIGNER 128, reportData 320) lie inside that range. |
| 3 | DEBUG rejection | `(_rawQuote[HEADER+48] & 0x02) == 0` | ✅ Correct (DEBUG = bit 1 of the little-endian ATTRIBUTES low byte). Debug-enclave memory — incl. the key in `reportData` — is host-readable, so this is essential. |
| 4 | **MRENCLAVE + MRSIGNER allowlist** | `if (checkLocalEnclaveReport) require(trustedUserMrEnclave[..] && trustedUserMrSigner[..])` | ✅ **The load-bearing control.** Requires *both* (AND). Default `true` (fail-closed). |
| 5 | TCB-status policy | `_isTcbStatusAccepted` | ⚠️ Accepts `OK`, `SW_HARDENING_NEEDED`, `CONFIG_AND_SW_HARDENING_NEEDED`, **`OUT_OF_DATE`**, **`OUT_OF_DATE_CONFIGURATION_NEEDED`**; rejects `CONFIGURATION_NEEDED`, `REVOKED`, `UNRECOGNIZED`. See risk #1. |
| 6 | Instance address binding | `address(bytes20(_rawQuote[HEADER+320:+20]))` | ✅ read from the bound region; `address(0)` rejected in `_addInstances`. |
| 7 | Re-registration / dedup | `addressRegistered` → `SGX_ALREADY_ATTESTED` | ✅ same address can't be re-registered to dodge expiry. |
| 8 | Validity window | `validSince <= now <= validSince + INSTANCE_EXPIRY (365d)`; `INSTANCE_VALIDITY_DELAY = 0` | ✅ expiry enforced; delay 0 = instant validity (risk #5). |
| 9 | Proof binding (proving) | `hashPublicInputs(aggHash, address(this), instance, taikoChainId)` + `ECDSA.recover == instance` + `_isInstanceValid` | ✅ bound to this verifier + chainId (no cross-chain/-verifier replay); permissionless proving is safe because it still needs the instance's private key. |

## Residual risks & recommendations (ranked)

**1. 🔴 Accepting `TCB_OUT_OF_DATE` on mainnet — my top recommendation to change.** An out-of-date TCB means the platform is missing microcode updates, and those updates are exactly what patch the disclosed SGX key-extraction classes (Foreshadow / SGAxe / ÆPIC-family). On such a platform the genuine enclave's signing key may be **extractable** — which *is* a way to obtain a controllable key for an allowlisted MRENCLAVE and forge proofs. This is the one path that bypasses the otherwise-sound model. For mainnet I'd **reject `OUT_OF_DATE`** (accept only `OK`, at most `SW_HARDENING_NEEDED`). The current set is also internally inconsistent — it rejects `CONFIGURATION_NEEDED` but accepts the strictly-worse `OUT_OF_DATE`. (Inherited verbatim from the old `AutomataDcapV3Attestation`, so it's a policy to revisit, not a regression.)

**2. 🟠 Operational invariants that MUST hold.** Permissionless registration is only as strong as:
   - `checkLocalEnclaveReport` stays **`true`** — if toggled off, *any* genuine SGX user can register a custom enclave with an attacker-known key → instant forgery. Treat `toggleLocalReportCheck()` as break-glass that is never on in production.
   - `trustedUserMrEnclave` / `trustedUserMrSigner` hold **only** the audited Taiko prover values (minimal set).

**3. 🟠 External trust in the Automata entrypoint + PCCS.** The verifier delegates *all* quote/TCB/collateral authenticity to `automataDcapAttestation`; its correctness and keeping PCCS collateral current (incl. CRLs) are assumed. Mitigated by the ToB-audited v1.1.0 / on-chain-pccs v1.1.1 and Taiko owning the entrypoint — but it's the single biggest dependency. The binding (#2) holds only if the entrypoint returns the verified report verbatim in `output`; worth a one-line confirmation against Automata's `Output` spec.

**4. 🟠 Owner is a full trust root.** Owner-only `addInstances()` registers an instance with **no attestation at all**, and the owner can rewrite the allowlist or disable it. Use a secure multisig / governance. (`registrar` is correctly *more limited* — it can only submit attested quotes, never bypass attestation.)

**5. 🟡 `INSTANCE_VALIDITY_DELAY = 0`.** A new instance is valid immediately, leaving no window to spot and `deleteInstances()` a rogue registration before it proves. A small non-zero delay adds reaction time (defense-in-depth).

**6. 🟢 Consider setting a `registrar` for mainnet.** Not required if the allowlist is correct, but with this PR the feature exists: pointing `registrar` at a Taiko-controlled address removes public registration entirely (kills spam/grief, shrinks the surface to a known submitter) while the allowlist stays the cryptographic backstop.

## Mainnet go-live checklist
- [ ] `checkLocalEnclaveReport == true`; only the audited MRENCLAVE/MRSIGNER trusted.
- [ ] Decide TCB policy — recommend dropping `OUT_OF_DATE` / `OUT_OF_DATE_CONFIGURATION_NEEDED`.
- [ ] Owner = multisig/governance; document that `addInstances` / `toggleLocalReportCheck` are privileged.
- [ ] Automata entrypoint + PCCS deployed, Taiko-owned, collateral/CRLs current; confirm `output` body == verified report.
- [ ] (Optional) set a `registrar`; (optional) non-zero `INSTANCE_VALIDITY_DELAY`.

**Bottom line:** the design is production-sound — a malicious user **cannot** forge proofs via permissionless registration as long as the allowlist is enforced and correctly set, because the cryptographic gate is the *attestation*, not the caller. The one change I'd make before mainnet is tightening the TCB policy to stop trusting out-of-date platforms.

---

davidtaikocha:
Re the attestation **freshness** gap that's still open after #21822:

#21822 only adds the DEBUG/PROVISION_KEY attribute reject + registrar gate; it leaves the legacy in-tree `AutomataDcapV3Attestation` untouched, which has **no collateral-freshness enforcement** — collateral (TCBInfo/QEIdentity) is owner-uploaded *without* Intel-signature verification, there's no on-chain `nextUpdate`/expiry field, the CRL is a static owner-maintained set, and the TCB gate accepts `OUT_OF_DATE`. Net effect: a DCAP quote is effectively a bearer token that never goes stale on-chain — the protocol can't reject it once the attested platform becomes known-bad (post-disclosure CVE, revoked PCK cert, decommissioned/sold hardware). Worst case is a **temporal replay**: a captured genuine, non-debug, allowlisted-MRENCLAVE quote whose signing key later becomes extractable gets registered as a fresh 365-day signing instance and used to forge proofs → fake L2 fork → drain.

**This PR fixes that.** Migrating to upstream Automata v1.1 (`automata-dcap-attestation@1.1.0` + `on-chain-pccs@1.1.1`) means collateral is Intel-signature-verified via on-chain PCCS, and `PCCSRouter` enforces the `issuedAt`/`nextUpdate` validity window (fail-closed: `TcbEvalExpiredOrNotFound` / `FmspcTcbExpiredOrNotFound` / `QEIdentityExpiredOrNotFound` / `CrlExpiredOrNotFound`), so stale/expired collateral and revoked certs get rejected once fresh collateral is on-chain. The strict immutable `MainnetSgxVerifier` (accepts only `UpToDate`/`SWHardeningNeeded`) also closes the `OUT_OF_DATE`-TCB acceptance path in the same move.

Two caveats worth noting for completeness:
- This is **collateral** freshness, not a per-quote nonce — non-interactive RA has no verifier challenge, so per-quote anti-replay remains the `reportData`→instance-address binding. It shrinks the temporal-replay window hard but doesn't make a quote provably "live".
- Freshness now depends on **keeping the on-chain PCCS collateral refreshed** (~30-day `nextUpdate` cadence); if it lapses it fails **closed** (a liveness/DoS concern, not a forgery one). Worth an ops owner — the Automata DCAP Dashboard automates this upkeep.

dantaik:
### Verification: external finding "Disabled SGX enclave-identity checks let arbitrary enclaves become proof signers" (high) is remediated by this PR

That finding was filed against the now-deleted in-tree `AutomataDcapV3Attestation.sol`, where the MRENCLAVE/MRSIGNER allowlist was gated behind `checkLocalEnclaveReport`, which defaulted to **false** (uninitialized storage bool; `init()` never set it). I traced the current code (head `f4ec315`) and the reported attack no longer holds:

- **Allowlist defaults ON, fail-closed.** `SgxVerifier.sol:143` sets `checkLocalEnclaveReport = true` in the **constructor**, and this is now a **non-upgradeable** contract (constructor, not initializer; deployed directly via `new SecureSgxVerifier(...)`), so the secure default always holds at deployment — it can't be left unset like the old upgradeable contract.
- **Registration fails closed.** With the check on, `registerInstance` requires `trustedUserMrEnclave[mrEnclave] && trustedUserMrSigner[mrSigner]` (`SgxVerifier.sol:276-283`); both mappings default to false, so no instance can register until the owner allowlists the audited identity — even with the default permissionless `registrar == address(0)`. The reported path (put attacker signer in `reportData` → register → sign proofs) now reverts `SGX_INVALID_ATTESTATION` at registration.
- **Plus** the verified-body hash binding, DEBUG-enclave rejection, and mainnet's SGX+ZK composition as additional backstops.

Confirmed empirically: `forge test --match-path test/layer1/verifiers/SgxVerifier.t.sol` → **44/44 pass**, including the two checks the report recommended (`test_registerInstance_RevertWhen_AllowlistOnAndMrEnclaveUntrusted` / `...MrSignerUntrusted`) and `test_constructor_enablesLocalReportCheckByDefault`.

Residual (not a vuln): the owner can still `toggleLocalReportCheck()` to disable enforcement, but that's now a deliberate privileged action emitting `LocalReportCheckToggled` — not the insecure-by-default posture the finding flagged.

---

davidtaikocha (approved):

smtmfft (approved):

PR #19443, opened 2025-05-13
feat(protocol): remove quota manager

dantaik:
The bridge has been in production for about a year. I suggest removing the quota manager to allow frictionless withdrawals.

PR #21828, opened 2026-06-23
feat(protocol): restore QuotaManager for ERC20Vault withdrawal quota

dantaik:
Restores the `QuotaManager` that was removed in #19443 and re-integrates it into the v3.0.0 line. Originally migrated from #21811.

## Base branch
This PR targets `codex/restore-bridge-eth-quota`, which independently restores the **bridge Ether** withdrawal quota. Both that branch and this one fork from the same `taiko-alethia-protocol-v3.0.0` tip, so they reconcile without merge conflicts.

- The **Files changed** tab shows the full 21-file set (+797/−16) because the merge-base is the shared v3.0.0 tip.
- The **net delta vs the codex base is 9 files / +326−8**: on top of codex's bridge-ETH-only work, this PR adds the **`ERC20Vault` token quota** integration, the **`DeployQuotaManager.s.sol`** deploy script, and the **`QuotaManager`** + **`ERC20Vault_quota`** test suites.

Note: in the v3.0.0 line the `Bridge` no longer implements `IEthMinter` (that interface does not exist here), so the restored import block adds **only** `IQuotaManager`.

## Quota integration
- Both `Bridge` (Ether) and `ERC20Vault` (ERC20) consume quota through an **immutable** `quotaManager` set in the constructor — no resolver lookup. The check is a no-op while `quotaManager == address(0)`.
- `QuotaManager` remains non-upgradeable; `consumeQuota` reverts on failure (`QM_OUT_OF_QUOTA`, `QM_PERMISSION_DENIED`).

## Debit only on actual release
Quota is debited exactly for funds that actually leave the bridge, fixing a double-debit where a failed-then-retried withdrawal consumed the value quota twice:
- `processMessage` debits `value + fee` only when the message reaches `DONE`, and `fee` only when it stays `RETRIABLE` (value remains in the bridge).
- `retryMessage` debits `value` only on a successful retry.
- `recallMessage` debits `value` (always released back to the source owner).
- `ERC20Vault` debits on release: `_consumeTokenQuota` runs in the same atomic call that transfers/mints the tokens, so a reverted delivery releases and debits nothing.

## Tests
- Restored/updated `QuotaManager` and `Bridge2_quota` suites (the latter asserts the raw `QM_OUT_OF_QUOTA` selector).
- `Bridge2_quotaAccounting` covers the Ether lifecycle; `ERC20Vault_quota` covers token release; new shared `CountingQuotaManager` test helper.

### Verification (Foundry v1.4.2)
- `compile:shared`, `compile:l1`, `compile:l2`, genesis build: ✅
- Quota suites: 16 passed (`QuotaManager`, `Bridge2_quota`, `Bridge2_quotaAccounting`, `ERC20Vault_quota`)
- Full shared bridge suite: 50 passed · full shared vault suite: 46 passed · layer2 `DelegateController`: 2 passed — no regressions

https://claude.ai/code/session_01TnurXsa1dDGcDu37fWKonB


packages/protocol/script/layer1/mainnet/DeployQuotaManager.s.sol
```diff
@@ -0,0 +1,48 @@
...
+        address bridge = IResolver(resolver).resolve(block.chainid, LibNames.B_BRIDGE, false);
+        address erc20Vault =
+            IResolver(resolver).resolve(block.chainid, LibNames.B_ERC20_VAULT, false);
+
+        QuotaManager qm = new QuotaManager(msg.sender, bridge, erc20Vault, 15 minutes);
+        console2.log("QuotaManager deployed:", address(qm));
+
+        // Config L2-to-L1 quota
+        uint104 value = 200_000; // USD
+        uint104 priceETH = 3100; // USD
```
dantaik:
```suggestion
        uint104 priceETH = 1750; // USD
```

packages/protocol/script/layer1/mainnet/DeployQuotaManager.s.sol
```diff
@@ -0,0 +1,48 @@
...
+        address erc20Vault =
+            IResolver(resolver).resolve(block.chainid, LibNames.B_ERC20_VAULT, false);
+
+        QuotaManager qm = new QuotaManager(msg.sender, bridge, erc20Vault, 15 minutes);
+        console2.log("QuotaManager deployed:", address(qm));
+
+        // Config L2-to-L1 quota
+        uint104 value = 200_000; // USD
+        uint104 priceETH = 1750; // USD
+        uint104 priceTKO = 5; // USD
```
dantaik:
```suggestion
        uint104 priceTKO = 1; // USD
```

davidtaikocha (approved):

davidtaikocha (approved):

PR #21833, opened 2026-06-23
fix(protocol): add Proposal0017 L1 recovery bundle

ggonzalez94:
Adds the Proposal0017 L1 recovery bundle for the Shasta incident response. The proposal upgrades SignalService, Bridge, ERC20Vault, and Inbox, disables the verified retriable message hashes, restores Inbox core state from the last pre-forgery block, and removes stale verifier trust from the existing RISC0, SP1, and SGX attesters.

The implementation, QuotaManager, and verifier addresses are now the real mainnet contracts deployed by `DeployHackRecoveryContracts` (chain 1, commit `b73608696`). They were fact-checked against live L1:

- the new immutable QuotaManager is deployed, owned by `admin.taiko.eth`, with the ETH/WETH/TKO/USDT/USDC recovery quotas set in its constructor (period `86400`s);
- the new Bridge and ERC20Vault implementations are deployed with that QuotaManager address;
- the new MainnetInbox implementation is deployed with the new MainnetVerifier and disabled `saveForcedInclusion` behavior;
- the new MainnetVerifier points at the newly deployed SGX-geth and SGX-reth verifier contracts and reuses the existing RISC0/SP1 verifiers.

Codediff links for every upgraded/replaced contract (Etherscan for the net-new QuotaManager and libraries, which have no predecessor to diff), plus reproducible build steps for the raiko2 v0.5.0 ZK image IDs and the raiko hotfix-1.16.1-2 SGX MRENCLAVE values, are in `Proposal0017.md`. All values in the doc were re-verified verbatim against those releases and against on-chain state.

This proposal does not upgrade QuotaManager and does not call `updateQuota`; the agreed quota values are constructor inputs on the new immutable QuotaManager.

Note: proving stays halted after execution until SGX instances are registered on the new verifiers (`nextInstanceId() == 0` on both) and proposals 18052+ are re-proven; that registration is a separate transaction.

Validated with `FOUNDRY_PROFILE=layer1 forge test --match-path test/layer1/proposals/Proposal0017.t.sol -vv`.

PR #21838, opened 2026-06-23
fix(protocol): remove _nextProposalId and _lastProposalBlockId from Inbox.init2

davidtaikocha:
## Summary

Removes the `_nextProposalId` and `_lastProposalBlockId` parameters from `Inbox.init2` (the one-time owner recovery function). Instead of overwriting these fields, `init2` now **preserves** them from the live core state and only resets the finalized pointer, timestamps, and last finalized block hash.

New signature:

```solidity
function init2(uint48 _lastFinalizedProposalId, bytes32 _lastFinalizedBlockHash)
```

## Why

Per [@davidtaikocha's review on #21833](https://github.com/taikoxyz/taiko-mono/pull/21833#discussion_r3457800252):

> CB did submit some new proposals after the forged proof transaction, so we should use the latest `coreState.nextProposalId` (`18118`), or just remove the `_nextProposalId` param in `init2`. Otherwise, we will have proposals with the same ID on-chain. And if we decided to open a PR to remove params in `init2`, I feel like we can remove `_lastProposalBlockId` too (but keep it does no harm).

### `nextProposalId` — the real reason (event stream)

Legitimate proposals were submitted after the forged-proof incident, so the live `nextProposalId` already advanced past the pre-incident value (e.g. `18118` vs `18059`), each emitting a `Proposed(uint48 indexed id, ...)` event for IDs `18059..18117`.

If recovery reset `nextProposalId` back to the pre-incident value, resumed proposing would emit **another `Proposed` event for an `id` that was already emitted** by those post-incident proposals. The result is more than one `Proposed` event with the same ID in the inbox event stream. Off-chain consumers replay this stream keyed on the (indexed) proposal `id` — the driver's derivation pipeline and the event indexer cannot reconcile two different proposals sharing one ID.

Preserving the live `nextProposalId` keeps the `Proposed` event stream monotonic and unambiguous.

### `lastProposalBlockId` — a separate, low-stakes reason

`lastProposalBlockId` is **not** related to the event-stream issue above. It is read in exactly one place: the one-proposal-per-L1-block guard `require(block.number > lastProposalBlockId)` in `propose` ([Inbox.sol:576](packages/protocol/contracts/layer1/core/impl/Inbox.sol#L576)), and it appears in no event, hash, proof, or L2 derivation.

So preserving the live value is simply harmless: it is a past L1 block, so the guard passes immediately when proposing resumes. Even if it were reset to the recovery block instead, the only consequence would be waiting one extra L1 block before the first post-recovery proposal. Either way there is no need to take it as a param or reset it — so it is dropped and left untouched.

## Changes

- `init2` drops from 4 params to 2.
- `nextProposalId` and `lastProposalBlockId` are read from `_coreState` and left untouched.
- Only `lastFinalizedProposalId`, `lastFinalizedTimestamp`, `lastCheckpointTimestamp`, and `lastFinalizedBlockHash` are reset.
- Dropped the validation `require`s that only guarded the removed parameters.
- Updated `test/layer1/core/inbox/InboxInit2.t.sol` to the new signature: asserts `nextProposalId`/`lastProposalBlockId` are preserved, replaces the removed-param tests with `RevertWhen_InboxNotActivated`, and exercises the unfinalized-range check via `vm.store`. All 9 tests pass.

## Notes

- `init2` is not part of the `IInbox` interface (impl-only), so no interface change is needed.
- The only on-branch caller of `Inbox.init2` is the test above; the `Proposal0017` recovery script that also calls it lives in #21833 (which targets `main`) and will need its `init2(...)` call updated when these branches converge.


packages/protocol/contracts/layer1/core/impl/Inbox.sol
```diff
@@ -225,32 +225,22 @@ contract Inbox is IInbox, ICodec, IForcedInclusionStore, IBondManager, Essential
...
-        _coreState = CoreState({
-            nextProposalId: _nextProposalId,
-            lastProposalBlockId: _lastProposalBlockId,
-            lastFinalizedProposalId: _lastFinalizedProposalId,
-            lastFinalizedTimestamp: uint48(block.timestamp),
-            lastCheckpointTimestamp: uint48(block.timestamp),
-            lastFinalizedBlockHash: _lastFinalizedBlockHash
-        });
+        currentState.lastFinalizedProposalId = _lastFinalizedProposalId;
+        currentState.lastFinalizedTimestamp = uint48(block.timestamp);
```
YoGhurt111, line 236: Shall we use `1782067079` in the tx https://etherscan.io/tx/0x15eedbbabe676aaba36456ecdf5330418502dc148f710f7f1d575fed79dce203 as well?

dantaik (approved):

dantaik:
I'd prefer to keep as much on-chain state data if possible, so LGTM, approved.

YoGhurt111 (approved):

ggonzalez94 (approved):

PR #21840, opened 2026-06-23
fix(protocol): enforce SGX enclave identity (MRENCLAVE) for instance registration

dantaik:
## Summary

Fixes **[3] Disabled SGX enclave-identity checks let arbitrary enclaves become proof signers** (high severity, CWE-345), reported by OpenAI in https://github.com/taikoxyz/taiko-mono/pull/21839

`AutomataDcapV3Attestation` only verified the application enclave's `MRENCLAVE`/`MRSIGNER` when the `checkLocalEnclaveReport` flag was on. That flag **defaults to `false` and is never enabled** on deployments — `Proposal0004` configures `setMrEnclave(...)` for both the SGX-geth and SGX-reth attesters but never calls `toggleLocalReportCheck()` (and never sets `MRSIGNER`). Combined with permissionless `registerInstance` (`registrar == address(0)` in all deploy scripts), an SGX-capable attacker could place their own signer in `reportData`, pass quote-only attestation, and register as a proof-signing instance — defeating the entire purpose of remote attestation.

## Why the fix is sound

`localEnclaveReport.mrEnclave`, `.mrSigner`, `.attributes` and `.reportData` are all authenticated by the Intel-rooted ECDSA quote signature (`V3Parser.validateParsedInput` → `packQEReport` → Step 8 sig check in `_verifyParsedQuote`). So once `verifyParsedQuote` returns true, reading `mrEnclave`/`mrSigner` off the struct is trustworthy — exactly the trust the code already places in `reportData` (the signer) and the existing DEBUG/PROVISION `attributes` check.

## Changes

**Layer 1 — root cause (`AutomataDcapV3Attestation`)**
- `MRENCLAVE` is now **always enforced**: a quote whose measurement isn't on the allowlist is rejected, failing closed when the allowlist is empty.
- `MRSIGNER` is **supplementary**: enforced only once at least one trusted signer is configured (tracked via `trustedMrSignerCount`), so the current MRENCLAVE-only mainnet config is not bricked.
- Removed the `toggleLocalReportCheck` footgun and its event. The `bool` storage slot is retained (renamed, deprecated) to preserve the upgradeable layout — verified slot-for-slot; the new `trustedMrSignerCount` consumes one slot from the former gap (`__gap` 39→38).

**Layer 2 — defense in depth (`SgxVerifier`)**
- `registerInstance` independently requires the attested `MRENCLAVE` to be trusted via the new `IAttestation.isMrEnclaveTrusted`, so a misconfigured/non-enforcing attestation can never register an arbitrary enclave. New error `SGX_UNTRUSTED_ENCLAVE`.

**Tests**
- `SgxVerifier.t.sol`: `test_registerInstance_RevertWhen_EnclaveUntrusted` (verifier guard).
- `AutomataDcapV3AttestationTest.t.sol`: mandatory-MRENCLAVE rejection, optional-MRSIGNER semantics, `trustedMrSignerCount` invariant, and an end-to-end registration invariant over the real attestation + a real SGX verifier.
- Updated the attestation test harness so each quote trusts its own measurement (the prior tests relied on the check being off).

**Scripts**: removed the now-defunct `--toggle-check` plumbing.

## Verification

- `forge test` — `test/layer1/verifiers/*` (67) and `test/layer1/automata-attestation/*` (16) all pass.
- Storage layout confirmed upgrade-compatible (`forge inspect` slot-by-slot).
- `forge fmt` applied; full `forge build` (layer1) clean.

> [!NOTE]
> The design (enforce in both layers; MRENCLAVE mandatory, MRSIGNER optional) was chosen explicitly to avoid bricking the live deployment, which pins identity via MRENCLAVE only. If the team prefers **requiring** MRSIGNER too, that's a one-line addition to the deployment proposals plus flipping the attestation policy.

---
_Generated by [Claude Code](https://claude.ai/code/session_013ewM4bAAVVRC4grXXDVmWQ)_


smtmfft:
Thanks for putting this together. My current read is that this should not be treated as a necessary fix for the current incident/recovery path.

On mainnet, our SGX attesters already have `checkLocalEnclaveReport` enabled, so the condition described here, arbitrary SGX enclave registration because local enclave identity checks were disabled, does not match the current deployment. We also have not seen evidence that arbitrary SGX enclaves were able to register through this path.

This change can make sense as defense-in-depth for a fresh attester deployment, but our plan is to move to the latest Automata release, where this hardening is already covered. Given that, I would prefer not to carry a local modification to `AutomataDcapV3Attestation` in this branch, especially since it changes the attestation policy/interface/storage surface.

For the current recovery, the required fix remains the SGX registration-side rejection of unsafe application-enclave attributes, especially DEBUG/PROVISION. I would suggest closing this PR, or reframing it as optional hardening outside the immediate recovery branch if we still want to keep a local variant.

dantaik:
Thanks @smtmfft — agreed, closing.

To confirm the reasoning for the record:
- If `checkLocalEnclaveReport` is already enabled on the mainnet attesters (you have the operational ground truth here; the repo proposals set MRENCLAVE but don't toggle the check, so it was presumably enabled out-of-band), then enclave identity is already enforced and this is **defense-in-depth, not a recovery necessity**.
- The recovery-relevant SGX hardening — rejecting unsafe application-enclave attributes (DEBUG/PROVISION) — is already on the branch via #21822, independent of this change.
- The broader enclave-identity hardening will be covered by moving to the upstream Automata release, so a local fork of `AutomataDcapV3Attestation` (policy/interface/storage surface) isn't worth carrying here.

Closing as not needed for the recovery branch. Happy to revive this as a standalone optional-hardening PR if useful once the Automata upgrade path is settled.

---

PR #21841, opened 2026-06-23
fix(protocol): prevent quota issuance overflow in QuotaManager

dantaik:
## Problem

Follow-up to #21821. In `QuotaManager.availableQuota`, the issuance calculation could overflow:

```solidity
uint256 issuance = q.quota * (block.timestamp + _leap - q.updatedAt) / quotaPeriod;
```

`block.timestamp - q.updatedAt` grows **without bound** over time (and `_leap` is a caller-supplied `uint256` on the public `availableQuota`). With a sufficiently large quota and elapsed time, the multiplication `q.quota * (...)` can exceed `uint256`, causing the call to revert. Since `consumeQuota` calls `availableQuota` on every release, an overflow would brick the quota mechanism and **halt the bridge**.

It is a minor risk in practice (timestamps grow slowly and quotas are unlikely to be configured that high), but the calculation is fragile and worth hardening.

## Fix

Cap the elapsed time at `quotaPeriod`:

```solidity
uint256 elapsed = (block.timestamp + _leap - q.updatedAt).min(quotaPeriod);
uint256 issuance = q.quota * elapsed / quotaPeriod;
return (issuance + q.available).min(q.quota);
```

This is **mathematically equivalent** to the original: once a full period has elapsed the quota is fully restored, and the result is already capped at `q.quota` by the final `.min(q.quota)`. Capping `elapsed` therefore never changes the returned value, while bounding the multiplication to `q.quota * quotaPeriod` (≤ 2¹⁰⁴ × 2²⁴ = 2¹²⁸), which can never overflow `uint256`.

## Tests

- `test_quota_manager_restores_fully_after_long_period` — quota fully restores and `consumeQuota` keeps working after warping ~100 years past the period.
- `test_quota_manager_no_overflow_with_max_quota_and_large_elapsed` — with `quota == type(uint104).max` and a large elapsed time, `availableQuota` returns the full quota instead of reverting. This test **panics with `arithmetic overflow (0x11)` on the pre-fix code** and passes with the fix, confirming it captures the reported bug.

All existing `QuotaManager` and `Bridge2_quota` tests continue to pass.

https://claude.ai/code/session_01GTM1Zt7VL5wwg9vvShtdvL

---
_Generated by [Claude Code](https://claude.ai/code/session_01GTM1Zt7VL5wwg9vvShtdvL)_

### DIFF
--- a/packages/protocol/contracts/shared/bridge/QuotaManager.sol
+++ b/packages/protocol/contracts/shared/bridge/QuotaManager.sol
@@ -81,19 +81,26 @@ contract QuotaManager is Ownable2Step, IQuotaManager {
 
     /// @notice Returns the available quota for a given token.
     /// @param _token The token address with Ether represented by address(0).
-    /// @param _leap Amount of seconds in the future.
+    /// @param _leap Number of seconds in the future to look ahead. Values greater than or equal
+    /// to `quotaPeriod` are treated as a full period (the quota is fully restored), so arbitrarily
+    /// large values are safe and never overflow.
     /// @return The available quota.
     function availableQuota(address _token, uint256 _leap) public view returns (uint256) {
         Quota memory q = tokenQuota[_token];
         if (q.quota == 0) return UNLIMITED_QUOTA;
         if (q.updatedAt == 0) return q.quota;
 
-        // Cap the elapsed time at `quotaPeriod`. Once a full period has passed the quota is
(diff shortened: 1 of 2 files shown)


davidtaikocha (approved):

dantaik:
@review

YoGhurt111 (approved):

PR #21845, opened 2026-06-23
feat(protocol): pin SGX enclave ATTRIBUTES per-MRENCLAVE in SecureSgxVerifier

dantaik:
## What

Hardens permissionless SGX instance registration so a *debug-mode* (or otherwise mis-configured) enclave can never become a trusted proof signer:

1. **Extends the universal forbidden-attribute floor** in `SgxVerifier` from `DEBUG | PROVISION_KEY` (`0x12`) to also reject **`EINITTOKEN_KEY`** (`0x20`) → mask `0x32`, on every network.
2. **Adds a fail-closed, per-MRENCLAVE ATTRIBUTES pin** to `SecureSgxVerifier`: each allowlisted enclave measurement must declare the exact `(mask, expected)` profile it may register with; an MRENCLAVE with no configured pin **cannot register**. The owner setter rejects any policy that could re-admit a forbidden bit.
3. **Deploy-script + doc note:** mainnet **and all public testnets** use `SecureSgxVerifier`; `InsecureSgxVerifier` (lenient TCB policy) is **local-devnet only**.

Scope is deliberately limited to the verifier's attribute checks (see *Out of scope*).

## Why — from first principles (no SGX background assumed)

- **The sealed room.** Intel SGX lets a program run inside a CPU-protected "enclave": its memory is hardware-encrypted and isolated, so privileged software (OS, hypervisor, root) and even a physical attacker can't *directly* read or modify it — *modulo* side-channel / microarchitectural attacks (Foreshadow, MDS, SGAxe, ÆPIC), which SGX explicitly does **not** defend against. The room can sign its output with a key that never leaves.
- **Proving you can't see inside it — attestation.** The room emits a *quote*: signed by the platform's attestation key, which is certified by a **PCK certificate chain rooting in an Intel CA**, attesting "this is a genuine SGX enclave running exactly this code at this TCB (patch) level." Two fingerprints identify the code: **MRENCLAVE** (hash of the enclave's code/data) and **MRSIGNER** (hash of the developer's signing key).
- **How Taiko uses it.** The prover runs *inside* the room, re-executes L2 blocks, and only signs a state transition that is genuinely valid. On L1, `SgxVerifier` trusts a signature from a registered key — accepting it **is** trusting the proof, since the contract can't re-run the L2. Anyone running a genuine prover registers their room's key permissionlessly via `registerInstance`.
- **The flaw.** A quote also carries an **ATTRIBUTES** field — mode switches set when the room is launched. One is **DEBUG**: a debug room is built for developers and **lets the host read and modify its memory** — including the signing key. Crucially, **a debug build and a production build of the same binary have the same MRENCLAVE** (DEBUG lives in ATTRIBUTES, which is *not* part of the measurement). So an attacker takes Taiko's real prover binary (correct MRENCLAVE/MRSIGNER → passes the allowlist), launches it in debug, reads out the signing key, and signs a *fraudulent* transition. The old code checked "genuine room, approved code?" (yes) but never "is the debug hatch open?" (it was). *(ATTRIBUTES is bound into the signed report and is also constrained at launch by the author's SIGSTRUCT `ATTRIBUTEMASK`, enforced at `EINIT` — which is why the attacker needed Taiko's real MRSIGNER, and why a complementary fix is to forbid DEBUG in the production SIGSTRUCT.)*
- **The other forbidden bits.** `PROVISION_KEY` (derive platform-identifying keys) and `EINITTOKEN_KEY` (mint launch tokens) are privileges a prover enclave must never hold. #21822 forbade DEBUG + PROVISION; **this PR adds the missing `EINITTOKEN_KEY`**.
- **Why pin per-MRENCLAVE.** A deny-list only catches bits you named. The robust check declares, per trusted binary, the *exact* allowed mode and rejects everything else — reserved/unknown bits, and a *missing* required bit (production must have `INIT` + 64-bit mode). It fails closed, so permissionless registration can't quietly admit a mode you didn't vet. The FLAGS bits (DEBUG/PROVISION/LAUNCH) occupy the first byte of the 16-byte ATTRIBUTES field, which is why the floor mask is `0x32` in that byte.

## Changes

| File | Change |
| --- | --- |
| `SgxVerifier.sol` | floor `0x12`→`0x32` (adds `EINITTOKEN_KEY`); new `_validateEnclaveAttributes` hook run after the floor (base no-op; production subclasses MUST override) |
| `SecureSgxVerifier.sol` | fail-closed per-MRENCLAVE `(mask, expected)` ATTRIBUTES pin via the hook; owner `setEnclaveAttributePolicy` / `removeEnclaveAttributePolicy`; setter rejects any policy that could re-admit a forbidden bit |
| `InsecureSgxVerifier.sol` | documented as **local-devnet only** |
| `DeployProtocolOnL1.s.sol`, `DeployShastaContracts.s.sol` | note: mainnet + all public testnets use `SecureSgxVerifier`; `InsecureSgxVerifier` is devnet only |
| `SgxVerifier.t.sol` | launch-key floor test (both subclasses) + 9 per-MRENCLAVE pin tests |

## Out of scope (intentionally; separate changes)

This PR is **only** the verifier's attribute checks. It does **not** add:
- a registration→activation delay, shorter `INSTANCE_EXPIRY`, or key rotation;
- a heterogeneous SGX-**and**-ZK proof composition;
- a bridge withdrawal delay / veto.

Those remain necessary defense-in-depth against the residual risk that a side-channel extracts a key from a fully-patched, non-debug enclave — but are tracked separately.

## Testing

- `forge test --match-path test/layer1/verifiers/SgxVerifier.t.sol` → **51/51 pass** (`SecureSgxVerifierTest` 30, `InsecureSgxVerifierTest` 21), incl. fail-closed, strict-pin accept/reject, setter validation, owner-gating, events, and removal.
- `forge fmt --check` clean; deploy scripts compile.

> [!NOTE]
> The live mainnet SGX verifiers are non-upgradeable, so this ships as a **new** deployment that the Inbox's composed verifier must be re-wired to. Because the pin fails closed, `setEnclaveAttributePolicy` must be configured for each production MRENCLAVE before go-live, or proving halts. `gas-reports/layer1-contracts.txt` should be regenerated with `pnpm snapshot:l1` before merge.


dantaik:
@smtmfft with this PR, do you know what parameters to use in order to call `setEnclaveAttributePolicy`

dantaik:
closed in favor of https://github.com/taikoxyz/taiko-mono/pull/21827

PR #21849, opened 2026-06-24
fix(taiko-client): handle empty blob case

YoGhurt111:


packages/taiko-client/pkg/rpc/blob_datasource.go
```diff
@@ -108,7 +108,7 @@ func (ds *BlobDataSource) GetBlobBytes(
 		b = append(b, bytes...)
 	}
 	if len(b) == 0 {
-		return nil, pkg.ErrSidecarNotFound
+		return nil, errors.Join(ErrInvalidBlobBytes, pkg.ErrSidecarNotFound)
```
davidtaikocha: Shall we just return `ErrInvalidBlobBytes`? since sidecar actually can be found? it just contains 0 bytes?

YoGhurt111: Sure

YoGhurt111:
```suggestion
		return nil, ErrInvalidBlobBytes
```

YoGhurt111: Fixed in [5182d9f](https://github.com/taikoxyz/taiko-mono/pull/21849/commits/5182d9f01c23c46425922cbdffaadf59bd760124)

davidtaikocha (approved):

PR #21852, opened 2026-06-24
fix(taiko-client): bound preconfirmation tx list decompression

failsafesecurity:
### Summary

This bounds zlib decompression for preconfirmation transaction lists before RLP decoding. The existing validation capped the compressed input size, but `utils.Decompress` read the decompressed stream without an output limit.

### Security impact

A compressed transaction list can expand beyond the expected preconfirmation tx-list budget during validation. Bounding the decompressed output avoids unnecessary memory growth before the payload is rejected.

### Changes

- Add `utils.DecompressWithLimit` while preserving existing `Decompress` behavior for other callers.
- Use a preconfirmation tx-list decompression ceiling in payload validation and cached-parent handling.
- Add a unit test covering oversized decompressed output rejection.

### Validation

- `git diff --check origin/main..HEAD` passed.
- Targeted Go tests were not run because `go`/`gofmt` are unavailable in this OpenClaw host environment.

### DIFF
--- a/packages/taiko-client/driver/preconf_blocks/server.go
+++ b/packages/taiko-client/driver/preconf_blocks/server.go
@@ -46,6 +46,8 @@ var (
 )
 
 const requestSyncMargin = uint64(128) // Margin for requesting sync, to avoid requesting very old blocks.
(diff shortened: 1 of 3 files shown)


davidtaikocha:
Thanks and since preconf server is protected by JWT secret, so the requests can be trusted.

PR #21853, opened 2026-06-24
fix(protocol): restrict preconfirmation operator additions

failsafesecurity:
### Summary

This restricts `PreconfWhitelist.addOperator` to the owner. Ejecters can still remove operators, but can no longer add new proposer/sequencer pairs.

### Security impact

Ejecters are documented and configured as an operator-removal role. Allowing the same role to add operators expands that authority and can make newly added proposers eligible after the activation delay.

### Changes

- Change `addOperator` from `onlyOwnerOrEjecter` to `onlyOwner`.
- Add a regression test confirming ejecters cannot add operators.

### Validation

- `git diff --check origin/main..HEAD` passed.
- Targeted Foundry tests were not run because `forge` is unavailable in this OpenClaw host environment.


ggonzalez94:
Thanks. But it is an intended design to allow this role to add operators

PR #21861, opened 2026-06-25
ci(taiko-client): keyless GAR auth via Workload Identity

kimo-ice:
## What

Migrates the taiko-client multi-arch image release workflow (`.github/workflows/taiko-client--docker.yml`) from a static GAR service-account key (`secrets.GAR_JSON_KEY`) to keyless **Workload Identity Federation**.

Both jobs (`build` matrix + `merge`) now:
- request `id-token: write`,
- authenticate via `google-github-actions/auth@v2` impersonating `gar-github-action@evmchain.iam.gserviceaccount.com`,
- `docker login` to `us-docker.pkg.dev` with `oauth2accesstoken` + the short-lived access token instead of the long-lived JSON key.

## Why

Part of the org GCP hardening effort: retiring long-lived downloadable SA keys in favor of short-lived, keyless OIDC tokens. `gar-github-action` is responsible for image builds across most repos, so its static key is high blast-radius.

## Verification

Proven before touching this workflow with a temporary smoke test (`wif-gar-smoke.yml`) that authenticated via the same WIF provider/SA and pushed a probe image to `us-docker.pkg.dev/evmchain/images`, on **both** `ubuntu-latest` and the self-hosted `arc-runner-set` / arm64 runners this workflow actually uses. All steps green (auth, login, GAR write). The smoke branch and probe images have been cleaned up.

The WIF binding `taikoxyz/taiko-mono -> gar-github-action@evmchain` (principalSet on `attribute.repository`) is already in place.

## Rollout / safety

- This workflow runs only on `push` to `main` and `taiko-alethia-client-v*` tags, so the cutover happens on merge.
- The `GAR_JSON_KEY` org secret is intentionally **left in place** as a fallback until the keyless pattern is rolled out to the remaining image workflows; it will be retired (and the `gar-github-action` keys deleted) afterwards.
- Revert = restore the previous two `Login to GAR` steps; the key is still valid.

This is the pilot; follow-up PRs will migrate the other taiko-mono docker workflows and sibling repos.

### DIFF
--- a/.github/workflows/taiko-client--docker.yml
+++ b/.github/workflows/taiko-client--docker.yml
@@ -21,6 +21,8 @@ permissions:
 
 env:
   REGISTRY_IMAGE: us-docker.pkg.dev/evmchain/images/taiko-client
+  WIF_PROVIDER: projects/790288402401/locations/global/workloadIdentityPools/terraform-pool/providers/github-provider
+  WIF_SERVICE_ACCOUNT: gar-github-action@evmchain.iam.gserviceaccount.com
 
 jobs:
   build:
@@ -44,20 +46,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
-      - name: Authenticate to Google Cloud (Workload Identity)
-        id: auth
-        uses: google-github-actions/auth@v2
-        with:
-          workload_identity_provider: projects/790288402401/locations/global/workloadIdentityPools/terraform-pool/providers/github-provider
-          service_account: gar-github-action@evmchain.iam.gserviceaccount.com
-          token_format: access_token
-      - name: Login to GAR
-        uses: docker/login-action@v4
-        with:
-          registry: us-docker.pkg.dev
-          username: oauth2accesstoken
-          password: ${{ steps.auth.outputs.access_token }}
-
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v4
 
@@ -77,6 +65,23 @@ jobs:
             type=ref,event=pr
             type=ref,event=tag
             type=sha
+
+      # Auth + login run immediately before the build/push step: the WIF access
+      # token is short lived (~1h), so we mint it as late as possible to leave
+      # the most headroom for long multi-arch builds before the push.
+      - name: Authenticate to Google Cloud (Workload Identity)
+        id: auth
+        uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: ${{ env.WIF_PROVIDER }}
+          service_account: ${{ env.WIF_SERVICE_ACCOUNT }}
+          token_format: access_token
+      - name: Login to GAR
+        uses: docker/login-action@v4
+        with:
+          registry: us-docker.pkg.dev
+          username: oauth2accesstoken
+          password: ${{ steps.auth.outputs.access_token }}
 
       - name: Build and push by digest
         id: build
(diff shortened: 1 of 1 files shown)


.github/workflows/taiko-client--docker.yml
```diff
@@ -39,12 +44,19 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
+      - name: Authenticate to Google Cloud (Workload Identity)
+        id: auth
+        uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: projects/790288402401/locations/global/workloadIdentityPools/terraform-pool/providers/github-provider
+          service_account: gar-github-action@evmchain.iam.gserviceaccount.com
+          token_format: access_token
```
```suggestion
          token_format: access_token
          access_token_lifetime: 3600s
```

.github/workflows/taiko-client--docker.yml
```diff
@@ -121,12 +133,19 @@ jobs:
             type=ref,event=tag
             type=sha
 
+      - name: Authenticate to Google Cloud (Workload Identity)
+        id: auth
+        uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: projects/790288402401/locations/global/workloadIdentityPools/terraform-pool/providers/github-provider
+          service_account: gar-github-action@evmchain.iam.gserviceaccount.com
+          token_format: access_token
```
RogerLamTd, line 152:
```suggestion
          token_format: access_token
          access_token_lifetime: 3600s
```

RogerLamTd (approved):

dantaik:
@review

.github/workflows/taiko-client--docker.yml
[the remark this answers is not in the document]
kimo-ice: Good catch, fixed in `32fc0501f`. Set `create_credentials_file: false` on both auth steps since the only thing we consume is `steps.auth.outputs.access_token` for `docker/login-action`. No `gha-creds-*.json` ADC file is written now, so nothing sensitive lands in the Docker build context.

kimo-ice: Applied in `32fc0501f` (added `access_token_lifetime: 3600s`). Also moved this build-job auth + GAR login to run immediately before the build/push step, so the token is minted as late as possible and gets the full hour of headroom before the push on long multi-arch builds.

.github/workflows/taiko-client--docker.yml
kimo-ice: Applied in `32fc0501f` (added `access_token_lifetime: 3600s`). The merge job already logs in right before its `imagetools create` push.

kimo-ice:
Thanks for the reviews. Changes since the first commit:

- **`create_credentials_file: false`** on both auth steps (Codex P2) — we only use the access-token output for `docker login`, so no `gha-creds-*.json` ADC file is written into the workspace / Docker build context.
- **`access_token_lifetime: 3600s`** on both auth steps (@RogerLamTd).
- **Token minted late:** the build job's auth + GAR login now run immediately before the build/push step, leaving the full hour of headroom before the push on long multi-arch builds (GLM warning re: token expiry).
- **DRY:** WIF provider + service account lifted to workflow-level `env` (`WIF_PROVIDER` / `WIF_SERVICE_ACCOUNT`) so the follow-up workflow migrations don't copy-paste them (GLM suggestion).
- **SHA-pinning the auth action** (DeepSeek): skipped here to stay consistent with the repo convention (every action uses a mutable tag, e.g. `actions/checkout@v6`). Better as a separate repo-wide pinning pass than a one-off on this file.

On the two red checks: **check-links** is failing on an external flake in the root `README.md` (`public-api.gitpoap.io/.../badge` → "Connection failed"), unrelated to this change; **merge-gatekeeper** is red only because it gates on check-links. Both should clear on a healthy re-run.

`GAR_JSON_KEY` stays in place as a fallback until the remaining image workflows migrate, then the keys get deleted and the extra `editor` role on `gar-github-action` dropped.

davidtaikocha (approved):

PR #21865, opened 2026-06-26
fix(protocol): enforce trusted MRENCLAVE/MRSIGNER in SgxVerifier.registerInstance

ggonzalez94:
## Why

Audit finding **F1 (Critical-in-code)**: application-enclave **code identity (MRENCLAVE/MRSIGNER) was only enforced when `AutomataDcapV3Attestation.checkLocalEnclaveReport` is enabled** — a storage bool that **defaults to `false`** and can be toggled off by the attestation owner. Everything else in `verifyParsedQuote` (QE identity, cert chain to the Intel root, TCB level, signatures) only proves the quote came from *genuine, non-debug SGX hardware* — none of it constrains *which code* runs in the application enclave.

So a verifier relying on that flag could accept a genuinely-attested enclave running **arbitrary attacker code**, register its key, and let it sign forged proofs — the same class of failure as the incident. On fresh deploys this is reachable by anyone (`registrar == address(0)` in the generic templates); on mainnet it's one mistaken toggle away.

## What

Pin MRENCLAVE and MRSIGNER **directly in `SgxVerifier.registerInstance`**, unconditionally, so code-identity enforcement no longer depends on any attestation-side toggle. This follows the existing pattern in this contract (the debug/provision forbidden-attributes check and the per-network TCB policy already live in the verifier).

```solidity
require(
    IAttestation(automataDcapAttestation).trustedUserMrEnclave(_attestation.localEnclaveReport.mrEnclave),
    SGX_UNTRUSTED_MRENCLAVE()
);
require(
    IAttestation(automataDcapAttestation).trustedUserMrSigner(_attestation.localEnclaveReport.mrSigner),
    SGX_UNTRUSTED_MRSIGNER()
);
```

- `SgxVerifier.sol`: two new errors + two `require`s after the forbidden-attributes check. Applies to **both** `SecureSgxVerifier` and `InsecureSgxVerifier`. The owner `addInstances()` break-glass path is intentionally unaffected.
- `IAttestation.sol`: expose `trustedUserMrEnclave` / `trustedUserMrSigner` view getters — already implemented by `AutomataDcapV3Attestation`'s `public` mappings.
- Tests: `MockAttestation` implements the getters (trust-all default + per-measurement overrides); 3 new tests (untrusted MRENCLAVE → revert, untrusted MRSIGNER → revert, both trusted → success).

## Why reading the parsed fields is sound

`mrEnclave`/`mrSigner` are read from the same calldata struct passed to `verifyParsedQuote`, and they are covered by the attestation signature it verifies: `signedQuoteData = header ++ packQEReport(localEnclaveReport)` (which includes `mrEnclave`+`mrSigner`) is signed by the attestation key → bound to the QE report → PCK cert chain → Intel root. A forged value changes `signedQuoteData` and fails verification. This is the exact same basis the existing forbidden-attributes check already relies on.

## Test plan

`FOUNDRY_PROFILE=layer1 forge test --match-path "test/layer1/verifiers/*"` → **71 passing**; `test/layer1/automata-attestation/*` → **13 passing**; `forge fmt --check` clean.

## ⚠️ Necessary but not sufficient — ship with allowlist curation

This makes pinning **mandatory**, but pinning only helps if the allowlist is **curated**. On the live mainnet attester proxies the **leaked MRSIGNER is still trusted** (separate finding F4). This PR should land together with the governance cleanup that (a) `setMrSigner(leakedSigner, false)`, and (b) registers the recovery enclave's MRENCLAVE/MRSIGNER — otherwise `registerInstance` will revert for the new prover, or still admit the known-bad-but-allowlisted signer.


davidtaikocha (approved):

ggonzalez94:
Based on multiple rounds of AI auditing. @smtmfft please help me review if:
1. this truly makes sense
2. I understand this is already set in storage in prod. But this is additional hardening for new deployments

davidtaikocha:
Closed due to the new SGX verifier we will use.

PR #21883, opened 2026-07-01
chore(repo): rollup-monitor

RogerLamTd:


packages/rollup-monitor/src/bindings.rs
```diff
@@ -0,0 +1,67 @@
...
+        uint256 amount
+    );
+
+    event Upgraded(address indexed implementation);
+    event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
+    event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender);
+    event RoleRevoked(bytes32 indexed role, address indexed account, address indexed sender);
+    event Paused(address account);
+    event Unpaused(address account);
+    event ExecutionSuccess(bytes32 txHash, uint256 payment);
```
davidtaikocha: **Critical — silent miss: the entire Safe-activity check is dead.**

The real Gnosis Safe events are `event ExecutionSuccess(bytes32 indexed txHash, uint256 payment)` / `...Failure(bytes32 indexed txHash, ...)` — `txHash` is **indexed**. `indexed` doesn't change `topic0`, so the signature still matches, but a real Safe log carries `txHash` in a topic (2 topics) with only 32 bytes of data, while this binding expects 0 indexed params and 64 bytes. `log_decode` therefore returns `Err`, and events.rs swallows it (`.is_ok()`), so **every guardian-Safe execution produces no `safe_transactions_total` increment and no alert** — the exact high-privilege actor this monitor exists to watch.

The unit test round-trips because it encodes the log using this same wrong definition.

```solidity
event ExecutionSuccess(bytes32 indexed txHash, uint256 payment);
event ExecutionFailure(bytes32 indexed txHash, uint256 payment);
```
Add a regression test that decodes a hand-built log with 2 topics (sig + txHash) and 32-byte data.

packages/rollup-monitor/src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+    new_int_counter_vec(
+        "rollup_monitor_sgx_anomalies_total",
+        "SGX or TEE registration anomalies observed",
+        &["chain", "instance", "reason"],
+    )
+});
+
+static RECONCILIATION_MISMATCHES_TOTAL: Lazy<IntCounterVec> = Lazy::new(|| {
+    new_int_counter_vec(
+        "rollup_monitor_reconciliation_mismatches_total",
```
davidtaikocha: **Critical — dead alert.** `reconciliation_mismatches_total` is registered here and listed in the README, but there is no `inc_reconciliation_mismatch` and no `record_alert` arm anywhere in the crate — so it can never increment. An operator who builds a Grafana alert on it gets a rule that is guaranteed never to fire.

Either implement the reconciliation check + its mutator, or remove the metric from registration and the README until the check exists (and mark it clearly as unimplemented).

packages/rollup-monitor/bin/rollup-monitor/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    #[cfg(unix)]
+    let term = async {
+        if let Ok(mut signal) = unix_signal(SignalKind::terminate()) {
+            signal.recv().await;
+        }
+    };
+    #[cfg(not(unix))]
+    let term = std::future::pending::<()>();
+
+    tokio::select! {
```
davidtaikocha, line 78: **Critical — silent monitor death.** This `select!` awaits only `ctrl_c`/SIGTERM, never `monitor_handle`. If `monitor.run()` returns `Err` (e.g. a bad `L1_HTTP_URL` fails `Url::parse` before the loop) or the task panics, the error is merely logged and the process keeps serving `/metrics`/`/healthz` with a frozen scan loop — k8s never restarts it.

Make the monitor task's completion a fatal branch:
```rust
res = &mut monitor_handle => {
    tracing::error!("monitor task exited: {res:?}");
    std::process::exit(1);
}
```
Pairs with the `/healthz` issue in server.rs.

packages/rollup-monitor/src/server.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    Ok(())
+}
+
+pub fn router() -> Router {
+    Router::new().route("/healthz", get(health)).merge(metrics::router())
+}
+
+async fn health() -> (StatusCode, &'static str) {
+    (StatusCode::OK, "rollup-monitor is running")
```
davidtaikocha: **Critical (with main.rs) — always-healthy while dead.** `/healthz` returns `200` unconditionally, with no reference to scan-loop liveness. Combined with the monitor-task-death issue in main.rs, a stalled or crashed scan loop leaves the pod `Ready` and serving a frozen `/metrics` forever.

Gate health on scan freshness — record a `last_successful_scan` timestamp and return `503` when `now - last_scan > k * poll_interval` — so a dead loop actually fails the liveness/readiness probe.

packages/rollup-monitor/src/monitor.rs
```diff
@@ -0,0 +1,433 @@
...
+            let Some(target) = self.contract_name(log.address()) else {
+                continue;
+            };
+            self.process_log("l1", chain_id, target.as_str(), &log);
+        }
+
+        Ok(())
+    }
+
+    async fn scan_eoa_transactions<P>(
```
davidtaikocha, line 158: **Important — no dedupe on EOA txs → overlap re-counts every cycle.** `scan_eoa_transactions` → `process_eoa_transaction` records alerts with no `SeenLogCache` check, while contract logs *are* deduped in `process_log`. With `OVERLAP_BLOCKS=20` rescanned each ~12s cycle, a single watched-EOA→unapproved-destination tx inflates `unexpected_eoa_transactions_total` ~20× — alert fatigue / metric corruption. A persistently-failing later chunk amplifies it (earlier chunks re-scanned on every retry).

Route EOA hits through the same dedupe cache keyed on `tx_hash`.

packages/rollup-monitor/src/monitor.rs
```diff
@@ -0,0 +1,433 @@
...
+    pub fn classify_and_record(&self, chain: &str, event: &ObservedEvent) -> Option<Alert> {
+        let alert = classify_event(&self.config, event)?;
+        record_alert(chain, &alert);
+        Some(alert)
+    }
+
+    pub fn process_log(&mut self, chain: &str, chain_id: u64, target: &str, log: &RpcLog) -> usize {
+        if let (Some(tx_hash), Some(log_index)) = (log.transaction_hash, log.log_index) {
+            let key = LogKey { chain_id, tx_hash, log_index };
+            if !self.seen_logs.insert(key) {
```
davidtaikocha, line 219: **Important — dedupe suppresses proposal-reorg detection.** The dedupe gate runs *before* `proposal_observation_from_log`, and `LogKey` is `(chain_id, tx_hash, log_index)` with **no `block_hash`**. `ProposalTracker::observe` explicitly compares `block_hash` to catch a reorg where the same proposal is re-mined at the same tx position in a different-hash block — but this gate rejects that re-observation first, so `proposal_reorgs_total` misses exactly that case. `process_log_detects_changed_proposal_observation` only exercises the different-`tx_hash` path, masking the gap.

Fix: include `block_hash` in `LogKey`, or observe proposals before the dedupe gate.

packages/rollup-monitor/src/monitor.rs
```diff
@@ -0,0 +1,433 @@
...
+        info!(from_block, safe_head, "scanning L1 rollup monitor range");
+        for (chunk_from, chunk_to) in
+            chunk_ranges(from_block, safe_head, self.config.max_block_range)
+        {
+            self.scan_contract_logs(provider, chain_id, chunk_from, chunk_to).await?;
+            self.scan_eoa_transactions(provider, chunk_from, chunk_to).await?;
+        }
+
+        self.l1_cursor.mark_scanned(safe_head);
+        metrics::set_scan_position("l1", safe_head, safe_head);
```
davidtaikocha: **Important — `scan_lag_blocks` is always 0.** Both `set_scan_position("l1", safe_head, safe_head)` calls (here and line 84) pass `safe_head` for *both* the `last_scanned` and `safe_head` args; metrics.rs computes `lag = safe_head.saturating_sub(last_scanned) = 0`. The one "am I keeping up?" signal is dead. Pass the true observed head (`latest_block`) as the second arg vs. `safe_head` as `last_scanned`.

packages/rollup-monitor/src/monitor.rs
```diff
@@ -0,0 +1,433 @@
...
+    {
+        if self.config.watched_eoas.is_empty() {
+            return Ok(());
+        }
+
+        for block_number in from_block..=to_block {
+            let Some(block) =
+                provider.get_block_by_number(BlockNumberOrTag::Number(block_number)).full().await?
+            else {
+                metrics::inc_scan_error("l1", "block_missing");
```
davidtaikocha, line 176: **Important — missing block silently skipped, then cursor advances past it.** On `get_block_by_number(...) == Ok(None)` this bumps `block_missing` and `continue`s, but the cycle still `mark_scanned`s past the block. A lagging node returning null for a block ≤ safe_head means that block's watched-EOA txs are missed forever.

Treat `Ok(None)` at/below safe_head as a chunk error (propagate `Err`) so the cursor doesn't advance and the block is retried.

packages/rollup-monitor/src/config.rs
```diff
@@ -0,0 +1,348 @@
...
+    value.split(',').map(str::trim).filter(|item| !item.is_empty())
+}
+
+fn normalize_name(name: &str) -> Result<String, String> {
+    let name = name.trim();
+    if name.is_empty() {
+        return Err("name cannot be empty".to_string());
+    }
+
+    Ok(name.to_string())
```
davidtaikocha: **Important — case-sensitive names cause a silent threshold drop.** `normalize_name` only trims; it does not lowercase. `classify_event` does `withdrawal_thresholds_wei.get(target)?` keyed on the watched-contract name, so `WATCHED_CONTRACTS=Inbox=…` + `WITHDRAWAL_THRESHOLDS_WEI=inbox=…` misses → the large-withdrawal alert is dropped entirely. Same hazard for `EXPECTED_OWNERS`/`EXPECTED_PROXY_IMPLEMENTATIONS`.

Canonicalize to lowercase here, or validate at startup that every expectation key exists in `watched_contracts`.

packages/rollup-monitor/src/config.rs
```diff
@@ -0,0 +1,348 @@
...
+    let mut parsed = HashMap::new();
+
+    for item in split_csv(value) {
+        let (name, address) = item
+            .split_once('=')
+            .ok_or_else(|| format!("expected name=address pair, got '{item}'"))?;
+        let name = normalize_name(name)?;
+        let address = Address::from_str(address.trim())
+            .map_err(|error| format!("invalid address for '{name}': {error}"))?;
+        parsed.insert(name, address);
```
davidtaikocha: **Important — duplicate keys silently last-win.** `parse_named_addresses` (and `parse_named_u128s` at line 175) use `insert`, so `inbox=0xAAA,inbox=0xBBB` silently discards the first → a watched contract / expected owner vanishes. Contrast `parse_named_b256_sets`, which correctly accumulates. Return `Err("duplicate name '{name}'")` for the single-value maps.

packages/rollup-monitor/src/events.rs
```diff
@@ -0,0 +1,835 @@
...
+
+pub fn classify_eoa_transaction(
+    watched_eoas: &HashSet<Address>,
+    signer: Address,
+    to: Option<Address>,
+) -> Option<ObservedEvent> {
+    watched_eoas.contains(&signer).then_some(ObservedEvent::UnexpectedEoaTransaction { signer, to })
+}
+
+pub fn decode_contract_log(target: &str, log: &RpcLog) -> Vec<ObservedEvent> {
```
davidtaikocha, line 231: **Important — decode failures are swallowed with no metric/log.** Every branch is `if let Ok(...) = log.log_decode()`. A log whose `topic0` matches a known event but fails to decode (the Safe-`indexed` bug, or future ABI drift) increments no `scan_error` and logs nothing — the silent miss is invisible. Since the scanner fetches all logs from watched addresses, a "no binding matched" is normal; but a **topic0-matched** log that fails to decode is a strong silent-miss signal. Match on `topic0` first, then `warn!` + `inc_scan_error` on decode failure for matched signatures.

packages/rollup-monitor/src/events.rs
```diff
@@ -0,0 +1,835 @@
...
+        let _ = decoded;
+        events.push(ObservedEvent::Pause { target: target.to_string(), paused: false });
+    }
+    if let Ok(decoded) = log.log_decode::<bindings::Upgraded>() {
+        events.push(ObservedEvent::ProxyUpgrade {
+            target: target.to_string(),
+            proxy: log.address(),
+            implementation: decoded.inner.data.implementation,
+        });
+        if target.contains("verifier") {
```
davidtaikocha: **Important — verifier-change keyed on a substring of a free-form name.** A `VerifierChange` is only emitted for an `Upgraded` log when the config target name literally contains `"verifier"`. Naming the contract `risc0`, `sgx_v`, or a typo silently drops the verifier-change signal (the plain `ProxyUpgrade` still fires). Key off an explicit verifier-target set derived from config instead.

(Minor, related — line 270 packs the 20-byte impl address into the same `verifier: B256` field used for real 32-byte `imageId`/`programVKey` values, so `expected_verifiers` entries for a proxy upgrade must be left-padded addresses: a non-obvious operator footgun.)

packages/rollup-monitor/src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+use prometheus::{Encoder, IntCounterVec, IntGaugeVec, Opts, Registry, core::Collector};
+
+static REGISTRY: Lazy<Registry> = Lazy::new(Registry::new);
+static METRICS_INIT: std::sync::Once = std::sync::Once::new();
+
+static PROXY_UPGRADES_TOTAL: Lazy<IntCounterVec> = Lazy::new(|| {
+    new_int_counter_vec(
+        "rollup_monitor_proxy_upgrades_total",
+        "Unexpected and expected proxy upgrade events observed",
+        &["chain", "target", "proxy", "implementation", "expected"],
```
davidtaikocha: **Important — unbounded label cardinality.** These counters carry raw addresses / proposal IDs as label values (`proxy`, `implementation` here; also `{prover}`, `{proposer}`, `{recipient}`, `{to}`, `{signer}`, `{instance}`, and monotonic `{proposal_id}` on `proposal_reorgs_total`). An attacker who can emit matching events — or just a busy chain — blows up Prometheus time-series cardinality, a DoS on the monitoring pipeline that can bury real alerts. Keep only bounded labels (`chain`, `target`, `expected`) on the counters; move high-cardinality identifiers to the log body.

.github/workflows/rollup-monitor--docker.yml
```diff
@@ -0,0 +1,143 @@
+name: "Build and Push Multi-Arch Docker Image (rollup-monitor)"
+permissions:
+  contents: read
+
+on:
+  push:
+    branches: [new-alerting-infra]
```
davidtaikocha: **Important — image only builds on a feature branch.** The workflow triggers only on `push` to `new-alerting-infra`. There's no `main` or tag trigger, yet the metadata-action is configured for `type=match,pattern=rollup-monitor-v(\d+\.\d+\.\d+)`. Once this merges, no image is ever built for `main` or a release tag. Add `main` + tag-push triggers.

packages/rollup-monitor/src/config.rs
```diff
@@ -0,0 +1,348 @@
...
+#[command(name = "rollup-monitor", version)]
+pub struct Config {
+    #[arg(long, env = "L1_HTTP_URL")]
+    pub l1_http_url: String,
+
+    #[arg(long, env = "L2_HTTP_URL")]
+    pub l2_http_url: Option<String>,
+
+    #[arg(long, env = "HTTP_PORT", default_value_t = DEFAULT_HTTP_PORT)]
+    pub http_port: u64,
```
davidtaikocha: **Important — `http_port` should be `u16`.** As `u64` it's only range-checked at bind time (`u16::try_from` in server.rs), and that error is swallowed into the spawned server task (see main.rs). A port > 65535 leaves a process with no HTTP server but a healthy-looking appearance to anything not probing the port. Make it `u16` so clap rejects it at startup.

davidtaikocha (commented):
## Code review — `rollup-monitor`

This is an **alert-only** monitor, so its dominant failure mode is a *silent miss*: a bug that makes a real security event never alert. The architecture is sound and the hard parts are done well, but several issues defeat that mission silently. Full findings are inline; summary below.

### Strengths
- Event signatures for the high-value events (Inbox `Proposed`/`Proved`, ERC20Vault `TokenSent`/`TokenReleased`, SGX `InstanceAdded`/`InstanceDeleted`, OZ admin events) were verified against the actual protocol contracts and are correct.
- Scan cursor never advances on error (`mark_scanned` runs only after the chunk loop) → failed ranges are retried, not skipped.
- No `unwrap`/`expect`/panicking index in the scan/decode path; `checked_sub`/`saturating_sub` throughout; the clippy deny-list is real and CI-enforced.
- Byte-based allowlist matching (no checksum-mismatch storm); empty-allowlist "alert on everything" fails safe.
- Metric registration is `Once`-guarded/idempotent; graceful shutdown wired; `.env` loading is prod-safe; `Cargo.lock` committed.

### Critical (each is a silent miss)
1. **Safe `ExecutionSuccess`/`ExecutionFailure` bindings missing `indexed` on `txHash`** → every real guardian-Safe execution log fails to decode → the entire "Safe activity" check is dead. (bindings.rs)
2. **`reconciliation_mismatches_total` is registered + documented but has no incrementer** → an alert built on it can never fire. (metrics.rs)
3. **A dead monitor task is invisible**: `main.rs`'s `select!` never awaits the monitor handle, and `/healthz` is unconditionally `200` → a crashed/failed scan loop keeps a "healthy" pod serving a frozen `/metrics`, no restart. (main.rs + server.rs)

### Important
4. Watched-EOA txs have no dedupe → 20-block overlap re-counts every EOA tx ~every cycle. (monitor.rs)
5. Proposal-reorg detection is suppressed by the dedupe cache (keyed without `block_hash`, run before `ProposalTracker::observe`). (monitor.rs)
6. `scan_lag_blocks` is structurally always 0. (monitor.rs)
7. A transiently-missing block is silently skipped for EOA scanning and never rescanned. (monitor.rs)
8. Config names are case-sensitive across maps → a case typo silently drops a withdrawal threshold. (config.rs)
9. Duplicate `name=…` keys silently last-win, dropping a watched contract / expected owner. (config.rs)
10. Decode failures on topic0-matched logs are swallowed with no metric/log → silent misses are invisible. (events.rs)
11. Unbounded Prometheus label cardinality (proposal_id + raw addresses) → monitoring-pipeline DoS. (metrics.rs)
12. Verifier-change detection keys off the substring `"verifier"` in the operator-supplied target name. (events.rs)
13. Docker image workflow triggers only on branch `new-alerting-infra` — no `main`/tag trigger. (workflow)
14. `http_port` is `u64`, range-checked only at bind time → out-of-range port fails silently. (config.rs)

### Minor
- `chain_id` falls back to `0` on RPC failure and is baked into dedupe keys — prefer fail-fast.
- `role_changes_total`'s `expected` label is hardcoded `"false"` — wire it or drop it.
- `Upgraded→VerifierChange` packs a 20-byte address into the same `B256` field used for 32-byte `imageId`/`programVKey` values — allowlist footgun.
- `contract_name` is an O(n) reverse-scan-and-clone per matched log — build an `address→name` map once.
- `chain` label is always `"l1"` (L2 not wired) — add a comment.
- `test` CI job lacks the `dependabot` guard the `lint` job has; Dockerfile has no `EXPOSE`/`HEALTHCHECK`.
- Test-only `execute_l1_cycle` advances the cursor without scanning and isn't `#[cfg(test)]`-gated.

**Verdict: not ready to merge.** #1–#3 are hard blockers (each silently defeats the tool's purpose); #4 and #5 should also land before this monitors anything real — ideally each with a regression test, since the current tests for #1 and #5 pass *around* the bugs.

_Automated review dispatched by David via Claude Code._

packages/rollup-monitor/src/monitor.rs
```diff
@@ -0,0 +1,576 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn contract_name(&self, address: Address) -> Option<String> {
+        self.contract_names.get(&address).cloned()
+    }
+
+    pub fn eoa_scan_range(&self, from_block: u64, to_block: u64) -> Option<(u64, u64)> {
```
davidtaikocha, line 194: **Important (new) — EOA tail-limiting silently drops head blocks on catch-up.** This caps the EOA scan to the last `eoa_scan_max_block_range` (25) blocks of the chunk, but `mark_scanned(safe_head)` still advances the cursor past the *whole* range. When a cycle advances > 25 blocks — cold start, or catch-up after a stall/RPC hiccup — the earlier blocks are EOA-scanned nowhere and never revisited.

Concrete (defaults, `eoa_scan_enabled=true`): cursor 997, `latest=1100 → safe_head=1097`, chunk `(977,1097)` span 121 → returns `(1073,1097)`; **blocks 977–1072 are EOA-scanned nowhere**, then `mark_scanned(1097)`. A watched-EOA tx there is a silent miss. Steady state (span ≈ 21 ≤ 25) is unaffected.

Fix: scan the full range for EOA, or bound the per-cycle *cursor advance* (not just the tail) so skipped blocks retry next cycle, or at minimum `warn!` + a dedicated `eoa_blocks_skipped` metric when `span > max_blocks` so the gap is observable. (Only matters when `EOA_SCAN_ENABLED`.)

packages/rollup-monitor/src/monitor.rs
```diff
@@ -0,0 +1,576 @@
...
+        if self.config.watched_eoas.is_empty() {
+            return Ok(());
+        }
+
+        for block_number in from_block..=to_block {
+            let Some(block) =
+                provider.get_block_by_number(BlockNumberOrTag::Number(block_number)).full().await?
+            else {
+                metrics::inc_scan_error("l1", "block_missing");
+                eyre::bail!("block {block_number} missing from EOA scan range");
```
davidtaikocha, line 177: **Important (new) — failing the whole cycle on a missing block can stall the cursor + restart-loop.** Replacing the old silent `continue` was right, but `bail!` here fails the entire cycle: the `?` propagates before `mark_scanned`, so a lagging RPC returning `Ok(None)` for a block ≤ safe_head stalls the cursor every tick *and* skips `mark_successful_scan()` → `/healthz` goes 503 → k8s restarts a pod whose only problem is a slow/pruned RPC (restart doesn't help). Consider distinguishing "RPC hasn't served this confirmed block yet" (retry the range without 503-ing / without a hard cycle failure) from a genuine permanent gap, e.g. bound the EOA range to blocks known-present, or treat a transient `None` as a soft retry that still refreshes liveness.

packages/rollup-monitor/src/monitor.rs
```diff
@@ -0,0 +1,576 @@
...
+            chunk_ranges(from_block, safe_head, self.config.max_block_range)
+        {
+            self.scan_contract_logs(provider, chain_id, chunk_from, chunk_to).await?;
+            if let Some((eoa_from, eoa_to)) = self.eoa_scan_range(chunk_from, chunk_to) {
+                self.scan_eoa_transactions(provider, chain_id, eoa_from, eoa_to).await?;
+            }
+        }
+
+        self.l1_cursor.mark_scanned(safe_head);
+        metrics::set_scan_position("l1", safe_head, latest_block);
```
davidtaikocha, line 126: **Important (new) — `scan_lag_blocks` is now a constant, not real backlog.** This passes `(last_scanned=safe_head, head=latest_block)`, so `lag = latest_block - safe_head = confirmations` every cycle — no longer literally 0, but it still can't detect the scanner falling behind, and because it's only updated on the success path it goes *stale* (not rising) when the loop stalls. Report `latest_block.saturating_sub(cursor_last_scanned)` using the persisted cursor position, updated every tick (including skipped/errored cycles), so a growing backlog actually shows up.

packages/rollup-monitor/bin/rollup-monitor/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    load_env_for_dev();
+    metrics::init();
+
+    let config = Config::parse();
+    info!("Starting rollup-monitor");
+
+    let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel::<()>();
+    let http_port = config.http_port;
+    let health = HealthState::new(std::time::Duration::from_secs(
```
davidtaikocha, line 43: **Minor (new) — cold-start 503 risk.** `max_stale = poll_interval*3` (36s default) is seeded at startup, but `mark_successful_scan()` fires only at the *end* of a full cycle. A first cold-start cycle with `START_BLOCK_LOOKBACK=7200` scans several `get_logs` chunks (plus full-block EOA polling if enabled); if that exceeds 36s on a slow RPC, `/healthz` 503s before the first cycle completes and k8s may kill it mid-catch-up. Mark freshness per-chunk / per successful RPC round-trip, or size the window off worst-case cold-start duration.

packages/rollup-monitor/src/config.rs
```diff
@@ -0,0 +1,411 @@
...
+    )]
+    pub start_block_lookback: u64,
+
+    #[arg(long, env = "OVERLAP_BLOCKS", default_value_t = DEFAULT_OVERLAP_BLOCKS)]
+    pub overlap_blocks: u64,
+
+    #[arg(long, env = "MAX_BLOCK_RANGE", default_value_t = DEFAULT_MAX_BLOCK_RANGE)]
+    pub max_block_range: u64,
+
+    #[arg(long, env = "EOA_SCAN_ENABLED", default_value_t = false)]
```
davidtaikocha, line 56: **Minor (new) — default-off EOA scan is a silent coverage change on upgrade.** Defaulting `eoa_scan_enabled=false` is reasonable for RPC budget (and documented in the README), but an existing deployment that already had `WATCHED_EOAS` set will silently stop EOA alerting after this upgrade until someone adds `--eoa-scan-enabled`. Emit a startup `warn!` when `watched_eoas` is non-empty but `eoa_scan_enabled` is false, so the disabled state is visible rather than surprising.

davidtaikocha (commented):
## Re-review (after `deepseek review fixes` + `apply review comments`)

Strong turnaround — I re-reviewed the two new commits with the same three-way split (event decoding, scan-loop/config, metrics/server/infra), verifying each prior finding was *actually* fixed (not superficially) and hunting for regressions the fixes introduced.

### ✅ All prior findings are genuinely fixed
**Criticals:** Safe `ExecutionSuccess`/`ExecutionFailure` now have `indexed txHash` — and the new regression test builds a **real 2-topic log** rather than round-tripping the binding's own definition. `reconciliation_mismatches_total` is fully removed (registration + README). A dead monitor/server task is now a fatal `process::exit(1)` `select!` branch, and `/healthz` returns `503` off an `AtomicU64` last-successful-scan timestamp (with a real staleness test).

**Importants:** EOA txs now dedupe via a `{chain_id, tx_hash}` cache; `LogKey` now includes `block_hash` so a same-tx/same-log-index reorg is no longer swallowed (new test covers it); config names are lowercased in `normalize_name`; duplicate single-value keys now hard-error; `http_port` is `u16`; label cardinality is bounded (addresses/proposal_id moved to log bodies); docker workflow triggers on `main` + `rollup-monitor-v*` tags; decode failures on a topic0-matched log now `warn!` + `inc_scan_error`. Every fix has a test that would fail under the old bug. Event signatures were re-verified against the protocol contracts — no drift introduced by the `events.rs` refactor.

### New issues the fixes introduced (all lower-severity than the originals)
Notably, the three EOA-related ones are **gated behind `EOA_SCAN_ENABLED` (default `false`)**, so the shipped default config is safe — these bite only when EOA scanning is turned on.

1. **[Important] EOA tail-limiting silently drops head blocks on catch-up.** `eoa_scan_range` caps the EOA scan to the last `EOA_SCAN_MAX_BLOCK_RANGE` (25) blocks of a chunk, but the cursor still advances past the *whole* chunk. When a cycle advances > 25 blocks (post-stall catch-up), the earlier blocks are EOA-scanned nowhere and never revisited — a silent miss, with no warn/metric. Steady state (span ≈ 21 ≤ 25) is fine.
2. **[Important] `bail!` on a missing block → stuck cursor + possible restart loop.** The prior silent-skip was correctly replaced, but failing the whole cycle on `Ok(None)` for a block ≤ safe_head means a lagging RPC stalls the cursor every tick *and* stops marking scan-freshness → `/healthz` flips to 503 → k8s restarts a pod whose only problem is a slow RPC.
3. **[Important] `scan_lag_blocks` is now a constant (= `confirmations`), not real backlog.** `set_scan_position("l1", safe_head, latest_block)` makes lag = `latest - safe_head` = confirmations every cycle, and it's only updated on success → on a stall the gauge goes *stale* rather than rising. Track the persisted cursor (`latest - last_scanned`, updated every tick) to get a real "falling behind" signal.
4. **[Minor] Cold-start 503 risk.** Health is marked only at cycle *end* with a 36s (`poll_interval×3`) window seeded at startup; a first cold-start cycle (`START_BLOCK_LOOKBACK=7200`, multiple `get_logs` chunks) that exceeds 36s on a slow RPC can 503 before it finishes. Mark freshness per-chunk, or size the window off worst-case cold-start.
5. **[Minor] EOA default-off is a coverage change on upgrade.** An existing deployment with `WATCHED_EOAS` set silently stops EOA alerting until `--eoa-scan-enabled` is added. Documented in the README, but a startup `warn!` when `watched_eoas` is set + scan disabled would prevent a surprise.
6. **[Minor] Non-monotonic staleness clock** (`SystemTime` can step backward) and **Dockerfile now installs `curl`** for `HEALTHCHECK` (small attack-surface bump). Both acceptable.

### Verdict
The three Criticals and every Important from the first pass are correctly resolved — nice work. Remaining items are all lower-severity edge cases clustered around the new EOA-scan + liveness machinery, and the default config side-steps most of them. I'd fix #1–#3 (or at least make #1's skipped-block gap observable) before enabling EOA scanning in production, but none are merge blockers for the default-off configuration.

_Automated re-review dispatched by David via Claude Code._

PR #21885, opened 2026-07-01
feat(taiko-client): align prover with raiko2 v4 api

YoGhurt111:
## Summary

- Align taiko-client prover requests with the raiko2 v4 proposal and aggregation API from taikoxyz/raiko2#119.
- Switch Raiko proof responses to the v4 string proof shape and remove the old V2 response struct names.
- Route RISC0 backlog status/clear calls through the v4 control-plane API with explicit `proof_type`.
- Share the v4 proof request path between the zkVM compose producer and SGX geth producer.

## Validation

```bash
go test -count=1 ./prover/proof_producer ./prover/proof_submitter
```


YoGhurt111:
The proposal IDs in response is only used for log.

.github/workflows/taiko-client--docker.yml
```diff
@@ -2,7 +2,7 @@ name: "Build and Push Multi-Arch Docker Image"
 
 on:
   push:
-    branches: [main]
+    branches: [main,codex/raiko2-api-119-prover ]
```
```suggestion
    branches: [main]
```

PR #21890, opened 2026-07-02
refactor(relayer): remove watchdog, detect forged messages in the indexer

davidtaikocha:
## Summary

The bridge `watchdog` is not (and will never be) the bridge pauser — its `pause()` calls revert with `ACCESS_DENIED`, which caused forged `MessageProcessed` events to loop through its queue forever. Its only remaining value was the `BridgeMessageNotSent` detection signal, so this **removes the watchdog entirely** and relocates that detection into the component that already sees every processed event: the `MessageProcessed` indexer.

Follows up on #21888 (the interim alert-only fix), which removed the pausing but kept the watchdog running; this removes the service outright.

## Changes

- **Relocate detection** (`indexer/handle_message_processed_event.go`): the indexer runs `destBridge.IsMessageSent(message)` inline (new `isForgedMessage` helper), before `saveEventToDB`, on the forward path only, and increments `BridgeMessageNotSent` when the originating bridge has no record of sending the message. The publish to the watchdog-only queue is removed. The indexer already holds `destBridge` (same address the watchdog checked, both directions) — no config change needed.
- **Delete the watchdog**: the `watchdog/` package, its `cmd/main.go` subcommand + import, `cmd/flags/watchdog.go`, and the `WATCHDOG_PRIVATE_KEY` requirement.
- **Metrics**: remove the now-dead `bridge_paused` / `bridge_paused_errors`; preserve `bridge_message_not_sent_opt_total`.
- Remove the orphaned `QueueMessageProcessedBody` type (no producer or consumer left).

## Testing

- `Test_isForgedMessage` (TDD): forged → increments; legit → doesn't, via a configurable mock bridge (no `prometheus/testutil` dependency).
- `go build ./packages/relayer/...`, `go vet ./packages/relayer/...`, `go test ./packages/relayer/indexer/...`, `golangci-lint` (0 issues).

## Deployment

A companion `k8s-configs` change removes the two watchdog ArgoCD apps + values. **Merge the relayer image first** so forged-message detection coverage is never dropped; then delete the two orphaned `*-MessageProcessed-queue`s.


packages/relayer/indexer/handle_message_processed_event.go
```diff
@@ -81,26 +95,21 @@ func (i *Indexer) handleMessageProcessedEvent(
...
-	}
+	return nil
+}
 
-	marshalledMsg, err := json.Marshal(msg)
+// isForgedMessage reports whether the bridge that should have originated this
+// message (destBridge) has no record of having sent it — the signature of a
+// forged message. Relocated from the removed watchdog.
+func (i *Indexer) isForgedMessage(message bridge.IBridgeMessage) (bool, error) {
+	sent, err := i.destBridge.IsMessageSent(nil, message)
```
RogerLamTd: should be srcBridge i think?

packages/relayer/indexer/handle_message_processed_event.go
```diff
@@ -66,12 +65,27 @@ func (i *Indexer) handleMessageProcessedEvent(
 		return nil
 	}
 
+	// Forged-message detection (relocated from the removed watchdog): if the
+	// bridge that should have originated this message has no record of sending
+	// it, alert. Runs before saveEventToDB so an RPC failure leaves no
+	// half-saved state and the event is cleanly re-indexed on retry.
+	forged, err := i.isForgedMessage(message)
+	if err != nil {
+		return errors.Wrap(err, "i.isForgedMessage")
```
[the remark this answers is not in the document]
davidtaikocha: Valid — fixed in 4bfb56e. `saveEventToDB` now runs before the forged-message check, and the check is best-effort: on an `IsMessageSent` RPC error it logs and returns nil instead of propagating, so a transient origin-chain (destBridge) outage no longer drops the event. The filter loop advancing `latestIndexedBlockNumber` on a swallowed error was exactly the drop path. Added a regression test asserting the event is still indexed when the check's RPC fails.

packages/relayer/indexer/handle_message_processed_event.go
davidtaikocha: I believe `destBridge` is correct here (the src/dest naming is genuinely confusing). This indexer filters `MessageProcessed` from `i.bridge` = `SrcBridgeAddress`, i.e. the chain where the event fires — which is the message's *destination* chain. The message was *sent* on the other chain = `DestBridgeAddress` = `i.destBridge`, so `destBridge.IsMessageSent(message)` asks the origin bridge "did you send this?". Using `srcBridge` would ask the *receiving* bridge if it sent the message — false for every legitimate message, so it would flag them all as forged.

This also matches the original watchdog verbatim (it used `destBridge.IsMessageSent` with the same addresses). Concretely for l1↔l2: `MessageProcessed` on L1 (`srcBridge` `0xd60247…`) are L2→L1 messages, so we check the L2 bridge = `destBridge` `0x1670…0001`.

Happy to add a clarifying code comment given how easy this is to misread — lmk.

YoGhurt111 (approved):

PR #21891, opened 2026-07-02
feat(relayer): add blacklist messagehash functionality

RogerLamTd:
Adds `ignoredMsgHashes` / `IGNORED_MSG_HASHES` support to the relayer indexer for known `MessageProcessed` hashes.

Ignored hashes are skipped only after removed-log and confirmation handling. Once skipped, they intentionally bypass forged-message alerting and `saveEventToDB`, so known forged historical messages do not keep producing alerts or persisted reindexed records.


packages/relayer/indexer/handle_message_processed_event.go
```diff
@@ -36,6 +36,11 @@ func (i *Indexer) handleMessageProcessedEvent(
 		return nil
 	}
 
+	if _, ignored := i.ignoredMsgHashes[event.MsgHash]; ignored {
+		slog.Warn("skipping ignored MessageProcessed msgHash", "msgHash", event.MsgHash)
```
davidtaikocha: **Logs as a raw byte array, not hex.** `event.MsgHash` is `[32]byte`, so slog renders `msgHash="[0 0 0 … 1]"`. Every other hash log in this package uses `.Hex()`, and this WARN is the only operator-facing signal that a blacklist entry fired — it needs to be greppable against the configured hash.

```suggestion
		slog.Warn("skipping ignored MessageProcessed msgHash", "msgHash", common.Hash(event.MsgHash).Hex())
```
Needs adding `"github.com/ethereum/go-ethereum/common"` to the imports.

packages/relayer/indexer/handle_message_processed_event.go
```diff
@@ -36,6 +36,11 @@ func (i *Indexer) handleMessageProcessedEvent(
 		return nil
 	}
 
+	if _, ignored := i.ignoredMsgHashes[event.MsgHash]; ignored {
```
davidtaikocha: **Design check (not a bug).** Placing the skip here suppresses both the forged-detection alert *and* `saveEventToDB`, so blacklisted hashes leave no DB record. The prior watchdog-relocation deliberately runs forged-detection *before* the save so forged events are still persisted. "Suppress the noise entirely" is reasonable for known-forged hashes, but since it changes what lands in the DB, could you confirm the intent and note it in the PR description? (Body is currently empty.)

packages/relayer/indexer/config.go
```diff
@@ -136,3 +146,36 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
...
+
+	hashes := make(map[common.Hash]struct{})
+
+	for _, raw := range strings.Split(value, ",") {
+		msgHash := strings.TrimSpace(raw)
+		if msgHash == "" {
+			continue
+		}
+
+		hexHash := strings.TrimPrefix(msgHash, "0x")
```
davidtaikocha: **Uppercase `0X` prefix is rejected.** `TrimPrefix(msgHash, "0x")` only strips lowercase, so a valid `0X…` hash fails the length check below with the misleading `expected 32-byte hex hash`. `common.HexToHash` (used elsewhere and in this PR's own test) accepts both cases.

```suggestion
		hexHash := strings.TrimPrefix(strings.ToLower(msgHash), "0x")
```

packages/relayer/indexer/config.go
```diff
@@ -136,3 +146,36 @@ func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
 		},
 	}, nil
 }
+
+func parseIgnoredMsgHashes(value string) (map[common.Hash]struct{}, error) {
```
davidtaikocha, line 150: **No direct tests for this parser.** Coverage is a single happy-path assertion in `TestNewConfigFromCliContext`. Since strictness is the whole point of a blacklist, the untested surface is where the risk lives — please add a table-driven `Test_parseIgnoredMsgHashes` covering: comma-separated multi-hash, duplicate collapse, wrong length → error, non-hex → error, `0x`/`0X` prefix, and empty/whitespace → nil. (Package CLAUDE.md targets >95%.)

packages/relayer/indexer/indexer.go
```diff
@@ -239,6 +241,7 @@ func InitFromConfig(ctx context.Context, i *Indexer, cfg *Config) (err error) {
 	i.ctx = ctx
 
 	i.minFeeToIndex = i.cfg.MinFeeToIndex
+	i.ignoredMsgHashes = cfg.IgnoredMsgHashes
```
davidtaikocha, line 244: **Minor — surface the blacklist size at boot.** A misconfigured/over-broad list silently drops legit `MessageProcessed` events (only a per-event WARN). An info log here lets operators sanity-check at startup:

```go
slog.Info("ignoredMsgHashes loaded", "count", len(i.ignoredMsgHashes))
```

packages/relayer/indexer/indexer_test.go
```diff
@@ -46,3 +50,30 @@ func newTestService(syncMode SyncMode, watchMode WatchMode) (*Indexer, relayer.B
 		eventName:        relayer.EventNameMessageSent,
 	}, b
 }
+
+func TestHandleMessageProcessedEventSkipsIgnoredMessageHash(t *testing.T) {
+	ignoredHash := common.HexToHash("0x1")
```
davidtaikocha: **Minor — use a value the parser would accept.** `common.HexToHash("0x1")` pads to a full hash, but production `parseIgnoredMsgHashes` would *reject* the short `0x1` form. The test still validates the mechanism (it injects the map directly), but a full 64-hex-char constant reflects a realistically-configured value. Optional: assert the mock `EventRepository.SavedCount()==0` to prove `saveEventToDB` was skipped directly, rather than via the `IsMessageSent` proxy.

davidtaikocha (commented):
## Review — blacklist messagehash

Verdict: **merge with fixes.** Core mechanism is correct, compiles, nil-safe, and the skip test genuinely guards the path (`IsMessageSentCalls==0` fails if the skip is removed). Nice work.

Two small code fixes, one design decision to confirm, plus minor nits — details inline. Nothing critical.

- **Fix:** `msgHash` logs as a raw byte array, not hex (breaks the file-wide `.Hex()` convention).
- **Fix:** uppercase `0X` prefix is rejected with a misleading "expected 32-byte hex hash" error.
- **Confirm:** the skip suppresses `saveEventToDB` too (not just the alert) — intended? Please note it in the PR description.
- **Coverage:** `parseIgnoredMsgHashes` has no direct tests (error paths / multi-hash / dedup).

Docs/CHANGELOG: no action needed — consistent with the preceding `MIN_FEE_TO_INDEX` flag and release-please auto-gen.

davidtaikocha (approved):

PR #21892, opened 2026-07-02
fix(protocol): reject SGX instance ids that overflow the uint32 proof field

dantaik:
## Problem

`SgxVerifier.verifyProof` decodes the instance id from a **4-byte (`uint32`)** field in the 89-byte proof:

```solidity
uint32 id = uint32(bytes4(_proof[:4]));
```

but ids are handed out from the **`uint256 nextInstanceId`** counter, which only ever increases and is never reused (`deleteInstances` does not free ids). Any instance assigned an id `>= 2**32` is written to storage correctly but can never be referenced from a proof — the id silently truncates to its low 32 bits, so `_isInstanceValid` looks up the wrong (or a nonexistent) slot.

This is a latent liveness gap, not an exploitable soundness issue: ids below `2**32` map 1:1, so no aliasing or collision is possible. It only matters after ~4.29B cumulative registrations, but the assumption that "every registered instance is reachable from a proof" is currently implicit.

## Fix

Make the invariant explicit by rejecting the first id that would not survive the `uint32` truncation, in `_addInstances` (the single path both `addInstances` and `registerInstance` funnel through):

```solidity
require(nextInstanceId <= type(uint32).max, SGX_INSTANCE_ID_OVERFLOW());
```

The largest `uint32`-representable id (`2**32 - 1`) still registers; the next one reverts with the new `SGX_INSTANCE_ID_OVERFLOW` error instead of creating an unreachable instance.

## Test

`test_addInstances_RevertWhen_InstanceIdOverflows` (runs against both `SecureSgxVerifier` and `InsecureSgxVerifier` via the shared suite): fast-forwards `nextInstanceId` to `type(uint32).max` with `stdstore`, asserts the max-id registration succeeds, then asserts the next registration reverts with `SGX_INSTANCE_ID_OVERFLOW`.

Full SGX verifier suite passes (151 tests).

---


dantaik:
@review

davidtaikocha (approved):

smtmfft (approved):

PR #21893, opened 2026-07-02
fix(protocol): let the owner recover a front-run SGX instance registration

dantaik:
Closing: the front-run this PR guarded against is not exploitable in either verifier configuration, so the `promoteInstances` recovery mechanism is unnecessary.

- **`registrar` set (transitional):** `registerInstance` reverts `SGX_NOT_REGISTRAR` for every caller except the registrar (the multisig), so nobody can front-run it. `registrar` is `immutable`, so this holds for the whole life of the deployment.
- **`registrar == 0` (end state):** registration is permissionless, but a replayed quote produces the *same* delayed instance bound to the *same* enclave key — the instance address comes from the attested `reportData`, not from `msg.sender`. The attacker can't steal it (no private key), can't block it (it becomes valid after the delay, and the enclave holds the key), and can't change its timing (a non-owner registration is delayed either way). Net effect is a wasted-gas revert of the prover's duplicate tx; the prover simply adopts the already-registered instance.

The residual item is off-chain, not on-chain: the prover client should treat `SGX_ALREADY_ATTESTED` + an existing `InstanceAdded(K)` as success and wait out the validity delay rather than erroring.


dantaik:
@review

PR #21894, opened 2026-07-02
fix(protocol): make SGX MRENCLAVE/MRSIGNER untrust permanent (no silent revival)

dantaik:
## Problem

`verifyProof` re-checks the *current* MRENCLAVE/MRSIGNER allowlist (via `_isEnclaveStillTrusted`), so `setMrEnclave(x, false)` / `setMrSigner(x, false)` revoke already-registered instances at proof time. But the re-check keys only off the current boolean flag:

```solidity
return mrEnclaveState[_instance.mrEnclave].trusted && trustedUserMrSigner[_instance.mrSigner];
```

So the revocation is **reversible**: if the owner untrusts a measurement/signer to evict a fleet and later re-trusts the *same* value — e.g. to onboard a new fleet under it — every previously-revoked instance silently revives, including ones whose keys were suspected compromised.

This is asymmetric with the attribute-policy channel, which was deliberately engineered so that a removed pin can **never** re-enable old instances (monotonic version, `removeEnclaveAttributePolicy` → re-add gets a fresh version). The allowlist channel lacked the equivalent guarantee.

## Fix

Make untrust permanent for both channels, matching the "never revive" philosophy already used for attribute policies:

- On a **trusted → untrusted** transition, record the value in `revokedMrEnclave` / `revokedMrSigner`.
- Reject re-trusting a revoked value (`SGX_MR_ENCLAVE_REVOKED` / `SGX_MR_SIGNER_REVOKED`).
- Untrusting a value that was **never** trusted stays a no-op (it does not poison the value).

An allowlist removal can therefore never be silently undone. Operators onboard a new enclave build under a **fresh** measurement (which they would do anyway for a new build); concrete compromised instances are still revoked irreversibly with `deleteInstances`.

Scope notes for reviewers:
- Toggling the entire allowlist off (`toggleLocalReportCheck`) remains, by design, a separate global action and is unchanged (still covered by `test_verifyProof_AllowlistDisabledSkipsReCheck`).
- **Alternative considered:** a monotonic per-value "trust epoch" (mirroring the attribute-policy version) that lets the same value be re-trusted for *new* registrations while keeping old instances dead. That is less restrictive but requires two new fields on the `Instance` struct (enclave + signer epoch) and a proof-path change. This PR takes the smaller, lower-risk permanent-revocation route; happy to switch to the epoch approach if you prefer re-trust to remain possible.

## Tests

- `test_verifyProof_RevokedInstanceCannotBeRevived`: register → untrust MRENCLAVE (proof now rejected) → re-trust reverts `SGX_MR_ENCLAVE_REVOKED`, so the instance can never come back.
- `test_setMrEnclave_RevertWhen_ReTrustAfterUntrust`, `test_setMrSigner_RevertWhen_ReTrustAfterUntrust`.
- `test_setMrEnclave_UntrustWhenNeverTrustedDoesNotRevoke` (guards the transition-only arming).

All run against both `SecureSgxVerifier` and `InsecureSgxVerifier`. Full SGX verifier suite passes (157 tests).

---


dantaik:
@review

smtmfft (approved):

smtmfft (approved):

PR #21895, opened 2026-07-02
feat(protocol): add optional SGX quote-freshness gate (bound quote age)

dantaik:
## Problem

`INSTANCE_EXPIRY` (90 days) bounds an SGX signing key's exposure only **after** registration — the contract comment frames it as side-channel protection "requiring periodic re-attestation with new keypairs." But nothing bounds how old the attested **quote** is at registration time: SGX enclave reports carry no timestamp, and `reportData` currently holds only the instance address.

So a quote generated a long time ago — or one whose in-enclave key was slowly side-channel-extracted over months — can be registered today and its key treated as fresh for the full delay + 90-day window. The "periodic re-attestation with new keypairs" guarantee only caps *post-registration* exposure; *pre-registration* key age is unbounded.

## Fix

Add an owner-toggled `requireQuoteFreshness` gate, **off by default** (fully backward compatible). When enabled, `registerInstance` requires the enclave to commit a recent L1 block into `reportData`, immediately after the 20-byte instance address:

```
reportData[0:20]  = instance address (unchanged)
reportData[20:28] = L1 block number (uint64, big-endian)
reportData[28:60] = that block's hash (bytes32)
```

and checks it on-chain:

```solidity
bytes32 actualBlockHash = blockhash(quoteBlock);
require(actualBlockHash != bytes32(0) && actualBlockHash == quoteBlockHash, SGX_STALE_QUOTE());
```

`blockhash` returns zero outside the most recent 256 blocks (~51 min on mainnet), so a non-zero match bounds the quote's age to that window. Both new fields live inside the enclave-report region already bound to the authenticated body by the existing verified-body-hash check, so they cannot be forged independently of the attestation.

### Why a toggle / what this ships

Enforcing freshness requires the prover (raiko) to embed the block commitment, which it does not do today. This PR ships the **on-chain half**: the check plus its owner toggle, defaulting off so current provers and all existing flows are unaffected. Enabling it is a coordinated change with prover support. The alternative (a hard requirement) would break registration immediately, so the gate is the backward-compatible path.

## Tests

- `test_registerInstance_SucceedsWhen_FreshQuote`: gate on, commit a recent correct block hash → registers.
- `test_registerInstance_RevertWhen_StaleQuote`: gate on, commit a block outside the 256-block window (hash reads back zero) → `SGX_STALE_QUOTE`.
- `test_registerInstance_RevertWhen_FreshQuoteHashMismatch`: recent block number, wrong hash → `SGX_STALE_QUOTE`.
- `test_registerInstance_FreshnessOffIgnoresBlockCommitment`: default-off path unaffected.
- `test_toggleQuoteFreshnessCheck_togglesAndEmits` / `_RevertWhen_NotOwner`.

All run against both `SecureSgxVerifier` and `InsecureSgxVerifier`. Full SGX verifier suite passes (161 tests).

---


dantaik:
@review

smtmfft (approved):
Looks good to me. This is useful forward-looking hardening for SGX quote registration. It is backward-compatible while disabled, and we will track the raiko2-side reportData change separately before enabling the freshness gate.

smtmfft:
Maybe one concern: this would also require a prover to register its instance within the 256-block window. Since we are currently still in multi-sig mode, this may not be achievable in practice.
I think this feature is useful. Should we merge it with the freshness check disabled for now, and enable it later once we move from multi-sig to permissionless mode? @ggonzalez94 @dantaik @davidtaikocha

dantaik:
@review

dantaik:
@smtmfft 现在如果registrar是address(0),那么就会要求freshness，不然就不要求。你再看一眼。

smtmfft (approved):
Reviewed the latest head. The implementation now matches the registrar-based policy: when , permissionless  requires the recent L1 block commitment; when a non-zero registrar is configured, only the registrar can call and the freshness gate is skipped. The dedicated registrar test covers the no-commitment path.\n\nCode looks good to me. Minor non-blocking note: the PR description still describes the earlier owner-toggle/default-off design, so it should be updated before merge to avoid operator confusion.

smtmfft:
Clarifying my approval comment above because shell interpolation stripped two inline code snippets from the review body.

I reviewed the latest head and the implementation matches the registrar-based policy:

- When `registrar == address(0)`, permissionless `registerInstance` requires the recent L1 block commitment and enforces the freshness gate.
- When `registrar != address(0)`, only the registrar can call `registerInstance`, and the freshness gate is skipped.
- `test_registerInstance_SkipsFreshnessWhen_RegistrarSet` covers the no-commitment registrar path.

Code looks good to me. Minor non-blocking note: the PR description still describes the earlier owner-toggle/default-off design, so it should be updated before merge to avoid operator confusion.

PR #21905, opened 2026-07-03
feat(bridge-ui): check if retry failed is wrapped quota error

RogerLamTd:


packages/bridge-ui/src/components/Dialogs/ClaimDialog/ClaimDialog.svelte
```diff
@@ -156,23 +183,16 @@
...
-            title: $t('bridge.errors.claim.quota_reached.title'),
-            message: $t('bridge.errors.claim.quota_reached.message'),
-          });
+          showQuotaReachedToast();
         } else {
-          errorToast({
-            title: $t('bridge.errors.unknown_error.title'),
-            message: $t('bridge.errors.unknown_error.message'),
-          });
+          if (!(await showQuotaToastIfClaimIsBlocked())) {
```
davidtaikocha: **❌ First-claim path is still uncovered — this reactive check can't catch it.**

This handles the retry case correctly (`retryMessage` reverts `B_RETRY_FAILED` → `ContractFunctionExecutionError` → here → quota toast). But when the message is still `NEW` (relayer hasn't processed it — e.g. a low/zero-fee withdrawal the relayer won't auto-claim), the user's claim runs `processMessage`, which **catches** the quota revert internally → status `RETRIABLE` → the tx **succeeds**. `handleClaim` then shows the success toast and `handleClaimError` is never invoked, so this check never runs. Result: false "success", no quota message — and an error-handler approach structurally can't catch a claim that doesn't throw.

**Complete fix: check quota proactively.** `checkQuota.ts` is already ideal for it — call `isClaimBlockedByQuota(bridgeTx)` at the *start* of the claim handler; if blocked, show the quota toast and skip the claim. That covers both paths and avoids burning gas on a doomed `processMessage` that just leaves a `RETRIABLE` mess. Keep this reactive check as a fallback.

ggonzalez94: Fixed in 3ded43134. Claims and retries now run the quota guard before submitting the transaction, and the reactive fallback remains for direct quota reverts or wrapped retry failures. Also added coverage for the proactive guard, wrapped retry fallback, and ETH quota lookup.

Verified locally with bridge-ui unit tests, svelte-check, and a skip-env build.

PR #21914, opened 2026-07-04
chore(taiko-client-rs): simplify and harden the test harness

davidtaikocha:
## What

A cleanup + robustness pass over the `just test` flow for `taiko-client-rs` (`justfile`, `tests/entrypoint.sh`, `tests/deploy.sh`, `tests/docker/docker-compose.test.yaml`, nextest config). No test logic changes — only the harness that stands up the L1/L2 docker stack, deploys contracts, and invokes nextest.

## Why

The harness had a silent-hang failure mode (unbounded readiness loops with suppressed errors), some actively harmful behavior (`restart: unless-stopped` resurrecting a crashed anvil as a fresh empty chain), and accumulated dead code.

## Changes

**Robustness**
- **Preflight + bounded waits** — check `cast`/`forge`/`jq` are on `PATH` up front, and cap each RPC readiness wait at 120s, dumping `docker compose logs` on timeout. A missing binary or a container that never comes up now fails loudly instead of spinning forever.
- **`trap cleanup EXIT`** only — `KILL` is untrappable and the extra `INT`/`ERR` handlers caused redundant teardown.
- **Drop `restart: unless-stopped`** from the test containers — a mid-run anvil/reth crash should surface as a failure, not silently restart as a fresh, empty chain.
- **Fail-loud env extraction** — split `declare`/`assign` for `L2_GENESIS_HASH` and `SHASTA_INBOX` so a failed `cast`/`jq` propagates under `set -e` (previously masked by `export`).

**Simplification / dead code**
- Remove seven exported env vars with **zero readers** in `crates/` or the scripts: `L2_HTTP_1`, `TREASURY`, `L1_CONTRACT_OWNER_PRIVATE_KEY`, `L1_PROVER_PRIVATE_KEY`, and the post-deploy `TAIKO_TOKEN` / `FORCED_INCLUSION_STORE` / `COMPOSE_VERIFIER` re-exports.
- Replace the `curl | jq | sed` genesis-hash fetch with `cast block 0 --field hash`; `jq | sed 's/\"//g'` → `jq -r`.
- Deduplicate the deploy key — the deploy script broadcasts via `vm.envUint("PRIVATE_KEY")`, so the inline `PRIVATE_KEY=$PRIVATE_KEY` re-export and the `--private-key` flag in `deploy.sh` were both redundant.
- Collapse the two identical `alethia-reth` L2 services into a YAML anchor.

**DX**
- `pull_policy: ${PULL_POLICY:-always}` — CI keeps `always`; locally `PULL_POLICY=missing just test` enables fast/offline runs.
- `just test *ARGS` forwards to nextest, so `just test -E 'test(some_name)'` targets a single test.
- Run the entrypoint from the package root (location-independent) and echo the resolved `PROTOCOL_DIR` (the local working-tree vs. CI pinned-tag protocol source is otherwise invisible).
- Move `nextest.toml` → `.config/nextest.toml` (nextest's default discovery path) and drop `--config-file`.

## Verification

- `bash -n` + `shellcheck` clean on both scripts.
- `docker compose -f tests/docker/docker-compose.test.yaml config` renders correctly with anchors expanded; `PULL_POLICY` override honored.
- `just --show test` resolves the `*ARGS` passthrough.
- Full end-to-end run is exercised by CI (this workflow).

## Follow-up (not in this PR)

Two worktrees can't run `just test` in parallel — fixed container names + host ports + a shared default compose project name mean `down -v` in one worktree tears down the other's stack. Fixing it properly (per-worktree compose project + dynamic host ports, which cascades into the ports the Rust harness connects to) is a larger, harness-facing change and will be a separate PR.


packages/taiko-client-rs/justfile
```diff
@@ -18,8 +18,8 @@ clippy-fix:
 udeps:
   cargo +{{toolchain}} udeps --all-targets
 
-test:
-  ./tests/entrypoint.sh
+test *ARGS:
+  ./tests/entrypoint.sh {{ARGS}}
```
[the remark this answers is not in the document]
davidtaikocha: Valid catch — confirmed and fixed in 81876ab0e. `{{ARGS}}` interpolated the space-joined variadic unquoted, so `just test -E 'test(some_name)'` died with a shell syntax error on the parens before the harness ran. Switched to `set positional-arguments` + `"$@"`, which forwards args as intact positional parameters (verified with filters containing spaces and parentheses).

YoGhurt111 (approved):

packages/taiko-client-rs/tests/entrypoint.sh
```diff
@@ -53,55 +65,49 @@ cleanup() {
     "${DOCKER_COMPOSE[@]}" -f "$COMPOSE_FILE" down -v
 }
 
+echo "Starting docker compose services..."
 "${DOCKER_COMPOSE[@]}" -f "$COMPOSE_FILE" up -d
-trap cleanup EXIT INT KILL ERR
+trap cleanup EXIT
```
YoGhurt111, line 70: Maybe we need to consider the fingdings `Missing INT / TERM trap may leave containers running on Ctrl+C`?

davidtaikocha: Good to double-check this one, but I'm fairly sure it's a false positive — the finding reads like a generic shellcheck/POSIX-`sh` rule, and it doesn't hold for a bash script with an `EXIT` trap.

In bash, `trap cleanup EXIT` already runs on Ctrl+C: the shell runs the EXIT trap and then re-raises the signal (so the exit status still reflects the signal). I verified it faithfully — SIGINT delivered to the foreground process group — on both bash 3.2 (macOS) and bash 5.2 (Linux):

- **SIGINT** (Ctrl+C) → cleanup runs ✅
- **SIGTERM** (`kill`, CI cancel, `timeout`) → cleanup runs ✅
- **SIGHUP** (terminal closed) → cleanup runs ✅
- **SIGKILL** → not run, but that's untrappable by anything (the old `KILL` sigspec was a no-op)

The POSIX-`sh` case the rule is worried about (EXIT not firing on signals) can't apply here — the script is bash-only anyway (arrays, `BASH_SOURCE`, `pipefail`).

Adding `INT`/`TERM` back would actually re-introduce the double-teardown this PR removes, since cleanup runs once per matching sigspec. On a single Ctrl+C:

| trap | `docker compose down -v` runs |
|---|:--:|
| `EXIT` (this PR) | 1× |
| `EXIT INT` | 2× |
| `EXIT INT KILL ERR` (original) | 3× |

So `EXIT`-only fires exactly once, still covers INT/TERM/HUP, and keeps the killed-by-signal exit status. Happy to drop in an inline comment noting the EXIT trap covers signals if that makes it clearer for the next reader.

PR #21924, opened 2026-07-06
feat(protocol): enable Unzen hardfork — re-enable forced inclusions, mandate ZK proofs

ggonzalez94:
## What

The L1 contract bundle for the Unzen hardfork: a new `MainnetInbox` implementation plus a fresh SGX/attestation stack. The DAO proposal that activates it lives in the stacked PR #21935 — it needs the deployed addresses, so it finalizes after this merges and the deploy scripts run.

1. **Forced inclusions re-enabled.** Reverts the incident-response disables (`fd8a69852`/`b73608696`): submissions accepted, proposers consume the queue, due-scan restored verbatim. Permissionless proposing and proving-by-age stay disabled — the latter was the incident's on-chain amplifier.
2. **Stale queue voided.** `Inbox.init3()` moves the queue head past the one pending entry, submitted mid-incident by a non-attacker address. Its blob leaves the EIP-4844 retention window ~2026-07-10, so it can never be derived — re-enabling the due-check without voiding it would force proposers to consume an inclusion whose data no longer exists.
3. **ZK proof mandated on every batch.** `ZkRequiredVerifier` replaces `MainnetVerifier` (now marked deprecated), which still accepts SGX-geth + SGX-reth with zero ZK proofs — the combination that finalized the June forged proofs. Accepted pairs: (SGX_GETH | SGX_RETH) + (RISC0 | SP1), or RISC0 + SP1. The SGX+SGX pair is structurally impossible, not just unconfigured.
4. **SGX verifiers replaced, both flavors.** Two new fail-closed `SecureSgxVerifier`s (geth + reth) carrying the post-v3.1.0 hardening (permanent untrust, uint32 id overflow rejection, quote-freshness gate), both wired to one audited **upstream** Automata DCAP attestation entrypoint — the MRENCLAVE/MRSIGNER allowlists live in the verifiers now, so per-flavor attesters are obsolete. The Proposal0017 verifiers are immutably wired to the old pre-incident vendored attestation; since every link in the chain is immutable (verifier → attestation, compose → sub-verifiers, Inbox → compose), this upgrade is the only deployment window that avoids a second Inbox upgrade later.

## Deployment: two scripts, both straight-line

The upstream Automata code only compiles under via_ir (quarantined in `profile.layer1o`), so deployment is deliberately two simple scripts with no branching:

1. `DeployAutomataDcapAttestation` (layer1o) — attestation entrypoint, fee pinned to zero, ownership to the DAO controller.
2. `DeployUnzenContracts` (layer1) — hard-requires `DCAP_ATTESTATION` to point at deployed code, then deploys both SecureSgxVerifiers → ZkRequiredVerifier → MainnetInbox bottom-up. The SGX verifiers deploy with **admin.taiko.eth as initial owner and registrar**: the multisig configures the trust registries manually (`setMrEnclave` → `setEnclaveAttributePolicy` → `setMrSigner`), registers the raiko instances (owner registrations skip the 24h validity delay), and transfers ownership to the DAO controller, which #21935 accepts.

This keeps the Taiko contracts on their canonical non-IR bytecode.

`LibL1Addrs` also gains the live Proposal0017 SGX verifier addresses and corrects the stale `SP1_RETH_VERIFIER` constant to the wired `0x73A0Db…` (verified on-chain).

## Testing

Restored pre-incident forced-inclusion suites, converted permissionless-propose tests to revert expectations, added `InboxInit3.t.sol` and ZkRequiredVerifier combo tests (geth+ZK accepted, SGX+SGX rejected). All layer1 tests pass; `MainnetInbox`/`DevnetInbox` storage layouts byte-identical to HEAD; both foundry profiles build.

## Before the fork can execute (not review blockers)

- Merge this → run the two deploy scripts on chain 1 → finalize #21935 (fill constants, generate the action file).
- Proposer client support for `numForcedInclusions` and raiko ZK-per-batch capacity must be in production first; un-updated proposers revert on `propose()` once an inclusion is due.
- Pre-execution: admin.taiko.eth (initial owner) sets the trust config and registers the raiko instances on both new SGX verifiers, then hands ownership to the DAO (accepted in #21935). RISC0+SP1 keeps proving alive if SGX registration lags.


packages/protocol/contracts/layer1/verifiers/compose/ZkRequiredVerifier.sol
davidtaikocha: I think we still allow SGX_GETH (gaiko2), right?

davidtaikocha (approved):

packages/protocol/contracts/layer1/verifiers/compose/ZkRequiredVerifier.sol
ggonzalez94: yes, it was missing in the proposal and now added on [ea81953](https://github.com/taikoxyz/taiko-mono/pull/21924/commits/ea81953bd9b49a38660c7b95f124796828b16b38)

davidtaikocha (approved):

PR #21934, opened 2026-07-07
test(taiko-client): isolate tx manager selector tests

weifanglab:
## Summary

Isolate the `TxMgrSelector` tests so they no longer share a package-level selector instance.

The previous tests mutated shared state through `RecordPrivateTxMgrFailed`, which made the package order-dependent under `go test -shuffle`. This keeps each test self-contained and adds coverage for:

- selecting the private transaction manager when available
- falling back to the normal transaction manager after a private transaction manager failure
- retrying the private transaction manager after the retry interval has elapsed

## Verification

- `go test -shuffle=on -count=20 ./pkg/utils`
- `go test -count=1 ./pkg/utils`
- `typos packages/taiko-client/pkg/utils/txmgr_selector_test.go`

### DIFF
--- a/packages/taiko-client/pkg/utils/txmgr_selector_test.go
+++ b/packages/taiko-client/pkg/utils/txmgr_selector_test.go
@@ -2,28 +2,63 @@ package utils
 
 import (
 	"testing"
+	"time"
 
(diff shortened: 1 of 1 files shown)


davidtaikocha:
Closed due to CI failure, and the improvement is not obvious.

PR #21943, opened 2026-07-10
chore(taiko-client-rs): harden whitelist preconfirmation ingress

davidtaikocha:
## Summary

- bind every queued preconfirmation to the authenticated parent hash and reject canonical-parent changes before engine injection
- return explicit `Inserted`, `AlreadyMaterialized`, and `Stale` outcomes so stale races cannot advance local status or be logged as successful inserts
- defer cached parent-mismatch races into normal ancestry recovery instead of aborting the drain pass as a fatal error
- reject stale envelopes before cache/recent/EOS admission, and evict stale recent envelopes before serving P2P responses
- restore the Go-compatible `/status` key `highestUnsafeL2PayloadBlockID`

## Root cause

Preconfirmation validation and engine injection were separated by asynchronous queueing without carrying the validated parent identity into the production path. At the same time, stale submissions returned undifferentiated success, so callers could mutate local tracking after event sync had already confirmed the same block height. Incoming P2P envelopes also entered caches before consulting the confirmed tip, and the status field relied on generic camel-case serialization that emitted `Id` instead of Go's `ID`.

## Safety and compatibility

- strengthens WLP-INV-003 and WLP-INV-004 by enforcing `block_number <= head_l1_origin` at ingress, queue processing, cache admission, and response serving
- strengthens WLP-INV-009 by checking the authenticated parent again inside the serialized production path
- preserves WLP-INV-002 readiness gating
- restores WLP-INV-010 JSON compatibility
- preserves Go REST compatibility: block-insertion failures remain HTTP 500 with a JSON `error` body; Catalyst retries all non-success responses through the same path
- does not change gossipsub topics, SSZ encoding, message IDs, signature validation, or discovery behavior, so the driver remains compatible with the Go `taiko-client` preconfirmation network

## Verification

- `just fmt`
- `just clippy-fix` (workspace clippy with `-D warnings` passed)
- `just test` (299 passed, 0 skipped)

The Foundry deployment phase still emits its existing source-definition parsing warnings; deployment and the complete test run exit successfully.

### DIFF
--- a/packages/taiko-client-rs/crates/driver/src/sync/event.rs
+++ b/packages/taiko-client-rs/crates/driver/src/sync/event.rs
@@ -205,13 +205,10 @@ fn resolve_reconnect_start_block(
         .map_or(startup_anchor_block_number, |finalized| overlap_start_block_number.min(finalized))
 }
 
-/// Return the terminal stale outcome when a target is at or below the confirmed tip.
+/// Return whether a preconfirmation target is at or below the confirmed tip.
 #[inline]
-fn stale_preconf_outcome(
-    block_number: u64,
-    confirmed_tip: u64,
-) -> Option<PreconfSubmissionOutcome> {
-    (block_number <= confirmed_tip).then_some(PreconfSubmissionOutcome::Stale)
+fn is_stale_preconf(block_number: u64, confirmed_tip: u64) -> bool {
+    block_number <= confirmed_tip
 }
 
 /// Responsible for following inbox events and updating the L2 execution engine accordingly.
@@ -411,8 +408,7 @@ impl EventSyncer {
                         continue;
                     }
                 };
-                if let Some(outcome) = stale_preconf_outcome(block_number, head_l1_origin_block_id)
-                {
+                if is_stale_preconf(block_number, head_l1_origin_block_id) {
                     DriverMetrics::preconf_stale_dropped_total().inc();
                     DriverMetrics::preconf_stale_dropped_ingress_total().inc();
                     warn!(
@@ -420,7 +416,7 @@ impl EventSyncer {
                         head_l1_origin_block_id,
                         "dropping stale preconfirmation payload in ingress loop"
                     );
-                    let _ = job.respond_to.send(Ok(outcome));
+                    let _ = job.respond_to.send(Ok(PreconfSubmissionOutcome::Stale));
                     DriverMetrics::preconf_queue_depth().set(rx.len() as f64);
                     continue;
                 }
@@ -841,14 +837,14 @@ impl EventSyncer {
             Some(origin) => origin.block_id.to::<u64>(),
             None => 0,
         };
-        if let Some(outcome) = stale_preconf_outcome(block_number, head_l1_origin_block_id) {
+        if is_stale_preconf(block_number, head_l1_origin_block_id) {
             DriverMetrics::preconf_stale_dropped_total().inc();
             DriverMetrics::preconf_stale_dropped_before_enqueue_total().inc();
             warn!(
                 block_number,
                 head_l1_origin_block_id, "dropping stale preconfirmation payload before enqueue"
             );
-            return Ok(outcome);
+            return Ok(PreconfSubmissionOutcome::Stale);
         }
         if is_materialized {
             return Ok(PreconfSubmissionOutcome::AlreadyMaterialized);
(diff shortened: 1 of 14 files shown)


packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -802,21 +813,21 @@ impl EventSyncer {
...
+    ) -> Result<PreconfSubmissionOutcome, DriverError> {
         let tx = self.preconf_tx.as_ref().ok_or(DriverError::PreconfirmationDisabled)?;
 
         // Reject early if strict ingress gating is not satisfied yet.
         if !self.preconf_ingress_ready.load(Ordering::Acquire) {
             return Err(DriverError::PreconfIngressNotReady);
         }
 
-        if preconfirmation_payload_is_materialized(&self.rpc, &payload).await? {
+        let is_materialized = preconfirmation_payload_is_materialized(&self.rpc, &payload).await?;
```
YoGhurt111: [P1] Bind the submission outcome to the exact block. This materialization check runs outside the router mutex, and both `AlreadyMaterialized` and `Inserted` return without the block hash that was actually observed or inserted. The API subsequently releases the serialized production path and reads the block again by height. A same-parent, same-height sibling can become canonical in between, causing the handler to sign sibling B's hash while constructing the envelope with request A's transactions. The receiver currently validates the signer and payload shape but does not recompute the block hash from the header/transactions, so the inconsistent envelope can be propagated and imported. Although this race predates the PR, the new outcome contract leaves it open in the exact path this PR is hardening. Please perform the materialization decision inside the serialized section and return the exact block hash (or the full `EngineBlockOutcome`) for both inserted and materialized cases; the handler should then fetch and validate by that hash rather than rereading by height.

packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/importer/ingress.rs
```diff
@@ -33,17 +33,35 @@ use super::{
...
+
 impl WhitelistPreconfirmationImporter {
     /// Cache a validated envelope and persist EOS epoch mapping when applicable.
     async fn ingest_validated_envelope(
         &mut self,
         envelope: Arc<WhitelistExecutionPayloadEnvelope>,
         ingress_source: &'static str,
-    ) {
+    ) -> Result<()> {
+        let confirmed_tip = self.head_l1_origin_block_id().await?;
```
YoGhurt111: [P2] Preserve an accepted envelope when the confirmed-tip RPC fails transiently. At this point gossipsub has already authenticated and accepted the message, but this new fallible lookup occurs before the envelope is written to the pending, recent, or EOS caches. A temporary provider error therefore drops a valid envelope that the old path retained for a later drain. Recovery is not guaranteed for a tip or EOS payload with no child to trigger an ancestry request, and duplicate suppression prevents immediate redelivery. Please keep the validated envelope in a non-servable retry/quarantine state until the confirmed-tip lookup succeeds, then decide whether it may enter recent/EOS; this preserves fail-closed stale handling without losing the message.

YoGhurt111 (commented):
I found two issues that should be addressed before merging: one residual identity-binding race in the submission outcome path, and one newly introduced availability regression in P2P ingress.

packages/taiko-client-rs/crates/driver/src/sync/event.rs
davidtaikocha: Fixed in 16c857cb2. `PreconfSubmissionOutcome::{Inserted, AlreadyMaterialized}` now carry the exact block hash observed by the submission path, and the build handler resolves the block with `get_block_by_hash(bound_hash)` instead of re-reading by height — the parent/number verifications now run against that bound block. In the ingress loop the materialization decision moved inside the router lock, and `Inserted` takes its hash from the `EngineBlockOutcome` returned by the production path. The pre-enqueue duplicate fast path stays outside the lock but returns the hash it compared against, so the response is still bound to the exact block that satisfied the check even if a sibling becomes canonical afterwards.

The cached-import path now also verifies the returned hash against the operator-signed `envelope.block_hash` and purges the envelope from the recent cache on mismatch, so a receiver no longer imports-and-re-serves an envelope whose payload does not produce its claimed hash — the propagation half of the scenario described here.

Covered by `queued_preconfirmation_binds_inserted_outcome_to_produced_block`, `materialized_preconfirmation_returns_observed_block_hash`, and the updated loop tests.

packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/importer/ingress.rs
davidtaikocha: Fixed in 16c857cb2. A failed confirmed-tip read at admission is now treated as an unknown tip: the envelope is admitted (with a warn log) instead of dropped, so the gossip-authenticated message survives transient provider errors exactly like the old path.

I went with admit-on-failure rather than a quarantine state: after this PR the admission check is a best-effort filter — the authoritative `block_number <= head_l1_origin` checks run again at cache drain, at response serving, and in the driver ingress loop, so an envelope admitted with an unknown tip can never be served or imported while stale. Same no-loss property, no new envelope state, and `ingest_validated_envelope` is infallible again.

Unit-testing the failure path itself needs an RPC seam the importer doesn't have yet; that's tracked as separate follow-up work.

YoGhurt111 (approved):

PR #21944, opened 2026-07-10
ci(taiko-client-rs): enable signer recovery for net builds

johntaiko:
## Why

- Standalone `protocol` builds with the default `net` feature fail because anchor validation calls `TxEnvelope::recover_signer()` without enabling Alloy's signer-recovery implementation.
- Workspace-wide feature unification can mask this failure, so downstream consumers such as raiko2 discover it only after updating their dependency pin.

## How

- Enable `alloy-consensus/k256` only through the `net` feature, preserving the no-default-features path used by zkVM guests.
- Check both default and no-default-features `protocol` builds independently in CI.

## Tests

- `cargo check -p protocol --no-default-features --locked` — passed
- `cargo check -p protocol --locked` — passed
- `just fmt-check` — passed
- `just clippy` — passed
- `just test` — 291 passed, 0 skipped

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex

### DIFF
--- a/.github/workflows/taiko-client-rs--test.yml
+++ b/.github/workflows/taiko-client-rs--test.yml
@@ -103,6 +103,12 @@ jobs:
         working-directory: packages/taiko-client-rs
         run: just fmt-check
 
+      - name: Check protocol feature isolation
+        working-directory: packages/taiko-client-rs
+        run: |
+          cargo check -p protocol --no-default-features --locked
+          cargo check -p protocol --locked
+
       - name: Check clippy
         working-directory: packages/taiko-client-rs
         run: just clippy
--- a/.github/workflows/taiko-client-rs--test.yml
+++ b/.github/workflows/taiko-client-rs--test.yml
@@ -103,12 +103,6 @@ jobs:
         working-directory: packages/taiko-client-rs
         run: just fmt-check
 
-      - name: Check protocol feature isolation
-        working-directory: packages/taiko-client-rs
-        run: |
-          cargo check -p protocol --no-default-features --locked
-          cargo check -p protocol --locked
-
       - name: Check clippy
         working-directory: packages/taiko-client-rs
         run: just clippy
@@ -241,6 +235,10 @@ jobs:
         uses: extractions/setup-just@v3
         with:
           just-version: 1.5.0
+
+      - name: Check protocol feature isolation
+        working-directory: packages/taiko-client-rs
+        run: just check-features
 
       - uses: taiki-e/install-action@nextest
 
--- a/packages/taiko-client-rs/crates/protocol/Cargo.toml
+++ b/packages/taiko-client-rs/crates/protocol/Cargo.toml
@@ -10,6 +10,7 @@ default = ["net"]
 net = [
     "alloy/default",
     "alloy/signers",
+    "alloy-consensus/k256",
     "bindings",
     "tokio",
     "tokio-stream",
--- a/packages/taiko-client-rs/justfile
+++ b/packages/taiko-client-rs/justfile
@@ -14,6 +14,10 @@ fmt-check:
   rustup toolchain install {{toolchain}} && \
   cargo +{{toolchain}} fmt --check
 
+check-features:
+  cargo check -p protocol --no-default-features --locked && \
+  cargo check -p protocol --locked
+
 clippy:
   cargo clippy --workspace --all-features --no-deps --exclude bindings --exclude test-harness -- -D warnings -D missing_docs -D clippy::missing_docs_in_private_items
 


.github/workflows/taiko-client-rs--test.yml
```diff
@@ -103,6 +103,12 @@ jobs:
         working-directory: packages/taiko-client-rs
         run: just fmt-check
 
+      - name: Check protocol feature isolation
+        working-directory: packages/taiko-client-rs
+        run: |
```
davidtaikocha: **The guard is skipped on exactly the PRs most likely to trip it.**

This step lives in the `lint` job, whose condition on line 39 excludes dependabot branches:

```yaml
if: ${{ github.event.pull_request.draft == false && !startsWith(github.head_ref, 'release-please') && !startsWith(github.head_ref, 'dependabot') }}
```

The `test` job (line 117) has no such clause. So dependabot PRs run tests but skip lint — and feature-unification drift is most often introduced precisely when a dependency bump flips a default feature. This exact regression class would sail straight through the guard.

Worth noting the breakage this PR fixes arrived via #21941, a dependency bump.

**Fix:** either move this step into the `test` job, or drop the `dependabot` clause from the `lint` condition.

---

Separately and much smaller: this is inlined as raw YAML rather than a `justfile` recipe, so it breaks the `just fmt-check` / `just clippy` / `just test` convention the surrounding steps follow, and a developer can't reproduce it locally before pushing. A `just check-features` recipe would address both.

.github/workflows/taiko-client-rs--test.yml
```diff
@@ -103,6 +103,12 @@ jobs:
...
+          cargo check -p protocol --no-default-features --locked
+          cargo check -p protocol --locked
```
davidtaikocha: Verified this is a sound guard: `-p protocol` selects only that package as the build root, so under resolver v2 (declared in `packages/taiko-client-rs/Cargo.toml`) sibling-member features never unify onto `alloy-consensus`. That's precisely why the base commit fails here while the full-workspace build passes.

One note on coverage: `cargo check` doesn't build dev-deps or test targets. That's the **right** call for the `--no-default-features` leg — protocol's `[dev-dependencies]` pull `alloy = { features = ["node-bindings"] }`, which would re-unify features onto the lib and re-mask a guest-path regression. So please don't be tempted to add `--all-targets` to that line.

packages/taiko-client-rs/crates/protocol/Cargo.toml
```diff
@@ -10,6 +10,7 @@ default = ["net"]
 net = [
     "alloy/default",
     "alloy/signers",
+    "alloy-consensus/k256",
```
davidtaikocha, line 13: Verified this is the correct feature, not merely *a* feature that happens to work.

The `SignerRecoverable for EthereumTxEnvelope` impl in alloy-consensus 1.8.3 (`src/transaction/envelope.rs:835`) is gated:

```rust
#[cfg(any(feature = "secp256k1", feature = "k256"))]
```

Neither backend is in alloy-consensus's `default = ["std"]`, and `alloy/default` + `alloy/signers` were already present in `net` on the base commit (which still failed) — so this line is **necessary**, not redundant.

It's also the minimal correct choice:
- `alloy-primitives/k256` alone wouldn't do it — the impl gate keys off alloy-consensus's *own* `k256` feature.
- `alloy/k256` would work but fans out far wider (`alloy-core/k256`, `alloy-network?/k256`, `alloy-rpc-types?/k256`, `alloy-signer-local?/secp256k1`, …).
- `secp256k1` would work but pulls a C library; `k256` is pure Rust, which is the better posture for the guest.

And since `anchor.rs:10` imports `alloy_consensus::` directly against protocol's own direct `alloy-consensus` 1.8.3 dep, this targets exactly the crate instance in use. (Note there are two alloy-consensus majors in the graph — 1.8.3 here, 2.0.4 under the reth stack — so getting the right one mattered.)

davidtaikocha (commented):
## Reviewed — the fix is correct and minimal. One follow-up worth doing before or right after merge.

I verified every claim in the PR body empirically rather than taking it on faith, by building the base and head commits in isolated worktrees.

### Confirmed

**The bug reproduces exactly as described.** On the base commit (`3a99ab18`), `cargo check -p protocol --locked` fails with `E0599: no method named 'recover_signer' found for &EthereumTxEnvelope<TxEip4844Variant>` at `crates/protocol/src/shasta/anchor.rs:270`. On this PR's head, both the `--no-default-features` and the default-feature checks pass.

**The masking mechanism is exactly what the description says.** `cargo tree -e features -i alloy-consensus@1.8.3 --workspace` shows `proposer` is the sole in-repo enabler — it declares `alloy-consensus = { workspace = true, features = ["k256"] }`, and resolver-v2 unification then hands the feature to `protocol` for free in any full-workspace build.

**Scoping to `net` is right.** The `anchor` module is gated `#[cfg(feature = "net")]` in `shasta/mod.rs:3` and holds the only `recover_signer` call in the crate, so the zkVM-guest path genuinely doesn't need `k256`. No lockfile change is required, and `--locked` passing on both legs proves it.

### The one thing I'd change

The guard is skipped on exactly the PRs most likely to trip it — see the inline comment on the new CI step.

### Checked and cleared

**The blast radius was wider than the description claims**, which only strengthens the case for merging. `proposer` is a *dev-dependency* of `driver`, and `cargo check` doesn't build dev-deps — so `cargo check -p driver --locked` was **also** broken on `main` with the same error. I verified it fails on the base and passes here. Anyone building a single crate locally, not just downstream raiko2, was hitting this.

**Making the direct `k256` dep optional isn't worth it.** It's only used by the `net`-gated `signer.rs`, so it looks like a hygiene win — but `cargo tree -p protocol --no-default-features -i k256` shows `k256 v0.13.4` already lands in the guest graph transitively via `alethia-reth-*` → `alloy-consensus 2.0.4`. Making protocol's edge optional would remove nothing from guest builds.

**The implicit optional-dep features are harmless.** `bindings`, `tokio`, `alloy-provider` et al. are public features (no `dep:` prefix), but the consuming code is gated on `net`, not on the individual deps. I checked four such `--no-default-features --features X` combinations; all compile. Pre-existing and benign.

<sub>Reviewed with Claude Code.</sub>

.github/workflows/taiko-client-rs--test.yml
```diff
@@ -236,6 +236,10 @@ jobs:
         with:
           just-version: 1.5.0
 
+      - name: Check protocol feature isolation
+        working-directory: packages/taiko-client-rs
+        run: just check-features
```
davidtaikocha, line 241: Confirmed this closes the gap. The `test` job's `if:` (line 111) omits the `dependabot` clause that `lint` carries (line 39), and — importantly — there's **no `needs:` in this workflow**, so `test` doesn't inherit `lint`'s skip. Dependabot PRs will now run this check.

Prerequisites are all satisfied at this point in the job: `Install Rust toolchain` (201), `Setup Rust cache` (207), `Setup just` (234). And running before `Checkout protocol for testing` (245) is safe — `bindings` has no `build.rs`, so nothing here needs Foundry artifacts. Verified by building in a worktree with no protocol checkout.

Only cost: this is now behind the heavy `test` setup rather than in the quick `lint` job, so a feature-isolation break surfaces later than it used to. Fine as-is; if it annoys you, running it in both jobs is cheap.

packages/taiko-client-rs/justfile
```diff
@@ -14,6 +14,10 @@ fmt-check:
   rustup toolchain install {{toolchain}} && \
   cargo +{{toolchain}} fmt --check
 
+check-features:
+  cargo check -p protocol --no-default-features --locked && \
+  cargo check -p protocol --locked
```
davidtaikocha, line 19: This is the right home for it — `just check-features` now reproduces the CI check locally, which was the parity gap.

I mutation-tested the recipe to make sure it isn't decorative: deleting `"alloy-consensus/k256"` from `protocol/Cargo.toml` makes this fail with the original `E0599: no method named 'recover_signer'` at `anchor.rs:270`, and `just` exits **101**, so the job fails. Fail-fast also works — the `&& \` makes both checks one shell command, so a failure in the `--no-default-features` leg short-circuits before the second.

Minor, non-blocking: `protocol` is hardcoded. It's genuinely the only crate in the workspace with a `no-default-features` guest path today, so the recipe is complete — but it won't self-extend if another crate grows one.

davidtaikocha (commented):
## Re-reviewed `89b936d` — both findings addressed correctly. LGTM.

I re-verified the fixes rather than eyeballing them, including a mutation test to confirm the guard is live and not decorative.

### The dependabot gap is closed

- The step now lives in the `test` job, whose condition (line 111) has no `dependabot` exclusion — only `draft` and `release-please`.
- I checked for the trap that would have silently defeated this: **there is no `needs:` anywhere in the workflow.** If `test` had declared `needs: lint`, then dependabot skipping `lint` would have skipped `test` along with it, and the move would have accomplished nothing. The jobs are independent, so the guard really does run.
- `test` has no `strategy`/`matrix`, so the check runs exactly once per PR, not N times.

### Placement is sound

The step at line 239 sits after `Install Rust toolchain` (201), `Setup Rust cache` (207), and `Setup just` (234) — all three prerequisites are in place. It runs *before* `Checkout protocol for testing` (245), which is fine: the `bindings` crate has no `build.rs` and needs no Foundry artifacts, so `cargo check -p protocol --locked` doesn't depend on that checkout. Confirmed by building it in a clean worktree with no protocol checkout present.

### The guard actually catches the regression — verified by mutation

Running `just check-features` at this head passes. I then deleted the `"alloy-consensus/k256"` line from `protocol/Cargo.toml` in a scratch worktree and re-ran it:

```
error[E0599]: no method named `recover_signer` found for reference
              `&EthereumTxEnvelope<TxEip4844Variant>` in the current scope
   --> crates/protocol/src/shasta/anchor.rs:270

just check-features exit code = 101
```

The first (`--no-default-features`) check passes, the second fails, and `just` propagates a non-zero exit — so CI fails the job. The guard would have caught the original bug.

### Nothing blocking

Two observations, neither worth holding the PR for:

1. **Feedback is now slower.** The guard moved from the fast `lint` job to the 50-minute `test` job, behind the mold/Foundry/pnpm/Node setup. That's the correct trade for dependabot coverage, but if the extra latency bites, the step is cheap enough to run in *both* jobs.
2. **`check-features` hardcodes `protocol`.** That's still the only crate in the workspace with a `no-default-features` guest path, so it's complete today. If another crate grows one, the recipe won't notice.

Nice turnaround. The `justfile` recipe is the right home for this — `just check-features` reproduces the CI check locally, which is what was missing.

<sub>Reviewed with Claude Code.</sub>

davidtaikocha (approved):

PR #21935, opened 2026-07-07
feat(protocol): Proposal0019 — Unzen hardfork DAO bundle

ggonzalez94:
## What

Adds the DAO bundle for the Unzen hardfork. It executes 22 L1 actions and no L2 actions:

- Rotate the trusted RISC0 image IDs and SP1 program vkeys.
- Rotate the trusted SGX MRENCLAVEs on the reused Proposal0017 attesters.
- Delete instance ID `0` from both reused SGX verifiers.
- Upgrade the Inbox with forced inclusions re-enabled and `ZkRequiredVerifier` baked in.
- Call `Inbox.init3()` to void the stale forced-inclusion entry.

## SGX sequencing

The MRENCLAVE rotation and instance deletions execute atomically with the verifier change. Until
`admin.taiko.eth` registers the replacement SGX instances, proving continues through the
`RISC0 + SP1` combination.

## Remaining release inputs

The new RISC0/SP1 IDs and SGX MRENCLAVEs remain guarded zero-value placeholders until the Unzen
raiko release is cut. Proposal action generation and the mainnet dry-run intentionally revert until
those values are populated.

## Verification

- `FOUNDRY_PROFILE=layer1 forge build script/layer1/proposals/Proposal0019.s.sol`
- `FOUNDRY_PROFILE=layer1 forge test --match-path test/layer1/verifiers/SgxVerifier.t.sol --match-test 'test_deleteInstances*' -vv`
- `forge fmt --check script/layer1/proposals/Proposal0019.s.sol`

PR #21945, opened 2026-07-10
feat(protocol): reuse the Proposal0017 SGX verifiers in the Unzen bundle

davidtaikocha:
## Why

`DeployUnzenContracts` deployed two fresh `SecureSgxVerifier`s wired to an upstream Automata DCAP attestation entrypoint, which `DeployAutomataDcapAttestation` had to deploy first under `FOUNDRY_PROFILE=layer1o`. That entrypoint reads Intel collateral from Automata's on-chain PCCS. **Automata does not maintain a provisioned PCCS router on Ethereum mainnet** — the current router holds no TCB-evaluation data, `getStandardTcbEvaluationDataNumber` reverts, and the deployment cannot succeed. The Unzen bundle was blocked on a third-party dependency.

Unzen's security objective — at least one ZK proof per proven batch — does not depend on that entrypoint. It is delivered by `ZkRequiredVerifier`, which structurally forbids the `SGX_GETH + SGX_RETH` pair that finalized the June 2026 forged proofs.

## What

Wire `ZkRequiredVerifier` to the SGX verifiers #21833 already deployed, rather than deploying new ones.

- `DeployUnzenContracts.s.sol` deploys two contracts instead of four: `ZkRequiredVerifier` and the `MainnetInbox` implementation that bakes it in. No `DCAP_ATTESTATION` env var, no prerequisite script, no `layer1o` step.
- `Proposal0019.s.sol` drops the two `acceptOwnership` actions and the `SGXGETH_VERIFIER_NEW` / `SGXRETH_VERIFIER_NEW` / `DCAP_ATTESTATION` constants: **16 actions → 14**.
- `Proposal0019.md` updated to match.

`ZkRequiredVerifier` receives the same four sub-verifiers the live `MainnetVerifier` (`0x71808449…`) already holds, so the behavioral deltas are exactly the intended ones: `SGX_GETH + SGX_RETH` stops satisfying the verifier, and `RISC0 + SP1` becomes sufficient without any TEE leg.

## Why this is safe (all read from mainnet)

- Both SGX verifiers' `owner()` is the DAO controller `0x75Ba7640…`; `pendingOwner()` is zero. No handover to accept.
- The MRENCLAVE/MRSIGNER registries live on the DAO-owned attester proxies `0x0ffa4A62…` / `0x8d7C9549…`, and the Unzen raiko release rebuilds the ZK guests but not the SGX enclaves — the measurements Proposal0017 trusted are still trusted and still correct.
- `nextInstanceId()` is `1` on each verifier: one registered raiko instance apiece, `validSince` 2026-06-29, valid to roughly 2027-06-29. SGX proving continues across the fork with no registration step, so there is no RISC0+SP1-only window.
- The deployed verifiers expose `verifyProof(uint256,bytes32,bytes)` (`0x14bcf3dd`), exactly what `ComposeVerifier` invokes.
- A fresh hardened `SecureSgxVerifier` cannot be pointed at the existing attester proxies: `SgxVerifier.registerInstance` calls `verifyAndAttestOnChain(bytes)` (`0x38d8480a`), which those implementations do not export — they expose `verifyParsedQuote(...)` (`0x089a168f`).

Migrating SGX to the hardened verifier + upstream Automata entrypoint remains desirable and is deferred to its own PR, once mainnet PCCS exists.

## Test plan

- `FOUNDRY_PROFILE=layer1 forge build` — green.
- `FOUNDRY_PROFILE=layer1 forge test --match-path test/layer1/verifiers/ComposeVerifier.t.sol` — 13 passed, 0 failed.
- `P=0019 pnpm proposal` reverts on the `MAINNET_INBOX_NEW_IMPL` guard (constants pending deployment), as designed.
- Verified on mainnet that the four addresses passed to `ZkRequiredVerifier` are byte-identical, and in the same roles, to the four sub-verifiers the live `MainnetVerifier` returns — i.e. this is a pure combination-logic change, not a re-wiring.

Stacked on #21935.

### DIFF
--- a/packages/protocol/script/layer1/proposals/Proposal0019.md
+++ b/packages/protocol/script/layer1/proposals/Proposal0019.md
@@ -205,21 +205,23 @@ RISC0/SP1 addresses were read live from the current `MainnetVerifier` (`risc0Ret
 
 ## Deployed Addresses
 
-> **TODO(unzen):** deployment is a single step:
->
-> ```bash
-> FOUNDRY_PROFILE=layer1 forge script DeployUnzenContracts ...
-> ```
->
-> Then fill this table, the two address constants **and the `NEW_RISC0_*` / `NEW_SP1_*` image IDs
-> from the Unzen raiko release** in [`Proposal0019.s.sol`](./Proposal0019.s.sol), and regenerate
-> `Proposal0019.action.md` (`P=0019 pnpm proposal`). The script reverts while any of these
-> constants are zero.
+Deployed in a single step with
+`FOUNDRY_PROFILE=layer1 forge script DeployUnzenContracts ...`:
 
-| Constant                 | Address | Contract                                        |
-| ------------------------ | ------- | ----------------------------------------------- |
-| `ZK_REQUIRED_VERIFIER`   | TBD     | `ZkRequiredVerifier` (`MainnetInbox` immutable) |
-| `MAINNET_INBOX_NEW_IMPL` | TBD     | `MainnetInbox` implementation                   |
+| Constant                 | Address                                      | Contract                                        |
+| ------------------------ | -------------------------------------------- | ----------------------------------------------- |
+| `ZK_REQUIRED_VERIFIER`   | `0x7284aaC05555Ae6559bdAd8B4221eC9584254Eec` | `ZkRequiredVerifier` (`MainnetInbox` immutable) |
+| `MAINNET_INBOX_NEW_IMPL` | `0x5253D4C91e80b880DdB54B78E74082Abe066F6b9` | `MainnetInbox` implementation                   |
+
+Both are verified on-chain: `ZK_REQUIRED_VERIFIER` returns the four production sub-verifiers (and
+`address(0)` for `tdxGethVerifier()` / `opVerifier()`), matching the outgoing `MainnetVerifier`
+getter-for-getter, and `MAINNET_INBOX_NEW_IMPL.getConfig().proofVerifier` equals
+`ZK_REQUIRED_VERIFIER`.
+
+> **TODO(unzen):** the `NEW_RISC0_*` / `NEW_SP1_*` image IDs in
+> [`Proposal0019.s.sol`](./Proposal0019.s.sol) are still `bytes32(0)`, pending the Unzen raiko
+> release. Fill them, then regenerate `Proposal0019.action.md` (`P=0019 pnpm proposal`). The
+> script reverts with `ZkImageIdNotSet()` while any of them is zero.
 
 ## Client Rollout Prerequisite
 
--- a/packages/protocol/script/layer1/proposals/Proposal0019.md
+++ b/packages/protocol/script/layer1/proposals/Proposal0019.md
@@ -10,10 +10,11 @@ proving images to the raiko release shipping with Unzen.
 The SGX verifiers are the ones Proposal0017 deployed, reused unchanged. This proposal does not
 deploy new SGX verifier contracts, register SGX instances, or transfer ownership. It does rotate
 the trusted SGX MRENCLAVE allowlist on the existing attester proxies to the raiko release shipping
-with Unzen; the trusted MRSIGNER remains unchanged (see
+with Unzen, and delete the instances registered under the retiring measurements; the trusted
+MRSIGNER remains unchanged (see
 [SGX verifiers reused](#3-sgx-verifiers-reused-from-proposal0017)).
 
-It executes **20 L1 actions** and **no L2 actions**:
+It executes **22 L1 actions** and **no L2 actions**:
 
 - **Actions 1–4**: rotate the trusted RISC0 image IDs — untrust the two live raiko2 v0.5.1
   IDs, trust the two new ones.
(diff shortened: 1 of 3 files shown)


johntaiko (approved):

packages/protocol/script/layer1/proposals/Proposal0019.md
```diff
@@ -71,76 +67,56 @@ The new `ZkRequiredVerifier` accepts exactly two sub-proofs in one of three comb
...
+No new SGX contract is deployed, so the bundle deploys exactly two contracts:
+`ZkRequiredVerifier`, and the `MainnetInbox` implementation that bakes it in as an immutable.
+
+This proposal therefore contains **no SGX action at all**. Three properties, each read from
+mainnet, are what make that safe:
+
+1. **Ownership is already where it belongs.** `owner()` on both verifiers, and on both attester
+   proxies, returns the DAO controller `0x75Ba76403b13b26AD1beC70D6eE937314eeaCD0a`;
+   `pendingOwner()` is zero on both. There is no handover to accept.
+2. **The trust registries are already configured for the Unzen raiko release.** MRENCLAVE and
```
YoGhurt111: Based on my testing experience on devnet, we changed endpoint from raiko1 to raiko2 from Unzen fork, so there should be new registries, correct me if I'm wrong. @smtmfft

packages/protocol/script/layer1/proposals/Proposal0019.md
```diff
@@ -191,73 +167,61 @@ contract (no refund path exists on-chain; the `ForcedInclusion` struct does not
...
+3. `RISC0_RETH_VERIFIER.setImageIdTrusted(NEW_RISC0_PROPOSAL_IMAGE_ID, true)`
+4. `RISC0_RETH_VERIFIER.setImageIdTrusted(NEW_RISC0_AGGREGATION_IMAGE_ID, true)`
+5. `SP1_RETH_VERIFIER.setProgramTrusted(OLD_SP1_PROPOSAL_PROGRAM_VKEY_BN256, false)`
+6. `SP1_RETH_VERIFIER.setProgramTrusted(OLD_SP1_PROPOSAL_PROGRAM_VKEY_HASH_BYTES, false)`
+7. `SP1_RETH_VERIFIER.setProgramTrusted(OLD_SP1_AGGREGATION_PROGRAM_VKEY_BN256, false)`
+8. `SP1_RETH_VERIFIER.setProgramTrusted(OLD_SP1_AGGREGATION_PROGRAM_VKEY_HASH_BYTES, false)`
+9. `SP1_RETH_VERIFIER.setProgramTrusted(NEW_SP1_PROPOSAL_PROGRAM_VKEY_BN256, true)`
+10. `SP1_RETH_VERIFIER.setProgramTrusted(NEW_SP1_PROPOSAL_PROGRAM_VKEY_HASH_BYTES, true)`
+11. `SP1_RETH_VERIFIER.setProgramTrusted(NEW_SP1_AGGREGATION_PROGRAM_VKEY_BN256, true)`
+12. `SP1_RETH_VERIFIER.setProgramTrusted(NEW_SP1_AGGREGATION_PROGRAM_VKEY_HASH_BYTES, true)`
```
YoGhurt111, line 213: And seems we also need to register the new SGX image(both geth and reth) for raiko2? @smtmfft

packages/protocol/script/layer1/proposals/Proposal0019.md
davidtaikocha: addressed in https://github.com/taikoxyz/taiko-mono/pull/21946

packages/protocol/script/layer1/proposals/Proposal0019.md
davidtaikocha: addressed in https://github.com/taikoxyz/taiko-mono/pull/21946

YoGhurt111 (approved):

packages/protocol/script/layer1/proposals/Proposal0019.md
davidtaikocha: **Correction to my reply above:** `deleteInstances` has been reverted out of this PR (5387dac1c). The bundle is **20 actions**, not 22 — the MRENCLAVE rotation (actions 13–18) is unchanged, but retiring the pre-Unzen SGX instances moves to a follow-up proposal.

The underlying fact still holds and is now written into Section 3 rather than acted on: untrusting an MRENCLAVE does not revoke an instance already registered under it, because the deployed `Instance` struct is `(address addr, uint64 validSince)` with no MRENCLAVE to re-check. So the raiko1 enclave keys stay accepted SGX signers until roughly 2027-06-29 unless a later proposal calls `deleteInstances([0])` on each verifier. That's now item 10 of the post-execution checklist, to be sequenced *after* the raiko2 registrations so SGX prover diversity is never lost.

Worth being explicit that this is bounded rather than harmless: `ZkRequiredVerifier` requires at least one ZK proof per batch, so a rogue SGX signature cannot finalize anything on its own.

packages/protocol/script/layer1/proposals/Proposal0019.md
davidtaikocha: **Correction to my reply above:** `deleteInstances` has been reverted out of this PR (5387dac1c). The bundle is **20 actions**, not 22 — the MRENCLAVE rotation (actions 13–18) is unchanged, but retiring the pre-Unzen SGX instances moves to a follow-up proposal.

The underlying fact still holds and is now written into Section 3 rather than acted on: untrusting an MRENCLAVE does not revoke an instance already registered under it, because the deployed `Instance` struct is `(address addr, uint64 validSince)` with no MRENCLAVE to re-check. So the raiko1 enclave keys stay accepted SGX signers until roughly 2027-06-29 unless a later proposal calls `deleteInstances([0])` on each verifier. That's now item 10 of the post-execution checklist, to be sequenced *after* the raiko2 registrations so SGX prover diversity is never lost.

Worth being explicit that this is bounded rather than harmless: `ZkRequiredVerifier` requires at least one ZK proof per batch, so a rogue SGX signature cannot finalize anything on its own.

ggonzalez94 (approved):

PR #21950, opened 2026-07-11
fix(taiko-client-rs): recover from scanner lag and reorged-out proposal logs in event sync

davidtaikocha:
## Summary

Two event-sync recovery fixes in the driver, both verified against `event-scanner` 1.1.1 semantics.

**1. Treat `ScannerError::Lagged` as a continuity break.**
`Lagged` is the only non-terminal scanner stream error: the subscription keeps running, but the lagged broadcast ranges were dropped by the scanner and are never re-fetched. The previous `log + continue` therefore silently skipped the proposals in those ranges, freezing derivation at the gap until a restart. The stream loop now breaks on `Lagged`, routing into the existing reconnect path, which closes preconfirmation ingress, rewinds the scanner start to `min(last_seen - 1, finalized)`, replays the gap, and re-opens ingress only after a fresh confirmed-sync probe.

**2. Orphaned-proposal-log classification is proven by canonical mismatch only.**
By-hash resolvability proves nothing about canonicality in either direction: nodes keep serving reorged-out blocks by hash (so a permanently failing log from a reorged-out block would retry forever and freeze event sync), while lagging or mixed RPC backends can transiently miss canonical blocks by hash (so treating "hash gone + head passed the height" as orphaned could permanently skip a valid proposal). The check now resolves the block height (from the stored block, else the log) and compares the canonical block hash at that height: only a proven mismatch classifies the log as permanently orphaned; unresolved canonical data — including a log with no resolvable height at all — stays retryable.

## Review round 1

- Dropped the remaining weak-evidence branch (`hash missing + head >= height` ⇒ orphaned) from the orphan recheck in favor of the canonical-mismatch proof above, removing the `get_block_number` probe entirely.
- Enforced the preconfirmation ingress gate inside the queue consumer, re-checked under the router lock: jobs already queued when the event loop closes ingress for a replay window — or racing past the submit-side check across its RPC awaits — are answered with `PreconfIngressNotReady` instead of being injected against a stale confirmed boundary. The whitelist importer's cached-drain path treats the error as abort-and-retry (envelope stays cached; the sync-ready poll retries after reopen), and the REST path already maps this error for callers.
- A full lag→replay→reopen e2e is not reachable through public APIs (the scanner is constructed internally and a real `Lagged` requires overflowing a 50k-message broadcast buffer); the gate behavior is covered by direct ingress-loop tests instead.

## Review round 2

- **Event loop is the sole ingress-gate owner.** The freshly spawned consumer previously stored the gate open unconditionally, so a reconnect close landing between spawn and the task's first poll was silently reverted, leaving ingress open for the whole replay window. The open now happens in the event loop right after the spawn (program-ordered with any later close), and the close takes the router lock: an in-flight injection that observed an open gate completes before the close lands, and every job dequeued afterwards observes the closed gate before replay derivation (serialized on the same lock) begins.
- **`scanner_live` is per scanner generation.** It was only reset before ingress had ever spawned, so a replacement scanner inherited the previous generation's live state and a readiness probe could reopen ingress before a fresh `SwitchingToLive`. It is now declared inside the reconnect loop, so every generation starts non-live and reopening always requires fresh liveness plus a passed confirmed-sync probe (WLP-INV-002).
- **No-height orphan fallback is retryable.** Missing by-hash data plus a missing log height is not a proven canonical mismatch; the fallback now returns retryable, matching the mismatch-only contract.

## Impacted invariants

- `WLP-INV-001`: lag recovery reuses the finalized-bounded reconnect anchor for replay.
- `WLP-INV-002`: scanner lag closes preconf ingress and re-opens it only after the scanner-live + confirmed-sync gate; gate transitions are event-loop-owned and serialized with the router lock, and scanner liveness is per generation.
- `WLP-INV-009`: event-driven reorg handling classifies orphaned logs by canonical-chain membership rather than block-store lookups.

## Testing

- Unit tests: canonical-hash mismatch → orphaned (stored and missing by-hash variants); same-hash canonical → retryable (including the missing-by-hash lagging-backend case); unresolved canonical height → retryable; missing height → retryable; `Lagged` classified as reconnect-worthy while terminal scanner errors are not; ingress loop rejects queued jobs while the gate is closed (including a consumer spawned with the gate never opened) and processes them once the event loop opens it.
- `just fmt-check` and `just clippy` (deny warnings, deny missing docs) clean.
- Full `just test` suite: 298/298 passed (unit + dockerized e2e).


packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -496,29 +522,28 @@ impl EventSyncer {
...
             .l1_provider
-            .get_block_number()
+            .get_block_by_number(BlockNumberOrTag::Number(block_number))
             .await
             .map_err(|err| SyncError::Rpc(RpcClientError::Provider(err.to_string())))?;
 
-        Ok(chain_head >= log_block_number)
+        // Only a proven canonical-hash mismatch classifies the log as permanently orphaned;
+        // unresolved canonical data stays retryable.
+        Ok(canonical_block.is_some_and(|canonical| canonical.header.hash != block_hash))
```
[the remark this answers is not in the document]
davidtaikocha: Not taking this one — the scenario is real but transient, it isn't introduced by this PR, and the proposed remediations don't fit the driver's recovery model. Reasoning:

**1. The skip-and-continue flow is pre-existing.** This PR only changes the evidence standard inside `is_permanently_orphaned_proposal_log` (by-hash resolvability → canonical-hash comparison at the log's height). The retry → orphan-recheck → skip → continue-batch flow is unchanged from `main`. Under the old classifier this exact scenario was strictly worse: a log from a reorged-out block could never be classified orphaned at all (nodes keep serving reorged-out blocks by hash), so the batch wedged forever behind the infinite retry — that's the bug this PR fixes.

**2. The stale-import window is inherent to optimistic import, not specific to the orphan path.** Any successfully derived log can be reorged out immediately *after* import, with no orphan proof anywhere in the picture — the same-batch case here is one instance of that general window. A canonicality pre-check on every remaining log is racy by construction (canonical-at-check ≠ canonical-at-import), adds an L1 RPC round-trip per log on the hot path, and still cannot close the window. The driver is deliberately optimistic-import + reorg-correct, the same model as the Go client.

**3. The correction machinery already covers this case — and the orphan proof's own trigger condition activates it.** For the orphan proof to fire, the reorg must already be visible on the L1 RPC (the canonical hash at the height differs). That is exactly the condition under which the scanner's next batch-boundary check (`reorg_handler.check(previous_batch_end)`, event-scanner 1.1.1 `block_range_scanner/common.rs:261`) detects the same reorg, emits `Notification::ReorgDetected`, and rewinds its own streaming position to the common ancestor (`common.rs:316`, `:340`, `:452`) — re-streaming the canonical range, which re-derives over any transiently imported stale proposal and moves the canonical head via FCU. The event loop additionally repairs the confirmed boundary on that notification (`reset_head_l1_origin_after_reorg`). If the stream dies instead of detecting, the reconnect path rewinds to `min(last_seen − 1, finalized)` and replays the entire reorg-unsafe window. Every path converges on re-derivation, so the import described here lives for roughly one batch boundary.

**4. WLP-INV-009 doesn't demand synchronous canonicality.** It states that event-driven and preconf-driven reorg handling are distinct flows and both required ("event path handles L1 reorg resets and proposal cursor rollback"). A transient import corrected by the event path's rewind is that invariant operating, not a violation of it.

On the remediations: abandoning the remainder of the batch after a proven orphan would be safe (processing is in-order, so every remaining log sits at or above the proven fork height, and the scanner's rewind re-serves the range) — but it only narrows a ~one-batch window by leaning on exactly the rewind machinery that already corrects it, so it's a follow-up hardening candidate at most, not a correctness requirement for this PR.

YoGhurt111: **[P1] Do not treat one cross-backend height mismatch as permanent orphan proof**

`self.rpc.l1_provider` and the event scanner are separate connections created from the same `SubscriptionSource`. With a load-balanced or WS endpoint during a short reorg split, the scanner can already follow canonical block `A` at height `h` while this provider is still pinned to the old fork `B`. If processing `A` fails because `A` is not yet resolvable by hash on `B`, this check reads `B` by number and immediately returns `true`, permanently skipping the canonical log.

This differs from the existing same-batch thread: the scanner remains on `A`, so it observes no reorg and emits no `ReorgDetected` or cursor rewind. When the ordinary provider later catches up, the skipped log is not replayed; the next proposal can then retry forever because the skipped proposal's batch mapping is absent. A single mismatch from an independent backend should remain retryable or force a scanner replay / consistent-view confirmation, rather than producing `SkippedOrphaned`.

packages/taiko-client-rs/crates/driver/src/sync/event.rs
```diff
@@ -1372,6 +1396,14 @@ impl SyncStage for EventSyncer {
                     }
                     Err(err) => {
                         DriverMetrics::event_scanner_errors_total().inc();
+                        if scanner_error_requires_reconnect(&err) {
+                            warn!(
+                                ?err,
+                                "event scanner dropped lagged block ranges; replaying from \
+                                 reconnect overlap"
+                            );
+                            break;
```
YoGhurt111: **[P2] Retry readiness even when no proposal events arrive**

This new `Lagged` reconnect path closes ingress, but reopening is attempted only while handling scanner items. `event-scanner` filters empty log batches and emits `SwitchingToLive` only once per scanner generation, so if the confirmed-sync probe transiently errors or is not ready on that notification, a quiet inbox produces no further item and ingress remains closed indefinitely. This leaves the new lag recovery path unable to complete even after the RPC state becomes healthy. Please retry the confirmed-sync probe on a timer while the current generation is live and the gate is closed, instead of relying on another proposal event.

YoGhurt111 (commented):
Two additional findings from reviewing the current head.

packages/taiko-client-rs/crates/driver/src/sync/event.rs
davidtaikocha: Fixed in d3958f713. The check now trusts a canonical-hash mismatch only at or below the finalized height: `canonical(h)` is immutable and identical across every honest view once finalized, so a trailing or load-balanced backend on the losing fork can no longer produce a false permanent skip, and the no-`ReorgDetected` gap you describe (the scanner stayed on the winning fork the whole time, so nothing would ever re-serve the skipped log) is closed — above finality every mismatch stays retryable.

Trade-off taken: a genuinely reorged-out log is now skipped only once finality passes its height, i.e. the retry loop spins with backoff for up to ~2 epochs before the proven skip lands (versus pre-#21950 spinning forever). Pre-finality chains treat geth's "finalized block not found" as finality-unavailable → retryable.

I went with the finalized guard over forcing a scanner replay because it adds no new control flow through the reconnect path and produces a proof no transient view split can contradict. If the ~2-epoch retry window on true orphans is a concern, wiring the unfinalized-mismatch case into the reconnect/replay path instead is a reasonable follow-up.

packages/taiko-client-rs/crates/driver/src/sync/event.rs
davidtaikocha: Fixed in d3958f713. All three premises check out against event-scanner 1.1.1 (empty log batches are filtered before reaching the stream, `SwitchingToLive` fires once per generation, and the probe only ran per received item), so reopening no longer depends on the next proposal event: the stream read is now a `tokio::select!` with a probe-retry interval arm (6s, `MissedTickBehavior::Skip`) guarded by `should_probe_confirmed_sync`, and the probe/spawn/open block moved into `try_open_preconf_ingress`, called from both the item path and the timer arm.

Both arms run in the same event-loop task, so the round-2 gate-ownership properties are unchanged: the gate keeps a single writer and opens stay program-ordered with reconnect closes. This also removes the same item-dependency from the *initial* open after historical sync, which had the identical shape.

YoGhurt111 (approved):

PR #21952, opened 2026-07-11
fix(taiko-client-rs): stop retrying fatal engine verdicts in event sync

davidtaikocha:
## Summary

The driver's event-sync retry loop (`process_log_batch`) retried **every** derivation failure with an unbounded exponential backoff — including deterministic engine verdicts like `INVALID`, which no retry can change. A content-level failure therefore turned into a silent no-progress stall (retrying forever at 12s max intervals) instead of surfacing to the operator, while the same verdict in beacon sync is treated as fatal.

The root enabler was error plumbing: raw provider errors were collapsed into `RpcClientError::Provider(err.to_string())` at ~15 call sites, contract errors were stringified (dropping revert data), and transient-vs-fatal conditions were detected by substring matching on stringified error chains — typed classification was impossible.

## Changes

**Stop retrying deterministic verdicts (behavior change)**
- `process_log_batch` now uses `RetryIf` with a typed `is_fatal_proposal_processing_error` predicate: an engine `INVALID` verdict (via derivation or preconf injection) aborts the retry loop and surfaces the error; transient failures (RPC transport, engine syncing) retry as before. Orphaned-log skipping is unchanged.

**Keep errors typed across the stack**
- `rpc`: `RpcClientError::Contract` now preserves the structured `alloy::contract::Error` (revert data survives) instead of `.to_string()`.
- `rpc`: new typed `RpcClientError::is_finalized_block_unavailable()` inspects the JSON-RPC error payload; replaces two `err.to_string().contains("finalized block not found")` substring matches (beacon + event sync). The geth-specific wording now lives in exactly one place.
- `driver`: `From<TransportError>` / `From<alloy::contract::Error>` conversions for `SyncError`, `DriverError`, and `EngineSubmissionError` — the `Provider(err.to_string())` sites become plain `?` with full error structure preserved.
- `driver`: the `DriverError` → `SyncError` conversion no longer downgrades typed errors into `anyhow`; non-sync/RPC variants stay typed behind a new `SyncError::Driver(Box<DriverError>)`.
- `driver`: removed the now-unused `EngineSubmissionError::Provider(String)` variant; the engine read-back path returns the typed RPC error.

## Not in scope

Draft #21949 (strict engine-status checks on newPayload / promotion FCU / read-back) is complementary and untouched: #21949 produces more typed `InvalidBlock` verdicts at the engine boundary; this PR makes the retry loop respond to them correctly.

## Testing

- `just fmt` + `just clippy` (`-D warnings -D missing_docs`) clean.
- `TEST_CRATE=driver just test`: **90/90 passed**, including the Dockerized e2e suites (`proposer_driver_e2e`, `shasta_event_sync`).
- New unit tests:
  - `process_log_batch_aborts_without_retry_on_engine_invalid_block` — asserts exactly **one** produce attempt on engine INVALID (previously retried until timeout).
  - `fatal_proposal_processing_error_*` — predicate classification (INVALID fatal; syncing/RPC/other transient).
  - `sync_error_from_driver_error_preserves_structure` — no more anyhow downgrade.
  - `finalized_block_unavailable_*` (rpc) — typed detection, including through contract errors, plus negative cases.


davidtaikocha:
Addressed review finding on the `Err(recheck_err)` arm in 249edc971: when derivation returns a fatal verdict but the orphaned-log recheck fails transiently, the closure now surfaces the retryable recheck error instead of the original fatal one — so `RetryIf` keeps going and a fatal verdict can only abort after `Ok(false)` confirms the log is still canonical. Added regression test `process_log_batch_keeps_retrying_invalid_block_until_orphan_recheck_resolves` (INVALID → recheck RPC failure → orphan confirmed → skipped, exactly 2 attempts). Re-ran `just fmt`, `just clippy`, and `TEST_CRATE=driver just test` — all green.

davidtaikocha:
Addressed the canonicality finding in 4f07f89c8: the orphan recheck now resolves the emitting block **by number** and compares the canonical hash at that height against the log's block hash — a match confirms canonicality, a mismatch confirms orphaning (even when the provider still serves the reorged-out block via `eth_getBlockByHash` from block storage), and a missing height keeps the existing head-guard so provider lag stays retryable. Logs without a block number fall back to the hash lookup. Added regression test `orphaned_proposal_log_detected_when_provider_retains_side_chain_block`. Verified: `just fmt` + `just clippy` clean, `TEST_CRATE=driver just test` 92/92 passed.

PR #21966, opened 2026-07-18
chore(taiko-client-rs): improve whitelist preconfirmation driver

davidtaikocha:
## Summary

A review of `crates/whitelist-preconfirmation-driver` against the Go client's preconf block server (`driver/preconf_blocks` + the pinned `taikoxyz/optimism` p2p fork) found five behavioral divergences. Each one can make a Rust node accept what every Go node rejects (or vice versa), splitting the preconf head on a mixed fleet, or silently break the API contract Catalyst-side tooling expects. This PR fixes all five; wire-format parity (topics, message-id, SSZ envelope layout, signing hash, validation ordering) was checked and needed no changes. An independent verification round against the Go/geth/echo-jwt sources then confirmed each fix and produced the tightenings in "Review follow-ups" below.

### 1. Compressed tx-list cap: 786,432 → 780,264 (`codec.rs`)

The cap was `131_072 * 6`, but Go checks `eth.MaxBlobDataSize * eth.MaxBlobsPerBlobTx` = `((4*31+3)*1024 − 4) * 6` = 780,264 (`server.go` `ValidateExecutionPayload`). A payload whose compressed list lands in the 6,168-byte gap was accepted and imported here while every Go node dropped it. The constant is now derived the same way Go derives it, with a test pinning the value.

### 2. Drop the wire-signature → embedded-signature substitution (`importer/validation.rs`, `importer/ingress.rs`)

`normalize_unsafe_payload_envelope` copied the wire signature (which signs the **SSZ envelope bytes**) into the envelope's embedded-signature slot (which peers verify against the **block hash** — a different domain) whenever the slot was empty. The substituted value was persisted to the L1 origin, passed the non-zero "can attest" check, and got served in response envelopes that **every** receiving peer rejects (the recovered address is never an operator). The Go client keeps the slot empty and declines to serve such blocks. Now: unsigned envelopes stay unsigned, `serve_envelope_by_hash` skips them in the recent cache (falling through to the L2 rebuild, which already declines on a zero L1-origin signature), and the L1 origin records a zero signature — all matching Go.

### 3. Reject recovery ids > 1 (`codec.rs`)

`recover_signer` parsed via alloy `Signature::from_raw_array`, which accepts and normalizes v ∈ {0, 1, 27, 28, 35+} — so anyone re-publishing a valid envelope with v=27 could make Rust and Go nodes disagree on acceptance at will. geth's `crypto.SigToPub` (what Go peers use) rejects v ≥ 4 outright and, for v ∈ {2, 3}, attempts a recovery that can only succeed when r < p − n (≈2^129 of a ~2^256 space); crafting such a signature that also recovers to an *allowlisted* address is cryptographically unreachable, so Go nodes reject every feasible v > 1 message too. Checking the v byte before parsing is therefore decision-equivalent to geth — not mechanism-identical, which an in-code comment now spells out — while closing the real 27/28/35+ divergence. (High-s signatures remain accepted, which matches Go's `Ecrecover` exactly.)

### 4. Validate JWT `exp`/`nbf` when present (`api/server/auth.rs`)

`validate_exp` was `false`, so an expired token stayed valid forever. The Go server (echo-jwt → golang-jwt v5 defaults) validates `exp`/`nbf` whenever present, with claims otherwise optional and zero leeway. jsonwebtoken's built-in validator cannot reproduce those semantics exactly — it treats present-but-malformed claims as absent, keeps `exp == now` valid, and rejects any token carrying an `aud` claim by default — so the library layer now verifies only the signature, and a manual `validate_temporal_claims` replicates golang-jwt v5: present claims must be numeric dates, truncated to whole seconds, expired once `now >= exp`, not yet valid while `now < nbf`, no leeway, and a numeric date of `0` counts as absent (golang-jwt's `parseNumericDate` returns nil for a zero float64). Tokens without claims continue to work.

### 5. Push `/ws` EOS notifications when a gossiped EOS block materializes, not on local build (`cache.rs`, `importer/*`, `api/service/*`)

The only notification site was the **API build path** — i.e. the node that built the EOS block notified its own subscriber, who already has the block from the 200 response. The Go client's only push site is the opposite: an EOS block received from the payload topic (`server.go` `OnUnsafeL2Payload`), pushed after `TryImportingPayload` succeeds — the signal the *incoming* sequencer actually needs.

Now: the EOS flag is trusted **only from the payload topic**, whose wire signature covers the whole SSZ envelope including the flag bytes. On the response topic the embedded signature covers only the block hash, so the flag is unauthenticated there — and Go's `OnUnsafeL2Response` has no EOS handling at all, so response-topic envelopes now contribute nothing to EOS state. A bounded `PayloadEosTracker` captures the flag at admission (the pending cache overwrites same-hash entries, so a later response envelope could otherwise flip it in either direction), and the `/ws` notification fires once the block **materializes** (`Inserted` with the operator-signed hash), carrying the wall-clock epoch at push time. The build path only records the epoch marker.

Two deliberate deviations from Go inside this fix, both local-API-only (no gossip effect):

- **Deferred imports still notify.** An EOS payload-topic block imported later (missing parent, or received while the node catches up) notifies on materialization. Go cannot: `envelopeFromMessage` drops the `EndOfSequencing` flag when caching and its push site is gated on an immediate fresh insert, so a Go sidecar silently misses those handovers.
- **Hash-mismatched insertions withhold the notification** (Go's push site never compares the envelope hash to its inserted block): the operator-signed EOS block never materialized, so notifying would announce a block peers cannot fetch.

## Review follow-ups (second round)

- **EOS markers are keyed by the wall-clock epoch at record time** at both record sites (payload-topic ingress and the build path), matching Go's `CurrentEpoch()`. The requesting operator looks the marker up with *its* wall-clock epoch at handover, so the previous block-timestamp keying missed an EOS block delivered just across an epoch boundary. Since `current_epoch()` is infallible, this also removes the build path's only post-insert failure mode (epoch derivation erroring after the block was already inserted and gossiped) — Go's build path cannot fail there either.
- **Rebuilt request-topic responses derive their EOS flag from the marker map** (`SharedPreconfState::is_end_of_sequencing_hash` — the same value scan Go runs over `sequencingEndedForEpochCache`) instead of the pending envelope cache, which a response-topic envelope could taint.
- **Envelopes whose import was rejected as invalid are purged from the recent serve cache**, matching the existing hash-mismatch purge; Go only ever serves imported blocks.
- **JWT extraction mirrors echo-jwt v4**: every `Authorization` value is considered (up to echo-jwt's 20-entry extractor cap; ours counts candidates where echo-jwt counts header indices, divergent only for requests with more than 20 `Authorization` values), the `Bearer ` prefix matches case-insensitively with a non-empty remainder, and any validating candidate authorizes. Verified against the pinned echo-jwt v4.4.0 that both missing and invalid tokens map to 401 there (`ErrJWTMissing` is `StatusUnauthorized`), so the middleware's existing single-401 behavior already matched and stays.
- **The ingress source is a two-variant enum** (`Payload`/`Response`), so a future ingress path cannot silently skip the EOS authentication gate, and the notify decision is an extracted, directly tested function (`eos_notification_due`).

## Not in scope (deliberate divergences left as-is)

- `/status` field set (`lookahead`/`totalCached` absent) and `canShutdown` semantics — consequences of the Catalyst-owns-handover architecture, worth documenting rather than changing.
- The Rust-only 8 MiB decompressed tx-list bound (zlib-bomb protection). Accurate risk statement: Go's `utils.Decompress` is unbounded, and a gas-legal block heavy in zero bytes (~11 MB decompressed at 4 gas per zero byte) can compress far under the 780,264 cap — Go imports such a block while Rust rejects it, stalling the Rust side of a mixed fleet on that head until L1 confirmation. Only an operator-signed payload can trigger this, so the bound is kept and the divergence is tracked as operator-signed-only rather than claimed "stricter is safer".

## Testing

- `cargo nextest run -p whitelist-preconfirmation-driver`: 127/127 pass.
- New tests across both rounds: cap value pinned to 780,264; recovery-id 27 rejected / 0-1 accepted (real `FixedKSigner` signature); expired-`exp` and future-`nbf` tokens rejected while claimless tokens pass, `exp: 0` treated as absent, fractional claims truncated to whole seconds, malformed present claims rejected; case-folded `bearer` schemes accepted, every `Authorization` value tried, candidate-free headers classed as missing; `SharedPreconfState` EOS notify/subscribe round-trip and marker-hash lookup; `eos_notification_due` composite decisions (notifies exactly once on a matching insert; never for unmarked hashes, mismatched insertions, `AlreadyMaterialized`, or `Stale`); `PayloadEosTracker` mark/take/discard/eviction semantics.
- `cargo clippy -p whitelist-preconfirmation-driver --all-targets`: no new warnings (the pre-existing `items_after_test_module` warning in `network/runtime.rs` is untouched by this PR).

### DIFF
--- a/packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/api/server/auth.rs
+++ b/packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/api/server/auth.rs
@@ -22,13 +22,17 @@ impl JwtAuth {
     /// Build a validator from a shared secret.
     pub(super) fn new(secret: &[u8]) -> Self {
         let mut validation = Validation::new(Algorithm::HS256);
-        // Claims stay optional (no claim is required to be present), but `exp`
-        // and `nbf` are validated whenever present with zero leeway — matching
-        // the Go client's echo-jwt defaults, which refuse expired tokens.
+        // The library layer only verifies the signature; temporal claims are
+        // checked by `validate_temporal_claims` with golang-jwt v5 semantics
+        // (present-but-malformed rejected, `exp == now` already expired),
+        // because jsonwebtoken treats malformed claims as absent and keeps
+        // `exp == now` valid. `aud` is not validated: the Go client's echo-jwt
+        // defaults only check an audience when one is configured, while
+        // jsonwebtoken would reject every token carrying an `aud` claim.
         validation.required_spec_claims.clear();
-        validation.validate_exp = true;
-        validation.validate_nbf = true;
-        validation.leeway = 0;
+        validation.validate_exp = false;
+        validation.validate_nbf = false;
+        validation.validate_aud = false;
         Self { decoding_key: DecodingKey::from_secret(secret), validation }
     }
 
@@ -45,9 +49,51 @@ impl JwtAuth {
             .strip_prefix("Bearer ")
             .ok_or_else(|| "authorization header must use bearer token".to_string())?;
 
-        decode::<serde_json::Value>(token, &self.decoding_key, &self.validation)
+        let token_data = decode::<serde_json::Value>(token, &self.decoding_key, &self.validation)
             .map_err(|err| format!("invalid bearer token: {err}"))?;
-        Ok(())
+        validate_temporal_claims(&token_data.claims)
+    }
+}
+
+/// Validate `exp`/`nbf` with the Go client's golang-jwt v5 semantics: claims are
+/// optional, but a present claim must be a numeric date, a token is expired once
+/// `now >= exp`, and not yet valid while `now < nbf` (no leeway).
+fn validate_temporal_claims(claims: &serde_json::Value) -> std::result::Result<(), String> {
+    // Backstop only: jsonwebtoken's own claim parsing already rejects
+    // non-object payloads today. Kept so a library change cannot silently turn
+    // the temporal checks below into no-ops (Go's MapClaims also rejects).
+    if !claims.is_object() {
+        return Err("token claims must be a JSON object".to_string());
+    }
+
+    let now = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map_err(|err| format!("system clock before unix epoch: {err}"))?
+        .as_secs_f64();
+
+    if let Some(exp) = numeric_date_claim(claims, "exp")? &&
+        now >= exp
+    {
+        return Err("token has expired".to_string());
+    }
+    if let Some(nbf) = numeric_date_claim(claims, "nbf")? &&
+        now < nbf
+    {
+        return Err("token is not valid yet".to_string());
+    }
+    Ok(())
+}
+
+/// Read an optional numeric-date claim, rejecting present-but-non-numeric values.
+fn numeric_date_claim(
+    claims: &serde_json::Value,
+    name: &str,
+) -> std::result::Result<Option<f64>, String> {
+    match claims.get(name) {
+        None => Ok(None),
+        Some(value) => {
+            value.as_f64().map(Some).ok_or_else(|| format!("claim {name} must be a numeric date"))
+        }
     }
 }
 
--- a/packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/api/server/auth.rs
+++ b/packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/api/server/auth.rs
@@ -85,6 +85,11 @@ fn validate_temporal_claims(claims: &serde_json::Value) -> std::result::Result<(
 }
 
 /// Read an optional numeric-date claim, rejecting present-but-non-numeric values.
+///
+/// A numeric value of zero is treated as an absent claim: golang-jwt v5's
+/// `MapClaims.parseNumericDate` returns nil for a zero float64 (a holdover
+/// from v4's struct zero-value semantics), so the Go client accepts `exp: 0`.
+/// `-0.0 == 0.0` on both sides, so negative zero follows the same rule.
 fn numeric_date_claim(
     claims: &serde_json::Value,
     name: &str,
@@ -92,7 +97,9 @@ fn numeric_date_claim(
     match claims.get(name) {
         None => Ok(None),
         Some(value) => {
-            value.as_f64().map(Some).ok_or_else(|| format!("claim {name} must be a numeric date"))
+            let seconds =
+                value.as_f64().ok_or_else(|| format!("claim {name} must be a numeric date"))?;
+            Ok((seconds != 0.0).then_some(seconds))
         }
     }
 }
--- a/packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/api/server/auth.rs
+++ b/packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/api/server/auth.rs
@@ -22,10 +22,13 @@ impl JwtAuth {
     /// Build a validator from a shared secret.
     pub(super) fn new(secret: &[u8]) -> Self {
         let mut validation = Validation::new(Algorithm::HS256);
-        // Validate signatures while keeping claims like `exp` optional.
+        // Claims stay optional (no claim is required to be present), but `exp`
+        // and `nbf` are validated whenever present with zero leeway — matching
+        // the Go client's echo-jwt defaults, which refuse expired tokens.
         validation.required_spec_claims.clear();
-        validation.validate_exp = false;
-        validation.validate_nbf = false;
+        validation.validate_exp = true;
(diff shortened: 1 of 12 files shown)


davidtaikocha:
All five findings confirmed against the jsonwebtoken 9.3.1 source and the Go `server.go` push path, and addressed in the follow-up commit:

**1. Malformed `exp`/`nbf` treated as absent** — Fixed, with a different mechanism than suggested. Typed optional claims (`Option<u64>`) would reject fractional numeric dates that RFC 7519 allows and Go accepts, and `Option<f64>` still misses the negative-`exp` case (jsonwebtoken's lenient parser turns it into `FailedToParse` → skipped). Instead the library layer now only verifies the signature, and `validate_temporal_claims` re-implements golang-jwt v5 semantics directly: claims optional; present claims must be numeric (string/null/array/bool → rejected); expired iff `now >= exp` (covers the `exp == now` boundary — note `reject_tokens_expiring_in_less_than = 1` was rejected as an alternative because `exp - 1` underflows u64 at `exp = 0`); not yet valid iff `now < nbf`; negative numeric dates parse like Go NumericDates (i.e. long expired); non-object claim payloads rejected like MapClaims. Tests cover malformed, null, negative, boundary, and fractional cases.

**2. `aud`-bearing tokens rejected** — Confirmed (`Validation::new` defaults `validate_aud: true` and `(Parsed, None) → InvalidAudience`); this predates this PR (main had the same default) but belongs here. `validate_aud = false` + audience-bearing token test.

**3. Explicit zero embedded signature served** — Fixed: the recent-cache gate is now `is_some_and(|sig| sig != [0u8; 65])`, falling through to the L2 rebuild whose zero-L1-origin-signature check declines — the same zero semantics Go applies (Go only ever serves from L2 state).

**4. EOS notification before materialization** — Fixed: the push moved from cache admission to `try_import_cached`, firing only on `Inserted` with a matching block hash (not `AlreadyMaterialized`, mirroring Go's already-canonical quit path which doesn't push). One deliberate deviation from exact parity: provenance is not restored, and deferred imports do notify. Exact parity is unattainable here — Go's envelope cache drops the EOS flag, so Go silently *loses* the notification for any EOS block that can't import on first delivery. The Rust push set is a strict superset of Go's, and never fires before the block exists locally. The epoch *marker* still records at admission (unchanged, documented) so EOS catch-up requests can be served for cached-but-not-yet-imported envelopes.

**5. `currentEpoch` from payload timestamp** — Fixed: the notification now carries `beacon_client.current_epoch()` read at push time, matching Go's `pushEndOfSequencingNotification` (which ignores its epoch argument and re-reads `CurrentEpoch()`). This also aligns the WS field with `/status`'s wall-clock epoch convention. The internal EOS *marker* keying remains timestamp-derived — a pre-existing, documented divergence outside this PR's scope.

davidtaikocha:
Both findings confirmed and fixed in the latest commit.

**Response-topic EOS spoofing** — Valid, and a real regression from the previous commit's "superset" decision: the EOS flag is operator-authenticated only on the payload topic (the wire signature covers the whole SSZ envelope, flag bytes included), while the response topic's embedded signature covers only the 32-byte block hash — the flag is attacker-controlled there. Fixed one step further than suggested: **all** EOS ingestion (the epoch marker feeding `/status`/EOS-request serving, not just the `/ws` notification) is now payload-topic-only, matching the Go client, which never ingests EOS state from responses. On provenance mechanics: the pending envelope cache overwrites same-hash entries, so re-reading the flag from the cached envelope at import time would still be forgeable in *both* directions — a later response could flip EOS on (spoofed handover) or off (suppressing a real one). The authenticated flag is therefore captured at admission into a bounded `PayloadEosTracker` keyed by block hash, consumed on `Inserted` with a matching hash. This makes the notification arrival-order-insensitive by construction (payload-then-response, response-then-payload, and duplicate re-marks all resolve to "notify iff the payload topic asserted EOS for the materialized hash"); the tracker's once-only/re-mark/eviction semantics are unit-tested. A full two-swarm arrival-order integration test remains blocked on the crate's known harness gap. Deferred payload-topic imports still notify — the one place this deliberately exceeds Go, whose envelope cache drops the EOS flag and silently loses those notifications.

**`exp: 0`** — Valid, confirmed in the pinned golang-jwt v5.2.1 source (`map_claims.go` `parseNumericDate`: a zero float64 returns nil, i.e. absent — a holdover from v4's struct zero-value semantics). A numeric-date value of zero is now treated as an absent claim for both `exp` and `nbf` (the same parse path governs both in Go), with parity tests. Negative zero follows the same rule (`-0.0 == 0.0` on both sides); negative dates remain "long expired", matching Go's NumericDate comparison. Worth noting for the record: v5's `json.Number` branch has *no* zero check, but echo-jwt's default JSON decoding produces float64, so the zero-as-absent branch is the live semantics.

PR #21970, opened 2026-07-20
chore(taiko-client-rs): surface async dial failures for configured whitelist peers

davidtaikocha:
## Motivation

While investigating why the hoodi `l2-node-reth` rs-driver sat at exactly 1 gossip peer for a week, we found its second configured bootnode (Chainbound, `35.239.142.239:4001`) has been dead the whole time — and nothing in logs or metrics showed it. The whitelist preconfirmation network runtime only observes errors `Swarm::dial` returns synchronously; connect timeouts/refusals arrive later as `OutgoingConnectionError` swarm events, which fell into the ignored catch-all. Result: ~16.4k dial timeouts over 5d17h (one per 30s retry tick), zero log lines, zero failure metrics. The only trace was `connected_configured_addr_count=1` in the periodic peer-status line.

Since the whitelist driver has no peer discovery (removed in #21913), the configured peers are the entire connectivity mechanism — a silently-dead configured peer is a silently-lost redundancy path, which showed up downstream as preconf response timeouts and ingestion gaps whenever the single remaining peer hiccuped.

## Changes

- Handle `SwarmEvent::OutgoingConnectionError` in the network runtime:
  - resolve each failed transport address to its configured source label (`bootnode` / `static peer`, else `unknown`),
  - count it under `whitelist_preconf_driver_network_dial_attempts_total{result="connect_failed"}`,
  - **warn once per outage** — edge-triggered via a `failing_configured_addrs` set so a peer that stays down produces one warning, not one per 30s tick; repeats log at debug,
  - log an info line when a previously-failing configured address connects again.
- Extract the bookkeeping into a pure `note_dial_failure` helper with unit tests, matching the module's existing test style.

## Testing

- `cargo test -p whitelist-preconfirmation-driver --lib` — 107 passed (2 new: first-failure edge-trigger + recovery re-arm, unconfigured addresses ignored)
- `just clippy` (workspace, `-D warnings -D missing_docs -D clippy::missing_docs_in_private_items`) — clean
- `cargo +nightly-2025-09-27 fmt` — applied

### DIFF
--- a/packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/network/runtime.rs
+++ b/packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/network/runtime.rs
@@ -262,6 +262,7 @@ impl WhitelistNetwork {
             event_tx,
             command_rx,
             connected_configured_addrs: HashSet::new(),
+            failing_configured_addrs: HashSet::new(),
             configured_peer_addrs,
             peer_retry_interval: delayed_interval(CONFIGURED_PEER_RETRY_INTERVAL),
             peer_status_log_interval: delayed_interval(PEER_STATUS_LOG_INTERVAL),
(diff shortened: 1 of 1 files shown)


davidtaikocha:
Superseded by #21971. Its configured-peer OutgoingConnectionError diagnostics, connect_failed metrics, once-per-outage warning/recovery bookkeeping, and regression tests were absorbed in f0f63f013, then hardened against stale overlapping failures in 7f18bd25e. The combined handler also preserves #21971’s discovered-peer dial-gate cleanup. Fresh final verification on #21971 passed 130 whitelist-driver tests, the targeted CLI regression, strict clippy, and the full Docker-backed just test suite.

PR #21971, opened 2026-07-20
feat(taiko-client-rs): optimize discv5 discovery + 20–30 peer management for the whitelist driver

davidtaikocha:
## Motivation

The Hoodi Rust whitelist-preconfirmation driver stayed at **1 gossip peer** for a week while the Go driver held roughly **23–28 peers**. The Rust driver had no working peer discovery, so a dead configured bootnode left Nethermind as a single connectivity path. That correlated with 304 preconfirmation response timeouts and repeated mid-flight ingestion stalls.

## What changed

This PR gives the Rust driver OP-stack-compatible discovery and peer management by reusing Kona components from the same pinned Optimism release already used by `kona-gossip`:

- `kona-disc` runs discv5 with an ENR signed by the same secp256k1 identity as libp2p, publishes the advertised IP/TCP/UDP endpoint, adds the OP-stack chain-ID entry, bootstraps `enr:` and `enode://...?discport=` records, and emits only chain-valid ENRs.
- `kona-peers` converts discovered ENRs into peer-ID-pinned TCP multiaddrs.
- `kona-gossip::ConnectionGater` supplies duplicate-dial suppression and redial throttling.
- Discovery dials candidates while the distinct connected-peer count is below `--p2p.peers.lo` (default 20).
- libp2p permits at most one established connection per `PeerId`, preventing duplicate connections from consuming the peer budget.
- A peer-aware soft high tide (`--p2p.peers.hi`, default 30) prunes unprotected distinct peers. Configured peer IDs are protected, pending disconnects are excluded from effective counts, and all-protected overflow is allowed.
- Public `NetworkConfig` validation rejects high tide 0, low tide above high tide, and discovery UDP port 0; low tide 0 remains supported.
- Direct TCP/0 bootnode targets are skipped while valid enode records remain available to discv5 through `discport`.

The ephemeral identity fallback changes from ed25519 to secp256k1 so the ENR identity and libp2p `PeerId` remain interoperable with Go peers.

## Absorbs #21970

This PR also absorbs #21970's configured-peer dial diagnostics:

- asynchronous `OutgoingConnectionError` transport failures increment `connect_failed` metrics with configured source labels,
- configured addresses warn once per outage and repeats log at debug,
- recovery re-arms the warning only for an exact configured-address connection,
- stale overlapping dial failures cannot reopen an outage after that address is connected,
- discovered-peer failures still clear Kona's in-flight dial state before classification so later discovery can retry.

## Flags and deployment contract

- `--p2p.discovery.addr` and `--p2p.disable-discovery` are active again.
- `--p2p.peers.lo` / `--p2p.peers.hi` default to 20/30.
- Discovery requires `--p2p.advertise.addr`; without it, the driver logs a warning and continues with configured peers.
- Deployment companion: taikoxyz/k8s-configs#1348 keeps the Go-driver static fallback while retaining the discovery flag, container UDP 30304, and Service UDP 30304.
- No deployment image tag is changed until an image containing this PR exists.

## Verification

- `cargo +nightly-2025-09-27 fmt --check`
- `cargo test -p whitelist-preconfirmation-driver --lib` — **130 passed**
- targeted CLI high-tide regression — passed
- strict workspace `just clippy` with warnings and missing docs denied — passed
- full Docker-backed `just test` — passed
- independent task reviews and final whole-branch review — no remaining Critical, Important, or Minor findings

Expected Hoodi steady state after image rollout: the Rust driver should top up toward 20 peers and remain roughly in the Go driver's 20–30-peer band, while retaining configured peers as protected fallback paths.


packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/network/behaviour.rs
```diff
@@ -44,15 +50,29 @@ impl From<identify::Event> for BehaviourEvent {
...
 pub(crate) fn build_behaviour(
     local_key: &identity::Keypair,
     topics: &Topics,
+    max_peers: usize,
 ) -> Result<TaikoBehaviour> {
     let mut gossipsub = build_gossipsub()?;
     topics.subscribe(&mut gossipsub).map_err(WhitelistPreconfirmationDriverError::p2p)?;
 
+    let limits = ConnectionLimits::default()
+        .with_max_established(Some(u32::try_from(max_peers).unwrap_or(u32::MAX)));
```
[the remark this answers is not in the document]
davidtaikocha: Addressed in c5f7e824e, 1bdbd9fb5, and 7f18bd25e. The global raw-connection cap is gone; libp2p now permits one established connection per PeerId, while a peer-aware soft high tide prunes distinct unprotected peers. Configured IDs are pre-protected, exact configured connections protect dynamically learned IDs, pending disconnects are excluded from effective counts, and all-protected overflow is allowed. Validation rejects hi=0. Fresh verification: 130 whitelist-driver tests, targeted CLI regression, strict clippy, and full Docker-backed just test all pass.

packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/network/discovery.rs
```diff
@@ -59,44 +74,201 @@ pub(crate) fn classify_bootnodes(bootnodes: Vec<String>) -> Vec<Multiaddr> {
...
+            "failed to convert P2P identity into discovery signing key: {err}"
+        ))
+    })
+}
+
+/// Convert a discovered ENR into a peer-id-pinned dialable TCP multiaddr.
+///
+/// Returns `None` for the local node itself and for ENRs without a nonzero dialable TCP
+/// socket or secp256k1 key.
+pub(crate) fn discovered_candidate(enr: &discv5::Enr, local_peer_id: &PeerId) -> Option<Multiaddr> {
```
YoGhurt111, line 197: **[P2] Reject special-purpose addresses from discovered ENRs**

`discovered_candidate` accepts any self-signed ENR with the matching chain ID, a secp256k1 identity, and a nonzero TCP port. Kona's ENR validation only enforces the `opstack` chain ID and also forwards matching `UnverifiableEnr` records, so an untrusted DHT participant can supply `0.0.0.0`, multicast, or broadcast TCP endpoints.

Those candidates reach `gater.dialing` and consume the low-tide in-flight budget until the connection-setup timeout. Rotating identities or addresses can repeatedly occupy the discovery slots and prevent legitimate candidates from being dialed; PeerId pinning only helps after TCP/Noise negotiation.

Please apply the same unspecified/multicast/broadcast validation used for the local advertised endpoint to discovered remote sockets before they enter the dial gate, and add regression tests for these address classes.

YoGhurt111 (commented):
Found one additional issue on the current head.

YoGhurt111:
The mainnet companion deployment work for this PR is tracked in [taikoxyz/k8s-configs#1359](https://github.com/taikoxyz/k8s-configs/issues/1359). It covers assigning an explicit whitelist-driver discovery UDP port, exposing the matching container/LoadBalancer port, rendering the mainnet manifests, and verifying live ENR reachability before rollout.

packages/taiko-client-rs/crates/whitelist-preconfirmation-driver/src/network/discovery.rs
davidtaikocha: Verified against the pinned Kona/libp2p revisions. The narrow observation is valid: `discovered_candidate` does not reject unspecified, multicast, or IPv4 broadcast TCP endpoints, while Kona's `EnrValidation` only checks the OP Stack payload and chain ID. Applying the same endpoint predicate here, with focused tests, is reasonable input hardening.

I do not think that predicate alone resolves the stated slot-exhaustion scenario. libp2p turns `0.0.0.0` into a ready transport error rather than holding it for the full setup timeout, multicast/broadcast may also fail before the timeout, and a rotating peer can still advertise global-unicast blackhole endpoints or accept TCP and stall Noise negotiation. That broader liveness threat requires candidate admission/fairness or rate-limiting controls beyond filtering these address classes.

YoGhurt111 (approved):

PR #21978, opened 2026-07-21
docs(protocol): create kimi-k3-audit.md

dantaik:


packages/protocol/audit/kimi-k3-audit.md
```diff
@@ -0,0 +1,89 @@
...
+/// @dev This function is callable by anyone to accept ownership without going through
+/// the TaikoDAO.
+function acceptOwnershipOf(address _contractToOwn) external nonReentrant {
+    Ownable2StepUpgradeable(_contractToOwn).acceptOwnership();
+}
+```
+
+**Attack scenario:**
+
+The DAO sends an L1→L2 governance message whose action batch includes `acceptOwnership()` on some contract (as `msg.sender = DelegateController`) followed by actions that depend on that ownership. A griefer front-runs message processing by calling `acceptOwnershipOf(target)` directly. The ownership transfer completes, but the in-batch `acceptOwnership()` call now reverts (`OwnableUnauthorizedAccount` — no pending transfer remains), reverting the **entire** message.
```
davidtaikocha, line 48: sounds not an attack for me, iirc this is a deliberate design we've made, and we should always avoid calling `acceptOwnership` directly in a batch and always use `acceptOwnershipOf`.

davidtaikocha (approved):

PR #21992, opened 2026-07-27
fix(taiko-client-rs): extend preconfirmation timeout

davidtaikocha:
## Summary

- increase the default preconfirmation payload submission timeout from 12 seconds to 24 seconds
- add a paused-time regression test covering the default response deadline

## Why

Masaya canonical proposal processing can hold the shared production router longer than 12 seconds. The preconfirmation caller can then return a retryable timeout while the queued job is still able to complete, causing Catalyst to retry the same height and increasing the risk of depth-1 reorgs.

Raising the timeout to 24 seconds puts the default above the observed proposal-processing tail. This is a mitigation only; cancellation and idempotency hardening remain separate follow-up work.

## Validation

- `just fmt-check`
- `cargo test -p driver --lib`
- `cargo clippy -p driver --all-targets --all-features --no-deps -- -D warnings`
- `git diff --check`

### DIFF
--- a/packages/taiko-client-rs/crates/driver/src/sync/event.rs
+++ b/packages/taiko-client-rs/crates/driver/src/sync/event.rs
@@ -166,7 +166,7 @@ impl ProposalRetryError {
 /// Default timeout for preconfirmation payload submission.
 ///
 /// Covers both the enqueue operation and awaiting the processing response.
-const PRECONFIRMATION_PAYLOAD_SUBMIT_TIMEOUT: Duration = Duration::from_secs(12);
+const PRECONFIRMATION_PAYLOAD_SUBMIT_TIMEOUT: Duration = Duration::from_secs(24);
 /// Timeout for best-effort `head_l1_origin` reset after an event-scanner reorg.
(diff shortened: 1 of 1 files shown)


YoGhurt111 (approved):

davidtaikocha:
This does not hang under Tokio’s paused-time runtime. Tokio auto-advances virtual time to the next pending timer when the runtime has no work to do, as documented in `tokio::time::pause`: https://docs.rs/tokio/latest/tokio/time/fn.pause.html

Here, the first outer timeout advances virtual time to 23 seconds and expires. The pinned submission remains alive with its internal deadline at 24 seconds. The second outer timeout then has a deadline at 25 seconds, so the internal 24-second deadline fires first and returns `PreconfResponseTimeout`.

Verified against this PR commit with:

```text
cargo test -p driver --lib default_preconfirmation_submission_waits_24_seconds_for_response -- --nocapture

1 passed; 0 failed; finished in 0.00s
```

No explicit `tokio::time::advance` is required, so no code change is needed for this feedback.

PR #21996, opened 2026-07-29
fix(bridge-ui): fix relayer claim path bug

RogerLamTd:


packages/bridge-ui/src/libs/relayer/RelayerAPIService.ts
```diff
@@ -276,17 +342,35 @@ export class RelayerAPIService {
...
+          bridgeTx.message = receiptMessage.message;
+        }
       }
 
-      if (!msgHash) return; //todo: handle this case
+      if (!bridgeTx.msgHash) return; //todo: handle this case
 
       const msgStatus = await RelayerAPIService._getBridgeMessageStatus({
-        msgHash,
+        msgHash: bridgeTx.msgHash,
```
davidtaikocha, line 583: **[P1] Keep the status field used by the relayer claim list in sync**

This lookup produces the canonical on-chain status, but the result is written only to `msgStatus`; `bridgeTx.status` remains from the relayer row that survived deduplication. `components/Relayer/Relayer.svelte` filters claimable transactions using `tx.status`, so a bad row marked `DONE` can still hide a corrected `NEW` message (and the reverse mismatch can expose a completed message). Assign the result to both fields, or migrate those consumers to `msgStatus`, and add a mismatched-status regression test.

packages/bridge-ui/src/libs/relayer/RelayerAPIService.ts
```diff
@@ -276,17 +342,35 @@ export class RelayerAPIService {
...
         srcChainId: Number(srcChainId),
         destChainId: Number(destChainId),
```
davidtaikocha: **[P1] Route the status lookup using the canonical message chain IDs**

`srcChainId` and `destChainId` were destructured from the possibly-corrupt relayer message before the receipt data was adopted. The hash/message can therefore be corrected while this status RPC still targets the stale destination route; the eventual claim code uses `bridgeTx.message` and may operate on a different route. After accepting `receiptMessage`, synchronize the top-level chain IDs and derive this lookup from the decoded message. Please add a destination-chain mismatch test.

packages/bridge-ui/src/libs/relayer/RelayerAPIService.ts
```diff
@@ -111,6 +120,63 @@ export class RelayerAPIService {
...
+          });
+          continue;
+        }
+
+        const normalizedMessage: Message = { ...message, gasLimit };
+        const user = getAddress(userAddress);
+        const senderMatch = getAddress(normalizedMessage.srcOwner) === user;
+        const receiverMatch = getAddress(normalizedMessage.destOwner) === user;
+
+        if (senderMatch || receiverMatch) {
```
davidtaikocha, line 255: **[P1] Do not bind blindly to the first user-owned `MessageSent` log**

A single source transaction can invoke `sendMessage` more than once, including multiple events for the same owner. In that case a valid relayer item describing a later log is silently overwritten with the first log's hash/message, so the UI can query or claim the wrong message. Prefer an exact match with the item's current `msgHash`; if no hash matches, only repair when there is exactly one user-owned candidate (or correlate by log index). Add a two-log regression test.

davidtaikocha (commented):
Inline review findings from the requested review. The existing processing-fee thread already covers that consistency issue, so I did not duplicate it.

davidtaikocha (approved):

ggonzalez94 (approved):
Reviewed at medium effort for bugs and architecture issues. The core approach — treating the on-chain `MessageSent` log as authoritative over the relayer row — is sound and well-tested. No funds-at-risk or architecture-breaking issues found, so approving.

One confirmed (non-blocking) display bug worth a follow-up:

- **Metadata wipe on msgHash correction** (`RelayerAPIService.ts:573`): when `msgHashChanged`, `destTxHash`/`claimedBy`/`fee` are wiped and never repopulated, so a corrected DONE transaction renders "Fee paid: 0 ETH", loses its dest-tx link, and gets mislabeled as relayer-claimed (`DesktopDetailsDialog.svelte:106` treats `claimedBy === undefined` + DONE as relayer). The wipe also fires when the relayer row's msgHash is merely absent rather than mispaired. Suggested guard: only wipe when both hashes are present and differ, and require non-null `claimedBy` for the isRelayer branch.

Plausible edge cases / cleanups (take or leave):
- Asset-detail overwrite runs unconditionally, so a decode fallback (stale vault address, older selector) can clobber correct relayer token metadata with the 0-ETH fallback even when the msgHash matched — consider only overwriting when `msgHashChanged`.
- The ambiguous-msgHash path drops the tx from the list with only a console.warn; combined with the transactionHash-only dedup, a multi-message tx with a corrupted relayer row becomes invisible.
- The missing-route hardening landed in the private `_getBridgeMessageStatus` copy; the shared `getMessageStatusForMsgHash` (used by the storage list and claim paths) still does the unguarded destructure.
- The hand-copied ABI fragments duplicate the generated `$abi` vault ABIs and the encoder tuples in `calculateMessageDataSize.ts` — importing from `$abi` would keep decoder and encoder in sync.

packages/bridge-ui/src/libs/relayer/RelayerAPIService.ts
```diff
@@ -1,12 +1,22 @@
...
-import { type Address, getAddress, type Hash, type Hex, numberToHex, type TransactionReceipt } from 'viem';
+import {
+  type Address,
+  decodeAbiParameters,
+  decodeEventLog,
+  decodeFunctionData,
+  getAddress,
+  type Hash,
+  type Hex,
+  numberToHex,
```
joker7333: J zu hykk jk

PR #22009, opened 2026-08-13
perf(protocol): avoid redundant fixed-k scalar multiplication

smtmfft:
## Summary
- use the known generator point and inverse scalar directly for the fixed k=1 signing attempt
- preserve the existing k=2 fallback through the generic path
- add coverage for both the normal k=1 result and the zero-s fallback case
- use the affine generator API supported by both standard k256 and the SP1-patched implementation

## Motivation
Taiko anchor transaction reconstruction intentionally uses fixed k=1. Running that known constant through generic secp256k1 generator multiplication and inversion is especially expensive inside zkVM guests. On a 192-block Shasta proposal, this change reduced RISC0 user cycles by another 801M after the KZG setup optimization, with identical guest output.

## Verification
- rustfmt +1.94.1 --edition 2024 --check packages/taiko-client-rs/crates/protocol/src/signer.rs
- cargo test -p protocol signer::tests -- --nocapture
- raiko2: just build-guest risc0 --force
- raiko2: just build-guest sp1 --bench --force
- raiko2: RISC0 and SP1 execution on proposal 23077 produced identical public output

### DIFF
--- a/packages/taiko-client-rs/crates/protocol/src/signer.rs
+++ b/packages/taiko-client-rs/crates/protocol/src/signer.rs
@@ -95,11 +95,17 @@ impl FixedKSigner {
         &self,
         hash: &[u8; 32],
     ) -> Result<SignatureWithRecoveryId, FixedKSignerError> {
-        for candidate in [Scalar::ONE, Scalar::from(2u64)] {
-            if let Ok(signature) = self.sign_with_specific_k(candidate, hash) {
-                debug!(?candidate, "generated signature with fixed k");
-                return Ok(signature);
-            }
+        if let Ok(signature) =
+            self.sign_with_k_components(AffinePoint::GENERATOR, Scalar::ONE, hash)
+        {
+            debug!(candidate = ?Scalar::ONE, "generated signature with fixed k");
+            return Ok(signature);
+        }
+
+        let candidate = Scalar::from(2u64);
+        if let Ok(signature) = self.sign_with_specific_k(candidate, hash) {
+            debug!(?candidate, "generated signature with fixed k");
+            return Ok(signature);
         }
         Err(FixedKSignerError::SigningFailed)
     }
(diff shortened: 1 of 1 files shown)


smtmfft:
The raiko2 integration and same-input benchmark are available in taikoxyz/raiko2#223. That PR rebuilds both RISC0 and SP1 guests from this commit and verifies identical proposal public output.

packages/taiko-client-rs/crates/protocol/src/signer.rs
```diff
@@ -162,6 +186,13 @@ impl FixedKSigner {
         let encoded = public_key.to_encoded_point(false);
         Address::from_raw_public_key(&encoded.as_bytes()[1..])
     }
+
+    fn ensure_golden_touch(&self) -> Result<(), FixedKSignerError> {
```
davidtaikocha: **Blocker:** Please add a doc comment for this private method. `crates/protocol/src/lib.rs` denies `clippy::missing_docs_in_private_items` outside tests, and the repository clippy command fails here with `missing documentation for a method`. I reproduced the failure on this PR head with `cargo +1.94.1 clippy -p protocol --all-features --no-deps -- -D warnings -D missing_docs -D clippy::missing_docs_in_private_items`.

packages/taiko-client-rs/crates/protocol/src/signer.rs
```diff
@@ -89,40 +94,59 @@ impl FixedKSigner {
...
     pub fn sign_with_predefined_k(
         &self,
         hash: &[u8; 32],
     ) -> Result<SignatureWithRecoveryId, FixedKSignerError> {
-        for candidate in [Scalar::ONE, Scalar::from(2u64)] {
-            if let Ok(signature) = self.sign_with_specific_k(candidate, hash) {
-                debug!(?candidate, "generated signature with fixed k");
-                return Ok(signature);
-            }
+        self.ensure_golden_touch()?;
```
davidtaikocha: **Blocker:** This guard is correct for fixed-`k` safety, but the current production WPD passes its operator P2P key into `FixedKSigner` (`runner.rs:123`). `build_preconf_block` then reaches `sign_digest`, so every otherwise-valid build using a non-golden operator key returns `UnsupportedSigner` here—after local insertion, but before `set_l1_origin_signature` and gossip publication. Please migrate the WPD production path to a normal ECDSA signer (`PrivateKeySigner` or an equivalent signer abstraction), promote the signer crates from dev-dependencies, and add a concrete service test covering signing and publication. This also needs security follow-up: known `k = 1` exposes a non-public ECDSA private key from one signature (accounting for low-`s` normalization), so any such key that successfully signed through this path should be rotated.

packages/taiko-client-rs/crates/protocol/src/signer.rs
smtmfft: Fixed in the updated PR: `sign_with_k_components` now documents the nonce-component invariant, and the k=1 fast path has an inline invariant comment.

packages/taiko-client-rs/crates/protocol/src/signer.rs
smtmfft: Updated this PR to be protocol-only again so it no longer carries the WPD production signer change. The WPD migration to `PrivateKeySigner` is handled in #22011, which I reviewed and verified separately. With this branch now only changing `protocol/src/signer.rs`, it should not reintroduce the WPD fixed-k path when #22011 merges first.

davidtaikocha (approved):
No blocking findings. The k=1 fast path is equivalent to the existing generic path (`kG = G`, `k^-1 = 1`), while the k=2 fallback behavior is preserved and covered by the zero-s test.

I also verified the PR merges cleanly with #22011 and ran the signer tests with default features and `--no-default-features`, plus strict protocol clippy, on the combined tree.

PR #22011, opened 2026-08-13
chore(taiko-client-rs): use standard preconfirmation signer

davidtaikocha:
## Summary

- replace the Rust whitelist preconfirmation driver's fixed-k signer with Alloy's standard `PrivateKeySigner`
- keep the existing preconfirmation signing domain and `r || s || yParity` wire encoding
- restrict `FixedKSigner` construction to the public golden-touch key used by `anchorV4`
- update the wrong-anchor-sender test to use a normal signer and add a regression guard for the whitelist service signer type

## Root cause

`FixedKSigner` uses the public, predefined nonce candidates `k = 1` and `k = 2`. That is acceptable for golden-touch anchor transactions because the golden-touch private key is public. It is unsafe for the secret whitelist operator key: one observed signature and its digest are sufficient to recover the private key.

The Go preconfirmation path already uses standard `crypto.Sign`; this change restores Rust/Go parity.

## Impact

Rust whitelist preconfirmation block-hash and gossip-envelope signatures now use standard RFC6979 secp256k1 signing. Signature verification and wire compatibility are unchanged.

This affects `WLP-INV-010` cross-implementation parity only. Ingress gating, stale boundaries, reorg handling, and execution-engine custom-table behavior are unchanged.

## Validation

- `just fmt`
- `just clippy-fix`
- `just unit`: 382 workspace tests and 43 `protocol --no-default-features` tests passed
- `cargo test -p whitelist-preconfirmation-driver --lib`: 146 passed
- `just test` was attempted but stopped before tests because Docker could not refresh the local `gcloud` credentials while pulling `us-docker.pkg.dev/evmchain/images/alethia-reth:main`

### DIFF
--- a/packages/taiko-client-rs/Cargo.lock
+++ b/packages/taiko-client-rs/Cargo.lock
@@ -10140,6 +10140,8 @@ dependencies = [
  "alloy-rpc-types 1.8.3",
  "alloy-rpc-types-engine 1.8.3",
  "alloy-rpc-types-engine 2.2.0",
+ "alloy-signer",
+ "alloy-signer-local",
  "arc-swap",
  "async-trait",
  "axum",
--- a/packages/taiko-client-rs/crates/protocol/src/signer.rs
+++ b/packages/taiko-client-rs/crates/protocol/src/signer.rs
@@ -45,7 +45,7 @@ pub enum FixedKSignerError {
     SigningFailed,
 }
 
-/// Deterministic secp256k1 signer.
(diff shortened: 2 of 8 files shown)


smtmfft (approved):
Looks good to me. This removes the WPD production use of FixedKSigner by switching the preconfirmation P2P signer to PrivateKeySigner, while keeping fixed-k signing scoped to golden-touch anchor transactions by making FixedKSigner::new private. I also verified the focused WPD signer test and protocol signer tests locally.

YoGhurt111 (approved):

Issue #19832, opened 2025-07-29
Explore L2 blob support


dantaik:
## Context

I'm considering adding native blob support on L2, where blobs are submitted directly to L2 and retained by L2 nodes for a limited period (e.g., a few weeks), without being relayed to L1.

## Motivation

This feature could enable cheaper temporary data availability for L2 applications, especially those that don’t require L1 finality or long-term persistence. It aligns with use cases like short-lived rollup state, preconfirmation metadata, or off-chain compute commitments.

## Challenges

Implementing L2-native blobs may require substantial research and design work, including:

- How to store, retrieve, and garbage-collect blob data efficiently on L2
- What kind of guarantees (availability, integrity) can be provided
- How to integrate blob usage with rollup state transitions or preconfirmations
- Whether changes to the L2 execution environment are needed

## Next steps

This issue is for exploration and early design. We should begin by surveying similar efforts, defining possible use cases, and identifying constraints in the current node and protocol architecture.

PR #22014, opened 2026-08-13
docs(protocol): add design spec for L2-native ephemeral blobs

### LINKED ISSUE #22015: taiko-geth: post-Unzen txpool accepts type-3 txs that every sealer drops, so proposers pay L1 DA for dead bytes
## Problem

Since Unzen activated Cancun/Prague/Osaka on L2 (mainnet fork time 2026-08-06 13:00 UTC), taiko-geth's blobpool is fork-active: a wrapped type-3 transaction submitted via `eth_sendRawTransaction` now passes txpool validation (`core/txpool/validation.go` only rejects blob txs pre-Cancun) and sits in the blobpool.

The proposal path then picks it up: `buildTransactionsLists` and `getPendingTxs` call `w.txpool.Pending(...)` **without** `OnlyPlainTxs: true` (`miner/taiko_worker.go:74-79, 377`), so the blob tx can be committed into a proposed txlist. But every sealer skips type-3 at block import (`miner/taiko_worker.go:277-281`, "Skip blob transactions"), so the transaction can never execute.

Net effect:

- The proposer pays L1 DA (manifest bytes in the proposal blob) for transactions that are guaranteed to be dropped by every node at sealing.
- Depending on the encoding path, a pool-resident blob tx RLP-encodes in wrapped form (~128 KiB sidecar per blob), inflating txlist bytes dramatically before pruning.
- The blob tx lingers in the blobpool indefinitely (accepted, never mined).

## Proposed fix

Reject `BlobTxType` in taiko-geth's txpool validation while native L2 blob support is not active (a `CHANGE(taiko)` gate alongside the existing Taiko checks in `core/txpool/validation.go`), and/or pass `OnlyPlainTxs: true` in both `Pending` calls in `miner/taiko_worker.go`. Either is a one-line-scale change; doing both is belt-and-suspenders.

When native L2 blob support ships (design under review in #22014, §8 "Pool" rows), the gate becomes fork-conditional (`!IsV7`-style) rather than unconditional.

## References

- Design spec PR: #22014 (`packages/protocol/docs/l2_native_blobs.md`, §8 and rollout phase 0)
- Prior context: #19832 (Explore L2 blob support), #19831 (EVM upgrade, shipped in Unzen)

dantaik:
## Summary

Adds `packages/protocol/docs/l2_native_blobs.md` — a detailed design specification for **native EIP-4844 type-3 transaction support on Taiko L2**, with blob bodies retained and served by Taiko's node network and **never posted to L1** ("ephemeral native blobs", the design sketched in #19832).

Core design points:

- **Invariant**: execution state depends only on `blobVersionedHashes` (which ride to L1 inside the ordinary txlist manifest), never on blob bodies — so derivation, proving, and the L1 contracts are untouched by body availability. A lost body strands its buyer; it cannot fork the chain or block a proof.
- **Retention: `BLOB_RETENTION_SECONDS_MIN = 6_291_456` (~72.8 days) — exactly 4× Ethereum's 4096-epoch (~18.2-day) sidecar window**, with a 90-day default for serving nodes. Retention is a service role (preconfers MUST, RPC providers SHOULD serve); other full nodes may prune freely.
- **Launch throughput/fee lane**: target 1 / max 3 blobs per block, EIP-4844 excess-blob-gas market with the EIP-7918 reserve mechanism (structure quoted from the Final EIP, including the leading early return; L1-identical arithmetic widths) and a Taiko-specific update fraction giving ×1.125-per-full-block max growth — deliberately the responsiveness of L1's original Cancun market (current L1 runs ≈×1.0817); blob fees burn at launch, with the routing decision due before mainnet and implementation deferred to a follow-up fork.
- **Availability guarantee at launch**: bonded-preconfer custody — one uniform acceptance predicate (cells recomputed from the blob, proofs batch-verified) so honest nodes can never disagree on an envelope, and the envelope signature commits to a fixed sha256 sidecar digest, making in-transit stripping detectable and never blamed on the operator. Documented as committee-grade DA (governance-grade enforcement) with a PeerDAS-style sampling upgrade path; the fake-hash/sidecar-less vector is analyzed and gated on the permissioned proposer set (§12.5).
- **IPFS archival profile (§7.4)**: an optional, off-by-default Kubo-sidecar profile (put-on-receipt, pin-on-seal) — blobs stored as single raw CIDv1 blocks, grouped into deterministic per-1000-block dag-cbor bundles whose **range-root** CIDs converge bit-for-bit across independent archivers (seal-log nodes carry lineage and are the DHT-announced pin roots; empty ranges are sealed so `skippedRanges` means only "incomplete"); bundle metadata outlives body retention so post-window retrieval has a discovery bootstrap; the `versionedHash → CID` index is untrusted because retrieval is doubly self-verifying (CID check + KZG re-derivation). Explicitly a replication/archival layer, not part of the availability guarantee.
- **Zero required L1 contract changes; modest, mostly upstream-inherited prover changes** (bodies are never proven; the type-3 STF rules come from revm/reth with mandatory parity vectors gating devnet; blob opcodes already carry Unzen zk-gas prices). The bulk of the work is preconf sidecar gossip, the per-node sidecar store/serving API, and execution-client fork gating — the `engine_getBlobsV2/V3` Osaka fork gate is already satisfied (Unzen mapped `OsakaTime`) and must simply not regress.
- **Forced-inclusion sources exclude type-3 at launch** with per-transaction skipping anchored to the on-chain `isForcedInclusion` flag, wallets/RPCs guarding users from paying forced-inclusion fees for unexecutable type-3 txs, and a security analysis covering withholding grief, storage-DoS bounds, fake-hash inclusion, and zk-gas truncation interplay. Appendix C consolidates the mandatory test-vector checklist.

Also flags an immediate hygiene item independent of this design, tracked in #22015: post-Unzen the blobpool is Cancun-active and `Pending()` isn't filtered, so a type-3 tx can enter a proposed txlist today only to be dropped by every sealer at import — the pool should reject type-3 (with a normative error message) until the fork activates.

## Review notes

@davidtaikocha — requesting your review, particularly on:

1. Launch parameters (§3, §15.2): `TARGET 1 / MAX 3` per block, `MAX_BLOBS_PER_TX = 2`, and the 90-day serving default over the 72.8-day minimum.
2. The sidecar hand-off path (§6): sealer includes envelopes → driver pulls sidecars via `engine_getBlobsV2` → preconf envelope carries `blobSidecars` — vs. alternatives you may prefer in the taiko-reth/Rust-driver architecture.
3. The §4.3.3 forced-inclusion restriction and the §12.4 censorship trade-off (plus the §15.3 envelope-only middle path).
4. Whether phase-0 hygiene (#22015) should land in taiko-geth now.
5. The IPFS profile (§7.4): duplicated Kubo blockstore (~2× disk, simple) vs filestore dedup — resolve before phase 3 — and whether sealed bundle CARs should feed a Filecoin/EthStorage incentivized tier (§15.7).
6. Blob-fee posture: launch burn with the routing decision due before mainnet activation (§5, §15.5).

## Testing

Documentation-only change; no code paths affected. Markdown/mermaid rendering verified.

https://claude.ai/code/session_01Jo9EivPkB9G8vxYB4rcVZU


packages/protocol/docs/l2_native_blobs.md
```diff
@@ -0,0 +1,378 @@
+# L2-Native Ephemeral Blobs
+
+This document specifies **ephemeral native blob support** for Taiko Alethia: accepting EIP-4844 type-3 (blob-carrying) transactions on the L2 itself, with blob bodies retained and served by Taiko's node network for a bounded window and **never posted to L1**.
+
+- **Status**: Draft for review
+- **Target**: the next hard fork after Unzen (placeholder `V7` throughout; fork gate `IsV7(timestamp)`)
+- **Depends on**: Unzen (Cancun+Prague+Osaka EVM on L2, zk-gas metering), Shasta derivation
+- **Related**: [#19832 — Explore L2 blob support](https://github.com/taikoxyz/taiko-mono/issues/19832), [Derivation.md](./Derivation.md), [zk_gas_spec.md](./zk_gas_spec.md)
```
dantaik, line 8: **Citation fix: the go.mod pin lives at the repo root, not `packages/taiko-client/go.mod`**

Verified against taiko-mono @ `8843a78`: there is no `packages/taiko-client/go.mod` — the monorepo has a single Go module at the root. The pin is the root `go.mod` replace directive:

```go
replace github.com/ethereum/go-ethereum v1.15.5 => github.com/taikoxyz/taiko-geth v1.18.1-0.20260715122422-ca164c06e00c
```

The cited revision itself is correct (it matches the taiko-geth tree I verified the other `file:line` citations against), so this is just a path fix so implementers can find the pin.

packages/protocol/docs/l2_native_blobs.md
```diff
@@ -0,0 +1,378 @@
...
+
+  ```python
+  def calc_excess_blob_gas(parent):
+      # reserve condition: blob base fee below the execution-linked floor (≈ baseFee/16)
+      if BLOB_BASE_COST * parent.base_fee_per_gas > GAS_PER_BLOB * fake_exponential(MIN_BLOB_BASE_FEE, parent.excess_blob_gas, BLOB_BASE_FEE_UPDATE_FRACTION):
+          # below the floor the excess never decays; it grows with usage, scaled by (MAX−TARGET)/MAX
```
dantaik: I cross-checked this section against the current EIP-7918 text (status: **Final**; requires EIP-4844 + EIP-7840). The condition matches the EIP verbatim:

```python
if BLOB_BASE_COST * parent.base_fee_per_gas > GAS_PER_BLOB * get_base_fee_per_blob_gas(parent):
```

- Both sides are total wei: `2**13 × base_fee_per_gas` vs `131072 × blob_base_fee`. Dividing by `GAS_PER_BLOB`, the floor condition is exactly `blob_base_fee < base_fee_per_gas / 16` — the doc's "≈ baseFee/16" annotation. It is a price-vs-price comparison; neither side is "per-gas" vs "per-blob".
- `BLOB_BASE_COST = 2**13` ✓, strict `>` ✓, no `max()` clamp on the price function ✓, reserve-branch scaling `blob_gas_used × (max − target) // max` ✓.

For the record: the "DeepSeek Code Review" comment claiming the comparison must be expressed in per-gas units is incorrect — the EIP compares totals on both sides, which is what this doc writes.

Two suggestions to make this section unassailable:

1. The "verified verbatim" claim invites exactly this kind of dispute; consider quoting the upstream `calc_excess_blob_gas` snippet (or deep-linking the Specification section) instead. Note one structural difference if quoting it: the EIP has an explicit early return `if parent.excess_blob_gas + parent.blob_gas_used < target_blob_gas: return 0` *before* the reserve branch. It is numerically identical to the doc's `max(0, …)` else-branch, so behavior matches, but "verbatim branch structure" isn't literally true.
2. Add the per-gas equivalent form — reserve active while `blob_base_fee < base_fee_per_gas / 16` — next to the inequality. One line, and it defuses the unit confusion permanently while making the §12.2 "the floor is not the anti-spam mechanism" argument easier to reason about on Taiko, where the execution base fee is small.

packages/protocol/docs/l2_native_blobs.md
```diff
@@ -0,0 +1,378 @@
...
+
+## 4. Consensus rules (provable, fork-gated on `IsV7`)
+
+### 4.1 Transaction validity
+
+A type-3 transaction's structure and validity follow EIP-4844 (Osaka revision) verbatim; Taiko's §3 schedule caps apply on top as consensus limits. In a V7+ block:
```
dantaik, line 69: Verified against the pinned taiko-geth (`eth/catalyst/api.go`): both `GetBlobsV2` and `GetBlobsV3` short-circuit to `null` while `config().LatestFork(head.Time) < forks.Osaka`. On Taiko that gate is subtle, because:

- the `forks` enum has **no Unzen ID** (ordering is `… Prague → Osaka → BPO1 …`), and `LatestFork` is computed purely from the standard fork times (`IsOsaka` etc.);
- `ChainConfig` carries a separate Taiko-only `UnzenTime` field (`params/config.go`), and the bundled `TaikoChainConfig` in `params/taiko_config.go` sets no `OsakaTime` at all.

So the §6.2 hand-off only works if the V7 network config maps `OsakaTime` (the fork the catalyst gate actually sees) to the Unzen/V7 activation. If it doesn't, every `engine_getBlobsV2` call on Taiko returns `null` forever and the builder's rebuild loop silently drops every type-3 transaction — a failure mode that would look exactly like "blobpool evicted everything". Worth one normative sentence in §6.2 ("at V7 the execution client MUST report `LatestFork ≥ Osaka`") plus a devnet checklist item.

The same fork-keying issue applies to the §8 row on `BlobScheduleConfig`: it is keyed by fork ID (`Cancun/Prague/Osaka/BPO1–BPO5`), so "replace with Taiko entries keyed to V7" should state which key the Taiko schedule attaches to (presumably the Osaka slot, since Unzen already activates the Osaka EVM) or whether a Taiko key must be added.

packages/protocol/docs/l2_native_blobs.md
```diff
@@ -0,0 +1,378 @@
...
+
+### 7.2 Retention and serving
+
+- **Requirement (this design's headline parameter): the network MUST retain and serve blob bodies for at least `BLOB_RETENTION_SECONDS_MIN = 6_291_456 s` (≈ 72.8 days).** The absolute seconds are the normative value; the 4×-Ethereum's-window relationship (Appendix A) is provenance as of 2026 and does not track future L1 changes. Taiko out-retains L1 by design; retention length is a product feature.
+- Node roles:
+  - **Serving nodes** (preconfers MUST; public RPC providers SHOULD): retain ≥ the minimum window (default config 90 days) and expose a **publicly reachable** serving API (or formally delegate serving to a public provider with equivalent retention — a private-only endpoint does not discharge the obligation), and SHOULD export retention/serving metrics so §7.1 under-retention is externally observable — at minimum `taiko_blob_store_oldest_age_seconds` (retention horizon), `taiko_blob_store_backfill_gaps` (missing bodies inside the window), and `taiko_blob_served_total{endpoint}`.
```
dantaik, line 177: The enforcement story in this section is "attributable operator misbehavior" with governance-grade slashing, and §12.5(b) leans on the same attribution. But the spec never says whether the preconf gossip signature covers `blobSidecars`. Today the driver's `Envelope.Signature` covers the gossip envelope bytes (`pkg/preconf/payload.go`), and §9 adds the sidecars as a sibling field. If sidecars ride along unsigned:

- an operator can plausibly claim the envelope it signed never included sidecars (he-said-she-said in a slashing dispute), and
- a malicious relayer can strip sidecars from a gossiped envelope, turning the §6.3 refusal rule into a liveness attack that also makes the *honest* operator look like the withholder to every peer behind the stripper.

Cheap fix: have the signed envelope commit to a **digest** of the sidecars (e.g., sha256 over the ordered per-hash sidecar digests, or a Merkle root) rather than the full ~400 KiB; §6.3 validation already recomputes each sidecar anyway, so verification cost is unchanged. Recommend a normative sentence: "the preconf signature MUST commit to the `blobSidecars` (directly or via a digest)", plus a note in the §9 codec-bump row that pre-V7 peers must tolerate the extra field. This also gives the §15.4 attestation idea a signed artifact to build on later.

packages/protocol/docs/l2_native_blobs.md
```diff
@@ -0,0 +1,378 @@
...
+- **Announcement.** Only seal-log roots receive DHT/IPNI provider records (Kubo `Reprovider.Strategy = "roots"`); range roots are reachable through them, and individual blob CIDs are never announced. Fetching one blob over IPFS means: obtain the range root (from a serving node's `GET /taiko/v1/blob_bundles?range=…`, an indexer, or via the DHT once a seal-log root CID is already known from the HTTP endpoint, an indexer, or a prior bundle's `prev` chain — the DHT locates providers for known CIDs; it is not a discovery bootstrap) → read the index → Bitswap the single raw leaf from a peer holding the bundle. Latency is accordingly worse than the HTTP API — acceptable for archival, wrong for the hot path.
+
+#### 7.4.2 Kubo sidecar with pin-on-receipt
+
+Reference deployment: a Kubo daemon co-located with the Taiko node (same pod / compose service), reached over its local RPC API. The driver's sidecar store gains an asynchronous IPFS worker; **no IPFS operation ever sits on the block-import, preconf-validation, or serving path.**
+
+```mermaid
```
dantaik, line 224: The post-window promise ("post-window retrieval gains a second network") depends on discovering a seal-log root CID *after* the 90-day window — but the only discovery surfaces are:

- `GET /blobs/{versionedHash}` → `bundleRoots` and `GET /taiko/v1/blob_bundles?range=…`, both served by §7.2 serving nodes whose retention the spec bounds at `BLOB_RETENTION_SECONDS_DEFAULT`, and
- the IPFS profile is off by default, so a default serving node never produces bundle metadata at all (its `bundleRoots` field will be empty).

So the pointers into IPFS can be pruned exactly when blob bodies are. The metadata is small relative to the blobs: at the launch target (1 blob/block) the sealed index is ≈120 B × 1000 entries per bundle ≈ 43 bundles/day ≈ **~5 MB/day** — negligible next to §12.2's numbers, and only range-root/seal-log nodes need to be pinned. Suggest specifying:

1. Serving/archiver nodes SHOULD retain the sealed bundle index (range → rangeRoot/sealLogRoot) indefinitely, independent of blob retention; or
2. If that obligation is unwanted, name the indexer/IPNI layer that owns long-term root discovery and state its retention expectation, so post-window retrieval has a defined bootstrap path.

Otherwise the DHT can only locate bundles for consumers who already learned a root CID from a source that no longer serves it.

packages/protocol/docs/l2_native_blobs.md
dantaik: Both suggestions adopted in 279d929: the pseudocode now quotes the Final EIP's structure including the leading early return, and the per-gas form (`blob_base_fee < base_fee_per_gas / 16`) is annotated inline on the condition.

One sharpening of point 1: the early return turns out to be slightly stronger than "numerically identical to the `max(0, …)` else-branch." Because it also *precedes* the reserve branch, it fires below the floor too — flushing sub-target residual excess to zero where the previous else-branch-only formulation would have held it. The corner is reachable from genesis: one exactly-target block from empty excess lands `excess = 87_381 < TARGET`; a following zero-usage block resets it to 0 under the quoted structure, but held 87,381 under the old form. The price difference is negligible (both ≈1 wei), but the `excessBlobGas` header values differ — consensus-relevant — so the corner is now called out in §5 and added to the mandatory test vectors. Good catch; quoting the EIP rather than paraphrasing it is exactly right.

---

PR #22016, opened 2026-08-13
fix(protocol): address review findings on Proposal0020

dantaik:
Applies the actionable findings from the [in-depth review of #22010](https://github.com/taikoxyz/taiko-mono/pull/22010#issuecomment-5278626312), stacked on `revamp-security-council`.

## Fixes

- **F1 — broken pre-approval verification command** (`Proposal0020.md`): the documented `getProposal` cast signature omitted the `uint16 approvals` return value, so the exact command approvers are told to run before approving failed with an ABI decode error. Corrected signature verified working against mainnet proposal 21.
- **F2 — stale `verify`-mode instruction**: `BuildDirectProposal.logProposalAction` still wrote "run the `verify` mode" (dropped in 6d57357) into every generated action file. The template now points at the `getProposal` comparison from the spec, and `Proposal0020.action.md` is regenerated — **all six action To/Value/Data fields are byte-identical**; only the instruction text changes.
- **F3 — silent invalid `MODE`**: `run()` now reverts with `InvalidMode(mode)` instead of logging and exiting 0.
- **F7 — EOA assumption**: `checkBaseline` now asserts `SC_GUSTAVO_GONZALEZ.code.length == 0`.
- **F4 — key-registration mechanism**: post-execution step 2 now names `setPublicKey` (the agent path) instead of implying `setOwnPublicKey`, which reverts `MustBeListed` for a non-listed agent.
- **F5 — `removeUnused()` accuracy**: the Action 6 comment and spec now state it prunes the account enumeration only; the removed members' `appointerOf`/agent mappings persist (harmlessly).
- **F6 — ordering precision**: spec and script comments now distinguish the contract-enforced dependency (Action 1 before Action 3) from the defensive placement of Actions 4–5.

## Verification

- Invalid mode: `MODE=bogus` now fails the script (`InvalidMode("bogus")`).
- `MODE=print` regeneration: committed `Proposal0020.action.md` matches; calldata diff vs the previous file is empty.
- `MODE=l1dryrun` on a mainnet fork: green end-to-end, including the new EOA baseline assertion.
- `forge fmt` with the repo config; no lines exceed 100 chars.

https://claude.ai/code/session_01NXZ8HWv6KYVV3fgC6MGJiv

---
_Generated by [Claude Code](https://claude.ai/code/session_01NXZ8HWv6KYVV3fgC6MGJiv)_

### DIFF
--- a/packages/protocol/script/layer1/governance/BuildDirectProposal.sol
+++ b/packages/protocol/script/layer1/governance/BuildDirectProposal.sol
@@ -35,6 +35,7 @@ abstract contract BuildDirectProposal is Script {
 
     error MissingEnv(string name);
     error CheckFailed(string what);
+    error InvalidMode(string mode);
 
     function run() external {
         string memory mode = vm.envString("MODE");
@@ -44,6 +45,7 @@ abstract contract BuildDirectProposal is Script {
             dryrunL1();
         } else {
             console2.log("Error: Invalid mode. Must be one of: print, l1dryrun");
+            revert InvalidMode(mode);
         }
     }
 
@@ -76,7 +78,10 @@ abstract contract BuildDirectProposal is Script {
             vm.toString(L1.DAO_STANDARD_MULTISIG),
             "`.\nThe UI pins the metadata and assembles `createProposal`; paste each action",
             " below into the\nUI's custom-action (calldata) form, in order. After creation,",
-            " run the `verify` mode\nagainst the new proposal id before approving.\n\n",
+            " compare the actions\nstored on-chain against this file before approving (the",
+            " `getProposal` command in\nProposal",
+            _proposalId,
+            ".md).\n\n",
             "- Destination plugin (set by the UI): `",
             vm.toString(L1.DAO_OPTIMISTIC_TOKEN_VOTING_PLUGIN),
             "`\n"
--- a/packages/protocol/script/layer1/governance/BuildDirectProposal.sol
+++ b/packages/protocol/script/layer1/governance/BuildDirectProposal.sol
@@ -33,10 +33,6 @@ abstract contract BuildDirectProposal is Script {
         bytes data;
     }
 
-    error MissingEnv(string name);
-    error CheckFailed(string what);
-    error InvalidMode(string mode);
-
     function run() external {
         string memory mode = vm.envString("MODE");
         if (keccak256(abi.encodePacked(mode)) == keccak256(abi.encodePacked("print"))) {
@@ -259,9 +255,17 @@ abstract contract BuildDirectProposal is Script {
         return "?";
     }
 
-    function readRaw(address _target, bytes memory _data) private view returns (bytes memory) {
+    function readRaw(address _target, bytes memory _data) internal view returns (bytes memory) {
         (bool ok, bytes memory ret) = _target.staticcall(_data);
         check(ok, "staticcall reverted");
         return ret;
     }
+
+    // ---------------------------------------------------------------
+    // Custom Errors
+    // ---------------------------------------------------------------
+
+    error MissingEnv(string name);
+    error CheckFailed(string what);
+    error InvalidMode(string mode);
 }
--- a/packages/protocol/script/layer1/governance/BuildDirectProposal.sol
+++ b/packages/protocol/script/layer1/governance/BuildDirectProposal.sol
@@ -264,6 +264,7 @@ abstract contract BuildDirectProposal is Script {
         returns (bytes memory)
     {
         (bool ok, bytes memory ret) = _target.staticcall(_data);
+        logIfReverted(ok, ret);
         check(ok, "staticcall reverted");
         return ret;
     }
--- a/packages/protocol/script/layer1/governance/BuildDirectProposal.sol
+++ b/packages/protocol/script/layer1/governance/BuildDirectProposal.sol
@@ -255,7 +255,14 @@ abstract contract BuildDirectProposal is Script {
         return "?";
     }
 
-    function readRaw(address _target, bytes memory _data) internal view returns (bytes memory) {
+    function readRaw(
+        address _target,
+        bytes memory _data
+    )
+        internal
+        view
+        returns (bytes memory)
+    {
         (bool ok, bytes memory ret) = _target.staticcall(_data);
         check(ok, "staticcall reverted");
         return ret;
--- a/packages/protocol/script/layer1/proposals/Proposal0020.s.sol
+++ b/packages/protocol/script/layer1/proposals/Proposal0020.s.sol
@@ -209,6 +209,20 @@ contract Proposal0020 is BuildDirectProposal {
             );
         }
 
+        // Action 6 effect: every account still enumerated by the registry is a listed
+        // signer, i.e. the removed members were pruned (their appointerOf/agent mappings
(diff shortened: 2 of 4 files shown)


packages/protocol/script/layer1/governance/BuildDirectProposal.sol
```diff
@@ -35,6 +35,7 @@ abstract contract BuildDirectProposal is Script {
 
     error MissingEnv(string name);
     error CheckFailed(string what);
+    error InvalidMode(string mode);
```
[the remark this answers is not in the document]
dantaik: Fixed in 2b57591 — all three custom errors (`MissingEnv`, `CheckFailed`, `InvalidMode`) now live in a final Custom Errors section, resolving the pre-existing layout violation as well.

---

dantaik:
Re the DeepSeek suggestions, disposition in 2b57591:

- **Taken — make the `removeUnused()` claim auditable**: `checkPostState` now asserts every account still enumerated by the registry is a listed signer (proving the pruning happened), and the spec's Action 6 note links the `removeUnused()` implementation. The persistence of `appointerOf` mappings stays documented prose rather than an assertion, since asserting implementation residue would make the dryrun fail if a future registry cleaned up more thoroughly.
- **Declined — mode constants/enum**: `BuildDirectProposal.run()` mirrors the existing `BuildProposal.run()` string-dispatch convention; diverging in one base only would be inconsistent. Happy to refactor both in a follow-up if the team wants it.
- **Declined — drop the log before `revert InvalidMode`**: kept intentionally; the log names the valid modes for the operator while the custom error carries the offending value.

---

dantaik:
Re the latest DeepSeek round — no changes made, with reasons:

- **The "overly broad invariant" warning is incorrect.** In the deployed `EncryptionRegistry`, `accountList` only ever gains *owner* accounts that are listed at registration time (`appointAgent` requires `isListed(msg.sender)`; `setOwnPublicKey`/`setPublicKey` require the owner to be listed; agents are never enumerated as accounts). `removeUnused()`'s postcondition is exactly "no unlisted account remains enumerated", and the assertion runs immediately after Action 6 in the same execution — so it asserts precisely the action's guarantee, not something stronger. Checking only the five removed seats would be strictly weaker.
- **`abi.encodeCall`**: would require vendoring the Aragon interfaces into taiko-mono; the whole `BuildDirectProposal` family deliberately uses string signatures against the deployed contracts (`settings()`, `multisigSettings()`, …). Keeping one style.
- **NatSpec for `InvalidMode`**: the protocol standard (`packages/protocol/AGENTS.md`) explicitly says "No natspec comments for errors"; the section is already consistent — all three errors are bare.

---

davidtaikocha (approved):

packages/protocol/script/layer1/proposals/Proposal0020.s.sol
```diff
@@ -160,6 +163,7 @@ contract Proposal0020 is BuildDirectProposal {
             !readBool(L1.DAO_SIGNER_LIST, "isListed(address)", L1.SC_GUSTAVO_GONZALEZ),
             "new member already listed"
         );
+        check(L1.SC_GUSTAVO_GONZALEZ.code.length == 0, "new member is not an EOA");
```
ggonzalez94, line 166: should we check this explicitely?

ggonzalez94 (approved):
LGTM. Just left one small comment

packages/protocol/script/layer1/proposals/Proposal0020.s.sol
dantaik: If "this" is the EOA property — it is now checked explicitly, on this exact line (`code.length == 0`, added by this PR).

If you mean the *agent precondition* (that `SC_GUSTAVO_GONZALEZ` is currently the Taiko Labs seat's appointed agent): that one is deliberately not pinned in `checkBaseline`, because it's expected to change before execution — Taiko Labs may rotate at any point after creation, and the dryrun must stay green in both worlds. `simulatePreExecution` handles it conditionally (rotates only if an appointer is still set), and `checkPostState` asserts the invariant that actually matters (`appointerOf(SC_GUSTAVO_GONZALEZ) == address(0)` once listed). An explicit `appointerOf == SC_TAIKO_LABS` baseline check would fail the rehearsal after a legitimate early rotation. Happy to add it as a warning-style log instead in a follow-up if you'd like visibility without the hard failure.

---

dantaik:
**Independent review: approve** ✅

I re-verified this PR from scratch — deployed mainnet sources (Etherscan-verified), live chain state, and a clean toolchain reproduction (pnpm 9 + repo-pinned Foundry v1.4.2).

**Findings confirmed fixed**

- **F1**: the deployed `getProposal` returns `(bool, uint16 approvals, ProposalParameters, bytes, IDAO.Action[], OptimisticTokenVotingPlugin)` — six values. Live call against mainnet proposal 21: the corrected cast signature decodes cleanly; the old one fails with `buffer overrun`. ✅
- **F2**: `MODE=print P=0020` regenerates `Proposal0020.action.md` byte-identical to the committed file; all six To/Value/Data fields are unchanged from the base branch — only the instruction text changed. ✅
- **F3**: `MODE=bogus` now fails the script with `InvalidMode("bogus")` instead of logging and exiting 0. ✅
- **F4/F5**: the Etherscan-verified `EncryptionRegistry` at `0x2eFD…42d1` is identical to `taikoxyz/dao-contracts` main. `setOwnPublicKey` reverts `MustBeListed` for a non-listed agent, so `setPublicKey(owner, key)` is indeed the agent path; `removeUnused()` prunes only `accountList`, leaving `appointerOf`/agent mappings intact (stale agent addresses stay reserved via the `AlreadyAppointed` check). The new spec wording is accurate. ✅
- **F6**: confirmed against deployed sources — only Action 1 → Action 3 is contract-enforced (`removeSigners` reverts when `newLength < minSignerListLength`; 10−5=5 < 8), while `_updateMultisigSettings` only requires `minApprovals ≤ addresslistLength()`, which 3 and 4 satisfy at every intermediate size (9, 10, 5). Actions 4–5 are defensive placement, exactly as now documented. ✅
- **F7**: `checkBaseline` now asserts the new member is an EOA, and the new `checkPostState` assertion verifies every account still enumerated by the registry is a listed signer (naming the offending address). ✅

**Independent reproduction**

- Recomputed all six calldata blobs with `cast calldata` — byte-identical to the committed `Proposal0020.action.md`.
- `forge fmt --check` passes on both changed files with the pinned toolchain (note: Foundry 1.5.x flags the `readRaw` header, but that's a formatter-version artifact — the repo-pinned v1.4.2 is clean).
- `MODE=l1dryrun` on a fresh mainnet fork: green end-to-end — baseline passes against live state, proposal 22 created, agent rotation simulated, all six actions executed as the DAO, and every post-state assertion (including the new pruning check) passes.
- Custom errors now sit in the end-of-file `// Custom Errors` section per the protocol conventions. 👍

**Nits (non-blocking)**

1. The `console2.log("Error: Invalid mode…")` in `run()` is redundant with the `InvalidMode(mode)` revert.
2. The spec links to `dao-contracts/blob/main/.../EncryptionRegistry.sol` — consider pinning to a specific commit for long-term reproducibility.

No functional concerns; with respect to the proposal calldata this is a docs/tooling-only change. Ship it.

PR #22017, opened 2026-08-13
fix(relayer): handle nil rabbitmq close notifications in subscribe

questfever:
## Summary

Handle nil RabbitMQ close notifications in `Subscribe`.

`amqp091-go` closes `NotifyClose` channels on graceful shutdown without sending an error first. Reading from the closed channel can therefore return `nil`, but `Subscribe` still dereferenced the value with `err.Error()` for both connection and channel close notifications.

This mirrors the nil-safe handling that was added to `Notify` in #15861, and applies the same protection to the remaining `Subscribe` paths.

## Verification

- `go test ./pkg/queue/rabbitmq`
- `go test ./pkg/queue/...`

### DIFF
--- a/packages/relayer/pkg/queue/rabbitmq/queue.go
+++ b/packages/relayer/pkg/queue/rabbitmq/queue.go
@@ -325,6 +325,14 @@ func (r *RabbitMQ) Nack(ctx context.Context, msg queue.Message, requeue bool) er
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


davidtaikocha (approved):

davidtaikocha:
Closed by https://github.com/taikoxyz/taiko-mono/actions/runs/31717017528/job/95096139043?pr=22017

PR #22018, opened 2026-08-14
docs(protocol): auction-based permissionless preconf design research

dantaik:
## What

Adds two research documents proposing an **auction-based, permissionless** preconfer/proposer mechanism for Taiko Alethia, as an alternative to the URC-based permissionless-preconf stack (parked on `permissionless-preconf`):

- `packages/protocol/docs/auction_based_permissionless_preconf.md` — the main design proposal ("Preconfer Auction")
- `packages/protocol/docs/urc_research_report.md` — companion deep-dive on eth-fabric/urc @ `132bc79`

## Why

- The URC path is blocked: not production-ready (see #22012), and its consensus-layer dependencies (validator lookahead, RLP header parsing, proposer commitments) are exactly the parts Ethereum is reshaping (ePBS/Glamsterdam H2 2026, FOCIL/Hegotá late 2026, SSLE ~2029).
- The proposal: an L1 contract runs a **perpetual standing-bid auction** for the exclusive right to propose/preconfirm. The top ETH bid is the winner **until outbid, quit, or expiry**; every transition takes effect **two epochs after placement**, so the current and next epoch's assignments are always final (≥ 1-epoch handover notice), and the bonded runner-up inherits automatically (empty list → total permissionless mode). Bids are a **per-epoch rate** charged lazily from a prepaid ETH balance — zero required transactions in steady state, no keeper.

## Design summary

- **Liveness #1:** winner posts a **TAIKO liveness bond**, slashed on stall via a checker-maintained, tx-atomic stall clock; a backup ladder (bonded runner-up → any bonded operator → anyone → forced inclusion) guarantees the chain never stalls. Stall slashes settle through a 1-epoch refute window where refutation must **disprove the recorded gap** (canonical proposal preimage against the Inbox ring-buffer hash) — reorg-robust and immune to stall-then-wake evasion.
- **Operator-signed blocks (S1)** enable cheap equivocation / malformed-block slashing; promise breaks (P1) and promise equivocation (P2) complete the fault catalogue.
- **Consensus-independent:** liveness faults provable from L1 state only — no beacon lookahead, no consensus structures; ePBS/FOCIL/SSLE-safe by construction (§4.7, §3.3).
- **O(1) read model** (§4.12): lazy per-epoch assignment cache, `effectiveEpoch = placedEpoch + 2` on every transition, current + next epochs always final, deeper reads provisional.
- Reuses in-repo primitives: `IProposerChecker`, `LibBonds`/bond-ledger pattern, forced inclusion, `ProverAuction` mechanics (worktree), the p2p preconf network.
- **Migration path:** enforce the existing (unenforced) `permissionlessInclusionMultiplier` escape hatch → shadow-mode auction on Hoodi → switch `proposerChecker` → full permissionless.

## Verification

- All cited config values verified against current code (`MainnetInbox`, `LibBonds`, `LibForcedInclusion`, `ProverAuction`, deployment logs, kimi-k3 audit I-01).
- Two design-review rounds integrated: an adversarial review (reorg-robust stall evidence, real Inbox escape-hatch enforcement, bond escrow + sizing) and a second review pass (perpetual-auction frequency, refute-rule fix, payment model, hot-standby duty, O(1) read model).
- Documentation only; no code or protocol behavior changes.

🤖 Research and drafting performed with AI assistance; findings verified against the repository.


dantaik:
Design review of `auction_based_permissionless_preconf.md` @ d5e6c1c against four target properties: (1) minimal auction frequency, (2) all liveness faults provable + slashable on L1 given signed L2 blocks, (3) O(1) reads of current/next-n epoch winners with `address(0)` when unassigned, (4) adequate join/quit time. Repo facts cited below were re-verified against current `main` (`IProposerChecker`, `Inbox._buildProposal`, `PreconfWhitelist`).

| # | Property | Verdict |
|---|---|---|
| 1 | Auction frequency as low as possible | ✅ Achieved by the perpetual standing-bid redesign — but §4.9 contradicts it, and the bid **payment model is unspecified** |
| 2 | Liveness faults provable + slashable on L1 | ⚠️ Core stall fault: yes, cleanly. **Critical flaw in the refute rule** makes every stall slash evadable as written |
| 3 | O(1) winner reads, `address(0)` fallback | ❌ Not specified — no getters, no read model, and the determinism horizon for "next n epochs" is only 48 s |
| 4 | Enough time to join and quit | ✅ Mostly — worst-case handover notice is 48 s; one parameter change fixes this and #3 together |

---

### 1. Auction frequency — ✅ with cleanups

The perpetual standing-bid design (top bid holds the right until outbid / quit / expiry / slash, handover quantized to epoch boundaries) is the correct answer: steady-state auction activity is **zero** — a transaction occurs only when someone actually wants to enter, displace, or leave. This is strictly better than the per-epoch auction of the earlier draft (13,500 auctions/yr) and can't be improved on without giving up contestability.

Three follow-ups:

- **§4.9 is stale and contradicts the redesign.** It still says "Per epoch: ~1-10 auction bids (commit + reveal…), one winner settlement". Commit-reveal was explicitly dropped (§4.3: sealed bids don't apply to a perpetual list), and a mandatory per-epoch "winner settlement" transaction would reintroduce per-epoch frequency *plus a keeper-liveness dependency*. The design should explicitly commit to **lazy assignment resolution**: winner-of-epoch must be derivable from stored state by pure computation; any storage normalization piggybacks on the next `bid`/`propose` call. No transaction may ever be *required* at an epoch boundary.
- **The bid payment model is unspecified**, and it determines whether per-epoch settlement sneaks back in. "The displaced winner keeps the current epoch (their bids for past epochs are already spent/settled)" implies the bid is an **ETH-per-epoch rate**, charged per epoch served — but nothing says how it's charged. Recommendation: define the bid as a rate, require prepaid ETH escrow covering ≥ k epochs, charge lazily (epochs-served × rate, settled on displacement/quit/expiry — O(1) multiplication), and auto-evict when the escrow can no longer cover the next epoch (this adds one more "effective-from" transition the read model in #3 must handle).
- **`tenureMax` expiry is the only mandatory periodic auction event left** (forced re-bid, 1/day). Its stated job — dead-bidder cleanup — needs only a heartbeat: make it a cheap incumbent-only *renewal* that changes nothing when exercised, rather than a lapse + re-rank + re-bid cycle. Also spec the quiet-expiry case: if the tenure lapses with no transaction at all, the lazy read model must still resolve "runner-up inherited at boundary B" in O(1).

(Also: the PR description still describes the old design — "sealed-bid first-price, commit-reveal", "one L1 epoch two epochs ahead". Worth updating to match d5e6c1c.)

### 2. L1-provable liveness — ⚠️ one critical fix needed

**What works.** The stall-fault evidence design is the strongest part of the doc: `lastProposalAt` written inside `checkProposer` is tx-atomic with the proposal, so evidence reorgs together with the chain; grace periods absorb shallow reorgs; fallback fires immediately while punishment is deferred. I verified this is compatible with the current interface: `IProposerChecker.checkProposer` is **not** `view` (the whitelist implementation narrows it to `view`, but the interface allows state writes), and the Inbox calls it unconditionally in `_buildProposal` — so the checker can legally maintain the stall clock with no Inbox change beyond the planned try/catch. The S1 class correctly exploits the signed-blocks assumption: equivocation and cheaply-checkable malformedness become slashable with one `ecrecover` + property check, and the EIP-712 domain / per-fault digest / bounded-slasher rules inherited from the URC post-mortem are the right hygiene.

**Critical — the refute rule as written makes stall slashing free to evade.** §4.4 says the escrow is released "if the winner re-proposes within the window **or** the fault condition no longer holds against the canonical chain at settlement". The first disjunct is the bug. Strategy: stall from epoch start; at +`STALL_GRACE` the runner-up proposes and escrows your slash; wake up, propose once during the 1-epoch refute window; escrow released. Cost: nothing. Repeatable every epoch you win — a guaranteed 48 s outage per epoch, provable but never punishable, and the challenger reward never settles (which also unfunds the runner-up's incentive to watch). Fix: refutation must **disprove the recorded gap, not demonstrate later activity** — release the escrow only on exhibiting a *canonical* proposal with `proposer == winner` and `timestamp` inside the recorded fault window. This is cheap to verify on-chain: proposals are hash-stored in the ring buffer and the `Proposal` struct contains `proposer` and `timestamp`, so the refuter presents the preimage against the stored hash — O(1). A reorged-out winner can always produce this (the reorg that restored their proposal is exactly what makes it canonical again); a genuinely silent winner never can. The "fault condition no longer holds" disjunct alone (second clause) is the correct semantics — the doc should delete the first.

**Boundary honesty — say what is *not* provable.** With signed blocks, the provable set is: stall (L1), window violation (L2), signed equivocation / malformed block (S1), broken signed promise (P1), unprovable proposal (V1). What remains structurally unprovable on L1 is **heartbeat censorship**: a winner who proposes a minimal/empty proposal every &lt;`STALL_GRACE`, consumes due forced inclusions, but serves no preconfs and excludes the L2 mempool. L1 cannot observe the L2 mempool or the absence of p2p messages — this is priced (bond ≥ 2× epoch revenue, §4.8) and floored (forced inclusion), not proved. That is a defensible position, but the doc should state it as an explicit non-goal in §4.4 so nobody reads "all liveness faults provable" as covering service quality. (The real fix is the sketched v2: make promise-root freshness an obligation checked in the proven state.)

Two smaller gaps:

- **Promise equivocation is missing from the fault table.** Two conflicting signed promises for the same `(epoch, seqNo)` is directly provable from the two signatures (same machinery as S1) and cheaper to punish than making the losing user wait to prove P1 non-inclusion. Add it as an S-class fault.
- **P1 v1 refutation evidence is unspecified.** "Winner may refute by paying" leaves the honest-winner path undefined against false claims. Concrete: refutation = Merkle inclusion proof of the promised tx against a *finalized* proposal's derived block within the promised range; the promise format must therefore bind a deadline/block range so late inclusion doesn't count.

Also worth one clarifying sentence in §4.5: since `lastProposalAt` tracks *any* proposer, once the runner-up takes over, the absent winner's continued silence is not separately punishable (the chain-level clock keeps resetting) — i.e., at most one stall slash per winner per epoch is the intended semantics. That's fine (capped at `livenessBond` anyway per §4.11), but make it explicit, including whether a slashed winner regains exclusivity if they return mid-epoch.

### 3. O(1) epoch-winner reads — ❌ needs a section

The doc never specs the read interface. Today's `PreconfWhitelist` already gives clients `getOperatorForCurrentEpoch()` / `getOperatorForNextEpoch()` with `address(0)` when the set is empty; the auction doc mentions getters only in passing (§4.6 "checker getter → sequencer address"). Three things need to be designed, not just implemented:

- **Getter spec.** `getProposerFor(epoch)` (plus sequencer + block-signing key variants) returning `address(0)` for unassigned epochs — which the design already has as a first-class state (§4.3 empty-list → permissionless mode), so the convention falls out naturally. Note the *checker* semantics differ from the whitelist's: `PreconfWhitelist.checkProposer` reverts when the operator is `address(0)`; the auction checker must instead **not revert** in unassigned mode (rung-2/3 rules apply). Deterministic top-of-list selection also removes the whitelist's `getBeaconBlockRootAtOrAfter` staticcall loop — which today is not even O(1) (it iterates over skipped slots) — so the auction getters can genuinely beat the incumbent on gas.
- **O(1) read model under lazy resolution.** With no keeper transaction (per #1), `getProposerFor(current + k)` must resolve from a constant number of SLOADs with no list iteration. This is achievable because boundary quantization guarantees a bounded number of pending transitions: keep an explicit assignment state machine — `(incumbent, rate, tenureEndEpoch, heir, heirEffectiveEpoch)` + ranked-list head — where every event (outbid, quit, expiry, ejection, escrow depletion) writes an *effective-from epoch* rather than mutating the assignment. Getters branch on epoch vs. these fields; list mutations pay their O(≤16) cost at write time. The doc should state the invariant outright: **reads are O(1) — constant SLOADs, no loops; writes pay the sorting cost.**
- **The determinism horizon is currently 48 s — decide whether that meets the "next n epochs" requirement.** With `freeze(E) = end(E) − 4 slots`, the winner of E+1 is *final* only during the last 48 s of E; before that, and for all E+2…E+n, any answer is provisional (current top of list, displaceable by a later bid). If clients need committed answers n epochs ahead, introduce `HANDOVER_NOTICE` = number of epochs between bid cutoff and control taking effect. Recommendation: **1 full epoch** (bids placed during E take effect at E+2), which makes `current` and `next` always final — exactly matching the two getters clients already consume — while deeper lookups return the provisional assignment plus a `finalized` flag. The cost is displacement latency of up to 2 epochs (~13 min), which is irrelevant next to `tenureMax` = 1 day; auction frequency is unaffected.

### 4. Join/quit time — ✅ with two adjustments

- **Join** is self-paced: deposit bond + bid, effective next boundary (or the one after the freeze); a bidder who wants more boot time simply bids earlier, and the incumbent cannot block entry (min-increment 5% + reserve floor + expiry keep the list contestable). Sound.
- **Quit** is well-designed: effective next boundary, must serve until then, going dark after quitting = stall slash — so clean handover is the only cheap exit, and quit notice is bounded at ≤ 1 epoch. Capital exit (escrow released after tenure end + refute window + 1-week withdrawal delay) matches mainnet norms and keeps slashing collateral available for exactly the window where faults from the final epoch can still settle. ✓
- **Soft spot: the worst-case handover notice is 48 s, and rung-1 takeover is 0 s.** When the winner quits/expires/is ejected, the runner-up inherits with as little as one freeze window of notice — and mid-epoch failover is immediate. That is coherent only if **listing itself is defined as an obligation**: standing on the list = hot-standby rung-1 duty at all times, bond at risk. The doc implies this but should state it as an explicit term of joining, because it changes what "enough time to join" means — you don't need time *after* winning, you must be ready *before* bidding. Raising `BID_FREEZE`/`HANDOVER_NOTICE` to a full epoch (see #3) also lifts the clean-displacement notice floor from 48 s to ≥ 384 s for the network (wallets/RPCs re-pointing to the new winner's keys), leaving only genuine failover on the fast path.
- Minor: spec the runner-up's escrow timing — a listed bidder who never actually served shouldn't inherit the winner's `tenure + refute window` escrow tail, just the plain withdrawal delay.

### Cross-cutting nits

- §4.7 claims "no `SLOT_TIME` hard-codes… only seconds-based grace periods", but §4.10 specs `BID_FREEZE` and the three graces in slots. Spec every parameter in seconds and define the epoch itself as a fixed-length timestamp window from a genesis anchor, so the consensus-independence claim survives slot-time changes (EIP-7782).
- §4.4's per-epoch slash cap (`livenessBond`, §4.11) should be stated once, normatively, covering L1+S1+V1 together — right now it lives in a scope note.

Overall: the perpetual standing-bid + epoch-quantized handover structure is the right shape for all four properties. The one thing I'd treat as a blocker before any spec work is the refute-rule fix in #2; #3 needs a new section rather than a tweak; #1 and #4 are wording/parameter cleanups.

---

dantaik:
Superseded by #22019, which now includes the design docs together with the implementation and tests.

PR #22035, opened 2026-08-21
feat(taiko-client-rs): tolerate non-archive L1 endpoints in finalized core-state reads

davidtaikocha:
## Problem

Path-scheme geth (the default full-node configuration) serves only ~128 recent blocks of state. The rust driver makes two finalized-block state reads during startup, and both were **fatal**:

1. **Beacon sync target** — `finalized_sync_target()` calls `Inbox.getCoreState()` at the `finalized` block tag (`sync/beacon.rs`). The existing fallback only matched geth's pre-finality error (`finalized block not found`), and the fail-fast gate makes any other error fatal before the first success.
2. **Event-sync finalized snapshot** — `try_finalized_l1_snapshot()` calls `getCoreState()` at the finalized block **hash** (`sync/event.rs`) and propagated any error out of the stage.

When L1 finality lags more than ~128 blocks (≈25.6 min) behind head, both reads fail on a non-archive endpoint with:

```
historical state 0x… is not available
```

Failure chain on cold start during such a finality delay:

```
finality gap > 128 blocks
→ driver pod (re)starts
→ getCoreState at finalized fails with historical-state error
→ error is fatal before first success (retryable_after_first_success)
→ sync pipeline exits → container crash-loops until finality recovers
```

A live retest against a mainnet path-scheme full geth confirmed the boundary: the call succeeds at `head-128` and fails at `head-129`. Both production rust-driver deployments are affected: `whitelist-preconfirmation-driver` embeds the full `SyncPipeline` (`PreconfIngressSync::start`), and the plain `driver` subcommand runs it directly.

### 1. Error matcher (`sync/mod.rs`)

`is_historical_state_unavailable` matches geth's `historical state <root> is not available`. geth >= 1.17 collapses every history-reader failure — state history disabled, history index still backfilling, root outside the `--history.state` window — into that single message, so one substring covers the class.

Deliberately **not** matched: a bare `missing trie node`. On geth >= 1.16.3 an old-block `eth_call` never surfaces it (the HistoricState fallback masks it), so today it means an incomplete or corrupt state DB and must stay on the fatal path. A negative test pins this.

### 2. Beacon sync target (`sync/beacon.rs`)

On a historical-state error, re-read `getCoreState()` at the **latest** block with a `warn`. The pre-existing pre-finality fallback does the same thing silently, which is correct — it is normal devnet behavior, not a degradation. Unrelated errors stay fatal, and errors from the fallback read itself stay fatal.

The target is a proof-finalized proposal recorded on L1 either way; only the read loses its reorg-proof anchoring during the degraded window. The checkpoint import still advertises `head = safe = finalized`, unchanged from today — see *Considered and dropped*.

### 3. Event-sync finalized snapshot (`sync/event.rs`)

The finalized **header** read needs no state and still succeeds on a non-archive endpoint, so the snapshot survives with only the state-derived field dropped:

```rust
/// Proposal id considered finalized-safe at this snapshot, or `None` when the endpoint
/// could not serve state at the finalized block.
finalized_safe_proposal_id: Option<u64>,
```

The startup decision then has three states, named so they cannot be collapsed again:

```rust
enum FinalizedClamp {
    Bounded(u64),   // clamp readable        -> target = min(resume, safe)
    Unknown,        // state unreadable      -> target = resume (unclamped)
    Unfinalized,    // chain never finalized -> target = 0 (replay from activation)
}
```

`Unknown` and `Unfinalized` both mean "no clamp value" but call for **opposite** responses, and this is the load-bearing part of the change. Mapping the unreadable case to `0` like the pre-finality case routes a synced node into a replay from the inbox activation block. That replay cannot complete against a non-archive **L2** — re-deriving an old proposal needs L2 state at its parent — and those failures retry against an unbounded backoff, so it trades a self-healing crash-loop for a driver that is up, never advances, and does not recover when L1 finality does. Only a chain that has never finalized replays from activation, where activation is near by construction.

The block anchor is kept for the same reason: the reconnect path then rewinds to the true finalized anchor instead of the far wider startup anchor.

## Considered and dropped

An earlier revision withheld the engine's `finalized` marking for latest-read beacon targets (`head`/`safe` only), on the theory that a finalized marking would block the rewind event sync needs after an L1 reorg. Dropped after checking the pinned reth (`f2eecc6`): there is no such guard. alethia-reth runs as OpStack, so the driver can reorg at will and a reorg below `finalized` only bumps a metric; a zero finalized hash is additionally a no-op there, since reth keeps the persisted value. The flag therefore changed nothing on the restart path it was written for, and only left fresh checkpoint nodes without a finalized marking. The derivation path already advertises finalized from `lastFinalizedProposalId` read at each proposal's own non-final L1 block, so the checkpoint import advertising `head`/`safe`/`finalized` matches the production path.

## Intentionally unchanged

Already non-fatal — they degrade with a `warn` on any error: the per-proposal finalized-hint lookup in the shasta derivation pipeline, and the post-reorg `head_l1_origin` reset.

## Tests

TDD: the two historical-state tests were written first and failed with the propagated error before the fix. 10 new tests.

`sync::beacon` — new test module, with a mocked L1 asserter and a `BeaconSyncer` test constructor:

- falls back to the latest core state on a historical-state error
- falls back to the latest core state before first L1 finality
- reads the finalized core state on the happy path
- surfaces unrelated RPC errors, asserting on the **original** message so the test cannot pass against an exhausted mock queue

`sync::event`:

- snapshot resolves both the block anchor and the clamp on the happy path
- keeps the block anchor with `finalized_safe_proposal_id: None` when state is unreadable
- `None` before first L1 finality (previously untested)
- unrelated core-state errors stay fatal
- `unreadable_finalized_state_keeps_resume_target_unclamped` pins the `Unknown` clamp against a non-zero resume

`sync`: matcher unit test, including the negative `missing trie node` case.

The three existing `resolve_target_with_optional_finalization` tests were updated for the new `FinalizedClamp` argument.

`cargo test -p driver --lib`: 136 passed, 0 failed. Note that no CI has run on this PR: every job in `taiko-client-rs--test.yml` gates on `draft == false`.

## Docs

WLP-INV-001 in `docs/agents/whitelist-preconfirmation-invariants.md` and the anchor row in `docs/agents/reference-map.md`, per that catalog's contract that behavior changes update the invariant text in the same change.

## Context

Found while reviewing taikoxyz/k8s-configs#1543 (moving L1 consumers from the archive pair onto path-scheme full geth). That rollout ultimately landed with `--gcmode archive` on the L1 endpoint, so the crash-loop is already mitigated in production. This PR is defence-in-depth: with it the rust driver no longer *requires* an archive/indexed L1 endpoint to survive a cold start during an L1 finality delay.


davidtaikocha:
Closing in favor of a replacement that preserves the original L1-finalized state boundary. The new implementation will keep the sync stages alive and retry finalized-state reads when a non-archive endpoint reports historical state unavailable, instead of falling back to latest or starting from an unclamped resume target. A superseding PR will follow.

PR #22056, opened 2026-08-25
fix(taiko-client): fail closed during preconf catch-up

davidtaikocha:
## Summary

- advance `highestUnsafeL2PayloadBlockID` when a validated preconfirmation envelope is cached, so `/status` reflects the highest payload known to the driver even before import
- skip missing-ancestor requests while the preconfirmation request topic has no connected peers, without consuming the request de-duplication entry
- keep the existing `/status` schema and Catalyst consumer unchanged

## Why

A restarted follower can receive and cache payloads above its execution-engine head while `highestUnsafeL2PayloadBlockID` remains pinned to that stale head. A consumer comparing those two stale values can incorrectly treat the node as synced and allow sequencing from the wrong parent.

This change makes the existing status check fail closed during that backlog. It also avoids issuing the one suppressed backfill request before the request-topic mesh has any peers.

## Scope

This fixes audit problems #1 and #3. It intentionally does not add timeout/backoff retries for a request published to an existing peer but never answered; that is audit problem #2 and should remain a separate change.

## Tests

- `go test -race ./packages/taiko-client/driver/preconf_blocks -run '^(TestCacheTestSuite|TestStatusReportsHighestCachedPayload|TestImportMissingAncientsDoesNotConsumeRequestWithoutTopicPeers|TestHasPreconfBlockRequestPeers)$' -count=1 -timeout=90s`
- `go vet ./packages/taiko-client/driver/preconf_blocks`
- `git diff --check`

### DIFF
--- a/packages/taiko-client/driver/preconf_blocks/server.go
+++ b/packages/taiko-client/driver/preconf_blocks/server.go
@@ -64,6 +64,10 @@ type preconfBlockChainSyncer interface {
 	InsertPreconfBlocksFromEnvelopes(context.Context, []*preconf.Envelope, bool) ([]*types.Header, error)
 }
 
+type gossipSubTopicPeerLister interface {
+	ListPeers(string) []peer.ID
+}
+
 // @title Taiko Preconfirmation Block Server API
 // @version 1.0
 // @termsOfService http://swagger.io/terms/
@@ -82,8 +86,9 @@ type PreconfBlockAPIServer struct {
 	anchorValidator               *validator.AnchorTxValidator
 	highestUnsafeL2PayloadBlockID uint64
 	// P2P network for preconfirmation block propagation
-	p2pNode   *p2p.NodeP2P
-	p2pSigner p2p.Signer
+	p2pNode             *p2p.NodeP2P
+	p2pSigner           p2p.Signer
+	gossipSubTopicPeers gossipSubTopicPeerLister
 	// WebSocket server for preconfirmation block notifications
 	ws *webSocketSever
 	// Lookahead information for the current and next operator
@@ -179,6 +184,7 @@ func New(
 // SetP2PNode sets the P2P node for the preconfirmation block server.
 func (s *PreconfBlockAPIServer) SetP2PNode(p2pNode *p2p.NodeP2P) {
 	s.p2pNode = p2pNode
+	s.gossipSubTopicPeers = p2pNode.GossipSub()
 }
 
 // SetP2PSigner sets the P2P signer for the preconfirmation block server.
@@ -783,6 +789,19 @@ func (s *PreconfBlockAPIServer) ImportMissingAncientsFromCache(
 			// If the parent payload is not found in the cache and chain is not syncing,
 			// we publish a request to the P2P network.
 			if !s.blockRequestsCache.Contains(currentPayload.Payload.ParentHash) {
+				if !s.hasPreconfBlockRequestPeers() {
+					log.Info(
+						"No peers on preconfirmation block request topic, skip publishing L2Request",
+						"blockID", parentNum,
+						"hash", currentPayload.Payload.ParentHash.Hex(),
+					)
+					return fmt.Errorf(
+						"failed to find parent payload in the cache, number %d, hash %s",
+						currentPayload.Payload.BlockNumber-1,
+						currentPayload.Payload.ParentHash.Hex(),
+					)
+				}
+
 				progress, err := s.rpc.L2ExecutionEngineSyncProgress(ctx)
 				if err != nil {
 					return fmt.Errorf("failed to get L2 execution engine sync progress: %w", err)
@@ -881,6 +900,15 @@ func (s *PreconfBlockAPIServer) ImportMissingAncientsFromCache(
 	metrics.DriverImportedPreconBlocksFromCacheCounter.Add(float64(len(payloadsToImport)))
 
 	return nil
+}
+
+func (s *PreconfBlockAPIServer) hasPreconfBlockRequestPeers() bool {
+	if s.gossipSubTopicPeers == nil || s.rpc == nil || s.rpc.L2 == nil || s.rpc.L2.ChainID == nil {
+		return false
+	}
+
+	topic := fmt.Sprintf("/taiko/%s/0/requestPreconfBlocks", s.rpc.L2.ChainID.String())
+	return len(s.gossipSubTopicPeers.ListPeers(topic)) > 0
 }
 
 // ImportChildBlocksFromCache tries to import the longest cached child envelopes from the cached payload queue.
@@ -1496,6 +1524,9 @@ func (s *PreconfBlockAPIServer) updateHighestUnsafeL2Payload(blockID uint64) {
 func (s *PreconfBlockAPIServer) tryPutEnvelopeIntoCache(msg *eth.ExecutionPayloadEnvelope, from peer.ID) {
 	id := uint64(msg.ExecutionPayload.BlockNumber)
 	h := msg.ExecutionPayload.BlockHash
+	if id > s.highestUnsafeL2PayloadBlockID {
+		s.updateHighestUnsafeL2Payload(id)
+	}
 	if s.envelopesCache.hasExact(id, h) {
 		return
 	}
--- a/packages/taiko-client/driver/preconf_blocks/server_test.go
+++ b/packages/taiko-client/driver/preconf_blocks/server_test.go
@@ -2,6 +2,7 @@ package preconfblocks
 
 import (
 	"context"
+	"encoding/json"
 	"math/big"
 	"net/http"
 	"net/http/httptest"
@@ -12,17 +13,29 @@ import (
 	"github.com/ethereum-optimism/optimism/op-service/eth"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/labstack/echo/v4"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/stretchr/testify/suite"
 
 	"github.com/taikoxyz/taiko-mono/packages/taiko-client/internal/testutils"
+	"github.com/taikoxyz/taiko-mono/packages/taiko-client/pkg/preconf"
 	"github.com/taikoxyz/taiko-mono/packages/taiko-client/pkg/rpc"
 )
 
 type PreconfBlockAPIServerTestSuite struct {
 	testutils.ClientTestSuite
 	s *PreconfBlockAPIServer
+}
+
+type stubTopicPeerLister struct {
+	peers     []peer.ID
+	lastTopic string
+}
+
+func (s *stubTopicPeerLister) ListPeers(topic string) []peer.ID {
+	s.lastTopic = topic
+	return s.peers
 }
 
 func (s *PreconfBlockAPIServerTestSuite) SetupTest() {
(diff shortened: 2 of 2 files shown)


davidtaikocha:
Reviewed against the incident write-up and the downstream consumer (`taikoxyz/Catalyst`). Verified on `0abe8f4`: `go build ./...`, `go vet`, `golangci-lint` and the three new tests are all clean, and I mutation-checked each new test.

**Verdict: the direction is right, but the invariant the fix depends on is not maintained — please don't merge this as the fix for Problem 1 yet.** The Problem 3 half is fine to keep, but it doesn't do what the PR body claims (see 4 below).

## What's right

- **Overloading `highestUnsafeL2PayloadBlockID` instead of adding a field is the correct call.** Catalyst's `TaikoStatus` has no `deny_unknown_fields`, so a new field would parse — but it wouldn't change the gate, which is a single equality check at `pacaya/src/node/operator/mod.rs:389` reused by `shasta/`, `realtime/` and `permissionless/`. Changing the field's *semantics* closes the hole with zero Catalyst change and zero release ordering. The documented contract at `api.go:333-335` already says "has received", so this makes the implementation match the doc.
- **It would have stopped the 2026-08-25 reorg.** During the 5m20s window the head was pinned at 10522943 with envelopes cached to 10523099, so the reported value would have been 10523099 and the gate would have refused ~3m50s before handover.
- **The zero-peer path deliberately does not consume the de-dup entry** (`server.go:792-803` returns before the `blockRequestsCache.Add`). That's the important detail and it's right — it's also a genuine partial mitigation of the declined Problem 2 for cold start.

## 1. Blocking — the new signal is erased on the exact path it exists to protect

`updateHighestUnsafeL2Payload` (`server.go:1505-1521`) **assigns**; it does not take a max. Four writers can lower it, and nothing ever re-derives the value from the envelope cache — which is the thing that actually holds the backlog. Two are reachable while a backlog is cached:

**(a) `server.go:943`, `ImportChildBlocksFromCache` — the tight one.**

```go
s.updateHighestUnsafeL2Payload(endBlockID)   // no `>` guard, unlike the sibling at :1468
```

`endBlockID` is the top of the **contiguous** run from `getChildren` (`queue.go:72-110` stops at the first hole). Anything cached above that hole is still unapplied, and the counter is driven down to `endBlockID` — which by construction equals the new EE head, so `/status == geth height` and Catalyst declares synced.

Incident-shaped instance: head 10522943, hole at 10522961, cache holds 10522983–10523099 (counter 10523099). Backfill closes 10522944–10522960, the walk imports them, `endBlockID = 10522960`, counter drops to 10522960 = EE head. **139 envelopes still cached and unapplied, and the node reports parity.** The `tryPutEnvelopeIntoCache` that follows at `:374` uses the same (now lower) number and does not restore it; only the next inbound gossip payload would — and at handover there is no next payload.

Two-line fix, and it's reorg-safe because `TryImportingPayload` already lowers for a reorg at `:1483` *before* calling this:

```go
if endBlockID > s.highestUnsafeL2PayloadBlockID {
    s.updateHighestUnsafeL2Payload(endBlockID)
}
```

**(b) `server.go:1285`, `recordLatestSeenProposal` — hard reset with no floor.**

`inserter.go:275` sets `PreconfChainReorged = true` unconditionally whenever L1 derivation actually inserts, i.e. exactly when the node is behind — the canonical short-circuit at `inserter.go:180-209` covers the healthy case and sends `false`. That reaches `server.go:1285`, which hard-assigns `highestUnsafeL2PayloadBlockID = proposal.LastBlockID`.

Since preconfirmations run ahead of L1 proposals by design, "derived tip below the gossip backlog" is the normal case, not a corner:

```
head 10522943, cache 10522944..10523099, counter 10523099   <- correct, fail-closed
L1 proposal derives 10522944..10523043
  head    -> 10523043
  counter := 10523043
Catalyst: 10523043 == 10523043 -> synced, while 56 validated envelopes sit unapplied
```

The window isn't short either: `ImportPendingBlocksFromCache` is one-shot, gated on `preconfImportsPending` (`chain_syncer.go:167`), so there is no periodic drain after derivation — it depends on the next inbound gossip, which fail-closed has just stopped.

**(c) `api.go:215` is the latch.** `// always update the highest unsafe L2 payload block ID` — not an entry point, but once (a) or (b) has opened the door and the node starts sequencing, every slot re-pins the counter to its own freshly built head, so the backlog signal never returns and the fork persists. That's the "rest of the network followed" step of the incident.

Suggestion: report `max(liveExecutionHead, envelopesCache.highestBlockID())` and drop the counter as the source of evidence — that removes (a), (b), item 2 below, and the pre-existing post-beacon-sync false negative in one change. If you'd rather keep the counter, make `updateHighestUnsafeL2Payload` monotonic and add an explicit `resetHighestUnsafeL2Payload` for the two genuine reorg sites, so lowering is opt-in and auditable.

## 2. Blocking — the counter is unbounded and never evicts; Catalyst responds by exiting the process

The envelope cache is a bounded 768-slot ring (`queue.go:16`), so a bad height ages out on its own. The counter has neither bound nor eviction. Nothing limits an inbound payload's block number from above: `ValidateExecutionPayload` (`server.go:951`) never looks at `BlockNumber`; `checkMessageBlockNumber` (`util.go:147`) is a lower bound only, and isn't even reached on the `!syncReady` (`:344`) and `IsSyncing` (`:355`) cache paths; the gossip validator rejects only `BlockNumber == 0`.

Gossip signature verification means this isn't an arbitrary-attacker vector — but one whitelist operator running a buggy build is enough. A single payload with an absurd height pins every receiving follower "behind" indefinitely, and Catalyst's `cancel_counter` (`mod.rs:305-313`) then calls `cancel_on_critical_error()` after `l2_slots_per_epoch/2` = 96 slots ≈ 192s, on every node regardless of role. Recovery needs the counter lowered, and both lowering paths are unreachable in that state — so: restart every driver. Worth a cheap sanity bound (e.g. ignore payloads more than `maxTrackedPayloads` above the current EE head for counter purposes).

## 3. Rust driver diverges — only Go nodes fail closed

`taiko-client-rs` fills the same JSON field from `reconcile_reported_head(l2_head)` (`api/service/status.rs:19`, `cache.rs:109-117`), where `l2_head` is `get_block_by_number(Latest)` — the **live EE head**, with a last-observed fallback. Catalyst compares that against its own head reading, so the gate is near-tautological there and a Rust node can never report itself behind. After this PR the same field means two different things depending on which driver a pod runs. If there's a sibling PR for the Rust side, please link it; otherwise this needs to land there too.

## 4. The mesh gate doesn't gate the mesh

`hasPreconfBlockRequestPeers` (`server.go:905`) uses `pubsub.PubSub.ListPeers`, which returns `p.topics[topic]` — peers that have **announced a SUBSCRIBE**, populated at connection handshake. It is unrelated to the gossipsub mesh (`gs.mesh[topic]`, GRAFTed peers); the router's own readiness test is `EnoughPeers` (`gossipsub.go:665`), which counts `len(gs.mesh[topic])` against `Dlo`.

The incident evidence settles it: the previous request on the same topic, 213 ms earlier from the same node, was received by 8 peers across both clusters — so `ListPeers` was comfortably non-zero when the failing publish went out into a 10-second-old mesh. **The gate as written would have passed that publish and the outcome would be unchanged.** It's still worth keeping for genuine cold start, but the PR body shouldn't claim it addresses the observed failure.

The real fix is the readiness option the write-up mentions: `PublishL2Request` (fork `op-node/p2p/gossip.go:1003`) currently passes no `PubOpt`; `pubsub.WithReadiness(pubsub.MinTopicSize(n))` blocks until the mesh is actually usable, which also addresses the root of Problem 2's "`nil` carries no delivery information".

## 5. A transient RPC error silently destroys the evidence

The counter advance lives in `tryPutEnvelopeIntoCache`, reachable only *after* `L2ExecutionEngineSyncProgress` (`server.go:349`), `checkMessageBlockNumber` (`:360`) and `TryImportingPayload` (`:366`) — each returns an error that drops the whole message, and gossipsub does not redeliver. A node that has just restarted is exactly when those RPCs are most likely to blip. One error on the highest payload and the node under-reports its backlog. Record into the cache / advance the counter right after `ValidateExecutionPayload` succeeds, before any fallible RPC.

## 6. Shipping Problem 1 without Problem 2

The write-up's own warning applies: fail-closed turns "sequences from a stale head" into "refuses to sequence", and Problem 2 is what turns a dark epoch back into a non-event. After the refusal nobody sequences, gossip stops, and `ImportMissingAncientsFromCache` is only re-driven by inbound gossip — meanwhile Catalyst exits after ~192s. The Problem 3 gate doesn't help, since it only decides whether to publish and is itself only reached from a gossip-driven call. The trade may well be right, but it should be a stated decision, and a retry ticker (a few seconds, gated on `seen > imported`) should land in the same release.

## Smaller things

- **Data race on the value that is now a safety signal.** `GetStatus` (`api.go:352`) takes `lookaheadMutex`; every writer holds `s.mutex` (distinct fields, `server.go:96` / `:111`). Pre-existing, but write frequency goes from per-import to per-envelope and the value is now the gate. `integration_test/entrypoint.sh:57` runs without `-race`, so CI won't see it. `atomic.Uint64` is preferable to taking `s.mutex` in `GetStatus`, which would put `/status` behind the gossip critical section.
- **`ListPeers` blocks on the pubsub event loop while `s.mutex` is held** (`pubsub.go:1374`, handler at `:622`). No deadlock — `notifySubs` drops to slow subscribers non-blockingly — but that's the concern: lengthening the critical section raises the odds of "subscriber too slow" drops, and a single dropped preconf message is what started this incident. Consider caching the count from the existing peer tick loop.
- **Topic literal duplicated with no cross-check.** `server.go:910` hardcodes what the fork's unexported `preconfBlocksRequestTopic` builds (`gossip.go:100-101`). It matches today — I diffed it. If the format ever drifts, `ListPeers` returns nil for an unknown topic, the gate is permanently closed and backfill requests stop entirely: a silent total loss of the recovery path. Best fix is a `PreconfBlocksRequestTopicPeers()` accessor in the fork alongside the existing `PreconfBlocksTopicV1Peers()` (`gossip.go:956-958`); at minimum pin the fork file:line in a comment.
- **Duplicated error return.** `server.go:798-802` is byte-identical to `:847-851`. `if !s.blockRequestsCache.Contains(h) && s.hasPreconfBlockRequestPeers() { … }` plus the log line collapses them.
- **Unmentioned control-flow change.** The gate precedes the `progress.IsSyncing() → return nil` at `:810`. `OnUnsafeL2Payload` already short-circuits on `IsSyncing`, but `OnUnsafeL2Response` doesn't — so for a beacon-syncing node with zero request-topic peers, an orphan response is now cached instead of dropped by a failing insert. That's an improvement; worth a line in the description.
- **Metric and log semantics.** `DriverHighestPreconfUnsafePayloadGauge` (`server.go:1520`) flips from "highest imported" to "highest seen", so dashboards comparing it to `DriverL2HeadHeightGauge` will show a standing gap. Every cached envelope also emits an INFO line at `:1507` (~117 extra lines in the incident's shape), as does the new zero-peer log at `:790`.

## Tests

Mutation-checked; the results are mixed:

- Reverting the `tryPutEnvelopeIntoCache` bump **fails** `TestStatusReportsHighestCachedPayload` correctly, and dropping the `>` guard also fails it. Good.
- Adding `blockRequestsCache.Add` before the zero-peer return **fails** `TestImportMissingAncients…` on the de-dup assertion, and changing the impl's topic to `/taiko/%s/1/…` fails it too. Good — but the topic assertion compares against the same literal the implementation builds, so it pins nothing about the *fork*, which is the drift that matters.
- **`TestHasPreconfBlockRequestPeers` is vacuous**: replacing the whole body of `hasPreconfBlockRequestPeers` with `return true` still passes. It only exercises the positive case.
- Deleting the gate entirely makes `TestImportMissingAncients…` **panic** at `:805` (`L2ExecutionEngineSyncProgress` on a stub client) rather than fail an assertion — it survives only by short-circuiting before the first real RPC, so move the gate and the test stops meaning anything.
- Not covered: the positive path (peers present → request published *and* de-dup entry consumed), so a permanently-closed gate is undetectable; the lowering paths in item 1; `/status` reporting behind when the head is below the cache. `TestStatusReportsHighestCachedPayload` also inserts a lower-numbered envelope and never asserts anything about it.
- Style: the three tests are standalone `func Test*` in a file that otherwise uses `PreconfBlockAPIServerTestSuite`. The deviation is justified (the suite needs a devnet, these don't) — worth a one-line comment saying so, otherwise it reads as an oversight.

## Suggested minimal set

1. `>` guard at `server.go:943` — the tightest hole, two lines.
2. Report `max(liveExecutionHead, envelopesCache.highestBlockID())` instead of a counter — removes items 1(b), 2, and the post-beacon-sync false negative.
3. Cache immediately after `ValidateExecutionPayload`, before any fallible RPC (item 5).
4. Align the Rust driver (item 3).
5. `WithReadiness` / exported topic accessor in the fork (item 4).
6. Retry ticker for Problem 2 in the same release (item 6).
7. `atomic.Uint64` for the counter.

PR #22057, opened 2026-08-25
feat(protocol): remove gas-schedule fragility (repricing headroom + transient storage)

dantaik:
## Summary

Audit of every gas-schedule assumption in `packages/protocol/contracts` against the Glamsterdam repricing candidates ([EIP-8038](https://eips.ethereum.org/EIPS/eip-8038) / [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) state-access repricing, calldata repricing proposals) plus the already-live [EIP-7623](https://eips.ethereum.org/EIPS/eip-7623) calldata floor (Pectra) — with the changes that are safe and beneficial today applied. The audit covers exposure on L1 and on Taiko L2 itself (as a type-1 rollup, Taiko is expected to adopt the same schedule in a subsequent L2 hard fork).

Cross-check against the [ethereum/repricing-impact](https://ethereum.github.io/repricing-impact/affected-contracts.html?schedule=eip-8038) dataset (2026-08-24, mainnet blocks 21,319,986–25,319,985): **no active Taiko L1 contract appears in the EIP-8038 "potentially broken" set** (all 188 addresses ever recorded in `deployments/mainnet-contract-logs-L1.md` verified against the per-address shards). The findings are therefore about *calibration* (constants drifting out of tune) and *cost*, not about anything breaking outright.

Reference deltas used below (EIP-8038 draft, subject to change while in peer review):

| Constant | Today | EIP-8038 | Note |
| --- | --- | --- | --- |
| Cold account access | 2,600 | 3,000 | +400 |
| Storage write (dirty-slot update component) | 2,800 | 10,000 | net cold update ≈ 5,000 → 12,100 |
| Account write (value transfer) | 6,700 | 9,000 | +2,300 |
| New account/create access | 7,000 | 12,000 | +5,000 |
| Fresh storage slot (create + write) | ≈ 22,100 | ≈ 22,000 | effectively unchanged |
| Access list: address / storage key | 2,400 / 1,900 | 2,900 / 2,000 | |
| EXTCODESIZE / EXTCODECOPY | — | +100 | extra warm-access read |

## Changes in this PR

[...]

PR #22058, opened 2026-08-26
refactor(protocol): unify L1 and L2 on the transient-storage reentrancy guard

dantaik:
## Summary

L1 and L2 now run the same EVM version (`foundry.toml` pins a Cancun-or-later `evm_version` for every profile), so the historical split — storage-based defaults in the shared contracts plus per-contract transient-storage overrides for the L1 mainnet deployments — no longer has a reason to exist. This PR makes the transient-storage (EIP-1153) implementation the single one and removes the wrapper contracts that existed only to carry the overrides.

## Changes

- **`EssentialContract._store/_loadReentryLock`** now keep the reentry lock in transient storage (same slot `LibFasterReentryLock` used) and are **no longer virtual** — one canonical reentrancy guard, structurally impossible to fork back into storage.
- **`Bridge._store/_loadContext`** move the call context to transient storage (same slot the `MainnetBridge` override used) and are no longer virtual.
- **`LibFasterReentryLock` is removed**, along with all per-contract overrides.
- **The shell wrappers `MainnetBridge`, `MainnetERC20Vault`, `MainnetERC721Vault` and `MainnetERC1155Vault` are deleted entirely** (contracts + generated `*_Layout.sol` + `gen-layouts.sh` entries — the base contracts in the shared layout list keep full coverage). Deploy scripts (`DeployProtocolOnL1`, `DeployHackRecoveryContracts`) construct `Bridge`/`ERC20Vault`/`ERC721Vault`/`ERC1155Vault` directly. The Taiko mainnet genesis balance invariant documented on `MainnetBridge` moves to `Bridge`. `MainnetInbox`, `DevnetInbox` and `MainnetDAOController` remain — they carry chain configuration or logic of their own.
- The `__reentry`/`__ctx` storage slots are retained for layout compatibility — **no storage layout change** (behavior on the L1 deployments is also unchanged: they already ran these exact transient slots via the overrides).

Transient-slot constants, independently recomputed (`cast keccak`):

| Constant | Derivation | Value |
| --- | --- | --- |
| `Bridge._CTX_SLOT` | `keccak256("bridge.ctx_slot")` | `0xe4ece82196de19aabe639620d7f716c433d1348f96ce727c9989a982dbadc2b9` ✅ |
| `EssentialContract._REENTRY_SLOT` | `keccak256("ownerUUPS.reentry_slot") + 1` | `keccak = …721a`, `+1 = 0xa5054f728453d3dbe953bdc43e4d0cb97e662ea32d7958190f3dc2da31d9721b` ✅ |

Both values are byte-identical to the ones the deployed `MainnetBridge`/`LibFasterReentryLock` have been running on L1.

## Effect

- L1 Shasta hot path (regenerated `snapshots/`): `propose_single` 71,083 → 65,988 (−7.2%), `prove_single` 85,315 → 82,220 (−3.6%).
- The L2 deployments (bridge `0x1670…0001`, vaults, signal service), which previously ran the storage-based defaults (~17k of lock/context storage writes per bridge message), pick up the savings at their next routine upgrade.

## Validation (foundry v1.4.2 per `.tool-versions`)

- `FOUNDRY_PROFILE=shared forge test`: 154/154 pass
- `FOUNDRY_PROFILE=layer1 forge test`: 374/374 pass
- `FOUNDRY_PROFILE=layer2 forge test`: 12/12 pass
- `forge build` clean on all profiles; `forge fmt` clean on all edited files
- Storage layouts unchanged; no test referenced the removed wrappers

Split out of #22057 (its sibling #22059 covers the gas-repricing headroom changes).

https://claude.ai/code/session_012xmvsjCrZyCJBVft82LjzJ


packages/protocol/contracts/shared/common/EssentialContract.sol
```diff
@@ -22,7 +28,8 @@ abstract contract EssentialContract is UUPSUpgradeable, Ownable2StepUpgradeable
 
     uint256[50] private __gapFromOldAddressResolver;
 
-    /// @dev Slot 1.
+    /// @dev Slot 1. __reentry is deprecated: the reentry lock lives in transient storage
+    /// (_REENTRY_SLOT); the storage slot is retained only for layout compatibility.
     uint8 internal __reentry;
```
dantaik: if it is deprecated, change it to "private"?

dantaik: Done in 1c3f4aa — `__reentry` is now `private`. Nothing reads or writes it any more; the solc storage layout (and the genesis-alloc generator, which addresses slots by layout name) is unaffected by the visibility change. All three test suites still pass.

---

dantaik:
Response to the DeepSeek review:

1. **"`context()` consumers previously saw the last stored `__ctx`"** — not observable: the storage implementation wrote the `_PLACEHOLDER` sentinel back after every invocation, and `context()` reverts with `B_INVALID_CONTEXT` for both the sentinel and the zero value, so an `eth_call` outside an invocation reverted before this change and still does. Behavior during an invocation is likewise identical. Now covered explicitly by the regression test below.

2. **EIP-1153 activation sequencing** — fair in the abstract; in practice Taiko Alethia has run a Cancun-or-later EVM since well before this change (`foundry.toml` pins `evm_version` per profile), and the L1 deployments already execute these exact tstore paths today via the former wrapper overrides.

3. **Regression test** — added in 87dfc18: asserts `context()` reverts outside an invocation, each `processMessage` call in the same transaction observes its own context, and the context is cleared again as soon as the invocation returns (per-invocation reset, not per-transaction).

4. **Layout regeneration** — the `*_Layout.sol` files document storage variables (name/type/slot), which this PR deliberately leaves unchanged; they were regenerated when the wrapper contracts were removed and carry no stale storage-based description. CI's generated-artifacts pass (4a7b284) confirms no drift.

5. **Snapshot trailing newline** — those files are written by forge's snapshot tooling; hand-editing them would just be reverted by the next regeneration/CI artifact commit.

---

dantaik (commented):
Reviewed exact head 87dfc1844591b22c1664bffc9a7c57a31b2fc16f against base 60e37a186c4975c7c47bc8017b7cc6a787eb6c1d.

No blocking findings.

I reviewed all 21 changed files, including EIP-1153 proxy/delegatecall semantics, exact transient-slot reuse, Bridge process/retry/recall context cleanup, storage-layout preservation, wrapper removal and deployment replacements, deleted-reference cleanup, snapshots, gas reports, and diff hygiene.

Validation: exact-head protocol and genesis CI are green. Focused offline Foundry coverage passed 70 unique tests across Bridge message paths, EssentialContract, shared vaults, Anchor, and DelegateController; the genesis profile also builds and the inspected Bridge layout retains the deprecated storage slots at their original locations. The local L1 profile could not compile fully offline because the exact Solady, Automata, Risc0, and SP1 sources were not cached; exact-head CI covers those suites.

Non-blocking test note: context_transient_lifecycle checks the externally visible lifecycle, but it would also pass the former storage-backed implementation. A future regression test could explicitly attempt callback reentrancy, assert REENTRANT_CALL and inNonReentrant during the callback, and assert the deprecated persistent lock/context slots remain unchanged. I do not see a production correctness issue in the implementation as submitted.

davidtaikocha (approved):

dantaik:
The suggested discriminating regression is implemented in f2d9008, going one step further than the lifecycle test: a `ReentrantProbe` target asserts `inNonReentrant()` is true during its callback, catches the reentrant `processMessage` and checks the revert selector is `REENTRANT_CALL`, and — the part the former storage-backed implementation could not satisfy — probes the deprecated persistent slots directly with `vm.load`, asserting the `__reentry` byte (slot 201, offset 0) and both `__ctx` slots (253–254) remain zero after processing. 12/12 in the suite.

---

PR #22061, opened 2026-08-26
feat(protocol): enable Osaka in genesis test runtime

davidtaikocha:
## Summary

- pin the Docker genesis node to Geth v1.17.2, matching the Geth version tracked by `packages/fork-diff`
- configure the node as post-Merge with Shanghai, Cancun, Prague, and Osaka active from genesis, including the required blob schedule
- remove obsolete Clique mining, account unlocking, and test key material
- add a live JSON-RPC probe that executes `TSTORE`/`TLOAD` and Osaka `CLZ` before the Foundry fork tests

## Context

Follow-up to the runtime compatibility finding on #22058. This branch is based directly on main and can merge independently.

The previous Geth genesis stopped at Berlin. Foundry executed forked tests in its locally configured EVM, so the Docker job could pass even though the actual node rejected newer opcodes. The live probe now verifies the real Geth runtime through Osaka.

## Validation

- Red case: the Osaka probe failed against the prior Cancun-only runtime with `invalid opcode: opcode 0x1e not defined`
- Green case: `pnpm --config.verify-deps-before-run=warn run genesis:test`
- `bash -n packages/protocol/test/genesis/genesis.test.sh packages/protocol/test/genesis/testnet/init_geth.sh`
- `docker compose -f packages/protocol/test/genesis/testnet/docker-compose.yml config --quiet`
- `git diff --check`


smtmfft (approved):
Reviewed the latest head a085ed0. The genesis test node is now post-Merge/Osaka-at-genesis, the old unlocked mining account path is removed, and the runtime probe exercises the target Osaka opcode path before Foundry runs.\n\nI also checked the automated comments: docker-compose only mounts the testnet directory now, not deleted keyfile/password paths, and genesis.test.sh runs with errexit so the Osaka probe is gating. CI is green for build-protocol and genesis-docker. No concerns from my side.

PR #22062, opened 2026-08-27
refactor: stream delimited configuration parsing

daixiheguu:
Replace two one-pass `strings.Split` loops with `strings.SplitSeq`:

strings.SplitSeq (introduced in Go 1.23)  returns a lazy sequence (strings.Seq), allowing gopher to iterate over tokens one by one without creating an intermediate slice.

It significantly reduces memory allocations and can improve performance for long strings.

More info: https://github.com/golang/go/issues/61901

- L2 execution-engine endpoint parsing in the client chain config;
- ignored message-hash parsing in the relayer indexer config.

Both callers consume tokens sequentially and do not need an intermediate slice, so this avoids unnecessary allocations while preserving order and empty-token behavior.

Tested with:

   ` go test ./packages/relayer/indexer ./packages/taiko-client/pkg/config`

### DIFF
--- a/packages/relayer/indexer/config.go
+++ b/packages/relayer/indexer/config.go
@@ -154,7 +154,7 @@ func parseIgnoredMsgHashes(value string) (map[common.Hash]struct{}, error) {
 
 	hashes := make(map[common.Hash]struct{})
 
-	for _, raw := range strings.Split(value, ",") {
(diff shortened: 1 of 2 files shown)


davidtaikocha:
Closed since don't see too many improvements here.

PR #22064, opened 2026-08-28
docs(protocol): Slot-Chain — preconfirmation protocol v2 design specification

dantaik:
Design and executable specification for Slot Chain V2, Taiko's seat-based perpetual reverse-auction preconfirmation protocol.

This PR is the sole owner of the design surface: normative LaTeX/PDF, executable Python models, golden vectors, adversarial model tests, and the reviewed multi-round contract implementation plan. Solidity and Foundry tests belong only in the stacked implementation PR.

## Current artifacts

- `packages/protocol/docs/preconfirmation-v2/slot-chain-spec.pdf`: 219-page reviewed design candidate.
- `packages/protocol/docs/preconfirmation-v2/tex/main.tex`: normative source.
- `commitment-model.py`: 764 golden vectors / 1,503 assertion sites.
- `lookahead-model.py`: 38 properties.
- `settlement-window-model.py`: 184 properties.
- `test-seat-market.py`: 114 adversarial tests.
- `test-economic-profile.py`: 38 adversarial tests.
- `test-settlement-window.py`: 277 adversarial tests.
- `docs/superpowers/plans/2026-09-01-slot-chain-v2-contract-implementation.md`: multi-round code/test implementation plan.

## Frozen architecture

- Seat-based, continuous/perpetual reverse auction with four bounded roster cells.
- Permissionless proof landing; aggregator seats are paid service roles, not proof-submission authority.
- Builder-independent forced-queue recovery and deterministic unsigned escape blocks.
- Proof-first, atomic and abortable version migration with exact legacy-resume and genesis journals.
- Fresh immutable V2 source/destination custody; no V1 Bridge/vault reuse and no relayer authority.
- Exact release/reclamation lifecycle, historical proof retention, and first/last builder/seat boundary handling.
- Disjoint build ownership: internal-only shared primitives are source-inline and checked by source/ABI/link invariants; deployables remain single-profile artifact-owned with deterministic creation/runtime/config/layout hashes.

## Latest terminal-lifecycle freeze

The schedule now derives a finite `lastManagedWindow` from the complete-window slot bound and the absolute genesis/evidence/reorg deadline. Admissions, reservations, seals, fork boundaries and reads are bounded to the managed interval. An exact STS1 Settlement view and permissionless O(1) `finalizeScheduleExpiryV1()` prevent cursor-backlog and post-`UINT64_MAX` liveness failures. Terminal normalization never narrows an unbounded wall-clock window, preserves the zero-leaf no-op, and terminal ACTIVE/LIABILITY generation release remains evidence-safe and independent of liability-ring capacity.

The PDF was rebuilt with Tectonic and all changed pages were visually inspected. The four executable model suites pass, including all 277 settlement adversarial tests. Independent architecture and model/spec challenge passes found no remaining Critical or Major defect in this round.

## Status

This is an audited design candidate and an implementation freeze, not a production release claim. The additive contract suite is being implemented and tested in the stacked PR without selecting V2 on any production path. Production readiness still requires the external gates retained in the specification and plan: real circuits/keys, measured proof and activation gas, calibrated economics, client/fork support, migration and archive-loss drills, multi-client testnet soak, independent conformance reproduction, and independent audits.


dantaik:
## Design review: Slot-Chain v2 (`packages/protocol/docs/preconfirmation-v2`)

**Scope.** I reviewed the full specification (`tex/main.tex`, 12,974 lines, v2.26 at `5387425`) and used the Python models to check the design. Method: the spec was split into 27 slices (every section plus cross-cutting passes for privileges, CL dependence, governance inaction, preconfirmation guarantees, model quality and internal consistency), each reviewed independently; 180 raw findings were merged into 120 candidates, and every candidate was then attacked by three independent verifiers (factual re-read of the cited lines, search of the whole spec/models for an existing mitigation, and severity calibration). 21 candidates were refuted and dropped; 99 survived. This comment reports the ones that matter, with the rest collapsed at the end. Line numbers refer to `tex/main.tex`.

**Requirements I checked against:** (R1) no party has special privileges beyond contract upgradability; (R2) no reliance on Ethereum's consensus layer, with EIP-4788 beacon-root reads accepted; (R3) the design is sound, complete and bug-free.

### Bottom line

- **R1 is not met.** Five distinct capabilities go beyond "upgrade the contracts", and in two of them governance *inaction* degrades or halts the chain: the finite fork-verifier horizon (C1), the 7-day migration-arm freeze (H5), the V2 bridge pauser (H6), the loss of the V1 SignalService checkpoint feed at cutover (H7), and the unspecified issuer of the builder lease token (M18).
- **R2 is met in spirit** (beacon root only, confined to `ScheduleOracle`), but the way it is used, fork-digest-routed SSZ verifier rows with a finite reviewed horizon, is what creates C1 and H12. An EL-only snapshot would remove that entire surface; see the design note under C1.
- **R3 is not met yet.** The core settlement machine, forced queue, registry lifecycle and escape lane are carefully constructed and most of the adversarial cases I tried are handled. But the preconfirmation guarantee is much weaker than the README states (H4), two normal-lane failure modes were reproduced in the executable model (H1, H2), the data-availability layer cannot carry the advertised 1-second cadence (H8, H9, H10), and the spec and models disagree on consensus-critical rules in at least six places (H1, M7, M8, M11, M13, F66), which §1 says blocks deployment.

I would not call this an implementation-freeze candidate yet. The Critical and High items below need design changes, not wording changes.

### Critical

**C1. Schedule liveness depends on recurring DAO action; a single missed renewal is irreversible.** Sealing a window requires `targetSlot(W)+64 < latestLastParentSlotExclusive` (L2222-2224), the latest fork-verifier row "retains its finite nonzero last slot" bounded by the next unreviewed consensus-schema boundary (L2483-2489), and an unsealed window is "permanently all-VACANT" (L2213-2214). Only the immutable DAO proposer can queue the three fork-change operations (L6056-6057), each with a 7-day delay and a 1-day execution window (L6104-6108). The spec itself says: "If governance does not extend it in time, targets ... fail closed and the ordinary post-deadline unsigned/VACANT escape remains available" (L2490-2493). Worse, once the wall clock passes horizon-8 windows, none of APPEND (needs `new.first = old.last` and a future protected target, L2423-2433), EXTEND (needs the old horizon above the future target, L2429-2431), REPLACE_PENDING (L2513-2516) or SPLIT (L2551-2554) can execute, root succession is "deliberately absent" (L1301) and a schedule-epoch transition is "not defined" (L2760-2762). Both tier 1 and tier 2 need a scheduled signature (L2846-2848), so a dormant, deadlocked or key-lost DAO permanently ends signed block production and every preconfirmation; the rollup becomes a forced-queue-only chain. §1, §2, §11 and §19 do not mention this precondition. *Fix:* make the latest row open-ended (or fail open to the last reviewed route) with SPLIT as the only shortening path, allow append after a lapse, and add the governance-liveness dependency to §1/§11.

*Design note on the beacon root (R2).* The beacon root is acceptable, but note that the SSZ layer is redundant even inside the current design: the Oracle already authenticates the carrier header via EIP-2935 and requires the payload `blockHash` to equal the carrier's `parentHash` (L2260-2261, L2691-2696), so `stateRoot`, `prevRandao` (`mixHash`) and `timestamp` are recoverable from the parent EL header RLP. The only SSZ-exclusive value is `BeaconBlock.slot`, which an EL timestamp bracket (`H_n.timestamp <= T < H_{n+1}.timestamp`, both headers EIP-2935-authenticated) replaces; the 8,191-block EIP-2935 horizon is far above the ~3,100 s target age. That variant deletes the fork-route registry, PVM kinds 2/5/6, the 1,811-window runway, the 12-second-slot and gindex pins, and with them C1 and H12. §17 item 7 rejects it in two sentences without argument.

### High

**H1. The aggregator seat duty is unreachable in steady state, and the model turns every duty into a recovery episode.** `targetTip = base + DELTA_RECOVERY_LAG` and `recoveryAt = baseTime + DELTA_RECOVERY_LAG` (L7817-7818), so `recoveryAt` is exactly the wall-clock time of slot `targetTip`. A prospective duty is refreshed only by a commit that "reaches the old target tip and lands no later than the old recovery equality" (L7836-7838), i.e. a candidate whose tip equals the wall-clock slot when it lands. With proofs at submission, 1,200 s windows and up to 900 s of proving, that never happens, so every primary's duty is allocated at `recoveryAt+1`. In `settlement-window-model.py` (L12346-12386, L14427-14440) allocation sets `Cause.SLA` and `_activate()` clears the normal window and opens recovery; three verifiers reproduced this (25 s of lag suffices). Result: with a paid aggregator installed, a healthy chain enters recovery roughly every `DELTA_RECOVERY_LAG` and revokes preconfirmations. This is also a spec/model contradiction: §7 lists two triggers (L3517-3523), §16's `sync()` uses an undefined `finalLagBreach` (L12613-12617), §9 adds a third seat trigger that "never waits for DELTA_FINAL_LAG" (L8054-8056), the model evaluates the `DELTA_FINAL_LAG` floor only when no primary is active (L14431-14436), and the `u8(causes)` byte hashed into `recoveryId` is defined nowhere in prose. *Fix:* target the tip at service start (or `base + LAG - allowed steady-state lag` with an enforced relation), never open recovery on duty allocation alone, and state one trigger rule in §7/§16/model.

**H2. A forced head becoming due mid-window cancels the window even when the stored best already consumes it (reproduced in the model).** `sync()` cancels an immature window whenever `dueAt[canonical.messageCursor] <= now` (L3521-3526, L12601-12603; model L14417-14419), keyed on the canonical cursor, not on `normalBest.endCursor`, and activation never checks the head's `dueAt`. Because a candidate's cutoff is frozen at the arm block, an envelope enqueued at `e` is first consumable by the *next* window, whose deadline exceeds `e+1500` whenever it activates more than ~300 s after `e`. For roughly three of four enqueue times the head becomes due inside that window, the already-proved best is dropped, and a `FORCE_DUE` episode opens; tier 2 needs slots `>= roundStartSlot`, so the cancelled window's promises are unlandable. Kind-1 bridge credits use the same queue, so ordinary traffic triggers it; one cheap kind-0 envelope per cycle keeps the chain in recovery. The relation `FORCE_DELAY >= W_SETTLE + T_INCLUDE_MAX` (L3509-3511) only covers appends after open. *Fix:* commit a covering `normalBest` early on force-due, or gate activation on `dueAt(head) > requiredThrough`, or size `FORCE_DELAY >= 2*W_SETTLE + arm gap + T_INCLUDE`.

**H3. Forced-queue drain capacity is per candidate while all flood items share one `dueAt`.** `dueAt = max(enqueuedAt+FORCE_DELAY, lastDueAt, recoveryExpiry+1)` (L4498) gives every item in a burst the same due time, while consumption is capped at 256 items / 80M accounted gas per candidate (L2878-2880) and the tier-3 escape is one block of 64 items / 20M gas (L2849-2856, L4575-4579), with one committing candidate per window or round. A normal candidate must prove `dueAt[endCursor] > requiredThrough` with no allowance for having exhausted the cap, so once the due backlog exceeds one cap the normal lane cannot close and recovery reopens after every commit. Deposits are priced per accounted gas with `accountedGas = max(gasLimit, intrinsic, MIN)` (L4430), and kind-1 credits are charged a flat 5,000,000 gas (L4434-4436) although their work is bounded to a handful of storage writes, capping bridge ingress at 4 credits per escape block. §8's disclaimer (L7003-7007) assumes the adversary must buy all capacity; round-bounded drain means ~17 heavy items per ~15 minutes suffices. *Fix:* space `dueAt` by prepaid work ahead, let a cap-exhausting candidate satisfy the boundary rule, allow multi-block tier-3 candidates, and price kind-1 at measured cost.

**H4. What a preconfirmation actually guarantees is far less than the README says.** Three verified mechanisms combine here. (a) The slashable promise namespace is `(builder, slot, contextId)` and `contextId` embeds `armBlockNumber/armBlockHash` (L2810-2813), so it changes every window, every cancel-and-rearm and every arm-block reorg; equivocation evidence requires equal context (L8181-8182, L8329-8333). Any block signed too late to be proved and landed before its window's deadline (under the spec's own 900 s proving bound, most of a 1,200 s window) must be re-signed under the next context, where a different body for the same slot is not slashable. (b) A successor builder may build on any ancestor within `G_MAX` (L2837-2843), and the candidate order `(blockCount desc, tipSlot desc)` (L3494-3496) makes a skip that re-includes the predecessor's transactions win at equal length; the spec calls this "Not slashable" (L8187-8188). Even a same-key conflict across two adjacent slots with the same parent, which is objectively provable from two headers, is excluded from slashing. (c) Any recovery activation, not only an escape commit, clears an L1-landed `normalBest` (L3527) and tier 2 cannot include pre-round slots (L2848-2849, L3570-3571), so up to ~3,600 slots of "provisionally proved" promises are revoked without slash. Net effect: a tier-1 signature commits to authorship and same-slot, same-window non-equivocation only. The README's "commitment to include and to order (§9)" (README L128) is not what §5.4 specifies. *Fix:* slot-scoped promise namespace (drop arm fields from the signed struct, bind the anchor at candidate level), slash same-key equal-parent conflicts across slots, and either let tier 2 include pre-round blocks that extend the canonical tip or state the revocation bound in §1/§5.4/§11.

**H5. `PUBLISH_MIGRATION_ARM` is a 7-day pause button (R1).** While the migration gate is ARMED/READY, `sync()` opens no context and never rolls an expired recovery round ("no context opens or rolls while armed", L12593-12594), a due queue head "is not drained or discarded" (L10706-10708), and forced ingress returns SYNCED. Arming needs no proof; the lease lasts exactly 604,800 s and only then may anyone abort (L6554, L10646-10648); a later-queued row can arm again after a fresh delay (L6601-6606). Kind-0 `validUntil` is at most 7 days, so items pending at arm expire as `EXPIRED_NO_TX`. Claims 2 and 3 in §1 (L116-135) and the 2,164 s recovery bound carry no migration carve-out. Since every profile, verifier and seat recalibration goes through this path, honest maintenance also costs liveness. *Fix:* keep the escape lane and forced ingress live during ARMED/READY (freeze only in the activation transaction), or pause `validUntil` clocks for pending items; add the carve-out to §1/§11.

**H6. An immutable bridge pauser with no bound, no unpause rule and no identity (R1).** Profile word 223 is `sourcePauser` (L8875) and the destination descriptor pins `address20(pauser)` (L11382); "Send/process/retry/manual-fail may be paused" (L5423-5425) and §11 calls it "Guardian pause" (L8572). The model's `set_paused` (L42672-42683, L52806-52816) checks only `caller == pauser`, with no duration cap, cooldown, self-expiry or permissionless unpause; the spec never says who holds the key or how a pause lifts, and §2 lists no such actor. Refund/expiry/withdraw paths are unpausable, so no theft, but a single key can freeze V2 DIRECT-ETH bridging in both directions indefinitely, by action or by loss. This contradicts "no owner-controlled executor ... is a protocol authority" (L6049-6050). *Fix:* remove it, or replace it with a bonded, self-expiring circuit breaker with a hard maximum and permissionless unpause, and name the role in §2/§13.

**H7. After genesis import, the legacy L1 SignalService receives no more checkpoints, stranding V1 L2-to-L1 withdrawals until a governance upgrade.** The campaign profile pins the SignalService's "implementation-embedded authorized syncer equal to the legacy Inbox" (L9821-9822), the legacy Inbox becomes permanently FROZEN (L10321-10322, L10570), and the new Settlement writes only its internal history ring with "no external publication call" (L12785-12786; L5927 "no checkpoint publisher exists in this revision"). The deployed `SignalService.saveCheckpoint` reverts for any other caller (`contracts/shared/signal/SignalService.sol:175`) and `proveSignalReceived` needs a checkpoint at the signal's L2 height. Launch V2 is L1-to-L2 DIRECT ETH only and every L2-to-L1 send stays V1 (L6963-6965), so every V1 ETH and Vault (ERC20/721/1155) withdrawal emitted after the imported height is unprovable on L1 until governance upgrades the SignalService (L6975-6977). That makes user-fund liveness for all bridged assets depend on unspecified governance action, and the abandoned value dwarfs the disclosed forced fees. *Fix:* specify a permissionless checkpoint path (authorize the immutable Router/target Settlement as syncer and write `(l2BlockNumber, tipHash, stateRoot)` on canonical commit), or disclose explicitly that V1 L2-to-L1 messaging ends at the last legacy checkpoint.

**H8. One blob per (block, chunk) with no packing.** The manifest array is one-to-one with the transaction's blobs (L3103-3108), the blob payload is `u32(len)||chunkBytes||zeroPadding` (L3369-3374), `chunkRoot` binds `fullBodyRoot||blockOrdinal||chunkIndex` (L11843-11848), and "Zero-byte padding still consumes a whole blob and therefore its whole blob-gas surcharge" (L3114-3115). Every nonempty L2 block therefore costs at least one full blob: one nonempty block per second needs ~12 blobs per L1 slot, and a 2 KB block pays the same DA as a 126 KB one. The 1-second cadence is only sustainable with mostly empty slots. *Fix:* a segment table per blob (`blockOrdinal, chunkIndex, offset, length`) under one KZG challenge, or chunks over a candidate-wide concatenated body stream.

**H9. The 16-session-per-candidate cap does not fit interleaved multi-builder placement.** A candidate may reference at most 16 sealed sessions (L2878, L3321-3322), only the owner may append (L3319), and the signed `dataManifestRoot` pre-commits `(sessionId, recordIndex)` (L11805-11809). Placement interleaves builders by hash order with non-adjacency (L2740-2748), so with 64 registrations the 17th distinct builder appears around slot 19. If each builder publishes its own data, the provable prefix stops there; if builders instead sign against shared publishers, each must predict a foreign session's next `recordIndex` and trust that publisher's append order, which reintroduces the optional aggregator as a hard dependency. The spec states neither topology. *Fix:* bound sessions per builder rather than per candidate, or reference records by content (`versionedHash`/leaf) so any publisher can post any builder's data.

**H10. Nothing relates `l2BlockGasLimit x 1 s` to proving throughput, and the failure mode is revocation, not slowdown.** `l2BlockGasLimit` (30M in the model and example profile) appears once in the spec (L8884) and in no §13 relation; `P_PROVE_MAX` is used only for the escape offset and session OPEN (L10855); the §14 proof gate (L11065-11067) fixes neither per-block gas nor opcode mix, so 4,096 empty blocks in 900 s satisfies it. Each window freezes `minAdmissibleSlot = activation - 1,200 s` (L3455) and the first block must extend the canonical tip, so blocks not landed in their own window are unlandable later; a prover fleet that falls behind full blocks does not degrade throughput, it forces builders to abandon the live chain each window or drives the tip to `DELTA_FINAL_LAG` and mass revocation. Related (M23): builders control proof cost with no proof-cost-aware schedule, so a minority of ticket holders can emit proof-hostile blocks that no candidate can prove in the window. *Fix:* an enforced relation between block gas, cadence, window length and a published proving-throughput word; a per-block proving budget; backpressure that degrades admissible gas under lag instead of revoking.

**H11. The builder's always-online signing key is also its only custody, exit and payout key.** "Registration is self-consenting: `builder=msg.sender`. Reservation and exit likewise require that same ECDSA address ... There is no `registerFor`, `tx.origin`, EIP-1271 delegation or caller-selected refund owner" (L1407-1410); `claimBuilderLeaseCreditV1(address recipient)` (L1585) is callable only by that key. A leaked 1 Hz signing key can equivocate in all up to 17 reserved windows (attacker keeps up to 20% as self-reporter, L8264-8266), revoke every outstanding promise, file the irrevocable one-shot exit (L1493), and then race the owner for the entire base bond (up to `BOND_MAX`); a lost key strands the bond forever. The seat market already has an operator/payout split (L7267); the far more exposed builder key does not. §11/§19 have no key-compromise row. *Fix:* separate signer from owner/payout at registration, add owner-only key revocation/rotation, and bound evidence to slots at or before the current L2 slot.

**H12. The fork-route machinery is fragile even with the beacon root accepted.** (a) The 12-second beacon slot is enforced by the protocol-lifetime Oracle itself (`targetTimestamp`, the 4788 scan step, "payload timestamp must equal `BEACON_GENESIS_TIME+12*parentSlot`" L2694-2695, and `u8(12)||u8(32)` in the constructor constant L1885), together with the EIP-4788/2935 runtime hashes, header field 19 and gindex 8; none of the six timelock kinds can re-pin them, so an L1 slot-time change (EIP-7782) or a system-contract redeploy ends sealing permanently. (b) Route intervals are exclusive, so a fork-date slip announced with less than 7 days' notice leaves every window whose target crosses the boundary VACANT for at least a week (L2513-2516, L2536-2537, L6104-6108), and the six-gindex row schema assumes the execution payload is an SSZ descendant of the beacon block, which EIP-7732 (Gloas/ePBS) breaks; the claim that a "reviewed verifier row" handles Gloas (L2314-2318) is therefore unsupported. (c) The 1,811-window correction runway is enforced only on APPEND (L2432-2434); SPLIT and REPLACE_PENDING rows need only `targetSlot(cw+8)+64 < first` (L2515-2518, L2551-2554), so a defective verifier installed that way is uncorrectable and yields at least 7 days of all-VACANT windows. *Fix:* the EL-only snapshot removes all of this; otherwise move slot time and system-contract identity into the delayed row, add a fail-open to the last reviewed route, and apply the runway to every new-verifier row.

### Medium

- **M1 (F09).** The 1,024-cell data-session ring is the only DA path for signed blocks (§17 rejects same-transaction blobs, L12758-12760). REFUND cells count against capacity for TTL + 86,400 s grace, owners can claim-and-reopen in one transaction so no FREE cell is ever observable, the two-per-owner cap is address-keyed, and BOND/BASE have only overflow relations (L8289-8291; null in the example). A Sybil can hold the whole ring for the price of refundable bonds plus base rent; §11 mislabels this "optional market" (L8564).
- **M2 (F10).** `postData` requires `msg.value == postRent` where `postRent` depends on `block.blobbasefee` (L3096-3112); a public-mempool blob post is correctly priced only for the immediately next block, otherwise it reverts and still burns the blob fee. Accept `>=` with refund or a caller-supplied fee bound.
- **M3 (F12).** L2 header/fork rules are unspecified: words 248-251 (`systemTransactionRulesHash`, `forcedInputRulesHash`, `stateTransitionAbiHash`, `legacyExecutionRulesHash`) have no defined preimage (the model fills them with placeholder labels, L31437-31440); word 247 "headerRulesHash" is keccak of the force-send label (L8911-8912); base-fee recipient, L2 type-3 acceptance, `prevRandao`, `extraData`, `parentBeaconBlockRoot` and coinbase per tier are never derived.
- **M4 (F29).** `FEE_NO_TX` ("underpriced", L4619-4622, L4634-4637) determines the receipt roots but has no comparison rule; `maxFee` (L3587) is never used; NONCE/FUNDS are equally undefined; no model implements the classification. A builder with consecutive slots can price out a targeted due head via the base fee.
- **M5 (F14).** Anyone may enter QUIESCENT at `quiesceNotBeforeBlock` (L10175-10177) with no requirement that the unfinalized proposal suffix be empty; activation abandons `[lastFinalized+1, nextProposalId)` (L10313-10316). The racer cannot enlarge the suffix but can freeze it at its maximum. Require an empty suffix or a deterministic latest-feasible entry.
- **M6 (F26).** `sealScheduleWindowV1` is heavy, permissionless and unpaid ("optional maintenance distributor", L2287, defined nowhere; L8453 says none is deployed) with a hard permanent deadline (L2213-2214). One missed seal blanks 384 slots. Add a normative seal reward and a grace path.
- **M7 (F28).** The model (`arm_normal_context`/`activate_normal_context`, L13734-13780) has no ARMED/OPEN sub-state and allows re-arm and re-activation over an open window once the arm is 255 blocks old, overwriting deadline/`minAdmissibleSlot`/`contextId` while keeping the old best; §16 forbids it (L12545). Fix the model and state the guard in §7.
- **M8 (F47).** The normative §16 `sync()` never sets `changed` after a non-ARMED normal commit or cancel (L12600-12618), contradicting the SYNCED rule (L363-368) and the model; a literal implementation can roll back a canonical commit through a later revert.
- **M9 (F24).** The "current dynamic floor" term of `globalMinReferencedSlot` (L501-503, L2144) is defined nowhere in prose or models, yet in `NORMAL_IDLE` it is the sole input to schedule expiry, tranche release, generation release and ring reuse.
- **M10 (F37).** `DELTA_RECOVERY_LAG` and `DELTA_SLASH_LAG` are consensus-critical seat clocks defined only at L7812-7823, absent from §13, and constrained only by ordering; nothing relates `slashLag` to `ESCAPE_OFFSET + DELTA_TIP + inclusion`, so a legal profile can make every SLA duty uncurable.
- **M11 (F38).** `premiumCap` prose (L7728-7731) caps accrual at objective failover; the model caps at `duty.recoveryAt` (L11965-11990) and its tests pin that. Up to 2,400 s of ask per duty differs.
- **M12 (F39).** A duty cured after strict failover is still SATISFIED and refundable (L8058-8061) while the failed-over primary's own max-ask standby is promoted with fresh clocks; no operator-uniqueness rule exists, so the collusion bond inequality (L8307-8318) assumes a forfeiture that never happens.
- **M13 (F31).** The normative `DestinationBridgeDescriptorV2` (L5529-5535) has eleven fields and no pauser, quota period or SignalService, but the model's descriptor carries them and `set_paused` authenticates against `descriptor.pauser`.
- **M14 (F33).** The destination native quota is a single immutable refilling bucket (cap fixed at tx0, L5662; no setter, model L21934-21936), debited `value+fee` on every DONE. Self-LP delivery to one's own address is allowed, so an attacker can pin available quota at zero each period for gas plus a recyclable capital lock, stalling all other V2 deliveries until they expire. Capacity changes need a new release.
- **M15 (F36).** `REGISTER_RELEASE` and `PUBLISH_MIGRATION_ARM` can be queued together and mature after one 7-day delay (L6083-6088, L6603-6607); a malicious verifier then finalizes arbitrary state. V2 has no immutable L2-to-L1 exit during that notice (every L2-to-L1 send is V1, see H7). Governance is a disclosed safety root, but the effective exit window is not stated in §1/§2.
- **M16 (F30).** Tier-3 escape rewards are paid only from voluntary `fundRewardClassV1` buckets (L8477-8481); no protocol flow feeds them (sinks are separate), and the forced-envelope fee words are static per profile, not indexed to base fee, so a price move needs a 7-day release and deposits ratchet monotonically (L4462-4464). Economic liveness of the censorship floor rests on altruism or timely governance.
- **M17 (F50).** §5.4 and §11 attribute revocation to escape commits only, but any recovery activation clears the landed best (L3527) and tier 2 excludes pre-round slots; the README says "recovery expires unfinalized preconfirmations". State one rule and the bound.
- **M18 (G01).** The builder lease token has no specified issuer, supply, distribution or explicit no-mint rule; the prohibition list appears twice with different contents (L607-611 vs L8250-8253) and the profile kind is just `ERC20_NO_HOOK`. Since ranking is purely by bond, whoever controls issuance or the initial float gates entry to the "permissionless" builder market. Denominate builder bonds in ETH or pin `totalSupply` and a launch distribution.
- **M19 (G02).** `leasePerWindowAtomic` is a lifetime constructor word of the ownerless Registry (L598-600, L640-643) in atomic units of an unpegged ERC-20, while everything at stake (MEV, revocation, bridge flows) is ETH; no relation, reference or repricing path exists.
- **M20 (G03).** The "meaningful entry barrier" (L8588-8589) is fully refundable escrow; skip/withhold is unslashable; a rented or thin-float capture of the schedule costs interest only. Consider a non-refundable per-window rent or minimum tenure.
- **M21 (G13).** `sessionId` derives from a contract-global `nextSessionSequence` (L2996-3003), so a 1-2 block L1 reorg that reorders opens changes the ID every signed `dataManifestRoot` pre-commits; re-signing the same slot with a different manifest is equivocation (L1753-1756). Use a per-owner sequence and state an open-stability depth.
- **M22 (G14).** Acceptance checks `expiry >= deadline + REORG_MARGIN` (L3323-3324) but commit checks `closingBlock.timestamp + REORG_MARGIN <= minDataExpiry` (L3503-3507) with no `T_INCLUDE_MAX` tolerance, so a minimum-expiry session makes the window commit only if `sync()` lands in the exact deadline block. A scheduled builder can use this to cancel windows.
- **M23 (G08).** No proof-cost-aware gas schedule, per-block proving budget, objective skip criterion or slashable offence for proof-hostile blocks (only forced envelopes carry `proofWeiPerAccountedGas`, L4456-4461); see H10.
- **M24 (G15).** `blockOrdinal` is the candidate-wide index and is hashed into `chunkRoot`, the KZG challenge and the MMR leaf (L3121, L3350-3351, L11755, L11844-11846), so any base change forces re-posting of every unlanded block's blobs, not just re-signing.
- **M25 (G10).** Covered in H12(c).
- **M26 (F20).** §2 omits the bridge pauser, the DAO proposer's exclusive queue/cancel right, the six immutable sinks (including the escape-block coinbase, L7082) and LPs; readers of §1-§2 cannot check R1.
- **M27 (F16).** Covered in H4(b).
- **M28 (F06).** Covered in H1.

<details>
<summary><b>Low and informational (56 items, click to expand)</b></summary>

- F19: §5 says the normal anchor header is authenticated via EIP-2935 while §3/§7/§12 use native `blockhash`; state one primitive.
- F22: `maximumLiableUntil` and `releaseWindow` are used as release predicates but never defined.
- F25: BuilderRegistry/ScheduleOracle/ProtocolRootFactory may not fit EIP-170; the fallback helper topology is unspecified.
- F27: winner-takes-all reward with strict-better replacement pays nothing for displaced proofs, driving last-moment submission.
- F34: the V2 bridge is hard-wired into the consensus proof statement (system tx at index 1, 541-byte rows); scope/complexity risk for a preconfirmation protocol.
- F40: zero-ask seat offers cannot be beaten or displaced once installed; four of them capture the waiting book.
- F41: `protocolCoinbaseSink` (profile word 71) has no source in the calibrated profile or sink table.
- F42: equivocation deterrent is one `L_LEASE` per (builder, window) regardless of ticket count; the base bond is never slashed.
- F43: the frozen profile schema requires nonzero `daoProposer`, `sourcePauser`, `aggregatorSeatMarket` and lease token, so a configuration without those roles is not expressible.
- F46: hard-coded 12-second L1 slot conversions outside the schedule (EIP-2935 arm bound, 256-cell history ring, blob retention).
- F48: the appendix cutover journal contradicts §8 and invariant 13 on the terminal VMC1 call and final write.
- F49: README says a preconfirmation is "a commitment to include and to order (§9)"; §5.4 says authorship and parent choice only.
- F52: model assurance is shallow on the properties that matter (signer-vs-schedule authority, proofs and gas are placeholders); most of the 135k lines are encoding fixtures.
- F54: 219 pages, v2.26, 155 commits, ~20 four-letter wire codes (MACT/MFRZ/MCAN/QMIG/MAPS/BRS1/BRD1/BRC1/ABR2/RTR2/...) and word-index cross-references make the design very hard to review or implement independently.
- F55: the signed `dataManifestRoot` pre-commits `sessionId`/`recordIndex`, so only the session owner can ever publish a promised body.
- F57: a stale arm activated at age up to 255 blocks freezes an old queue cutoff; an opener can burn a 1,200 s window.
- F60: placement prose never defines `otherRemaining` (whether VACANT counts); only the model disambiguates.
- F61: the seed is a single L1 `prevRandao`, biasable by ~one bit per adjacent proposer; not analysed.
- F62: "rejects zero/ill-typed fields" contradicts required-zero and naturally-zero statement fields.
- F63: §13 omits values used in enforced relations (`CLOCK_SKEW`, seal margin, `REORG_STABILITY_SECONDS`).
- F65: Settlement hard-codes point-evaluation precompile gas, address and return format; an L1 repricing needs a new root.
- F66: §7 says round refresh promotes no successor; the model promotes the selected successor.
- F68: kind-0 forced ingress requires `msg.sender == recovered L2 sender`, excluding contract accounts and relayers from the censorship floor.
- F69: `forcedExpirySink` is declared but no rule routes value to it.
- F70: the source-bundle QuotaManager is deployed and hashed into every release but never consulted.
- F71: real-CALL attempts debit quota only after target execution, so LPs burn full `gasLimit` on quota misses.
- F73: an unexecuted PENDING INVALIDATE/MIGRATION_CANCEL intent silently makes `stageBest` revert.
- F74: seat waiting-book griefing via threshold-free displacement and immediate pending exit.
- F75: migration arm "excuses unresolved duties" in prose; the model excuses only OPEN duties.
- F76: §11 says an exhausted reward vault pays "zero or partial"; §10 says insufficient funding reverts the claim.
- F77: reward metering pays per builder-controlled `gasUsed` with no cost-linked relation.
- F78: 256-cell per-class receipt rings with 88,200 s occupancy make silent reward forfeiture routine.
- F81: profile `genesisTimestamp` (word 8) is not required to equal the Registry/Oracle genesis timestamp.
- F82: the L2-side EIP-2935 history contract (words 245/246) has no deployment or per-block system-call rule.
- F85: every legacy mutator/resume/upgrade path fails closed on Router LGC1/profile-leaf reads with no time-based fallback.
- F86: snapshot geometry "strictly exceeds" contradicts both models, which enforce `>=`.
- F87: §14 names an "emergency governance path" the spec never defines.
- F92: §1 says legacy proof-finalization stays live during the whole campaign; §12 blocks proving at QUIESCENT.
- F93: README, PR description and §18 disagree on model/test counts (see model audit).
- F94: "either executable model" leaves the normative model set unspecified.
- F95: the immutable DAO proposer is an unconstrained, non-rotatable address (may be an EOA or a proxy).
- G11: only the latest fork row is correctable; a mistaken non-latest pending row can be shrunk only to one slot.
- F100 (info): the seat market is undeployable as committed (nearly every parameter null) and buys ~1,000 lines/three contracts for a bond that binds only if nobody recovers for `DELTA_SLASH_LAG`.
- F101 (info): the reporter reward is capturable by any front-runner; deterrence holds, monitoring incentive does not.
- F102 (info): worst-case ~94 min (lag trigger) / ~61 min (forced trigger) to the first recovered block.
- F103 (info): a hostile DAO can chain back-to-back genesis campaigns to keep legacy forced ingress fenced most of the time.
- F64, F72, F79, F83, F84, F90, F91, F97, F98, F99 (info): minor internal inconsistencies (reclaimSurplus predicate stated twice, profile size caps, 128-call scan inclusion assumption, "byte-identical legacy configuration" vs replaced verifier, `winningDataCommitment` domain at genesis, `F_FINAL` is EL depth not CL finality, thin-market VACANT coverage, `firstSlot` binding, surplus reclaim blockable by one unwithdrawn pull).

</details>

### Model audit

All six suites pass here and match the README: `settlement-window-model.py` 186 properties, `test-settlement-window.py` 293 tests (21,340 s, about 6 hours of wall-clock, so this will not run in CI per push), `commitment-model.py` 812 vectors / 1,593 assertion sites, `test-seat-market.py` 114, `lookahead-model.py` 38, `test-economic-profile.py` 38. The PR description still says 764/1,503, 184 and 277; please regenerate those numbers from the models in the same commit. Two caveats on what a passing run proves: signatures, validity proofs, EVM execution and gas are placeholders, and a large share of the corpus checks encodings against values produced by the same code. Where the models and prose disagree on consensus rules (H1, M7, M8, M11, M13, F66, F75, F86) the spec's own §1 rule says deployment is blocked, so those need to be resolved before the freeze label is accurate.

### Checked and found sound

For balance, candidate findings that did not survive verification: monopolising the four-move replacement budget (a replacement needs a strictly greater bond than the global smallest live entry, so it costs real capital); an escape-vs-tier-2 race (the escape slot is frozen at `roundStart + 1,900 s`, giving tier 2 a head start); builder-forced aggregator breach via unprovable blocks (the tier-3 escape is a builder-independent cure); post-launch runtime setters (every economic change goes through typed timelock operations); an exit-FIFO deadlock (the model persists `exit_head_sequence`); prover-chosen builder sets (Settlement authenticates every schedule row against ScheduleOracle SWV1, L2157-2170); and a lying-verifier safety hole (explicitly scoped out in §2).

### Suggested order of work

1. Resolve C1 and H12 together: either adopt the EL-only snapshot or make the fork-route registry fail open with an open-ended latest row.
2. Decide what a preconfirmation promises (H4) and make the signed struct and slashing rules match; then fix the README.
3. Fix the two reproduced state-machine defects (H1, H2) and the drain/`dueAt` mismatch (H3), and reconcile the recovery trigger across §7/§9/§16/model.
4. Remove or bound the three non-upgrade privileges (H5, H6, H7) and add a privileged-roles table to §2.
5. Rework DA for the 1-second cadence (H8, H9, H10, M1, M2, M24) or lower the advertised cadence.
6. Add the missing consensus definitions (M3, M4, M9, M10) and the builder key/owner split (H11).
7. Calibrate the economics before calling this a freeze: the example profile is `UNCALIBRATED` with every bond, rent and seat value null, so incentive compatibility cannot be assessed today.

---

PR #22065, opened 2026-08-29
feat(bridge-ui): offer a private relay when a claim's fee is contestable

dantaik:
## Problem

`processMessage(message, proof)` is permissionless and pays the processing fee to whoever lands it first. When a user claims manually, the transaction sitting in the public mempool carries everything a competitor needs — message plus merkle proof — to submit the same call and take the fee. The user's own transaction then reverts on the status check, so they lose the fee refund *and* pay gas for the failure.

This is narrower than it first looks. The bridge only pays a third party when both of these hold:

- **`gasLimit != 0`** — otherwise `processMessage` reverts with `B_PERMISSION_DENIED` for any sender but the destination owner. The UI's "only I can claim" checkbox sets exactly this.
- **`fee > 0`** — otherwise there is no prize. Claiming someone else's message delivers their funds and costs the sender gas.

So the exposed case is a user manually claiming a message that still carries a fee — typically because no relayer picked it up. `isClaimFeeContestable` encodes both conditions and the notice renders only then.

## Change

Offer Flashbots Protect from the claim pre-check, on the chains it covers. It relays transactions straight to block builders instead of gossiping them, takes no service fee, and drops transactions that would revert — so a claim that loses the race costs nothing. The endpoint pins `hint=hash`, its maximum privacy setting, so MEV-Share searchers only ever see the transaction hash and never the message or the proof.

Scope note: the wallet broadcasts the transaction, not this app, so the UI can only prompt — via `wallet_addEthereumChain`. The user is free to reject it and claim over their own endpoint; nothing in the flow is blocked either way. Wallets give no way to read back which endpoint they actually use, so an accepted prompt is remembered per chain to avoid asking again.

Only Ethereum has a private relay market today, so claims landing on Taiko are unaffected and keep using the wallet's own endpoint.

Not applied to the release step: `recallMessage` pays no one — it returns the message value to the source owner — so front-running a release buys nothing.

### Files

- `src/libs/mev/` — the relay table (`privateRpc.ts`), the wallet prompt (`enablePrivateRpc.ts`), the per-chain preference (`privateRpcPreference.ts`), and the contestability predicate (`isClaimFeeContestable.ts`), with unit tests.
- `src/components/MevProtection/MevProtectionNotice.svelte` — the opt-in card; renders nothing on chains with no relay, or once the user has accepted.
- `ClaimPreCheck.svelte` — renders the notice when the wallet is on the destination chain and the fee is contestable.
- `app.config.ts`, `i18n/en.json`, `libs/error/errors.ts` — storage key, copy, and one error type.

## Alternatives considered

MEV Blocker (`/fullprivacy`) is the close runner-up: also free and permissionless, and fastest to inclusion in the [independent RPC benchmark](https://arxiv.org/html/2505.19708v1). Flashbots wins here on covering Sepolia as well as mainnet, so the flow can be exercised on a testnet. `getPrivateRpc` is a plain table, so adding MEV Blocker later — or letting the user pick — is a small change.

Rejected: bloXroute Protect (requires an account and shares transactions with searchers via BackRunMe) and Merkle (74% inclusion success in the same benchmark).

## Out of scope

The off-chain relayer in `packages/relayer` sends the same `processMessage` call and is the party with real, continuous exposure — it competes for fees on every message by definition. It would benefit from the `privateTxMgr` / `TxMgrSelector` pattern that `taiko-client` already uses. This PR is deliberately limited to `packages/bridge-ui`.

## Testing

- `pnpm --filter bridge-ui test:unit` — 12 new tests pass; the 17 pre-existing failures in this environment are unrelated (they need the generated `$chainConfig`, which is built from deployment env vars).
- `pnpm --filter bridge-ui svelte:check` — 0 errors, 0 warnings.
- `pnpm --filter bridge-ui build` — succeeds.
- `prettier --check` and `eslint` clean on all touched files.

Not exercised against a live wallet: the `wallet_addEthereumChain` prompt is covered by unit tests against a mocked wallet client, but the actual prompt UX should be checked in MetaMask before this leaves draft, since wallets vary in how they handle adding an endpoint to a network that already exists.

### DIFF
--- a/packages/bridge-ui/src/app.config.ts
+++ b/packages/bridge-ui/src/app.config.ts
@@ -19,6 +19,7 @@ export const pendingTransaction = {
 export const storageService = {
   bridgeTxPrefix: 'transactions',
   customTokenPrefix: 'custom-tokens',
+  mevProtectionPrefix: 'mev-protection',
 };
 
 export const bridgeTransactionPoller = {
--- a/packages/bridge-ui/src/components/Dialogs/ReleaseDialog/ReleaseSteps/ReleasePreCheck.svelte
+++ b/packages/bridge-ui/src/components/Dialogs/ReleaseDialog/ReleaseSteps/ReleasePreCheck.svelte
@@ -4,6 +4,7 @@
 
   import { ActionButton } from '$components/Button';
   import { Icon } from '$components/Icon';
+  import { MevProtectionNotice } from '$components/MevProtection';
   import type { BridgeTransaction } from '$libs/bridge';
   import { getChainName } from '$libs/chain';
   import { config } from '$libs/wagmi';
@@ -55,6 +56,9 @@
       </div>
     </div>
   </div>
+  {#if correctChain}
+    <MevProtectionNotice chainId={Number(tx.srcChainId)} />
+  {/if}
   {#if !canContinue && !correctChain}
     <div class="h-sep" />
     <div class="f-col space-y-[16px]">
--- a/packages/bridge-ui/src/components/Dialogs/ReleaseDialog/ReleaseSteps/ReleasePreCheck.svelte
+++ b/packages/bridge-ui/src/components/Dialogs/ReleaseDialog/ReleaseSteps/ReleasePreCheck.svelte
@@ -4,7 +4,6 @@
 
   import { ActionButton } from '$components/Button';
   import { Icon } from '$components/Icon';
-  import { MevProtectionNotice } from '$components/MevProtection';
   import type { BridgeTransaction } from '$libs/bridge';
   import { getChainName } from '$libs/chain';
   import { config } from '$libs/wagmi';
@@ -56,9 +55,6 @@
       </div>
     </div>
   </div>
-  {#if correctChain}
-    <MevProtectionNotice chainId={Number(tx.srcChainId)} />
-  {/if}
   {#if !canContinue && !correctChain}
     <div class="h-sep" />
     <div class="f-col space-y-[16px]">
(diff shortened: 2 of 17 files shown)


dantaik:
Closing: the fix belongs in the off-chain relayer, not here.

The exposure this PR addresses turned out to be narrow. `processMessage` pays a third party only when `gasLimit != 0` (otherwise it reverts with `B_PERMISSION_DENIED` for any sender but the destination owner) **and** `fee > 0`. So a user's manual claim is only contestable when they are claiming a message that still carries a fee — and losing that race costs them the fee refund they were trying to save, plus gas on the reverted transaction, while their funds still arrive.

The UI also cannot actually route the transaction: the wallet broadcasts it, so this could only prompt via `wallet_addEthereumChain`, which rewrites the user's RPC for all of Ethereum mainnet rather than just this claim, cannot be verified afterwards, and varies by wallet.

The relayer in `packages/relayer` has the real, continuous exposure — it competes for these fees on every message — and it holds its own key, so pointing `eth_sendRawTransaction` at a private relay there is deterministic. `taiko-client` already has the pattern (`--l1.private` plus `TxMgrSelector`, which falls back to the public txmgr when the private one fails).

Analysis of the relay options, for whoever picks this up: Flashbots Protect at `?hint=hash` and MEV Blocker at `/fullprivacy` are both free with no service fee; the max-privacy setting matters because the default on both shares transaction data with searchers. bloXroute Protect needs an account and forwards transactions to searchers via BackRunMe; Merkle landed 74% of transactions in the [2025 RPC benchmark](https://arxiv.org/html/2505.19708v1).

Note that this only covers L2→L1 messages. For L1→L2 the destination is Taiko, where no private relay exists.

---

PR #22071, opened 2026-08-31
fix(relayer): bound HTTP RPC requests and preserve progress monitoring

davidtaikocha:
> Draft — coordinated with [`k8s-configs#1609`](https://github.com/taikoxyz/k8s-configs/pull/1609), which switches the deployed relayer RPC URLs from WebSocket to HTTP. Merge and publish this relayer image before deploying that configuration change.

## What

- Keeps the existing 10-second deadline on the indexer's `HeaderByNumber("latest")` call.
- Adds a shared relayer RPC dialer for API, bridge, indexer, and processor.
- Gives every HTTP JSON-RPC request a configurable overall deadline (`ETH_CLIENT_REQUEST_TIMEOUT`, default `5m`).
- Reuses the processor's bounded source RPC connection for both raw RPC and `ethclient` calls.
- Replaces the metrics-only `SubscribeNewHead` loop with a 10-second `HeaderByNumber` poll, so `blocks_scanned_ops_total` and its existing alerts continue to work over HTTP.
- Makes the block-progress retry loops stop when their component context is canceled.

WebSocket and IPC URLs still use go-ethereum's native transports. The new client-level timeout is applied only to HTTP/HTTPS.

## Why

On 2026-08-30 an internal-devnet relayer stayed `Running` and `Ready` but stopped processing for two hours after the node pods behind its Kubernetes Service were replaced. Its RabbitMQ consumer remained attached, 113k messages accumulated, and the process emitted no new logs. The relayer had no overall timeout on several execution-critical RPC calls, so a stale connection could block a worker indefinitely.

Changing the URLs to HTTP alone is insufficient: Go's HTTP transport normally reuses healthy connections. This PR supplies the missing recovery bound. A stuck request is canceled after at most five minutes; normal RPCs such as fuzz-heavy `FilterLogs` batches that can legitimately exceed ten seconds still have enough time to finish. The dedicated head query keeps its tighter ten-second deadline.

## Metrics compatibility

HTTP does not support `eth_subscribe`, but Hoodi and mainnet alert on:

```promql
increase(blocks_scanned_ops_total[15m]) < 1
```

The new polling loop records actual source-chain head changes rather than counting every poll. The first successful poll establishes a baseline, and later hash changes increment the existing counter. This preserves the alert's progress semantics without requiring a separate WebSocket connection.

## Configuration

| flag | environment variable | default | purpose |
|---|---|---:|---|
| `ethClientTimeout` | `ETH_CLIENT_TIMEOUT` | `10s` | existing short per-call/contract deadline |
| `ethClientRequestTimeout` | `ETH_CLIENT_REQUEST_TIMEOUT` | `5m` | new HTTP client ceiling for every JSON-RPC request |

## Verification

- Affected relayer package tests pass, including new HTTP timeout and polling metric tests.
- `go vet` passes for all changed packages.
- The relayer command builds successfully.
- The repository-wide relayer test command still has two unrelated baseline/environment failures: `pkg/repo` resolves a Moby `Port` API without `Int`, and `pkg/metrics` cannot bind its test port in the restricted sandbox.

## Rollout

1. Merge this PR and wait for the relayer Docker workflow to publish its `sha-<commit>` image.
2. Update `k8s-configs#1609` to that exact image tag.
3. Deploy one internal-devnet relayer first and confirm head-progress metrics, queue acknowledgements, and bridge processing.
4. Roll through Hoodi, then mainnet.

### DIFF
--- a/packages/relayer/api/api.go
+++ b/packages/relayer/api/api.go
@@ -13,6 +13,7 @@ import (
 	"github.com/labstack/echo/v4"
 	"github.com/taikoxyz/taiko-mono/packages/relayer/pkg/http"
 	"github.com/taikoxyz/taiko-mono/packages/relayer/pkg/repo"
+	"github.com/taikoxyz/taiko-mono/packages/relayer/pkg/rpcclient"
 	"github.com/taikoxyz/taiko-mono/packages/relayer/pkg/utils"
 	"github.com/urfave/cli/v2"
 )
@@ -49,12 +50,12 @@ func InitFromConfig(ctx context.Context, api *API, cfg *Config) (err error) {
 	ctxDial, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
 
-	srcEthClient, err := ethclient.DialContext(ctxDial, cfg.SrcRPCUrl)
+	srcEthClient, err := rpcclient.DialEthClient(ctxDial, cfg.SrcRPCUrl, cfg.ETHClientRequestTimeout)
 	if err != nil {
 		return err
 	}
 
-	destEthClient, err := ethclient.DialContext(ctxDial, cfg.DestRPCUrl)
+	destEthClient, err := rpcclient.DialEthClient(ctxDial, cfg.DestRPCUrl, cfg.ETHClientRequestTimeout)
 	if err != nil {
 		return err
 	}
--- a/packages/relayer/api/config.go
+++ b/packages/relayer/api/config.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"strings"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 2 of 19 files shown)


smtmfft (approved):
Reviewed the latest head `579993385`.

The relayer HTTP timeout path looks consistent: HTTP/HTTPS clients get a bounded request timeout, WS/IPC continue through go-ethereum's native transports, and the source head monitor now uses polling so the existing progress metric still works after the planned HTTP URL rollout. I also checked the remaining review points around metric semantics and retry shutdown; they match the PR description and do not look blocking to me.

Verification: `git diff --check origin/main...HEAD` is clean; local `go test ./pkg/rpcclient ./pkg/utils ./cmd/flags` and `go test ./api ./bridge ./indexer ./processor` both passed; GitHub `lint-relayer` and `test-relayer` are green.

No concerns from my side.

PR #22063, opened 2026-08-28
chore(taiko-client,taiko-client-rs): remove Masaya network support

davidtaikocha:
## Summary

- remove Masaya network registration and fork scheduling from the Go client
- make the Rust protocol reject Masaya chain ID `167011` through the existing unsupported-chain error
- align taiko-client with taiko-geth `taiko` at `fe4f45c83f7a`
- align taiko-client-rs with alethia-reth `main` at `4992a91cfb54`
- update eventindexer and relayer MySQL test helpers for the testcontainers-go v0.44 mapped-port API
- preserve historical release notes

## Testing

- `PACKAGE=... make test`
- `PACKAGE=pkg/config make test`
- `PACKAGE=bindings/manifest make test`
- `PACKAGE=../eventindexer/pkg/repo make test` — passed with 63.8% coverage
- `PACKAGE=../relayer/pkg/repo make test` — passed with 42.5% coverage
- `just fmt`
- `just clippy-fix`
- `just test` — 411 passed, 0 skipped
- `cargo test --locked -p protocol --no-default-features` — 43 unit tests and 1 integration test passed
- fresh CI: `lint-eventindexer`, `test-eventindexer`, `lint-relayer`, and `test-relayer` passed

## Notes

- `make lint` was attempted, but the repository tooling currently installs `golangci-lint v2.13.2` with Go 1.26 and then fails because a checked file requires Go 1.27.

PR #22072, opened 2026-08-31
fix(eventindexer): make transfer-log balance updates replay-idempotent

davidtaikocha:
## Problem

The indexer deliberately re-processes a trailing range of blocks on every restart. `setInitialIndexingBlockByMode` rewinds the cursor to `MAX(events.emitted_block_id) - 1` and `filter()` resumes at `cursor + 1`:

```go
// indexer/set_initial_processing_block_height.go
if latest != 0 {
    startingBlock = latest - 1
}
// indexer/filter.go
for j := i.latestIndexedBlockNumber + 1; j <= endBlockID; j += i.blockBatchSize {
```

Only protocol events advance `events.emitted_block_id`. On L2 that means Bridge events, which are sparse — measured on mainnet L2: **37 logs from `0x1670…0001` across the last 150,000 blocks, most recent one 3,035 blocks behind tip**. So the replayed range is routinely thousands of blocks, not the trailing one.

NFT and ERC20 balances are maintained by read-modify-write with no log identity:

- `increaseBalanceInDB` does `b.Amount += opts.Amount` on the row it finds
- `decreaseBalanceInDB` returns `nil, nil` when the row is gone
- `nft_balances` / `erc20_balances` carry only **non-unique** indexes

Replay an ERC-721 `A -> B` and B goes to 2, while the already-deleted A cannot be decremented again. ERC-1155 amounts and ERC20 amounts double the same way. It is silent, and it never self-corrects.

That the codebase already knows replay happens: `pkg/repo/transaction.go:69` swallows duplicate-key errors on `transactions` for exactly this reason. The balance tables were the unfinished half of that defence.

Credit to @smtmfft, who found this reviewing taikoxyz/k8s-configs#1606.

## Fix

Claim each balance-mutating unit of a transfer log in a new `processed_transfer_logs` table, **inside the same transaction as the mutation**, keyed by `(chain_id, tx_hash, log_index, batch_index, kind)`:

```go
res := tx.WithContext(ctx).Clauses(clause.Insert{Modifier: "IGNORE"}).Create(p)
...
return res.RowsAffected == 1, nil
```

The unique key plus `INSERT IGNORE` means a replayed log is dropped by the storage engine and affects no rows, so the balance mutation is skipped. Sharing the transaction means a rolled-back mutation also releases the claim, and the existing deadlock-retry loop keeps working.

The `IGNORE` modifier is load-bearing and the first revision of this PR got it wrong. `clause.OnConflict{DoNothing: true}` reads as the idiomatic equivalent, but gorm's MySQL dialector rewrites it to `ON DUPLICATE KEY UPDATE id=id` (it sets `DoNothing = false` and synthesises the assignment from the primary key). That statement's affected-row count is 0 only while the connection lacks `CLIENT_FOUND_ROWS`; with `clientFoundRows=true` in the DSN the no-op update reports one row, every replay reads as new, and the mutation is applied twice — exactly the corruption the claim exists to prevent. The production DSN does not set the flag, so nothing was ever corrupted by it, but replay correctness should not hinge on an implicit driver setting.

`batch_index` is what keeps the units of a single ERC-1155 `TransferBatch` distinct — they share a `(tx_hash, log_index)` and would otherwise collapse into one claim, so only the first token id in each batch would ever be applied. `kind` separates the NFT and ERC20 paths so neither can mask the other.

## Scope: ERC20 is fixed too

`saveERC20Transfer` has the identical defect, and unlike the NFT path **it is enabled in production today** — hoodi's L2 eventindexer runs `INDEX_ERC20S=true`, so hoodi's ERC20 balances are already exposed to this. Fixing only the NFT half would leave a live instance of the same bug behind. Happy to split it out if reviewers prefer two PRs.

## The first restart after this deploys: guarded

`processed_transfer_logs` starts empty, so logs applied by a *pre-upgrade* process are unclaimed and that one restart would still double-apply its trailing range. Nothing records what the old process applied, so the indexer cannot repair this — but it can refuse to do it silently.

`UnclaimedBalanceReplayRisk` reports risk when a kind has no claims for the chain **and** the balance table was last written at or after the block the cursor resumes from:

```
claims(chain, kind) == 0
AND MAX(balance.updated_at) >= transacted_at(MAX(events.emitted_block_id))
```

`Indexer.Start()` checks it before `setInitialIndexingBlockByMode`, so it fails before any RPC work.

Recency rather than mere existence is what makes it usable. A chain whose balance indexing was switched off long ago and is only now being re-enabled — mainnet, whose `nft_balances` rows predate 2025-05-31 — has a replay range entirely newer than anything already written, so it cannot double count and must not be blocked. A coarse "balance table is non-empty" test would refuse exactly the deploy this work is aimed at.

Ways out: `--allowUnclaimedBalanceReplay` / `ALLOW_UNCLAIMED_BALANCE_REPLAY=true` accepts a single overcount with a warning, or reset the affected balances so they rebuild from claims.

Environments as they stand: mainnet has both flags off, nothing fires. hoodi runs `INDEX_NFTS=true` + `INDEX_ERC20S=true` against live balances and will trip the guard on the upgrade restart — intended, and a decision for deploy time.

## Also in this PR: unbreaking `pkg/repo` tests

`pkg/repo/containers_test.go` has not compiled since #21998 (2026-08-15) bumped testcontainers-go v0.40 → v0.44, where `MappedPort` changed from `nat.Port` to `moby/api/types/network.Port` and lost `.Int()`:

```
pkg/repo/containers_test.go:70:38: port.Int undefined
    (type "github.com/moby/moby/api/types/network".Port has no field or method Int)
```

`eventindexer--test.yml` skips branches starting with `dependabot`, so the bump never ran this suite and the break landed unnoticed — the whole `pkg/repo` package (every `TestIntegration_*`) has been silently unbuildable for two weeks. One-line fix here (`port.Int()` → `port.Num()`) because the regression tests below cannot otherwise run. Worth a separate look at that CI exclusion.

## Tests

`TestIntegration_NFTBalance_RestartReplayDoesNotDoubleCount` walks the exact scenario: mint to alice, transfer alice → bob, replay the same log, then a genuinely new bob → carol log to prove the claim is not over-blocking.

Mutation-verified — with the guard removed the test fails on precisely the predicted corruption:

```
Error: Expected nil, but got: &eventindexer.NFTBalance{ID:2, ChainID:1,
       Address:"0xbob", Amount:2, TokenID:1, ContractAddress:"0xnftcontract",
       ContractType:"ERC721"}
Error: Not equal: expected: 1, actual: 2
```

`TestIntegration_NFTBalance_BatchIndexSeparatesUnitsOfOneLog` covers the ERC-1155 batch case, and `TestIntegration_ERC20Balance_RestartReplayDoesNotDoubleCount` mirrors the first for ERC20.

`TestIntegration_UnclaimedBalanceReplayRisk` covers the bootstrap guard across seven cases including per-kind scoping. Mutation-verified: degrading it to the coarse "balances non-empty" predicate fails exactly the `balances_predate_the_block_being_resumed_from` case.

`TestIntegration_TransferLog_ReplayDetectedWithClientFoundRows` pins the driver setting described above: `testMysql` now takes extra DSN params, and this test runs both the raw claim and a full replay against `clientFoundRows=true`. Also mutation-verified — swap `Insert{Modifier: "IGNORE"}` back for `OnConflict{DoNothing: true}` and it fails with bob at `Amount:2`.

Full suite green locally (the CI command), `golangci-lint` clean:

```
ok  github.com/taikoxyz/taiko-mono/packages/eventindexer/api        0.659s
ok  github.com/taikoxyz/taiko-mono/packages/eventindexer/indexer    1.276s
ok  github.com/taikoxyz/taiko-mono/packages/eventindexer/pkg/http   1.827s
ok  github.com/taikoxyz/taiko-mono/packages/eventindexer/pkg/metrics 3.329s
ok  github.com/taikoxyz/taiko-mono/packages/eventindexer/pkg/repo  235.146s
```

## Notes for review

- `processed_transfer_logs` grows without bound. At current mainnet L2 volume (1 ERC-721 `Transfer` per 100,000 blocks) that is nothing, but it will need pruning if NFT/ERC20 traffic ever picks up. Rows older than the deepest possible replay window are safe to drop.
- This does **not** address the historical `nft_balances` gap on mainnet (~9.58M blocks with NFT indexing off). That is separate work, tracked in taikoxyz/k8s-configs#1606.
- An alternative fix would be a cursor that does not depend on the `events` table. That is a larger change and still would not survive a crash mid-batch, which per-log identity does.

### DIFF
--- a/packages/eventindexer/cmd/flags/indexer.go
+++ b/packages/eventindexer/cmd/flags/indexer.go
@@ -67,6 +67,14 @@ var (
 		Category: indexerCategory,
 		EnvVars:  []string{"INDEX_NFTS"},
 	}
+	AllowUnclaimedBalanceReplay = &cli.BoolFlag{
+		Name: "allowUnclaimedBalanceReplay",
+		Usage: "Start even when balances predate the transfer log claim table, " +
+			"accepting a single double count on this restart",
+		Required: false,
+		Category: indexerCategory,
+		EnvVars:  []string{"ALLOW_UNCLAIMED_BALANCE_REPLAY"},
+	}
 	IndexERC20s = &cli.BoolFlag{
 		Name:     "indexERC20s",
 		Usage:    "Whether to index erc20 transfer events or not",
@@ -87,4 +95,5 @@ var IndexerFlags = MergeFlags(CommonFlags, []cli.Flag{
 	SyncMode,
 	IndexNFTs,
 	IndexERC20s,
+	AllowUnclaimedBalanceReplay,
 })
--- a/packages/eventindexer/erc20_balance.go
+++ b/packages/eventindexer/erc20_balance.go
@@ -38,6 +38,7 @@ type UpdateERC20BalanceOpts struct {
 type ERC20BalanceRepository interface {
 	IncreaseAndDecreaseBalancesInTx(
 		ctx context.Context,
+		ref TransferLogRef,
 		increaseOpts UpdateERC20BalanceOpts,
 		decreaseOpts UpdateERC20BalanceOpts,
 	) (increasedBalance *ERC20Balance, decreasedBalance *ERC20Balance, err error)
--- a/packages/eventindexer/indexer/config.go
+++ b/packages/eventindexer/indexer/config.go
@@ -30,31 +30,36 @@ type Config struct {
 	SyncMode                SyncMode
 	IndexNFTs               bool
 	IndexERC20s             bool
-	Layer                   string
-	OpenDBFunc              func() (db.DB, error)
+
+	// AllowUnclaimedBalanceReplay lets the indexer start even when
+	// UnclaimedBalanceReplayRisk reports that this restart may double count once.
+	AllowUnclaimedBalanceReplay bool
+	Layer                       string
+	OpenDBFunc                  func() (db.DB, error)
 }
 
 // NewConfigFromCliContext creates a new config instance from command line flags.
 func NewConfigFromCliContext(c *cli.Context) (*Config, error) {
 	return &Config{
-		DatabaseUsername:        c.String(flags.DatabaseUsername.Name),
-		DatabasePassword:        c.String(flags.DatabasePassword.Name),
-		DatabaseName:            c.String(flags.DatabaseName.Name),
-		DatabaseHost:            c.String(flags.DatabaseHost.Name),
-		DatabaseMaxIdleConns:    c.Uint64(flags.DatabaseMaxIdleConns.Name),
-		DatabaseMaxOpenConns:    c.Uint64(flags.DatabaseMaxOpenConns.Name),
-		DatabaseMaxConnLifetime: c.Uint64(flags.DatabaseConnMaxLifetime.Name),
-		MetricsHTTPPort:         c.Uint64(flags.MetricsHTTPPort.Name),
-		ETHClientTimeout:        c.Uint64(flags.ETHClientTimeout.Name),
-		L1TaikoAddress:          common.HexToAddress(c.String(flags.L1TaikoAddress.Name)),
-		BridgeAddress:           common.HexToAddress(c.String(flags.BridgeAddress.Name)),
-		BlockBatchSize:          c.Uint64(flags.BlockBatchSize.Name),
-		SubscriptionBackoff:     c.Uint64(flags.SubscriptionBackoff.Name),
-		RPCUrl:                  c.String(flags.IndexerRPCUrl.Name),
-		SyncMode:                SyncMode(c.String(flags.SyncMode.Name)),
-		IndexNFTs:               c.Bool(flags.IndexNFTs.Name),
-		IndexERC20s:             c.Bool(flags.IndexERC20s.Name),
-		Layer:                   c.String(flags.Layer.Name),
+		DatabaseUsername:            c.String(flags.DatabaseUsername.Name),
+		DatabasePassword:            c.String(flags.DatabasePassword.Name),
+		DatabaseName:                c.String(flags.DatabaseName.Name),
+		DatabaseHost:                c.String(flags.DatabaseHost.Name),
+		DatabaseMaxIdleConns:        c.Uint64(flags.DatabaseMaxIdleConns.Name),
+		DatabaseMaxOpenConns:        c.Uint64(flags.DatabaseMaxOpenConns.Name),
+		DatabaseMaxConnLifetime:     c.Uint64(flags.DatabaseConnMaxLifetime.Name),
+		MetricsHTTPPort:             c.Uint64(flags.MetricsHTTPPort.Name),
+		ETHClientTimeout:            c.Uint64(flags.ETHClientTimeout.Name),
+		L1TaikoAddress:              common.HexToAddress(c.String(flags.L1TaikoAddress.Name)),
+		BridgeAddress:               common.HexToAddress(c.String(flags.BridgeAddress.Name)),
+		BlockBatchSize:              c.Uint64(flags.BlockBatchSize.Name),
+		SubscriptionBackoff:         c.Uint64(flags.SubscriptionBackoff.Name),
+		RPCUrl:                      c.String(flags.IndexerRPCUrl.Name),
+		SyncMode:                    SyncMode(c.String(flags.SyncMode.Name)),
+		IndexNFTs:                   c.Bool(flags.IndexNFTs.Name),
+		AllowUnclaimedBalanceReplay: c.Bool(flags.AllowUnclaimedBalanceReplay.Name),
+		IndexERC20s:                 c.Bool(flags.IndexERC20s.Name),
+		Layer:                       c.String(flags.Layer.Name),
 		OpenDBFunc: func() (db.DB, error) {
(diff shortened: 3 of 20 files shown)


smtmfft (commented):
I have one rollout-safety concern, so I am not approving this yet.

The new claim table only protects logs after they have been claimed once. On the first restart after this deploys, `processed_transfer_logs` starts empty, while `setInitialIndexingBlockByMode` still rewinds from `MAX(events.emitted_block_id) - 1` and `filter()` resumes at `cursor + 1`. For an environment that already has ERC20/NFT balance rows and has balance indexing enabled, that deployment restart will replay pre-upgrade transfer logs, insert fresh claims for them, and apply the balance mutation again once.

The PR description acknowledges this as a known gap, but it is still a silent data-corruption mode triggered by the deploy/restart itself. I think we should either add an explicit mitigation/guard in code, or tie this PR to a concrete rollout step that prevents balance indexing from processing the replay range until claims are backfilled/reset. For example, fail fast when balance indexing is enabled, the relevant balance table is non-empty, and `processed_transfer_logs` is empty; or provide a one-time migration/backfill/controlled reset path for the affected environment.

The per-log claim implementation itself looks directionally right, and the new regression tests cover the replay case after the table is populated. My concern is specifically the first upgraded process start against existing balance data.

davidtaikocha:
Guard added in da0702f. The gap is real — I flagged it in the description but you're right that "documented" is not a mitigation for a silent corruption mode triggered by the deploy itself.

`UnclaimedBalanceReplayRisk` refuses to start when a kind has no claims for the chain **and** the balance table was last written at or after the block the cursor resumes from. `Indexer.Start()` calls it before `setInitialIndexingBlockByMode`, so it fails before any RPC work.

I deviated from your exact predicate on one point, and it matters. You proposed *"balance indexing is enabled, the relevant balance table is non-empty, and `processed_transfer_logs` is empty"*. Existence alone false-positives on the deploy this whole chain of work is aimed at: mainnet has `nft_balances` rows written before 2025-05-31 and NFT indexing off ever since, so re-enabling it (taikoxyz/k8s-configs#1606) would trip the guard for nothing — the replay range is entirely newer than anything the old process wrote, and cannot be double counted. A guard that cries wolf on its first real deploy just teaches people to set the override.

So the signal is recency instead:

```
claims(chain, kind) == 0
AND MAX(balance.updated_at) >= transacted_at(MAX(events.emitted_block_id))
```

Balances written at or after the block being resumed from mean a process was applying them *inside* the range about to be replayed. `events.transacted_at` gives the resume block's time without an extra RPC. Balance `updated_at` is DB wall-clock and lags block time slightly, which biases toward firing — the safe direction.

Escape hatches are `--allowUnclaimedBalanceReplay` / `ALLOW_UNCLAIMED_BALANCE_REPLAY=true` (logs a warning and proceeds, accepting one overcount), or resetting the affected balances so they rebuild from claims.

`TestIntegration_UnclaimedBalanceReplayRisk` covers seven cases including per-kind scoping (an erc20 claim must not clear nft risk). Mutation-verified against your version specifically: degrading the check to the coarse "balances non-empty" predicate fails exactly one case — `balances_predate_the_block_being_resumed_from`.

Environments as they stand: mainnet has both flags off, so nothing fires. hoodi runs `INDEX_NFTS=true` + `INDEX_ERC20S=true` with live balances and will trip the guard on the upgrade restart — that is the intended outcome, and it needs a decision at deploy time rather than a silent overcount.

Unrelated, but it came up while rebasing: #22063 fixed `port.Int()` → `port.Num()` in `packages/relayer/pkg/repo/containers_test.go`. Same testcontainers v0.44 break this PR fixes in `packages/eventindexer`, so that one was not a one-off — the dependabot CI exclusion is worth its own issue.

Full suite green, `golangci-lint` clean.

smtmfft (approved):
Reviewed the updated head (`da0702f19f8856715ce4729c3fdc9f39e1892fda`).

The first-upgrade replay concern I raised on the previous head is addressed now. `Indexer.Start()` checks the enabled balance kinds before setting the replay cursor, and `UnclaimedBalanceReplayRisk` makes the dangerous case fail closed when the claim table is empty and existing balances were written at or after the resume block timestamp. That turns the deploy-time silent corruption path into an explicit operator decision/reset path.

I also re-checked the transfer-log claim path: the claim is inserted with `INSERT IGNORE` inside the same transaction as the balance mutation, and duplicate claims skip both increase/decrease. ERC1155 batch units are separated by `batch_index`, and ERC20/NFT namespaces are separated by `kind`.

Local verification:
- `git diff --check base-main...HEAD`
- `gofmt -l ...` on the touched Go files
- `go test ./pkg/repo -run 'TestIntegration_UnclaimedBalanceReplayRisk|TestIntegration_TransferLog_ReplayDetectedWithClientFoundRows|TestIntegration_NFTBalance_RestartReplayDoesNotDoubleCount|TestIntegration_NFTBalance_BatchIndexSeparatesUnitsOfOneLog|TestIntegration_ERC20Balance_RestartReplayDoesNotDoubleCount'`
- `go test ./...` from `packages/eventindexer`

CI has `lint-eventindexer` and `test-eventindexer` passing on this head. No blocking concerns from my side.

PR #22075, opened 2026-08-31
fix(relayer): prevent crawler block range underflow

davidtaikocha:
## What

- saturate both crawler window subtractions when the source-chain head is lower than the configured window
  - the start window (`numLatestBlocksStartWhenCrawling`) no longer wraps to ~2^64
  - the end window (`numLatestBlocksEndWhenCrawling`) now clamps to 0 instead of declining to subtract
- retain the Shasta genesis or block-zero floor, and leave the mature-chain path bit-identical
- add a table-driven regression test covering both windows across the mature, exact-boundary, shorter-than-window and unconfigured regimes

## Why

On internal-devnet, `relayer-crawler-l1-to-l2` remained Running and Ready while logging `latestIndexedBlockNumber` near `2^64`. The L1 head was below the default 50,400-block crawler lookback, so the unsigned subtraction wrapped and the batch loop never executed — the crawler skipped all useful historical work.

Fixing the start window made the low-height path reachable for the first time, which exposed the same defect one line down. The end-exclusion guard avoided its own underflow by declining to subtract at all:

```go
if endBlockID > i.numLatestBlocksEndWhenCrawling {
    endBlockID -= i.numLatestBlocksEndWhenCrawling
}
```

At low height that means "do not skip the recent blocks" — the opposite of the knob's intent. With the production default `numLatestBlocksEndWhenCrawling=300` and a head of 10, `endBlockID` stayed at 10 and the crawler indexed straight through the head block, re-opening the double-queue race the exclusion window exists to prevent. The `>` was also wrong at the exact boundary `head == N`, where every block is inside the window and the answer should be 0. Both are now saturated the same way.

## Validation

- both halves mutation-verified: reverting the start-window clamp reproduces the wrapped cursor (`expected: 0x7, actual: 0xffffffffffff3b2a`); reverting the end-window clamp fails exactly the two end-window cases and nothing else
- `go test ./packages/relayer/indexer -run TestFilterCrawlPastBlocksClampsCrawlWindowsToAvailableHistory -count=1 -v` — 6/6 subtests pass
- full `go test ./...` for `packages/relayer` — all packages pass
- `go vet ./...`, `gofmt -l`, and `golangci-lint run --config=.golangci.yml --timeout=4m` — 0 issues

## Rollout

No Kubernetes or configuration change is included. A relayer image build and internal-devnet rollout are separate follow-up actions.


smtmfft (approved):
Reviewed exact head `ed31bb829366e30b58271623782242754cc48302` and its effective merge with current `main` (`e93f7828ee6268059ee40ed9e7d6ed0f25f2d159`). No blocking concerns.

Both crawler-window subtractions now saturate correctly at low height and exact boundaries while preserving mature-chain behavior. The automated warning about a persisted wrapped cursor does not apply: crawl mode forces `Resync`, and each non-target poll resets this in-memory cursor before applying the window. Focused regression coverage and the full relayer Go suite passed on the effective merge; current relayer CI and lint are green.

PR #22077, opened 2026-08-31
fix(protocol): budget the bridge's Ether sends for EIP-8037 state-creation repricing

dantaik:
## Problem

EIP-8037 (state-creation repricing, scheduled for Glamsterdam) makes creating a fresh storage slot cost **64 state bytes × 1,530 gas/byte = 97,920 gas** (vs 22,100 today). State gas is charged to a transaction-level reservoir first, but the reservoir is funded only by transaction gas *above* the 16.7M per-transaction execution cap — so for every realistic claim transaction the reservoir is empty and the charge falls through to **the current call frame's own gas**.

The bridge sends all capped Ether transfers with `_SEND_ETHER_GAS_LIMIT = 35_000`:

- `processMessage` → fee-remainder refund to `destOwner` (reverts the whole claim on failure, on **both** the relayer path and the self-claim path, for any nonzero-fee message);
- `processMessage` → relayer fee payout to `msg.sender` (reverts on failure);
- `recallMessage` → value back to `srcOwner` (reverts on failure);
- `retryMessage` → value to `destOwner` on the invocation-prohibited branch (non-reverting).

Post-Glamsterdam, a smart-wallet `destOwner` that writes a single fresh storage slot when receiving Ether would run out of gas inside the 35k cap. Because the failed send reverts `processMessage` before the status update, the message stays `NEW` forever: the relayer can't process it, and the owner can't self-claim either (the full fee still routes to `destOwner` through the same capped send). The bridged Ether would be unclaimable. The existing comment on the constant only reasoned about EIP-8038 (state-*access* repricing, caller-side costs) and predates 8037's finalization.

## Fix

Raise `_SEND_ETHER_GAS_LIMIT` from 35,000 to **135,000** = the legacy 35,000 callee budget + one EIP-8037 slot-creation charge (97,920), rounded up. This preserves the same guarantee across the fork: every wallet that fit in the old cap still fits, provided its receive path creates at most one fresh storage slot (the old cap likewise only ever covered one 22.1k slot plus overhead). On chains before the fork the extra headroom is inert for honest recipients, and the cap continues to bound how much relayer gas a malicious recipient can burn.

### Alternatives considered

- **Non-reverting refunds with a claimable-credit mapping (pull pattern)** — removes gas assumptions entirely and survives any future repricing, but adds storage to an upgradeable contract, a new external claim function, and new audit surface. Worth considering as a follow-up if we want zero coupling to gas schedules.
- **Forwarding all remaining gas on refunds** — rejected: unbounded relayer gas-griefing by a malicious `destOwner`.
- **Relying on claimers to buy reservoir gas (tx gas above the 16.7M execution cap)** — rejected: forces every relayer and self-claimer to front ~16.7M gas × basefee per claim and breaks standard estimation flows.

### Follow-ups (out of scope here)

`GAS_RESERVE` (800k), `GAS_OVERHEAD` (120k) and `_GAS_REFUND_PER_CACHE_OPERATION` (20k) are relayer-fee calibration constants explicitly meant to be retuned with production data; Glamsterdam inflates the bridge's own fresh-slot writes (message status, signal cache) too, so they should be recalibrated from Platåberget/testnet measurements before the fork. Those affect relayer economics, not claimability. Bridge UI / relayer default `gasLimit` suggestions for messages invoking contract recipients will also need raising, since invocation gas is user-chosen per message.

## Tests

New helper `MessageReceiver_CreatingFreshStorageSlots` writes N fresh slots per Ether receive. With N=4 (~112k gas today, representative of a one-fresh-slot wallet post-8037: 97,920 + overhead — and well above the old 35k cap, so these tests fail on `main`'s constant):

- `test_bridge2_processMessage__refund_to_storage_creating_wallet` — pure capped-refund path (invocation-prohibited target) delivers full value to the wallet;
- `test_bridge2_processMessage__storage_creating_wallet_claims_with_fee` — end-to-end relayer claim: invocation pays value, capped send refunds the fee remainder, relayer keeps its cut, balances conserve;
- `test_bridge2_recallMessage_storage_creating_wallet_srcOwner` — recall returns value to a storage-writing `srcOwner`;
- `test_bridge2_processMessage__refund_receiver_exceeding_gas_cap` — N=7 (~179k) still fails with `ETH_TRANSFER_FAILED` and leaves the message `NEW`, pinning the cap from above.

https://claude.ai/code/session_01LbPMYpG2UWajo5ehLdmAvi

---
_Generated by [Claude Code](https://claude.ai/code/session_01LbPMYpG2UWajo5ehLdmAvi)_

### DIFF
--- a/packages/protocol/contracts/shared/bridge/Bridge.sol
+++ b/packages/protocol/contracts/shared/bridge/Bridge.sol
@@ -61,11 +61,21 @@ contract Bridge is EssentialResolverContract, IBridge {
     /// suggest. State-access repricing forks such as EIP-8038 raise mostly caller-side costs
     /// (value-transfer account write, cold-recipient access) that are paid by this contract
     /// before forwarding and do not draw on this budget, so no headroom bump is needed for them.
+    /// State-creation repricing (EIP-8037, scheduled for Glamsterdam) does draw on this budget:
+    /// creating a fresh storage slot charges 64 state bytes * 1,530 gas/byte = 97,920 gas, and
+    /// unless the transaction buys gas beyond the 16.7M per-transaction execution cap (which no
+    /// realistic claim transaction does, leaving its state-gas reservoir empty), that charge is
+    /// deducted from the callee frame's own gas. A smart wallet that writes a single fresh slot
+    /// when receiving Ether would then run out of gas under the previous 35,000 cap, and since a
+    /// failed send reverts processing, its messages would become unclaimable. The cap is
+    /// therefore the legacy 35,000 callee budget plus one EIP-8037 slot-creation charge
+    /// (97,920), rounded up — wallets that fit before the fork still fit after it, as long as
+    /// their receive path creates at most one storage slot.
     // - EOA gas used is < 21000
     // - For Loopring smart wallet, gas used is about 23000
     // - For Argent smart wallet on Ethereum, gas used is about 24000
     // - For Gnosis Safe wallet, gas used is about 28000
-    uint256 private constant _SEND_ETHER_GAS_LIMIT = 35_000;
+    uint256 private constant _SEND_ETHER_GAS_LIMIT = 135_000;
 
     /// @dev Place holder value stored in the context slots between message invocations.
     uint256 private constant _PLACEHOLDER = type(uint256).max;
--- a/packages/protocol/test/shared/bridge/Bridge2_processMessage.t.sol
+++ b/packages/protocol/test/shared/bridge/Bridge2_processMessage.t.sol
@@ -442,6 +442,8 @@ contract TestBridge2_processMessage is TestBridge2Base {
         message.fee = 5_000_000;
         message.value = 2 ether;
         message.destOwner = address(wallet);
+        // With empty message.data the invocation is a plain value-bearing call that hits the
+        // wallet's receive() — no onMessageInvocation implementation is required.
         message.to = address(wallet);
 
         eBridge.processMessage(message, FAKE_PROOF);
--- a/packages/protocol/test/shared/bridge/Bridge2_processMessage.t.sol
+++ b/packages/protocol/test/shared/bridge/Bridge2_processMessage.t.sol
@@ -2,7 +2,9 @@
 pragma solidity ^0.8.24;
 
 import "./TestBridge2Base.sol";
-import "test/shared/bridge/helpers/MessageReceiver_CreatingFreshStorageSlots.sol";
+import {
+    MessageReceiver_CreatingFreshStorageSlots
+} from "test/shared/bridge/helpers/MessageReceiver_CreatingFreshStorageSlots.sol";
 
 contract Target is IMessageInvocable {
     uint256 public receivedEther;
--- a/packages/protocol/test/shared/bridge/Bridge2_processMessage.t.sol
+++ b/packages/protocol/test/shared/bridge/Bridge2_processMessage.t.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.24;
 
 import "./TestBridge2Base.sol";
+import "test/shared/bridge/helpers/MessageReceiver_CreatingFreshStorageSlots.sol";
 
 contract Target is IMessageInvocable {
     uint256 public receivedEther;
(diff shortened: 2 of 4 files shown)


dantaik:
Re: the DeepSeek review — triaged each point:

**"Helper doesn't implement `IMessageInvocable`, so the fee-claim test would revert or isn't exercising invocation"** — incorrect. The test sends **empty `message.data`**, and `_unableToInvokeMessageCall` only prohibits calls when `data.length >= 4` with a selector other than `onMessageInvocation`. With empty calldata, `_invokeMessageCall` performs a plain value-bearing call that lands in the wallet's `receive()` — which is exactly how bridged ETH reaches a smart wallet in production. The test's `assertEq(wallet.receiveCount(), 2)` proves both the invocation receive and the refund receive execute (traces show 111,292 and 89,392 gas respectively). Added a comment in the test making this explicit (`7ca4379`).

**"Tests aren't fork-aware"** — fair as a maintenance note: the slot counts are deliberately calibrated to the pre-Glamsterdam schedule (4 slots ≈ one post-8037 slot). The helper now documents how to recalibrate (4 → 1, 7 → 2) if/when the test EVM adopts EIP-8037 pricing (`7ca4379`).

**"Use N=5 to pin the cap nearer the boundary"** — declined: N=5 measures ~134.6k on first receive, within ~1k of the 135k cap, which makes the test flaky across solc/optimizer changes. The current pair (111k passes / ~179k fails) brackets the cap robustly.

**"Add a `retryMessage` capped-send test"** — declined: that branch (`_unableToInvokeMessageCall` == true inside `retryMessage`) is unreachable for messages processed under the current rules — prohibited-invocation messages are marked `DONE` in `processMessage` and never become `RETRIABLE`. It exists only for legacy messages left `RETRIABLE` under pre-upgrade prohibition rules, would require `vm.store` surgery on the status mapping to construct, and is non-reverting (`sendEther`, not `sendEtherAndVerify`), so it was never a stuck-funds vector — the cap bump only makes it more likely to deliver.

**Fee-calibration constants (`GAS_RESERVE`, `GAS_OVERHEAD`, `_GAS_REFUND_PER_CACHE_OPERATION`)** — acknowledged; intentionally out of scope, per the PR description. They affect relayer economics (relayers simulate off-chain), not claimability, and are explicitly marked for retuning with production/testnet data closer to the fork. Shipping the cap early is safe: pre-fork it is inert for honest recipients.

---

davidtaikocha:
## Review summary — max-effort pass

**Verdict first: `135_000` is the right number, and this is mergeable.** I verified the sizing against the EIP-8037 spec itself: `CPSB = 1,530` × 64 state bytes = **97,920** per fresh slot, charged to the callee frame when the reservoir is empty — exactly as the comment reasons. One refinement: the spec *replaces* the legacy 20,000 SSTORE-set execution cost (`GAS_STORAGE_SET`: 20,000 → 0 execution + 97,920 state), so the true preservation requirement is `35,000 − 20,000 + 97,920 = 112,920`, and 135,000 carries **~22k of real headroom** — more conservative than the additive "132,920 rounded up" derivation suggests. The tests genuinely pin the behavior: on `main`'s 35,000 all three success-path tests fail with `ETH_TRANSFER_FAILED` (mutation-verified), and the helper's gas margins are compiler-stable (≤132 gas spread across optimizer/`evm_version` sweeps).

Everything below was produced by a 10-angle review with adversarial verification of each candidate; all load-bearing numbers are **measured** (traces, bisection, mutation runs), not estimated. None of it challenges the constant — the findings cluster around the deferral framing, comment/fixture accuracy, and test tightness.

### Main asks

1. **`GAS_OVERHEAD = 120_000` breaks on day one, not at the fork** (`Bridge.sol:49` × `:353-370`). `gasUsedInFeeCalc` is snapshotted *before* both capped sends, so a hostile `destOwner` can now burn 137,300 (was 37,300) that is never charged to the fee — and the honest-case slack in `GAS_OVERHEAD` is only ~7k today (measured on this PR's own fee test: `gasCharged = 275,729` while the refund receive burned 89,392 after the snapshot, unreimbursed). The grief is near-free for the attacker (the unspent fee refunds to their own `destOwner`) and `ProfitableOnly` defaults to `false` in the relayer. Post-8037 it stops being adversarial-only: the status write becomes a 97,920 fresh-slot charge *after* the snapshot, so the fixed tail (~117k) consumes essentially all of `GAS_OVERHEAD` on every claim, EOA recipients included. The follow-ups paragraph should therefore (a) state that the widening lands at merge time on today's schedule, and (b) give the recalibration a hard *with-or-before-the-fork* deadline. Two factual fixes in the same paragraph: `GAS_RESERVE` is not a "relayer-fee calibration constant" (it never enters the fee formula — only `getMessageMinGasLimit`/`_invocationGasLimit`), and `_GAS_REFUND_PER_CACHE_OPERATION` is dead code (`SignalService.proveSignalReceived` unconditionally `return 0`, so `numCacheOps` is always 0).

2. **The helper's fork-recalibration NOTE is off by one and would turn the safety tests red at exactly the wrong moment** (`MessageReceiver_CreatingFreshStorageSlots.sol:10-12`). A first receive creates `numSlots + 1` fresh slots (the `receiveCount` slot is fresh too, as the file's own lines 8-9 say). Following "4 → 1" gives 2 × 97,920 = 195,840 against the 137,300 callee budget, so both must-pass first-receive tests revert `ETH_TRANSFER_FAILED` — and the natural misread of a red gas test is "cap too small", inviting a spurious bump. Correct guidance is non-uniform: **4 → 0** for the two first-receive tests, **1** for the fee test's warm-counter second receive, and **7 → 1** for a tight upper pin.

3. **A one-line, in-idiom escape hatch would remove the fund-loss cliff entirely — worth taking now** (`Bridge.sol:370`, also `:268`). The cap exists to protect a third-party relayer, but it is applied uniformly to the self-claim refund (`msg.sender == destOwner`) and the srcOwner self-recall, where the caller can only grief itself. `stats.processedByRelayer ? _SEND_ETHER_GAS_LIMIT : gasleft()` mirrors line 321 exactly, needs no storage, and makes any wallet self-rescuable under any future repricing (the N=7 test is itself proof the cliff survives the bump: that wallet is unclaimable by anyone, owner included). The body's rejection of gas forwarding ("unbounded relayer gas-griefing") is refuted by the codebase's own dominant path: vault-originated recalls already forward **all** gas to an arbitrary `srcOwner` via permissionless `recallMessage` → `onMessageRecalled` → the vaults' uncapped `sendEtherAndVerify` — the line-268 cap is the outlier. Verified safe: all five mutating entry points share the transient reentrancy lock, nothing in-repo reads `messageStatus` mid-flight, and a greedy callee only wastes its own tx. Companion: hoist `_updateMessageStatus` to between lines 365 and 370 (matching `recallMessage`'s write-before-call ordering; hoisting above 337 would pull the SSTORE into the fee snapshot).

4. **The constant's comment misstates the guarantee in both directions** (`Bridge.sol:64-73`). (a) "At most one storage slot" is not a sufficient condition: EIP-8037 prices *all* callee-frame state creation. Harness-verified counterexample: a wallet whose `receive()` forwards value to a never-existing address costs 36,600 today (fits 37,300, zero slots) but needs ≥183,600 post-fork (`STATE_BYTES_PER_NEW_ACCOUNT` = 120 × 1,530, charged in the wallet's frame); same for `CREATE` in receive. Add "…and creates no other new state (no transfers to fresh addresses, no CREATE)". (b) The additive derivation double-counts the replaced 20,000 — restating on the replace model (requirement 112,920, headroom ~22k) makes the margin argument *stronger*. The helper docstring's "a single fresh slot costs 97,920" is also short by the surviving 2,100 access charge (≈100,020 total).

### Test tightening

5. **The suite doesn't pin the constant: any cap in [108,992, 175,720] is fully green** (mutation-bisected; 120,000 — below the documented 132,920 derivation — and even 110,000 — below the real 112,920 floor — pass all 159 shared tests). The earlier N=5 flakiness rebuttal doesn't hold up empirically: N=5's true minimum is **131,235** (margin 3,765, spread 87 gas across compiler settings), not "~134.6k within 1k". The one-character `4 → 5` (with `7 → 6` on the fail side) closes 93% of the window.

6. **The flagship fee test's capped leg only exercises 89,392 gas** — below the 97,920 single-slot charge the PR is calibrated to. The heavy 111,292 first receive goes through the *uncapped* invocation (`_invocationGasLimit = 193,344`); the capped refund hits a warm `receiveCount`. The test passes with the cap as low as ~90k, and the docstring's "~112k" describes the wrong leg. Fix: fresh wallet for the refund leg, or stride `receiveCount` fresh per receive.

7. **The fee test is silently coupled to `GAS_RESERVE`**: bisected, `GAS_RESERVE ≥ 884,353` makes exactly this one test fail (invocation OOG → RETRIABLE), with a symptom pointing at the send cap — and raising `GAS_RESERVE` is this PR's own stated follow-up. Derive the message gasLimit from `getMessageMinGasLimit()` as `Bridge2_failMessage.t.sol:40` already does.

8. **All four new tests run the relayer path under unlimited tx gas, with `getQuotaManager() → address(0)`.** The self-claim path (which ignores `message.gasLimit` and the documented 102% rule entirely; measured worst case ~9.47M tx gas to fund the tail — an expansion over `main`, not a regression, but untested) deserves one bounded-gas test, and `Bridge2_quotaAccounting.t.sol` is the natural home for a storage-creating-`destOwner` variant with a real quota manager. `retryMessage:392` — the fourth changed site — has no test (defensible: reachable only for pre-#20939 legacy messages, and the flip there is benign-to-positive, but then the body's "four sites" framing ships with three tested).

9. Minor: the recall test misses value-conservation (a 1-wei leak mutation passes it while the file's `getBalanceForAccounts` idiom catches it — two lines to add); the negative test name omits the `RevertWhen_` infix that `packages/protocol/CLAUDE.md` mandates (mixed local precedent, your call); the post-`expectRevert` asserts at :486-487 are vacuous (all discriminating power is in the `expectRevert` itself — fine to keep, just not coverage).

### Out of scope, worth a separate issue

10. **A second instance of this PR's own failure mode sits one line above the sends** (`Bridge.sol:335`, pre-existing): any Ether claim with `value + fee` above the QuotaManager ceiling (250 ETH on mainnet) reverts `QM_OUT_OF_QUOTA` *before* the status write — `availableQuota` is capped at `q.quota`, so waiting never helps — wedging the message at NEW with no FAILED signal and no recall path until the owner raises the ceiling. The bridge-ui compounds it (`checkQuota.ts` ignores `fee` and compares only against `availableQuota`; the UI copy says "wait for the quota to replenish", which is false and unbounded here). The dead `StatusReason.OUT_OF_ETH_QUOTA` enum suggests a graceful path was once intended.

### Deployment note

11. The constant is bytecode-baked: it takes effect only via impl upgrades of **both** Bridge proxies, and the L1 one must land **before Glamsterdam** (mainnet target Nov 4, 2026) or the exact failure this PR describes materializes until a later upgrade. Worth one line in the body naming the proxies and the deadline — `grep -r Glamsterdam` outside this PR's files currently returns nothing.

### Checked and clean (so nobody re-litigates)

- Relayer/bridge-ui gas-padding "mis-keying" on `Message.To` vs `destOwner` is harmless: the txmgr always `CallContract`-preflights the exact fixed gas limit (fail-closed pre-broadcast) and the UI simulates before signing.
- The widened refund callback adds no reentrancy capability: the invocation call already runs in the same `status == NEW` window with millions of attacker-authored gas, and the global transient lock covers every mutating entry point.
- No other capped or stipend-limited Ether send exists anywhere in `contracts/` (vaults/Inbox/Anchor all use the `gasleft()` overload); no stale 35k references anywhere; no gas-snapshot/CI artifact is invalidated; changelog is release-please; the earlier named-imports feedback is resolved.

davidtaikocha (approved):

PR #21697, opened 2026-05-18
chore(eventindexer): remove legacy L1 fork indexing code

YoGhurt111:
## Summary

This PR removes legacy L1 fork-specific indexing support from `packages/eventindexer` for the pre-Shasta TaikoL1 flows, including Ontake and Pacaya.

The indexer no longer creates or stores the legacy `TaikoL1` binding, no longer exposes the `l1TaikoAddress`, Ontake fork height, or Pacaya fork height CLI/config entries, and no longer switches across default, Ontake, or Pacaya filter paths. It also deletes the legacy TaikoL1 event save handlers, the Ontake v2 binding, the Pacaya `TaikoInbox` binding, and the fork-specific abigen scripts.

The remaining runtime indexing paths are bridge events, raw block data, and Shasta inbox events. The repository layer keeps its historical `BlockProposed` / `TransitionProved` and `BatchProposed` / `BatchesProved` query compatibility so existing stored data can still be read by APIs that fall back to older event names.

## Validation

- `go test ./packages/eventindexer/... -run '^$'`
- `go test ./packages/eventindexer/api ./packages/eventindexer/indexer ./packages/eventindexer/pkg/http ./packages/eventindexer/pkg/metrics`
- `git diff --check`

Full `go test ./packages/eventindexer/...` still fails in `packages/eventindexer/pkg/repo` because the local testcontainers setup cannot find rootless Docker (`rootless Docker not found`). This same failure was present before the code changes.

PR #22078, opened 2026-08-31
fix(eventindexer): restore layer-aware inbox startup

### LINKED ISSUE #22076: eventindexer: indexer cannot start on v0.15.0+ — l1TaikoAddress is now required and is the only inbox source (#21697 regression)
## Summary

Since #21697 (`a78580e5c`, 2026-05-21, *"chore(eventindexer): remove legacy L1 fork indexing code"*) the eventindexer **cannot start as an indexer** in any of our environments. It has gone unnoticed for three months because every deployment is still pinned to a pre-#21697 image — `eventindexer-v0.15.0` was the first release containing the change and was never rolled out.

Discovered while rolling `eventindexer:sha-45fe0a7` onto internal-devnet (taikoxyz/k8s-configs#1606). Both indexer pods went straight to `CrashLoopBackOff`:

```
Required flag "l1TaikoAddress" not set
```

The `api` command is unaffected — those pods rolled out cleanly and serve normally.

## What changed

#21697 did three things together:

1. deleted the `shastaInboxAddress` flag and its `SHASTA_INBOX_ADDRESS` env var
2. flipped `l1TaikoAddress` (`L1_TAIKO_ADDRESS`) from `Required: false` to `Required: true`
3. made it the **only** source of the inbox binding

```go
// indexer/indexer.go
if cfg.L1TaikoAddress.Hex() != ZeroAddress.Hex() {
    inboxContract, err = inbox.NewInbox(cfg.L1TaikoAddress, ethClient)
}
```

and `setInitialIndexingBlockByMode` now calls it **unconditionally, for both layers**, before anything else:

```go
// indexer/set_initial_processing_block_height.go
func (i *Indexer) setInitialIndexingBlockByMode(ctx context.Context, mode SyncMode) error {
	startingBlock, err := i.getFirstShastaBlockHeight(ctx)   // ← always
	if err != nil {
		return err
	}
	switch mode {
	case Sync:
		latest, err := i.eventRepo.FindLatestBlockID(ctx, i.srcChainID)
		...
		if latest != 0 {
			startingBlock = latest - 1                        // ← and in Sync mode the result is then discarded
		}
```

```go
func (i *Indexer) getFirstShastaBlockHeight(ctx context.Context) (uint64, error) {
	if i.inbox == nil {
		return 0, errors.New("inbox contract not configured")
	}
	ts, err := i.inbox.ActivationTimestamp(nil)
	...
```

So startup now hard-requires an inbox contract that answers `activationTimestamp()` — and in `Sync` mode (the default) its value is thrown away immediately afterwards whenever the DB already has events.

### 1. The env-var contract changed and nothing was migrated

`SHASTA_INBOX_ADDRESS` was removed from the flag set but is still what our charts pass the Shasta inbox in, and it is still in this repo's own example env files:

```
packages/eventindexer/.default.env:11:SHASTA_INBOX_ADDRESS=
packages/eventindexer/.l1.env:11:SHASTA_INBOX_ADDRESS=
```

Result across our three environments on `sha-45fe0a7` — the addresses that are actually configured as `l1TaikoAddress` do not answer the call the binary now makes:

| deployment | configured `L1_TAIKO_ADDRESS` | outcome |
|---|---|---|
| internal-devnet l1 + l2 | absent | `Required flag "l1TaikoAddress" not set` (observed) |
| hoodi l2 | absent (chart sets a `l2_TAIKO_ADDRESS` the binary never reads) | same |
| hoodi l1 | `0xf6eA848c7d7aC83de84db45Ae28EAbf377fe0eF9` | `activationTimestamp()` reverts |
| mainnet l2 | `""` | `inbox contract not configured` |
| mainnet l1 | `0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a` (dormant Pacaya inbox) | `activationTimestamp()` reverts |

Only the Shasta inboxes answer, and those are exactly the values being passed via the removed variable:

```
$ cast call 0x6f21C543a4aF5189eBdb0723827577e1EF57ef1f "activationTimestamp()(uint64)" --rpc-url <mainnet L1>
1775135903
$ cast call 0xeF4bB7A442Bd68150A3aa61A6a097B86b91700BF "activationTimestamp()(uint64)" --rpc-url <hoodi L1>
1770296640
$ cast call 0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a "activationTimestamp()(uint64)" --rpc-url <mainnet L1>
Error: execution reverted
```

That part is arguably ours to fix in the charts — but it is a silent breaking change to the deployment contract, so it is worth a note in the release/migration guide either way.

### 2. An L2 indexer cannot satisfy this at all

This is the part we cannot fix from configuration. The inbox is bound to the indexer's **own** client:

```go
ethClient, err := ethclient.Dial(cfg.RPCUrl)
...
inboxContract, err = inbox.NewInbox(cfg.L1TaikoAddress, ethClient)
```

For an `l2` indexer `RPC_URL` is the L2 node, where an L1 inbox address has no code. So there is no value of `L1_TAIKO_ADDRESS` that lets an l2 indexer start:

- absent → `Required flag "l1TaikoAddress" not set`
- empty → `inbox contract not configured`
- the real L1 Shasta inbox → `ActivationTimestamp()` against the L2 RPC

Before #21697 an l2 indexer did not need this at all.

## Suggested directions

Not prescribing an implementation, but the shape seems to be one of:

- skip `getFirstShastaBlockHeight()` when `layer == l2`, and/or when `Sync` mode is going to resume from `FindLatestBlockID` anyway — the value is discarded in that path today
- give the inbox its own L1 client (a separate `L1_RPC_URL`) rather than binding it to `RPC_URL`
- keep `l1TaikoAddress` optional for `l2`, or reinstate a distinct inbox-address flag

Whatever the fix, `packages/eventindexer/.default.env` and `.l1.env` should stop advertising `SHASTA_INBOX_ADDRESS`.

## Environment

- broken: `eventindexer-v0.15.0` and later, including `sha-45fe0a7` (current `main`)
- last good: `eventindexer-v0.14.0` (does not contain `a78580e5c`); internal-devnet was on `sha-960c15b`, a 2026-01-23 build
- our deployments are being held on pre-#21697 images until this is resolved

davidtaikocha:
## Summary

- restore the dedicated `SHASTA_INBOX_ADDRESS` configuration removed by #21697
- let L2 indexers start from their checkpoint or block zero without an L1 Inbox
- skip `activationTimestamp()` when sync can resume from an existing checkpoint
- validate the L1 Inbox address, the sync mode, and the layer before opening the
  database or RPC client
- include the Shasta activation block itself in fresh and resync L1 scans

## Regression cause

#21697 intended to remove the legacy `l1TaikoAddress`, but the merged code kept
it as a required flag, removed `shastaInboxAddress`, and reused the retained
address as the Shasta Inbox binding. Startup then queried that L1 binding before
checking the layer or database checkpoint, which made L2 startup impossible and
made L1 startup depend on the wrong deployed address.

## Startup behavior

`latestIndexedBlockNumber` is the last block that has been filtered, and
filtering resumes at the block after it. The table below lists the first block
actually filtered; the stored cursor is one below that value.

| Layer | Mode | Checkpoint | First block filtered | Activation call |
| --- | --- | --- | --- | --- |
| L1 | sync | present | latest | no |
| L1 | sync | absent | Shasta activation block | yes |
| L1 | resync | ignored | Shasta activation block | yes |
| L2 | sync | present | latest | no |
| L2 | sync | absent | 1 | no |
| L2 | resync | ignored | 1 | no |

Current charts already provide `SHASTA_INBOX_ADDRESS`, so this fix requires no
chart migration. Deployments that still pass the removed `L1_TAIKO_ADDRESS` are
unaffected, because it reaches the binary as an environment variable rather than
a CLI flag and is simply ignored.

## Configuration validation

`Config.validate()` runs before `InitFromConfig` opens the database or dials an
RPC endpoint, and rejects three things:

- a sync mode that is neither `sync` nor `resync` (`ERR_INVALID_MODE`)
- a layer that is not exactly `l1` or `l2` (`ERR_INVALID_LAYER`)
- an L1 indexer with no Shasta Inbox address (`ERR_NO_SHASTA_INBOX_ADDRESS`)

The layer check matters because the Inbox binding is now selected by layer
rather than by address. Without it, any other `LAYER` value — including a
set-but-empty `LAYER=`, which overrides the flag's `l1` default — left the
indexer with no Inbox binding. A checkpointed `sync` would then start
successfully and silently never index Shasta `Proposed` or `Proved` events,
while the same configuration on an empty database failed loudly. Unsupported
layers are now a startup error in both cases.

## Shasta activation block

A fresh `sync` or any `resync` resolved the activation block `N` and stored it
as the cursor, so filtering began at `N + 1` and never requested `N` itself.
That block is not empty: `Inbox.activate` emits the genesis `Proposed` event
there, and because activation sets `lastProposalBlockId` to `1` while
`_buildProposal` only requires the current block to be greater than that value,
a regular `propose` may land in the same L1 block. Repeating `resync` recomputed
the same cursor, so it could not recover the omission.

The fallback now resolves to `N - 1`, guarded at zero so the cursor cannot
underflow — `getBlockByTimestamp` can return `0` through its
`targetTimestamp >= latestTimestamp` early return, and wrapping a `uint64` would
leave the filter loop unable to make progress.

This off-by-one predates this PR; `main` has the identical assignment and it
traces back through #21201. It is fixed here because the startup matrix
introduced by this PR would otherwise have pinned `N` as the expected cursor.

Whether existing deployments also need a backfill depends on which branch of
`LibInboxSetup.activate` set `activationTimestamp`. A self-stamped activation
(`block.timestamp`) puts `N` exactly on the activate block, so the genesis
`Proposed` was missed; a preset activation timestamp puts `N` at or before it,
in which case the first scan may already have covered activation. That is a
per-environment check against the deployed inbox, not a code question.

## Test plan

- `go test ./packages/eventindexer/indexer`
- `go test ./packages/eventindexer/... -run '^$'`
- `go vet ./packages/eventindexer/...`
- `golangci-lint run --config=.golangci.yml --timeout=10m` from `packages/eventindexer`
- `git diff --check origin/main...HEAD`

Startup selection is covered by a table-driven matrix over
layer x mode x checkpoint, including the zero-block guard and the assertion that
a checkpointed `sync` never invokes the Inbox resolver.
`TestFirstFilteredBlockIsShastaActivationBlock` composes cursor selection with
the filter's resume offset through `Indexer.nextFilterStartBlock`, so the first
requested block is asserted rather than the stored cursor.

Fixes #22076


smtmfft (changes requested):
Blocking: include the Shasta activation block in fresh/resync L1 scans

For an empty-DB `sync` or any `resync`, `initialIndexingBlockByMode` returns the activation block `N` from `firstShastaBlock`, and `setInitialIndexingBlockByMode` stores `N` in `latestIndexedBlockNumber`. The filter treats that field as the last processed cursor and starts at `N + 1`, so block `N` is never queried.

That block is not empty: [`Inbox.activate`](https://github.com/taikoxyz/taiko-mono/blob/d330cdbf3afc456b70081a05ca3535a1daaa285f/packages/protocol/contracts/layer1/core/impl/Inbox.sol#L188-L200) emits the genesis `Proposed` event there. A normal `propose` may also follow in the same L1 block because activation initializes `lastProposalBlockId` to `1`, while `_buildProposal` only requires the current block to be greater than that value. The event indexer has no special exclusion for either log, so fresh/resync startup silently omits data and repeating `resync` cannot recover it.

Please make the activation result a pre-scan cursor (`N - 1`, guarded for block zero), or otherwise make the first filter range include `N`. The new startup-matrix tests currently assert `N` and do not compose that result with the filter's `+1`; add a regression covering the first requested range / a `Proposed` log at `N`.

davidtaikocha:
Fixed in 91484600d.

Confirmed the whole chain rather than just the symptom:

- `filter()` resumes at `latestIndexedBlockNumber + 1`, and startup stored the activation block `N` in that field, so `N` was never requested.
- `Inbox.activate` emits the genesis `Proposed` in that block via `_emitProposedEvent(proposal)`.
- `LibInboxSetup.activate` sets `lastProposalBlockId = 1`, and `_buildProposal` only requires `block.number > _lastProposalBlockId`, so a regular `propose` can share the same L1 block. Both claims check out.
- `getBlockByTimestamp` returns the largest block with `time <= target`, so `N` is a block still to be scanned, not one already done.

The fallback now resolves to `N - 1`, guarded at zero. The guard matters for more than block 0: `getBlockByTimestamp` can return `0` through its `targetTimestamp >= latestTimestamp` early return on a chain holding only genesis, and `N - 1` on a `uint64` would wrap to 2^64-1, after which `filter`'s `j <= endBlockID` never holds and indexing silently no-ops.

On the test point — the matrix was asserting the cursor in isolation and never composing it with the `+1`. Rather than restate the offset in the test, I extracted `Indexer.nextFilterStartBlock()`, used it in `filter()`, and threaded the first-Shasta-block resolver through `setInitialIndexingBlockByMode` so `TestFirstFilteredBlockIsShastaActivationBlock` runs the real path from resolver to first requested block. Mutation-checked all three ways: restoring the old `return firstBlock`, dropping the zero guard, and changing the filter resume to `+2` each fail it. The `+2` case is the one the previous tests could not have caught.

Also added an `l1 resync at first block` row to the matrix for the zero guard, and documented the cursor contract on the field itself.

Two notes beyond the fix:

**Severity depends on how `activationTimestamp` was set.** `LibInboxSetup.activate` has two branches. Self-stamped (`_activationTimestamp == 0` -> `block.timestamp`) puts `N` exactly at the activate block, so the genesis `Proposed` is definitely lost. Preset (non-zero, `activate` called later in the window) stores the preset value, so `N <= activateBlock` and a strictly-smaller `N` would already have covered activation. Worth checking per environment whether the block whose timestamp equals `activationTimestamp()` actually contains `InboxActivated` — that decides whether mainnet/hoodi also need a backfill or whether this is only a going-forward fix. The fix is correct either way.

**This predates the PR.** `origin/main` has the identical assignment and it traces back through #21201's `getGenesisBlockHeight`. Fixing it here anyway, because the startup-matrix tests this PR adds were pinning `N` as the expected cursor — that turns an accidental off-by-one into a codified invariant, which is the stronger reason not to defer it.

Left the L2 fresh path returning `0` (scan starts at block 1). That matches pre-#21697 behavior and L2 genesis carries no logs, but flagging it as the same class in case you want the two paths made symmetric.

`go build`, `go vet`, `golangci-lint` (0 issues), and the full `indexer` package all clean.

smtmfft (approved):
Re-reviewed updated head `91484600d297ab66df87a2af0b6726ff3a65051b`. The previous activation-block blocker is fixed.

Fresh/resync L1 startup now converts activation block `N` to the last-processed cursor `N - 1`, with a zero guard, and `filter()` resumes through `nextFilterStartBlock()` at exactly `N`. The new regression composes the real cursor-selection and filter-resume path for both `sync` and `resync`, so it catches either side drifting again. Checkpointed sync and L2 behavior remain unchanged.

I also reviewed the complete PR diff and its current merge with `main` (`f5543c71adfab1fa46397c94ca568afe83ae89f9`). Eventindexer indexer tests, all-package compile checks, `go vet ./...`, diff checks, and current CI/lint passed. No remaining blocking concerns.

PR #22082, opened 2026-08-31
test(protocol): tighten the EIP-8037 send-cap tests and correct its comments

davidtaikocha:
## What this is

The zero-semantic-change subset of the review on #22077 ([review comment](https://github.com/taikoxyz/taiko-mono/pull/22077#issuecomment-5476345767)): comment corrections and test tightening. **`Bridge.sol` changes comments only** — verified mechanically, and that is the whole argument for why this needs no protocol review:

```
git diff main...HEAD -- packages/protocol/contracts/shared/bridge/Bridge.sol \
  | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*(///|$)'   # prints nothing
```

Stripping all comments from `Bridge.sol` at base and at tip yields byte-identical files.

## The main deliverable

Before this branch the whole suite passed for **any** `_SEND_ETHER_GAS_LIMIT` in `[108,992, 175,720]` — including values *below* what the constant exists to guarantee. Bisected, at both ends:

| | admissible cap values | width |
|---|---|---|
| `main` (159 tests) | `[108,992, 175,720]` | 66,729 |
| this branch, first pass (163 tests) | `[134,557, 135,557]` | 1,001 |
| this branch, after review (164 tests) | **`[134,957, 135,057]`** | **101** |

A 660× tightening. The old floor sat 13,928 gas *below* the post-fork requirement of 122,920; the new one clears it by 12,037.

(Counts are the whole `test/shared/**` suite — `pnpm test:shared` — not the bridge directory alone, which is 68 → 73. The intervals are identical under either scoping.)

What closes it is a new fixture that records `gasleft()` in its `receive()`, so one test measures the forwarded allowance directly (`135,000 + 2,300` stipend, measured 137,243) instead of inferring it from how much work a recipient can afford. That test also needs **no recalibration at the fork** — it measures the allowance, not gas consumed under a particular SSTORE price.

## Comment corrections

- **The sizing derivation was wrong on both forks.** Amsterdam ships EIP-8038 alongside EIP-8037, and 8038's `STORAGE_WRITE` (2,800 → 10,000) is charged in the frame doing the write — so it draws on this budget, which the previous text excluded as caller-side. A fresh cold slot goes `2,100 + 20,000 = 22,100` today to `2,100 + 10,000 + 97,920 = 110,020`; preserving a legacy one-slot wallet costs `35,000 − 22,100 + 110,020 = **122,920**`, leaving 135,000 with **~12k** of headroom. The cap was sized as `35,000 + 97,920` rounded up, before 8038 was accounted for — it still clears the corrected requirement, but by half the margin previously believed.
- **The compatibility claim is gone, not narrowed.** Three successive counterexamples falsified three successive narrowings, so the comment now states what the cap was sized for plus an explicit non-guarantee. The three: forwarding value to a never-used address costs ~36,600 today but ≥183,600 after (account creation, zero storage slots); one fresh cold SSTORE + one cold existing write + ~75 warm `EXTCODESIZE` reads costs ~34,600 today and ~137,120 after (8038 bills `EXTCODESIZE` as two warm accesses; an earlier draft of this line said 34,900/137,420, which carried 300 gas of unlisted overhead — the delta, and so the conclusion, is unchanged); and one fresh slot plus a single existing slot rewritten as `x→y→x→z→x→w` costs 33,100 today and 142,520 after, because `STORAGE_WRITE` is charged on every departure from a slot's transaction-start value while the restoring credit goes only to the transaction refund counter, never back to the frame's `gas_left`.
- **The fixture's recalibration note was off by one.** A first receive creates `numSlots + 1` fresh slots — the counter slot is fresh too. The old note's "`4 -> 1`" is a two-slot wallet (220,040 gas post-fork). Under a profile whose transaction gas limit leaves EIP-8037's state-gas reservoir empty, following it would have turned the must-pass tests red at exactly the fork it was written for; under the repo's current `gas_limit = u64::MAX` they would instead stay *green for the wrong reason*, because the reservoir absorbs the state charges before the callee frame sees them. The note now records both — the count rule and the fact that counts alone are not sufficient without such a profile.

## Test changes

- Bounds raised `4 -> 5` / `7 -> 6`. `N=4` needed a 108,992 operand — below the 122,920 requirement — so the old lower bound did not demonstrate the guarantee. `N=5` (131,235) clears it; `N=6` (153,478) still fails and brackets the cap more tightly.
- The fee test's capped leg was a *warm-counter second* receive costing 89,392, below what a single fresh slot costs post-fork. Separate wallets for `to` and `destOwner` make it a first receive. Its `gasLimit` is now derived from `getMessageMinGasLimit(0)`, decoupling it from `GAS_RESERVE` — the hardcoded `1_000_000` made it fail at `GAS_RESERVE >= 884,353` with a symptom pointing at the send cap.
- New coverage for the self-claim path with a *cap-sensitive* receiver, on both the prohibited-invocation and `gasleft()` arms. Three pre-existing tests already reached that path (`transactBy(Alice)` with `destOwner = Alice`), but all with EOA `destOwner`s, so none of them put the capped send's budget under any pressure. Also new: the capped send under a live `QuotaManager` — every prior test that drove a storage-creating recipient through that send inherited `getQuotaManager() == address(0)`.
- **The relayer fee payout was entirely unpinned.** `Bridge.sol`'s `msg.sender` send is a separate capped send from the destOwner refund, and every test paid it to an EOA, where the 2,300 stipend alone suffices — setting that send's gas operand to zero left the whole suite green. A new test makes a storage-creating contract the relayer with an EOA `destOwner`, so the payout is the only leg needing the budget.
- The recall test gained the file's value-conservation idiom — a 1-wei leak to a third account previously passed it — and the one negative test was renamed to the `RevertWhen_` form `packages/protocol/CLAUDE.md` specifies.

Each new or changed test was mutation-checked against the full 164-test shared suite. Dropping `stats.processedByRelayer &&` from the payout guard fails exactly the two self-claim tests. Collapsing the invocation-budget ternary to its relayer branch fails the invocation test (alongside three pre-existing tests that already caught it) — the weaker `gasleft() -> 1` mutation cited earlier did not establish this, and the test was strengthened in review to make it do so. Setting the relayer fee payout's gas operand to zero fails exactly the new relayer test. The new quota test is uniquely sensitive to a send-budget regression conditioned on a wired QuotaManager; it is *not* uniquely sensitive to mutations of the quota accounting itself, which pre-existing tests already cover.

## Changes from review

An independent review of this branch confirmed the headline claims (bytecode identity, both bisection endpoints, every gas figure against the EIP text) and produced four fixes, now pushed:

- **`220,140` → `220,040`** in the fixture note, and the recalibration guidance was strengthened. It said counts alone are "not sufficient"; the truth is stronger — post-fork the fixture's granularity is one fresh slot (110,020), coarser than the ~12,000 of headroom, so **no** count brackets the requirement. `N=0` lands *below* 122,920 and `N=1` far above the cap. The fixture is a pre-fork instrument; after the fork only the pin test constrains the constant.
- **The pin test's ±500 tolerance was the sole constraint on the interval**, and ~9× wider than the noise it absorbs (frame entry is 57 gas). It was also centred on 137,300, which the measurement can never reach, so half the band was dead. Now a named `_EXPECTED_SEND_BUDGET` with an upper bound (the allowance can never exceed cap + stipend) and a −100 lower bound: 1,001 admissible values → 101.
- **`self_claim_with_invocation` did not pin the arm its docstring claimed.** Collapsing the ternary left it green, because `gasLimit = 1_000_000` gave the relayer branch 193,344 — well above the 131,235 needed. `gasLimit` is now derived as `getMessageMinGasLimit(0) + 1`, so the alternate arm gets one gas plus the 2,300 stipend. An earlier revision used `+ 120_000`, which works today but not after the fork: the receive's six fresh-slot state charges then come from the transaction's state-gas reservoir, leaving only ~73.5k visible to the frame, which a 120,000 arm would satisfy — the mutation would have stopped being caught at exactly the fork this branch is about. One gas cannot complete the receive under any schedule.
- **`MessageReceiver_RecordingGasBudget` pre-initialises its slot**, keeping a slot creation out of the gas-capped frame so a too-low cap fails on the assertion rather than on `ETH_TRANSFER_FAILED`, which would report the wrong cause. (An earlier revision priced that update at 12,100 — the cold existing-slot figure. The constructor runs in the same transaction as the call under test, so the slot is warm and already dirty when the send arrives and the update is billed at `WARM_ACCESS`.)

Plus the relayer-payout coverage above, and magnitude corrections in the `Bridge.sol` comment: `EXTCODESIZE` is +100 warm (+500 cold), needing ~75 to matter, so it no longer leads the list; the per-departure rule is distinguished from per-write (`x→y→z` pays the write once); and the closing hedge now names EIP-8037 too, since it is also in Review and its 1,530 gas/byte feeds every figure.

## Deliberately excluded

- **`GAS_OVERHEAD` recalibration.** The fee snapshot precedes both capped sends, so the widened burn is unreimbursed. Real recalibration needs production `MessageProcessed` data.
- **A self-claim/self-recall escape hatch for the cap.** A ~3-line semantic change; needs auditor sign-off, so it belongs in its own PR.
- **The pre-existing quota brick** — an Ether claim whose `value + fee` exceeds the QuotaManager ceiling reverts before the status write, wedging the message at NEW. Pre-existing, deserves its own issue.
- **`retryMessage`'s capped send appears to be dead code.** `Status.RETRIABLE` is assigned in exactly one place, inside the branch where `_unableToInvokeMessageCall` is false; that function is pure in the message and the `immutable` `signalService`, so it must return false again on retry. Only an upgrade swapping the `signalService` immutable between the two transactions could reach it. Pre-existing and out of scope here, but worth its own issue.

## Three notes for reviewers

1. **CI could silently invalidate this PR's headline claim.** `.github/workflows/protocol.yml` runs `pnpm fmt:sol` — a *writing* formatter — then auto-commits and force-pushes the result onto the PR branch. `.tool-versions` pins foundry v1.4.2, under which `Bridge.sol` is already clean; but if a runner ever resolves a newer foundry, that step would push a non-comment `Bridge.sol` reformat here. Worth a glance at the diff after any CI auto-commit.
2. The Protocol workflow is gated on `draft == false`, so the 164-test gate will not run in CI while this stays a draft. Every check on this PR currently reports `SKIPPED` except `gitleaks` — the suite this PR is about has never run in CI here.
3. **"No semantic change" is not "no deployment consequence."** The `Bridge` creation and runtime bytecode are byte-identical to `main` with `bytecodeHash=none`, but under the repo default (`ipfs`) they are *not*: NatSpec feeds solc's devdoc, which is hashed into the CBOR metadata trailer. [Proposal0022](https://github.com/taikoxyz/taiko-mono/pull/22083) authenticates deployed implementations with `cast codehash` against a local build, so its deploy and its verification must be done from the same commit relative to this merge.

Also note `lefthook.yml:10` invokes `pnpm -F protocol lint:sol`, which does not exist in `packages/protocol/package.json` (`fmt:sol` does). It is broken on `main` too, so every Solidity commit here used `--no-verify`.

### DIFF
--- a/packages/protocol/contracts/shared/bridge/Bridge.sol
+++ b/packages/protocol/contracts/shared/bridge/Bridge.sol
@@ -61,29 +61,28 @@ contract Bridge is EssentialResolverContract, IBridge {
     /// suggest. Glamsterdam reprices state on two axes and both reach this budget. EIP-8038
     /// (state access) raises ACCOUNT_WRITE 6,700 -> 9,000 and COLD_ACCOUNT_ACCESS 2,600 -> 3,000,
     /// which this contract pays before forwarding, but it also raises STORAGE_WRITE 2,800 ->
-    /// 10,000 in the frame doing the write. That one is net-metered per transaction — a slot pays
-    /// it once, when it first moves away from its transaction-start value — and adds 7,100 to a
-    /// cold existing-slot write, 7,200 to a warm one. EIP-8037 (state creation) meters a *fresh*
-    /// slot as 64 state bytes * 1,530 gas/byte = 97,920 of state gas; unless the transaction buys
-    /// gas beyond the 16.7M per-transaction execution cap (which no realistic claim transaction
-    /// does, leaving its state-gas reservoir empty), that charge also falls on the callee frame.
-    /// A fresh cold slot therefore goes from 2,100 + 20,000 = 22,100 today to
-    /// 2,100 + 10,000 + 97,920 = 110,020. A smart wallet writing one fresh slot on receive would
-    /// run out of gas under the previous 35,000 cap, and since a failed send reverts processing
-    /// its messages would become unclaimable. Preserving that wallet's budget costs
-    /// 35,000 - 22,100 + 110,020 = 122,920; the cap was sized as 35,000 + 97,920 rounded up,
-    /// before EIP-8038 was accounted for, and still clears that requirement with roughly 12,000
-    /// to spare.
-    /// This is NOT a general "fits before, fits after" guarantee. It holds only for a receive
-    /// path whose SSTOREs are its only repriced operations, creating at most one fresh slot and
-    /// changing at most one further existing slot. EIP-8038 reprices ordinary reads too —
-    /// EXTCODESIZE now bills two warm accesses rather than one — so a path doing one fresh cold
-    /// SSTORE, one cold existing write and ~75 warm EXTCODESIZE reads costs ~34,900 today and
-    /// ~137,420 after the fork: inside the old budget, outside this one, while still meeting the
-    /// storage conditions above. EIP-8037 likewise reprices account creation to 120 state bytes
-    /// = 183,600 gas, so forwarding value to a never-before-used address, or running CREATE,
-    /// fits today and does not after. EIP-8038's parameters are still under review; re-check
-    /// these figures before the fork.
+    /// 10,000 in the frame doing the write: +7,100 on a cold existing-slot write, +7,200 on a
+    /// warm one. EIP-8037 (state creation) meters a *fresh* slot as 64 state bytes * 1,530
+    /// gas/byte = 97,920 of state gas; unless the transaction buys gas beyond the 16.7M
+    /// per-transaction execution cap (which no realistic claim transaction does, leaving its
+    /// state-gas reservoir empty), that charge also falls on the callee frame. A fresh cold slot
+    /// therefore goes from 2,100 + 20,000 = 22,100 today to 2,100 + 10,000 + 97,920 = 110,020.
+    /// A smart wallet writing one fresh slot on receive would run out of gas under the previous
+    /// 35,000 cap, and since a failed send reverts processing its messages would become
+    /// unclaimable. Preserving that wallet's budget costs 35,000 - 22,100 + 110,020 = 122,920;
+    /// the cap was sized as 35,000 + 97,920 rounded up, before EIP-8038 was accounted for, and
+    /// still clears that requirement with roughly 12,000 to spare.
+    /// That margin is what this number buys, and it is NOT a general "fits before, fits after"
+    /// guarantee — do not read it as one. Glamsterdam reprices enough distinct operations that a
+    /// receive path within the old budget can exceed this one while looking modest: EXTCODESIZE
+    /// now bills two warm accesses rather than one; account creation costs 120 state bytes =
+    /// 183,600, so forwarding value to a never-used address or running CREATE fits today and not
+    /// after; and STORAGE_WRITE is charged per departure from a slot's transaction-start value,
+    /// its restore credited only to the transaction refund counter, so a slot rewritten several
+    /// times bills 10,000 each time while the frame's own gas never gets the credit back.
+    /// Recipients doing more than one fresh slot's worth of bookkeeping should be measured
+    /// against the new schedule rather than assumed safe. EIP-8038's parameters are still under
+    /// review; re-check these figures before the fork.
     // - EOA gas used is < 21000
     // - For Loopring smart wallet, gas used is about 23000
     // - For Argent smart wallet on Ethereum, gas used is about 24000
--- a/packages/protocol/contracts/shared/bridge/Bridge.sol
+++ b/packages/protocol/contracts/shared/bridge/Bridge.sol
@@ -61,23 +61,29 @@ contract Bridge is EssentialResolverContract, IBridge {
     /// suggest. Glamsterdam reprices state on two axes and both reach this budget. EIP-8038
     /// (state access) raises ACCOUNT_WRITE 6,700 -> 9,000 and COLD_ACCOUNT_ACCESS 2,600 -> 3,000,
     /// which this contract pays before forwarding, but it also raises STORAGE_WRITE 2,800 ->
-    /// 10,000, charged in the frame doing the write, so every value-changing SSTORE the callee
-    /// makes costs 7,200 more. EIP-8037 (state creation) meters a *fresh* slot as 64 state bytes
-    /// * 1,530 gas/byte = 97,920 of state gas; unless the transaction buys gas beyond the 16.7M
-    /// per-transaction execution cap (which no realistic claim transaction does, leaving its
-    /// state-gas reservoir empty), that charge also falls on the callee frame. A fresh cold slot
-    /// therefore goes from 2,100 + 20,000 = 22,100 today to 2,100 + 10,000 + 97,920 = 110,020.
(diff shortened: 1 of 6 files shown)


davidtaikocha:
Re-review of `f8bca8967059ebdb5b051fc63f1c806619af58d1`: the original EIP-8038 arithmetic and `_RevertWhen_` naming findings are fixed correctly. `Bridge.sol` remains executable-semantics-identical to base.

I still see three documentation issues before merge:

1. **The compatibility guarantee remains too broad** (`Bridge.sol:75-79`). It bounds fresh/existing storage changes, but not other callee-side EIP-8038 repricings. For example, one fresh cold SSTORE, one existing cold first write, and 75 unrolled warm `ADDRESS; EXTCODESIZE; POP` operations cost approximately `34,900` before Glamsterdam versus `137,420` afterward. That fits the old `35,000 + 2,300` allowance and exceeds the new `135,000 + 2,300`, while satisfying the stated storage/new-state conditions. I suggest removing the universal “wallets that fit before still fit after” claim, or limiting it to receive paths whose modeled SSTOREs are their only repriced callee operations.

2. **The future fixture recalibration will not work with the configured test gas limit** (`MessageReceiver_CreatingFreshStorageSlots.sol:10-17`, `foundry.toml:4`). The tests run with `gas_limit = u64::MAX`. Under EIP-8037, gas above the execution cap fills the transaction-global state-gas reservoir, and state charges draw from that reservoir before the callee frame. The reservoir would absorb the `97,920` state charges, so changing the fixture counts to `0`/`1` would not make the two-slot case bracket the 135k CALL operand. This needs a dedicated test profile/top-level gas limit that leaves the reservoir empty, or the count-only recalibration advice should be removed.

3. **The PR description is still stale.** It still states `112,920`, `~22k`, `200,040`, and the derived `3,928`/`21,637` margins. The corrected values are `122,920`, `~12k`, `220,140`, `13,928` below, and `11,637` above.

Minor wording: “every value-changing SSTORE ... costs 7,200 more” is not the EIP-8038 charging rule. `STORAGE_WRITE` applies when a slot first moves away from its transaction-start value; the cold existing-slot delta is `7,100`, and the warm delta is `7,200`.

Verification at the updated head:

- shared tests: **163/163 passed**
- shared bridge tests: **72/72 passed**
- recorded forwarded budget: **137,243**
- mechanical non-comment `Bridge.sol` diff: empty

Tests ran offline with Forge 1.7.1 / Solc 0.8.30. The repository pins Foundry 1.4.2, and protocol CI remains skipped while the PR is a draft.

References: [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), [EIP-8038](https://eips.ethereum.org/EIPS/eip-8038).

dantaik (approved):

PR #22083, opened 2026-08-31
feat(protocol): Proposal0023 upgrades the L1 and L2 bridges and ERC20 vaults

davidtaikocha:
## What

Proposal0023 ships two merged changes to production by upgrading four proxies, two per chain. **No contract logic changes** — under `contracts/` only the address constants in `LibL1Addrs`/`LibL2Addrs` move (see Review notes); the rest is deploy scripts, the DAO proposal, a runbook and tests.

- **Bridges** — the `_SEND_ETHER_GAS_LIMIT` 35,000 → 135,000 change from #22077.
- **ERC20 vaults** — the EIP-2612 `permit` and Permit2 send paths from #22093. Permit2 lives at `0x000000000022D473030F116dDEE9F6B43aC78BA3` on both chains.
- **The L1 `bridged_erc20`** — a pre-existing defect found on the way (below): the L1 resolver names a July 2024 `BridgedERC20` the live vault can no longer initialise, so first-time token deliveries to L1 revert today. One more L1 action registers a `BridgedERC20V2` built from `main`.

> **All eight contracts are deployed and verified**, each address checked on-chain before being written in: the bridge-side four on 2026-08-31 (redeployed 2026-09-01 after #22082), the two `ERC20Vault` and the two `BridgedERC20V2` implementations on 2026-09-02. (The plain `BridgedERC20` deployments `0xFcbc02A2…` and `0x3505a070…` from earlier that day lack EIP-2612 permit and are superseded; nothing references them.) `Proposal0023.action.md` carries the executable calldata, and `forge verify-bytecode` reports `Creation code matched with status full` for every implementation (runtime code too for all but the two bridge-side L2 runs, which hit a foundry chain-alias bug after the creation match).
>
> | | Address | |
> | --- | --- | --- |
> | L1 `Bridge` impl | `0xA15dca0A72da684f20e0FC708DECFb230a715462` | [codediff](https://codediff.taiko.xyz/?addr=0xd60247c6848B7Ca29eDdF63AA924E53dB6Ddd8EC&newimpl=0xA15dca0A72da684f20e0FC708DECFb230a715462&chainid=1) |
> | L1 `ERC20Vault` impl | `0x32E47c04E8c329E8c10062731448e7658aDEEB8e` | [codediff](https://codediff.taiko.xyz/?addr=0x996282cA11E5DEb6B5D122CC3B9A1FcAAD4415Ab&newimpl=0x32E47c04E8c329E8c10062731448e7658aDEEB8e&chainid=1) |
> | L1 `BridgedERC20V2` impl | `0x9ccB9eBa4335096c5B64f050C3c734632D497c3b` | `LibL1Addrs.BRIDGED_ERC20`; registered as `bridged_erc20` on the L1 resolver, not a proxy target |
> | L2 `Bridge` impl | `0xa200c2268d77737a8Fd2CA1698dA6eeab2a85CEb` | [codediff](https://codediff.taiko.xyz/?addr=0x1670000000000000000000000000000000000001&newimpl=0xa200c2268d77737a8Fd2CA1698dA6eeab2a85CEb&chainid=167000) |
> | L2 `ERC20Vault` impl | `0xa01d464ca3982DAa97B19fa7F8a232eB11A9DDb3` | [codediff](https://codediff.taiko.xyz/?addr=0x1670000000000000000000000000000000000002&newimpl=0xa01d464ca3982DAa97B19fa7F8a232eB11A9DDb3&chainid=167000) |
> | L2 `BridgedERC20V2` impl | `0xD6601cdea5857338EbdEE4CF38298aff43f01431` | `LibL2Addrs.BRIDGED_ERC20`; registered as `bridged_erc20` on the new L2 resolver, not a proxy target |
> | L2 `DefaultResolver` proxy | `0x2ea05A9CD06984Cf533a1829d8b0BE6289a43984` | `LibL2Addrs.SHARED_RESOLVER`; impl `0x8Af4669E3068Bae96b92cD73603f5D86beD07a9a` |
>
> **Read the L2 addresses on L2 only** — every one of them has, or collides with, unrelated code on L1 (the bridge-side ones with vault implementations retired in 2024, `0xa01d464c…` with a 170-byte proxy); the same deployer reached the same nonces on both chains.

The proposal executes **4 top-level L1 actions** (two `upgradeTo`, one `registerAddress`, plus the `sendMessage` `BuildProposal` appends) and **7 L2 actions**.

## L1 — both legs are like-for-like

The live implementations `0x1c94D798…` (bridge) and `0x024253C6…` (vault), both from Proposal0017, are a `MainnetBridge` and a `MainnetERC20Vault`, the L1-only subclasses #22058 deleted. Their transient slots were folded into the unified `Bridge`/`ERC20Vault` at **byte-identical constants** — the reentry constant `0xa5054f72…` is present verbatim in the live vault bytecode. Diffing each implementation's dependency tree from that commit to `main`:

- bridge: `Bridge.sol`, `EssentialContract.sol`, an unused `LibNames` constant → the only behavioural delta is the send cap;
- vault: `ERC20Vault.sol` + the new `IPermit2.sol` (#22093), `EssentialContract.sol`, the same `LibNames` constant → the only behavioural delta is the two permit entrypoints.

The new L1 vault carries the same `resolver` (`0x8Efa0156…`) and `quotaManager` (`0xBaCb003f…`) immutables the live proxy answers; `DeployERC20VaultUpgradeL1` read both back from the live proxy before broadcasting. The L1 shared resolver already carries the vault's `bridge` and `erc20_vault` entries; its `bridged_erc20` entry is the one stale name, corrected by the third L1 action (see "Also fixed" below). Explorers show `Bridge` / `ERC20Vault` instead of the `Mainnet*` names.

## L2 — the vault makes the same two-year jump, and needs one more contract

The L2 vault `0x1670…0002` runs the protocol 1.10.0 implementation `0xb96AbB41…` (commit `9345f14`), the same vintage as the L2 bridge, and resolves everything through the legacy `AddressManager` `0x1670…0006`, which only answers `getAddress(uint64,bytes32)`. `ERC20Vault` on `main` resolves through `IResolver.resolve(uint256,bytes32,bool)` at six sites — its `onlyFromNamed(B_BRIDGE)` guards, the `erc20_vault` counterpart on chain 1 for every delivery and send, the bridge it sends through, and `bridged_erc20` when it deploys a token — so the vault moves onto the resolver the first version of this proposal deployed. That adds three registrations (`erc20_vault` for chains 1 and 167000, `bridged_erc20` for 167000), and the chain-167000 `bridge` entry, registered "for symmetry" before, becomes load-bearing.

**`bridged_erc20` has to be built from `main`, not the legacy `0x98161D67…`, and it is the `BridgedERC20V2` flavour.** The new vault initialises each bridged token through the six-argument `IBridgedERC20Initializable.init` (selector `0x6c0db62b`); the legacy implementation's bytecode carries only the seven-argument, address-manager based init (`0xbb86ef93`) and no fallback, so the proxy constructor's delegatecall would revert and the bridge would park every first-time token delivery as `RETRIABLE`. Even with a matching initializer the legacy token would authorise `mint`/`burn` through `IAddressManager.getAddress`, which `DefaultResolver` does not have. `BridgedERC20` on `main` takes the vault as an immutable instead, and the registered implementation is deployed with the vault **proxy**, so tokens deployed after this proposal survive future vault upgrades too. It is `BridgedERC20V2` — same constructor, same six-argument `init`, plus EIP-2612 `permit` — because the July 2024 implementation has `permit` and the vault's new `sendTokenWithPermit` reverts `VAULT_PERMIT_NO_ALLOWANCE` on a token without it; plain `BridgedERC20` would have taken that away from every bridged token deployed from now on. Existing bridged tokens are unaffected: they resolve `erc20_vault` through the legacy registry, which still names the unchanged proxy.

Storage layout is slot-for-slot identical between `forge inspect ERC20Vault storage-layout` at `9345f14` and `ERC20Vault_Layout.sol` on `main` (the same `addressManager` → gap, `lastUnpausedAt` → unused pattern as the bridge). The `BridgeTransferOp` struct, all five events and the errors are identical too, so `sendToken`'s selector, the relayer, the indexer and the bridge UI are unaffected by the jump.

The L2 order is five registrations, the vault upgrade (a plain owner upgrade — the vault is not on the call stack), then the bridge self-upgrade **last**, so the batch makes no further call once the bridge's own code has been swapped under its frame. All seven execute in one transaction; the order inside it is defensive.

## Evidence

- **The fork rehearsal deploys nothing and executes the committed calldata.** It builds the actions from `Proposal0023`'s constants and the address libraries — the same `_buildAllActions` output `Proposal0023.action.md` pins — against the deployed implementations on both forks, and takes the L2 message by decoding the appended `sendMessage` calldata of that batch. `BuildProposal.sol` is untouched: `_buildAllActions` is private, so `Proposal0023Harness` reproduces its wrapping verbatim, and the action-file pin (generated by the real tooling) is what keeps the two from drifting. On L1 that is both upgrades, the `bridged_erc20` registration and the appended `sendMessage` from the DAO controller through the just-upgraded bridge, then a WETH `sendToken` through the new vault; a never-seen L2-canonical token delivered before the batch parks `RETRIABLE`, and retrying that same message after the batch lands `DONE`, minting from the new `BridgedERC20V2`, which is then sent back with `sendTokenWithPermit`. On L2, the seven-action batch is delivered through `processMessage` (executor and relayer cases, the latter pinned at a 4,160,064-gas invocation budget), then USDT is delivered through the existing bridged token, a never-seen token is delivered, deploys a `BridgedERC20V2` proxy from the registered implementation and is sent back through `sendTokenWithPermit` (a signed EIP-2612 permit, no approve), and bridged USDT is sent back through the burn path. Both legs pass against live mainnet state as of 2026-09-02.
- **Upgrade safety re-verified on 2026-09-03 from the exact deployed commits** (runbook, _Upgrade Safety_). Storage layouts regenerated with `forge inspect` at `b73608696` (the live L1 `MainnetBridge` / `MainnetERC20Vault`) are identical to the new implementations, entry for entry; at `9345f14` (the live L2 1.10.0 code) the only differences are `addressManager` becoming gap and the dropped `lastUnpausedAt` bytes, neither read by `main`. The live proxies' slots were read: `_initialized` 3 / 1 / 1 / 1, owners intact, `__paused` byte `1`, which is `_FALSE` in all three `EssentialContract` versions; no initializer runs and none is needed. `_CTX_SLOT` and `_REENTRY_SLOT` are byte-identical to the live L1 code, and `IBridgedERC20` is unchanged, so existing bridged tokens keep minting and burning. The new L2 code is the first on L2 to use transient storage, so an `eth_call` of `sendMessage` on the new L2 implementation through the live RPC was checked to revert `B_INVALID_CHAINID` (after the `nonReentrant` `tstore`), and Prague and Osaka builds of all four L2 contracts are byte-identical minus metadata. The `bridged_erc20` registrations upgrade no proxy: existing tokens keep the implementation in their own proxy, and the July 2024 `0x65666141…` and the new `0x9ccB9eBa…` are both `BridgedERC20V2` with the same external function set.
- Both dry runs, `MODE=l1dryrun` and `MODE=l2dryrun`, simulated against the deployment RPCs revert with `DryrunSucceeded()` as designed. `Controller.dryrun` is permissionless and always reverts, so the simulation is the whole check.
- **The one 1.10.0 → `main` change that touches messages already in flight is documented, audited and pinned.** `_unableToInvokeMessageCall` no longer exempts recipients without code (#20939, EIP-7702), so an L1→L2 message to an EOA carrying non-`onMessageInvocation` calldata is refunded to `destOwner` as `INVOCATION_PROHIBITED` instead of delivered to `to` — L1 made the same transition in Proposal0017. Every `MessageSent` of the L1 bridge through block 25,889,001 was audited (40,494 messages): 20 carry such calldata, 19 of them to contracts that both versions refuse alike, and the one code-less target was delivered long ago with `destOwner == to`. None is outstanding; the runbook says to re-run the query right before execution. The fork rehearsal now pins both sides on the live 1.10.0 bridge and the upgraded one, and the runbook's new "What else changes on L2" section lists every other observable difference.
- Unit tests pin the encoding of all ten actions, the revert guard for each missing address, the shape and size of the L2 message (2,052 bytes, from which the relayer budget above is derived), that the no-argument builders forward the deployed addresses (written as literals), and that `Proposal0023.action.md` matches what the code builds (a missing file now fails rather than skips).

## Also fixed: the L1 `bridged_erc20`

The L1 shared resolver answers `bridged_erc20` with `0x65666141…`, a July 2024 implementation with only the seven-argument init, while the live L1 vault has called the six-argument init since Proposal0017. Verified on an L1 fork: a first-time delivery to L1 of a token canonical on another chain lands `RETRIABLE` today. It is the same mismatch the L2 leg has to avoid, so it gets the same fix: L1 action 3 registers a `BridgedERC20V2(L1.ERC20_VAULT)` on the L1 resolver, which the DAO controller owns. Existing L1 bridged tokens are untouched. The fork rehearsal pins both sides: the delivery parks `RETRIABLE` before the batch and the retry lands `DONE` after it, so messages this defect has already parked become claimable by retrying them.

## Review notes

- Error declarations sit at the top of the contract, matching `Proposal0021.s.sol` and `BuildProposal.sol`, rather than `CLAUDE.md:45`'s "errors at end". Deliberate, not a slip.
- The deployment addresses travel as `L1Deployment`/`L2Deployment` structs with named fields, so the transposed-forward hazard the earlier tests guarded against can no longer compile silently.
- Per review, the addresses the resolvers name from this proposal on live in the address libraries, updated ahead of execution: `LibL2Addrs.SHARED_RESOLVER` is new, and `LibL1Addrs.BRIDGED_ERC20` / `LibL2Addrs.BRIDGED_ERC20` take the `BridgedERC20V2` values under their existing names, each with a comment naming the July 2024 address it replaces. Only the four proxy implementations remain `Proposal0023` constants. Nothing else in the repository reads either `BRIDGED_ERC20` constant, and the unit tests keep the deployed values as literals so a library edit cannot be mirrored silently.
- None of the four deploy scripts should be re-run; each would deploy a second set of contracts that the constants do not point at.

---

**Renumbered 0022 → 0023** per @dantaik's review: #21964 already claims Proposal0022. Verified 0023 is unused on `main` and unclaimed by any open PR.


packages/protocol/script/layer1/proposals/Proposal0023.s.sol
```diff
@@ -0,0 +1,113 @@
...
+// To print the proposal action data: `P=0022 pnpm proposal`
+// To dryrun the proposal on L1: `P=0022 pnpm proposal:dryrun:l1`
+// To dryrun the proposal on L2: `P=0022 pnpm proposal:dryrun:l2`
+/// @custom:security-contact security@taiko.xyz
+contract Proposal0022 is BuildProposal {
+    /// @dev Placeholder until `DeployBridgeUpgradeL1` has run on Ethereum mainnet.
+    address public constant BRIDGE_NEW_IMPL_L1 = address(0);
+
+    /// @dev Placeholders until `DeployBridgeUpgradeL2` has run on Taiko Alethia.
+    address public constant BRIDGE_NEW_IMPL_L2 = address(0);
```
dantaik: TODO here.

dantaik:
use 0023 instead, as 0022 is created here https://github.com/taikoxyz/taiko-mono/pull/21964

davidtaikocha:
> use 0023 instead, as 0022 is created here #21964

updated in https://github.com/taikoxyz/taiko-mono/pull/22083/changes/36dcab123607b4753167d177718fbabace2a39e2

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
davidtaikocha: deployed in https://github.com/taikoxyz/taiko-mono/pull/22083/changes/dbb6016b0521be0ce91c160061758f2409253a0d

davidtaikocha:
Went through the DeepSeek review point by point. Two are actionable and now fixed; one is a known tradeoff; the first one is incorrect and I want to record why, so nobody "fixes" a mock that is already right.

### The signal mock is correct — the review's premise is wrong

> `ISignalService.proveSignalReceived` returns `bool`, and `abi.encode(uint256(0))` decodes as `false`.

It returns `uint256`, not `bool` — on `main` and on the protocol 1.10.0 implementation actually deployed at `0x1670…0001`:

```solidity
function proveSignalReceived(uint64, address, bytes32, bytes calldata)
    external
    returns (uint256 numCacheOps_);
```

So `abi.encode(uint256(0))` is the correct encoding for `numCacheOps_ = 0`. There is also no truthiness check anywhere on the path. 1.10.0 consumes it as:

```solidity
try ISignalService(_signalService).proveSignalReceived(...) returns (uint256 numCacheOps) {
    numCacheOps_ = uint32(numCacheOps);
} catch { revert B_SIGNAL_NOT_RECEIVED(); }
```

Success is "did not revert". The suggested `abi.encode(true)` would encode `uint256(1)`, i.e. `numCacheOps = 1`, which 1.10.0 turns into an extra 20,000 gas of relayer refund — a silent semantic change, not a fix.

On selector-prefix matching: `vm.mockCall` matches calls whose calldata *starts with* the given bytes, so a bare selector is the documented idiom. Verified empirically rather than argued — deleting the mock makes both L2 tests fail immediately:

```
[FAIL: B_SIGNAL_NOT_RECEIVED()] test_l2_selfUpgradeThroughProcessMessage()
[FAIL: B_SIGNAL_NOT_RECEIVED()] test_l2_selfUpgradeThroughProcessMessage_byRelayer()
```

That proves the mock is matched and load-bearing. The unmutated tests assert `status == DONE` and `reason == INVOCATION_OK`; had the proof check not been satisfied, `processMessage` would have reverted and they could not pass. So "the critical L2 self-upgrade path is not convincingly exercised" is contradicted by the tests' own assertions.

### Committed calldata now has an automated check — fixed

This was the review's best point, and nothing covered it. Added in 2c100e2f7: a test parses the calldata out of the committed `Proposal0023.action.md` and compares it against `_buildAllActions()`. That needed `_buildAllActions` widened from `private` to `internal` — reconstructing the array in the test instead would restate the implementation, including the bridge message wrapping the L2 batch, which is the one part no other test covers.

Mutation-tested both directions: reverting `BRIDGE_NEW_IMPL_L1` to the superseded address without regenerating fails with `Proposal0023.action.md is stale`, and flipping one hex character in the committed file fails the same way.

### Stale address table in the PR description — fixed

Correct catch. The description still listed the pre-redeployment addresses; the files were updated but the description was not. Now updated.

### Fork tests not running in CI — known, and deliberate

Accurate, and already stated in the runbook. This repository has no fork tests and CI configures no RPC endpoints, so adding a fork job means putting RPC secrets into CI — a repository-level decision rather than something this PR should make unilaterally. Worth noting the direction of travel: this evidence used to be terminal output pasted into the runbook, and is now a test anyone can re-run with two environment variables.

dantaik:
- [x] @davidtaikocha  We shall also upgrade ERC20Vault after https://github.com/taikoxyz/taiko-mono/pull/22093 is merged.

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
```diff
@@ -0,0 +1,238 @@
...
+
+    /// @dev Deployed by `DeployBridgeUpgradeL2` on Taiko L2.
+    /// https://codediff.taiko.xyz/?addr=0x1670000000000000000000000000000000000001&newimpl=0xa200c2268d77737a8Fd2CA1698dA6eeab2a85CEb&chainid=167000
+    address public constant BRIDGE_NEW_IMPL_L2 = 0xa200c2268d77737a8Fd2CA1698dA6eeab2a85CEb;
+
+    /// @dev The new L2 resolver, an ERC1967 proxy over implementation
+    /// `0x8Af4669E3068Bae96b92cD73603f5D86beD07a9a`, owned by the DelegateController. It has no
+    /// predecessor to diff against: the legacy registry `0x1670…0006` is a different contract that
+    /// stays in place for the L2 NFT vaults.
+    address public constant L2_SHARED_RESOLVER = 0x2ea05A9CD06984Cf533a1829d8b0BE6289a43984;
```
dantaik: shall we add this constant to LibL2Addrs?

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
```diff
@@ -0,0 +1,238 @@
...
+
+    /// @dev Deployed by `DeployERC20VaultUpgradeL2` on Taiko L2.
+    /// https://codediff.taiko.xyz/?addr=0x1670000000000000000000000000000000000002&newimpl=0xa01d464ca3982DAa97B19fa7F8a232eB11A9DDb3&chainid=167000
+    address public constant ERC20_VAULT_NEW_IMPL_L2 = 0xa01d464ca3982DAa97B19fa7F8a232eB11A9DDb3;
+
+    /// @dev Deployed by `DeployBridgedERC20V2L2` on Taiko L2: a `BridgedERC20V2` (EIP-2612 `permit`
+    /// included) with the vault proxy as its `erc20Vault` immutable. It is registered as
+    /// `bridged_erc20` rather than set as a proxy target, so there is no predecessor to diff
+    /// against.
+    address public constant BRIDGED_ERC20_NEW_IMPL_L2 = 0xD6601cdea5857338EbdEE4CF38298aff43f01431;
```
dantaik: also add this constant to LibL2Addrs

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
```diff
@@ -0,0 +1,238 @@
...
+
+    /// @dev Deployed by `DeployERC20VaultUpgradeL1` on Ethereum mainnet.
+    /// https://codediff.taiko.xyz/?addr=0x996282cA11E5DEb6B5D122CC3B9A1FcAAD4415Ab&newimpl=0x32E47c04E8c329E8c10062731448e7658aDEEB8e&chainid=1
+    address public constant ERC20_VAULT_NEW_IMPL_L1 = 0x32E47c04E8c329E8c10062731448e7658aDEEB8e;
+
+    /// @dev Deployed by `DeployBridgedERC20V2L1` on Ethereum mainnet: a `BridgedERC20V2` (EIP-2612
+    /// `permit` included) with the vault proxy as its `erc20Vault` immutable, registered as
+    /// `bridged_erc20` on the L1 shared resolver in place of the July 2024 implementation
+    /// `0x65666141…` that the live vault can no longer initialise.
+    address public constant BRIDGED_ERC20_NEW_IMPL_L1 = 0x9ccB9eBa4335096c5B64f050C3c734632D497c3b;
```
dantaik: also add this constant to LibL1Addrs

dantaik: Actually update the value in the file, use the same var name.

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
dantaik: Actually update the value in the file, use the same var name.

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
```diff
@@ -0,0 +1,238 @@
...
+        // 0: Upgrade the mainnet bridge to the implementation carrying the EIP-8037 Ether send cap.
+        actions[0] = buildUpgradeAction(L1.BRIDGE, _d.bridgeImpl);
+
+        // 1: Upgrade the mainnet ERC20 vault to the implementation carrying EIP-2612 permit and
+        // Permit2 support (#22093). The new implementation is built with the same resolver and
+        // quota manager immutables the live one carries, and the L1 shared resolver already holds
+        // every name the vault reads, so no registration accompanies this upgrade.
+        actions[1] = buildUpgradeAction(L1.ERC20_VAULT, _d.erc20VaultImpl);
+
+        // 2: Point the L1 shared resolver's `bridged_erc20` at a `BridgedERC20V2` built from `main`.
```
dantaik, line 103: the comment is incorrect, it's about the bridged token.

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
```diff
@@ -0,0 +1,238 @@
...
+        // 0-4: Populate the new resolver before actions 5 and 6 make the implementations that read
+        // it live. The legacy L2 registry `0x1670…0006` predates IResolver and cannot serve any of
+        // these lookups, so a missing entry reverts the call that needs it. All seven actions
+        // execute in one transaction, so the order inside the batch is defensive; what matters is
+        // that the registrations and the upgrades ship in the same bundle.
+        //
+        // 0: `bridge` for chain 1 — what the new bridge implementation reads on every sendMessage
+        //    and processMessage. Its three resolver lookups (isDestChainEnabled,
+        //    _proveSignalReceived, _isSignalReceived) all pass the counterparty id, which on L2 is 1.
+        actions[0] = _registerAction(_d.sharedResolver, _L1_CHAIN_ID, LibNames.B_BRIDGE, L1.BRIDGE);
```
dantaik, line 160: why do we need to register the bridge address again?

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
```diff
@@ -0,0 +1,238 @@
...
+        //    new vault deploys, read on the first delivery of a canonical token it has not seen
+        //    before. This must be built from `main`: the new vault initialises the token through
+        //    the six-argument IBridgedERC20Initializable.init and the token authorises minting
+        //    through its `erc20Vault` immutable, while the legacy implementation `0x98161D67…`
+        //    registered on the old registry only implements the seven-argument, address-manager
+        //    based init. Registering the legacy one would make every first-time token delivery to
+        //    L2 revert. `BridgedERC20V2` rather than plain `BridgedERC20` so new bridged tokens keep
+        //    the EIP-2612 `permit` the legacy implementation has and `sendTokenWithPermit` needs.
+        actions[4] = _registerAction(
+            _d.sharedResolver, _L2_CHAIN_ID, LibNames.B_BRIDGED_ERC20, _d.bridgedErc20Impl
```
dantaik, line 187: I think all contract addresses have been wired, we only need to re-register the new L2 bridged erc20 impl, right?

dantaik (commented):
@davidtaikocha I think the L1 and L2 actions should match, no?
My understanding is we only need to do these on both L1 and L2:
- register a new BridgedERC20 impl
- upgrade Bridge
- upgrade ERC20Vault.

dantaik:
We should be really careful with this proposal, maybe we should upgrade Hoodi first. @davidtaikocha 

A lot of changes for the bridged erc20 token: https://codediff.taiko.xyz/?addr=0x65666141a541423606365123Ed280AB16a09A2e1&newimpl=0x9ccB9eBa4335096c5B64f050C3c734632D497c3b&chainid=1

We shall also compare the other upgrade contract code.

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
davidtaikocha: updated: https://github.com/taikoxyz/taiko-mono/pull/22083/changes/54a60306b458a35723c809409445d15a0f9c2c68

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
davidtaikocha: updated: https://github.com/taikoxyz/taiko-mono/commit/54a60306b458a35723c809409445d15a0f9c2c68

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
davidtaikocha: updated: https://github.com/taikoxyz/taiko-mono/commit/54a60306b458a35723c809409445d15a0f9c2c68

davidtaikocha:
> @davidtaikocha I think the L1 and L2 actions should match, no? My understanding is we only need to do these on both L1 and L2:
> 
> * register a new BridgedERC20 impl
> * upgrade Bridge
> * upgrade ERC20Vault.

The main reason for the mismatch is because the [L2 bridge](https://taikoscan.io/address/0x1670000000000000000000000000000000000001#readProxyContract) / ERC20Vault are too old (which still import the legacy `addressManager`), so we have to deploy the new `DefaultResolver` and do the registration on L2 first, before we upgrade those two on L2.

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
davidtaikocha: Since the bridge on L2 is too old and still uses the legacy `AddressManager`, so we have to deploy a new `DefaultResolver` on L2 and do the registrations again before we deploy a new Bridge which uses this new `DefaultResolver`.

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
davidtaikocha: same reason: https://github.com/taikoxyz/taiko-mono/pull/22083#discussion_r3915123291

packages/protocol/script/layer1/proposals/Proposal0023.s.sol
davidtaikocha: Emm i think it's about pointing to the new bridged token impl in comments already? or i missed anything.

davidtaikocha:
> We should be really careful with this proposal, maybe we should upgrade Hoodi first. @davidtaikocha
> 
> A lot of changes for the bridged erc20 token: https://codediff.taiko.xyz/?addr=0x65666141a541423606365123Ed280AB16a09A2e1&newimpl=0x9ccB9eBa4335096c5B64f050C3c734632D497c3b&chainid=1
> 
> We shall also compare the other upgrade contract code.

Yes, planning to do so once this PR gets reviewed.

dantaik:
@davidtaikocha — independent verification of the deployed addresses, constructor args, resolver wiring and tests, done from `main` @ `53776d6` against live L1/L2 state (2026-09-03). Nothing execution-breaking; a handful of items to fix or clarify at the end.

## What was verified (all clean)

For each of the eight new contracts I pulled the Etherscan/Taikoscan verified standard-JSON, diffed every source file against `main` (and the OpenZeppelin 4.9.6 packages pinned in `pnpm-lock.yaml`), recompiled with solc 0.8.30, and compared against the creation-tx input and the live runtime code.

| Chain | Address | Verified name | Sources vs `main` | Creation bytecode | Constructor args (read back on-chain) |
|---|---|---|---|---|---|
| L2 | `0x2ea05A9C…3984` | `ERC1967Proxy` | identical | full match incl. metadata | impl `0x8Af4669E…`, `init(0xfA06E15B…)` |
| L2 | `0x8Af4669E…7a9a` | `DefaultResolver` | identical | full match | — |
| L2 | `0xa200c226…5CEb` | `Bridge` | identical | full match | resolver `0x2ea05…`, signalService `0x1670…0005`, quotaManager `0`, pauser `0` |
| L2 | `0xa01d464c…DDb3` | `ERC20Vault` | identical | full match | resolver `0x2ea05…`, quotaManager `0` |
| L2 | `0xD6601cde…1431` | `BridgedERC20V2` | identical | full match | erc20Vault `0x1670…0002` |
| L1 | `0xA15dca0A…5462` | `Bridge` | identical | full match | `0x8Efa0156…`, `0x9e0a2496…`, `0xBaCb003f…`, `0x9CBeE534…` (== live `MainnetBridge` args) |
| L1 | `0x32E47c04…EB8e` | `ERC20Vault` | identical | full match | `0x8Efa0156…`, `0xBaCb003f…` (== live `MainnetERC20Vault` args) |
| L1 | `0x9ccB9eBa…7c3b` | `BridgedERC20V2` | identical | full match | erc20Vault `0x996282cA…` |

- Every constructor-arg address is a verified contract of the type the parameter name implies (`SignalService` proxies, `QuotaManager`, `GnosisSafeProxy` for the pauser, `DefaultResolver` proxies, the vault proxies, `DelegateController` proxy as resolver owner with `l2Bridge = 0x1670…0001`, `daoController = 0x75Ba…`, `l1ChainId = 1`).
- L2 `SHARED_RESOLVER`: owner is the DelegateController, impl slot `0x8Af4669E…`, zero entries today (no `AddressRegistered` events, every `resolve(…, true)` returns zero). The committed calldata registers exactly `(1,bridge)=0xd60247c6…`, `(167000,bridge)=0x1670…0001`, `(1,erc20_vault)=0x996282cA…`, `(167000,erc20_vault)=0x1670…0002`, `(167000,bridged_erc20)=0xD6601cde…`, each pointing at a verified proxy/impl of the right kind. Every name the new `Bridge` (`(1,bridge)`) and `ERC20Vault` (`(167000,bridge)`, `(1,erc20_vault)`, `(167000,bridged_erc20)`) can read is covered, and there are no conflicts with the immutables (bridged token's `erc20Vault` == registered `erc20_vault`; registered `bridge` == DelegateController's `l2Bridge`).
- L1 `bridged_erc20` defect confirmed: the runtime code of `0x65666141…` contains only the 7-arg init selector `0xbb86ef93`, not the 6-arg `0x6c0db62b` the live vault calls. The L1 resolver's owner is the DAO controller, so action 3 is authorized.
- L1 like-for-like claim confirmed: diffing the live `MainnetBridge`/`MainnetERC20Vault` source trees against `main` leaves only `Bridge.sol` (send cap + transient ctx), `EssentialContract.sol` (transient reentry lock) and an unused `LibNames` constant; both slot constants appear verbatim in the live L1 bytecode; L1 storage layouts identical. L2 layouts (1.10.0 recompiled with solc 0.8.27 vs `main`) differ only in `addressManager`→gap at slot 151 and the dropped `lastUnpausedAt`. Live slot reads match the runbook (`_initialized` 3/1/1/1, `__paused` byte 1).
- Transient storage works on Taiko: a raw `TLOAD` in `eth_call` executes, and `sendMessage` on `0xa200…` reverts `B_INVALID_CHAINID` after the `nonReentrant` `tstore`. Osaka vs Prague runtime bytecode is identical minus metadata for all four L2 contracts (for `ERC20Vault` the only extra diff is the embedded `ERC1967Proxy` metadata).
- Legacy L2 `AddressManager` still holds `(167000,erc20_vault)` and `(167000,bridge)` (genesis-populated, no events), so existing bridged tokens keep authorizing mint/burn.
- Tests: `Proposal0023.t.sol` 8/8 pass; `Proposal0023Fork.t.sol` 3/3 pass against live L1 and Taiko state (L2 block 10,907,691), including both `processMessage` paths of the self-upgrade.

## Please fix / clarify

1. **Unused resolver entry.** `Proposal0023.s.sol:171-176` — action 3 registers `(167000, erc20_vault)`, which neither the new `Bridge` nor the new `ERC20Vault` reads (all `erc20_vault` lookups use the counterparty chain id). Harmless, but the reviewer criterion was "every entry is read by one of the two new implementations": either drop it or keep the comment and explicitly call it out as intentionally unused in the runbook.
2. **Runbook deploy dates don't match the chain.** On-chain creation timestamps: the bridge-side four (`0x2ea05…`, `0x8Af4…`, `0xa200…`, `0xA15d…`) were created **2026-09-01 05:49 UTC**; the vault impls on 2026-09-02 09:08 UTC; the `BridgedERC20V2` impls on 2026-09-02 12:57 UTC (all by `0x56706f11…`). `Proposal0023.md:352` still says "Bridge side, run on 2026-08-31", while `:33-34` and `:469-470` say redeployed on 09-01. Since `DeployBridgeUpgradeL2` deploys the resolver proxy and the bridge impl in one run, the 09-01 re-run necessarily minted an abandoned first resolver proxy from 08-31 — please state which run produced `0x2ea05…` and that the earlier one is unused (ties in with `:405-406` "none should be re-run").
3. **Test comment overstates pinning.** `Proposal0023.t.sol:160-163` claims the expected addresses are literals not read from the address libraries, but the proxy and registered addresses in those assertions come from `LibL1Addrs`/`LibL2Addrs` (`L1.BRIDGE`, `L2.ERC20_VAULT`, …). Only the four implementation addresses and the three `DEPLOYED_*` constants are literals; a wrong edit to e.g. `LibL2Addrs.BRIDGE` is caught only by `test_actionFileMatchesTheBuiltCalldata`. Either make them literals as the comment says or reword the comment.
4. **Fork-test coverage gaps (worth a sentence in the test header, optionally a small addition).** `Proposal0023Fork.t.sol:193` and `:311` mock `proveSignalReceived` wholesale on both forks, so the upgraded bridges' proof path against the live signal services is not exercised (the resolver lookup that feeds it still runs). And `_rehearseL2Upgrade` (`:276`) never delivers a second governance message after the bridge swap; the DelegateController depends on the new transient `context()`, which the post-upgrade vault deliveries in the same test do exercise, so this is coverage rather than risk — a no-op `registerAddress` delivered after the batch would close it cheaply.
5. **Missing L1 precondition.** `Proposal0023.md:99-100` lists only "L2 bridge not paused". The L1 `sendMessage` (action 4) is `whenNotPaused` too, so a paused L1 bridge reverts the whole L1 batch (atomically, but the proposal fails). Add it.
6. **Stale hardcoded resolvers on `main`.** `contracts/layer1/mainnet/SharedResolver.sol:26` and `MainnetSharedResolver.sol:26` hardcode `0x65666141…` for `bridged_erc20`. Mainnet runs a `DefaultResolver` so they're not on the execution path, but they become wrong after this proposal — update to `0x9ccB9eBa…` (or delete them) in this PR or a follow-up.
7. **Deployment logs.** `deployments/mainnet-contract-logs-L1.md` / `-L2.md` are untouched (the L2 log still carries the "deploy and register BridgedERC20V2" todo). Fine if it's done in a follow-up like Proposal0019 (#21991), just flagging so it isn't lost.
8. **DeepSeek's "resolver has no admin so it cannot be upgraded" is incorrect** and shouldn't drive any change: `DefaultResolver` is UUPS with `_authorizeUpgrade onlyOwner`, and the owner is the DelegateController. The ERC1967 admin slot is irrelevant.
9. **Cosmetic.** Taikoscan doesn't recognise `0x2ea05…` as a proxy (`Proxy: 0`), so the read/write-as-proxy tabs are missing; submitting the proxy verification would help future readers.

---

davidtaikocha:
@dantaik — all nine handled. Items 1–6 landed on this branch via #22119 (merged); 7 is a separate draft; 8 and 9 need no change here.

| # | Disposition |
| --- | --- |
| 1 | Kept on purpose. The runbook's "only the names the migrated contracts read" sentence now names the exception (L2 action 4, the vault's own `erc20_vault`) and why it stays: the L1 resolver carries its own chain's `bridge`/`erc20_vault` the same way, it is harmless, and dropping it would only churn the committed calldata and the 2,052-byte / 4,160,064-gas pins. |
| 2 | Runbook dates corrected: the bridge-side four were deployed on 2026-09-01, and the creation-tx table now lists them (`0x02f4a79f…`, `0x59cb3c09…`, `0x80c14fa5…`, `0x552d1e5b…`). The 2026-08-31 run's L1 `Bridge` `0x8636d970…`, L2 `Bridge` `0x097BBBef…`, L2 `DefaultResolver` proxy `0x2dfef033…` (empty, DelegateController-owned) and implementation `0x4F750D13…` are unused and unreferenced; not narrated in the runbook. |
| 3 | Test comment reworded: the seven `DEPLOYED_*` literals are what the no-argument builders forward; proxies and pre-existing registration targets come from `LibL1Addrs`/`LibL2Addrs` and are pinned by the action-file test. |
| 4 | Fork-test header now states that `proveSignalReceived` is mocked on both forks and that the Hoodi rehearsal (#22118) exercised the live proof path through the upgraded bridges in both directions; the Hoodi L2 `SignalService` implementation (`0x22efa191…`) is the same build as mainnet's `0x18b27428…` (identical runtime bytecode apart from the immutables). The L2 rehearsal also delivers a second governance message after the batch (a no-op `registerAddress`) and asserts `Controller.ActionExecuted` on the DelegateController, so its `context()` read is exercised against the new transient-storage implementation. Both fork tests pass against live L1 and L2 (2026-09-03); mutation-checked. |
| 5 | Precondition added: neither bridge may be paused, since the appended L1 `sendMessage` (action 4) is `whenNotPaused` too. |
| 6 | `SharedResolver.sol` and `MainnetSharedResolver.sol` now return `0x9ccB9eBa…` for `(1, bridged_erc20)`. |
| 7 | Draft #22120 (base `main`) pre-writes the contract-log entries as they will read after execution, with `TBD` placeholders for the date and the two transaction hashes, plus a `note:` under each `shared_address_manager (sam)` and `shared_resolver` saying which contracts read it. |
| 8 | Agreed, no change. |
| 9 | Done: https://taikoscan.io/address/0x2ea05A9CD06984Cf533a1829d8b0BE6289a43984#code now recognises the proxy. |

dantaik (approved):

PR #22080, opened 2026-08-31
fix(bridge-ui): fix audit findings and unify the ETH/ERC20/ERC721/ERC1155 bridges

dantaik:
An in-depth audit of `packages/bridge-ui`, its fixes, and the refactoring that followed. Nothing outside `packages/bridge-ui` is touched.

The audit produced 99 raw findings from 12 parallel reviewers, de-duplicated to 64, each re-verified by two adversarial reviewers against the source. This branch started as the critical + high fixes; the medium/low PR ([#22081](https://github.com/taikoxyz/taiko-mono/pull/22081)) was merged into it, and everything since came from CI review rounds on this branch, three rounds of refactoring, and a whole-package human re-audit.

### Critical
- **Recall proof gate compared block numbers from two different chains** (`libs/proof/BridgeProver.ts`) — `getEncodedSignalProofForRecall` required `latestSyncedBlock >= bridgeTx.blockNumber`, a destination-chain height against a source-chain block. Whenever the source chain was ahead (the normal case for an L1→L2 bridge released on L1), releasing funds from a FAILED message threw `BlockNotSyncedError` forever. The FAILED signal lives on the destination chain at an unknown block, so there is nothing to gate on; the gate is removed and `getProof` still rejects while the signal is absent from synced state.

### High
- **Server-side NFT repository leaked one user's NFTs to another** (`MoralisNFTRepository.server.ts`) — the `/api/nft` singleton kept one `cursor`/`nfts`/`hasFetchedAll` for all requests and its address-change guard was dead code. Pagination state is now keyed per wallet+chain behind a bounded cache.
- **Valid pending transactions were deleted from local history on transient failures** (`Transactions/Status/Status.svelte`) — any polling-start failure removed the stored transaction. Removal now also requires the transaction to be past a 24h stale-age threshold.
- **USDT allowance-reset flow could never trigger** — detection keyed on symbol `tUSDT` while the reset UI gated on the mainnet USDT address, so mainnet USDT users with a partial allowance were stuck in an approve loop. Both sides now use a shared `tokenNeedsAllowanceReset`.
- **Escape closed the recipient dialog keeping unconfirmed edits** — Escape bypassed the validation Confirm enforces, and Cancel clobbered a configured destOwner. Escape now cancels, cancel restores both recipient and destOwner from an open-time snapshot, and the listener is removed on close.
[...]

PR #22094, opened 2026-09-01
fix(bridge-ui): fix audit findings and unify the ETH/ERC20/ERC721/ERC1155 bridges

dantaik:
Fixes for an in-depth audit of `packages/bridge-ui`, plus the refactoring that followed. Nothing outside `packages/bridge-ui` is touched.

Supersedes #22080, which carried the review history; the branch and commits are unchanged.

## Fixes

**Critical** — the recall proof gate compared a destination-chain synced height against a source-chain block number, so releasing funds from a FAILED message threw `BlockNotSyncedError` forever whenever the source chain was ahead (the normal case for an L1→L2 bridge released on L1).

**High**
- The server-side NFT repository served one user's NFTs to another: the `/api/nft` singleton shared one cursor across all requests. Pagination state is now keyed per wallet+chain behind a bounded cache.
- Valid pending transactions were deleted from local history on any transient polling failure.
- The USDT allowance-reset flow could never trigger — detection keyed on a symbol while the UI gated on an address.
- The recipient and destination-owner dialogs committed unconfirmed edits: Escape kept them, async validation had no generation guard, and Confirm keyed on a stale flag rather than on what the box held. The destination owner is the one address that can process a `gasLimit: 0` message.
- ERC20 amounts were formatted with `formatEther` in the details dialogs — 100 USDC shown as `0.0000000001`.

**Medium / low** — around fifty smaller defects. The recurring themes:
- **Stored XSS** through an unsanitized on-chain token symbol interpolated into a `{@html}` toast.
- **Message invariants** (`minGasLimit`, the fee/gas-limit pairing, the pause check) enforced in one place on every send path.
- **Strict amount parsing** replacing permissive `parseUnits`, which accepted hex, negatives and silent rounding.
- **Async races** — generation counters across the relayer search, NFT scan, balance reads, approval polling, the account watcher and both address dialogs.
- **Settings outliving their transfer** — a zero gas limit, a custom fee draft, a token's balance and a previous account's recipient could each ride into a transfer nobody set them for.
- **Transaction identity** keyed by message rather than transaction hash, with `MessageSent` lookups matched to the emitting transaction.
- **Error discrimination** — a revert, a timeout, an unreadable receipt and a truncated history are distinct; only a revert fails an action, and a storage failure can no longer be reported as a failed bridge.

## Refactoring

The UI supports ETH, ERC20, ERC721 and ERC1155; three rounds pulled out what they share.

- `Bridge.prepareSend` — one send preamble (wallet check, pause assertion, contract, gas-limit estimation, fee, destination owner) replacing four copies.
- `NFTBridge` — an abstract base for ERC721 and ERC1155, which were near-copies; 358 lines out of the subclasses, 233 into the base.
- `dialogTransactionFlow` — one send → await → classify → toast for claim, retry and release.

## Tests

**621 tests across 82 files**, up from 190 on `main`. Component tests mount real Svelte components; `vitest.config.ts` gained the aliases that needed. Every regression test was checked by reverting the fix and confirming it fails for the right reason.

## Validation

`pnpm lint`, `pnpm svelte:check` (0 errors, 0 warnings), `SKIP_ENV_VALIDATION=true pnpm test:unit --run`, `SKIP_ENV_VALIDATION=true pnpm build`, and `typos` from the repo root — all pass.

## Notes

- One known gap, deliberately not fixed here: relayer rows whose on-chain enhancement fails are dropped without telling the user. The counting that fixes the silence is in #22087, stacked on this branch alongside the RPC batching that reduces the failures.
- `.github/workflows/repo--deepseek-review.yml` keys its concurrency group on the PR number for both `pull_request` and `issue_comment` with `cancel-in-progress`, so any comment cancels the review of the push it comments on. Appending `-${{ github.event_name }}` to the group fixes it; that file is out of scope here.

https://claude.ai/code/session_017SSZC4UvEbDKAYZXjt3qfA

---
_Generated by [Claude Code](https://claude.ai/code/session_017SSZC4UvEbDKAYZXjt3qfA)_


dantaik:
Thanks — the P1 on the fee dialog is correct, and it is a regression I introduced in `832ba57`. Fixed in `31cb87e`. The other two do not hold; evidence below.

## Fixed

**Commit edits to an existing custom fee** — right in every detail, including the mechanism. `confirmChanges` does not always go through the update reactive: when the committed method is already `CUSTOM` it calls `updateProcessingFee` directly, because assigning `$processingFeeMethod` the value it already holds re-runs nothing. `closeModal()` runs *after* that call, so my `modalOpen` guard saw the dialog still open and dropped the edit — reopening a committed custom fee, typing a new amount and confirming kept the old fee while the dialog reported success.

That is worse than the leak the guard was added to close, and on the primary path rather than an edge case. Confirm now says so explicitly with a `committing` argument that the reactive never passes; focus is skipped on that path since the dialog is closing. The new test fails against the previous version with the confirmed `0.01` reverting to `0.002`.

## Not changing

**Preserve zero when parsing zero-decimal amounts** — `BigInt('')` is `0n`, not a throw. An empty string is valid input to `StringToBigInt` and converts to zero, the same way `Number('')` is `0`, so `parseDecimalAmount('0', 0)` already returns `{ ok: true, value: 0n }` and `inputAmount` never exits through an exception.

```
$ node -e "console.log(BigInt(''))"
0n
```

Both levels are pinned by tests that pass on this head: `parseDecimalAmount.test.ts` covers `'0'` at 0 and 18 decimals, `'0.0'`, and `'000'`; `TokenAmountInput.component.test.ts` types `5` then `0` and asserts `$enteredAmount` is `0n` with no error — which is exactly the ERC1155 sequence described, and it would fail on `enteredAmount` still holding `5` if the parse threw.

**Resume monitoring after a receipt wait times out** — the dialog is not stuck: `closeDialog` sets `dialogOpen = false` *before* calling `reset()`, so the early return in `reset` cannot hold it open, and the close button, Escape and outside-click all work while a transaction is pending. What survives is narrower and deliberate: reopening for that same message will not offer a second Claim, because a claim that is on chain and unresolved must not be sent twice. The outcome is not unobserved either — the transactions list polls the message status independently and shows it, and the contract rejects processing an already-processed message regardless. A watcher inside the dialog would duplicate the poller that already exists.

---

Validation on `31cb87e`: `pnpm lint`, `pnpm svelte:check` (0 errors, 0 warnings), 622 tests across 82 files passing, `pnpm build`, `typos` from the repo root. Pushing after this comment.

---

dantaik:
Three of these are real and are fixed in `17cf3f0` — all three in code I changed earlier today, so this round was worth having.

## Fixed

**`NFTCardGrid` deselects by `tokenId` only** — correct, and it is a half-finished change of mine: I moved the membership test to `isSameNFT` and left the removal filter comparing `tokenId` alone. Deselecting one of two same-numbered tokens from different collections cleared both. The filter now removes by the identity the branch above matched on.

**`DestOwner` does not prefill the committed owner on reopen** — also correct, and worse than "pre-existing". `Recipient.svelte` and `NFTBridge.resetForm` both write `$destOwnerAddress`, so the dialog is routinely opened on a value its own input never saw. That was cosmetic until *my* change made Confirm require a validation matching the draft — after which the box opened empty and Confirm was dead until the same address was retyped. It now seeds from the store and asks for the validation explicitly, since `setAddress` deliberately does not announce itself.

**`DestOwner` does not invalidate `validated` when the destination chain changes** — right again, and the same shape: `canConfirmDestOwner` rejects a classification made for another chain, but nothing superseded it, so Confirm went dead with nothing on screen explaining why. The address is re-checked against the new chain now, as the recipient dialog does.

Both new DestOwner tests fail against the previous version — an empty box, and no re-check on the chain switch.

## Not changing

**`ETHBridge._prepareTransaction` references a removed `processingFee`** — the suspicion in the finding is the right one to have had. It is destructured in both places it is used:

```
$ grep -n "processingFee\|value = senderAmount" src/libs/bridge/ETHBridge.ts
37:      value = senderAmount;
73:    const { value: callValue, fee: processingFee } = message;
75:    const value = callValue + processingFee;
88:    const { value: callValue, fee: processingFee } = message;
90:    const value = callValue + processingFee;
```

Line 37 is the zero-fee branch and uses no fee at all. `svelte:check` reports 0 errors, which an undeclared variable would not survive.

**`_transformTransaction` can throw on a decimal `amount`** — the finding describes the guard working as designed. A row whose numeric fields cannot be represented is skipped rather than taking the whole page down; that was a deliberate earlier fix, and the relayer's schema does not produce decimals for these fields.

## Worth doing, not here

The remaining warnings are fair but are UX additions rather than defects, and each widens this diff: a toast on a failed "load more" in `ScannedImport`, a `catch` around `updateTransactions`, clearing the search box in `Relayer.reset()`, and a timeout on the pause reads. Happy to take any of them if you would rather they land in this PR.

---

Validation on `17cf3f0`: `pnpm lint`, `pnpm svelte:check` (0 errors, 0 warnings), 624 tests across 82 files passing, `pnpm build`, `typos` from the repo root. Pushing after this comment.

---

dantaik:
The critical and one warning are real; fixed in `2f69399`. The RetryDialog warning does not hold.

## Fixed

**Recipient dialog accepts a smart contract as destination owner** — accurate. `validateDestOwner` commits `$destOwnerAddress` with the `isSmartContract` call commented out, while `DestOwner.svelte` refuses a contract and says so on screen. Two paths to the same store, disagreeing — and the embedded field is the one shown *for a contract recipient*, which is exactly when the destination owner matters, since it is the fallback processor for a message the recipient cannot claim.

One correction to the framing: this is pre-existing rather than something this PR introduced — the same block is commented out on `main`. What the PR changed is that hardening the standalone dialog made the disagreement matter, which is how it surfaced.

The check is live again, with the generation guard its now-async form needs, and `canConfirmRecipient` blocks on it so Confirm agrees with the store. Covered by a predicate case and a component test that types a contract into the destination-owner field and asserts Confirm stays disabled with nothing committed.

**`awaitDialogTransaction` treats arbitrary errors as `pending`** — right, and it contradicted this branch's own reasoning. `ConfirmationStep` names the two errors that mean the wait gave up and carries a comment warning against inferring them from "not a failure"; the shared dialog helper held the weaker rule, so an unexpected error was dressed up as "may still confirm" and left the action disabled with the real error buried. It names the same two now. The test that pinned the old default is updated to the corrected rule rather than deleted.

While fixing the first one I found the Recipient component tests mocked `isSmartContract` with a blanket `true`, which classified the destination owner as a contract as well. They are address-aware now, which is what production does — worth noting because that mock was hiding the difference between the two fields.

## Not changing

**RetryDialog clears `retrying` after a successful send** — it does not. `handleClaimClick` delegates to `claimWithQuotaGuard`, whose only `setClaiming(false)` after the quota check is inside `catch`, immediately before rethrowing:

```ts
  try {
    await claim();
  } catch (error) {
    // claim() normally reports through its own events; a throw escaping it would
    // otherwise leave the dialog's claiming flag stuck on
    setClaiming(false);
    throw error;
  }
```

That is the ReleaseDialog pattern the finding asks for, already shared by both dialogs. The other two `retrying = false` assignments in the file are in the error handler and in `reset()`.

**Relayer decimal `amount`/`fee`** — asked and answered on the previous round: a row whose numeric fields cannot be represented is skipped rather than taking the page down, which was a deliberate earlier fix.

---

Validation on `2f69399`: `pnpm lint`, `pnpm svelte:check` (0 errors, 0 warnings), 626 tests across 82 files passing, `pnpm build`, `typos` from the repo root. Pushing after this comment.

---

dantaik:
Confirmed and fixed in `51ce2d2`. Every mechanical claim in this one holds: the status is emitted first, the `DONE` check sits after the receipt read, and the catch that keeps the interval alive is reached before it — so a source RPC that cannot serve the receipt leaves a finalised message being re-read, re-emitted and re-polled forever.

Reproduced before fixing. Against the previous ordering the regression test sees **four `DONE` emissions across three intervals** instead of one, with the poller still running.

The reasoning about the receipt is also right, and it is what makes the fix safe: it only fills in `bridgeTx.blockNumber`, which is needed by the proof paths, and a processed message has none of those left. So the terminal status is acted on as soon as it is read, before the receipt is attempted — which also saves the RPC call entirely for completed rows.

I took the second half of the suggestion too: the receipt read now has its own `try`/`catch`, matching the processable read at the top of the same function, whose comment already says one failing must not skip the other. That way a receipt failure cannot skip anything else in the tick even for the non-terminal statuses.

One correction to the scope, which does not change the verdict: "permanently retain an interval, emitter, and RPC traffic" is stronger than what happens. `destroy` calls `stopPolling` once the last listener is removed, so the poller is torn down with the transaction row that owns it — the leak lasts as long as the transactions page stays open, not the life of the process. Still wrong, still worth fixing; just bounded.

---

Validation on `51ce2d2`: `pnpm lint`, `pnpm svelte:check` (0 errors, 0 warnings), 627 tests across 82 files passing, `pnpm build`, `typos` from the repo root. Pushing after this comment.

---

dantaik:
The critical is correct and is mine — a gap I opened in `2f69399` an hour ago. Fixed in `206b003`, along with one of the warnings. The other warning does not hold.

## Fixed

**Destination-owner validation not invalidated on a destination-chain change** — right, and it is the direct consequence of making `validateDestOwner` async in the previous commit without wiring it into the handler that already superseded the recipient's. The hazard the finding names is exactly the one the check exists for: an address that is an EOA on the old chain and a contract on the new one.

The handler now bumps the owner's generation, clears `validatedDestOwner` and `destOwnerIsSmartContract`, and re-checks the address on screen against the new chain.

Two things fell out of writing the test, both worth recording:

- **It needed a first-run guard.** Clearing on the mount pass wiped the owner just seeded from the store — a value that passed its own validation when it was committed — which broke the existing "already validated before a remount" test. That test earning its keep is the reason the regression did not ship.
- **My first version of the new test was wrong.** It mocked the classification per-address only, so the re-check after the switch simply re-approved the owner and the test failed against the *fixed* code. Modelling the real case (EOA on the old chain, contract on the new) is what makes it meaningful — and it fails against the previous version with the stale owner committed.

**NFT ownership read leaves a stale `allApproved`** — confirmed. `checkOwnershipOfNFT` sat outside any `try`/`catch`, so a failed read threw with a previous token's `allApproved = true` still set, leaving Bridge enabled for an NFT whose approval state could not be read at all. It clears the flag before rethrowing, the way the not-owner branch immediately below it already does.

## Not changing

**`$computingBalance` left raised after `reset()`** — the success path does clear it. `reset()` ends with:

```ts
      if (!published) return;
      log('tokenBalance', $tokenBalance);
    } else {
      balance = '0.00';
    }
    $computingBalance = false;
  };
```

The winning path falls through the `if` block to that last line. Only the *superseded* path returns early without clearing, which is deliberate and is what the comment above it describes: clearing there would stop the spinner while the read that replaced it is still in flight. The account-change branch clears it explicitly because it returns from inside its own guard rather than falling through.

**`Relayer.reset()` does not clear the bound search address** — fair, and unchanged from the previous round: a UX addition rather than a defect, and the search stays disabled until revalidation. Happy to take it if you would rather it land here.

---

Validation on `206b003`: `pnpm lint`, `pnpm svelte:check` (0 errors, 0 warnings), 629 tests across 82 files passing, `pnpm build`, `typos` from the repo root. Pushing after this comment.

---

dantaik:
Both the critical and the first warning are correct. Fixed in `14db1b9` — and since this is the third round landing on the same seam, I fixed the shape rather than another path.

## The root cause

`Recipient.svelte` carries two generation counters but `supersedePendingValidation()` bumped only the recipient's — and that helper is what *every* invalidation path calls: cancel, destroy, `clearRecipient`, both draft watchers, and the destination-chain change I added last round. All six left a pending destination-owner lookup free to resolve afterwards and write the shared store.

That also explains the pattern in these reviews: I made `validateDestOwner` async two commits ago and have been retrofitting invalidation one site at a time. The helper now bumps both counters, which is what makes the *next* invalidation path correct without having to remember there are two — the thing my previous two fixes each got wrong in turn.

**Trusting `$destOwnerAddress` across a chain change on remount** — same class, fixed the same way. The validated owner now carries the chain it was classified on, as the validated recipient already did, and `canConfirmRecipient` compares it. Without that, a destination change while the component was unmounted left the owner restored from the store looking validated, skipping the contract refusal on exactly the field that decides who can process a `gasLimit: 0` message.

Both verified by reverting: the cancel test fails with the cancelled owner committed, the predicate case with a chain-mismatched owner accepted.

Worth recording that my first version of the cancel test passed for the wrong reason — I mounted the component without opening the dialog, so `modalOpen` was false and Escape did nothing. It only became a real test once it opened the dialog the way the trigger does.

## Not changing

**The zero address is accepted until send time** — `messageInvariants` rejects it before any transaction is signed, so nothing reaches the chain. Adding it to the dialog predicates would be a second place to state the same rule, and the value of catching it a step earlier is a slightly better error message rather than a different outcome. I would rather keep one owner for that invariant than two that can drift — which is the same argument as the fix above.

---

Validation on `14db1b9`: `pnpm lint`, `pnpm svelte:check` (0 errors, 0 warnings), 631 tests across 82 files passing, `pnpm build`, `typos` from the repo root. Pushing after this comment.

---

dantaik:
Superseded by #22097 — same branch (`claude/bridge-ui-code-audit-0ljtmq`), same commits, shorter description. Closed at the author's request.

The review history stays here. What was found and fixed on this PR, for anyone following the thread there:

- **Codex** — the `modalOpen` guard I added for the fee dialog dropped the edit on the path that matters (`confirmChanges` calls the updater directly when the method is unchanged), fixed in `31cb87e`.
- **Round 1** — the NFT deselect filter left half-converted to `isSameNFT`, and the destination-owner dialog opening empty on a value committed elsewhere, fixed in `17cf3f0`.
- **Round 2** — a contract could be committed as destination owner through the recipient dialog while the standalone one refused it (pre-existing; the check was commented out on `main`), and `awaitDialogTransaction` treating every unrecognised rejection as "still pending", fixed in `2f69399`.
- **Round 3** — a finalised message polled forever when its source receipt could not be read, fixed in `51ce2d2`; then the destination-owner classification surviving a chain change, fixed in `206b003`.
- **Round 4** — the root cause behind the last two: `supersedePendingValidation` bumped one of the two generation counters while every invalidation path called it, fixed in `14db1b9`.

Declined with reasons, also recorded here: the `BigInt('')` claim (it returns `0n`), the RetryDialog flag (`claimWithQuotaGuard` already clears only on an escaping throw), `$computingBalance` after `reset()` (the success path falls through to it), and the `ETHBridge` "undeclared variable" (destructured at both use sites).

---

